PR #465, opened 2020-02-15
Encapsulate fatomic_fetch_add and ifloordiv as functions

k-ye:
Issue #396 

As discussed before, we don't have to repeatedly generate these instructions.

### DIFF
--- a/taichi/backends/codegen_metal.cpp
+++ b/taichi/backends/codegen_metal.cpp
@@ -9,6 +9,10 @@ namespace {
 constexpr char kKernelThreadIdName[] = "utid_";  // 'u' for unsigned
 constexpr char kGlobalTmpsBufferName[] = "global_tmps_addr";
 constexpr char kArgsContextName[] = "args_ctx_";
+// Atomic fetch add for float32
+constexpr char kFAtomicFetchAddFunc[] = "fatomic_fetch_add";
+// Floor div for int32
+constexpr char kIFloordivFunc[] = "ifloordiv";
 
 class MetalKernelCodegen : public IRVisitor {
  public:
@@ -236,15 +240,8 @@ class MetalKernelCodegen : public IRVisitor {
     const auto bin_name = bin->raw_name();
     if (bin->op_type == BinaryOpType::floordiv) {
       if (is_integral(bin->element_type())) {
-        const auto intm = fmt::format("{}_intermediate_", bin_name);
-        emit("const {} {} = ({} / {});", dt_name, intm, lhs_name, rhs_name);
-        // Should we construct an AST for this?
-        const auto expr_str = fmt::format(
-            "(({lhs} * {rhs} < 0) && ({rhs} * {intm} != {lhs})) ? ({intm} - 1) "
-            ": {intm}",
-            fmt::arg("lhs", lhs_name), fmt::arg("rhs", rhs_name),
-            fmt::arg("intm", intm));
-        emit("const {} {} = ({});", dt_name, bin_name, expr_str);
+        emit("const {} {} = {}({}, {});", dt_name, bin_name, kIFloordivFunc,
+             lhs_name, rhs_name);
       } else {
         emit("const {} {} = floor({} / {});", dt_name, bin_name, lhs_name,
              rhs_name);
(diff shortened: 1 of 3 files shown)


taichi/backends/codegen_metal.cpp
```diff
@@ -410,6 +388,59 @@ class MetalKernelCodegen : public IRVisitor {
...
+        "(({lhs} * {rhs} < 0) && ({rhs} * {intm} != {lhs})) ? ({intm} - 1) "
+        ": {intm}",
+        fmt::arg("lhs", kLhs), fmt::arg("rhs", kRhs), fmt::arg("intm", kIntm));
+    emit("return ({});", expr_str);
+    pop_indent();
+    emit("}}");
+    emit("");
+  }
+
+  void gen_fatomic_fetch_add_func() {
```
yuanming-hu: This does look cleaner than the original approach!

Will that make things even cleaner, if we directly emit a multiline string of the function `fatomic_fetch_add`, or even put all these functions into a "header" and then "include" it? I.e. directly include the emitted shader code by `gen_fatomic_fetch_add_func` instead of emitting and formatting the lines.

I've got good experience when taking this approach in the legacy backend. E.g. here I implement CPU atomics using CAS
https://github.com/taichi-dev/taichi/blob/1cc91f52acebb2bb0e39a49baecbb459b60389d7/taichi/arithmetics.h#L1045 and this file is then directly included as a header. Not sure if this is a good idea on the Metal backend though.

k-ye: Thanks for the suggestion! While I don't like to emit these function as string literals either, unfortunately, I don't think that can work on Metal, because of the usage of [`newLibraryWithSource:options:completionHandler`](https://developer.apple.com/documentation/metal/mtldevice/1433351-newlibrarywithsource). The doc says that 

> The source may only import the Metal standard library. There is no search path to find other functions.

X(

k-ye: Let me remove all those `fmt::format()` stuffs and just emit a multiline string lietral..

yuanming-hu: To me, it's now much easier to read! It's a pity that Metal shaders cannot do `#include`... If we are getting more of these functions, we might consider manually "including" something like `helpers.metal` by prepending to the emitted code, so that we get rid of the `emit("...")`. Of course, you know the Metal backend much better than I do so it's your decision. I'm just talking :-)

k-ye: Great :-)

> Metal shaders cannot do #include... If we are getting more of these functions

Ah, Metal shaders can, but then you will have to output these source code into an actual file, then invoke the Metal compiler, which I think is the legacy approach? We can switch to that if these helpers continue to grow.

> we might consider manually "including" something like helpers.metal by prepending to the emitted code

+1. I guess one way to do this is the `inc.h` approach, where we put these helpers into a header, surrounded by some macro. And during the codegen time, that macro expands to `emit()`... Will need more time playing on this idea...

k-ye:
I'll squash the commits if there's no other comments..?

taichi/backends/codegen_metal.cpp
yuanming-hu: > Will need more time playing on this idea...

Yeah I thought a little bit about this. One solution could be creating a `helpers.cpp` with contents like

```C++
std::string metal_helpers = R("

f32 atomic_add(f32 *dest, f32 val) {
  do (...) {
    ...  
  } while (CAS...);
}

...other functions...
");
```

then we prepend `metal_helpers` to the emitted code. This is very close to invoke a single `emiit` with a multiline string literal though. Anyway, I think we are good for now. Merging...

yuanming-hu:
I can do the squashing (just one click on my end). (Appveyor is so slow...bypassing it)

Screenshot from 2020-02-14 23-51-33

taichi/backends/codegen_metal.cpp
k-ye: One sec, let me obey the new rules and squash it 😂

yuanming-hu: (Sorry about merging too fast... XD)

> Yeah, i think the macro can expand to std::string metal_helpers = R(" only inside codegen, and will expand to nothing in other cases.

That's a good idea! I think having a *header* with just the function definition as the body (without the quotes) will enable the IDE to syntax highlight to some extent.

>  Then we can even use c++ compiler to make sure the code compilers.. (On the other hand, there are a few Metal specific keywords and functions, so maybe this approach will be very noisy)

I agree having the C++ compile checking syntax correctness for us is good, but as you said it might be noisy...

yuanming-hu: Btw it seems that there's no way for me to merge in https://github.com/taichi-dev/taichi/pull/465/commits/a753899b29b25fbc157e1562af7497aa798e9b1f ... Please feel free to open a new PR for that (so sorry about the trouble...)

k-ye: I think that commit belongs to my local branch, and on master these two commits are already squashed into one? See abbf5b13537ab5e0f1d951fb3502b65a4f509bb1 :)

PR #513, opened 2020-02-21
Taichi build on macOS with Python 3.6 to Python 3.8 covered #409

Detavern:
Related issue id = #409

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,15 +7,15 @@ matrix:
   - os: osx
     osx_image: xcode10.3
     env:
-      - MATRIX_EVAL="CC=clang && CXX=clang++ && PYENV_VERSION=3.6.1 && PYTHON=python"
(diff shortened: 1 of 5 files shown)


Detavern:
Hi, as far as I know, [python builds are not available on the macOS environment](https://docs.travis-ci.com/user/languages/python/), which is same even using generic travis builds.

Currently, python version on osx depends on `Xcode` image version.

In my approach, `homebrew` and `pyenv` are used to build specific python version on macOS (quite tricky).

Additionally, since python 3.8, `platform.platform()` will use `platform.mac_ver()` instead.

```ipython
# Python 3.6, 3.7
In [1]: import platform

In [2]: platform.platform()
Out[2]: 'Darwin-19.3.0-x86_64-i386-64bit'

# Python 3.8
In [1]: import platform

In [2]: platform.platform()
Out[2]: 'macOS-10.15.3-x86_64-i386-64bit'
```

Finally covered 🍺

Issue #524, opened 2020-02-24
Use a condition in a kernel crashes taichi


medakk:
I'm not sure if I'm using taichi incorrectly, but the following program crashes:

```python
import taichi as ti

ti.init(arch=ti.cuda) # Run on GPU by default

n = 100
pixels = ti.var(dt=ti.i32, shape=(n, n))

@ti.kernel
def paint(t: ti.i32):
    if t < 100:
        for i, j in pixels:
            pixels[i, j] = 1

gui = ti.GUI("Automata", (n, n))
for t in range(n):
    paint(t)
    gui.set_image(pixels)
    gui.show()
```

```
[Release mode]
[Taichi version 0.5.2, cuda 10.1, commit 4d56959a]
[E 02/23/20 18:21:46.543] [statements.inc.h:visit@19] Not supported.
[E 02/23/20 18:21:46.543] Received signal 6 (Aborted)
***********************************
* Taichi Compiler Stack Traceback *
***********************************
/home/sci/karthik/.virtualenvs/deep/lib/python3.6/site-packages/taichi/core/../lib/taichi_core.so: taichi::signal_handler(int)
/lib64/libc.so.6(+0x361e0) [0x7f468dc031e0]
/lib64/libpthread.so.0: raise
/home/sci/karthik/.virtualenvs/deep/lib/python3.6/site-packages/taichi/core/../lib/taichi_core.so: taichi::Tlang::IRVisitor::visit(taichi::Tlang::StructForStmt*)
/home/sci/karthik/.virtualenvs/deep/lib/python3.6/site-packages/taichi/core/../lib/taichi_core.so: taichi::Tlang::CodeGenLLVM::visit(taichi::Tlang::Block*)
/home/sci/karthik/.virtualenvs/deep/lib/python3.6/site-packages/taichi/core/../lib/taichi_core.so: taichi::Tlang::CodeGenLLVM::visit(taichi::Tlang::IfStmt*)
/home/sci/karthik/.virtualenvs/deep/lib/python3.6/site-packages/taichi/core/../lib/taichi_core.so: taichi::Tlang::CodeGenLLVM::visit(taichi::Tlang::Block*)
/home/sci/karthik/.virtualenvs/deep/lib/python3.6/site-packages/taichi/core/../lib/taichi_core.so: taichi::Tlang::CodeGenLLVMGPU::visit(taichi::Tlang::OffloadedStmt*)
[...]

PR #525, opened 2020-02-24
Update doc around how the loops at the outermost scope is parallelized

k-ye:
Related issue id = #524


k-ye:
Looks like I screwed up the formatting 😂 I'll find an RST previewer

Screen Shot 2020-02-24 at 12 33 27

Issue #504, opened 2020-02-19
Add ti.u8 and ti.i16 support #505


archibate:
**Concisely describe the proposed feature**
I would like to add more data types (eg. `ti.u8`, `ti.i16`) to the compiler so that audio and graphic users could be happy dealing with their raw data.

**Describe the solution you'd like**
`u8` is already in `taichi/tlang_util.h`.
You may want to start from `python/taichi/lang/util.py` (?)

**Additional comments**
Just the basic supports, not need too much things like `pow_i8`.

PR #527, opened 2020-02-24
Make test_types.py work for the Metal backend

k-ye:
Related issue id = #504 #396 

Hi @archibate 

Sorry i changed your tests quite a bit, in order to make it work for the Metal backend as well. The problem with Metal is that it doesn't support 64bit buffers, so we had to separate the tests into 64-bit vs the rest. 

I found a nice decorator `@pytest.mark.parametrize()` ([doc](http://doc.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize)) that can somewhat simplify the code. 

Also I saw that you got this "variables are expected to be declared before kernel invocation" problem,  therefore had to create another wrapper inside each test? IIUC, `ti.all_archs(_with)` resets the test case here:
https://github.com/taichi-dev/taichi/blob/57287cf0493f5719e0e8b021fe18cdfe0c20c718/python/taichi/lang/__init__.py#L224

So by putting `@ti.all_archs` at the bottom of the decorator chain, we no longer need to wrap it, because it resets Taichi every time the test case is called. (Unfortunately, this doesn't apply to this PR, as explained below)

At last, I tried something like this:

```py
def _test_foo(x):
  ...

@pytest.mark.parametrize('x', [1, 2, 3])
@ti.all_archs
def test_foo(x):
  _test_foo(x)
```

But `@pytest.mark.parametrize` didn't work well with `@ti.all_archs`  this way. So I had to once again, wrap the `_test_foo()` inside `test_foo()`, so that `@ti.all_archs` can correctly apply the decoration:

```py
@pytest.mark.parametrize('x', [1, 2, 3])
def test_foo(x):
  @ti.all_archs
  def run():
    _test_foo(x)
  run()
```

### DIFF
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -330,6 +330,48 @@ def must_throw(ex):
   return decorator
 
 
+# Poor man's parametrized test decorator
+# The usage is identical to how @pytest.mark.parametrize is commonly used. For
+# example:
+#
+# @ti.parametrize('foo', [1, 2, 3])
+# @ti.all_archs
+# def test_xx(foo):
+#    ...
+#
+# @ti.parametrize('foo,bar', [
+#   (1, 'a'),
+#   (2, 'b'),
+#   (3, 'c'),
+# ])
+# @ti.all_archs
+# def test_yy(foo, bar):
+#    ...
+def parametrize(argnames: str, argvalues):
+  # @pytest.mark.parametrize only works for canonical function args, and doesn't
+  # support *args or **kwargs. This makes it difficult to play along with other
+  # decorators like @ti.all_archs. As a result, we implement our own.
+  argnames = [s.strip() for s in argnames.split(',')]
+  def iterable(x):
+    try:
+      _ = iter(x)
+      return True
+    except:
+      return False
+
+  def decorator(test):
+    def wrapped(*test_args, **test_kwargs):
+      for vals in argvalues:
+        if isinstance(vals, str) or not iterable(vals):
+          vals = (vals, )
+        kwargs = {k: v for k, v in zip(argnames, vals)}
+        assert len(kwargs.keys() & test_kwargs.keys()) == 0
+        kwargs.update(test_kwargs)
+        test(*test_args, **kwargs)
+    return wrapped
+  return decorator
+
+
 def complex_kernel(func):
 
   def decorated(*args, **kwargs):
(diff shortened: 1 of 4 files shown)


archibate:
> Also I saw that you got this "variables are expected to be declared before kernel invocation" problem, therefore had to create another wrapper inside each test? IIUC, ti.all_archs(_with) resets the test case.

Sounds to be a good solution for this!

> But @pytest.mark.parametrize didn't work well with @ti.all_archs this way.

So that the problem is, ti.all_archs didn't pass x to test_foo in the first case?

archibate:
That's strange, in:
https://github.com/taichi-dev/taichi/blob/57287cf0493f5719e0e8b021fe18cdfe0c20c718/python/taichi/lang/__init__.py#L211
`test` arguments should be recognized...

k-ye:
> That's strange, in:
> https://github.com/taichi-dev/taichi/blob/57287cf0493f5719e0e8b021fe18cdfe0c20c718/python/taichi/lang/__init__.py#L211
> 
> 
> `test` arguments should be recognized...

Yeah, I think `pytest.mark.parametrize` only knows about the normal function arguments, no defaults, no `*args`, no `**kwargs`... https://github.com/pytest-dev/pytest/issues/3221 Unfortunately we have to put `@ti.all_archs` at the bottom of this chain :(

archibate:
Maybe we can customize our own `parametrize` like:
```python
def parametrize(xs):
  def decorator(foo):
    def wrapped(*args, **kwargs):
      for x in xs:
        foo(x, *args, **kwargs)
    return wrapped
  return decorator

@parametrize([1, 2, 3])
@ti.all_archs
def test_func(x):
  assert x == 2333
```

k-ye:
> Maybe we can customize our own parametrize like

SG! While I generally don't prefer maintaining our own utilities for such things that are not essential to Taichi, I guess having a `@ti.parametrize` will work better for our tests, because most of them need to use `@ti.all_arch`. Thanks!

archibate:
Cool implementation! I opened an issue for this on https://github.com/pytest-dev/pytest/issues/6810.

archibate:
Their reply: https://github.com/pytest-dev/pytest/issues/6810#issuecomment-590371372.
They say we can solve this by adding `functools.wraps` to `ti.all_archs`... how do you think?

k-ye:
Thank you so much for following up with the pytest team! Yeah, after decorating the tests with `@functools.wraps`, we can now integrate with `@pytest.mark.parametrize`!! Again, really appreciate your efforts! (BTW I'm not available during working hours, so I will have to catch up new comments later tonight..)

Issue #523, opened 2020-02-23
raise python errors instead of c++ abort #533


archibate:
I would like to throw python exception instead of getting c++ abort for some of taichi errors (eg. imread-file-not-found, new-var-after-materialization) so that people could catch them in python and deal with it.

**Describe the solution you'd like**
> When C++ code invoked from Python throws an `std::exception`, it is automatically converted into a Python Exception.

So, throw `std::exception` in `TI_ERROR`, instead of `std::raise`. Doing this, you'll got a `RuntimeError` in python-scope.

**Additional comments**
See https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html for more details.

archibate:
Also helps `pytest` to __show messages__ when error occur in C++ scope instead of exit silently with code 255.

archibate:
TODO: @archibate you gonna work on this tonight!

archibate:
TODO: gdb trigger config var.

archibate:
Seems already solved by @yuanming-hu. Closing now.

PR #533, opened 2020-02-25
throw exception instead of SIGABRT to obtain RuntimeError in Python-scope #523

archibate:
Related issue id = #523

### DIFF
--- a/taichi/core/logging.cpp
+++ b/taichi/core/logging.cpp
@@ -69,7 +69,7 @@ Logger::Logger() {
     try {
       if (p) std::rethrow_exception(p);
     } catch (const std::string &e) {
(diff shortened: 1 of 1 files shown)


archibate:
Travis CI: a unexpected crash while runing test.

archibate:
TODO: @archibate will figure out if this is an accident or not and improve this PR tomorrow.

archibate:
Look like the problem... It would be better if pytest could tell what type of crash it is... eg. segfault or assert fail.

archibate:
After this, gdb won't be triggered any more. May add runtime config vars to control if gdb should be triggered (pls default to OFF to make pytest display error message!!!)

PR #541, opened 2020-02-26
Fix Metal crashes due to the usage of dynamic snode

k-ye:
Metal doesn't support `dynamic` or any other sparse snodes... :(


k-ye:
No worries at all XD

PR #557, opened 2020-03-02
Do not raise error if the `for` loop is static

xumingkuan:


k-ye:
i'm actually confused. Do we need to enforce Kernel Simplicity Rule or not (https://taichi.readthedocs.io/en/latest/differentiable_programming.html#differentiable-programming)?

PR #564, opened 2020-03-04
[refactor] Construct `MetalKernelArgsAttribute` directly from `Kernel::Arg`s

k-ye:
Related issue id = #396


k-ye:
Actually there's no need to change `program/kernel.h|cpp`. Will merge this directly after CI..

PR #583, opened 2020-03-11
Support break in non-parallel for statements by translate range-for into while #578

archibate:
Related issue id = #578

### DIFF
--- a/cmake/TaichiCore.cmake
+++ b/cmake/TaichiCore.cmake
@@ -34,7 +34,7 @@ endif()
 
 option(USE_STDCPP "Use -stdlib=libc++" OFF)
 option(TI_WITH_CUDA "Build with the CUDA backend" OFF)
-option(TI_WITH_OPENGL "Build with the OpenGL backend" OFF)
+option(TI_WITH_OPENGL "Build with the OpenGL backend" ON)
 option(GLEW_USE_STATIC_LIBS OFF)
 
 include_directories(${CMAKE_SOURCE_DIR})
--- a/cmake/TaichiCore.cmake
+++ b/cmake/TaichiCore.cmake
@@ -34,7 +34,7 @@ endif()
 
 option(USE_STDCPP "Use -stdlib=libc++" OFF)
 option(TI_WITH_CUDA "Build with the CUDA backend" OFF)
-option(TI_WITH_OPENGL "Build with the OpenGL backend" ON)
+option(TI_WITH_OPENGL "Build with the OpenGL backend" OFF)
 option(GLEW_USE_STATIC_LIBS OFF)
 
 include_directories(${CMAKE_SOURCE_DIR})
--- a/taichi/analysis/detect_for_with_break.cpp
+++ b/taichi/analysis/detect_for_with_break.cpp
@@ -15,21 +15,18 @@ class DetectForWithBreak : public BasicStmtVisitor {
   }
 
   void visit(FrontendWhileStmt *stmt) override {
-    TI_INFO("FOUND WHILE");
     loop_stack.push_back(stmt);
     stmt->body->accept(this);
     loop_stack.pop_back();
   }
 
   void visit(FrontendForStmt *stmt) override {
-    TI_INFO("FOUND FOR");
     loop_stack.push_back(stmt);
     stmt->body->accept(this);
     loop_stack.pop_back();
   }
 
   void visit(FrontendIfStmt *stmt) override {
-    TI_INFO("FOUND IF");
     if (stmt->true_statements)
       stmt->true_statements->accept(this);
     if (stmt->false_statements)
@@ -37,7 +34,6 @@ class DetectForWithBreak : public BasicStmtVisitor {
   }
 
   void visit(FrontendBreakStmt *stmt) override {
-    TI_INFO("FOUND BREAK!");
     TI_ASSERT_INFO(loop_stack.size() != 0, "break statement out of loop scope");
     auto loop = loop_stack.back();
     if (loop->is<FrontendForStmt>())
--- a/taichi/analysis/detect_for_with_break.cpp
+++ b/taichi/analysis/detect_for_with_break.cpp
@@ -1,5 +1,6 @@
 #include "taichi/ir/ir.h"
-#include <set>
+
+#include <unordered_set>
 
 TLANG_NAMESPACE_BEGIN
 
@@ -8,7 +9,7 @@ class DetectForWithBreak : public BasicStmtVisitor {
   using BasicStmtVisitor::visit;
 
   std::vector<Stmt *> loop_stack;
-  std::set<Stmt *> fors_with_break;
+  std::unordered_set<Stmt *> fors_with_break;
   IRNode *root;
 
   DetectForWithBreak(IRNode *root) : root(root) {
@@ -40,14 +41,14 @@ class DetectForWithBreak : public BasicStmtVisitor {
       fors_with_break.insert(loop);
   }
 
-  std::vector<Stmt *> run() {
+  std::unordered_set<Stmt *> run() {
     root->accept(this);
-    return std::vector<Stmt *>(fors_with_break.begin(), fors_with_break.end());
+    return fors_with_break;
   }
 };
 
 namespace irpass {
-std::vector<Stmt *> detect_fors_with_break(IRNode *root) {
+std::unordered_set<Stmt *> detect_fors_with_break(IRNode *root) {
   DetectForWithBreak detective(root);
   return detective.run();
 }
--- a/taichi/analysis/detect_for_with_break.cpp
+++ b/taichi/analysis/detect_for_with_break.cpp
@@ -1,0 +1,60 @@
+#include "taichi/ir/ir.h"
+#include <set>
+
+TLANG_NAMESPACE_BEGIN
+
+class DetectForWithBreak : public BasicStmtVisitor {
+ public:
+  using BasicStmtVisitor::visit;
+
+  std::vector<Stmt *> loop_stack;
+  std::set<Stmt *> fors_with_break;
+  IRNode *root;
+
+  DetectForWithBreak(IRNode *root) : root(root) {
+  }
+
+  void visit(FrontendWhileStmt *stmt) override {
+    TI_INFO("FOUND WHILE");
+    loop_stack.push_back(stmt);
+    stmt->body->accept(this);
+    loop_stack.pop_back();
+  }
+
+  void visit(FrontendForStmt *stmt) override {
+    TI_INFO("FOUND FOR");
+    loop_stack.push_back(stmt);
+    stmt->body->accept(this);
+    loop_stack.pop_back();
+  }
+
+  void visit(FrontendIfStmt *stmt) override {
+    TI_INFO("FOUND IF");
+    if (stmt->true_statements)
+      stmt->true_statements->accept(this);
+    if (stmt->false_statements)
+      stmt->false_statements->accept(this);
+  }
+
+  void visit(FrontendBreakStmt *stmt) override {
+    TI_INFO("FOUND BREAK!");
+    TI_ASSERT_INFO(loop_stack.size() != 0, "break statement out of loop scope");
+    auto loop = loop_stack.back();
+    if (loop->is<FrontendForStmt>())
+      fors_with_break.insert(loop);
+  }
+
+  std::vector<Stmt *> run() {
+    root->accept(this);
+    return std::vector<Stmt *>(fors_with_break.begin(), fors_with_break.end());
+  }
+};
+
+namespace irpass {
+std::vector<Stmt *> detect_fors_with_break(IRNode *root) {
+  DetectForWithBreak detective(root);
+  return detective.run();
+}
+}  // namespace irpass
+
+TLANG_NAMESPACE_END
--- a/taichi/transforms/lower_ast.cpp
+++ b/taichi/transforms/lower_ast.cpp
@@ -196,8 +196,8 @@ class LowerAST : public IRVisitor {
         auto mask = std::make_unique<AllocaStmt>(DataType::i32);
         new_while->mask = mask.get();
         auto &stmts = new_while->body;
-        for (int i = 0; i < (int)flattened.size(); i++) {
-          stmts->insert(std::move(flattened[i]), i);
+        for (int i = 0; i < (int)flattened2.size(); i++) {
+          stmts->insert(std::move(flattened2[i]), i);
         }
         // insert break
         stmts->insert(
--- a/taichi/transforms/lower_ast.cpp
+++ b/taichi/transforms/lower_ast.cpp
@@ -168,7 +168,7 @@ class LowerAST : public IRVisitor {
       begin->flatten(flattened);
       end->flatten(flattened);
       static bool is_good_range_for = true; // TODO(archibate): detect if outer-most one
-      if (1 || is_good_range_for) { // #578
+      if (is_good_range_for) { // #578
         is_good_range_for = false;
         auto &&new_for = std::make_unique<RangeForStmt>(
             stmt->parent->lookup_var(stmt->loop_var_id[0]), begin->stmt,
@@ -187,6 +187,9 @@ class LowerAST : public IRVisitor {
         auto loop_var_load = Stmt::make<LocalLoadStmt>(loop_var_addr);
         auto cond_stmt = Stmt::make<BinaryOpStmt>(BinaryOpType::cmp_lt,
             std::move(loop_var_load).get(), end->stmt);
+        flattened.push_back(std::move(loop_var_load));
+        flattened.push_back(std::move(cond_stmt));
+        auto cond_stmt_got = flattened.back().get();
 
         auto &&new_while = std::make_unique<WhileStmt>(std::move(stmt->body));
         auto mask = std::make_unique<AllocaStmt>(DataType::i32);
@@ -197,7 +200,7 @@ class LowerAST : public IRVisitor {
         }
         // insert break
         stmts->insert(
-            std::make_unique<WhileControlStmt>(new_while->mask, std::move(cond_stmt).get()),
+            std::make_unique<WhileControlStmt>(new_while->mask, cond_stmt_got),
             flattened.size());
         stmt->insert_before_me(std::make_unique<AllocaStmt>(DataType::i32));
         auto &&const_stmt =
--- a/taichi/transforms/lower_ast.cpp
+++ b/taichi/transforms/lower_ast.cpp
@@ -167,7 +167,9 @@ class LowerAST : public IRVisitor {
       auto end = stmt->end;
       begin->flatten(flattened);
       end->flatten(flattened);
-      if (0 /* is_good_range_for */) { // #578
+      static bool is_good_range_for = true; // TODO(archibate): detect if outer-most one
+      if (1 || is_good_range_for) { // #578
+        is_good_range_for = false;
         auto &&new_for = std::make_unique<RangeForStmt>(
             stmt->parent->lookup_var(stmt->loop_var_id[0]), begin->stmt,
             end->stmt, std::move(stmt->body), stmt->vectorize, stmt->parallelize,
@@ -176,11 +178,15 @@ class LowerAST : public IRVisitor {
       } else {
         // transform into a structure as
         // while (1) { cond; if (no active) break; original body...}
-        //auto loop_var = Stmt::make<AllocaStmt>(DataType::i32); // may use stmt->parent->lookup_var(stmt->loop_var_id[0])?
+        //auto loop_var = Stmt::make<AllocaStmt>(DataType::i32);
         auto loop_var = stmt->parent->lookup_var(stmt->loop_var_id[0]);
         flattened.push_back(Stmt::make<LocalStoreStmt>(loop_var, begin->stmt));
-        // Alloca Assign loop_var from begin->stmt;
-        auto cond_stmt = Stmt::make<BinaryOpStmt>(BinaryOpType::cmp_lt, loop_var, end->stmt);
+        auto loop_var_addr = LaneAttribute<LocalAddress>(
+            LocalAddress(loop_var->as<AllocaStmt>(), 0));
+        // auto loop_var_store = Stmt::make<LocalStoreStmt>(loop_var_addr, begin->stmt);
+        auto loop_var_load = Stmt::make<LocalLoadStmt>(loop_var_addr);
+        auto cond_stmt = Stmt::make<BinaryOpStmt>(BinaryOpType::cmp_lt,
+            std::move(loop_var_load).get(), end->stmt);
 
         auto &&new_while = std::make_unique<WhileStmt>(std::move(stmt->body));
         auto mask = std::make_unique<AllocaStmt>(DataType::i32);
@@ -203,6 +209,7 @@ class LowerAST : public IRVisitor {
             std::make_unique<LocalStoreStmt>(new_while->mask, const_stmt_ptr));
         new_while->body->mask_var = new_while->mask;
         stmt->parent->replace_with(stmt, std::move(new_while));
+        throw IRModified();
       }
     } else {
       std::vector<Stmt *> vars(stmt->loop_var_id.size());
--- a/taichi/transforms/lower_ast.cpp
+++ b/taichi/transforms/lower_ast.cpp
@@ -167,11 +167,43 @@ class LowerAST : public IRVisitor {
       auto end = stmt->end;
       begin->flatten(flattened);
       end->flatten(flattened);
-      auto &&new_for = std::make_unique<RangeForStmt>(
-          stmt->parent->lookup_var(stmt->loop_var_id[0]), begin->stmt,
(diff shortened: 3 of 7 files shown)


archibate:
Why transform my BinaryOpStmt(cmp_lt) into `const [0]`.

archibate:
> Why transform my BinaryOpStmt(cmp_lt) into `const [0]`.

May because that cond_stmt not inserted into stmts.

archibate:
Too hard, can't do this at all.

yuanming-hu:
> Too hard, can't do this at all.

What exactly stopped you? :-)

archibate:
> > Too hard, can't do this at all.
> 
> What exactly stopped you? :-)

Don't have a clear image of how Stmt::insert and scoping works.

taichi/transforms/lower_ast.cpp
```diff
@@ -167,11 +167,50 @@ class LowerAST : public IRVisitor {
...
+        }
+        // insert break
+        stmts->insert(
+            std::make_unique<WhileControlStmt>(new_while->mask, std::move(cond_stmt).get()),
+            flattened.size());
+        stmt->insert_before_me(std::make_unique<AllocaStmt>(DataType::i32));
+        auto &&const_stmt =
+          std::make_unique<ConstStmt>(TypedConstant((int32)0xFFFFFFFF));
+        auto const_stmt_ptr = const_stmt.get();
+        stmt->insert_before_me(std::move(mask));
```
archibate: What on earth does `mask` mean? An alias of `cond`?

taichi/transforms/lower_ast.cpp
```diff
@@ -167,11 +167,50 @@ class LowerAST : public IRVisitor {
...
+        auto &&new_while = std::make_unique<WhileStmt>(std::move(stmt->body));
+        auto mask = std::make_unique<AllocaStmt>(DataType::i32);
+        new_while->mask = mask.get();
+        auto &stmts = new_while->body;
+        for (int i = 0; i < (int)flattened.size(); i++) {
+          stmts->insert(std::move(flattened[i]), i);
+        }
+        // insert break
+        stmts->insert(
+            std::make_unique<WhileControlStmt>(new_while->mask, std::move(cond_stmt).get()),
```
archibate: How does `WhileControlStmt` control `while` by `mask`?

yuanming-hu: `mask` is for vectorization on CPUs. It indicates which lanes of a SIMD vector are active.

taichi/transforms/lower_ast.cpp
yuanming-hu: `mask` is for vectorization on CPUs. It indicates which lanes of a SIMD vector are active.

taichi/transforms/lower_ast.cpp
archibate: So it's `(mask & 1) == cond` on single CPU?

archibate: > WhileControlStmt will just set the lane masks to 0 so that inactive lanes (i.e. those already exited) will not get executed.

Cool impl! will vectorized ones do atomic read mask to become zero only if mask bit nr > current bit nr?

archibate:
Can I just translate it into FrontendWhileStmt instead of WhileStmt?

yuanming-hu:
> Can I just translate it into FrontendWhileStmt instead of WhileStmt? (in somewhere before lower_ast)

Actually, after taking a look at your current implementation, I think it's very very close! We should probably stick to the current one.

archibate:
> > Can I just translate it into FrontendWhileStmt instead of WhileStmt? (in somewhere before lower_ast)
> 
> Actually, after taking a look at your current implementation, I think it's very very close! We should probably stick to the current one.

Close? Don't understand why it make my `BinaryOpStmt` becomes `const [0]`.. May be closer if this mystery could reason..

taichi/transforms/lower_ast.cpp
```diff
@@ -167,11 +167,50 @@ class LowerAST : public IRVisitor {
...
+        // transform into a structure as
+        // while (1) { cond; if (no active) break; original body...}
+        //auto loop_var = Stmt::make<AllocaStmt>(DataType::i32);
+        auto loop_var = stmt->parent->lookup_var(stmt->loop_var_id[0]);
+        flattened.push_back(Stmt::make<LocalStoreStmt>(loop_var, begin->stmt));
+        auto loop_var_addr = LaneAttribute<LocalAddress>(
+            LocalAddress(loop_var->as<AllocaStmt>(), 0));
+        // auto loop_var_store = Stmt::make<LocalStoreStmt>(loop_var_addr, begin->stmt);
+        auto loop_var_load = Stmt::make<LocalLoadStmt>(loop_var_addr);
+        auto cond_stmt = Stmt::make<BinaryOpStmt>(BinaryOpType::cmp_lt,
```
archibate: May be it wants me push this two into flattened? std::unique_ptr so annoying! Don't know if get() work after std::move.

archibate:
Seems whilestmt used a diff way from forstmt to insert: insert-before-me and flattened. Go bed now.

yuanming-hu:
Thanks. I'll take a look after a meeting.

taichi/transforms/lower_ast.cpp
yuanming-hu: Yes, you need to store the instructions in the block.

yuanming-hu:
Is there a way to reproduce your issue? Which test case should I run? I think your implementation looks promising. Just need to figure out what's wrong in detail.

yuanming-hu:
Added a simple test and rebase your commits after code reformatting to avoid future merge conflicts. Please update locally.

tests/python/test_for_break.py
```diff
@@ -0,0 +1,23 @@
...
+
+    ti.root.dense(ti.i, N).place(x)
+
+    @ti.kernel
+    def func():
+        for i in range(N):
+            for j in range(N + 5):
+                if j >= i:
+                    break
+                x[i] = j
```
archibate: Maybe you want `x[i, j] = [i, j]` to track all assigns?

yuanming-hu: Right, that makes the test stronger.

archibate:
You've made TAB=4???!

archibate:
Done! Run `ninja -C build && ti test for_break -v --arch opengl` to reproduce.

archibate:
> Seems whilestmt used a diff way from forstmt to insert: insert-before-me and flattened. Go bed now.

Exactly this caused problem, fixed using flattened2.

archibate:
Next: how to detect if is_good_range_for.

archibate:
Can we disable print_ir for snode accessors?

yuanming-hu:
> Can we disable print_ir for snode accessors?

I think that's already disabled by default?

archibate:
> I think that's already disabled by default?
 
No, still print something like 2d_ten_arr.

archibate:
Hello? Have time tonight.

archibate:
Fixed one, stupid error.

archibate:
If `begin < end`, we enumerate `[begin, end)`.
If `end > begin`, we enumerate `[end, begin)` or `(begin, end]`?

archibate:
Do we have `IncStmt` for something like `x++`? Or at least `AugAssignStmt` for `x += 1`?

yuanming-hu:
> If `begin < end`, we enumerate `[begin, end)`.
> If `end > begin`, we enumerate `[end, begin)` or `(end, begin]`?

Sorry I'm on a meeting - quick reply: If `end > begin`, we should just do nothing



> Do we have `IncStmt` for something like `x++`? Or at least `AugAssignStmt` for `x += 1`?

You can use `AtomicOpStmt` or a combination of local load, add, local store

archibate:
Waiting CI test pass..

archibate:
> > If `begin < end`, we enumerate `[begin, end)`.
> > If `end > begin`, we enumerate `[end, begin)` or `(end, begin]`?
> 
> Sorry I'm on a meeting - quick reply: If `end > begin`, we should just do nothing

Good then, so do my impl.

> > Do we have `IncStmt` for something like `x++`? Or at least `AugAssignStmt` for `x += 1`?
> 
> You can use `AtomicOpStmt` or a combination of local load, add, local store

We don't need that to be atomic, since loop_var is only used locally in this loop.

yuanming-hu:
Do we need the pass to distinguish good fors from bad ones (those with break)? We still want to keep the good fors since that will allow more compiler optimizations later.

archibate:
Btw, format server says:
`FileNotFoundError: [Errno 2] No such file or directory: '/home/taichigardener/repos/taichi/tests/python/test_cond_grad.py'`.
Seems skipping is not syncing current file names..

archibate:
> Do we need the pass to distinguish good fors from bad ones (those with break)? We still want to keep the good fors since that will allow more compiler optimizations later.

Oh right! That's for optimization! Reasonable, but how to detect if a for has break?

yuanming-hu:
> Btw, format server says:
> `FileNotFoundError: [Errno 2] No such file or directory: '/home/taichigardener/repos/taichi/tests/python/test_cond_grad.py'`.
> Seems skipping is not syncing current file names..

It seems that the format server doesn't handle deleted files correctly.

yuanming-hu:
> > Do we need the pass to distinguish good fors from bad ones (those with break)? We still want to keep the good fors since that will allow more compiler optimizations later.
> 
> Oh right! That's for optimization! Reasonable, but how to detect if a for has break?

You can use something like

```C++
class DetectForsWithBreak: public BasicStmtVisitor {
pubilc:
  void visit(FrontendForStmt *stmt) {
    current_for = stmt;
  }
  void visit(FrontendBreakStmt *stmt) {
    fors_with_break.insert(current_for);
  }
}

```

archibate:
> > Btw, format server says:
> > `FileNotFoundError: [Errno 2] No such file or directory: '/home/taichigardener/repos/taichi/tests/python/test_cond_grad.py'`.
> > Seems skipping is not syncing current file names..
> 
> It seems that the format server doesn't handle deleted files correctly.

Not sure how `git.Repo` works.. will `repo.index.diff('abc233')` checkout to the commit `abc233` instead of just return some informations about it?

archibate:
Okay here, have a careful look at what I wrote, may buggy since the time is late here.

yuanming-hu:
> Not sure how `git.Repo` works.. will `repo.index.diff('abc233')` checkout to the commit `abc233` instead of just return some informations about it?

It'll just return the diff information without checking out.

yuanming-hu:
> Okay here, have a careful look at what I wrote, may buggy since the time is late here.

Thank you so much! I'll take a look later after a meeting (which I need to prepare for...) in the afternoon

archibate:
> > Not sure how `git.Repo` works.. will `repo.index.diff('abc233')` checkout to the commit `abc233` instead of just return some informations about it?
> 
> It'll just return the diff information without checking out.

That might be the source of problem, FormatFile is called will search for real existing files(?)

archibate:
Travis, error.

yuanming-hu:
> > > Not sure how `git.Repo` works.. will `repo.index.diff('abc233')` checkout to the commit `abc233` instead of just return some informations about it?
> > 
> > 
> > It'll just return the diff information without checking out.
> 
> That might be the source of problem, FormatFile is called will search for real existing files(?)

Right. That file was probably modified (and deleted)...

archibate:
> > > > Not sure how `git.Repo` works.. will `repo.index.diff('abc233')` checkout to the commit `abc233` instead of just return some informations about it?
> > > 
> > > 
> > > It'll just return the diff information without checking out.
> > 
> > 
> > That might be the source of problem, FormatFile is called will search for real existing files(?)
> 
> Right. That file was probably modified (and deleted)...

And the format didn't applied to the changed file, we may want another issue+PR pair to fix this.

archibate:
have to sleep now,
```
=========================== short test summary info ===========================
FAILED tests/python/test_loop_grad.py::test_loop_grad - RuntimeError: [make_a...
FAILED tests/python/test_offload.py::test_reduce_merged - RuntimeError: [make...
FAILED tests/python/test_stop_grad.py::test_stop_grad2 - RuntimeError: [make_...
================= 3 failed, 285 passed in 1466.13s (0:24:26) ==================
```
ill try fix tmr.

archibate:
make_adjoint doesn't support WhileStmt?

yuanming-hu:
> make_adjoint doesn't support WhileStmt?

For now, it doesn't.

archibate:
That for doesn't have a break, seems we detected good for as bad for.

archibate:
> > make_adjoint doesn't support WhileStmt?
> 
> For now, it doesn't.

Is that possible to support WhileStmt? Even if it have break?

yuanming-hu:
> > > make_adjoint doesn't support WhileStmt?
> > 
> > 
> > For now, it doesn't.
> 
> Is that possible to support WhileStmt? Even if it have break?

Yes, but that's non-trivial. I'm working on supporting complex if stmts right now.

archibate:
> have to sleep now, err found related to make_adjoint/grad,
> 
> ```
> =========================== short test summary info ===========================
> FAILED tests/python/test_loop_grad.py::test_loop_grad - RuntimeError: [make_a...
> FAILED tests/python/test_offload.py::test_reduce_merged - RuntimeError: [make...
> FAILED tests/python/test_stop_grad.py::test_stop_grad2 - RuntimeError: [make_...
> ================= 3 failed, 285 passed in 1466.13s (0:24:26) ==================
> ```
> 
> ill try fix tmr.
> btw, why not `mv tests/python/* tests`.

Why? This locally passed.

archibate:
May be mergeable or may not, don't know why CI randomly breaking down test..

archibate:
Travis CI:
```
An error occurred while generating the build script.
```
First time 3.6&3.7, second time 3.8 breaks...
Why?

archibate:
Hello? If that's another issue you can merge PR as you want.

yuanming-hu:
> Travis CI:
> 
> ```
> An error occurred while generating the build script.
> ```
> 
> First time 3.6&3.7, second time 3.8 breaks...
> Why? Was this related to my code?

It's probably not your fault. Not sure what makes Travis unhappy recently.

yuanming-hu:
> Hello? If that's another issue you can merge this PR as you want.

Sorry I haven't got a chance to review this. Will do it now.

cmake/TaichiCore.cmake
```diff
@@ -34,7 +34,7 @@ endif()
 
 option(USE_STDCPP "Use -stdlib=libc++" OFF)
 option(TI_WITH_CUDA "Build with the CUDA backend" OFF)
-option(TI_WITH_OPENGL "Build with the OpenGL backend" OFF)
+option(TI_WITH_OPENGL "Build with the OpenGL backend" ON)
```
yuanming-hu: Maybe we should keep it `OFF`?

taichi/analysis/detect_for_with_break.cpp
```diff
@@ -0,0 +1,56 @@
...
+    loop_stack.pop_back();
+  }
+
+  void visit(FrontendForStmt *stmt) override {
+    loop_stack.push_back(stmt);
+    stmt->body->accept(this);
+    loop_stack.pop_back();
+  }
+
+  void visit(FrontendIfStmt *stmt) override {
```
yuanming-hu: Nice implementation here! Maybe we should put this in `BasicStmtVisitor` in a future PR, something like `Make BasicStmtVisitor support frontend statements`.

archibate:
> > Travis CI:
> > ```
> > An error occurred while generating the build script.
> > ```
> > 
> > 
> > First time 3.6&3.7, second time 3.8 breaks...
> > Why? Was this related to my code?
> 
> It's probably not your fault. Not sure what makes Travis unhappy recently.

I once found it didn't download llvm.zip correctly...

tests/python/test_for_break.py
```diff
@@ -0,0 +1,23 @@
+import taichi as ti
+
+
+@ti.all_archs_with(print_ir=True)
```
yuanming-hu: Let's remove `print_ir=True` since we are no longer debugging it :-) Thanks.

taichi/transforms/lower_ast.cpp
```diff
@@ -101,9 +103,9 @@ class LowerAST : public IRVisitor {
   }
 
   void visit(FrontendBreakStmt *stmt) override {
-    TI_ASSERT_INFO(
-        capturing_loop->is<WhileStmt>(),
-        "The loop capturing 'break' must be a while loop instead of for loop.");
+    TI_ASSERT(//TI_ASSERT_INFO(
```
yuanming-hu: How about removing all the comments here?

taichi/transforms/lower_ast.cpp
```diff
@@ -167,11 +169,67 @@ class LowerAST : public IRVisitor {
...
+        == detected_fors_with_break.end(); // for without break
+      if (is_good_range_for) {  // #578
+        auto &&new_for = std::make_unique<RangeForStmt>(
+            stmt->parent->lookup_var(stmt->loop_var_id[0]), begin->stmt,
+            end->stmt, std::move(stmt->body), stmt->vectorize,
+            stmt->parallelize, stmt->block_dim, stmt->strictly_serialized);
+        flattened.push_back(std::move(new_for));
+      } else {
+        // transform into a structure as
+        // i = begin; while (1) { if (i > end) break; original body; i += 1; }
```
yuanming-hu: Should it be `if (i >= end) break`?

taichi/transforms/lower_ast.cpp
```diff
@@ -167,11 +169,67 @@ class LowerAST : public IRVisitor {
...
+        // auto loop_var = Stmt::make<AllocaStmt>(DataType::i32);
+        auto loop_var = stmt->parent->lookup_var(stmt->loop_var_id[0]);
+        flattened.push_back(Stmt::make<LocalStoreStmt>(loop_var, begin->stmt));
+        auto loop_var_addr = LaneAttribute<LocalAddress>(
+            LocalAddress(loop_var->as<AllocaStmt>(), 0));
+        VecStatement flattened2;
+        // auto loop_var_store = Stmt::make<LocalStoreStmt>(loop_var_addr,
```
yuanming-hu: Please remove the lines that are commented out.

taichi/transforms/lower_ast.cpp
```diff
@@ -167,11 +169,67 @@ class LowerAST : public IRVisitor {
...
       begin->flatten(flattened);
       end->flatten(flattened);
-      auto &&new_for = std::make_unique<RangeForStmt>(
-          stmt->parent->lookup_var(stmt->loop_var_id[0]), begin->stmt,
-          end->stmt, std::move(stmt->body), stmt->vectorize, stmt->parallelize,
-          stmt->block_dim, stmt->strictly_serialized);
-      flattened.push_back(std::move(new_for));
+      bool is_good_range_for = capturing_loop == nullptr || // the outer-most for
+        std::find(detected_fors_with_break.begin(), detected_fors_with_break.end(), stmt)
+        == detected_fors_with_break.end(); // for without break
```
yuanming-hu: Let's add a comment to explain what `good range for` is. `for without break` is slightly vague.

taichi/transforms/lower_ast.cpp
```diff
@@ -167,11 +169,67 @@ class LowerAST : public IRVisitor {
...
+        auto &&new_while = std::make_unique<WhileStmt>(std::move(stmt->body));
+        auto mask = std::make_unique<AllocaStmt>(DataType::i32);
+        new_while->mask = mask.get();
+        auto &stmts = new_while->body;
+        for (int i = 0; i < (int)flattened2.size(); i++) {
+          stmts->insert(std::move(flattened2[i]), i);
+        }
+
+        VecStatement flattened3;
+        flattened3.push_back(Stmt::make<ConstStmt>(TypedConstant((int32)1)));
```
yuanming-hu: You can actually use `auto const_one = flattened3.push_back<ConstStmt>(TypedConstant((int32)1)`
Then you don't need `Stmt::make` and `flattened3.back().get()` :-)

taichi/analysis/detect_for_with_break.cpp
archibate: Potentially a boolean like `visit_container = true`?

yuanming-hu: `BasicStmtVisitor` already means it visits all the containers. We just need to add support for frontend containers for it.

cmake/TaichiCore.cmake
archibate: Can I make GL default to ON in something ignored by `.gitignore`? It's not easy to keep in mind to specify that long argument.

taichi/transforms/lower_ast.cpp
```diff
@@ -101,9 +103,7 @@ class LowerAST : public IRVisitor {
   }
 
   void visit(FrontendBreakStmt *stmt) override {
-    TI_ASSERT_INFO(
-        capturing_loop->is<WhileStmt>(),
-        "The loop capturing 'break' must be a while loop instead of for loop.");
+    TI_ASSERT(capturing_loop->is<WhileStmt>());
```
yuanming-hu: I'm slightly confused - now that we support `break` in certain for loops, why do we still need this assertion?

taichi/transforms/lower_ast.cpp
```diff
@@ -167,11 +167,62 @@ class LowerAST : public IRVisitor {
...
+            stmt->parallelize, stmt->block_dim, stmt->strictly_serialized);
+        flattened.push_back(std::move(new_for));
+      } else {
+        // transform into a structure as
+        // i = begin; while (1) { if (i >= end) break; original body; i += 1; }
+        auto loop_var = stmt->parent->lookup_var(stmt->loop_var_id[0]);
+        flattened.push_back<LocalStoreStmt>(loop_var, begin->stmt);
+        auto loop_var_addr = LaneAttribute<LocalAddress>(
+            LocalAddress(loop_var->as<AllocaStmt>(), 0));
+        VecStatement flattened2;
```
yuanming-hu: Can we rename `flattened2`/`flattened3` into `load_and_compare` etc.?

yuanming-hu:
(Added more tests and your implementation passes them. Good job!)

taichi/transforms/lower_ast.cpp
archibate: `error: no member named 'find' in 'std::vector<taichi::lang::Stmt *, std::allocator<taichi::lang::Stmt *> >'`

taichi/transforms/lower_ast.cpp
```diff
@@ -167,11 +167,62 @@ class LowerAST : public IRVisitor {
       auto end = stmt->end;
       begin->flatten(flattened);
       end->flatten(flattened);
-      auto &&new_for = std::make_unique<RangeForStmt>(
-          stmt->parent->lookup_var(stmt->loop_var_id[0]), begin->stmt,
-          end->stmt, std::move(stmt->body), stmt->vectorize, stmt->parallelize,
-          stmt->block_dim, stmt->strictly_serialized);
-      flattened.push_back(std::move(new_for));
+      bool is_good_range_for = capturing_loop == nullptr ||
+        std::find(detected_fors_with_break.begin(), detected_fors_with_break.end(), stmt)
```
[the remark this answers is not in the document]
archibate: `error: no member named 'find' in 'std::vector<taichi::lang::Stmt *, std::allocator<taichi::lang::Stmt *> >'`

taichi/transforms/lower_ast.cpp
archibate: Yes, `as<T>()` will do that anyway!

taichi/transforms/lower_ast.cpp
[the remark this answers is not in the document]
yuanming-hu: Oh I see. I thought you are using an `std::set` as in `DetectForWithBreak` - why change it to an `std::vector`? I think `std::set` is better.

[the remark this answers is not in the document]
archibate: See taichi/analysis/gather_deactivations.cpp#L26, any reason it use vector as return but store as set?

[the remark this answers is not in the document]
yuanming-hu: Thanks for pointing that out. I think we should change that to `std::set` as well.

taichi/analysis/detect_for_with_break.cpp
yuanming-hu: I realized that we need to record the capturing loops here, so a general-purpose frontend statement visitor won't help much. You current impl looks good.

yuanming-hu:
Actually maybe `unordereds_set` is better for these analysis passes. I changed that and will merge once CI passes. Thanks!

archibate:
Can we merge this now?

archibate:
`unordered_set`?? Wasn't `set` itself already means unordered?

yuanming-hu:
> `unordered_set`?? Wasn't `set` itself already means unordered? Sorry, I'm not a C++ expert..

Well, `set` is actually ordered: https://en.cppreference.com/w/cpp/container/set *std::set is an associative container that contains a **sorted** set of unique objects of type Key.*


> Can we merge this now?

Wiil do it once CI passes. This will be shipped with v0.5.9 tonight/tomorrow.

PR #587, opened 2020-03-14
add static ptr assignment for funcs too

KLozes:
This allows ti.static assignment for funcs and classfuncs as well.


KLozes:
No problem!

PR #627, opened 2020-03-20
add atomic min/max/bit_and/bit_or/bit_xor

KLozes:
Also hooked up atomic sub and regular xor :)

Related issue id = #606 


KLozes:
Thank you, I'm flattered! And No problem! Glad to help out!

tests/python/test_local_atomics.py
```diff
@@ -29,3 +29,155 @@ def func():
...
+        a = -1000
+        for i in range(10):
+            ti.atomic_max(a, i)
+        A[None] = a
+
+    func()
+    assert A[None] == 9
+
+
+@ti.all_archs
```
archibate: Great work! I think @k-ye and me will add these atomic ops to metal and opengl later so they'll pass this all_archs.

k-ye: Yep, created #632

PR #634, opened 2020-03-21
Add atomic_bit_or/xor/and for OpenGL

archibate:
Related issue id = #606 
This is a continue of #632 


archibate:
Much more easy as I thought.. since GL already have atomicAnd/Or instructions.

PR #691, opened 2020-04-01
[metal] Fix bug in `listgen` where it goes beyond `ListManager`'s capacity

k-ye:
FYI, without this change, the added test could actually fail on Metal...

Related issue = #593 


k-ye:
Right I remember that now! Merging :)

PR #666, opened 2020-03-27
[OpenGL] Support NVIDIA GLSL compiler

archibate:
Related issue = #633 #664 

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

Issue #603, opened 2020-03-15
[OpenGL] performance profiling&improving


archibate:
**Concisely describe the proposed feature**
GL too slow, only doubled FPS compared to x64 for my n-body example. Please improve it.

**Describe the solution you'd like (if any)**
Most time costed on sync (data transfer from&to GPU) thanks to GL's great API design, we may want to figure out how to use it more efficiently.

**Additional comments**
Related = #492.

archibate:
Currently root buffer is constantly stayed in GPU memory, so no much sync time worrying now. Only the external_ptr is very large one consideration.
Also note that NV has less sync overhead compared to mesa.

PR #692, opened 2020-04-02
[opengl] random use different seed in each launch

archibate:
So that `examples/sdf_renderer.py` works.

Related issue = #492 

### DIFF
--- a/taichi/backends/opengl/opengl_api.cpp
+++ b/taichi/backends/opengl/opengl_api.cpp
@@ -244,7 +244,8 @@ GLSSBO *root_ssbo;
 void create_glsl_root_buffer(size_t size) {
   // if (root_ssbo) return;
   initialize_opengl();
-  root_ssbo = new GLSSBO;  // TODO(archibate): mem leaking
+  root_ssbo = new GLSSBO;  // TODO(archibate): mem leaking, use std::optional instead
+  size += 2 * sizeof(int);
   void *buffer = std::calloc(size, 1);
   root_ssbo->bind_data(buffer, size, GL_DYNAMIC_READ);
   root_ssbo->bind_index(0);
--- a/examples/nbody_oscillator.py
+++ b/examples/nbody_oscillator.py
@@ -1,51 +1,42 @@
 import taichi as ti
 
 ti.init(
-    arch=ti.cuda)  # Try to run on GPU, use ti.opengl if you don't have CUDA
+    arch=ti.opengl)  # Try to run on GPU, use ti.opengl if you don't have CUDA
 
 N = 8000
-pos = ti.var(dt=ti.f32, shape=(N, 2))
-vel = ti.var(dt=ti.f32, shape=(N, 2))
+pos = ti.Vector(2, dt=ti.f32, shape=N)
+vel = ti.Vector(2, dt=ti.f32, shape=N)
 bounce = 0.8
 
 
 @ti.kernel
 def initialize():
     for i in range(N):
-        pos[i, 0] = ti.random() * 0.5 - 0.25 + 0.5
-        pos[i, 1] = ti.random() * 0.5 - 0.25 + 0.5
-        vel[i, 0] = ti.random() * 2 - 1
-        vel[i, 1] = ti.random() * 2 - 1
+        for k in ti.static(range(2)):
+            pos[i][k] = ti.random() * 0.5 - 0.25 + 0.5
+            vel[i][k] = ti.random() * 2 - 1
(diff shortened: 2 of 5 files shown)


yuanming-hu:
Thanks! I'll test it on my end. Meanwhile, let's finish https://github.com/taichi-dev/taichi/pull/666 first so that we get rid of the redundant commits.

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -236,7 +244,8 @@ GLSSBO *root_ssbo;
 void create_glsl_root_buffer(size_t size) {
   // if (root_ssbo) return;
   initialize_opengl();
-  root_ssbo = new GLSSBO;  // TODO(archibate): mem leaking
+  root_ssbo = new GLSSBO;  // TODO(archibate): mem leaking, use std::optional instead
```
archibate, line 247: A very little change (< 10 CoL) can fix this mem leakage, should I put in this PR?

yuanming-hu: If it's just 10 LoC then let's fix it here.

archibate:
We may also want to add a test case for this, to check if `ti.random()` return same when called again. e.g.
```py
@ti.kernel
def func(i: ti.i32):
  x[i] = ti.random()
for i in range(5):
  func(i)
  assert x[i] != x[i - 1]
```

yuanming-hu:
> We may also want to add a test case for this, to check if `ti.random()` return same when called again. e.g.
> 
> ```python
> @ti.kernel
> def func(i: ti.i32):
>   x[i] = ti.random()
> for i in range(5):
>   func(i)
>   assert x[i] != x[i - 1]
> ```

This might be a brittle test since there's a tiny probability that `x[i] == x[i - 1]`. It will be better to, say repeat 5 times and assert there are at least 4 different results.

yuanming-hu:
Screenshot from 2020-04-02 12-00-36

I confirm this works on my end. It reaches 90 samples per pixel per second, which is 1.8x faster than the CUDA backend!

archibate:
That's amazing! Couldn't believe we are now faster than CUDA! Does this mean that CUDA still has lot of space of optimizations to be done?

yuanming-hu:
I guess it's because CUDA is not yet fully optimized (and the OpenGL backend is well done). I haven't got a chance to work on it.

archibate:
Seems #603 is just 杞人忧天ing... the performance of current OpenGL backend is really utilized now.
NEXT: to support sparse data structures on GL, so that the thought-to-be-weak GLSL can really 平起平坐 with CUDA :)

archibate:
#603 says:

> Most time costed on sync (data transfer from&to GPU) thanks to GL's great API design, we may want to figure out how to use it more efficiently.

In fact I found sync is only costly on mesa one, not NV, maybe their GPUs have really faster communication with CPU?

yuanming-hu:
> #603 says:
> 
> > Most time costed on sync (data transfer from&to GPU) thanks to GL's great API design, we may want to figure out how to use it more efficiently.
> 
> In fact I found sync is only costly on mesa one, not NV, maybe their GPUs have really faster communication with CPU?

I guess it's because the NVIDIA driver is better implemented than mesa for NVIDIA GPUs.

archibate:
Could you also test `examples/nbody_oscillator.py` on NV-GL & CUDA? I guess CUDA is better when dealing with memory intensive kernels (doesn't it?)

yuanming-hu:
I have to change `N` to `80000` to make a difference. Then OpenGL reaches 15 FPS and CUDA 5 FPS. OpenGL is 3x faster :-) It seems me that there's a performance bug in the CUDA backend, which I should fix as soon as possible...

yuanming-hu:
I updated https://github.com/taichi-dev/taichi/pull/692/files#diff-733c00d9e85a14e095b788b8b985e75c

Btw do you want to fix the memory leakage here? If not I'll go merge this PR.

archibate:
> It seems me that there's a performance bug in the CUDA backend, which I should fix as soon as possible...

Sounds a serious issue, good luck!

> Btw do you want to fix the memory leakage here? If not I'll go merge this PR.

No, let's just keep this PR small and trackable :)

yuanming-hu:
Cool, thanks. Merging this in now...

taichi/backends/opengl/opengl_api.cpp
archibate: Ok.

PR #716, opened 2020-04-06
[Lang] Support `continue` on all backends

k-ye:
Related issue = #708 

### DIFF
--- a/taichi/codegen/codegen_llvm.cpp
+++ b/taichi/codegen/codegen_llvm.cpp
@@ -89,7 +89,10 @@ class CodeGenStmtGuard {
     setter_(saved_stmt_);
   }
 
- protected:
+  CodeGenStmtGuard(CodeGenStmtGuard &&) = default;
+  CodeGenStmtGuard &operator=(CodeGenStmtGuard &&) = default;
+
+ private:
   llvm::BasicBlock *saved_stmt_;
   Setter setter_;
 };
--- a/taichi/codegen/codegen_llvm.cpp
+++ b/taichi/codegen/codegen_llvm.cpp
@@ -31,8 +31,7 @@ void OffloadedTask::compile() {
   func = (task_fp_type)kernel_symbol;
 }
 
-// FunctionCreationGuard
-
+// TODO(k-ye): Hide FunctionCreationGuard inside cpp file
 FunctionCreationGuard::FunctionCreationGuard(
     CodeGenLLVM *mb,
     std::vector<llvm::Type *> arguments)
@@ -74,6 +73,42 @@ FunctionCreationGuard::~FunctionCreationGuard() {
     mb->entry_block = old_entry;
   }
 }
+
+namespace {
+
+class CodeGenStmtGuard {
+ public:
+  using Getter = std::function<llvm::BasicBlock *(void)>;
+  using Setter = std::function<void(llvm::BasicBlock *)>;
+
+  explicit CodeGenStmtGuard(Getter getter, Setter setter)
+      : saved_stmt_(getter()), setter_(std::move(setter)) {
+  }
+
+  ~CodeGenStmtGuard() {
+    setter_(saved_stmt_);
+  }
+
+ protected:
+  llvm::BasicBlock *saved_stmt_;
+  Setter setter_;
+};
+
+CodeGenStmtGuard make_loop_reentry_guard(CodeGenLLVM *cg) {
+  return CodeGenStmtGuard([cg]() { return cg->current_loop_reentry; },
+                          [cg](llvm::BasicBlock *saved_stmt) {
+                            cg->current_loop_reentry = saved_stmt;
+                          });
+}
+
+CodeGenStmtGuard make_while_after_loop_guard(CodeGenLLVM *cg) {
+  return CodeGenStmtGuard([cg]() { return cg->current_while_after_loop; },
+                          [cg](llvm::BasicBlock *saved_stmt) {
+                            cg->current_while_after_loop = saved_stmt;
+                          });
+}
+
+}  // namespace
 
 // CodeGenLLVM
 
@@ -664,29 +699,44 @@ void CodeGenLLVM::visit(ConstStmt *stmt) {
 void CodeGenLLVM::visit(WhileControlStmt *stmt) {
   BasicBlock *after_break =
       BasicBlock::Create(*llvm_context, "after_break", func);
-  TI_ASSERT(while_after_loop);
+  TI_ASSERT(current_while_after_loop);
   auto cond =
       builder->CreateICmpEQ(llvm_val[stmt->cond], tlctx->get_constant(0));
-  builder->CreateCondBr(cond, while_after_loop, after_break);
+  builder->CreateCondBr(cond, current_while_after_loop, after_break);
   builder->SetInsertPoint(after_break);
+}
+
+void CodeGenLLVM::visit(ContinueStmt *stmt) {
+  if (stmt->as_return()) {
+    builder->CreateRetVoid();
+  } else {
+    TI_ASSERT(current_loop_reentry != nullptr);
+    builder->CreateBr(current_loop_reentry);
+  }
+  // Stmts after continue are useless, so we switch the insertion point to
+  // /dev/null. In LLVM IR, the "after_continue" label shows "No predecessors!".
+  BasicBlock *after_continue =
+      BasicBlock::Create(*llvm_context, "after_continue", func);
+  builder->SetInsertPoint(after_continue);
 }
 
 void CodeGenLLVM::visit(WhileStmt *stmt) {
   BasicBlock *body = BasicBlock::Create(*llvm_context, "while_loop_body", func);
   builder->CreateBr(body);
   builder->SetInsertPoint(body);
+  auto lrg = make_loop_reentry_guard(this);
+  current_loop_reentry = body;
 
   BasicBlock *after_loop =
       BasicBlock::Create(*llvm_context, "after_while", func);
-  auto old_while_after_loop = while_after_loop;
-  while_after_loop = after_loop;
+  auto walg = make_while_after_loop_guard(this);
+  current_while_after_loop = after_loop;
 
   stmt->body->accept(this);
 
   builder->CreateBr(body);  // jump to head
 
   builder->SetInsertPoint(after_loop);
-  while_after_loop = old_while_after_loop;
 }
 
 llvm::Value *CodeGenLLVM::cast_pointer(llvm::Value *val,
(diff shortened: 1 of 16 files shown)


taichi/codegen/codegen_llvm.cpp
```diff
@@ -664,29 +699,44 @@ void CodeGenLLVM::visit(ConstStmt *stmt) {
...
-  TI_ASSERT(while_after_loop);
+  TI_ASSERT(current_while_after_loop);
   auto cond =
       builder->CreateICmpEQ(llvm_val[stmt->cond], tlctx->get_constant(0));
-  builder->CreateCondBr(cond, while_after_loop, after_break);
+  builder->CreateCondBr(cond, current_while_after_loop, after_break);
   builder->SetInsertPoint(after_break);
 }
 
+void CodeGenLLVM::visit(ContinueStmt *stmt) {
```
k-ye, line 709: @yuanming-hu for LLVM

taichi/codegen/codegen_opengl.cpp
```diff
@@ -671,6 +671,14 @@ class KernelGen : public IRVisitor {
     emit("if ({} == 0) break;", stmt->cond->short_name());
   }
 
+  void visit(ContinueStmt *stmt) override {
```
k-ye, line 674: @archibate for OpenGL

taichi/transforms/offload.cpp
```diff
@@ -414,9 +482,10 @@ OffloadedResult offload(IRNode *root) {
     PromoteIntermediate::run(root, result.local_to_global_offset);
     PromoteLocals::run(root, result.local_to_global_offset);
   }
-  irpass::insert_gc(root);
-  irpass::typecheck(root);
-  irpass::re_id(root);
+  insert_gc(root);
+  AssociateContinueScope::run(root);
```
k-ye, line 486: I put this pass here so that all backends get this step for free before they are integrated with #700 . Do you think we should create a new pass for this in the future?

taichi/ir/ir.h
```diff
@@ -940,6 +941,23 @@ class WhileControlStmt : public Stmt {
...
+ public:
+  // This is the loop on which this continue stmt has effects. It can be either
+  // an offloaded task, or a for/while loop inside the kernel.
+  Stmt *scope;
+
+  ContinueStmt() : scope(nullptr) {
+    TI_STMT_REG_FIELDS;
+  }
+
+  bool as_return() const {
```
k-ye, line 954: @yuanming-hu I just realized that this will require us to encapsulate the actual kernel body as a function (LLVM  does this, but Metal doesn't). Otherwise when we use one thread to handle a few elements, this could prematurely terminate the entire kernel thread.

taichi/codegen/codegen_llvm.cpp
yuanming-hu: This part looks good to me! Thanks!

taichi/ir/ir.h
archibate: Using `continue` as a shortcut of `return` in top-level? Nice idea!

> Alternatively, do you folks think it even makes sense to support continue at the top level? I mean Taichi users are supposed to well understand that the top-level loop as parallelized, yet continue makes sense only for serial loops.

We want to keep this easy-to-use shortcut since **they acts same as the expected in python**.
In comparison, we don't want `break` in top-level is because **it can't act same as the expected in python**.

taichi/transforms/ir_printer.cpp
```diff
@@ -202,6 +206,14 @@ class IRPrinter : public IRVisitor {
     print("while control {}, {}", stmt->mask->name(), stmt->cond->name());
   }
 
+  void visit(ContinueStmt *stmt) override {
+    if (stmt->scope) {
+      print("{} continue (as_return={})", stmt->name(), stmt->as_return());
```
archibate, line 211: This `(as_return={})` seems unnecessary since we always know it's as_return by seeing it's in the top-level.

taichi/ir/ir.h
k-ye: > The other option is to directly translate this into a continue in the source backends. But we have to be very careful when doing that...

I guess this means we assume the kernel body is embedded inside a "meta" for-loop implemented by the Taichi runtime: https://github.com/taichi-dev/taichi/blob/58c64baf24b33898d217154d0026543e2aa0bc2e/taichi/runtime/llvm/runtime.cpp#L1074-L1077

Then I think wrapping the kernel into a function is a much more robust and cleaner approach...

taichi/transforms/ir_printer.cpp
yuanming-hu: That's true, but for debugging the IR we may want to display as much information about the statements as possible.

tests/python/test_continue.py
```diff
@@ -0,0 +1,84 @@
...
+def test_unconditional_continue():
+    x = ti.var(ti.i32, shape=n)
+
+    @ti.kernel
+    def run():
+        # Launch just one thread
+        for _ in range(1):
+            for j in range(n):
+                continue
+                # pylint: disable=unreachable
```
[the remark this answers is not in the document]
k-ye: Yep. Actually I didn't notice "too many issues", but maybe that's because I'm adapted to ignore them..

taichi/transforms/offload.cpp
k-ye: Ack, i plan to make it a new pass once we all switch to this.. https://github.com/taichi-dev/taichi/blob/213a3c7c02823a95d67ec2c9cf562c7f0389746c/taichi/transforms/compile_to_offloads.cpp#L7

tests/python/test_continue.py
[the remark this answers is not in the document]
yuanming-hu: Sounds good :-) Let's plan to adopt the two linters in v0.7.0: https://github.com/taichi-dev/taichi/issues/677

taichi/ir/ir.h
k-ye: OK, sounds like I should keep this support. Thanks guys :)

taichi/transforms/ir_printer.cpp
k-ye: Would displaying `stmt->scope->name()` more useful? On the other hand, you can also just eyeball to figure out which loop the continue acts on..

taichi/ir/ir.h
yuanming-hu: > Then I think wrapping the kernel into a function is a much more robust and cleaner approach...

Right, then I guess the source-to-source backends have to support something like non-linear code emission since now parallel loop bodies are outlined into functions. Pragmatically maybe using a `continue` is still the easiest way.

taichi/transforms/simplify.cpp
```diff
@@ -1066,6 +1066,10 @@ class BasicBlockSimplify : public IRVisitor {
     return;
   }
 
+  void visit(ContinueStmt *stmt) override {
+    return;
+  }
+
   static bool is_global_write(Stmt *stmt) {
     return stmt->is<GlobalStoreStmt>() || stmt->is<AtomicOpStmt>();
```
archibate, line 1074: Btw, is `AtomicOpStmt` always about global variables?

taichi/codegen/codegen_opengl.cpp
archibate: Thanks! Test passing.

taichi/transforms/ir_printer.cpp
yuanming-hu: > Would displaying `stmt->scope->name()` more useful?

I think that would be useful. In general, it will be a good idea to ensure we can easily reconstruct the IRNodes from the printed IR string. We might need to implement an IR parser in the future.

taichi/transforms/simplify.cpp
yuanming-hu: Before offloading it can also be used on locals. E.g.

```python
@ti.kernel
def test():
  a = 0
  for i in range(100): # parallel loop
    a += i # atomic add
```

k-ye: I think it's either SNode variables or global temporary variables. I forgot whether if it supports arg variables..

taichi/ir/ir.h
```diff
@@ -940,6 +941,23 @@ class WhileControlStmt : public Stmt {
...
+    return scope->is<OffloadedStmt>();
```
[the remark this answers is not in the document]
k-ye: Yep, done

taichi/transforms/ir_printer.cpp
k-ye: OK, fixed.

taichi/ir/ir.h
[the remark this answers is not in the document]
archibate: We may also want `TI_ASSERT(scope->task_type == range_for || scope->task_type == struct_for)` :)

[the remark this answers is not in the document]
k-ye: Done, let's write as defensively as possible..

taichi/transforms/ir_printer.cpp
```diff
@@ -208,7 +208,7 @@ class IRPrinter : public IRVisitor {
 
   void visit(ContinueStmt *stmt) override {
     if (stmt->scope) {
-      print("{} continue (as_return={})", stmt->name(), stmt->as_return());
+      print("{} continue (scope={})", stmt->name(), stmt->name());
```
archibate, line 211: Will that be good if we also add this to `FrontendBreakStmt`?

taichi/ir/ir.h
k-ye: > Right, then I guess the source-to-source backends have to support something like non-linear code emission since now parallel loop bodies are outlined into functions.

+1

> Pragmatically maybe using a continue is still the easiest way.

Right. I just realized a problem on Metal. Its `range_for` doesn't use grid-stride loop, yet its `struct_for` does, gahhh... 😦

taichi/transforms/ir_printer.cpp
k-ye: Yeah i think so. Let's do it in another PR

taichi/transforms/simplify.cpp
archibate: I thought arg atomics is also supported...? What type of variables are not supported with atomic op?

taichi/transforms/ir_printer.cpp
k-ye: I just realized you said `FrontendBreakStmt`, maybe more useful to add this for `BreakStmt`?

k-ye:
(May be related to #689 ) This came up when I was writing this PR, would it be good if we split `FrontendStmt` and `Expression` into its own, higher layer of IR? I think the only pass that cares about both sets of IRs is `lower_ast`?

yuanming-hu:
Right, ideally FrontendIR and Expressions are unique to the Taichi language, which should be translated into the CHI IR/execution engine via `lower_ast`.

This separation will improve modularity since now the frontend and middle-end IR (CHI) are separate into two modules. In fact, `lower_ast` should not even use `IRVisitor` (which is for the middle-end IR). We should ideally have a FrontendASTVisitor...

PR #717, opened 2020-04-06
[OpenGL] 64-bit data type support

archibate:
Related issue = #664 

Currently I'm just adding float64 atomics... and didn't work.


archibate:
Why I can't done atomic of float64? Because I found `floatBitsToInt` not work for `int64`.
Maybe I want to only support atomic of float64 when `GL_NV_shader_atomic_float64` built-in instrins found?

archibate:
To test (didn't work well for now):
```py
import taichi as ti

ti.init(arch=ti.opengl)

x = ti.var(dt=ti.i64, shape=())

@ti.kernel
def func():
    x[None] = 233

func()
print(x[None])
```
and good night, Zzz..

archibate:
Fixed, ready for merge!

archibate:
Of course!

PR #764, opened 2020-04-13
[cli] ti test --arch argument not work, fix it

archibate:
Related issue = #... (if any)


archibate:
Not working because pytest doesn't pass the same taichi instance to forked children task now.

archibate:
> Cool! If I understand correcetly, you are assuming all the child process forked by pytest will inherit the `TI_WANTED_ARCHS` envvar?

Yes, I believe.

PR #790, opened 2020-04-15
[doc] tell arch users to install ncurses5-compat-libs for libtinfo5

archibate:
Related issue = #791

### DIFF
--- a/docs/dev_install.rst
+++ b/docs/dev_install.rst
@@ -28,7 +28,8 @@ Installing Depedencies
 
   .. code-block:: bash
 
-    sudo pacman -S ncurses5-compat-libs  # /usr/lib/libtinfo.so.5
+    sudo pacman -S yaourt
+    yaourt -S ncurses5-compat-libs  # /usr/lib/libtinfo.so.5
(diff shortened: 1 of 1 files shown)


docs/dev_install.rst
```diff
@@ -28,6 +28,7 @@ Installing Depedencies
 
   .. code-block:: bash
 
+    sudo pacman -S ncurses5-compat-libs  # /usr/lib/libtinfo.so.5
```
archibate, line 31: Actually this is an package in only `archlinuxcn`:
```
[bate@archit gpu]$ sudo pacman -Ss libtinfo5
archlinuxcn/ncurses5-compat-libs 6.1-1
    System V Release 4.0 curses emulation library, ABI 5
```
Should I tell them to add archlinuxcn source to their `/etc/pacman.conf` to install this dep, even if they're not chinese at all?

archibate:
Long terms short, what libtinfo5 does? can we remove libtinfo5 dep?

archibate:
Replaced by #791.

docs/dev_install.rst
archibate: Fixed by using `yaourt`, install from aur/ncurse5-compat-libs.

archibate:
Let's go for #791 attempt.

PR #796, opened 2020-04-16
[example] Add an Euler equation solver

KLozes:
Here's a compressible Euler equation solver! Its just a simple finite volume scheme with a HLLC Riemann solver. It already creates some nice images, but needs to be polished up a bit. I'll work on that tomorrow night :)

Related issue = #776 

### DIFF
--- a/examples/euler.py
+++ b/examples/euler.py
@@ -9,9 +9,11 @@ N = 512 # grid resolution
 CFL = .8
 IC_type = 0  # 0:sod
 BC_type = 0 # 0:walls
-img_field = 0 # 0:density, 1: schlieren, 2:vorticity, 3: velocity mag, 4: temperature
+img_field = 1 # 0:density, 1: schlieren, 2:vorticity, 3: velocity mag
 res = 1024  # gui resolution
-cmap_name = 'magma_r' # python colormap
+cmap_name = 'jet' # python colormap
+use_fixed_caxis = 0 # 1: use fixed caxis limits, 0: automatic caxis limits
+fixed_caxis = [0.0, 5.0] # fixed caxis limits
 
 
 Q = ti.Vector(4, dt=real, shape=(N,N)) # [rho, rho*u, rho*v, rho*e] consv vars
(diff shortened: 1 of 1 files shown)


yuanming-hu:
WOW! As a graphics person, I've been mainly playing with incompressible fluid simulations - i.e. those with a Chorin-style projection. I didn't realize compressible fluids can have stunning visual results as well! Thanks for adding this.

The implementation looks great to me. Just one nit: would it be possible to remove the dependency on `matplotlib`?

Another quick question: does the Taichi implementation with the CUDA backend have a comparable performance to https://www.youtube.com/watch?v=k72_YeNO7yU? I'm just curious about the current CUDA backend performance in Taichi.

examples/euler.py
```diff
@@ -0,0 +1,329 @@
...
+img = ti.var(dt=ti.f32, shape=(res, res))
+
+gamma = 1.4  # ratio of specific heats
+h = 1.0 / (N - 2)  # cell size
+vol = h * h  # cell volume
+
+
+@ti.func
+def is_interior_cell(i, j):
+    return i > 0 and i < N - 1 and j > 0 and j < N - 1
```
yuanming-hu, line 35: Taichi supports `0 < i < N - 1` :-)

KLozes: Oh neat! I had no idea

KLozes:
Yeah! The shocks and acoustic waves really add beautiful visuals that you cant really get in incompressible flow. I'm also excited to improve this with some better discretization schemes and AMR. See https://youtu.be/txk-VO1hzBY?t=2079. 

I can get rid of the matplotlib dependency. I guess I'll just make a colormap manually. Perhaps we could add colormap as an optional argument to set_image()? We could create our own library of colormaps if we need to. 

I'll try time them both tonight and let you know. If everything is working right with Taichi, I think it should easily beat my CUDA version.

archibate:
> I'll try time them both tonight and let you know. If everything is working right with Taichi, I think it should easily beat my CUDA version.

Many thanks! If possible, could you also try timing against `ti.opengl`? CUDA somtimes mysteriously run 2x slower than OpenGL according to @yuanming-hu.

yuanming-hu:
> > I'll try time them both tonight and let you know. If everything is working right with Taichi, I think it should easily beat my CUDA version.
> 
> Many thanks! If possible, could you also try timing against `ti.opengl`? CUDA sometimes mysteriously run 2x slower than OpenGL... according to @yuanming-hu.

I tried OpenGL, yet it seems to get stuck at compilation...

yuanming-hu:
> I can get rid of the matplotlib dependency. I guess I'll just make a colormap manually. Perhaps we could add colormap as an optional argument to set_image()? We could create our own library of colormaps if we need to.

If that's > 5 lines of code let's simply do something like
```python
try:
  import matplotlib
except:
  print('Please pip install matplotlib`)
  exit(-1)
```
Installing `matplotlib` is no big deal, and it's a nice demonstration on how to use `taichi` with other libraries anyway.



> I'll try time them both tonight and let you know. If everything is working right with Taichi, I think it should easily beat my CUDA version.

Thanks! I think this grid-based solver can be a great benchmark for `taichi`. Having timings on a handwritten CUDA reference is also awesome. We can even compare the lines of code in both implementations - this will definitely attract CDF people who are still hand-writting CUDA or using matlab.

archibate:
> I tried OpenGL, yet it seems to get stuck at compilation...

It's not stuck, it's just taking a **very long time** for NV GL to compile, once a shader is compiled, the next time you launch eular.py won't take time since NV will cache it's compile result. (yes, even if you exit python, NV GL keep your shader caches!)

KLozes:
I think this is ready! I also added a fancy modern discretization scheme that seems to have almost 4x higher resolution than the classical muscl method :). I'll work on some timing comparisons tomorrow. heres a nice pic!
euler_equations

PR #834, opened 2020-04-21
[Lang] Add math constants like `ti.pi`

archibate:
Related issue = #831 

### DIFF
--- a/python/taichi/misc/constants.py
+++ b/python/taichi/misc/constants.py
@@ -1,0 +1,19 @@
+import math
+
+pi = math.pi
+pi_2 = pi / 2
+pi_4 = pi / 4
+tau = math.tau
+E = math.e
+ln2 = math.log(2)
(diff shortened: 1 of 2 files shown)


python/taichi/misc/constants.py
```diff
@@ -0,0 +1,19 @@
+import math
+
+pi = math.pi
+pi_2 = pi / 2
+pi_4 = pi / 4
+tau = math.tau
+E = math.e
```
archibate, line 7: I use uppercase `E` due to the fear of name conflict, WDYT?

archibate:
But it's convenient, right? No extra `import math` required. And `ti.pi` is shorter than `math.pi`.
> The only reason all three modules provide a pi value is so if you are using just one of the three modules, you can conveniently have access to pi without having to import another module. They're not providing different values for pi.

PR #839, opened 2020-04-21
[IR] Use JIT compilation/evaluation for systematic constant folding 

archibate:
Related issue = #510 

We now support more operation types rather than just add and mul.
Also add a lot of unary ops rather than just casting.


yuanming-hu:
Cool! @xumingkuan could you help review this? I'm recently swamped in too many todo items... I think the ultimate solution is to use JIT while doing constant fold.

taichi/transforms/constant_fold.cpp
```diff
@@ -1,9 +1,72 @@
...
+  PER_OP(log, std::log)
+  PER_OP(exp, std::exp)
+  PER_OP(abs, std::abs)
+  PER_OP(sin, std::sin)
+  PER_OP(cos, std::cos)
+  PER_OP(tan, std::tan)
+  PER_OP(asin, std::asin)
+  PER_OP(acos, std::acos)
+  PER_OP(tanh, std::tanh)
+  PER_OP(rsqrt, 1 / std::sqrt)
```
xumingkuan, line 24: These are incongruent with `this->val_i32 = o(rhs.val_i32);`... We need to be cautious about the return type of unary operators.

xumingkuan:
One reason I didn't implement #510 is that constant folding has a low priority for now. Our python frontend already does constant folding if users write something like `a = 2 * 3`, and there are not many constants to fold if there are not many other optimization passes producing constants.

Would you like to help to migrate to JIT as this comment https://github.com/taichi-dev/taichi/issues/510#issuecomment-589479296?

xumingkuan:
For the failing test, please take a look at `test_simplify.cpp::simplify_linearized_with_trivial_inputs` and try uncommenting the print statements.

taichi/transforms/constant_fold.cpp
```diff
@@ -1,9 +1,72 @@
...
+    case BinaryOpType::op: this->val_i32 = lhs.val_i32 o rhs.val_i32; break;
+#define PER_OF(op, f) \
+    case BinaryOpType::op: this->val_i32 = f(lhs.val_i32, rhs.val_i32); break;
+  switch (op) {
+  PER_OP(add, +)
+  PER_OP(sub, -)
+  PER_OP(mul, *)
+  PER_OP(div, /)
+  // PER_OP(truediv, /) // XXX: is this same as div?
+  // PER_OP(floordiv, /) // XXX: do we have std::floordiv?
```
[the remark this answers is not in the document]
archibate: Oh, I see, this is i/i=i so all are same, should consider more if i/i=f or f/f=f.

archibate:
> JIT Compile the operation with operand types into a function and invoke that function to do constant fold

I'd like to. Is this `function` mean invoked at runtime or compile-time?

xumingkuan:
> > JIT Compile the operation with operand types into a function and invoke that function to do constant fold
> 
> I'd like to. Is this `function` mean invoked at runtime or compile-time?

This "function" is invoked at compile-time. For example, when we visit `a + b` in the constant folding pass, we compile a function that does `a + b`, and then replace the `BinaryOpStmt` with the result.

archibate:
So all `BinaryOpStmt` are replaced by `FuncCallStmt`? How could this help constant folding?

xumingkuan:
> So all `BinaryOpStmt` are replaced by `FuncCallStmt`? How could this help constant folding?

Oh no, it's not that. I think the "function" should take two `ConstStmt`s and one `BinaryOp` as input (or, take one `ConstStmt` and one `UnaryOp`), and return one `ConstStmt`.

archibate:
> > So all `BinaryOpStmt` are replaced by `FuncCallStmt`? How could this help constant folding?
> 
> Oh no, it's not that. I think the "function" should take two `ConstStmt`s and one `BinaryOp` as input (or, take one `ConstStmt` and one `UnaryOp` as input), and return one `ConstStmt`.

This is basically what `TypedConstant::from_binary_op` does? What's the difference?

archibate:
```
kernel {
  <gen*x1> $72 = get root
  <i32 x1> $79 = const [0]
  <gen*x1> $74 = [S2root][root]::lookup($72, $79) activate = false
  <gen*x1> $75 = get child [S2root->S3dense] $74
  <i32 x1> $81 = const [1]
  <i32 x1> $87 = const [0]
  <i32 x1> $88 = const [0]
  <i32 x1> $84 = const [4]
  <i32 x1> $89 = const [0]
  <i32 x1> $90 = const [0]
  <gen*x1> $78 = [S3dense][dense]::lookup($75, $90) activate = true
}
kernel {
  <gen*x1> $72 = get root
  <i32 x1> $79 = const [0]
  <gen*x1> $74 = [S2root][root]::lookup($72, $79) activate = false
  <gen*x1> $75 = get child [S2root->S3dense] $74
  <i32 x1> $90 = const [0]
  <gen*x1> $78 = [S3dense][dense]::lookup($75, $90) activate = true
}
```
DIE didn't elimate $90, should combine $90 with $79.

xumingkuan:
The difference is that the "function" should be just like compiling a *kernel*, if I understand @yuanming-hu 's idea correctly, so that we can reuse our `typecheck` and other passes to deal with various types. 

On the other hand, `TypedConstant::from_binary_op` needs us to write every combination of types if we want to support all types.

xumingkuan:
> ```
> kernel {
>   <gen*x1> $72 = get root
>   <i32 x1> $79 = const [0]
>   <gen*x1> $74 = [S2root][root]::lookup($72, $79) activate = false
>   <gen*x1> $75 = get child [S2root->S3dense] $74
>   <i32 x1> $81 = const [1]
>   <i32 x1> $87 = const [0]
>   <i32 x1> $88 = const [0]
>   <i32 x1> $84 = const [4]
>   <i32 x1> $89 = const [0]
>   <i32 x1> $90 = const [0]
>   <gen*x1> $78 = [S3dense][dense]::lookup($75, $90) activate = true
> }
> kernel {
>   <gen*x1> $72 = get root
>   <i32 x1> $79 = const [0]
>   <gen*x1> $74 = [S2root][root]::lookup($72, $79) activate = false
>   <gen*x1> $75 = get child [S2root->S3dense] $74
>   <i32 x1> $90 = const [0]
>   <gen*x1> $78 = [S3dense][dense]::lookup($75, $90) activate = true
> }
> ```
> 
> DIE didn't elimate $90, should combine $90 with $79.

I'm not quite sure why this happens, but one more `simplify` pass can eliminate one of the `const [0]`s.

archibate:
> one more simplify pass can eliminate one of the const [0]s.

It did and passed, should we add this to the compile_to_offloads?

archibate:
> The difference is that the "function" should be just like compiling a _kernel_, if I understand @yuanming-hu 's idea correctly, so that we can reuse our `typecheck` and other passes to deal with various types.
> 
> On the other hand, `TypedConstant::from_binary_op` needs us to write every combination of types if we want to support all types.

You mean, we've already write every combinations in type_check, so we can reuse it? by something like `class ConstantFold : TypeCheckVisitor`?

xumingkuan:
> > one more simplify pass can eliminate one of the const [0]s.
> 
> It did and passed, should we add this to the compile_to_offloads?

We've already had enough `simplify` passes in `compile_to_offloads`.

xumingkuan:
> > The difference is that the "function" should be just like compiling a _kernel_, if I understand @yuanming-hu 's idea correctly, so that we can reuse our `typecheck` and other passes to deal with various types.
> > On the other hand, `TypedConstant::from_binary_op` needs us to write every combination of types if we want to support all types.
> 
> You mean, we've already write every combinations in type_check, so we can reuse it? by something like `class ConstantFold : TypeCheckVisitor`?

No... `type_check` is just a pass in `compile_to_offloads`. Maybe this is more similar to what I mean: *compile* this kernel when we visit a `BinaryOpStmt`.
```
$1 = const [2]
$2 = const [3]
$3 = mul $1 $2
(something like return $3)
```

But it's probably not trivially calling `compile_to_offloads` recursively: this will cause dead loops in the constant folding pass.

archibate:
What if `BinaryOpStmt` has something other than `ConstStmt` as oprand? Like `2 + (3 + 4)`. How does it extinguish `2 + (3 + 4)` and `2 + (3 + x)`? It can't be pre-seen without recursion.

> But it's probably not trivially calling compile_to_offloads recursively: this will cause dead loops in the constant folding pass.

Let's **not call simplification passes** in these constant folding functions :)

xumingkuan:
> What if `BinaryOpStmt` has something other than `ConstStmt` as oprand? Like `2 + (3 + 4)`.
> How does it distinguish `2 + (3 + 4)` and `2 + (3 + x)`? It can't be pre-seen without recursion.
> For the `2 + (3 + x)`, do we want:
> 
> ```
> $1 = const [2]
> $2 = const [3]
> $3 = expr [x]  # ????
> $4 = add $2 $3
> $5 = add $1 $4
> return $5
> ```
> 
> ?
> 
> > But it's probably not trivially calling compile_to_offloads recursively: this will cause dead loops in the constant folding pass.
> 
> Let's **not call simplification passes** for these constant folding functions :)

We don't have any `expr`s in the constant folding pass... They only live in the frontend AST.

xumingkuan:
For `2 + (3 + 4)`, we may need to JIT compile two times.

Anything other than `ConstStmt` should not be folded. For example, `2 + (3 + x)` is **not** equivalent to `5 + x`.

xumingkuan:
> > But it's probably not trivially calling compile_to_offloads recursively: this will cause dead loops in the constant folding pass.
> 
> Let's **not call simplification passes** for these constant folding functions :)

I'm not sure if this is feasible and if the code will be maintainable. Could you please describe your solution if you have already come up with one?

xumingkuan:
> Maybe this is more similar to what I mean: _compile and launch_ this kernel when we visit a `BinaryOpStmt`.
> 
> ```
> <i32 x1> $1 = const [2]
> <i32 x1> $2 = const [3]
> <i32 x1> $3 = mul $1 $2
> (something like return $3)
> ```
> 
> But it's probably not trivially calling `compile_to_offloads` recursively: this will cause dead loops in the constant folding pass.

@yuanming-hu May I take you some minutes to verify if I understood your idea of "JIT Compile the operation with operand types into a function and invoke that function to do constant fold" correctly in the quote above?

yuanming-hu:
> > Maybe this is more similar to what I mean: _compile and launch_ this kernel when we visit a `BinaryOpStmt`.
> > ```
> > <i32 x1> $1 = const [2]
> > <i32 x1> $2 = const [3]
> > <i32 x1> $3 = mul $1 $2
> > (something like return $3)
> > ```
> > 
> > 
> > But it's probably not trivially calling `compile_to_offloads` recursively: this will cause dead loops in the constant folding pass.
> 
> @yuanming-hu May I take you some minutes to verify if I understood your idea of "JIT Compile the operation with operand types into a function and invoke that function to do constant fold" correctly in the quote above?

Yes, that's basically the idea. We need `compile-cache-launch`. Let's say you have a BinaryOp, generate a kernel like
```
a = GetArg 0
b = GetArg 0
c = a OP b
return c
```
(See the implementation of SNode readers for how to implement return. The const folder kernel would look pretty much the same as SNode accessor kernels.)

and launch this kernel with your operands in `ConstStmt` to evaluate it. To prevent too many kernel compilation, cache the kernels with the same Op and operand data type.

xumingkuan:
> Yes, that's basically the idea. We need `compile-cache-launch`. Let's say you have a BinaryOp, generate a kernel like
> 
> ```
> a = GetArg 0
> b = GetArg 0
> c = a OP b
> return c
> ```
> 
> (See the implementation of SNode readers for how to implement return. The const folder kernel would look pretty much the same as SNode accessor kernels.)
> 
> and launch this kernel with your operands in `ConstStmt` to evaluate it. To prevent too many kernel compilation, cache the kernels with the same Op and operand data type.

Thanks for the quick reply! Given that `a` and `b` are `GetArg`s, this kernel will not be constant folded again.

archibate:
> > > But it's probably not trivially calling compile_to_offloads recursively: this will cause dead loops in the constant folding pass.
> > 
> > 
> > Let's **not call simplification passes** for these constant folding functions :)
> 
> I'm not sure if this is feasible and if the code will be maintainable. Could you please describe your solution if you have already come up with one?

We can have an option pass to compile_to_offload:
```
// In constant_fold:
compile_to_offload(ir, optimized=false);
// In main code:
compile_to_offload(ir, optimized=true);
```

archibate:
> and launch this kernel with your operands in ConstStmt to evaluate it. To prevent too many kernel compilation, cache the kernels with the same Op and operand data type.

You mean, tiny-kernels like snode_read/write? (which is also cached)

xumingkuan:
> We can have an option pass to compile_to_offload:
> 
> ```
> // In constant_fold:
> compile_to_offload(ir, optimized=false);
> // In main code:
> compile_to_offload(ir, optimized=true);
> ```

Let's use this kernel instead:

> ```
> a = GetArg 0
> b = GetArg 0
> c = a OP b
> return c
> ```

Without *arguments*, it's not a *function*.

xumingkuan:
> > and launch this kernel with your operands in ConstStmt to evaluate it. To prevent too many kernel compilation, cache the kernels with the same Op and operand data type.
> 
> You mean, tiny-kernels like snode_read/write? (which is also cached)

Yes, I think.

archibate:
> Without arguments, it's not a function. (Actually it still is a function, but not reusable)

We should make all `ConstStmt` arguments? So it's reusable.

yuanming-hu:
> > and launch this kernel with your operands in ConstStmt to evaluate it. To prevent too many kernel compilation, cache the kernels with the same Op and operand data type.
> 
> You mean, tiny-kernels like snode_read/write? (which is also cached)

Yes

xumingkuan:
> > Without arguments, it's not a function. (Actually it still is a function, but not reusable)
> 
> We should make all `ConstStmt` arguments? So it's reusable.

Exactly! Just need to make the arguments have the same type with the `ConstStmt`s.

archibate:
> > > Without arguments, it's not a function. (Actually it still is a function, but not reusable)
> > 
> > 
> > We should make all `ConstStmt` arguments? So it's reusable.
> 
> Exactly! Just need to make the arguments have the same type with the `ConstStmt`s.

Do we already have argument type-templating infrastructure in snode_rw? We can set up a general tiny-kernel-launchers including both snode_rw and jit_const_fold.
I gradually understand what this mean now, will do these this night to prevent a brain fry.

xumingkuan:
> > > > Without arguments, it's not a function. (Actually it still is a function, but not reusable)
> > > 
> > > 
> > > We should make all `ConstStmt` arguments? So it's reusable.
> > 
> > 
> > Exactly! Just need to make the arguments have the same type with the `ConstStmt`s.
> 
> Do we already have argument type-templating infrastructure in snode_rw? We can set up a general tiny-kernel-launchers including both snode_rw and jit_const_fold.
> I gradually understand what this mean now, will do these this night to prevent a brain fry.

I guess we don't need to set up a new launcher -- it would be fine to just imitate 
`Kernel &Program::get_snode_reader(SNode *snode)`.

taichi/transforms/constant_fold.cpp
```diff
@@ -45,8 +45,8 @@ bool TypedConstant::from_binary_op(BinaryOpType op,
   PER_OP(sub, -)
   PER_OP(mul, *)
   PER_OP(div, /)
-  // PER_OP(truediv, /) // XXX: is this same as div?
-  // PER_OP(floordiv, /) // XXX: do we have std::floordiv?
+  PER_OP(truediv, /) // XXX: is this same as div?
```
xumingkuan, line 48: I think `truediv` is something like `(double)a / b` where `a` and `b` are `int`s in C++... Anyway, after we upgrade to JIT, these will become unnecessary :)

archibate: Maybe we should check return type...if return type is f32, assert it's not truediv.

archibate:
How do I add IRs to a kernel?

xumingkuan:
> How do I add IRs to a kernel?

Maybe like this?
https://github.com/taichi-dev/taichi/blob/a03fa13139c8f48eead8af8d41e1518a659be30c/taichi/program/program.cpp#L446

taichi/transforms/constant_fold.cpp
```diff
@@ -90,10 +142,65 @@ class ConstantFold : public BasicStmtVisitor {
...
+
+  ConstantFoldJIT() : BasicStmtVisitor() {
+  }
+
+  void visit(UnaryOpStmt *stmt) override {
+    uops.push_back(stmt);
+  }
+
+  void visit(BinaryOpStmt *stmt) override {
+    bops.push_back(stmt);
```
xumingkuan, line 159: Please notice that we can only fold statements with all operands being `ConstStmt`s :)

And therefore, ultimately, we need to run it iteratively like `ConstantFold::run(IRNode *node)`.

archibate:
```py
import taichi as ti

ti.init(debug=True)
ti.set_gdb_trigger(False)
x = ti.var(ti.i32)

ti.root.dynamic(ti.i, 16, 4).place(x)

@ti.kernel
def func():
    x[3] = 1

func()
```
Why it fails?
`RuntimeError: [program.cpp:check_runtime_error@315] Assertion failure: Accessing Tensor of Size [16] with indices (3)`

archibate:
With cfd:
```
kernel {
  $0 = offloaded  {
    <i32 x1> $1 = const [0]
    2 : assert $1, "Accessing Tensor of Size [3] with indices (%d)", $1
    <f32 x1> $3 = const [0.0]
    <gen*x1> $4 = get root
    <gen*x1> $5 = [S0root][root]::lookup($4, $1) activate = false
    <gen*x1> $6 = get child [S0root->S1dense] $5
    <i32 x1> $7 = bit_extract($1 + 0, 0~2)
    <gen*x1> $8 = [S1dense][dense]::lookup($6, $7) activate = false
    <f32*x1> $9 = get child [S1dense->S2place_f32] $8
    <f32*x1> $10 : global store [$9 <- $3]
  }
}
```
without:
```
kernel {
  $0 = offloaded  {
    <i32 x1> $1 = const [0]
    <i32 x1> $2 = const [1]
    <i32 x1> $3 = cmp_ge $1 $1
    <i32 x1> $4 = const [3]
    <i32 x1> $5 = cmp_lt $1 $4
    <i32 x1> $6 = bit_and $3 $5
    <i32 x1> $7 = bit_and $2 $6
    8 : assert $7, "Accessing Tensor of Size [3] with indices (%d)", $1
    <f32 x1> $9 = const [0.0]
    <gen*x1> $10 = get root
    <gen*x1> $11 = [S0root][root]::lookup($10, $1) activate = false
    <gen*x1> $12 = get child [S0root->S1dense] $11
    <i32 x1> $13 = bit_extract($1 + 0, 0~2)
    <gen*x1> $14 = [S1dense][dense]::lookup($12, $13) activate = false
    <f32*x1> $15 = get child [S1dense->S2place_f32] $14
    <f32*x1> $16 : global store [$15 <- $9]
  }
}
```

taichi/transforms/constant_fold.cpp
```diff
@@ -44,28 +116,49 @@ class ConstantFold : public BasicStmtVisitor {
...
 
+  static bool jit_from_binary_op(TypedConstant &tc, BinaryOpType op,
+      const TypedConstant &lhs, const TypedConstant &rhs)
+  {
+    if (lhs.dt != DataType::i32 || rhs.dt != DataType::i32 || tc.dt != DataType::i32)
+      return false;
+    auto kernel_name = fmt::format("jit_constexpr_{}", 0);
+    auto func = [&] () {
+      auto lhstmt =
+        Stmt::make<ConstStmt>(LaneAttribute<TypedConstant>(lhs));
```
xumingkuan, line 127: Let's use `ArgLoadStmt` for caching the JIT kernels.

If we compile a kernel for each statement that can be constant folded, it will be very time-consuming --- we'll have to compile O(n) kernels (where n is the number of statements).

On the other hand, if we cache the kernels for each combination of types, the number of kernels we need to compile will be a constant. This will reduce compile time, especially for large programs.

taichi/program/compile_config.h
```diff
@@ -33,6 +33,7 @@ struct CompileConfig {
   bool fast_math;
   bool use_unified_memory;
   bool async;
+  bool no_cp2o;
```
xumingkuan, line 36: If `ArgLoadStmt` is used, this will become unnecessary. I think this boolean variable is not a good design.

Of course, feel free to use these expedient variables for debugging before merging into master :)

taichi/transforms/constant_fold.cpp
```diff
@@ -1,9 +1,81 @@
 #include "taichi/ir/ir.h"
+#include "taichi/program/program.h"
+#include "taichi/ir/snode.h"
 #include <deque>
 #include <set>
+#include <cmath>
 
 TLANG_NAMESPACE_BEGIN
 
+bool TypedConstant::from_unary_op(UnaryOpType op, const TypedConstant &rhs) {
```
xumingkuan, line 10: Please remove this after completing the JIT constant folding pass.

archibate:
crash on test_out_of_bound, will fix tmr, good night!

archibate:
Do we eliminate statements like `assert(0)`? So that no assertion on constant indices.

archibate:
Why I seperately test, it passed. When combine together in `test_assert.py`, it failed?

xumingkuan:
> Do we eliminate statements like `assert(1)`? So that no assertion on constant indices.

Not yet, but it's easy for me to write this optimization.

archibate:
todo: map key by (op, dt1, dt2)

archibate:
How do I use `std::unordered_map` for custom object like `BinaryEvaluatorId` or `std::pair`?

archibate:
```
[I 04/26/20 17:37:18.395] [constant_fold.cpp:jit_from_binary_op@89] !!!IN
kernel {
  $0 = offloaded  {
    <i32 x1> $1 = arg[1]
    <i32 x1> $2 = arg[2]
    <i32 x1> $3 = cmp_ge $1 $2
    <i32 x1> $4 : store arg 0 <- $3
  }
}
Fatal Python error: Segmentation fault

Current thread 0x00007ffad0a53740 (most recent call first):
  File "/home/bate/Develop/taichi/python/taichi/lang/kernel.py", line 389 in func__
  File "/home/bate/Develop/taichi/python/taichi/lang/kernel.py", line 415 in __call__
  File "/home/bate/Develop/taichi/python/taichi/lang/kernel.py", line 484 in wrapped
  File "/home/bate/Develop/taichi/tests/python/test_assert.py", line 51 in test_out_of_bound
  File "/home/bate/Develop/taichi/python/taichi/lang/__init__.py", line 445 in test
  File "/home/bate/Develop/taichi/python/taichi/lang/__init__.py", line 455 in func__
  File "/usr/lib/python3.8/site-packages/_pytest/python.py", line 167 in pytest_pyfunc_call
  File "/usr/lib/python3.8/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/usr/lib/python3.8/site-packages/pluggy/manager.py", line 84 in <lambda>
  File "/usr/lib/python3.8/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/usr/lib/python3.8/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/usr/lib/python3.8/site-packages/_pytest/python.py", line 1445 in runtest
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 134 in pytest_runtest_call
  File "/usr/lib/python3.8/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/usr/lib/python3.8/site-packages/pluggy/manager.py", line 84 in <lambda>
  File "/usr/lib/python3.8/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/usr/lib/python3.8/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 210 in <lambda>
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 237 in from_call
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 209 in call_runtest_hook
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 185 in call_and_report
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 99 in runtestprotocol
  File "/usr/lib/python3.8/site-packages/_pytest/runner.py", line 84 in pytest_runtest_protocol
  File "/usr/lib/python3.8/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/usr/lib/python3.8/site-packages/pluggy/manager.py", line 84 in <lambda>
  File "/usr/lib/python3.8/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/usr/lib/python3.8/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/usr/lib/python3.8/site-packages/_pytest/main.py", line 271 in pytest_runtestloop
  File "/usr/lib/python3.8/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/usr/lib/python3.8/site-packages/pluggy/manager.py", line 84 in <lambda>
  File "/usr/lib/python3.8/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/usr/lib/python3.8/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/usr/lib/python3.8/site-packages/_pytest/main.py", line 247 in _main
  File "/usr/lib/python3.8/site-packages/_pytest/main.py", line 197 in wrap_session
  File "/usr/lib/python3.8/site-packages/_pytest/main.py", line 240 in pytest_cmdline_main
  File "/usr/lib/python3.8/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/usr/lib/python3.8/site-packages/pluggy/manager.py", line 84 in <lambda>
  File "/usr/lib/python3.8/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/usr/lib/python3.8/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/usr/lib/python3.8/site-packages/_pytest/config/__init__.py", line 92 in main
  File "/home/bate/Develop/taichi/python/taichi/main.py", line 56 in test_python
  File "/home/bate/Develop/taichi/python/taichi/main.py", line 173 in main
  File "/home/bate/Develop/taichi/bin/ti", line 4 in <module>
[E 04/26/20 17:37:18.400] Received signal 11 (Segmentation fault)


段错误 (核心已转储)
```
Why compiling a simple jit kernel cause segf?

archibate:
Hi, I'm applying `ArgLoadStmt` now, but I don't dear to go forward without solving the segf.

archibate:
I found that if I enable cache: 1 of 2 fails; if I disable cache 2 of 2 fails. Interesting...

archibate:
I realized: thank you unique_ptr.

archibate:
Maybe I should move `cache` into `class Program`?

archibate:
Fixed now! Btw, why I see `Merging is blocked`?

archibate:
AppVeyor:
```
================================== FAILURES ===================================
__________________________ tests/python/test_abs.py ___________________________
[gw0] win32 -- Python 3.6.8 C:\Python36-x64\python.exe
worker 'gw0' crashed while running 'tests/python/test_abs.py::test_abs'
=========================== short test summary info ===========================
FAILED tests/python/test_abs.py::test_abs
================== 1 failed, 346 passed in 377.39s (0:06:17) ==================
Command exited with code 1
```
I see no such failure on my local machine.

archibate:
Can we make `pytest` retry crashed test twice for stability? It would make less contributors' confuse and help us to determine whether a commit is working or not correctly.

archibate:
NEXT: also apply to UnaryOpStmt.

archibate:
Segv on `simplify_linearized_with_trivial_inputs`:
```
Running C++ tests...

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test is a Catch v2.0.1 host application.
Run with -? for options                                                                      

-------------------------------------------------------------------------------
alg_simp
  simplify_multiply_zero_fast_math
-------------------------------------------------------------------------------
/home/bate/Develop/taichi/tests/cpp/test_alg_simp.cpp:69
...............................................................................

/home/bate/Develop/taichi/tests/cpp/test_alg_simp.cpp:117: FAILED:
  CHECK( block->size() == 8 )
with expansion:
  10 == 8

/home/bate/Develop/taichi/tests/cpp/test_alg_simp.cpp:125: FAILED:
  CHECK( block->size() == 4 )
with expansion:
  10 == 4

[I 04/27/20 00:59:48.368] [constant_fold.cpp:get_binary_op_jit_eval_kernel@71] HEL
-------------------------------------------------------------------------------
simplify
  simplify_linearized_with_trivial_inputs
-------------------------------------------------------------------------------
/home/bate/Develop/taichi/tests/cpp/test_simplify.cpp:9
...............................................................................

/home/bate/Develop/taichi/tests/cpp/test_simplify.cpp:9: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal

===============================================================================
test cases:   6 |   4 passed | 2 failed
assertions: 161 | 158 passed | 3 failed

[E 04/27/20 00:59:48.368] Received signal 11 (Segmentation fault)


***********************************
* Taichi Compiler Stack Traceback *                                                          
***********************************                                                          
/tmp/taichi-9cezs9ld/taichi_core.so: taichi::Logger::error(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool)                                  
/tmp/taichi-9cezs9ld/taichi_core.so: taichi::signal_handler(int)                             
/usr/lib/libc.so.6(+0x3bd70) [0x7fe8bcf2ad70]                                                
/tmp/taichi-9cezs9ld/taichi_core.so: taichi::lang::ConstantFold::get_binary_op_jit_eval_kernel(taichi::lang::ConstantFold::BinaryEvaluatorId const&)                                      
/tmp/taichi-9cezs9ld/taichi_core.so: taichi::lang::ConstantFold::visit(taichi::lang::BinaryOpStmt*)                                                                                       
/tmp/taichi-9cezs9ld/taichi_core.so: taichi::lang::BasicStmtVisitor::visit(taichi::lang::Block*)                                                                                          
/tmp/taichi-9cezs9ld/taichi_core.so: taichi::lang::irpass::constant_fold(taichi::lang::IRNode*)                                                                                           
/tmp/taichi-9cezs9ld/taichi_core.so(+0x8d5508) [0x7fe8ad368508]                              
/tmp/taichi-9cezs9ld/taichi_core.so: Catch::RunContext::invokeActiveTestCase()               
/tmp/taichi-9cezs9ld/taichi_core.so: Catch::RunContext::runCurrentTest(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&)                                             
/tmp/taichi-9cezs9ld/taichi_core.so: Catch::RunContext::runTest(Catch::TestCase const&)      
/tmp/taichi-9cezs9ld/taichi_core.so: Catch::Session::runInternal()                           
/tmp/taichi-9cezs9ld/taichi_core.so: Catch::Session::run()                                   
/tmp/taichi-9cezs9ld/taichi_core.so: taichi::run_tests(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >)                                
/tmp/taichi-9cezs9ld/taichi_core.so: taichi::RunTests::run(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)
/tmp/taichi-9cezs9ld/taichi_core.so(+0x806d48) [0x7fe8ad299d48]
/tmp/taichi-9cezs9ld/taichi_core.so(+0x696cfc) [0x7fe8ad129cfc]
/usr/lib/libpython3.8.so.1.0: PyCFunction_Call
/usr/lib/libpython3.8.so.1.0: _PyObject_MakeTpCall
/usr/lib/libpython3.8.so.1.0(+0xfeb97) [0x7fe8bcc2db97]
/usr/lib/libpython3.8.so.1.0: _PyEval_EvalFrameDefault
/usr/lib/libpython3.8.so.1.0: _PyEval_EvalCodeWithName
/usr/lib/libpython3.8.so.1.0: _PyFunction_Vectorcall
/usr/lib/libpython3.8.so.1.0(+0x1e0aeb) [0x7fe8bcd0faeb]
/usr/lib/libpython3.8.so.1.0: PyObject_Call
/usr/lib/libpython3.8.so.1.0: _PyEval_EvalFrameDefault
/usr/lib/libpython3.8.so.1.0: _PyFunction_Vectorcall
/usr/lib/libpython3.8.so.1.0: _PyEval_EvalFrameDefault
/usr/lib/libpython3.8.so.1.0: _PyEval_EvalCodeWithName
/usr/lib/libpython3.8.so.1.0: _PyFunction_Vectorcall
/usr/lib/libpython3.8.so.1.0: _PyEval_EvalFrameDefault
/usr/lib/libpython3.8.so.1.0: _PyEval_EvalCodeWithName
/usr/lib/libpython3.8.so.1.0: PyEval_EvalCode
/usr/lib/libpython3.8.so.1.0(+0x2668c8) [0x7fe8bcd958c8]
/usr/lib/libpython3.8.so.1.0(+0x26aba3) [0x7fe8bcd99ba3]
/usr/lib/libpython3.8.so.1.0: PyRun_FileExFlags
/usr/lib/libpython3.8.so.1.0: PyRun_SimpleFileExFlags
/usr/lib/libpython3.8.so.1.0: Py_RunMain
/usr/lib/libpython3.8.so.1.0: Py_BytesMain
/usr/lib/libc.so.6: __libc_start_main
/usr/bin/python3: _start
```
Have to sleep to prevent a class drop now, good night!

xumingkuan:
> Can we make `pytest` retry crashed test twice for stability? It would make less contributors' confuse and help us to determine whether a commit is working or not correctly.

This is probably the same issue as #828...

I don't think we should make `pytest` retry: in most cases, if CI fails, a commit is not working correctly. For me, I run `ti test` (or `ti test abs` if only `test_abs.py` is failed) whenever I find my commit fails on CI, and if it passes, I realize that it's probably that issue. I think we can just trigger CI again in this case...

taichi/transforms/constant_fold.cpp
```diff
@@ -42,30 +47,86 @@ class ConstantFold : public BasicStmtVisitor {
...
+
+  struct BinaryEvaluatorId
+  {
+    BinaryOpType op;
+    DataType ret, lhs, rhs;
+
+    explicit operator int() const // make STL map happy
+    {
+      return (int)op | (int) ret << 8 | (int) lhs << 16 | (int) rhs << 24;
+    }
```
xumingkuan, line 61: You may use something like this (this is for `std::pair`; in our case, you may change the hash function in the following code to something you like, and `template` is not necessary):
```C++
struct hash_pair { 
    template <class T1, class T2> 
    size_t operator()(const pair<T1, T2>& p) const
    { 
        auto hash1 = hash<T1>{}(p.first); 
        auto hash2 = hash<T2>{}(p.second); 
        return hash1 ^ hash2; 
    } 
}; 
unordered_map<pair<int, int>, bool, hash_pair> um; 
```

taichi/transforms/constant_fold.cpp
```diff
@@ -42,30 +47,86 @@ class ConstantFold : public BasicStmtVisitor {
...
+  static bool jit_from_binary_op(TypedConstant &ret, BinaryOpType op,
+      const TypedConstant &lhs, const TypedConstant &rhs)
+  {
+    BinaryEvaluatorId id{op, ret.dt, lhs.dt, rhs.dt};
+    auto *ker = get_binary_op_jit_eval_kernel(id);
+    auto &ctx = get_current_program().context;
+#define PER_TYPE(x) ctx.set_arg(1, lhs.val_##x);
+    PER_TYPE(i32) PER_TYPE(i64)
+    PER_TYPE(f32) PER_TYPE(f64)
+#undef PER_TYPE
```
xumingkuan, line 106: I don't understand if this is the same as
```C++
ctx.set_arg(1, lhs_val_i32);
ctx.set_arg(1, lhs_val_i64);
```
Won't it be set multiple times?

taichi/transforms/constant_fold.cpp
```diff
@@ -42,30 +47,86 @@ class ConstantFold : public BasicStmtVisitor {
...
+#define PER_TYPE(x) ctx.set_arg(2, rhs.val_##x);
+    PER_TYPE(i32) PER_TYPE(i64)
+    PER_TYPE(f32) PER_TYPE(f64)
+#undef PER_TYPE
```
xumingkuan, line 110: I would suggest extracting it into a function like `void set_arg(const TypedConstant&)` and avoid the duplicated code.

In that function, I think it would be clearer to avoid using macros. Writing `if` 4 times for 4 types is fine.

xumingkuan:
I understand that debugging may take more time than writing code. You may use the stack traceback, the auto-triggered gdb (if you're using Linux), `irpass::print` and `irpass::analysis::verify` to locate the bug in one statement, and then in most cases, you'll find why there the bug is. Don't assume the bug is in the master branch unless you locate it somewhere you didn't modify in the PR -- at least all tests pass in the master branch.

Two months ago, as I'm using Windows and I'm not familiar with gdb, and taichi didn't give me the stack traceback on Windows, and I haven't implemented `irpass::analysis::verify`, `SIGSEGV` was really frustrating. I believe with these tools, you'll be able to debug without my help.

archibate:
Thanks! I finally locate the problem: `get_current_program()` returns `nullptr`!!!

archibate:
May I create a `Program` for `test_cpp`? Without `Program`, how can we launch JIT kernels?!

yuanming-hu:
> May I create a `Program` for `test_cpp`? Without `Program`, how can we launch JIT kernels?!

Good idea. We should assume a `Program` and `Kernel` always exist when manipulating IRs. Also, the global `get_current_program` should be removed (however that is nontrivial refactoring and should be done in a separate PR)...

archibate:
> > May I create a `Program` for `test_cpp`? Without `Program`, how can we launch JIT kernels?!
> 
> Good idea. We should assume a `Program` and `Kernel` always exist when manipulating IRs. Also, the global `get_current_program` should be removed (however that is nontrivial refactoring and should be done in a separate PR)...

Why remove it? We can make `get_current_program` per-thread to enable real multi-thread testing instead of multi-process testing.

xumingkuan:
Glad to see you're making progress! 

I don't know how to create a `Program` like `ti.init` either, but this error may be related to LLVM modules (which I'm also not familiar with):
https://github.com/taichi-dev/taichi/blob/3206dbd52d1d0a380e37a7d80a6ad852090fc71f/taichi/llvm/llvm_context.cpp#L636

xumingkuan:
Another solution can be just don't write tests for constant folding in C++... Just write in python with `ti.init(print_ir=True)`, and compare the result with/without the constant folding pass.

archibate:
> Another solution can be just don't write tests for constant folding in C++... Just write in python with `ti.init(print_ir=True)`, and compare the result with/without the constant folding pass.

That sounds less systematic, also what if we apply JIT method to other C++ passes? Cancel them too? Not sounding very good imo.

xumingkuan:
> > Another solution can be just don't write tests for constant folding in C++... Just write in python with `ti.init(print_ir=True)`, and compare the result with/without the constant folding pass.
> 
> That sounds less systematic, also what if we apply JIT method to other C++ passes? Cancel them too? Not sounding very good imo.

Yes, create a `Program` is the better solution. But if you're not sure if the constant folding pass is correct and you see some errors, it will be hard to find if the bug is in the constant folding pass or in creating the `Program`. I think you can write some python tests to make sure that the constant folding pass is working correctly in various cases, while (probably) waiting for @yuanming-hu to give you some advice about "how to create a full-working-well `Program` like python `ti.init`".

xumingkuan:
I'm going to sleep now. Good night!

archibate:
All done except for the `module` nrf!

archibate:
Could you help me to find a way to initialize `Program`? Many thanks!!

archibate:
Where is `main_thread_data` initialized!?

archibate:
get_this_thread_module -> main_thread_data <- get_this_thread_data <- per_thread_data

archibate:
I added `ti.init()` in `test_cpp()` to fix, but got another:
```
alg_simp
  simplify_multiply_zero_fast_math
-------------------------------------------------------------------------------
/home/bate/Develop/taichi/tests/cpp/test_alg_simp.cpp:71
...............................................................................

/home/bate/Develop/taichi/tests/cpp/test_alg_simp.cpp:71: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  bad optional access

[I 04/29/20 00:11:48.957] [llvm_context.cpp:get_this_thread_data@624] create!!!!!!!!!
[I 04/29/20 00:11:48.957] [llvm_context.cpp:get_this_thread_data@626] !!!!!!!!!
[I 04/29/20 00:11:48.961] [llvm_context.cpp:get_this_thread_data@624] create!!!!!!!!!
[I 04/29/20 00:11:48.961] [llvm_context.cpp:get_this_thread_data@626] !!!!!!!!!
[I 04/29/20 00:11:48.961] [constant_fold.cpp:run@170] 1
[I 04/29/20 00:11:48.961] [constant_fold.cpp:run@174] 2
kernel {
  $88 = arg[1]
  $89 = arg[2]
  $90 = mul $88 $89
  $91 : store arg 0 <- $90
}
[I 04/29/20 00:11:48.961] [constant_fold.cpp:jit_from_binary_op@110] k
-------------------------------------------------------------------------------
simplify
  simplify_linearized_with_trivial_inputs
-------------------------------------------------------------------------------
/home/bate/Develop/taichi/tests/cpp/test_simplify.cpp:9
...............................................................................

/home/bate/Develop/taichi/tests/cpp/test_simplify.cpp:9: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  bad optional access
```
Why bad? Why optional???

archibate:
F  I  X  E  D  !

archibate:
TODO: context return bad.

archibate:
When running jit_evaluator on test_cpp, the return value is random, seems context broken?

xumingkuan:
Looks like something not initialized or undefined behavior...

archibate:
I added an `assert(0)` to the kernel, and no error shows. Seems the kernel isn't executed at all?

archibate:
I'm giving up this unless someone could point out my problem solver. Or I had to remove constant_fold test.

yuanming-hu:
> I'm giving up this unless someone could point out my problem solver. Or I had to remove constant_fold test.

Really sorry about my absence this week. I'm rushing for a quarterly report. I'll have more time this weekend, and will definitely take a look at this PR. Thanks for your hard work!

yuanming-hu:
We are on our way to abandoning `ArgStoreStmt`, and return values are now stored in a specific buffer instead of part of the context. The old `ArgStoreStmt` strategy is an abuse of the arguments that are supposed to be constants.

It's too late in my timezone now but I'll detail this tomorrow. See the snode reader for how to fetch the results.

archibate:
> We are on our way to abandoning `ArgStoreStmt`, and return values are now stored in a specific buffer instead of part of the context. The old `ArgStoreStmt` strategy is an abuse of the arguments that are supposed to be constants.
> 
> It's too late in my timezone now but I'll detail this tomorrow. See the snode reader for how to fetch the results.

So my problem maybe, `ArgStoreStmt` is not working on LLVM now? And it works on OpenGL because I didn't update for that abandon? Thank for you info.

archibate:
On Thu, Apr 30, 2020 at 09:38:04PM -0700, Yuanming Hu wrote:
> We are on our way to abandoning `ArgStoreStmt`, and return values are now stored in a specific buffer instead of part of the context. The old `ArgStoreStmt` strategy is an abuse of the arguments that are supposed to be constants.
> 
> It's too late in my timezone now but I'll detail this tomorrow. See the snode reader for how to fetch the results.
> 
> -- 
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub:
> https://github.com/taichi-dev/taichi/pull/839#issuecomment-622244367

I still see `FrontendArgStoreStmt` in snode_reader...

archibate:
On Thu, Apr 30, 2020 at 09:38:04PM -0700, Yuanming Hu wrote:
> We are on our way to abandoning `ArgStoreStmt`, and return values are now stored in a specific buffer instead of part of the context. The old `ArgStoreStmt` strategy is an abuse of the arguments that are supposed to be constants.
> 
> It's too late in my timezone now but I'll detail this tomorrow. See the snode reader for how to fetch the results.
> 
> -- 
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub:
> https://github.com/taichi-dev/taichi/pull/839#issuecomment-622244367

test reply.

archibate:
On Thu, Apr 30, 2020 at 09:38:04PM -0700, Yuanming Hu wrote:
> We are on our way to abandoning `ArgStoreStmt`, and return values are now stored in a specific buffer instead of part of the context. The old `ArgStoreStmt` strategy is an abuse of the arguments that are supposed to be constants.
> 
> It's too late in my timezone now but I'll detail this tomorrow. See the snode reader for how to fetch the results.
> 
> -- 
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub:
> https://github.com/taichi-dev/taichi/pull/839#issuecomment-622244367

test

yuanming-hu:
> On Thu, Apr 30, 2020 at 09:38:04PM -0700, Yuanming Hu wrote: We are on our way to abandoning `ArgStoreStmt`, and return values are now stored in a specific buffer instead of part of the context. The old `ArgStoreStmt` strategy is an abuse of the arguments that are supposed to be constants. It's too late in my timezone now but I'll detail this tomorrow. See the snode reader for how to fetch the results. -- You are receiving this because you were assigned. Reply to this email directly or view it on GitHub: [#839 (comment)](https://github.com/taichi-dev/taichi/pull/839#issuecomment-622244367)
> I still see `FrontendArgStoreStmt` in snode_reader...

Right, but note that we are transitioning to something like a `KernelReturnStmt` #909 

In fact, in LLVM it's already implemented so:

https://github.com/taichi-dev/taichi/blob/f5373b1395a66c78506de0fd3c172dffe0d444d0/taichi/codegen/codegen_llvm.cpp#L867-L882

archibate:
> I took a look at your implementation and it looks pretty neat :-) I think we are very close, once #917 is done.

Great, so we won't solve/merge this until #917 is done?

archibate:
Could you tell me if I could deal with this without #917? Let's not making this PR huge like #535.

yuanming-hu:
> Could you tell me if I could deal with this without #917? Let's not making this PR huge like #535.

Sure, if you would go without #917, please see here:

https://github.com/taichi-dev/taichi/blob/2fa77350e336491159da0609e5220b7fe6b5572d/taichi/ir/snode.cpp#L137

archibate:
Let's first make backends unified to `result_buffer` and `KernelReturnStmt` now.

archibate:
```
__________________________________ test_type_tensor64[dt0] __________________________________
[gw2] linux -- Python 3.8.2 /usr/bin/python3

dt = DataType.int64

    @pytest.mark.parametrize('dt', _TI_64_TYPES)
    @ti.require(ti.extension.data64)
    @ti.archs_excluding(ti.opengl)
    def test_type_tensor64(dt):
>       _test_type_tensor(dt)

tests/python/test_types.py:88: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/python/test_types.py:74: in _test_type_tensor
    func(i, j)
python/taichi/lang/kernel.py:500: in wrapped
python/taichi/lang/kernel.py:431: in __call__
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = (0, 0), tmps = [], callbacks = [], actual_argument_slot = 2

    def func__(*args):
        assert len(args) == len(
            self.arguments), '{} arguments needed but {} provided'.format(
                len(self.arguments), len(args))
    
        tmps = []
        callbacks = []
    
        actual_argument_slot = 0
        for i, v in enumerate(args):
            needed = self.arguments[i]
            if isinstance(needed, template):
                continue
            provided = type(v)
            # Note: do not use sth like needed == f32. That would be slow.
            if id(needed) in real_type_ids:
                if not isinstance(v, (float, int)):
                    raise KernelArgError(i, needed, provided)
                t_kernel.set_arg_float(actual_argument_slot, float(v))
            elif id(needed) in integer_type_ids:
                if not isinstance(v, int):
                    raise KernelArgError(i, needed, provided)
                t_kernel.set_arg_int(actual_argument_slot, int(v))
            elif self.match_ext_arr(v, needed):
                dt = to_taichi_type(v.dtype)
                has_torch = has_pytorch()
                is_numpy = isinstance(v, np.ndarray)
                if is_numpy:
                    tmp = np.ascontiguousarray(v)
                    tmps.append(tmp)  # Purpose: do not GC tmp!
                    t_kernel.set_arg_nparray(actual_argument_slot,
                                             int(tmp.ctypes.data),
                                             tmp.nbytes)
                else:
    
                    def get_call_back(u, v):
                        def call_back():
                            u.copy_(v)
    
                        return call_back
    
                    assert has_torch and isinstance(v, torch.Tensor)
                    tmp = v
                    taichi_arch = self.runtime.prog.config.arch
    
                    if str(v.device).startswith('cuda'):
                        # External tensor on cuda
                        if taichi_arch != taichi_lang_core.Arch.cuda:
                            # copy data back to cpu
                            host_v = v.to(device='cpu', copy=True)
                            tmp = host_v
                            callbacks.append(get_call_back(v, host_v))
                    else:
                        # External tensor on cpu
                        if taichi_arch == taichi_lang_core.Arch.cuda:
                            gpu_v = v.cuda()
                            tmp = gpu_v
                            callbacks.append(get_call_back(v, gpu_v))
                    t_kernel.set_arg_nparray(
                        actual_argument_slot, int(tmp.data_ptr()),
                        tmp.element_size() * tmp.nelement())
                shape = v.shape
                max_num_indices = taichi_lang_core.get_max_num_indices()
                assert len(
                    shape
                ) <= max_num_indices, "External array cannot have > {} indices".format(
                    max_num_indices)
                for i, s in enumerate(shape):
                    t_kernel.set_extra_arg_int(actual_argument_slot, i, s)
            else:
                assert False
            actual_argument_slot += 1
        # Both the class kernels and the plain-function kernels are unified now.
        # In both cases, |self.grad| is another Kernel instance that computes the
        # gradient. For class kerenls, args[0] is always the kernel owner.
        if not self.is_grad and self.runtime.target_tape and not self.runtime.inside_complex_kernel:
            self.runtime.target_tape.insert(self, args)
    
>       t_kernel()
E       RuntimeError: [program.h:get_current_kernel@148] current_kernel
```
Why??

archibate:
https://github.com/taichi-dev/taichi/blob/31a925126371fd4888453e11c9bc2d6bb8240eb0/taichi/program/program.h#L132-L137
What a smart design... So I have to initialize this? And what more have I do to compile a kernel?

yuanming-hu:
TBH I don't think that design is smart - we should remove `current_kernel` later. For now, we set `current_kernel` only when we are defining/compiling the kernel:
https://github.com/taichi-dev/taichi/blob/31a925126371fd4888453e11c9bc2d6bb8240eb0/taichi/program/kernel.cpp#L25

https://github.com/taichi-dev/taichi/blob/31a925126371fd4888453e11c9bc2d6bb8240eb0/taichi/program/kernel.cpp#L43

archibate:
On Fri, May 08, 2020 at 05:41:28AM -0700, Yuanming Hu wrote:
> TBH I don't think that design is smart - we should remove `current_kernel` later. For now, we set `current_kernel` only when we are defining/compiling the kernel:
> https://github.com/taichi-dev/taichi/blob/31a925126371fd4888453e11c9bc2d6bb8240eb0/taichi/program/kernel.cpp#L25
> 
> https://github.com/taichi-dev/taichi/blob/31a925126371fd4888453e11c9bc2d6bb8240eb0/taichi/program/kernel.cpp#L43
> 
> -- 
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub:
> https://github.com/taichi-dev/taichi/pull/839#issuecomment-625796490

So I simply removed `current_kernel = nullptr` and the problem gone, it's a `smart` design anyway. (and encounter another problem then)

archibate:
For now I just push current_kernel into stack to make cfd work, see 8c8b3e4.

archibate:
Have problems with non-32-bit types like i8 and f64 in test_types.py, too tired have to fix tmr.

archibate:
I found that it's `debug = True` make cfd not working well:
```py
import taichi as ti

ti.init(debug=True) # OK when set to `False`

@ti.kernel
def func():
    n = 15
    assert n >= 0

func()
```

archibate:
@yuanming-hu Hello? All local test passing and I believe the CI test will pass too. Therefore the PR is ready for review now.

tests/cpp/test_alg_simp.cpp
```diff
@@ -108,11 +114,13 @@ TI_TEST("alg_simp") {
 
     irpass::typecheck(block.get());  // insert 2 casts
     TI_CHECK(block->size() == 10);
+    // irpass::print(block.get());
```
yuanming-hu, line 117: Shall we remove these debug prints?

tests/cpp/test_simplify.cpp
```diff
@@ -35,6 +37,7 @@ TI_TEST("simplify") {
     irpass::constant_fold(block.get());
     irpass::alg_simp(block.get(), CompileConfig());
     irpass::die(block.get());  // should eliminate consts
+    irpass::simplify(block.get());
     // irpass::print(block.get());
```
yuanming-hu, line 41: Let's remove this as well (although not introduced by you... thanks!)

taichi/transforms/constant_fold.cpp
```diff
@@ -11,61 +14,161 @@ class ConstantFold : public BasicStmtVisitor {
...
+
+    explicit operator int() const // make STL map happy
+    {
+      return (int)op | (int)!!is_binary << 7
+        | (int) ret << 8 | (int) lhs << 16 | (int) rhs << 24;
+    }
+
+    UnaryOpType unary_op() const
+    {
+      return (UnaryOpType) op;
```
yuanming-hu, line 31: Maybe `TI_ASSERT(!is_binary)`?

taichi/transforms/constant_fold.cpp
```diff
@@ -11,61 +14,161 @@ class ConstantFold : public BasicStmtVisitor {
...
+    }
+
+    BinaryOpType binary_op() const
+    {
+      return (BinaryOpType) op;
```
yuanming-hu, line 36: Same here, maybe an assert will make this function safer.

taichi/transforms/constant_fold.cpp
```diff
@@ -11,61 +14,161 @@ class ConstantFold : public BasicStmtVisitor {
...
       stmt->parent->erase(stmt);
       throw IRModified();
-    } else if (stmt->op_type == BinaryOpType::mul) {
-      new_constant.val_int32() =
-          lhs->val[0].val_int32() * rhs->val[0].val_int32();
+    }
+  }
+
+  void visit(UnaryOpStmt *stmt) override {
+    auto lhs = stmt->operand->cast<ConstStmt>();
```
yuanming-hu, line 164: `lhs` -> `operand`?

taichi/transforms/constant_fold.cpp
```diff
@@ -11,61 +14,161 @@ class ConstantFold : public BasicStmtVisitor {
...
+    explicit ContextArgSaveGuard(Context &ctx_) : ctx(ctx_) {
+      std::memcpy(old_args, ctx.args, sizeof(old_args));
     }
+
+    ~ContextArgSaveGuard() {
+      std::memcpy(ctx.args, old_args, sizeof(old_args));
+    }
+  };
+
+  static bool jit_from_binary_op(TypedConstant &ret, BinaryOpStmt *stmt,
```
yuanming-hu, line 107: `jit_evaluate_binary_op`?

taichi/transforms/constant_fold.cpp
```diff
@@ -11,61 +14,161 @@ class ConstantFold : public BasicStmtVisitor {
...
+    ContextArgSaveGuard _(ctx); // save input args, prevent override current kernel
+    ctx.set_arg<int64_t>(0, lhs.val_i64);
+    ctx.set_arg<int64_t>(1, rhs.val_i64);
+    irpass::print(ker->ir);
+    (*ker)();
+    ret.val_i64 = get_current_program().fetch_result<int64_t>(0);
+    return true;
+  }
+
+  static bool jit_from_unary_op(TypedConstant &ret, UnaryOpStmt *stmt,
```
yuanming-hu, line 127: `jit_evaluate_unary_op`?

taichi/transforms/constant_fold.cpp
```diff
@@ -11,61 +14,161 @@ class ConstantFold : public BasicStmtVisitor {
...
+  {
+    if (!is_good_type(ret.dt))
+      return false;
+    JITEvaluatorId id{(int)stmt->op_type, ret.dt, lhs.dt, stmt->cast_type,
+      false};
+    auto *ker = get_jit_evaluator_kernel(id);
+    auto &ctx = get_current_program().get_context();
+    ContextArgSaveGuard _(ctx); // save input args, prevent override current kernel
+    ctx.set_arg<int64_t>(0, lhs.val_i64);
+    irpass::print(ker->ir);
```
yuanming-hu, line 138: Shall we remove this?

taichi/transforms/constant_fold.cpp
```diff
@@ -11,61 +14,161 @@ class ConstantFold : public BasicStmtVisitor {
...
+    if (!is_good_type(ret.dt))
+      return false;
+    JITEvaluatorId id{(int)stmt->op_type, ret.dt, lhs.dt, rhs.dt,
+      true};
+    auto *ker = get_jit_evaluator_kernel(id);
+    auto &ctx = get_current_program().get_context();
+    ContextArgSaveGuard _(ctx); // save input args, prevent override current kernel
+    ctx.set_arg<int64_t>(0, lhs.val_i64);
+    ctx.set_arg<int64_t>(1, rhs.val_i64);
+    irpass::print(ker->ir);
```
yuanming-hu, line 121: pls remove this?

taichi/transforms/constant_fold.cpp
```diff
@@ -11,61 +14,161 @@ class ConstantFold : public BasicStmtVisitor {
...
+  {
+    int op;
+    DataType ret, lhs, rhs;
+    bool is_binary;
+
+    explicit operator int() const // make STL map happy
+    {
+      return (int)op | (int)!!is_binary << 7
+        | (int) ret << 8 | (int) lhs << 16 | (int) rhs << 24;
+    }
```
yuanming-hu, line 27: Maybe we should use `JITEvaluatorId` itself as the key of the JIT evaluation kernel cache. You'll have to define a few more members for hashing and comparing two `JITEvaluatorId`, but that will make the system more robust.

archibate:
Thank for the review, I have to sleep to prevent a dreamless dream now, cutmr.

taichi/transforms/constant_fold.cpp
```diff
@@ -11,61 +14,161 @@ class ConstantFold : public BasicStmtVisitor {
...
+  };
+
+  static Kernel *get_jit_evaluator_kernel(JITEvaluatorId const &id)
+  {
+    auto &cache = get_current_program().jit_evaluator_cache;
+    int iid = int(id);
+    auto it = cache.find(iid);
+    if (it != cache.end()) // cached?
+      return it->second.get();
+    static int jic = 0; // X: race?
```
yuanming-hu, line 49: Let's make it an `std::atomic<uint64>` member of `Program`.

taichi/transforms/constant_fold.cpp
```diff
@@ -11,61 +14,161 @@ class ConstantFold : public BasicStmtVisitor {
...
+    ~ContextArgSaveGuard() {
+      std::memcpy(ctx.args, old_args, sizeof(old_args));
+    }
+  };
+
+  static bool jit_evaluate_binary_op(TypedConstant &ret, BinaryOpStmt *stmt,
+      const TypedConstant &lhs, const TypedConstant &rhs)
+  {
+    // ConstStmt of `bad` types like `i8` is not supported by LLVM.
+    // Dis: https://github.com/taichi-dev/taichi/pull/839#issuecomment-625902727
```
yuanming-hu, line 113: Let's move this comment to the definition of `is_good_type`.

yuanming-hu:
Crazy test case for your new constant folding pass:

Screenshot from 2020-05-09 14-31-06

If you compute the partial sum of the first 10 terms of this series, it will give you  `3.1415905109380797`. Is the constant folding pass now capable of folding that into pi? :-)

archibate:
```py
import taichi as ti
ti.init(print_ir=True)

@ti.kernel
def calc_pi() -> ti.f32:
    term = 1.0
    sum = 0.0
    divisor = 1
    for i in ti.static(range(10)):
        sum += term / divisor
        term *= -1/3
        divisor += 2
    return sum * ti.sqrt(12.0)

print(calc_pi())
```
```
[I 05/10/20 10:59:26.885] [compile_to_offloads.cpp:operator()@18] Simplified III:
kernel {
  $0 = offloaded  {
    <f32 x1> $1 = const [3.1415904]
    <f32 x1> $2 : kernel return $1
  }
}
3.1415903568267822
```
Yes :) we did.

taichi/transforms/constant_fold.cpp
archibate: Here I used `cache.size()` instead of `jic`, and I realized that we may need a mutex for `cache`?

taichi/transforms/constant_fold.cpp
archibate: I used to have this, but I give up because, it requires a `hash` structure that knows all member of `JITEvaluatorId`, while I want to keep `JITEvaluatorId` private in this file (instead of in header).

yuanming-hu: Sounds good. I think we are good for now, but keep in mind that if `lhs`, `rhs`, `ret` grows bigger than 256, then this hash value is no longer unique.

xumingkuan:
BTW, please let @taichi-gardener enforce code formatting before merging into master :)

taichi/transforms/constant_fold.cpp
archibate: `!!is_binary` prevents the situation that `is_binary` is none of 0 or 1, e.g. `is_binary = 2` will cause a serious problem here.

archibate:
https://github.com/taichi-dev/taichi/pull/839/commits/cf451f0d1ed64409276a8436f22e12c971f7d3c0 by @taichi-gardener is making too much unrelated changes, can make tracking bug harder, especially given that the PR is a highly dedicated since it involves IR. Please revert.
Btw, maybe we should add a GitHub hook to enforce a @taichi-gardener commit before every PR merges? So that no dirty code from last PR involved in this one.

taichi/codegen/codegen_llvm.cpp
```diff
@@ -691,6 +691,7 @@ void CodeGenLLVM::visit(ConstStmt *stmt) {
     llvm_val[stmt] = llvm::ConstantInt::get(
         *llvm_context, llvm::APInt(64, val.val_int64(), true));
   } else {
+    TI_P(data_type_name(val.dt));
```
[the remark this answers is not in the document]
archibate: Let's tell the user what type they involved that make `ConstStmt` broken before the following `TI_NOT_IMPLEMENTED`.

yuanming-hu:
> [cf451f0](https://github.com/taichi-dev/taichi/commit/cf451f0d1ed64409276a8436f22e12c971f7d3c0) by @taichi-gardener is making too much unrelated changes, can make tracking bug harder, especially given that the PR is a highly dedicated since it involves IR, please revert.
> Btw, maybe we should add a GitHub hook to enforce a @taichi-gardener commit before every PR merges? So that no dirty code from last PR involved in this one.
> For now let's make a separate PR to get rid of the dirty codes.

Right, having a GitHub action to reformat the code automatically would be great! I have no idea how that should be implemented, though...

taichi/transforms/constant_fold.cpp
xumingkuan: I think `(int)is_binary` can only have a value of 0 or 1.
https://stackoverflow.com/questions/2725044/can-i-assume-booltrue-int1-for-any-c-compiler

xumingkuan: BTW, if `rhs` is at least 128 (not 256), the result of `(int) rhs << 24` can be undefined (although I believe it's correct in most compilers)
https://stackoverflow.com/questions/7622/are-the-shift-operators-arithmetic-or-logical-in-c

archibate: How about `using JITEvaluatorIdType = uint32_t`?

xumingkuan: Sounds good.

archibate:
> Looks good! But it's a little bit inconsistent with the comment (256 should be 128).
> 
> I would suggest changing to this without modifying the comment:
> 
> ```c++
> return (JITEvaluatorIdType)ret | (JITEvaluatorIdType)lhs << 8
>         | (JITEvaluatorIdType)rhs << 16 | (JITEvaluatorIdType)op << 24
>         | (JITEvaluatorIdType)is_binary << 31;
> ```

Thanks! But I think we have more `Binary/UnaryOpType`'s than `DataType`'s, so let's make 256 for the `OpType`'s then.

archibate:
Great! I'll merge as soon as CI passed.

xumingkuan:
Benchmark:
**Before**:
benchmark20200510
**After**:
benchmark20200510_constant_fold
(@yuanming-hu Shall we add this pass to `advanced_optimization`?)

yuanming-hu:
Yes, let's make it part of advanced optimizations. Please also mention this in your 10-minute talk (and of course, make it clear that this is implemented by @archibate).

archibate:
> advanced_optimization

+1, how to add?

archibate:
Moved to adv now, btw, 怎么看benchmark的这个图表的?

xumingkuan:
> Moved to adv now, btw, 怎么看benchmark的这个图表的？~1%是一个好的结果吗？

Each bar represents the number of remaining statements of codegen of a unit test after all "advanced optimizations" divided by the number of statements without advanced optimizations.

1% is surely a good result -- #857 also reduced the number of statements by ~1%. #897 and #898 had optimization effect <1%.

PR #879, opened 2020-04-26
[misc] Fix `ti.cfg.gdb_trigger` not found

archibate:
Related PR id = #769

I made an error in #769: there is no `ti.cfg.gdb_trigger` at all!
Now the problem is where should `gdb_trigger` be, `CompileConfig` or `CoreState`?
The first seems only accessible only by compiler parts. So `get_current_program().config` stand for only compiler configuations so it's not found anywhere in `logger.cpp`?

### DIFF
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -142,7 +142,13 @@ def init(arch=None,
     environ_config("print_benchmark_stat")
     environ_config("device_memory_fraction", float)
     environ_config("device_memory_GB", float)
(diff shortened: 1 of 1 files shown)


archibate:
I realized that in `test_assert.py`, you set gdb trigger to `False`. And that is applied to other test threads too.

archibate:
CI passing! I'm merging this now to prevent other problem if users fork from master. Sorry for introducing an error in master branch.. I guess will soon be revoked from write-access if I keep merging failure codes into master...

PR #890, opened 2020-04-27
[Example] Add Multiple Importance Sampling to the cornell box scene

k-ye:
I added MIS to the cornell box example so as to denoise the scene. Below are the comparisons, with each running for 1000 iterations.

* Reference (img variance=0.1455)
ref_var=0 14548

* New code with MIS (img variance=0.1094)
new_var=0 10938

Here's a result after running 15K iterations:

new_iters=15k

---

The MIS is done by doing two samples per intersection. Note that this is carried out only when the ray hits a lambertian material (i.e. the boudaries)

* One ray that shoots toward a random point on the light source (`sample_area_light`)
* One ray that samples from a cosine hemisphere distribution (`sample_brdf`).

This is basically what the PBR book Chapter 15 has described. I didn't look much into the special handling of sampler on specular materials (delta functions), but the result looks fine..

A few useful references I've found:

* the PBR book (Chapter 8, 14, 15)
* [Progressive Path Tracing with Explicit Light Sampling](https://computergraphics.stackexchange.com/questions/5152/progressive-path-tracing-with-explicit-light-sampling/5153#5153)
* [Cornell CS 6630 - Path Tracing](https://www.cs.cornell.edu/courses/cs6630/2015fa/index.shtml)

Also Kajiya-1986, Veach-[1995, 1997] may also be worth a read, but I've had enough fun with ray tracers for now...

---

I don't want to waste much of your time reviewing the details here, and absolutely no rush on this :)

Related issue = N/A

### DIFF
--- a/examples/cornell_box.py
+++ b/examples/cornell_box.py
@@ -1,5 +1,6 @@
 import taichi as ti
 import time
+import sys
 import math
 import numpy as np
 from renderer_utils import ray_aabb_intersection, intersect_sphere, ray_plane_intersect, reflect, refract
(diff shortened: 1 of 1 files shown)


k-ye:
Hmm, were you running on CUDA? If so, this *might* expose a problem in the random number generator...

FYI, here's what I got on `x64` and `metal` with `max_depth=1`

* x64
x64_depth=1

* metal
metal_depth=1

At last, to demonstrate that the change is somewhat substantial, I'm attaching a comparison of sampling 15K iterations (left: reference, it actually ran for 16K, but there was hardly any improvement;, right: with MIS)

IMG_2805

k-ye:
BTW, do you still see that ring artifacts even with a larger `max_ray_depth` and sampling iterations?

yuanming-hu:
cuda, max_ray_depth=2
Screenshot from 2020-04-28 19-18-46

On CPU the pattern is less pronounced.

Usually, this kind of pattern is due to numeric precision.

Also it seems that the image with MIS has a brighter floor. Writing a correct MIS path tracer is hard :-)

Anyway, please feel free to merge.

k-ye:
> Usually, this kind of pattern is due to numeric precision.

I see. Last time I saw this effect (not as obvious), it was because I forgot to add the epsilon when shooting a new ray from the hit point. But we are doing that here, so not sure what's happening...

> Also it seems that the image with MIS has a brighter floor. Writing a correct MIS path tracer is hard :-)

Can't agree more.. I can probably read on Veach-1995 and go through this again 🙁

PR #898, opened 2020-04-29
[opt] Set `has_global_side_effect` to `false` for some statements

xumingkuan:
Related issue = #656 (maybe not directly related)

This PR sets `has_global_side_effect` to `false` for `LoopIndexStmt, GlobalTemporaryStmt, StackAllocaStmt, StackLoadTopStmt, StackLoadTopAdjStmt`.

Benchmark:
**Before**:
benchmark20200428

**After** (Notice that this also affect the numbers when `advanced_optimization = false`!):
benchmark20200428_3

### DIFF
--- a/taichi/ir/statements.h
+++ b/taichi/ir/statements.h
@@ -377,4 +377,3 @@ class StackAccAdjointStmt : public Stmt {
 };
 
 TLANG_NAMESPACE_END
-        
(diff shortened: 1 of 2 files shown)


xumingkuan:
> Cool! Actually, I think all `Stack` statements have no global side effects (including those that push/pop). They only work on a local stack. Maybe we mark those as well.

As they are not *used as operands*, `DIE` will eliminate them if we mark as no global side effects...

xumingkuan:
BTW why GitHub doesn't give a reasonable diff?

PR #904, opened 2020-04-30
[test] Add an argument -r to rerun failed test

archibate:
Related issue = N/A

So that we don't get confused with random failures.
See https://github.com/pytest-dev/pytest-rerunfailures.

### DIFF
--- a/python/taichi/main.py
+++ b/python/taichi/main.py
@@ -34,7 +34,7 @@ def test_python(args):
     if args.verbose:
         pytest_args += ['-s', '-v']
     if args.rerun:
-        pytest_args += ['--reruns', '2']
+        pytest_args += ['--reruns', args.rerun]
     if int(pytest.main([os.path.join(root_dir, 'misc/empty_pytest.py'),
                         '-n1', '-q'])) == 0:  # test if pytest has xdist or not
         try:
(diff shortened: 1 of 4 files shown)


python/taichi/main.py
```diff
@@ -32,10 +31,12 @@ def test_python(args):
     else:
         # run all the tests
         pytest_args = [test_dir]
-    if verbose:
+    if args.verbose:
         pytest_args += ['-s', '-v']
+    if args.rerun:
+        pytest_args += ['--reruns', '2']
```
k-ye, line 37: I'm fine with this approach, just wondering why not simply let users decide how many times they want to rerun?

python/taichi/main.py
```diff
@@ -72,6 +73,10 @@ def make_argument_parser():
                         '--verbose',
                         action='store_true',
                         help='Run with verbose outputs')
+    parser.add_argument('-R',
```
k-ye, line 76: Please use `-r` to be consistent, unless there is a conflict?

python/taichi/main.py
archibate: I thought a user will never want to rerun more than 3 times since already proof the test to be wrong and keep rerun only wasting time to get the result.

archibate: Anyway, it doesn't matter at all, let's just use `-r2`.

Issue #902, opened 2020-04-29
Sparse computation with kernels


mathusalem2020:
Dear Community, 

If I were to simulate diffusion over a 2D grid with a square hole in the middle, how could I make use of sparse structures to avoid populating my kernel with a bunch of if statements ? 

```
import taichi as ti
import numpy as np
import matplotlib.pyplot as plt
%matplotlib inline
#running this in jupyter 
import cv2
ti.reset() #running this inside a jupyter notebook helps relaunch. 
#Otherwise compiler (?) doesn't let jupyter rerun.
ti.cfg.arch = ti.cuda 

real = ti.f32
scalar = lambda: ti.var(dt=real)
vec = lambda: ti.Vector(2, dt=real)

steps=15000
n_grid=256
p=scalar()
ti.root.dense(ti.l, steps).dense(ti.ij, n_grid).place(p) #

thick=10
init_cond=np.ones((n_grid,n_grid))
init_cond[thick:-thick,thick:-thick]=0
plt.imshow(init_cond)
#the initial condition has a thick border of concentration 1. The rest is 0.

   
@ti.func 
def laplacian(t, i, j):
    return  (-4 * p[t, i, j] + p[t, i, j - 1] + p[t, i, j + 1] + p[t, i + 1, j] + p[t, i - 1, j])


###Request####
#Here below, I avoid computing the diffusion on a square 25x25 area.
#How can I engineer the field p() to be sparse, so the kernel does not care
#about those values
@ti.kernel
def diffuse(t: ti.i32):
    for i in range(n_grid):
        for j in range(n_grid):
            if not ((i>100)&(i<125)&(j>100)&(j<125)):
                lp = laplacian(t-1,i,j)
                p[t,i,j] = p[t-1,i,j]+0.24*lp
####Request#####
# I would not need to replenish the initial conditions if the borders
# of my field were also inactive in the sparse structure of p()
@ti.kernel
def replenish(t: ti.i32):
    for i in range(n_grid):
        for j in range(n_grid):
            if ((i<thick) or (i>n_grid-thick)) or  ((j<thick) or (j>n_grid-thick)):
                p[t,i,j]=1




@ti.kernel                
def init_condition(arr: ti.ext_arr()):
    for i in range(n_grid):
        for j in range(n_grid):
            p[0,i,j]=arr[i,j]
            

@ti.kernel
def put_into_img(t: ti.i32, arr: ti.ext_arr()):
    for i in range(n_grid):
[...]

PR #900, opened 2020-04-29
[Doc] Add documentions for layout

archibate:
Related issue = #736 #896 #902

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

PR #905, opened 2020-04-30
[Example] Add example/bitmasked.py

archibate:
Related issue = #902


k-ye:
Actually we have `taichi_sparse.py`, which should be easy to be use `bitmasked`:

https://github.com/taichi-dev/taichi/blob/eef5139edd18c0d8741637f028cef9df31c0595c/examples/taichi_sparse.py#L10-L13

Could you consider having a flag in that example, e.g. `use_bitmasked`, then instantiate `block1, 2, 3` as either `bitmasked` or `pointer`, depending on the flag?

Also, as a general advice, can you include a screeenshot when adding a new example?

archibate:
Will consider tmr.
tbh taichi_sparse is a little complicated make people can't catch the point...
before having a good doc on sparse, let's add a very simple example for them to study then.

examples/bitmasked.py
```diff
@@ -0,0 +1,46 @@
...
+    for i, j in ti.ndrange(n, n):
+        x[i, j] = ti.sin(t)
+
+
+activate()
+
+gui = ti.GUI('bitmasked', n)
+for frame in range(10000):
+    paint(frame * 0.05)
+    #paint2(frame * 0.05)  # try this!!
```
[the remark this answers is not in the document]
archibate: Nice name! But I think `use_sparsity` make the example more hard to understand, so I didn't add yet.

archibate:
Thank for your valuable input! With this and #900, I believe we can make user understand sparse computation more clear now.

examples/bitmasked.py
```diff
@@ -0,0 +1,42 @@
+import taichi as ti
+ti.init()
```
k-ye, line 2: `arch=ti.gpu`?

examples/bitmasked.py
```diff
@@ -0,0 +1,42 @@
...
+    # Let's activate elements in the rectangle now!
+    for i, j in ti.ndrange((100, 125), (100, 125)):
+        x[i, j] = 0  # assign any value to activate the element at (i, j)
+
+
+@ti.kernel
+def paint_active_pixels(t: ti.f32):
+    # struct-for syntax: loop over active pixels, inactive pixels are excluded
+    for i, j in x:
+        x[i, j] = ti.sin(t)
```
k-ye, line 25: Because this can go negative, half of the time the graph is pure black. WDYT if we map the range from [-1, 1] to [0, 1]?

examples/bitmasked.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@ti.kernel
+def paint_all_pixels(t: ti.f32):
+    # range-for syntax: loop over all pixels, no matter active or not
+    for i, j in ti.ndrange(n, n):
+        x[i, j] = ti.sin(t)
```
k-ye, line 32: ditto

archibate:
Thank for your valuable input! With this and #900, I believe we can make user understand sparse computation more clear now.

> LGTM thanks! a few more nits, also did you get the same aliases on your machine?

Yes, and it's random.

archibate:
I found why: **we can't use bitmasked tesnor as a argument for `gui.set_image`** (or `to_numpy`). I'm not sure this is a bug or an expected feature. @yuanming-hu

archibate:
I see you use `np.empty` in `to_numpy`, and struct-for in `tensor_to_ext_arr`, so the inactive elements are not initialized at all!!!! Solution:
1. use `np.zeros`, so that inactive elements are initialized zero.
2. use range-for in `tensor_to_ext_arr`, so that all elements are copied, including inactive ones.

Concern for 1: Values in inactived elements will not be copied, is this ok?
Concern for 2: what if `pointer`? Will accessing an inactive elements cause a segfault? (null pointer)

Which one would you think is better? @yuanming-hu @k-ye

archibate:
Great! I'll merge once CI pass :) Btw, will we add a test for sparse tensor numpy I/O later (in a sep PR)?

k-ye:
> Btw, will we add a test for sparse tensor numpy I/O later (in a sep PR)?

Yep, that will be great, thank you!

PR #919, opened 2020-05-04
[ir][refactor] Move all Expression subclasses to `frontend_ir.h`

k-ye:
This PR completes all the frontend IR migration.

---

As @archibate and @xumingkuan have pointed out, there are some legacy constructs like `For`, `If` that are left inside `ir.h/cpp`. Unfortunately, some of them are still in use, so they are not immediately removable, see https://github.com/taichi-dev/taichi/blob/0c3358eae2b2305be991b529ba98b80651860f16/taichi/math/svd.h#L160-L162

Maybe we should either move them to `frontend.h/cpp` (**not** `frontend_ir.h`), or place them into a new file (e.g. `legacy_frontend.h/cpp`).

WDYT?

---

Related issue = #689 


k-ye:
The idea looks very appealing to me!

PR #936, opened 2020-05-09
[ir][refactor] Deprecate `ArgStoreStmt` (both frontend and CHI)

k-ye:
Also removed some helpers that are unused now:

* `mark_arg_return_value()`
* `Kernel::Arg::is_return_value`

Related issue = #909 

### DIFF
--- a/taichi/program/kernel.cpp
+++ b/taichi/program/kernel.cpp
@@ -213,7 +213,7 @@ void Kernel::set_arch(Arch arch) {
 }
 
 int Kernel::insert_arg(DataType dt, bool is_nparray) {
-  args.push_back(Arg{dt, is_nparray, 0, false});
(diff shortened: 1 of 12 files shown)


taichi/program/kernel.cpp
```diff
@@ -196,10 +196,6 @@ int64 Kernel::get_ret_int(int i) {
   }
 }
 
-void Kernel::mark_arg_return_value(int i, bool is_return) {
-  args[i].is_return_value = is_return;
```
archibate, line 200: Thanks! Could you also remove is_return_value here?

k-ye: Good catch! Done

k-ye:
> Can we also remove ArgStoreStmt too? I believe it's no longer used anywhere now.

Yep, done!

PR #951, opened 2020-05-11
[opengl] Use `compile_to_offloads` for IR lowering

archibate:
Related PR id = #700

### DIFF
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -718,6 +718,7 @@ void OpenglCodeGen::lower() {
                               /*ad_use_stack=*/false, config.print_ir,
                               /*lower_global_access*/true);
   global_tmps_buffer_size_ = res.total_size;
+  TI_TRACE("[glsl] Global temporary buffer size {} B", global_tmp_buffer_size_);
 #ifdef _GLSL_DEBUG
   irpass::print(ir);
 #endif
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -1,4 +1,4 @@
-//#define _GLSL_DEBUG 1
+#define _GLSL_DEBUG 1
 #include "codegen_opengl.h"
 
 #include <string>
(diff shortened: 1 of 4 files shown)


taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -1,4 +1,4 @@
-//#define _GLSL_DEBUG 1
+#define _GLSL_DEBUG 1
```
xumingkuan, line 1: Is this to be merged into master?

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -828,6 +709,21 @@ FunctionType OpenglCodeGen::gen(void) {
...
+  auto res = irpass::compile_to_offloads(ir, config,
+                              /*vectorize=*/false, kernel_->grad,
+                              /*ad_use_stack=*/false, config.print_ir,
+                              /*lower_global_access*/true);
+  global_tmps_buffer_size_ = res.total_size;
+  TI_TRACE("[glsl] Global temporary buffer size {} B", global_tmps_buffer_size_);
+#ifdef _GLSL_DEBUG
+  irpass::print(ir);
+#endif
+}
```
xumingkuan, line 725: What about the passes after `offload`?

taichi/backends/opengl/codegen_opengl.cpp
archibate: Nice catch!

taichi/backends/opengl/codegen_opengl.cpp
archibate: What do you mean by `offload`? Can't get your point.. Sorry..

xumingkuan: Oh, I understand now. The return value of `compile_to_offloads` is obtained by `irpass::offload`, this doesn't mean that `ir` is also returned at that time... (I was wondering where are the passes after `irpass::offload` (such as `full_simplify` and `demote_atomics`).)

xumingkuan:
I just thought about an optimization: among all fields of the return value of `compile_to_offloads`, if only `res.total_size` is used, we can reduce the number of global temps in `full_simplify` after `irpass::offload`, and then reduce `total_size` by relabeling all global temps. (e.g. if `global temp[offset=4]` is eliminated, we can let `global temp[offset=8]` take the place of `global temp[offset=4]`)

archibate:
> I just thought about an optimization: among all fields of the return value of `compile_to_offloads`, if only the variable `res.total_size` is used to allocate memory, we can reduce the number of global temps in `full_simplify` after `irpass::offload`, and then reduce `total_size` by relabeling all global temps. (e.g. if `global temp[offset=4]` is eliminated, we can let `global temp[offset=8]` take the place of `global temp[offset=4]`)

Sounds great! How do I do then? Basically re-run `IdentifyValuesUsedInOtherOffloads` at the end of `compile_to_offloads`? It's cool but we may want to add this in a separate PR?

k-ye:
FYI, the max global temporaries buffer size is actually fixed at `1MB`, see
https://github.com/taichi-dev/taichi/blob/1045bc24e9f7e98687302d8864bc1ad177fe6445/taichi/inc/constants.h#L9

So if we just don't care about waste a bit of memory, I think we should just remove `OffloadedResult`.

archibate:
> So if we don't care about wasting just a bit of memory, I think we should just remove OffloadedResult.

So, metal can simply consider gtmp as a part of root, just need to be zero-initialized per-launch? Or even zero-initialization is not needed?

xumingkuan:
> > So if we don't care about wasting just a bit of memory, I think we should just remove OffloadedResult.
> 
> So, metal can simply consider gtmp as a part of root, just need to be zero-initialized per-launch? Or even zero-initialization is not needed?

Zero-initialization is not needed. Now I think we don't care about wasting just a bit of memory...

k-ye:
> So, metal can simply consider gtmp as a part of root, just need to be zero-initialized per-launch? Or even zero-initialization is not needed?

For both Metal and LLVM, they are just assuming all the global tmps can fit in 1MB, so we don't have to care about precisely how much memory it will take...

> Zero-initialization is not needed.

Exactly..

archibate:
> > So, metal can simply consider gtmp as a part of root, just need to be zero-initialized per-launch? Or even zero-initialization is not needed?
> 
> For both Metal and LLVM, they are just assuming all the global tmps can fit in 1MB, so we don't have to care about precisely how much memory it will take...
> 
> > Zero-initialization is not needed.
> 
> Exactly..

Thank for the information, I'll perform a TPRT to see if this harms performance.

PR #985, opened 2020-05-14
[opengl][refactor] RangeSizeEvaluator -> KernelParallelAttrib to store more infomation

archibate:
…e more infomation

Related issue = #... (if any)


taichi/backends/opengl/opengl_api.h
```diff
@@ -47,8 +51,7 @@ struct CompiledProgram {
 
   void add(const std::string &kernel_name,
            const std::string &kernel_source_code,
-           int num_groups,
-           RangeSizeEvaluator rse,
+           KernelParallelAttrib kpa,
```
k-ye, line 54: nit: If you want to make sure the caller uses `std::move`, you can change the type to `KernelParallelAttrib&&`

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -81,9 +71,11 @@ class KernelGen : public IRVisitor {
   std::string root_snode_type_name_;
   std::string glsl_kernel_prefix_;
   int glsl_kernel_count_{0};
-  int num_threads_{1};
-  int num_groups_{1};
-  RangeSizeEvaluator range_size_evaluator_;
+
+  bool is_top_level_{true};
+  LineAppender line_appender_, line_appender_header_;
```
k-ye, line 76:
```suggestion
  LineAppender line_appender_;
  LineAppender line_appender_header_;
```

archibate: I'm fine with this, just ask if this is a suggested C++ coding style to never use `,` when declaring variables?

taichi/backends/opengl/opengl_api.h
archibate: Cool C++11! Thank for telling me that, I finally realized the difference between `&` and `&&`: `&&` must be moved lhs, while `&` only need to be lhs.

taichi/backends/opengl/codegen_opengl.cpp
k-ye: It can be misused, e.g. when you declare an object and its pointer type in the same line. Also see https://wiki.sei.cmu.edu/confluence/display/c/DCL04-C.+Do+not+declare+more+than+one+variable+per+declaration . Personally I think this is controversial, so the simplest rule is just not to use it.

archibate:
Thank for the suggestion! To clarify, `KernelParallelAttrib kpa` is an extent of `int num_threads`, so it's only used for parallel kernels.

Issue #992, opened 2020-05-15
[Backend] DirectX (D3D11, HLSL) Backend


quadpixels:
**Concisely describe the proposed feature**
I would like to add a DirectX (I'm thinking of D3D11 for now) backend to the compiler so that I can have one more way to utilize the integrated GPU in my laptop.

**Describe the solution you'd like (if any)**
I feel inclined to follow the OpenGL backend's approach (emitting GLSL source) since HLSL (the shader language used by D3D11 and D3D12) is similar to GLSL in many ways.

**Additional comments**
I had been working on a little game with both D3D11 and OpenGL rendering paths, and it appears to me that in Windows, the overhead for creating buffers and issuing draw calls with D3D11 is much lower than with OpenGL (by profiling with GPUView and RenderDoc), so I'm curious if DX can bring a slightly better user experience for Taichi in Windows.
I will start by studying the existing backends (standing on the shoulders of the giants)  (:

yuanming-hu:
Welcome on board! Adding a DirectX backend would be great for Windows users.

I think following the OpenGL backend's implementation shoulds reasonable. I'm not familiar with DirectX, but HLSL sounds like the only mature way to create DX shaders. There are other projects like https://github.com/microsoft/DirectXShaderCompiler - not sure if they provide a more useable alternative approach for shader code generation (e.g. compiling LLVM IR to DX shaders, if there's any chance...)

> I will start by studying the existing backends (standing on the shoulders of the giants) (:

Experts on this are @archibate and @k-ye. If you speak Chinese, welcome to our 0-th TaichiCon event: https://zhuanlan.zhihu.com/p/139627994 We will share some development experience there and uploaded the slides after the event :-)

quadpixels:
Thanks! Yes I actually attended the 0th TaichiCon evnet (:
I remember a participant noted SPIR-V, and potential paths to support multiple backends such as from Vulkan to D3D12 & OpenGL & WebGL, and I feel that might be great for a long-term plan (b/c I'm not familiar with Vulkan) ..
I guess my priority now might be to have something woring first so I'll try to come up with something with D3D11 and HLSL. Archibate and K-ye's presentation gave me a lot of inspirations & I will read into the Taichi language and try to get started.    (:

k-ye:
Nice! FYI, all the backends are in this directory: https://github.com/taichi-dev/taichi/tree/master/taichi/backends (you probably already know)..

quadpixels:
[...]

PR #993, opened 2020-05-15
[opengl][refactor] refactor stuff into one "SnodeInfo" for better furture sparse support

archibate:
Related issue = #711 #992

### DIFF
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -249,7 +249,7 @@ class KernelGen : public IRVisitor {
 
     emit("int {} = {} + {} * {}; // {}", stmt->short_name(),
          parent->short_name(),
-         struct_compiled_->snode_map[parent_type].elem_stride,
+         struct_compiled_->snode_map.at(parent_type).elem_stride,
          stmt->input_index->short_name(), stmt->snode->node_type_name);
   }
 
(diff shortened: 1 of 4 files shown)


taichi/backends/opengl/struct_opengl.cpp
```diff
@@ -32,41 +31,41 @@ void OpenglStructCompiler::collect_snodes(SNode &snode) {
...
-    stride_map_[node_name] = data_type_size(snode.dt);
+    snode_map_[node_name].stride = data_type_size(snode.dt);
   } else if (snode.type == SNodeType::dense || snode.type == SNodeType::root) {
-    const int n = (snode.type == SNodeType::dense) ? snode.n : 1;
-    length_map_[node_name] = n;
-    stride_map_[node_name] =
-        stride_map_[node_name + "_ch"] * length_map_[node_name];
-    class_children_map_[node_name] = stride_map_.at(node_name + "_ch");
+    const int n = (snode.type == SNodeType::root) ? 1 : snode.n;
+    snode_map_[node_name].length = n;
```
k-ye, line 65: Similar to what you've done for L38, you can write `auto &snode_info = snode_map_[node_name];` once?

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -258,8 +258,8 @@ class KernelGen : public IRVisitor {
   void visit(GetChStmt *stmt) override {
     emit("int {} = {} + {}; // {}", stmt->short_name(),
          stmt->input_ptr->short_name(),
-         struct_compiled_
-             ->class_get_map[stmt->input_snode->node_type_name][stmt->chid],
+         struct_compiled_->snode_map[stmt->input_snode->node_type_name]
```
k-ye, line 261: FYI: If you do not want to accidentally insert a default value into the map when the key is missing, consider using [`find`](http://www.cplusplus.com/reference/unordered_map/unordered_map/find/) over `operator[]`.

archibate: I guess [`at`](www.cplusplus.com/reference/map/map/at) should also work?

k-ye: Yes (sorry i slipped through this PR..)

Issue #1004, opened 2020-05-17
[misc] Support access vector/matrix elements dynamically


k-ye:
**Concisely describe the proposed feature**

I found that the following code didn't work:

```py
import taichi as ti

ti.init(arch=ti.gpu)

@ti.func
def f():
    a = ti.Vector([0.0, 0.0, 0.0])
    # for i in ti.static(range(3)):  # OK
    for i in range(3):  # throw an exception!
        a[i] = i
    return a

@ti.kernel
def test():
    f()
 
test()
```

Running it will spit the following error:

```
Traceback (most recent call last):
  File "r2.py", line 17, in <module>
    test()
  File "/Users/yekuang/Github/taichi/python/taichi/lang/kernel.py", line 500, in wrapped
    return primal(*args, **kwargs)
  File "/Users/yekuang/Github/taichi/python/taichi/lang/kernel.py", line 430, in __call__
    self.materialize(key=key, args=args, arg_features=arg_features)
  File "/Users/yekuang/Github/taichi/python/taichi/lang/kernel.py", line 310, in materialize
    taichi_kernel = taichi_kernel.define(taichi_ast_generator)
  File "/Users/yekuang/Github/taichi/python/taichi/lang/kernel.py", line 307, in taichi_ast_generator
    compiled()
  File "r2.py", line 15, in test
    f()
  File "/Users/yekuang/Github/taichi/python/taichi/lang/kernel.py", line 52, in __call__
    ret = self.compiled(*args)
  File "r2.py", line 10, in f
    a[i] = i
  File "/Users/yekuang/Github/taichi/python/taichi/lang/impl.py", line 93, in subscript
    return value.subscript(*indices)
  File "/Users/yekuang/Github/taichi/python/taichi/lang/matrix.py", line 221, in subscript
    return self(i, j)
  File "/Users/yekuang/Github/taichi/python/taichi/lang/matrix.py", line 192, in __call__
    return self.entries[self.linearize_entry_id(*args)]
TypeError: list indices must be integers or slices, not Expr
```

I initially thought this to be a bug. But then after looking at `Matrix` implementation,

https://github.com/taichi-dev/taichi/blob/91b8164e4a881d7d588daf02ca5083e097fa0fec/python/taichi/lang/matrix.py#L190-L192

 I saw that `entries` is a python list, so the index must be known at compile time?

So I think it will be great if we can support accessor index on `ti.Vector` and `ti.Matrix` at runtime.

[...]

PR #1013, opened 2020-05-18
[Lang] Improve error message for Matrix/Vector accessors

k-ye:
Related issue = #1004 

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -185,11 +185,11 @@ class Matrix(TaichiOperations):
             args = args + (0, )
         assert 0 <= args[0] < self.n
         assert 0 <= args[1] < self.m
-        # TODO(#1004): See if it's possible to support this at runtime
+        # TODO(#1004): See if it's possible to support indexing at runtime
         for i, a in enumerate(args):
             assert isinstance(
                 a, int
-            ), f'The {i}-th index of a Matrix/Vector must be a compile-time constant, got {a}'
+            ), f'The {i}-th index of a Matrix/Vector must be a compile-time constant integer, got {a}'
         return args[0] * self.m + args[1]
 
     def __call__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


python/taichi/lang/matrix.py
```diff
@@ -185,6 +185,11 @@ def linearize_entry_id(self, *args):
             args = args + (0, )
         assert 0 <= args[0] < self.n
         assert 0 <= args[1] < self.m
+        # TODO(#1004): See if it's possible to support this at runtime
+        for i, a in enumerate(args):
+            assert isinstance(
+                a, int
+            ), f'The {i}-th index of a Matrix/Vector must be a compile-time constant, got {a}'
```
archibate, line 192:
```suggestion
        # TODO(#1004): See if it's possible to support indexing at runtime
        for i, a in enumerate(args):
            assert isinstance(
                a, int
            ), f'The {i}-th index of a Matrix/Vector must be a compile-time constant integer, got {a}'
```

python/taichi/lang/matrix.py
```diff
@@ -185,6 +185,11 @@ def linearize_entry_id(self, *args):
             args = args + (0, )
         assert 0 <= args[0] < self.n
         assert 0 <= args[1] < self.m
+        # TODO(#1004): See if it's possible to support indexing at runtime
+        for i, a in enumerate(args):
+            assert isinstance(
+                a, int
+            ), f'The {i}-th index of a Matrix/Vector must be a compile-time constant integer, got {a}'
```
archibate, line 192:
```suggestion
            ), f'The {i+1}-th index of a Matrix/Vector must be a compile-time constant integer, got {a}'
```
Btw, does 0-th make sense in English? Like ground floor in UK vs first floor in US?

xumingkuan: I prefer `The {i}-th index`. "Zeroth" makes sense, especially to programmers.

PR #1019, opened 2020-05-19
[refactor] Add an always_false class trait

k-ye:
As you can see, this can be helpful for implementing the fallback branch of an `if constexpr`

Related issue = N/A
WANT_LGTM=any

### DIFF
--- a/taichi/math/arithmetic.h
+++ b/taichi/math/arithmetic.h
@@ -1,16 +1,8 @@
 #pragma once
 
-#include <type_traits>
+#include "taichi/common/trait.h"
 
 namespace taichi {
-
-namespace detail {
-template <class T>
-struct always_false : std::false_type {};
-
-template <class T>
-inline constexpr bool always_false_v = always_false<T>::value;
-}  // namespace detail
 
 // Round up |a| to the closest multiple of |b|, works only for integers.
 template <typename T>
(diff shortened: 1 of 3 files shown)


taichi/math/arithmetic.h
```diff
@@ -1,26 +1,17 @@
...
-
 // Round up |a| to the closest multiple of |b|, works only for integers.
 template <typename T>
 T iroundup(T a, T b) {
   if constexpr (std::is_integral_v<T>) {
     return ((a + b - 1) / b) * b;
   } else {
-    // Have to be type dependent: https://stackoverflow.com/a/53945549/12003165
-    static_assert(detail::always_false_v<T>, "Must be integral type");
+    static_assert(always_false_v<T>, "Must be integral type");
```
archibate, line 13: Intresting, so I guess the difference:
`static_assert(0)` always fail, even `if constexpr`.
`static_assert(always_false_v<T>)` only fail when `if constexpr` branches to there.

archibate: But I guess:
```suggestion
    static_assert(std::false_type<T>::value, "Must be integral type");
```
is more direct? What make you decide to add `always_false_v`? Could you clarify in `// comments`?

k-ye: I believe this is the recommended style in C++17, because they've added a `_v` for probably all the metaprogramming helpers. For example: `std::is_same` vs `std::is_same_v`: https://en.cppreference.com/w/cpp/types/is_same

PR #1029, opened 2020-05-20
[Lang][ir] "print" now support multiple scalar arguments

archibate:
Related issue = #923

- [x] print for only one expr
- [x] print multiple expr
<del>- [ ] print string + expr (iapr)</del>
<del>- [ ] print vector&matrix (iapr)</del>

note: iapr = in another pull request, no more note for this on core devs.

[[Click to format]](http://kun.csail.mit.edu:31415/1029)

### DIFF
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -113,7 +113,7 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
     std::vector<llvm::Value *> values;
 
     std::string formats;
-    for (auto &&content : stmt->contents) {
+    for (auto content : stmt->contents) {
       if (formats.size() != 0)  // not the first expr?
         formats += " ";         // add seperator
       formats += data_type_format(content->ret_type.data_type);
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -109,41 +109,36 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
 
     TI_ASSERT(stmt->width() == 1);
 
-    auto content = stmt->contents[0];
+    std::vector<llvm::Type *> types;
+    std::vector<llvm::Value *> values;
 
-    auto value_type = tlctx->get_data_type(content->ret_type.data_type);
+    std::string formats;
+    for (auto &&content : stmt->contents) {
+      if (formats.size() != 0)  // not the first expr?
+        formats += " ";         // add seperator
+      formats += data_type_format(content->ret_type.data_type);
 
-    std::string format;
+      auto value_type = tlctx->get_data_type(content->ret_type.data_type);
+      auto value = llvm_val[content];
+      if (content->ret_type.data_type == DataType::f32)
+        value = builder->CreateFPExt(value, value_type);
 
-    auto value = llvm_val[content];
-
-    if (content->ret_type.data_type == DataType::i32) {
-      format = "%d";
-    } else if (content->ret_type.data_type == DataType::i64) {
-      format = "%lld";
-    } else if (content->ret_type.data_type == DataType::f32) {
-      value_type = llvm::Type::getDoubleTy(*llvm_context);
-      value = builder->CreateFPExt(value, value_type);
-      format = "%f";
-    } else if (content->ret_type.data_type == DataType::f64) {
-      format = "%.12f";
-    } else {
-      TI_NOT_IMPLEMENTED
+      types.push_back(value_type);
+      values.push_back(value);
     }
 
-    std::vector<llvm::Type *> types{value_type};
+    auto format_str = "[debug] " + formats + "\n";
     auto stype = llvm::StructType::get(*llvm_context, types, false);
-    auto values = builder->CreateAlloca(stype);
-    auto value_ptr = builder->CreateGEP(
-        values, {tlctx->get_constant(0), tlctx->get_constant(0)});
-    builder->CreateStore(value, value_ptr);
-
-    auto format_str = "[debug] " + format + "\n";
-
+    auto value_arr = builder->CreateAlloca(stype);
+    for (int i = 0; i < values.size(); i++) {
+      auto value_ptr = builder->CreateGEP(
+          value_arr, {tlctx->get_constant(0), tlctx->get_constant(i)});
+      builder->CreateStore(values[i], value_ptr);
+    }
     llvm_val[stmt] = LLVMModuleBuilder::call(
         builder.get(), "vprintf",
         builder->CreateGlobalStringPtr(format_str, "format_string"),
-        builder->CreateBitCast(values,
+        builder->CreateBitCast(value_arr,
                                llvm::Type::getInt8PtrTy(*llvm_context)));
   }
 
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -105,6 +105,8 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
   }
 
   void visit(PrintStmt *stmt) override {
+    // TODO(@archibate): ask @yuanming-hu about how to deal with CUDA print
+
     TI_ASSERT(stmt->width() == 1);
 
     auto content = stmt->contents[0];
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -105,8 +105,6 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
   }
 
   void visit(PrintStmt *stmt) override {
-    // TODO(@archibate): ask @yuanming-hu about how to deal with CUDA print
-
     TI_ASSERT(stmt->width() == 1);
 
     std::vector<llvm::Type *> types;
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -107,21 +107,23 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
   void visit(PrintStmt *stmt) override {
     TI_ASSERT(stmt->width() == 1);
 
-    auto value_type = tlctx->get_data_type(stmt->stmt->ret_type.data_type);
+    auto content = stmt->contents[0];
+
+    auto value_type = tlctx->get_data_type(content->ret_type.data_type);
 
     std::string format;
 
-    auto value = llvm_val[stmt->stmt];
+    auto value = llvm_val[content];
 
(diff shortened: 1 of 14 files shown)


taichi/transforms/vector_split.cpp
```diff
@@ -241,8 +242,11 @@ class BasicBlockVectorSplit : public IRVisitor {
 
   void visit(PrintStmt *stmt) override {
     for (int i = 0; i < current_split_factor; i++) {
-      current_split[i] =
-          Stmt::make<PrintStmt>(lookup(stmt->stmt, i), stmt->str);
+      std::vector<Stmt *> new_contents;
+      std::transform(stmt->contents.begin(), stmt->contents.end(),
+                     new_contents.begin(),
+                     [&](auto &&x) { return lookup(x, i); });
```
k-ye, line 248: FYI: Note that this is actually capturing `i` by **reference**, which could be dangerous sometimes. Fortunately, `std::transform` finishes synchronously. Otherwise, the `i` in `return lookup(x, i);` could be different from when the lambda was constructed. I don't think this is a problem here, but I would prefer `[=]` just to be safe.

k-ye, line 248: Again, i think it's fine to just pass by value, i.e. `(auto x)`

taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -105,41 +105,40 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
...
+
     TI_ASSERT(stmt->width() == 1);
 
-    auto value_type = tlctx->get_data_type(stmt->stmt->ret_type.data_type);
+    std::vector<llvm::Type *> types;
+    std::vector<llvm::Value *> values;
 
-    std::string format;
+    std::string formats;
+    for (auto &&content : stmt->contents) {
```
k-ye, line 116: Here and below: Why `&&`? Note that `&&` usually conveys special meanings in C++ (e.g. move), which isn't the case here. Can we use `auto`, given that `x` is just a raw pointer and is cheap to copy?

taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -105,41 +105,40 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
...
+      if (formats.size() != 0)  // not the first expr?
+        formats += " ";         // add seperator
```
k-ye, line 118: Just FYI, there's a `fmt::join()` API, see https://fmt.dev/latest/api.html?highlight=join#_CPPv2N3fmt4joinERK5Range11string_view. But I'm fine with this approach..

taichi/ir/ir.h
```diff
@@ -805,7 +806,7 @@ class Block : public IRNode {
  public:
   Block *parent;
   std::vector<std::unique_ptr<Stmt>> statements, trash_bin;
-  std::map<Identifier, Stmt *> local_var_alloca; // Only used in frontend
+  std::map<Identifier, Stmt *> local_var_alloca;  // Only used in frontend
```
k-ye, line 809: I guess this is changed due to the formatter, but since I see it now, how about:

```suggestion
  std::unordered_map<Identifier, Stmt *> local_var_alloca;  // Only used in frontend
```

taichi/transforms/ir_printer.cpp
```diff
@@ -182,16 +182,15 @@ class IRPrinter : public IRVisitor {
   }
 
   void visit(FrontendPrintStmt *print_stmt) override {
-    print("print \"{}\", {}", print_stmt->str, print_stmt->expr.serialize());
+    print("print {}", print_stmt->contents[0].serialize());
```
k-ye, line 185: Since this is powerful enough to support multiple args, WDYT if we print the number of args as well?

taichi/ir/ir.h
archibate: Good idea! But we want to focus in this PR and not getting unwanted bugs into here.

taichi/backends/cuda/codegen_cuda.cpp
archibate: Cool! Btw, does C++ have something like the Python `map()`?

k-ye: Yes, and i thought you just used... `std::transform`

taichi/transforms/vector_split.cpp
k-ye: Sorry let me just add a few more comments on this: I'm suggesting to use `[=]` because this is inside a loop. If you used Go before, you probably know that there's a nasty bug, where if you create an anonymous function that captures the index `i` of the loop, it is captured by reference. Since lots of the times these anonymous functions are goroutines, they run asynchronously, and the index value would be changed inside the function. See https://yourbasic.org/golang/gotcha-data-race-closure/

taichi/backends/cuda/codegen_cuda.cpp
archibate: I see, but can I use it as a return value? e.g. `std::vector<T> transform(std::vector<T>)`?

taichi/transforms/ir_printer.cpp
archibate: Implemented using `fmt::join`.

taichi/backends/cuda/codegen_cuda.cpp
k-ye:
```cpp
std::vector<Foo> in;
std::vector<Bar> out;
// Make sure out.size() >= in.size();
std::transform(in.begin(), in.end(), out.begin(), [](const Foo&)->Bar {...});
```

taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -105,41 +105,40 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
...
     TI_ASSERT(stmt->width() == 1);
 
-    auto value_type = tlctx->get_data_type(stmt->stmt->ret_type.data_type);
+    std::vector<llvm::Type *> types;
+    std::vector<llvm::Value *> values;
 
-    std::string format;
+    std::string formats;
+    for (auto content : stmt->contents) {
+      if (formats.size() != 0)  // not the first expr?
```
k-ye, line 117: OK, since you did use `fmt::join` (thx!), could you also do the same here?

taichi/codegen/codegen_llvm.cpp
```diff
@@ -619,58 +619,35 @@ llvm::Value *CodeGenLLVM::create_print(std::string tag,
...
-  } else {
-    TI_NOT_IMPLEMENTED
+  std::string formats;
+  for (auto content : stmt->contents) {
+    auto value = llvm_val[content];
+    if (content->ret_type.data_type == DataType::f32)
+      value = builder->CreateFPExt(value, tlctx->get_data_type(DataType::f64));
+    args.push_back(value);
+
+    if (formats.size() != 0)  // not the first expr?
```
k-ye, line 643: and here..?

taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -105,41 +105,40 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
   }
 
   void visit(PrintStmt *stmt) override {
+    // TODO(@archibate): ask @yuanming-hu about how to deal with CUDA print
```
k-ye, line 108: Is this still necessary?

taichi/transforms/lower_ast.cpp
```diff
@@ -115,10 +115,13 @@ class LowerAST : public IRVisitor {
 
   void visit(FrontendPrintStmt *stmt) override {
     // expand rhs
-    auto expr = load_if_ptr(stmt->expr);
+    std::vector<Stmt *> stmts;
     auto fctx = make_flatten_ctx();
-    expr->flatten(&fctx);
-    fctx.push_back<PrintStmt>(expr->stmt, stmt->str);
+    for (auto &&x : stmt->contents) {
```
k-ye, line 120:
```suggestion
    for (auto x : stmt->contents) {
```

taichi/transforms/vector_split.cpp
```diff
@@ -241,8 +242,11 @@ class BasicBlockVectorSplit : public IRVisitor {
 
   void visit(PrintStmt *stmt) override {
     for (int i = 0; i < current_split_factor; i++) {
-      current_split[i] =
-          Stmt::make<PrintStmt>(lookup(stmt->stmt, i), stmt->str);
+      std::vector<Stmt *> new_contents;
+      std::transform(stmt->contents.begin(), stmt->contents.end(),
```
k-ye, line 246: I just realized that this is wrong: `std::transform` will not automatically call `push_back` for `new_contents`. You need to either make sure `new_contents`'s size is the same as `stmt->contents`, or you can try [`std::back_inserter`](https://en.cppreference.com/w/cpp/iterator/back_inserter). Example in https://en.cppreference.com/w/cpp/algorithm/transform

taichi/backends/cuda/codegen_cuda.cpp
archibate: No, this is not the case. We will have to use `fmt::format("[debug] {}\n", fmt::join(formats))` for this, while we already do the format works in `lang_util.h:data_type_format()` and `vprintf` at runtime.

k-ye: Sorry I don't get your point. What's the problem with using `fmt::format("[debug] {}\n", fmt::join(formats, " "))`?

archibate: Overkill.

xumingkuan: > Sorry I don't get your point. What's the problem with using `fmt::format("[debug] {}\n", fmt::join(formats, " "))`?

`data_type_format(content->ret_type.data_type) for content in stmt->contents` doesn't form an `std::vector`.

k-ye: We could just change `formats` to `std::vector<std::string>`..

taichi/ir/frontend_ir.h
```diff
@@ -75,11 +75,12 @@ class FrontendIfStmt : public Stmt {
 
 class FrontendPrintStmt : public Stmt {
  public:
-  Expr expr;
-  std::string str;
```
k-ye, line 79: BTW, do you know what `str` was for? Was it used as a formatting string in the print statement?

archibate:
> Very nice! Do you plan to support `string` in the near future? That shouldn't be too hard given what you already have now. We should also support printing extra data types such as `u64` in a different PR.

Yes, we will, iapr. Let's merge this for now.

taichi/ir/frontend_ir.h
archibate: It's the variable name, e.g.:
`print(x)`, then `str = "x"`.

PR #1034, opened 2020-05-21
[refactor] Remove the global begin_stmt and end_stmt

k-ye:
"One less global variable is a bit less evil in the world." -- *Nobody*  233 B.C.

As part of the refactor, I also:

* Reordered a few static function args so that they match the corresponding Constructor's order
* Added `const &` to a few `stmt_to_offloaded` maps.

Related issue = N/A
WANT_LGTM = any

### DIFF
--- a/taichi/transforms/offload.cpp
+++ b/taichi/transforms/offload.cpp
@@ -14,7 +14,7 @@ namespace {
 // Offloaded local variables to its offset in the global tmps memory.
 using StmtToOffsetMap = std::unordered_map<const Stmt *, std::size_t>;
 
-struct OffloadedToBeginEndStmts {
+struct OffloadedRanges {
   using Map = std::unordered_map<const OffloadedStmt *, Stmt *>;
   Map begin_stmts;
   Map end_stmts;
@@ -23,8 +23,8 @@ struct OffloadedToBeginEndStmts {
 // Break kernel into multiple parts and emit struct for listgens
 class Offloader {
  public:
-  static OffloadedToBeginEndStmts run(IRNode *root) {
-    OffloadedToBeginEndStmts offloaded_to_begin_end;
+  static OffloadedRanges run(IRNode *root) {
+    OffloadedRanges offloaded_ranges;
 
     auto root_block = dynamic_cast<Block *>(root);
     auto root_statements = std::move(root_block->statements);
@@ -52,14 +52,14 @@ class Offloader {
           offloaded->const_begin = true;
           offloaded->begin_value = val->val[0].val_int32();
         } else {
-          offloaded_to_begin_end.begin_stmts.insert(
+          offloaded_ranges.begin_stmts.insert(
               std::make_pair(offloaded.get(), s->begin));
         }
         if (auto val = s->end->cast<ConstStmt>()) {
           offloaded->const_end = true;
           offloaded->end_value = val->val[0].val_int32();
         } else {
-          offloaded_to_begin_end.end_stmts.insert(
+          offloaded_ranges.end_stmts.insert(
               std::make_pair(offloaded.get(), s->end));
         }
         offloaded->block_dim = s->block_dim;
@@ -77,7 +77,7 @@ class Offloader {
       }
     }
     assemble_serial_statements();
-    return offloaded_to_begin_end;
+    return offloaded_ranges;
   }
 
  private:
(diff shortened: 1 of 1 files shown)


taichi/transforms/offload.cpp
```diff
@@ -331,9 +340,11 @@ class FixCrossOffloadReferences : public BasicStmtVisitor {
  private:
   FixCrossOffloadReferences(
       const StmtToOffsetMap &local_to_global_offset,
-      std::unordered_map<Stmt *, Stmt *> stmt_to_offloaded)
+      const std::unordered_map<Stmt *, Stmt *> &stmt_to_offloaded,
```
archibate, line 343: Any reason why `const &` here?

taichi/transforms/offload.cpp
```diff
@@ -14,20 +14,18 @@ namespace {
 // Offloaded local variables to its offset in the global tmps memory.
 using StmtToOffsetMap = std::unordered_map<const Stmt *, std::size_t>;
 
-std::unique_ptr<std::unordered_map<OffloadedStmt *, Stmt *>> begin_stmt,
-    end_stmt;
+struct OffloadedToBeginEndStmts {
```
archibate, line 17: Name nit: why `to`?
```suggestion
struct OffloadedRanges {
```

taichi/transforms/offload.cpp
```diff
@@ -77,9 +77,11 @@ class Offloader {
       }
     }
     assemble_serial_statements();
+    return offloaded_to_begin_end;
```
archibate, line 80:
```suggestion
    return offloaded_ranges;
```

taichi/transforms/offload.cpp
k-ye: Good point. If you look at the entire call stack, it was like:

```
offload()
\- FixCrossOffloadReferences::run()
    \- FixCrossOffloadReferences::FixCrossOffloadReferences()
```

Before this PR, only the last call used `move`, while the first two didn't (likely because we forgot to), so we still made two unnecessary copies. While move is a modern feature, IMO the easiest way to avoid copy remains using `const &`. (`move` also means transferring ownership, which is really useful when data cannot be copy constructed, e.g. `std::unique_ptr`). Does this make sense?

taichi/transforms/offload.cpp
k-ye: Ditto

taichi/transforms/offload.cpp
k-ye: Nice! Since many places need to be changed, I will make the change locally.

k-ye:
@archibate If you do approve, could you hit the Approve button?

k-ye:
> why OffloadedRanges *const offloaded_ranges_; is used rather than const OffloadedRanges &offloaded_ranges_; -- I think it's better to use a const reference if we don't modify it.

It's because we do modify the value of the hash maps in `const OffloadedRanges&`, therefore we cannot mark it const (unfortunately)

xumingkuan:
> > why OffloadedRanges *const offloaded_ranges_; is used rather than const OffloadedRanges &offloaded_ranges_; -- I think it's better to use a const reference if we don't modify it.
> 
> It's because we do need the non-const value of the hash maps in `const OffloadedRanges&`, therefore we cannot mark it const (unfortunately). See https://github.com/taichi-dev/taichi/pull/1034/files#diff-3960fe96f8a142ecb79201fd25fa3335R212-R213 and
> 
> https://github.com/taichi-dev/taichi/blob/28ae40c8b652edd61311a27646a9312bb7068105/taichi/transforms/offload.cpp#L223
> 
> .
> BTW it seems that `test_and_allocate()`'s arg can be made const, but maybe we should clean that up later.

I see. So if `const OffloadedRanges&` is used, `begin->second` will have type `const Stmt *`?

BTW is there any advantage of `OffloadedRanges *const` over `OffloadedRanges *`? Does it only imply that the pointer itself can't be modified?

k-ye:
> I see. So if const OffloadedRanges& is used, begin->second will have type const Stmt *?

Yes, see [find](https://en.cppreference.com/w/cpp/container/unordered_map/find). There are two overloads on it, based on the constness of `unordered_map` itself. If the map is const, then only (2) is available, which implies constness of its kv as well.

>  Does it only imply that the pointer itself can't be modified?

Yes, since `T* const` is basically a `T&`, this is also used to convey the fact that the object doesn't own the pointer.

PR #1036, opened 2020-05-22
[refactor] Use unordered_map for Block::local_var_alloca

archibate:
Related issue = https://github.com/taichi-dev/taichi/pull/1029#discussion_r428620940


archibate:
Maybe there's some reason why we used ordered_`map`...

PR #1061, opened 2020-05-26
[doc] Emphasize some ideas on how to communicate effectively

k-ye:
I think it's the shared responsibility on all of us to make the community a friendly place to discuss and work. As a start, I'm adding a few examples in the contributor guidelines. Please take a look, and feel free to add more advices if you feel necessary. Thanks!

Screen Shot 2020-05-26 at 20 35 14

Related issue = N/A


docs/contributor_guide.rst
```diff
@@ -49,6 +51,14 @@ Effective communication
...
 - If you are not a native speaker, consider using a spell checker such as `Grammarly <https://app.grammarly.com/>`_.
 
+Please base your discussion and feedback on facts, and not personal feelings. It is very important for all of us to maintain a friendly and blame-free community. Some examples:
+
+.. tip::
+  (Positive) This design could be confusing to new Taichi users.
+
+.. warning::
+  (Negative) This design is terrible.
+
```
KLozes, line 62: Maybe the ```tip``` and ```warning``` should be changed to ```recommended``` and ```not recommended``` , or ```acceptable``` and ```not acceptable```. I'm not sure the word ```warning``` here is conveying that we should avoid such unfriendly comments. Otherwise, LGTM!

k-ye: Totally agreed. The unfortunate thing is that, rst predefined these admonitions, and `tip` and `warning` offers the color I wanted... I also see that you can define custom admonitions, but then I'm not sure how to customize the color.

KLozes: I too am not a CSS expert. I think changing the ```(Positive)``` and ```(Negative)``` to ```(Acceptable)``` and ```(Not Acceptable)``` would work too.

k-ye: Cool, done

PR #1064, opened 2020-05-26
[Lang] Add any()/all() functions for Matrix, move casting ops to common_ops

KLozes:
I added ```any()/all()``` functions to Matrix to check if any/all matrix entries are non-zero. It works on float and int types. They can be called in using builtin syntax or as a method of a matrix. 
```python
any(mat) or mat.any() 
all(mat) or mat.all()
```

I also moved __ti_int__ and __ti_float__ to common ops, which allows casting of matrices with 
``` python
int(mat)
float(mat)
```

Related issue = #1050 


KLozes:
Thanks! I agree the the negative integers were confusing and unexpected.

PR #1078, opened 2020-05-29
[metal] Make the comparison result -1 for true

k-ye:
Related issue = #577


k-ye:
> Btw, I think -1 for true is really smart because: bit_not(-1) = 0.

Yeah, i think that's what @yuanming-hu explained for vectorized computation (https://github.com/taichi-dev/taichi/commit/222bc726e3dfb43f6cc94f276cadddfde004e1be#r39531260)

Issue #1076, opened 2020-05-29
[LLVM] GC system doesn't maintain free list correctly


k-ye:
**Describe the bug**

I think there's a memory leak in the GC system. It seems that `NodeManager::free_list` is not maintained correctly?

https://github.com/taichi-dev/taichi/blob/022ee53da2ef3786e5b40971f1e075b217a446b4/taichi/runtime/llvm/runtime.cpp#L616

Instead of `clear()`, we probably should resize `free_list` to be `max(free_list->size() - free_list_used, 0)`.

**Log/Screenshots**

N/A

**To Reproduce**

I will make a PR to add a test case for this.

PR #1079, opened 2020-05-29
[llvm] [sparse] Fix a memory leak bug in GC's free list on CPU

k-ye:
Related issue = #1076 

### DIFF
--- a/taichi/runtime/llvm/internal_function.h
+++ b/taichi/runtime/llvm/internal_function.h
@@ -1,3 +1,12 @@
+#define _CHECK(cond, r)                        \
+  do {                                         \
+    TI_ASSERT(cond);                           \
+    if ((r)->error_code) {                     \
+      printf("%s", (r)->error_message_buffer); \
+      abort();                                 \
+    }                                          \
+  } while (0)
+
 i32 do_nothing(Context *context) {
   return 0;
 }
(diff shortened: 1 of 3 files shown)


k-ye:
This isn't completely ready yet, because I want to merge #1078  first to eliminate all the test failures. But you can probably get a sense of where the problem is..

taichi/runtime/llvm/internal_function.h
```diff
@@ -1,3 +1,12 @@
+#define _CHECK(cond, r)                        \
```
xumingkuan, line 1: Let's use `TI` as the prefix of all macros.

tests/python/test_internal_func.py
```diff
@@ -50,3 +50,11 @@ def test():
 
     test()
     test()
+
+
+def _test_node_manager_gc():
```
xumingkuan, line 55: Why does it begin with `_`?

k-ye: The bug is not fixed, so running this test will actually fail.

taichi/runtime/llvm/internal_function.h
k-ye: Done

tests/python/test_internal_func.py
k-ye: I added the fix now

k-ye:
Thanks for the quick review!

FYI, later I'd like to make the `InternalFuncStmt` support simple arguments, then we can pass in things like `ti.cpu`, `ti.gpu` labels to the test function..

taichi/runtime/llvm/internal_function.h
```diff
@@ -1,3 +1,14 @@
+// Note that TI_ASSERT provided by runtime doesn't check fail immediately. As
+// a hack, we just check the last error code whenever we call TI_ASSERT.
+#define TI_TEST_CHECK(cond, r)                 \
+  do {                                         \
+    TI_ASSERT(cond);                           \
+    if ((r)->error_code) {                     \
+      printf("%s", (r)->error_message_buffer); \
```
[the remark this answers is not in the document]
yuanming-hu: This may lead to JIT linking issues. In some cases `printf` here conflicts with the `printf` in the main thread. That's why we use `taichi_printf`.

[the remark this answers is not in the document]
k-ye: Ah thanks!

PR #1080, opened 2020-05-29
[opengl] Make the comparison result -1 for true

archibate:
Related issue = #577 #1078


archibate:
Yes! It worked.

PR #1098, opened 2020-06-01
[cli] [example] Support running examples with python file extension

TH3CHARLie:
Related issue = #... (if any)

### DIFF
--- a/python/taichi/main.py
+++ b/python/taichi/main.py
@@ -10,6 +10,7 @@ from colorama import Fore, Back, Style
 from taichi.tools.video import make_video, interpolate_frames, mp4_to_gif, scale_video, crop_video, accelerate_video
 from pathlib import Path
 import runpy
+from typing import Tuple, Set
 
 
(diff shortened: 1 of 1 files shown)


TH3CHARLie:
Running `ti example foo.py` would be a common thing to do for new users, so let's support it.

python/taichi/main.py
```diff
@@ -98,6 +98,9 @@ def get_available_examples() -> set:
 def run_example(name: str):
     """Run an example based on the example NAME."""
     all_example_names = get_available_examples()
+    # try to find and remove python file extension once
+    if name.endswith('.py'):
+        name = name.split('.')[0]
```
archibate, line 103:
```suggestion
        name = name[:-3]
```
In case of `example.name.py` -> `example`.

TH3CHARLie: Thanks for the comment! I think I will refactor this PR once #1085 is merged(as it requires to rebase)

Issue #1095, opened 2020-06-01
[Lang] make Matrix.__init__ less response by introducing multiple initializers


archibate:
**Concisely describe the proposed feature**
I strongly argue about the ultra-massive-centralized-multiplexer in `ti.Matrix.__init__`:
```py
def __init__(self, n=1, m=1, dt=None, shape=None,
             empty=False, layout=None, needs_grad=False,
             keep_raw=False, rows=None, cols=None):
   ... # 100+ lines of if-else-if's, within just **one function**
```

> Functions should be short and sweet, and do just one thing. They should fit on one or two screenfuls of text (the ISO/ANSI screen size is 80x24, as we all know), and do one thing and do that well.

... said the [Linux Kernel Coding Style](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#functions), and I believe the 80x24 rule also works in python too.

> Another measure of the function is the number of local variables. They shouldn’t exceed 5-10, or you’re doing something wrong. Re-think the function, and split it into smaller pieces. A human brain can generally easily keep track of about 7 different things, anything more and it gets confused. You know you’re brilliant, but maybe you’d like to understand what you did 2 weeks from now.

Not to say lobal variables, just the arguments alone, there are 11.

**Describe the solution you'd like (if any)**

Also note that the API initializing a `ti.Matrix` is highly centralized, GAMES201 students can easily get confused by error messages when they're not using the correct argument specification:
```py
ti.Matrix(n, m, dt, shape)  # matrix tensor
ti.Matrix(n, dt, shape)  # vector tensor
ti.Matrix([[a, b], [c, d]])  # taichi-scope matrix
ti.Matrix([a, b])  # taichi-scope vector
ti.Matrix(cols=[u, v])  # taichi-scope matrix, with col vectors
ti.Matrix(rows=[u, v])  # taichi-scope matrix, with row vectors
ti.Matrix(n, m, empty=True)  # empty matrix
```
Don't you think a better design could be:
```py
ti.Matrix.var(n, m, dt, shape)  # matrix tensor
ti.Vector.var(n, dt, shape)  # vector tensor
ti.Matrix([[a, b], [c, d]])  # taichi-scope matrix
ti.Vector([a, b])  # taichi-scope vector
ti.Matrix.cols([u, v])  # taichi-scope matrix, with col vectors
ti.Matrix.rows([u, v])  # taichi-scope matrix, with row vectors
ti.Matrix.empty(n, m)  # empty matrix
```
[...]

PR #1107, opened 2020-06-02
[Lang] [refactor] ti.Matrix initializers

liaopeiyuan:
Addressing #1095 .

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -585,9 +585,9 @@ class Matrix(TaichiOperations):
                           [ti.sin(alpha), ti.cos(alpha)]])
 
     @staticmethod
-    def var(n, m, dt, shape):
+    def var(n, m, dt, **kwargs):
         import taichi as ti
-        return ti.Matrix(n=n, m=m, dt=dt, shape=shape)
+        return ti.Matrix(n=n, m=m, dt=dt, **kwargs)
 
     @staticmethod
     def rows(rows):
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -596,7 +596,7 @@ class Matrix(TaichiOperations):
         mat.n = len(rows)
         if isinstance(rows[0], Matrix):
             for row in rows:
-                assert row.m == 1, "inputs must be vectors, ie. size n by 1"
+                assert row.m == 1, "Inputs must be vectors, i.e. m == 1"
                 assert row.n == rows[
                     0].n, "Input vectors must share the same shape"
             mat.m = rows[0].n
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -6,6 +6,7 @@ import numpy as np
 from .util import to_numpy_type, to_pytorch_type, deprecated
 from .common_ops import TaichiOperations
 from collections.abc import Iterable
+import warnings
 
 
 class Matrix(TaichiOperations):
@@ -28,12 +29,28 @@ class Matrix(TaichiOperations):
         self.grad = None
 
         if rows is not None or cols is not None:
-            raise Exception(
-                "ad-hoc usage of row- and column- vectors is deprecated." +
-                "This message will be removed in the next version.")
-        if empty == True:
-            raise Exception(
-                "instantiation of empty matrix via ti.Matrix is deprecated.")
+            warnings.warn(
+            f"ad-hoc usage of row- and column- vectors is deprecated.",
+            PendingDeprecationWarning,
+            stacklevel=3)
+            if rows is not None and cols is not None:
+                raise Exception("cannot specify both rows and columns")
+            self.dt = dt
+            import taichi as ti
+            mat = ti.Matrix.cols(cols) if cols is not None else ti.Matrix.rows(rows)
+            self.n = mat.n
+            self.m = mat.m
+            self.entries = mat.entries
+            return
+
+        elif empty == True:
+            warnings.warn(
+            f"instantiation of empty matrix via ti.Matrix is deprecated.",
+            PendingDeprecationWarning,
+            stacklevel=3)
+            self.dt = dt
+            self.entries = [[None] * m for _ in range(n)]
+            return
 
         elif isinstance(n, (list, tuple, np.ndarray)):
             if len(n) == 0:
@@ -594,7 +611,6 @@ class Matrix(TaichiOperations):
         else:
             raise Exception(
                 "cols/rows must be list of lists or lists of vectors")
-        print(mat.entries, mat.n, mat.m)
         return mat
 
     @staticmethod
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -46,8 +46,8 @@ class Matrix(TaichiOperations):
 
         elif empty == True:
             warnings.warn(
-                f"instantiation of empty matrix via ti.Matrix is deprecated.",
-                PendingDeprecationWarning,
+                f"ti.Matrix(n, m, empty=True) is deprecated, use ti.Matrix.empty(n, m) instead",
+                DeprecationWarning,
                 stacklevel=3)
             self.dt = dt
             self.entries = [[None] * m for _ in range(n)]
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -598,7 +598,7 @@ class Matrix(TaichiOperations):
             for row in rows:
                 assert row.m == 1, "inputs must be vectors, ie. size n by 1"
                 assert row.n == rows[
-                    0].n, "input vectors must be the same shape"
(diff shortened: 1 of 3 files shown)


liaopeiyuan:
Current thought: __init__ only defaults to ``ti.Matrix([[a, b], [c, d]])  # taichi-scope matrix``, and other usages are invoked by var, rols, cols, empty, etc, and passed into __init__ via flags.

python/taichi/lang/matrix.py
```diff
@@ -565,6 +565,26 @@ def rotation2d(alpha):
...
 
+    @staticmethod
+    def var(n, m, dt, shape):
+        import taichi as ti
+        return ti.Matrix(n=n, m=m, dt=dt, shape=shape)
+
+    @staticmethod
+    def cols(columns):
+        import taichi as ti
+        return ti.Matrix(cols=columns)
```
archibate, line 576: Thank for refactor, could you move the codes in if-else-if in `__init__` to here?

archibate:
Thank for refactor, but could you show a warning (but still working) instead of raising an exception? Users needs time to get accustomed with the changes, we don't want to break backward-compatibility until, say, v0.7.0.

The `@ti.layout` can be a good example:
https://github.com/taichi-dev/taichi/blob/73c405f9fe2c660903a6ba4e73cb03774efe5117/python/taichi/lang/impl.py#L282-L285

python/taichi/lang/matrix.py
```diff
@@ -627,14 +662,28 @@ def outer_product(self, b):
...
+                yield ', '
+            yield '['
+            for i in range(self.n):
+                if i:
+                    yield ', '
+                yield self(i, j)
+            yield ']'
+
+        if self.m != 1:
+            yield ']'
```
archibate, line 689: Remove?

tests/python/test_linalg.py
```diff
@@ -263,6 +263,37 @@ def test_init_matrix_from_vectors():
...
+        for i in range(3):
+            a = ti.Vector([1.0, 4.0, 7.0])
+            b = ti.Vector([2.0, 5.0, 8.0])
+            c = ti.Vector([3.0, 6.0, 9.0])
+            m1[i] = ti.Matrix.rows([a, b, c])
+            m2[i] = ti.Matrix.cols([a, b, c])
+            m3[i] = ti.Matrix.rows([[1.0, 4.0, 7.0], [2.0, 5.0, 8.0],
+                                    [3.0, 6.0, 9.0]])
+            m4[i] = ti.Matrix.cols([[1.0, 4.0, 7.0], [2.0, 5.0, 8.0],
+                                    [3.0, 6.0, 9.0]])
```
archibate, line 277: Nice work! Would you mind also update these usages in `docs/matrix.rst`?

python/taichi/lang/matrix.py
```diff
@@ -590,6 +584,47 @@ def rotation2d(alpha):
...
 
+    @staticmethod
+    def var(n, m, dt, shape):
+        import taichi as ti
+        return ti.Matrix(n=n, m=m, dt=dt, shape=shape)
+
+    @staticmethod
+    def rows(rows):
+        import taichi as ti
+        mat = ti.Matrix()
```
archibate, line 595:
```suggestion
        mat = ti.Matrix().empty()
```
since it's previous content won't be used, just a placeholder.

python/taichi/lang/matrix.py
```diff
@@ -590,6 +584,47 @@ def rotation2d(alpha):
...
+                "cols/rows must be list of lists or lists of vectors")
+        return mat
+
+    @staticmethod
+    def cols(cols):
+        import taichi as ti
+        return ti.Matrix.rows(cols).transpose()
+
+    @staticmethod
+    def empty(n, m, dt=None):
```
archibate, line 623: Mark: btw, I find that `dt` seems only work for tensor-of-matrices, not for matrices. `self.dt` might needs refactor.

archibate:
Sorry about the CI failure, now fixed in #1120.

liaopeiyuan:
Where is .empty() defined?

archibate:
Given that `empty` is hardly visible for end-users, it's not yet in the doc, sadly... but feel free to add you entry about it if you like to.

python/taichi/lang/matrix.py
```diff
@@ -25,38 +26,33 @@ def __init__(self,
...
                  rows=None,
                  cols=None):
-        # TODO: refactor: use multiple initializer like `ti.Matrix.cols([a, b, c])`
-        # and `ti.Matrix.empty(n, m)` instead of ad-hoc `ti.Matrix(cols=[a, b, c])`.
         self.grad = None
+
         if rows is not None or cols is not None:
+            warnings.warn(
+                f"ad-hoc usage of row- and column- vectors is deprecated.",
+                PendingDeprecationWarning,
```
archibate, line 34:
```suggestion
                f"ti.Matrix(rows=[...]) or ti.Matrix(cols=[...]) is deprecated, use ti.Matrix.rows([...]) or ti.Matrix.cols([...]) instead.",
                DeprecationWarning,
```

python/taichi/lang/matrix.py
```diff
@@ -25,38 +26,33 @@ def __init__(self,
...
+                rows)
+            self.n = mat.n
+            self.m = mat.m
+            self.entries = mat.entries
+            return
+
+        elif empty == True:
+            warnings.warn(
+                f"instantiation of empty matrix via ti.Matrix is deprecated.",
+                PendingDeprecationWarning,
```
archibate, line 50:
```suggestion
                f"ti.Matrix(n, m, empty=True) is deprecated, use ti.Matrix.empty(n, m) instead",
                DeprecationWarning,
```

python/taichi/lang/matrix.py
```diff
@@ -25,38 +26,33 @@ def __init__(self,
...
+            self.entries = mat.entries
+            return
+
+        elif empty == True:
+            warnings.warn(
+                f"ti.Matrix(n, m, empty=True) is deprecated, use ti.Matrix.empty(n, m) instead",
+                DeprecationWarning,
+                stacklevel=3)
+            self.dt = dt
+            self.entries = [[None] * m for _ in range(n)]
```
xumingkuan, line 53: Will this cause some issue because of the shallow copying in python?

liaopeiyuan: What do you mean by shallow copying? Each `None` in entries should be unique, e.g.

```
>>> a = [[None]*4 for _ in range(4)]
>>> a
[[None, None, None, None], [None, None, None, None], [None, None, None, None], [None, None, None, None]]
>>> a[1][1]=2
>>> a
[[None, None, None, None], [None, 2, None, None], [None, None, None, None], [None, None, None, None]]
>>> 
```

python/taichi/lang/matrix.py
```diff
@@ -627,14 +662,11 @@ def outer_product(self, b):
...
-Vector.unit = Matrix.unit
-Vector.normalized = Matrix.normalized
+class Vector(Matrix):
+    def __init_(n, dt=None, shape=None, offset=None, **kwargs):
+        super(Vector, self).__init__(n,
+                                     1,
+                                     dt=dt,
+                                     shape=shape,
+                                     offset=offset,
+                                     **kwargs)
```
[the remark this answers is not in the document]
archibate: Disagree, unless you give me a Pros vs Cons comparison.

python/taichi/lang/matrix.py
archibate: No worry. None is not a pointer so it's ok.

python/taichi/lang/matrix.py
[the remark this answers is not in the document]
yuanming-hu: Thanks for your reply.

The *subclass approach* has been discussed in #1046 yet no consensus has been made. This indicates neither of the options (using a subclass or not)  has dominating advantages over the other. In practice, probably either option is fine.

We should be decisive here to move things forward. As the BDFL of Taichi, I would do my job to stop the debt and make a decision: we should not use the *subclass approach*, and Vectors are not substantially different from matrices. 

Therefore, in this PR we should stick to the original version and make `Vector` a function for now.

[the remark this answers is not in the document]
k-ye: I want to reiterate the guideline that we should focus one PR for one thing. And the thing we need to focus here is to improve the `Matrix` initializers.

Speaking of vector vs matrix, if this causes so much discussion, i wonder if we should follow numpy and just provide an `ti.array`...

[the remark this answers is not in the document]
archibate: Good idea, we may ultimately come to `Array` in #777, and no such convention needed.

liaopeiyuan:
I have reverted the changes and CI should be working right now.

archibate:
Thank for the hard work! @yuanming-hu I'm merging this now to prevent from being stale.

Issue #1128, opened 2020-06-03
Something wrong with Cornell box example


leo-luo-lzy:
- os: macOS Catalina 10.15.5
- Python version: 3.7.0
- taichi version: 0.6.7

It seems that there is something wrong with the example 'Cornell box' .
Thanks for your help in advance.
![Screenshot 2020-06-03 at 4 38 35 PM](https://user-images.githubusercontent.com/66378748/83664341-74509080-a5fc-11ea-863a-b7d5e3436dc6.png)
![Screenshot 2020-06-03 at 4 44 15 PM](https://user-images.githubusercontent.com/66378748/83664381-7f0b2580-a5fc-11ea-9198-7e89b1891555.png)

rexwangcc:
The error itself seems like a numpy warning indicating `img / img.mean() * 0.24` got a negative number therefore all `var` become `nan`. But I failed to reproduce it with:
```
[Taichi] mode=release
[Taichi] version 0.6.7, supported archs: [cpu, metal], commit ca4d9dda, python 3.7.7
```

TH3CHARLie:
Would you mind sharing your numpy info via `pip show numpy`? I can not reproduce this on my mac either.

leo-luo-lzy:
> pip show numpy

![Screenshot 2020-06-04 at 1 26 21 AM](https://user-images.githubusercontent.com/66378748/83668392-71589e80-a602-11ea-9e18-9b7d172ac768.png)
Here it is. Thank you!

k-ye:
Hi,

Thank you for reporting this!

This is an issue due to the graphics card. My MBP has two cards:

1. AMD Radeon Pro 5500M
2. Intel(R) UHD Graphics 630

I can repro this on 2, but not 1. My wild guess would be that this was due to some floating point arithmetics differences. Maybe some denominators were close to zero and led to this problem? (@yuanming-hu have you encountered such kind of errors before? I.e. numerical rounding errors led to NaN..? )

Unfortunately, my understanding of Metal is not enough for me to easily spot where the problem was :( It is even possible that Apple had a poor support on the integrated graphics cards.. (We once observed that Metal's memory order went completely nonsense on an old MBP model)

I could make a patch to disable direct light sampling in this case. The result will be more noisy, but hopefully it can at least present a reasonable result.

bobcao3:
Can not repro, if this is still a issue please reopen

PR #1139, opened 2020-06-04
[metal] Add APIs to list all the available Metal devices

k-ye:
When debugging #1128 , I find it useful to be able to select which GPU to run. This PR adds the necessary APIs for us to choose GPUs.

Related issue = #1128 

### DIFF
--- a/taichi/backends/metal/api.cpp
+++ b/taichi/backends/metal/api.cpp
@@ -9,6 +9,7 @@ namespace metal {
 
 extern "C" {
 id MTLCreateSystemDefaultDevice();
+id MTLCopyAllDevices();
 }
 
 namespace {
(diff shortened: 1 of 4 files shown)


k-ye:
> how do I specify which device to use?

You cannot yet. I'm not adding that functionality here.

I think this is a general problem for all GPU backends, not just Metal. Pytorch allows you to choose which device to run by a [`set_device()`](https://pytorch.org/docs/stable/cuda.html#torch.cuda.set_device) API. Maybe we can have similar things in Taichi, or even simpler `ti.init(arch=ti.gpu, device_id=0, ...)`.

archibate:
Yes, it would be a useful feature, but highly backend-dependent.
FYI, for OpenGL, I have to choose NV / Intel-integrated card by the `optirun` util provided by NV:
`optirun python x.py` (NV card)
`python x.py` (Intel card)

k-ye:
> use device instead of arch

I guess you mean to use `device` in addition to `arch`, not to replace `arch`..? Otherwise that's an aggressive step..

> I have to choose NV / Intel-integrated card by the optirun util provided by NV:

I see, that is unfortunate. TBH, I think OpenGL is the outlier here, due to historical reasons. I guess this global state machine design didn't really take into account the scenarios of multiple GPU devices.. All major platforms, like CUDA, Metal or Vulkan, have provided a way to select devices.

Anyway, I have filed #1154 , we can move the discussion there.

PR #1144, opened 2020-06-04
[llvm] Cache llvm runtime bitcode to reduce start up time in dev mode

### LINKED ISSUE #875: [misc] Cache compiled runtime bitcode runtime_x64.bc
Currently `compile_runtime_bitcode` cost a long time to compile.
Also note that `compile_runtime_bitcode` is called even if I use non-LLVM backend like `ti.opengl`.
I would like to cache the compiled bitcode so that it won't recompile every time so that startup time can be reduced.

Cache `runtime_{x64,cuda}.bc`, instead of putting it in `/tmp/taichi-xxx`.
Also please disable `compile_runtime_bitcode` on non-LLVM backend.

**Additional comments**
My built-from-scratch `clang-8` is super slow (bug?) when compiling Taichi... but fine with single `cpp` files compared to `gcc`.
And that's why this issue become serious on my local machine..

archibate:
Related issue = close #875 

I hate debugging script in dev mode, just because this huge long start up time compiling bitcode everytime I restart..

### DIFF
--- a/docs/global_settings.rst
+++ b/docs/global_settings.rst
@@ -7,19 +7,19 @@ Backends
 - To specify which Arch to use: ``export TI_ARCH=cuda`` or ``ti.init(arch=ti.cuda)``.
 - To specify which GPU to use for CUDA: ``export CUDA_VISIBLE_DEVICES=[gpuid]``.
 - To specify pre-allocated memory size for CUDA: ``export TI_DEVICE_MEMORY_GB=0.5`` or ``ti.init(device_memory_GB=0.5)``.
-- To not use unified memory for CUDA: ``export TI_USE_UNIFIED_MEMORY=0``.
+- To disable unified memory usage on CUDA: ``export TI_USE_UNIFIED_MEMORY=0``.
 
 Compilation
 ***********
 
 - Disable advanced optimization to save compile time & possible erros: ``ti.core.toggle_advanced_optimization(False)``.
 - To print intermediate IR generated: ``export TI_PRINT_IR=1`` or ``ti.init(print_ir=True)``.
-- To print preprocessed Python codes: ``export TI_PRINT_PREPROCESSED=1`` or ``ti.init(print_preprocessed=True)``..
+- To print preprocessed Python code: ``export TI_PRINT_PREPROCESSED=1`` or ``ti.init(print_preprocessed=True)``..
 
 Runtime
 *******
 
-- Restart the Taichi runtime system (clear memory, destroy all variables and kernels): ``ti.reset()``.
+- Restart the entire Taichi system (destroy all tensors and kernels): ``ti.reset()``.
 - To start program in debug mode: ``export TI_DEBUG=1`` or ``ti.init(debug=True)``.
 
 Logging
--- a/docs/global_settings.rst
+++ b/docs/global_settings.rst
@@ -1,14 +1,35 @@
 Global settings
 ---------------
 
-- Disable advanced optimization to save compile time: ``ti.core.toggle_advanced_optimization(False)``
-- Restart the Taichi runtime system (clear memory, destroy all variables and kernels): ``ti.reset()``
-- Eliminate verbose outputs: ``ti.get_runtime().set_verbose(False)``
-- To not trigger GDB when crashes: ``export TI_GDB_TRIGGER=0``
-- To not use unified memory for CUDA: ``export TI_USE_UNIFIED_MEMORY=0``
-- To specify pre-allocated memory size for CUDA: ``export TI_DEVICE_MEMORY_GB=0.5``
-- Show more detailed log (TI_TRACE): ``export TI_LOG_LEVEL=trace``
-- To specify which GPU to use for CUDA: ``export CUDA_VISIBLE_DEVICES=[gpuid]``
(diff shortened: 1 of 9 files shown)


archibate:
Feel free to say no without a reason if you don't like this, just give me a walk around to disable these redundant compiling process..

yuanming-hu:
> Feel free to say no without a reason if you don't like this, just give me a walk around to disable these redundant compiling process..

I think maybe we can add an environment variable `TI_CACHE_RUNTIME_BITCODE` to control this? By default, it shouldn't be cached. Improving your development experience is important, but we also want to simplify the workflow for people who don't care about start up time.

> Btw, guess what, I found that I'm becoming the most frequent commiter this week according to https://github.com/taichi-dev/taichi/pulse...

Thanks for everything you have done for Taichi :-)

archibate:
> TI_CACHE_RUNTIME_BITCODE

I'm fine with this approach. But why we neeed compile rtbc each time restart? Can we just let Makefile care this?

archibate:
Done, btw, I found no test about `@` (matmul) in linalg, is this intensional? Also note that `z = z @ w` seems to fail due to l-value problems.

archibate:
May I merge this now? I suck the compile time so much that I have to watch bilibili before a program starts, please approve me asap if you feel everything is ok now.

docs/global_settings.rst
```diff
@@ -1,14 +1,35 @@
...
-- To specify which Arch to use: ``export TI_ARCH=cuda``
-- To print intermediate IR generated: ``export TI_PRINT_IR=1``
-- To print verbose details: ``export TI_VERBOSE=1``
+Backends
+********
+
+- To specify which Arch to use: ``export TI_ARCH=cuda`` or ``ti.init(arch=ti.cuda)``.
+- To specify which GPU to use for CUDA: ``export CUDA_VISIBLE_DEVICES=[gpuid]``.
+- To specify pre-allocated memory size for CUDA: ``export TI_DEVICE_MEMORY_GB=0.5`` or ``ti.init(device_memory_GB=0.5)``.
+- To not use unified memory for CUDA: ``export TI_USE_UNIFIED_MEMORY=0``.
```
yuanming-hu, line 10:
```suggestion
- To disable unified memory usage on CUDA: ``export TI_USE_UNIFIED_MEMORY=0``.
```

docs/global_settings.rst
```diff
@@ -1,14 +1,35 @@
...
+
+Compilation
+***********
+
+- Disable advanced optimization to save compile time & possible erros: ``ti.core.toggle_advanced_optimization(False)``.
+- To print intermediate IR generated: ``export TI_PRINT_IR=1`` or ``ti.init(print_ir=True)``.
+- To print preprocessed Python codes: ``export TI_PRINT_PREPROCESSED=1`` or ``ti.init(print_preprocessed=True)``..
```
yuanming-hu, line 17:
```suggestion
- To print preprocessed Python code: ``export TI_PRINT_PREPROCESSED=1`` or ``ti.init(print_preprocessed=True)``..
```

docs/global_settings.rst
```diff
@@ -1,14 +1,35 @@
...
+
+Runtime
+*******
+
+- Restart the Taichi runtime system (clear memory, destroy all variables and kernels): ``ti.reset()``.
```
yuanming-hu, line 22:
```suggestion
- Restart the entire Taichi system (destroy all tensors and kernels): ``ti.reset()``.
```

k-ye:
This PR has broken the CI, see:

* https://travis-ci.com/github/taichi-dev/taichi/jobs/343699535
* https://github.com/taichi-dev/taichi/runs/744663826
* https://github.com/taichi-dev/taichi/pull/1155/checks?check_run_id=744629126

Please be mindful of `[skip ci]`, abusing it could hide the presubmit errors

archibate:
Sorry for breaking CI! Maybe I shouldn't add `[skip ci]` on the last commit of before merge...
Any idea why `<filesystem>` is not found on OS X? It should be supported since C++17 according to https://en.cppreference.com/w/cpp/header/filesystem, or OS X doesn't support C++17 yet?

k-ye:
Hmm, my feeling is that it's a Travis CI problem, maybe their clang is stuck at a super old version? Currently the master branch compiles just fine on my local env. 

I found this post (https://stackoverflow.com/a/49600206/12003165): `<filesystem>` used to be in `<experimental/filesystem>`, so maybe we need to use that on Mac?  

> Maybe I shouldn't add [skip ci] on the last commit of before merge...

+1...

Rullec:
2 weeks ago I met with this problem as well. It seems that <filesystem>
this header can only be used directly on macOS 10.15 Catalina, instead of
my 10.14 Mojave. It's fully about the version.

Ye Kuang <notifications@github.com> 于2020年6月6日周六 下午3:52写道：

k-ye:
>  It seems that <filesystem>
this header can only be used directly on macOS 10.15 Catalina, instead of
my 10.14 Mojave. It's fully about the version.

Thanks for the info. Yeah I saw that Travis CI is stilling using [XCode 10.3](https://developer.apple.com/documentation/xcode_release_notes/xcode_10_3_release_notes?language=objc), which only covers up to Mac 10.14.6

PR #1145, opened 2020-06-04
[IR] Avoid promoting both operands of "pow" to the same type

### LINKED ISSUE #1141: [ir] How does the specialized pow_{type}() gets picked?
**Describe the bug**

I'm not sure how `fast_pow_{type}` gets picked in cases where the fast version can be used.

For example, I'm testing with @archibate 's example in https://zhuanlan.zhihu.com/p/145222094:

```py
@ti.kernel
def calc_pi() -> ti.f32:
  sum = 0.0
  for i in range(66666666):
    n = 2 * i + 1
    sum += pow(-1.0, i) / n
  return sum * 4
```

I thought this `pow` will be specialized to `fast_pow_f32`, because lhs is an `f32` and rhs is an `i32`. However, the generated IR inserts a `cast_value` to the rhs, converting it from an `i32` to `f32`:

```
<i32 x1> $5 = loop $4 index 0
...
<f32 x1> $9 = cast_value<f32> $5  <-- !
<f32 x1> $10 = const [1.0]
<f32 x1> $11 = pow $10 $9
```

Therefore I wonder if these fast pows are ever used in practice?

xumingkuan:
Related issue = fix #1141

I think this solution looks a bit ad-hoc, and it may break other backends...

Do you have better solutions?

**Test case:**
```python
@ti.kernel
def calc_pi() -> ti.f32:
  sum = 0.0
  for i in range(66666666):
    n = 2 * i + 1
    sum += pow(-1.0, i) / n
  return sum * 4
```
Before:
```
[I 06/04/20 17:17:42.365] [compile_to_offloads.cpp:taichi::lang::irpass::co
mpile_to_offloads::<lambda_44aeda62b30289419801af51381ceafc>::operator ()@2
1] Simplified III:
kernel {
  $0 = offloaded  {
    <f32*x1> $1 = global tmp var (offset = 0 B)
    <f32 x1> $2 = const [0.0]
    <f32*x1> $3 : global store [$1 <- $2]
  }
  $4 = offloaded range_for(0, 66666666) block_dim=adaptive {
    <i32 x1> $5 = loop $4 index 0
    <i32 x1> $6 = add $5 $5
    <i32 x1> $7 = const [1]
    <i32 x1> $8 = add $6 $7
    <f32 x1> $9 = cast_value<f32> $5
    <f32 x1> $10 = const [-1.0]
    <f32 x1> $11 = pow $10 $9
    <f32 x1> $12 = cast_value<f32> $8
    <f32 x1> $13 = div $11 $12
    <f32*x1> $14 = global tmp var (offset = 0 B)
    <f32 x1> $15 = atomic add($14, $13)
  }
  $16 = offloaded  {
    <f32*x1> $17 = global tmp var (offset = 0 B)
    <f32 x1> $18 = global load $17
    <f32 x1> $19 = const [4.0]
    <f32 x1> $20 = mul $18 $19
    <f32 x1> $21 : kernel return $20
  }
}
3.141597032546997
```
After:
```
[I 06/04/20 17:54:50.889] [compile_to_offloads.cpp:taichi::lang::irpass::co
mpile_to_offloads::<lambda_44aeda62b30289419801af51381ceafc>::operator ()@2
1] Simplified III:
kernel {
  $0 = offloaded  {
    <f32*x1> $1 = global tmp var (offset = 0 B)
    <f32 x1> $2 = const [0.0]
    <f32*x1> $3 : global store [$1 <- $2]
  }
  $4 = offloaded range_for(0, 66666666) block_dim=adaptive {
    <i32 x1> $5 = loop $4 index 0
    <i32 x1> $6 = add $5 $5
    <i32 x1> $7 = const [1]
    <i32 x1> $8 = add $6 $7
    <i64 x1> $9 = cast_value<i64> $5
    <f32 x1> $10 = const [-1.0]
    <f32 x1> $11 = pow $10 $9
    <f32 x1> $12 = cast_value<f32> $8
    <f32 x1> $13 = div $11 $12
    <f32*x1> $14 = global tmp var (offset = 0 B)
    <f32 x1> $15 = atomic add($14, $13)
  }
  $16 = offloaded  {
    <f32*x1> $17 = global tmp var (offset = 0 B)
    <f32 x1> $18 = global load $17
    <f32 x1> $19 = const [4.0]
    <f32 x1> $20 = mul $18 $19
    <f32 x1> $21 : kernel return $20
  }
}
3.141596555709839
```

### DIFF
--- a/taichi/runtime/llvm/runtime.cpp
+++ b/taichi/runtime/llvm/runtime.cpp
@@ -268,6 +268,38 @@ f64 pow_f64(f64 a, f64 b) {
   return std::pow(a, b);
 }
 
+f32 pow_f32_i64(f32 x, i64 n) {
+  if (n > 512)
+    return std::pow(x, n);
+  bool neg = n < 0;
+  n = abs(n);
+  f32 tmp = x;
+  f32 ans = 1;
+  while (n) {
+    if (n & 1)
+      ans *= tmp;
+    tmp *= tmp;
+    n >>= 1;
+  }
+  return neg ? 1.0 / ans : ans;
+}
+
+f64 pow_f64_i64(f64 x, i64 n) {
+  if (n > 512)
+    return std::pow(x, n);
+  bool neg = n < 0;
+  n = abs(n);
+  f64 tmp = x;
+  f64 ans = 1;
+  while (n) {
+    if (n & 1)
+      ans *= tmp;
+    tmp *= tmp;
+    n >>= 1;
+  }
+  return neg ? 1.0 / ans : ans;
+}
+
 f32 __nv_sgnf(f32 x) {
   return sgn_f32(x);
 }
--- a/taichi/codegen/codegen_llvm.cpp
+++ b/taichi/codegen/codegen_llvm.cpp
@@ -450,9 +450,9 @@ void CodeGenLLVM::visit(BinaryOpStmt *stmt) {
     if (arch_is_cpu(current_arch())) {
       if (ret_type == DataType::f32) {
         if (stmt->rhs->ret_type.data_type == DataType::f32) {
-          llvm_val[stmt] =
-              create_call("pow_f32", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
-        } else if (stmt->rhs->ret_type.data_type == DataType::i64){
+          llvm_val[stmt] = create_call(
+              "pow_f32", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
+        } else if (stmt->rhs->ret_type.data_type == DataType::i64) {
           llvm_val[stmt] = create_call(
               "pow_f32_i64", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
         } else {
@@ -461,9 +461,9 @@ void CodeGenLLVM::visit(BinaryOpStmt *stmt) {
         }
       } else if (ret_type == DataType::f64) {
         if (stmt->rhs->ret_type.data_type == DataType::f64) {
-          llvm_val[stmt] =
-              create_call("pow_f64", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
-        } else if (stmt->rhs->ret_type.data_type == DataType::i64){
+          llvm_val[stmt] = create_call(
+              "pow_f64", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
+        } else if (stmt->rhs->ret_type.data_type == DataType::i64) {
           llvm_val[stmt] = create_call(
               "pow_f64_i64", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
         } else {
--- a/taichi/codegen/codegen_llvm.cpp
+++ b/taichi/codegen/codegen_llvm.cpp
@@ -449,11 +449,27 @@ void CodeGenLLVM::visit(BinaryOpStmt *stmt) {
   } else if (op == BinaryOpType::pow) {
     if (arch_is_cpu(current_arch())) {
       if (ret_type == DataType::f32) {
-        llvm_val[stmt] =
-            create_call("pow_f32", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
+        if (stmt->rhs->ret_type.data_type == DataType::f32) {
+          llvm_val[stmt] =
+              create_call("pow_f32", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
+        } else if (stmt->rhs->ret_type.data_type == DataType::i64){
+          llvm_val[stmt] = create_call(
+              "pow_f32_i64", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
+        } else {
+          TI_P(data_type_name(stmt->rhs->ret_type.data_type));
+          TI_NOT_IMPLEMENTED
+        }
       } else if (ret_type == DataType::f64) {
-        llvm_val[stmt] =
-            create_call("pow_f64", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
+        if (stmt->rhs->ret_type.data_type == DataType::f64) {
+          llvm_val[stmt] =
+              create_call("pow_f64", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
+        } else if (stmt->rhs->ret_type.data_type == DataType::i64){
+          llvm_val[stmt] = create_call(
+              "pow_f64_i64", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
+        } else {
+          TI_P(data_type_name(stmt->rhs->ret_type.data_type));
+          TI_NOT_IMPLEMENTED
+        }
       } else if (ret_type == DataType::i32) {
         llvm_val[stmt] =
             create_call("pow_i32", {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
--- a/taichi/transforms/type_check.cpp
+++ b/taichi/transforms/type_check.cpp
@@ -248,15 +248,24 @@ class TypeCheck : public IRVisitor {
       stmt->op_type = BinaryOpType::div;
     }
 
+    auto ret_type = stmt->lhs->ret_type.data_type;
     if (stmt->lhs->ret_type.data_type != stmt->rhs->ret_type.data_type) {
-      auto ret_type = promoted_type(stmt->lhs->ret_type.data_type,
-                                    stmt->rhs->ret_type.data_type);
+      ret_type = promoted_type(stmt->lhs->ret_type.data_type,
+                               stmt->rhs->ret_type.data_type);
       if (ret_type != stmt->lhs->ret_type.data_type) {
-        // promote rhs
+        // promote lhs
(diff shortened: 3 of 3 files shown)


taichi/runtime/llvm/runtime.cpp
```diff
@@ -268,6 +268,38 @@ f64 pow_f64(f64 a, f64 b) {
   return std::pow(a, b);
 }
 
+f32 pow_f32_i64(f32 x, i64 n) {
```
archibate, line 271: Suggestion: pow_f32_i32?
We can clamp if rhs is i64 since pow result will overflow if rhs > INT_MAX anyway.

taichi/runtime/llvm/runtime.cpp
```diff
@@ -268,6 +268,38 @@ f64 pow_f64(f64 a, f64 b) {
...
+    if (n & 1)
+      ans *= tmp;
+    tmp *= tmp;
+    n >>= 1;
+  }
+  return neg ? 1.0 / ans : ans;
+}
+
+f64 pow_f64_i64(f64 x, i64 n) {
+  if (n > 512)
```
archibate, line 288: if (abs(n) > 512)

archibate:
Btw, could you post a benchmark time result to the PR desc?

taichi/runtime/llvm/runtime.cpp
xumingkuan: Oh yes. (1 + 2^-23)^(2^31) = 1.5e+111 is not representable in f32.

xumingkuan:
> We can clamp if rhs is i64

Shall we cast it to i64 in typecheck, and then clamp it back in runtime.cpp?

I think casting to i32 in typecheck is better. But what happens if a number like 2^50 is cast to i32? Will pow(2, cast(2^50, i32)) result in INF as it should be?

xumingkuan:
> Btw, could you post a benchmark time result to the PR desc?

I think my laptop is not very suitable for time benchmarking... Maybe you can do it on desktop computers for a more accurate result?

xumingkuan:
BTW, as you can see, this PR changes the result of the test case. Will it be an issue?

archibate:
> pow(2, cast(2^50, i32))

I guess cast(2^50, i32) == 0?

> I think my laptop is not very suitable for time benchmarking... Maybe you can do it on desktop computers for a more accurate result?

Sure :) Will do once I have time.

> BTW, as you can see, this PR changes the result of the test case. Will it be an issue?

What do you mean by changes the result? SPRT?

archibate:
> I'm curious about why i64 instead of i32, and yes -- this would be problematic on Metal, since it's not supporting 64-bit data yet. And I think most of the times we would default to 32-bit data?

+1, OpenGL without GL_ARB_gpu_shader_int64 extension will also be broken on this.
I think the best solution could be:
1. narrowing `f ** i64` to `f ** i32`, since rhs `2**32` is overflow anyway.
2. expanding `f ** i64` to `f ** f64`, wdyt?

archibate:
```
**      Taichi Programming Language      **
x64::root_listgen_________________________________________
time_avg                      1.2e+01 ->     9.5    -20.7%

x64::nested_struct_listgen_8x8____________________________
time_avg                        1e+01 ->   1e+01     -1.2%

x64::nested_struct_listgen_16x16__________________________
time_avg                      1.1e+01 -> 1.1e+01     -1.3%

x64::nested_struct_fill_and_clear_________________________
time_avg                      9.3e+01 -> 9.5e+01     +2.1%

x64::nested_struct________________________________________
time_avg                      2.5e+01 -> 2.4e+01     -2.1%

x64::nested_range_blocked_________________________________
time_avg                      1.4e+01 -> 1.3e+01     -8.1%

x64::nested_range_________________________________________
time_avg                      1.3e+01 -> 1.2e+01     -1.9%

x64::memcpy_______________________________________________
time_avg                      1.4e+02 -> 1.4e+02     +1.4%

x64::flat_struct__________________________________________
time_avg                      1.2e+01 -> 1.1e+01     -2.5%

x64::flat_range___________________________________________
time_avg                      1.2e+01 -> 1.2e+01     +1.4%

opengl::root_listgen______________________________________
time_avg                       0.0051 ->  0.0056     +9.8%

opengl::nested_struct_listgen_8x8_________________________
time_avg                       0.0044 ->  0.0047     +6.8%

opengl::nested_struct_listgen_16x16_______________________
time_avg                       0.0044 ->  0.0045     +2.3%

opengl::nested_struct_____________________________________
time_avg                       0.0042 ->  0.0044     +4.8%

opengl::nested_range_blocked______________________________
time_avg                       0.0054 ->  0.0053     -1.9%

opengl::flat_struct_______________________________________
time_avg                       0.0067 ->  0.0068     +1.5%

opengl::flat_range________________________________________
time_avg                       0.0065 ->  0.0064     -1.5%

cuda::root_listgen________________________________________
time_avg                          4.4 ->     4.4     -0.0%

cuda::nested_struct_listgen_8x8___________________________
time_avg                          4.4 ->     4.4     +0.1%

cuda::nested_struct_listgen_16x16_________________________
time_avg                          4.4 ->     4.4     -0.0%

cuda::nested_struct_fill_and_clear________________________
time_avg                        2e+02 ->   2e+02     +0.0%

cuda::nested_struct_______________________________________
time_avg                          7.1 ->     7.1     -0.4%

cuda::nested_range_blocked________________________________
time_avg                      1.5e+01 -> 1.5e+01     +0.0%

cuda::nested_range________________________________________
time_avg                          4.4 ->     4.4     -0.0%

cuda::flat_struct_________________________________________
time_avg                          4.4 ->     4.4     +0.0%

cuda::flat_range__________________________________________
time_avg                          4.5 ->     4.5     -0.2%
```
sdf: 7.01 -> 7.01
no obvious difference... strange... I guess it's because `std::pow` is already optimized when rhs close to int?

xumingkuan:
> I guess it's because `std::pow` is already optimized when rhs close to int?

Yes. See https://baptiste-wicht.com/posts/2017/09/cpp11-performance-tip-when-to-use-std-pow.html.

So do we enable -ffast-math by default? If the answer is yes, maybe this PR is unnecessary.

archibate:
I'm not an expert on this... but ultimately we should done this in alg_simp, right?
Btw, I found `abs` raised and warning about `int64_t` when compile, could you rename it to `std::abs` to fit int64_t?

xumingkuan:
I think we should close this PR and improve the strength reduction for `pow(a, b)` where `b` is a **compile-time constant** in `alg_simp` (specifically, extend this optimization to `b` being an integer with `real` data type), and just leave the work of optimizing `pow(a, b)` where `b` is not a compile-time constant but an integer to `std::pow`. WDYT?

archibate:
>  think we should close this PR and improve the strength reduction for pow(a, b) where b is a compile-time constant in alg_simp

Yes, thanks, **directly do opts in alg_simp** sounds better to me.

> specifically, extend this optimization to b being an integer with real data type

Btw, will we opt `pow(x, 0.5)` -> `sqrt(x)`? And `pow(x, 2.5)` -> `x * x * sqrt(x)`?

xumingkuan:
> > specifically, extend this optimization to b being an integer with real data type
> 
> Btw, will we opt `pow(x, 0.5)` -> `sqrt(x)`? And `pow(x, 2.5)` -> `x * x * sqrt(x)`?

Sounds good. I'll have a try in the next week.

BTW, because of our typecheck cast the value to float32/64, and (of course) alg_simp needs constant_fold to JIT do the cast (otherwise it won't know that it's a compile-time const), I want to make sure that for every small integer `x` in [-32, 32], after we JIT cast it to float32, it's still an integer (not something like `x + eps`) on all backends?

archibate:
I just realized that, we can't apply this to non constant rhs int if using the `alg_simp` approach. e.g. the example of `pow(-1, i)` in my zhuanlan.

archibate:
> I want to make sure that for every small integer x in [-32, 32], after we JIT cast it to float32, it's still an integer (not something like x + eps) on all backends?

It should be, feel free to add a C++ test to check this, on all_archs.

xumingkuan:
> I just realized that, we can't apply this to non constant rhs int if using the `alg_simp` approach. e.g. the example of `pow(-1, i)` in my zhuanlan.

Yes, although `pow(-(1 + eps), 2^50)` is `INF` and `pow(-(1 - eps), 2^50)` is `0`, we can't just clamp the rhs to int32 because `pow(-1, n)` needs the parity of `n`.

archibate:
> because pow(-1, n) needs the parity of n.

Interesting, but I think `n & (2**32 - 1)` has the same parity with `n`? So narrowing down to i32 doesn't affect real numbers like `-1`, right?

k-ye:
Can we conclude that:

* None of this, #1141 or fast pow specializations are necessary (sorry about the wasted efforts..!)
* We should aim to simplify `pow` in the alg simplification pass, e.g. (x ** 2 --> x * x, although i think this is already done?). Maybe open a separate issue to track this?

xumingkuan:
> > because pow(-1, n) needs the parity of n.
> 
> Interesting, but I think `n & (2**32 - 1)` has the same parity with `n`? So narrowing down to i32 doesn't affect real numbers like `-1`, right?

`pow(1.1, 2**32)` is `INF`, but `pow(1.1, (2**32)&(2**32 - 1))` is `1`...

xumingkuan:
Closing because `std::pow` already provides some optimizations.

PR #1152, opened 2020-06-05
[GUI] Add ti.GUI.EXIT event to close a window

### LINKED ISSUE #1118: [GUI] closing a window cause crash
Currently closing a window using the 'x' button right-top will cause the program to crash.
This confuses the jupyter users which doesn't re-import taichi after restart.
Also note that if user close window with 'esc', the window won't disappear for jupyter users due to the same reason.

Make window close a event/exception. If the event is not processed/catched, automatically break-the-loop/exit-the-program.

**Additional comments**
As mentioned in the GAMES 201 WeChat group.

archibate:
Related issue = #1118

@xumingkuan Could you do me a favor by testing this sample code:
```py
gui = ti.GUI("Hello")
while gui.running:
   gui.show()

print('GUI closed')
```
when you have time? Many thanks :)

@k-ye Would you mind to add this feature to cocoa following the win32&x11 as example itpr / iapr? No rush at all, any time before v0.6.8 is ok :)


taichi/gui/gui.h
```diff
@@ -777,6 +779,18 @@ class GUI : public GUIBase {
...
     last_frame_time = taichi::Time::get_time();
     redraw();
+    // Some old examples / users don't even provide a `break` statement for us
+    // to terminate loop. So we have to terminate the program with RuntimeError
+    // if ti.GUI.EXIT event is not processed. Pretty like SIGTERM, you can hook
+    // it, but you have to terminate after your handler is done.
+    if (should_close) {
+      if (++should_close >
+          5) {  // if the event is not processed in 5 frames, raise RuntimeError
+        throw std::string(
```
k-ye, line 789: I haven't followed up on the Jupyter issue, so my comment here might not cover the whole picture. Is throwing an exception the best way to solve whatever Jupyter' issue is.? Below are what I know so far:

1. Please don't use exception in C++, it has caused lots of headaches. See #1059 , #655
2. According to the comment, I believe it's the examples that need to be fixed...This logic doesn't seem generic enough to be placed at the GUI level. Based on your understanding, is it possible to fix the examples for Jupyter users?

taichi/gui/x11.cpp
```diff
@@ -114,6 +123,10 @@ void GUI::create_window() {
                ButtonPressMask | ExposureMask | KeyPressMask | KeyReleaseMask |
                    ButtonPress | ButtonReleaseMask | EnterWindowMask |
                    LeaveWindowMask | PointerMotionMask);
+  wmDeleteMessage = std::malloc(sizeof(Atom));
```
k-ye, line 126: RAII: Use `std::vector<char>` to allocate a memory of `sizeof(Atom)`?

xumingkuan:
I can't close the window :(

Hitting the "close" button has no effect.

taichi/gui/gui.h
archibate: It's not the examples need to be fixed, it's our users' legacy codes need to be fixed.
They may write:
```
while True:
  gui.show()
```
I want them to write `while gui.running`, but I can't, since they may not know this change.
We have to throw an exception in `gui.show()` to break the infinite loop, no other option, sadly.
BTW FYI `std::exception` here is not a C++ exception, it's a `RuntimeError` in Python exception, according to pybind11.

taichi/gui/win32.cpp
```diff
@@ -122,8 +122,11 @@ LRESULT CALLBACK WindowProc(HWND hwnd,
                                               gui->cursor_pos});
       break;
     case WM_CLOSE:
-      exit(0);
-      break;
+      gui->key_events.push_back(GUI::KeyEvent{GUI::KeyEvent::Type::press,
+                                              "WMClose", gui->cursor_pos});
+      // https://stackoverflow.com/questions/3155782/what-is-the-difference-between-wm-quit-wm-close-and-wm-destroy-in-a-windows-pr
```
archibate, line 127:
```suggestion
      gui->should_close++;
      // https://stackoverflow.com/questions/3155782/what-is-the-difference-between-wm-quit-wm-close-and-wm-destroy-in-a-windows-pr
```

archibate:
> I can't close the window :(
> 
> Hitting the "close" button has no effect.

Sorry! It should works now.

xumingkuan:
> > I can't close the window :(
> > Hitting the "close" button has no effect.
> 
> Sorry! It should works now.

Interesting... `GUI closed` is printed after I hit the button, but the GUI is not closed.

archibate:
How about this? `DestroyWindow`?

xumingkuan:
Same behavior after the new commits...

`GUI closed` is printed and the window becomes "no response" after I hit the close button.

xumingkuan:
It works!!!

archibate:
Great! I'll merge this for now once @yuanming-hu approves me, and please add a Mac version **before Jun 7 we release v0.6.8** if someone feel interested :)

docs/gui.rst
```diff
@@ -276,6 +276,24 @@ A *event filter* is a list combined of *key*, *type* and *(type, key)* tuple, e.
...
+    :parameter gui: (GUI)
+    :return: (bool) ``True`` if ``ti.GUI.EXIT`` event occurred, vice versa
+
+    For example:
+
+    ::
+
+        while gui.running:
+            if gui.get_event(ti.GUI.ESCAPE):
+                gui.running = False
```
xumingkuan, line 290: Do we automatically update `gui.running`, or should users update it in these lines?

xumingkuan: Oh it's `ti.GUI.ESCAPE`, not the close button.

archibate: We do, the `gui.running` is a getter + setter. Thank for pointing out, we should make the doc more clear here.

taichi/gui/gui.h
k-ye: I won't change my opinion that "if closed 5 times, quit the GUI" is an application level logic and doesn't belong to GUI. But I see your point that we need this to make the old code work. I'm fine with having this for Games 201, but I think eventually, we should aim for:

* make `running` read-only
* mark `running` to `false` if we've received the close event
* update the doc so that users know they should use `while gui.running:` if they want the program to be interactive

Hopefully this makes the code cleaner?

> FYI std::exception here is not a C++ exception, it's a RuntimeError in Python exception, according to pybind11.

Ah i see.

archibate: t4po, will iapr.

> update the doc so that users know they should use while gui.running: if they want the program to be interactive

This is already done in both doc and error-message.

taichi/gui/x11.cpp
```diff
@@ -73,6 +74,14 @@ void GUI::process_event() {
     switch (ev.type) {
       case Expose:
         break;
+      case ClientMessage:
+        // https://stackoverflow.com/questions/10792361/how-do-i-gracefully-exit-an-x11-event-loop
+        if (ev.xclient.data.l[0] == *(Atom *)wmDeleteMessage.data()) {
+          key_events.push_back(
```
k-ye, line 80: Could you wrap this logic as a member function for `GUI`? I implemented something for Cocoa, and had to reuse this bit of logic..

docs/gui.rst
```diff
@@ -201,6 +201,37 @@ A *event filter* is a list combined of *key*, *type* and *(type, key)* tuple, e.
     gui.get_event((ti.GUI.PRESS, ti.GUI.ESCAPE), (ti.GUI.RELEASE, ti.GUI.SPACE))
 
 
+.. attribute:: gui.running
+
+    :parameter gui: (GUI)
+    :return: (bool) ``True`` if ``ti.GUI.EXIT`` event occurred, vice versa
+
+    ``ti.GUI.EXIT`` occurs when you click on the close button / X button of a widnow.
```
xumingkuan, line 209:
```suggestion
    ``ti.GUI.EXIT`` occurs when you click on the close (X) button of a window.
```

PR #1155, opened 2020-06-05
[refactor] Move log2int into arithmatic.h

k-ye:
Related issue = N/A
WANT_LGTM=any


k-ye:
I actually have a question. Can `runtime.cpp` include a header whose implementation is in `.cpp`..?

taichi/math/arithmetic.h
```diff
@@ -14,4 +14,19 @@ T iroundup(T a, T b) {
   }
 }
 
+template <typename T>
+uint32_t log2int(T value) {
+  if constexpr (!std::is_integral_v<T>) {
+    static_assert(always_false_v<T>, "Must be integral type");
+  }
```
archibate, line 21: Why not just
```suggestion
  static_assert(std::is_integral_v<T>, "Must be integral type");
```
? Sorry, I'm not a C++ expert..

archibate:
> I actually have a question. Can `runtime.cpp` include a header whose implementation is in `.cpp`..?

I don't think so. The `runtime.bc` will only contain codes from `runtime.cpp`, right? So we should make those impl inline in `.h`.

k-ye:
> I don't think so. The runtime.bc will only contain codes from runtime.cpp, right? So we should make those impl inline in .h.

I also feel like so, which is why I made this a function template. But i guess it's easy to test out.

archibate:
You may merge with master again to fix CI :)

PR #1166, opened 2020-06-06
[OpenGL] support print 1 scalar in OpenGL kernels

### LINKED ISSUE #1158: [OpenGL] support print statements
I would like to add `print` support to the OpenGL backend.
It can help non-CUDA users debug their program on GPU. Also help me to debug the backend.

1. Add `used.print` to `class UsedFeature`, when `PrintStmt` used, `used.print = true`.
2. Add a message buffer to `struct runtime`.

Not sure how CUDA transport string message from GPU to CPU, but GLSL doesn't support `char` at all.
So we have to maintain a string table in CPU-side, and left the string index on GPU-side.
@yuanming-hu Could you provide more infromation about the CUDA implementation on `PrintStmt`?

**Additional comments**
@k-ye After I done with this, feel free to add `PrintStmt` support to Metal too.

archibate:
Related issue = close #1158

It was considered impossible for OpenGL to do this, and I found a way, to do it :)
Now the problem comes to: how to print strings? Sleepy now, cutmr.


taichi/backends/opengl/shaders/print.glsl.h
```diff
@@ -0,0 +1,13 @@
+// vim: ft=glsl
+// clang-format off
+#include "taichi/util/macros.h"
+STR(
+void _msg_push_i32(int x) {
+  int i = atomicAdd(_msg_count_, 1);
```
k-ye, line 6: There could be a buffer overflow? How about move the CPU side check to here, i.e. write `x` only if `i < MAX_PRINT_ENTRIES`

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -474,24 +496,24 @@ struct CompiledProgram::Impl {
...
+      auto rt_buf = (GLSLRuntime *)launcher->impl->runtime_ssbo->map();
+      auto msg_count = rt_buf->msg_count;
+      if (msg_count > MAX_PRINT_ENTRIES) {
+        TI_WARN(
+            "[glsl] Too much print within one kernel: {} > {}, cutting tail",
+            msg_count, MAX_PRINT_ENTRIES);
+        msg_count = MAX_PRINT_ENTRIES;
+      }
+      for (int i = 0; i < msg_count; i++) {
+        std::cout << rt_buf->msg_buf[i] << std::endl;
```
k-ye, line 509: How to distinguish `float` from `int`?

archibate: 1. An extra bitmap. Can be overhead-saving but hard to maintain, and can't extend to `f64` or `i64`.
2. An extra int array. Can cause large overhead copying message, but easy to extend, even to `string`.
3. An extra char array. No, OpenGL doesn't support `u8` at all.
WDYT?

k-ye: I like 2, since you mentioned `string` support as well. We can have a 32-bit integer for every data element to be printed. This 32-bit integer can encode both the type and the size.

For example, if you need to support `float`, `int` and `str`, then the type info needs 2bits, with the rest 30bits being the size

* For an integer `42`, type=`0`, size doesn't matter, since we know it's int32
* For a float `233.0`, type=`1`, size doesn't matter for the same reason.
* Flor a str `abcde`, type=`2`, size=`5`

archibate: > 1MB

Does 1MB data really printable on a terminal? I think 32KB is enough.

k-ye: I don't expect the buffer to be fully filled for most of the time, but this is really a number I came up randomly. Please feel free to choose a reasonably large size (512bytes seems too small to me)

taichi/backends/opengl/shaders/print.glsl.h
```diff
@@ -0,0 +1,32 @@
...
+STR(
+int _msg_allocate_slot() {
+  return atomicAdd(_msg_count_, 1);
+}
+
+void _msg_set(int msgid, int contid, int type, int value) {
+  int base = msgid << 5; // MSG_SIZE = (1 << 5)
+  _mesg_i32_[base + contid] = value;
+  _mesg_i32_[base + 30] |= (type & 1) << contid;
+  _mesg_i32_[base + 31] |= (type >> 1) << contid;
```
archibate, line 13: `base + 30` for `type_bitmap_low`.
`base + 31` for `type_bitmap_high`.
combine together, obtains `type`.
I know it looks trick here, but it enable us to print 29 contents, instead of just 16 contents. And yet I don't find a better solution to utilize memory.

taichi/backends/opengl/shaders/print.glsl.h
```diff
@@ -0,0 +1,32 @@
...
+}
+
+void _msg_set_end(int msgid, int contid) {
+  int base = msgid << 5;
+  _mesg_i32_[base + 29] = contid;
```
archibate, line 18: `base + 29` for `contents.size()`.

taichi/backends/opengl/runtime.h
```diff
@@ -0,0 +1,27 @@
...
+};
+)
+
+// clang-format on
+#else
+
+struct GLSLRuntime {
+  int rand_state;
+  int msg_count;
+  int msg_buf[MAX_MESSAGES * MSG_SIZE];
```
archibate, line 24: Mark: may use `struct GLSLMsgEntry msg_buf[MAX_MESSAGES]` here? Although it doesn't make sence in GLSL, but we can still make the reviewer relief by making C++ version clearer.

taichi/backends/opengl/shaders/print.glsl.h
archibate: Update: I improved readability in C++ version of `GLSLRuntime`, so although limitations make me can't improve GLSL version, reviewers still can get my point in the C++ version.

taichi/backends/opengl/runtime.h
```diff
@@ -0,0 +1,51 @@
...
+    int type = (type_bm_lo >> i) & 1;
+    type |= ((type_bm_hi >> i) & 1) << 1;
+    return type;
+  }
+} __attribute__((packed));
+
+struct GLSLRuntime {
+  int rand_state;
+  int msg_count;
+  GLSLMsgEntry msg_buf[MAX_MESSAGES * MSG_SIZE];
```
k-ye, line 48: I'm kind of confused of why the buffer size is specified like this, why not simply use `MAX_MESSAGES`?

archibate: Typo, sorry!

PR #1170, opened 2020-06-07
[Opt] Improve strength reduction optimization for "pow"

xumingkuan:
Related issue = #1141 #944

Related PR = #1065

WANT_LGTM = any


xumingkuan:
I think for any `int32 x`, `x == (float64)x`. Or shall we optimize (say) `a ** (2 + eps)` to `a * a` when `fast_math=True`?

k-ye:
> I think for any int32 x, x == (float64)x.

I see, glad to learn about this..

> Or shall we optimize (say) a ** (2 + eps) to a * a when fast_math=True?

I'll leave this up to you :)

PR #1161, opened 2020-06-06
[mac] [misc] Fix Travis CI: no <filesystem>, use <unistd.h> instead

archibate:
Related issue = https://github.com/taichi-dev/taichi/pull/1144#issuecomment-639999779

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

PR #1173, opened 2020-06-07
[misc] Simplify the usage of <filesystem>

k-ye:
I think we can just choose to use `<experimental/filesystem>` when the system version is too old.

Related issue = #1161

### DIFF
--- a/taichi/llvm/llvm_context.cpp
+++ b/taichi/llvm/llvm_context.cpp
@@ -38,12 +38,20 @@
 #include "taichi/jit/jit_session.h"
 #include "taichi/common/task.h"
 #include "taichi/util/environ_config.h"
-#ifdef _WIN32
-// Travis CI seems doesn't support <filesystem>...
-#include <filesystem>
+
+#ifdef TI_PLATFORM_OSX
+#include "Availability.h"
+#ifndef __MAC_10_15
+// Travis CI is not yet on Mac 10.15...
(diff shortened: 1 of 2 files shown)


taichi/llvm/llvm_context.cpp
```diff
@@ -38,12 +38,20 @@
...
+
+#ifdef TI_PLATFORM_OSX
+#include "Availability.h"
+#ifndef __MAC_10_15
+// Travis CI is not yet on Mac 10.15...
+#define _TI_USE_EXPERIMENTAL_FILESYSTEM 1
+#endif  // __MAC_10_15
+#endif  // TI_PLATFORM_OSX
+
+#ifdef _TI_USE_EXPERIMENTAL_FILESYSTEM
```
archibate, line 50: Can't we just always use `<experimental/filesystem>`? It works on new machines too.

k-ye: Thanks. <experimental/filesystem> on my local env generated warnings, and it didn't fix the problem anyway. I'm now trying to just roll up XCode in Travis CI and see if that fixes the error. Will let you know once this is ready.

k-ye:
> Will this cause people with xcode10.3 fail to compile?

Yeah, I guess we still need to support `Mac 10.14` users. Maybe we should leave this as is for now.

PR #1187, opened 2020-06-08
[test] Let pytest-rerunfailures pytest-xdist to be hard dependences

archibate:
Related issue = (reported by GAMES 201 students)


rexwangcc:
Btw, maybe we could remove the `misc/empty_pytest.py` now since we don't need to use them for plugin probes?

PR #1184, opened 2020-06-08
[bug] [lang] Fix an issue which breaks the binary operations between global Vectors in Taichi-scope 

rexwangcc:
Related issue = #... (if any)

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

PR #1188, opened 2020-06-08
[Lang] [refactor] Support Python-scope scalar functions / matrix operations, e.g. ti.sqrt(2)

archibate:
Related issue = #1167, replace #1184, address issues in #1051

This PR aims to refactor/replace the current python-scope operator system in #1051 which heavy rely on `Matrix.to_numpy`.

With this PR, you can:
```py
>>> ti.sqrt(2)
1.4142135623730951
```
You see `ti.sqrt` behave exactly same as `math.sqrt` for python numbers, just like the `np.sqrt(2)` would does.

You may also do matrix / vector operations in Python-scope:
```py
>>> ti.Vector([3, 4]).normalized()
[0.6 0.8]
>>> ti.cos(ti.Vector([3, 4]))
[-0.9899925  -0.65364362]
```

Not sure how this affects constant folding process, though... I guess it actually helps do CFD on Python-side and reduce compile-time in launching JIT kernels?

I think we can release a hot-fix for this in ~1 day, given that GAMES 201 students are willing to operate `ti.Vector` in Python-scope.


archibate:
@codecov Thank you! With your kind reminder, I managed to polish my test over and over again, and fixed many potential bugs! :heart:

python/taichi/lang/ops.py
```diff
@@ -256,84 +274,209 @@ def pow(self, power):
...
 
 
 @binary
 def raw_div(a, b):
-    return Expr(taichi_lang_core.expr_div(a.ptr, b.ptr), tb=stack_info())
+    if is_taichi_expr(a) or is_taichi_expr(b):
+        a, b = wrap_if_not_expr(a), wrap_if_not_expr(b)
+        return Expr(taichi_lang_core.expr_div(a.ptr, b.ptr), tb=stack_info())
+    else:
+        return a // b  # TODO: Is this correct???
```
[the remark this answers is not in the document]
archibate: `int / int` is floordiv and `float / float` is truediv in C++.. Maybe we should:
```suggestion
        return a // b if isinstance(a, int) and isinstance(b, int) else a / b
```
?

python/taichi/lang/ops.py
```diff
@@ -256,84 +274,209 @@ def pow(self, power):
...
 
 
 @binary
 def raw_mod(a, b):
-    return Expr(taichi_lang_core.expr_mod(a.ptr, b.ptr), tb=stack_info())
+    if is_taichi_expr(a) or is_taichi_expr(b):
+        a, b = wrap_if_not_expr(a), wrap_if_not_expr(b)
+        return Expr(taichi_lang_core.expr_mod(a.ptr, b.ptr), tb=stack_info())
+    else:
+        return a - b * int(float(a) / b)
```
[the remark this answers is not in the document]
archibate: Your recall is correct :) See
https://github.com/taichi-dev/taichi/blob/18e17ffd709eddaad5a37cd8f183b0dd690fac69/tests/python/test_mod.py#L38-L39

python/taichi/lang/matrix.py
```diff
@@ -339,35 +321,55 @@ def z(self, value):
...
+                    ret.entries[i * self.mat.m + j] = self.mat(i,
+                                                               j)[self.index]
+            return ret
+
+    # host access & python scope operation
     @python_scope
-    def __getitem__(self, index):
-        return Matrix.Proxy(self, index)
+    def __getitem__(self, indices):
+        if self.is_global():
```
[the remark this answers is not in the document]
archibate: Yes, when `is_global() == False`, it can be either matrix of constants or **matrix of local-variables**.

archibate:
@yuanming-hu Hello? Is this to be merged before v0.6.10? This PR is error-prone and needs your decision.

Best,
Yubin Peng

python/taichi/lang/matrix.py
```diff
@@ -263,45 +247,40 @@ def w(self):
         else:
             return self[3]
 
+    # since Taichi-scope use v.x.assign() instead
```
xumingkuan, line 250: Which function will it invoke if I call `v.x = 233` in Taichi-scope?

python/taichi/lang/matrix.py
```diff
@@ -539,51 +529,47 @@ def make_grad(self):
...
-    @taichi_scope
     def all(self):
         import taichi as ti
-        ret = self.entries[0] != ti.expr_init(0)
+        ret = ti.cmp_ne(self.entries[0], 0)
         for i in range(1, len(self.entries)):
-            ret = ret + (self.entries[i] != ti.expr_init(0))
-        return -(ret == ti.expr_init(-len(self.entries)))
+            ret = ret + ti.cmp_ne(self.entries[i], 0)
+        return ti.cmp_eq(ret, -len(self.entries))
```
xumingkuan, line 572: Does it return -1 before this PR but return 1 after this PR when the result is `True`?

python/taichi/lang/matrix.py
```diff
@@ -138,52 +138,39 @@ def is_global(self):
...
-        ret = Matrix(self.n, other.m)
+        ret = Matrix.empty(self.n, other.m)
         for i in range(self.n):
             for j in range(other.m):
-                ret(i, j).assign(self(i, 0) * other(0, j))
+                ret_ij = ret.linearize_entry_id(i, j)
+                ret.entries[ret_ij] = self(i, 0) * other(0, j)
                 for k in range(1, other.n):
-                    ret(i, j).assign(ret(i, j) + self(i, k) * other(k, j))
+                    ret.entries[ret_ij] = ret(i, j) + self(i, k) * other(k, j)
```
yuanming-hu, line 173: Will this lead to code bloat? We want to store the results to a local variable instead of in a set of very complex `Expr`.

python/taichi/lang/matrix.py
archibate: OIC, should we make expr_cmp_eq return -1 one True all the time?

python/taichi/lang/matrix.py
archibate: It's first `v.x.getter` then `.assign()`, see `python/lang/transformer.py`.
```py
In [2]: @ti.kernel 
   ...: def func(x: ti.f32, y: ti.f32): 
   ...:     v = ti.Vector([x, y]) 
   ...:     v.x = 233  # translated into v.x.assign(233)
   ...:     print(v) 
   ...:                                                                                                                                                                                      

In [3]: func(666, 777)                                                                                                                                                                       
[233.000000, 777.000000]
```

python/taichi/lang/matrix.py
```diff
@@ -138,52 +138,39 @@ def is_global(self):
...
-        if self.is_pyconstant() and in_python_scope():
-            return self.make_from_numpy(self.to_numpy() @ other.to_numpy())
-
         assert self.m == other.n, f"Dimension mismatch between shapes ({self.n}, {self.m}), ({other.n}, {other.m})"
-        ret = Matrix(self.n, other.m)
+        ret = Matrix.empty(self.n, other.m)
         for i in range(self.n):
             for j in range(other.m):
-                ret(i, j).assign(self(i, 0) * other(0, j))
+                ret_ij = ret.linearize_entry_id(i, j)
```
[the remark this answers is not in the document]
archibate: Maybe it's just faster...

python/taichi/lang/matrix.py
xumingkuan: We should return 1 on True after we have `ti.u1`. For now, I think we return -1 for efficiency reasons, as we can optimize `-1 and a` to `a`.

archibate:
Btw, I noticed that `expr_cmp_eq(x, y)` returns -1 while `x == y` returns 1 thanks to `ti.chain_compare`? Is that intentional to convert -1 in backend to 1 in frontend?

xumingkuan:
> Btw, I noticed that `expr_cmp_eq(x, y)` returns -1 while `x == y` returns 1 thanks to `ti.chain_compare`? Is that intentional to convert -1 in backend to 1 in frontend?

Yeah I implemented `ti.chain_compare` to be 1 because I thought `ti.u1` will be implemented soon... Maybe we should change that to -1 (just change `True` to `-1` here) to make it more consistent for now?
https://github.com/taichi-dev/taichi/blob/6e78dcf75888f850786c71373e7dc6ad01912e3c/python/taichi/lang/impl.py#L87

archibate:
I'm fine with turning all `True` to `-1`, but notice that `ti.is_active` return 1 on `True`... so user will obtain an error if `ti.is_active(blk, 0) and x > y` since it's `1 & -1`, LOL.

python/taichi/lang/matrix.py
xumingkuan: Let's simply don't change the behavior of this function in this PR. Although we're not consistent on the value of True now, this PR already contains a large changeset.

archibate:
Hello? I feel more want to merge this before v0.6.11.

archibate:
Could you review on this asap? My knowledge on this PR is getting rust, I won't be able to make any change after ~2 weeks.

tests/python/test_matrix.py
```diff
@@ -32,66 +41,109 @@ def test_python_scope_matrix_operations():
...
+    print(v.normalized())
+    print(v.cross(w))
+    w.y = v.x * w[0]
+    r.x = r.y
+    r.y = r.z
+    r.z = r.w
+    r.w = r.x
+    print(w)
+    s[0, 1] = 2
+    print(s[0, 1])
```
yuanming-hu, line 94: Apart from should we also `assert` their results?

If the code is intentionally designed this way just to test if `print` works, please leave a comment.

python/taichi/lang/ops.py
```diff
@@ -107,83 +108,124 @@ def sqr(obj):
...
+    if is_taichi_expr(a):
+        return Expr(taichi_lang_core.expr_sin(a.ptr), tb=stack_info())
+    else:
+        return math.sin(a)
 
 
 @unary
-def asin(expr):
-    return Expr(taichi_lang_core.expr_asin(expr.ptr), tb=stack_info())
+def cos(a):
```
yuanming-hu, line 127: Thanks for modifying everything here. The changes have too much repetition and are kind of error-prone. 

Instead of repeating yourself 20+ times in a single file, we should probably consider extracting the common parts.  Something like
```python
def unary_operation(taichi_op, python_op):
    ....

@unary
def cos(a):
    return unary_operation(taichi_lang_core.expr_cos, math.cos)
```

The same for binary operators.

tests/python/test_matrix.py
archibate: Yes, I added these just to make sure Python-scope functions don't crash Taichi, and we're good enough if so. A good `assert`'ed test is really hard to be systematically (like I did in `test_element_wise`) can be added iapr?

python/taichi/lang/ops.py
archibate: I used `_unary_operation` to prevent name issue against the evil `from . import *` pattern.
Btw, is that helpful to make a code generator for this situation to anti-error (next time)?

yuanming-hu: I actually think simply use a function or a better decorator will do the job :-)  A code generator is too complex for the case here.

tests/python/test_matrix.py
yuanming-hu: I see. I understand `print` is hard to test systematically for now. Given the title is `test_constant_matrices`, should we test if the return values are correct using `asserts`?

python/taichi/lang/ops.py
yuanming-hu: Do you have local commits that are not pushed? I don't see the `_unary_operation` function in the latest code.

python/taichi/lang/matrix.py
```diff
@@ -801,14 +786,13 @@ def cross(self, b):
...
 
-    @taichi_scope
     def outer_product(self, b):
         assert self.m == 1
         assert b.m == 1
         c = Matrix(self.n, b.n)
         for i in range(self.n):
             for j in range(b.n):
-                c(i, j).assign(self(i) * b(j))
+                c.set_entry(i, j, self(i) * b(j))
```
yuanming-hu, line 795: I think this changes the behavior in Taichi-scope. We probably want `assign` there.

python/taichi/lang/matrix.py
```diff
@@ -138,52 +138,39 @@ def is_global(self):
...
-        ret = Matrix(self.n, other.m)
+        ret = Matrix.empty(self.n, other.m)
         for i in range(self.n):
             for j in range(other.m):
-                ret(i, j).assign(self(i, 0) * other(0, j))
+                acc = self(i, 0) * other(0, j)
                 for k in range(1, other.n):
-                    ret(i, j).assign(ret(i, j) + self(i, k) * other(k, j))
+                    acc = acc + self(i, k) * other(k, j)
+                ret.set_entry(i, j, acc)
```
yuanming-hu, line 173: Is `set_entry` same as `assign` in Taichi-scope?

archibate: No, thank for pointing out, should we make `set_entry` call `assign` in Taichi-scope and `=` in Python-scope? Not sure this is a good design, but it's the most strightforward makes matmul work in Python-scope, which is the PR aim.

python/taichi/lang/ops.py
archibate: Oh sorry! Didn't push... too much PRs and I forget what is what.

python/taichi/lang/matrix.py
yuanming-hu: Yeah I think that's a good idea for now. At least it makes me feel safer since we are not changing Taichi-space behavior.

archibate: Thank for approving my idea. BTW I just come up with a more crazy idea: all exprs are not-assignable, just like LLVM IR does, which permanantly prevent such issues when doing `a = a.transpose()` or more complicated assignment dependency issues (iapr of course).

python/taichi/lang/matrix.py
```diff
@@ -772,19 +763,24 @@ def empty(n, m):
         mat = ti.Matrix([[None] * m for _ in range(n)])
         return mat
 
+    @staticmethod
+    def new(n, m):
+        if impl.inside_kernel():
+            return Matrix(n, m)
+        else:
+            return Matrix.empty(n, m)
```
rexwangcc, line 771: 🤔 The use cases of `new` looked more like this would be a `classmethod` to me, but I'm not 100% sure. e.g.

```suggestion
    @classmethod
    def new(cls, n, m):
        if impl.inside_kernel():
            return cls(n, m)
        else:
            return cls.empty(n, m)
```

python/taichi/lang/ops.py
yuanming-hu: Cool, now everything looks much better. Thanks.

python/taichi/lang/matrix.py
yuanming-hu: We still need to have mutable local variables, so I'm afraid that won't work very well.

tests/python/test_matrix.py
archibate: Added now, although not covered all functions, but can be a good walk around before we come to a systematic test.

python/taichi/lang/matrix.py
archibate: Thank for teaching me Python knowledge! I never know this `@classmethod` decorator before, it looks so useful, thanks! @yuanming-hu Will we make other initializers like `Matrix.cols/rows/var` a class method too? Or `TaichiClass.var` should be?

k-ye:
I did a bisect and found that this PR has changed the rendering result of `cornell_box.py`. Before this one, it was like:

Screen Shot 2020-06-25 at 21 23 04

With this one, it becamse
Screen Shot 2020-06-25 at 21 21 05

Note that olive shape white shade pointed by the arrow. Can you think of any arithmetic or behavior change that could have resulted in this difference?

k-ye:
Seems like the problem is around https://github.com/taichi-dev/taichi/blob/13462ff09560de75fc780de3ee6b0519086ee3b9/python/taichi/lang/ops.py#L213-L215

archibate:
Since @xumingkuan has already done C++ side pow optimization, let's deprecate raw_pow then.

k-ye:
I has dug a bit more into the problem. I think the real culprit is an incorrect propagation of `ti.random()`. See #1329 .

k-ye:
`mpm_lagrangian_forces.py` is also broken.

PR #1194, opened 2020-06-09
[IR] Move "lower_access" after "offload"

xumingkuan:
Related PR = #1180

benchmark20200608_4

In order to achieve similar performance, I added some optimization passes.


xumingkuan:
Removing `Simplify II` here won't affect the resulting number of statements, but it looks weird to me not having `full_simplify` before `offload`.

PR #1201, opened 2020-06-10
[Lang] Export .ply file

Eydcao:
Related issue = #1016 

Hi, with the help of @TroyZhai and his [repository](https://github.com/TroyZhai/np2ply) we are working to build a pipeline for exporting results into `.ply` files so that people can use Houdini for rendering and viewing. A simple todo list is below

- a note for myself to learn this [addon](https://github.com/neverhood311/Stop-motion-OBJ) in blender for time series play.

- [x] A note on windows dev install
- [x] Resolve the conflict about win dev install with #1202
- [x] Import Zhai's repo to taichi and successfully build an interface to export a single file
- [x] Update with Zhai's update concerning face's piece, color's alpha, etc.

- [x] Houdini support single file
- [x] Houdini support time series
- [x] Pipeline for Houdini time series generation
- [x] Tutorial for ti to Houdini

- [x] Blender support single file
- [x] Blender support time series (with the help of stop-motion-object)
- [x] Pipeline for Blender time series generation
- [x] Tutorial for ti to Blender

Please feel free to give more suggestions any time~


examples/export_ply.py
```diff
@@ -0,0 +1,23 @@
...
+x = x.to_numpy()
+writer = ti.PLYWriter(num_vertices=n_particles)
+writer.add_vertex_pos(x[:, 0], x[:, 1], x[:, 2])
+
+# export binary ply file
+writer.export("example.ply")
+os.remove("example.ply")
+# export ascii ply file
+writer.export_ascii("example_ascii.ply")
+os.remove("example_ascii.ply")
```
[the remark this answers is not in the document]
Eydcao: sure~~~I was just trying to keep repo clean. Notice this is far away complete, so the script may change many times before everything settles down

examples/export_ply.py
```diff
@@ -0,0 +1,37 @@
...
+import os
+
+ti.init(arch=ti.cpu)
+
+x = ti.Vector(3, dt=ti.f32, shape=(10, 10, 10))
+color = ti.var(ti.f32, shape=(10, 10, 10, 3))
+
+
+@ti.kernel
+def init():
```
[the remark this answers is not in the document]
Eydcao: Got it! `place` should be better.

Eydcao:
> Hi @Eydcao, I'd like to add a few more to-dos to make this feature more practical.
> 
> * [x]   Support time sequences in Houdini
> 
> I think this is already supported if we name the ply sequence as xxx_0001, xxx_0002, ... we just need to make a pipeline to generate those sequences, so
> 
> * [ ]   Generate time sequences
> 
> For most people, it's probably not that straightforward to understand all the channels. 'pos', 'normal', 'rgba' are probably fine, but 'id', 'piece' (see updates in my repo) and custom channels will require some explanations or I could foresee that they can get used very rarely. Therefore,
> 
> * [ ]   Make a nice tutorial to document the pipeline from taichi to Houdini
> 
> Lastly, there could be a lot of requests to have Blender supported, so
> 
> * [ ]  Test the exported files in Blender
> * [ ]  Test time sequences in Blender
> * [ ]   Make a nice tutorial to document the pipeline from taichi to Blender
> 
> Cheers, guys

Hi Zhai, thanks for your detailed suggestion!

Concerning the detailed pipeline for generating time series output. I would suggest using a similar manner as in many CAE software that user define deltaFrame# (like deltaTime). Any other thoughts (just output every frame)?

The reason why I wrapped your functions previously (xyz to vector) is that seems convenient to pass a ti.matrix to the writer, but sure I will keep your interface for those who appreciate more control over details.

All the todos have been added. Thanks guys.

Yadi

python/taichi/misc/np2ply.py
```diff
@@ -128,6 +125,19 @@ def add_vertex_rgba(self, r: np.array, g: np.array, b: np.array, a: np.array):
         self.add_vertex_channel("blue", "float", b)
         self.add_vertex_channel("Alpha", "float", a)
 
+    # pass ti vector/matrix tensor directly
+    def add_vertex_rgba(self, rgba):
+        if not isinstance(rgba, np.ndarray):
```
TroyZhai, line 130: could we add an assert here to specify what exactly type we're expecting 'rgba' to be?

python/taichi/misc/np2ply.py
```diff
@@ -128,6 +125,19 @@ def add_vertex_rgba(self, r: np.array, g: np.array, b: np.array, a: np.array):
         self.add_vertex_channel("blue", "float", b)
         self.add_vertex_channel("Alpha", "float", a)
 
+    # pass ti vector/matrix tensor directly
+    def add_vertex_rgba(self, rgba):
+        if not isinstance(rgba, np.ndarray):
+            rgba = rgba.to_numpy()
+        channels = rgba.shape[rgba.ndim-1]
+        assert channels == 4, "The dimension for rgba should be 4"
+        n = rgba.size // channels
```
TroyZhai, line 134: I think self.num_vertex is better here.

python/taichi/misc/np2ply.py
```diff
@@ -128,6 +125,19 @@ def add_vertex_rgba(self, r: np.array, g: np.array, b: np.array, a: np.array):
...
         self.add_vertex_channel("Alpha", "float", a)
 
+    # pass ti vector/matrix tensor directly
+    def add_vertex_rgba(self, rgba):
+        if not isinstance(rgba, np.ndarray):
+            rgba = rgba.to_numpy()
+        channels = rgba.shape[rgba.ndim-1]
+        assert channels == 4, "The dimension for rgba should be 4"
+        n = rgba.size // channels
+        rgba = np.reshape(rgba, (n, channels))
```
TroyZhai, line 135: If we're going to reshape rgba, could we do that at first, at least before we check its shape in line 132?

python/taichi/misc/np2ply.py
```diff
@@ -236,3 +246,21 @@ def export_ascii(self, path):
                 for j in range(self.num_face_channels):
                     f.write(str(self.face_data[j][i]) + " ")
                 f.write("\n")
+
+    def export_for_time_series_ascii(self, series_num: int, path):
```
TroyZhai, line 250: Can we make sure path here is a string?

TroyZhai: Also could you make the name shorter considering this could be the main entrance of this feature? Something like 'export_frame' and 'export_frame_ascii' maybe?

python/taichi/misc/np2ply.py
```diff
@@ -99,20 +99,17 @@ def add_vertex_normal(self, normal):
         self.add_vertex_channel("ny", "float", normal[:, 1])
         self.add_vertex_channel("nz", "float", normal[:, 2])
 
-    def add_vertex_color(self, r: np.array, g: np.array, b: np.array):
+    def add_vertex_rgb(self, r: np.array, g: np.array, b: np.array):
```
TroyZhai, line 102: Can we keep a 'add_vertex_color' method in this class? It can be calling add_vertex_rgb directly, but it's really intuitive to use. Thanks

python/taichi/misc/np2ply.py
Eydcao: Sure, I would suggest `rgba` should either assert `np.ndarray` or 'ti.vector/Matrix'. Any other type would you suggest?

python/taichi/misc/np2ply.py
Eydcao: This makes sense that `n` must equal to `self.num_vertex`. Nonetheless, I would prefer to calculate `n` then add an assertion here. Because people may pass a `ti.vector` tensor with the wrong size without realizing this, so we should inform them here.

What would you suggest?

python/taichi/misc/np2ply.py
Eydcao: To me I would insist on checking the last dimension first, just to make sure the tensor people passed at least has the same dimension as `rgba` required. Otherwise, if we reshape first i.e. a `100*100*2` array into `5000*4` one, and by chance `vertex_num = 5000` then an unnoticeable runtime bug is there.

python/taichi/misc/np2ply.py
Eydcao: > Can we make sure path here is a string?

Right!

Eydcao: > Also could you make the name shorter considering this could be the main entrance of this feature? Something like 'export_frame' and 'export_frame_ascii' maybe?

Haha, I apologize for some of my previous experiences from those industrial standards. (like `exportForThisTimeInASCII`).

python/taichi/misc/np2ply.py
Eydcao: Sure, what about this way, this will be easier for user to distinct the difference input list. But I can also do it the old way (using overload).
```
def add_vertex_rgb(self, r, g, b)

def add_vertex_color(self, color: np.ndarray/ti.vector)
```

Similarly we can define `add_vertex_xyz` and `add_vertex_pos`.

TroyZhai: I think from ti to np should be fairly easy so it's should be alright if we stick everything with np. Overloading sounds great but can it be done gracefully in python?

python/taichi/misc/np2ply.py
TroyZhai: Yeah, I was worries about the same thing, I think that's exactly we don't want to calculate n before we do the assert

python/taichi/misc/np2ply.py
TroyZhai: Well I have different opinions. It doesn't matter we pass 100*100*2 or 5000*4, it should work in both situations, or the reshape loses its meaning.

python/taichi/misc/np2ply.py
TroyZhai: Hahaha, in python everything should be cool

python/taichi/misc/np2ply.py
Eydcao: haha I assume this problem is the same as that in the conversation above. For a professional programmer, even the function needs i.e. `vector<vector4D>`, she/he could store it whatever the shape outside, as long as its convenient.

For a rookie programmer like myself I would appreciate less freedom in this input list just to prevent errors T_T. Yeah, I think I will adopt your suggestion here by assuming the programmer has handled the size well.

python/taichi/misc/np2ply.py
Eydcao: I think our disagreement here is somewhat the same as the next conversation, I will address it there. Thanks

python/taichi/misc/np2ply.py
Eydcao: Haha, conversion between ti and np was not my point. I did not well illustrate the term 'overloading' previously, its actually because in the current commit I have both `add_vertex_pos(x,y,z)` and `add_vertex_pos(pos: vector)` which is the overload. Though it works well, it can be confusing for people and pylint.

So following your previous suggestion about keeping both `add_vertex_color` and `add_vertex_rgb`, I was suggesting we can define them as below to both avoid overloading and distinguish the input list, meanwhile keeping both.

```
def add_vertex_rgb(self, r, g, b)

def add_vertex_color(self, colorVector)

def add_vertex_xyz(self, x, y, z)

def add_vertex_pos(self, posVector)
```

Sorry for the confusion :>

examples/export_ply.py
```diff
@@ -0,0 +1,44 @@
...
+pos = ti.Vector(3, dt=ti.f32, shape=(10, 10, 10))
+rgba = ti.var(ti.f32, shape=(10, 10, 10, 4))
+
+
+@ti.kernel
+def place_pos():
+    for i, j, k in pos:
+        pos[i, j, k][0] = 0.1*i
+        pos[i, j, k][1] = 0.1*j
+        pos[i, j, k][2] = 0.1*k
```
yuanming-hu, line 17:
```suggestion
        pos[i, j, k] = 0.1* ti.Vector([i, j, k])
```

examples/export_ply.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@ti.kernel
+def move_particles():
+    for i, j, k in pos:
+        pos[i, j, k][0] = pos[i, j, k][0]+0.01
+        pos[i, j, k][1] = pos[i, j, k][1]+0.01
+        pos[i, j, k][2] = pos[i, j, k][2]+0.01
```
yuanming-hu, line 25:
```suggestion
        pos[i, j, k] += ti.Vector([0.1, 0.1, 0.1])
```

examples/export_ply.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@ti.kernel
+def fill_rgba():
+    for I in ti.grouped(rgba):
+        rgba[I] = ti.random()
```
yuanming-hu, line 31:
```suggestion
    for i, j, k, l in rgba:
        rgba[i, j, k, l] = ti.random()
```

Just in case the reader doesn't understand `ti.grouped`.

python/taichi/misc/np2ply.py
TroyZhai: To be honest I still prefer only have setters for pos, color and rgba, and have all channels individually passed in. It's the least confusing way, I would suggest to keep it simple and less over designed 😁

python/taichi/misc/np2ply.py
TroyZhai: I think in Python there is a unspoken rule that if something will get reshaped eventually, its input dimension doesn't matter any more. We could stick to this convention here

examples/export_ply.py
Eydcao: Those are pretty nice improvements! I didn't understand grouped as well.:>

python/taichi/misc/np2ply.py
Eydcao: > unspoken rule

That frightens me a bit T_T, okay I will leave this conversation unresolved just in case related improvement can be done in the future. For now, I will keep it a little strict on the input requirement just in case for errors.

python/taichi/misc/np2ply.py
Eydcao: We both got a point and timeline is urgent, we still have tutorials to build, so for now I will try to do it my fashion and try to make it clear on the documentation
@__@ T__T although I have applied many suggestions, it makes me feel a little down to not accept all of other's kind suggestions. But time is limited, I will leave this conversation open if we got feedback from the students later.

python/taichi/misc/np2ply.py
TroyZhai: The issue with this approach is that if the user passes something else that has exactly 4 channels, it will only get an error from reshape, and that is very discouraging for a user. For most, people want a transparent yet robust tool, not delving into our tool and dealing with all the unexpected errors. We should at least ensure that our code can handle all the unexpected situations without any uncaptured error message directly from runtime.

python/taichi/misc/np2ply.py
TroyZhai: Well, no offenses here but the fact is that no one wills to remember what is the difference between color and rgb, pos and xyz. **It was not designed that way and shouldn't be like that either**. We have to consider the user when we do something like this. You can say, ok for the first version let's cut the corners and leave some confusions that we promise to fix when we got feedback. However, when we fix that, there'll be more **silent dissatisfaction** from the users that have got used to our tangled logic. Therefore it's really important to have a design before having it done, and that's is what I did last weekend. I agree that we could make this feature more convenient by all your suggestions and revisions, but sorry that's exactly why I tried to avoid in my code - forget about convenience in the first version, make it robust, **do whatever it means to do without discouragement**. Convenience can be compromised and delayed to the future, but not the above qualities.

Eydcao: Got your point, I will adopt your suggestion concerning the user design issue and comment out the compact/reloaded ones.

Thanks for all the patience :>

python/taichi/misc/np2ply.py
Eydcao: > ...if the user passes something else that has exactly 4 channels, it will only get an error from reshape...

Although this will be addressed as in the screenshot, I would like to adopt your final suggestion about which api to keep that explicitly indicating all input channels by the user is the most robust.

image

I will comment out those "compact version api" and add a comment note in the code, just in case they will be reactive in the future.

Thanks for this long patience, I learned a lot :>

docs/export_results.rst
```diff
@@ -173,3 +173,166 @@ Install ``ffmpeg`` on OS X
...
+        # create a PLYWriter
+        writer = ti.PLYWriter(num_vertices=num_vertices)
+        writer.add_vertex_pos(np_pos[:, 0], np_pos[:, 1], np_pos[:, 2])
+        writer.add_vertex_rgba(
+            np_rgba[:, 0], np_rgba[:, 1], np_rgba[:, 2], np_rgba[:, 3])
+        writer.export_frame_ascii(frame, series_prefix)
+
+After running the code above, you will find the output sequence of ``ply`` files in the current running folder. Next, we will break up the process of using ``ti.PLYWriter`` into 4 steps and show some examples.
+
+- Setup ``ti.PLYWriter``
```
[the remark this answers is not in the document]
Eydcao: I have never used example -p before, will give a try and see how to implement properly.

Thanks,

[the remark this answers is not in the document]
yuanming-hu: Or you can just include a runnable `export_mesh.py` in the `examples` folder for now. No need to update the doc if you don't have `ti example -p`. Let's try to merge this as soon as possible :-)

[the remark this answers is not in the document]
Eydcao: Sorry for the delayed reply I was touching on sth else. Sure I am adding the second part to a run able ``export_mesh.py``. Just a sec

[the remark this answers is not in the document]
Eydcao: Hi Yuanming! there is some bug in conversion between normal data into binary form in the current version. This is concerning the face ids only (so the moving cube example is okay) and was not covered in my previous tests. I will try to fix asap but as respect to timeline, I have disabled the binary output format in our doc and example.

If urgent please merge and I will open an issue, so sorry for not trying all possible channels.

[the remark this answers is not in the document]
Eydcao: image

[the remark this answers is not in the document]
yuanming-hu: No problem. In the worse case, we can ask the students to use particles only for now. No rush.;_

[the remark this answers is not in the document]
Eydcao: Not that bad, ascii is all good T_T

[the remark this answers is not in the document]
Eydcao: Solved!

PR #1212, opened 2020-06-10
[CLI] Add TI_ADVANCED_OPTIMIZATION config env var

archibate:
Related issue = #... (if any)

---

Btw. is that possible to move `advanced_optimization` to `CompilerConfig`?

### DIFF
--- a/docs/global_settings.rst
+++ b/docs/global_settings.rst
@@ -12,7 +12,7 @@ Backends
 Compilation
 ***********
 
-- Disable advanced optimization to save compile time & possible erros: ``ti.core.toggle_advanced_optimization(False)``.
+- Disable advanced optimization to save compile time & possible erros: ``ti.core.toggle_advanced_optimization(False)`` or ``export TI_ADVANCED_OPTIMIZATION=0``.
 - To print intermediate IR generated: ``export TI_PRINT_IR=1`` or ``ti.init(print_ir=True)``.
 - To print preprocessed Python code: ``export TI_PRINT_PREPROCESSED=1`` or ``ti.init(print_preprocessed=True)``..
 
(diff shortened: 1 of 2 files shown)


archibate:
> > IMO the current name is good enough :)
> > @xumingkuan seems the obstacle of moving `advanced_optimization` to `CompilerConfig` is that we don't pass a `CompilerConfig` var to all transforms/passes? I understand the difficulty, but I think ultimately we should make all transfroms a class method of `Compiler`, who has a member variable of `CompilerConfig`?
> 
> Please checkout #988 and https://github.com/taichi-dev/taichi/pull/1058/files#diff-60ee40c377644d6a6a4ff0cedad49110R182: we can obtain the `CompileConfig` in all passes by `root->get_kernel()->program.config` now.

So, should I refactor all `advanced_optimization` to `root->get_kernel()->program.config.advanced_optimization` in this PR?

xumingkuan:
> > > IMO the current name is good enough :)
> > > @xumingkuan seems the obstacle of moving `advanced_optimization` to `CompilerConfig` is that we don't pass a `CompilerConfig` var to all transforms/passes? I understand the difficulty, but I think ultimately we should make all transfroms a class method of `Compiler`, who has a member variable of `CompilerConfig`?
> > 
> > 
> > Please checkout #988 and https://github.com/taichi-dev/taichi/pull/1058/files#diff-60ee40c377644d6a6a4ff0cedad49110R182: we can obtain the `CompileConfig` in all passes by `root->get_kernel()->program.config` now.
> 
> So, should I refactor all `advanced_optimization` to `root->get_kernel()->program.config.advanced_optimization` in this PR?

I think that should be in another PR.

archibate:
Great! Let's merge this then,

PR #1226, opened 2020-06-12
[Opt] Add binary ops optimization pass

TH3CHARLie:
Related issue = #938 (if any)

- [[Click here for the format server]](http://kun.csail.mit.edu:31415/)
- [[Click here for coverage report]](http://codecov.io/gh/taichi-dev/taichi/)

PR #1231, opened 2020-06-13
[ir] Don't do optimization for JIT evaluator kernels

xumingkuan:
Related issue = N/A


xumingkuan:
> Nice work! Will we also do print_evaluator_ir=False in this PR?

No, as it's already done in #1226.

Issue #1243, opened 2020-06-14
Block::kernel isn't set correctly in the async computation


k-ye:
**Describe the bug**

`Block::kernel` isn't set correctly in the async computation. In the default mode, this works because the root block maps to a Taichi kernel, which has its kernel set. In the async mode, however, the root block maps to an offloaded task.

Other problems so far:

* [ ] fuse need to respect memory barrier implicitly established by the kernel boundary.
* [ ] constant folding??

k-ye:
One thing I forgot to mention: We should probably be a bit more careful when using default args, e.g. https://github.com/taichi-dev/taichi/blob/24f683655d82c47f9ca152a8994a284358189ad2/taichi/ir/transforms.h#L37

I feel like this has somehow complicated the code flow when debugging these things.

xumingkuan:
I wonder why we still need to pass these parameters given that we can get the `Kernel *` from an `IRNode *`...

k-ye:
> I wonder why we still need to pass these parameters given that we can get the Kernel * from an IRNode *...

Well, this is probably added before we added `IRNode::get_kernel()`? But given that it's not working reliably...

k-ye:
This particular bug is fixed by #1245..

PR #1245, opened 2020-06-14
[ir] Set root block kernel after cloning

k-ye:
I've managed to fix all the `kernel == nullptr` problems that I've found so far. Note that this PR alone doesn't fix all the problems. In order to run `mpm99`, you would still need to disable kernel fusion and constant folding. What's more interesting is that, the main constant folding workflow is fine, the problem was around https://github.com/taichi-dev/taichi/blob/1402c4e5501984e765cacab566117dd07ede07cc/taichi/transforms/constant_fold.cpp#L213-L215

Honestly I think `IRNode::get_kernel()` is mostly broken in the async mode, since it depends on the assumption that root is a block.. 

Related issue = #1243 

----

### DIFF
--- a/taichi/analysis/clone.cpp
+++ b/taichi/analysis/clone.cpp
@@ -114,14 +114,18 @@ class IRCloner : public IRVisitor {
     if (kernel == nullptr) {
       kernel = root->get_kernel();
     }
+    TI_ASSERT(kernel != nullptr);
     std::unique_ptr<IRNode> new_root = root->clone();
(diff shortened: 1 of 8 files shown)


k-ye:
> but shall we modify the behavior in async mode so that the root is a Block containing an OffloadedStmt instead of a raw OffloadedStmt?

Good point. I feel like this is a bit ad-hoc, but I cannot think of how this could break things in any other way.. Allow me to merge this one first

PR #1246, opened 2020-06-14
[test] Add coverage threshold to make codecov not fail so easily

archibate:
Related issue = #

----
According to https://docs.codecov.io/docs/commit-status.

### DIFF
--- a/tests/python/test_matrix.py
+++ b/tests/python/test_matrix.py
@@ -10,7 +10,7 @@ test_vector_arrays = (np.array([42, 42]), np.array([24, 24]))
 
 
 @ti.host_arch_only
-def _test_python_scope_vector_operations():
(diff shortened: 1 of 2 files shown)


tests/python/test_matrix.py
```diff
@@ -19,7 +19,7 @@ def test_python_scope_vector_operations():
 
 
 @ti.host_arch_only
-def test_python_scope_matrix_operations():
+def _test_python_scope_matrix_operations():
```
k-ye, line 22: Were these `_` added in order to test the coverage delta? Shall we revert this file?

archibate: Yes, and as you see, codecov didn't report a failure! We succeed!

PR #1248, opened 2020-06-15
[Opt] Store-to-load forwarding with Control-flow graph

xumingkuan:
Related issue = #926 #656 
Related PR = #932
Replaces #859 #857 (stage 1)

Benchmark on the number of statements: unchanged.

If we remove all `variable_optimization` passes:
benchmark20200616

----


xumingkuan:
```
$1 = offloaded range_for(0, 16) {
  $2 = global ptr ...
  $3 = global load $2
  $4 = abs $3
  $5 = global store [$2 <- $4]
}
```
Currently, the store-to-load forwarding pass finds that `$5` is the only store of `$2`, and replace `$3` with `$4`...

xumingkuan:
A hacky solution: create a temporary `Block` with `GlobalStoreStmt` to everything that is `GlobalLoad`ed in the IR and attach it to the start node of the CFG. Is it safe destructing the `Block` together with its `GlobalStoreStmt`s when destructing the CFG? Or do you have better solutions?

xumingkuan:
Another similar issue:
```
$1 = alloca
$2 = ...
$3 = if $2 {
  $4 = local store [$1 <- $2]
}
$5 = local load $1
```
Maybe we want to make allocas not zero-initialized by default so that $5's UD-chain is not only $4.

xumingkuan:
TODO:
1. mark global ptr & alloca as sources of data
2. CSE for global ptr in another PR

xumingkuan:
I just found there is the 4th type of variable: `ExternalPtrStmt`.

taichi/ir/control_flow_graph.cpp
```diff
@@ -56,6 +78,163 @@ bool CFGNode::erase_entire_node() {
...
+        reach_gen.insert(stmt);
+        reach_kill.insert(data_source_ptr);
+      }
+    }
+  }
+}
+
+bool CFGNode::reach_kill_variable(Stmt *var) const {
+  // Does this node (definitely) kill a definition of var?
+  // return reach_kill.find(var) != reach_kill.end();
```
xumingkuan, line 107: If we return this directly, the running time of `reaching_definition_analysis` here https://github.com/taichi-dev/taichi/issues/926#issuecomment-645092561 will decrease from `45.799 ms` to `14.094 ms`.

yuanming-hu:
Awesome! This looks much cleaner than `variable_optimization`. I'll have to review it carefully tomorrow.

taichi/transforms/cfg_optimization.cpp
```diff
@@ -6,14 +6,16 @@
...
-void cfg_optimization(IRNode *root) {
+void cfg_optimization(IRNode *root, bool after_lower_access) {
   TI_AUTO_PROF;
   auto cfg = analysis::build_cfg(root);
   while (true) {
     bool modified = false;
     cfg->simplify_graph();
     if (cfg->unreachable_code_elimination())
       modified = true;
+    if (!after_lower_access && cfg->store_to_load_forwarding())
```
yuanming-hu, line 17: Maybe we can still do local store forwarding after lower_access?

taichi/analysis/build_cfg.cpp
```diff
@@ -130,26 +130,20 @@ class CFGBuilder : public IRVisitor {
...
-        CFGNode::add_edge(before_offload,
-                          graph->nodes[block_begin_index].get());
-      }
+      int offload_stmt_id = current_stmt_id;
+      auto block_begin_index = graph->size();
+      stmt->body->accept(this);
+      prev_nodes.push_back(graph->back());
+      // Container statements don't belong to any CFGNodes.
+      begin_location = offload_stmt_id + 1;
+      CFGNode::add_edge(before_offload, graph->nodes[block_begin_index].get());
```
yuanming-hu, line 139: I have a question here: are we building a CFG for the body of the OffloadedStmt here that is completely unconnected to other parts of the graph? (If this is the case everything will be easier to understand.)

taichi/analysis/data_source_analysis.cpp
```diff
@@ -0,0 +1,42 @@
+#include "taichi/ir/ir.h"
+#include "taichi/ir/analysis.h"
+#include "taichi/ir/statements.h"
+
+TLANG_NAMESPACE_BEGIN
+
+namespace irpass::analysis {
+
+Stmt *get_data_source(Stmt *store_stmt) {
```
yuanming-hu, line 9:
```suggestion
// Returns the *data* being stored to the destination, in store_stmt.
Stmt *get_data_source(Stmt *store_stmt) {
```

Or maybe rename it to `get_store_data`?

taichi/analysis/data_source_analysis.cpp
```diff
@@ -0,0 +1,42 @@
...
+  } else if (auto local_store = store_stmt->cast<LocalStoreStmt>()) {
+    return local_store->data;
+  } else if (auto global_store = store_stmt->cast<GlobalStoreStmt>()) {
+    return global_store->data;
+  } else {
+    return nullptr;
+  }
+}
+
+Stmt *get_data_source_pointer(Stmt *store_stmt) {
```
yuanming-hu, line 23: `get_store_destination`?

taichi/analysis/build_cfg.cpp
xumingkuan: No. For example, in this case, 
```
$0 = offload list gen ...
$1 = offload {
  ...
}
$2 = offload {
  ...
}
$3 = offload {
  ...
}
```
the CFG will be
```
Node 0 (the start node of CFG): $0, next={Node 1}
Node 1: the first node of $1's body, next=...
...
Node a: the last node of $1's body, next={Node a+1}
Node a+1: the first node of $2's body, next=...
...
Node b: the last node of $2's body, next={Node b+1}
Node b+1: the first node of $3's body, next=...
...
Node c: the last node of $3's body
```

taichi/analysis/data_source_analysis.cpp
xumingkuan: Yeah `get_store_data` and `get_store_destination` are better names.

taichi/transforms/cfg_optimization.cpp
xumingkuan: Yes, then I think we'll have to pass `bool after_lower_access` to `store_to_load_forwarding()` and even `reaching_definition_analysis()`...

taichi/analysis/build_cfg.cpp
xumingkuan: (And so I need to check if the stored data is visible to the load_stmt before doing store forwarding...)

xumingkuan: What about making each OffloadedStmt's CFG unconnected, and make `ControlFlowGraph::start_node` an `std::vector` rather than a `const int`?

xumingkuan: We can perform some cross-offload analyses and optimizations if we make every node connected and if we don't consider each global ptr as a source of data... 

For example:
```
$1 : offload {
  $2 = global ptr a
  $3 = arg load [0]
  $4 : global store $2 <- $3
}
$5 : offload {
  $6 : global ptr a
  $7 : arg load [0]
  $8 : if ... {
    $9 : global store $6 <- $7
  }
  $10 : global load $6 (can be replaced with $7 if we don't consider $6 as a source of data))
}

File: taichi/ir/control_flow_graph.h [hunk 6]
```diff
@@ -15,17 +19,27 @@ class CFGNode {
   CFGNode *prev_node_in_same_block;
   CFGNode *next_node_in_same_block;
 
+  // Edges in the graph
   std::vector<CFGNode *> prev, next;
 
+  // Reaching definition analysis
+  std::unordered_set<Stmt *> reach_gen, reach_kill, reach_in, reach_out;
```
yuanming-hu, line 26: We should probably point people to something like https://en.wikipedia.org/wiki/Reaching_definition

taichi/analysis/build_cfg.cpp
yuanming-hu: I think we can basically stick to what you have now. If we should do cross-offload optimization here needs more discussions.

taichi/transforms/cfg_optimization.cpp
xumingkuan: Oh I've just done it in this PR.

yuanming-hu: OK sounds good. No problem. Thanks!

taichi/ir/control_flow_graph.cpp
```diff
@@ -92,6 +271,85 @@ CFGNode *ControlFlowGraph::back() {
...
+    }
+    auto old_out = std::move(now->reach_out);
+    now->reach_out = now->reach_gen;
+    for (auto stmt : now->reach_in) {
+      if (!now->reach_kill_variable(
+              irpass::analysis::get_store_destination(stmt))) {
+        now->reach_out.insert(stmt);
+      }
+    }
+    if (!(now->reach_out == old_out)) {
```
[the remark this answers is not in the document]
xumingkuan: https://en.cppreference.com/w/cpp/container/unordered_set/operator_cmp

I thought `operator!=` will be removed in C++20, but maybe writing `!(a == b)` in unnecessary... https://stackoverflow.com/questions/58319928/why-is-operator-removed-in-c20-for-many-standard-library-types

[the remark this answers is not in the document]
yuanming-hu: OK that makes sense. Either is fine then.

(Hopefully we can upgrade to C++20 in before 2025.)

xumingkuan:
I feel like we shouldn't consider `GlobalPtrStmt`s as data sources -- because it will not only cause losing some optimization opportunities but also make live variable analysis problematic...
```
$1 = global ptr a
$2 = global store $1 <- 233
$3 = global ptr a
$4 = global load $3
```
If we consider $3 as a data source, $2 will be eliminated...

xumingkuan:
I just came up with an idea to fix this. Will open a PR.

PR #1251, opened 2020-06-15
[Opt] [ir] [refactor] Remove exceptions from IR check_out_of_bound and constant_fold

TH3CHARLie:
Related issue = #1059

----


xumingkuan:
I just took a look -- `check_out_of_bound` and `constant_fold`'s refactoring looks simple enough so that it's fine to merge several passes like this at one time. But passes like `simplify` would be rather complicated, and that should be a single PR.

PR #1257, opened 2020-06-16
[Metal] Add AD stack to support the local mutables when using AutoDiff

k-ye:
I noticed that Metal hasn't supported this yet. It seems like all I need to do is adding a small stack to support the complex AD kernels? I guess OpenGL can do similar things by simulating a per-thread stack, with each entry being a `(primal, adjoint)` pair (@archibate )

Related issue = #581 

----

### DIFF
--- a/taichi/backends/metal/codegen_metal.cpp
+++ b/taichi/backends/metal/codegen_metal.cpp
@@ -15,6 +15,7 @@ namespace {
 
 namespace shaders {
 #define TI_INSIDE_METAL_CODEGEN
+#include "taichi/backends/metal/shaders/ad_stack.metal.h"
 #include "taichi/backends/metal/shaders/helpers.metal.h"
 #include "taichi/backends/metal/shaders/runtime_kernels.metal.h"
 #undef TI_INSIDE_METAL_CODEGEN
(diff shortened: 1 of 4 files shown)


taichi/backends/metal/codegen_metal.cpp
```diff
@@ -551,6 +552,65 @@ class KernelCodegen : public IRVisitor {
     TI_WARN("Cannot print inside Metal kernel, ignored");
   }
 
+  void visit(StackAllocaStmt *stmt) override {
```
k-ye, line 555: Reference implementation from LLVM: https://github.com/taichi-dev/taichi/blob/035bcb99851908fb3d5fb5f2250b907fdbd51392/taichi/codegen/codegen_llvm.cpp#L1421-L1479

taichi/backends/metal/shaders/ad_stack.metal.h
```diff
@@ -0,0 +1,76 @@
...
+// Autodiff stack for local mutables
+
+// clang-format off
+METAL_BEGIN_RUNTIME_STRUCTS_DEF
+STR(
+    // clang-format on
+    using AdStackPtr = thread byte *;
+
+    inline thread uint32_t *
+    mtl_ad_stack_n(AdStackPtr stack) {
```
k-ye, line 31: Reference implementation: https://github.com/taichi-dev/taichi/blob/035bcb99851908fb3d5fb5f2250b907fdbd51392/taichi/runtime/llvm/runtime.cpp#L1305-L1328

taichi/backends/metal/codegen_metal.cpp
```diff
@@ -551,6 +552,65 @@ class KernelCodegen : public IRVisitor {
...
+    TI_ASSERT(stmt->width() == 1);
+
+    const auto &var_name = stmt->raw_name();
+    emit("byte {}[{}];", var_name, stmt->size_in_bytes());
```
archibate, line 559: So we have to use byte array here? Then it might be hard for OpenGL which doesn't support pointers or unions. Or might lose 64-bit support when AD.

taichi/backends/metal/codegen_metal.cpp
```diff
@@ -551,6 +552,65 @@ class KernelCodegen : public IRVisitor {
...
+
+  void visit(StackPopStmt *stmt) override {
+    emit("mtl_ad_stack_pop({});", stmt->stack->raw_name());
+  }
+
+  void visit(StackPushStmt *stmt) override {
+    auto *stack = stmt->stack->as<StackAllocaStmt>();
+    const auto &stack_name = stack->raw_name();
+    const auto elem_size = stack->element_size_in_bytes();
+    emit("mtl_ad_stack_push({}, {});", stack_name, elem_size);
```
archibate, line 571: If I'm thinking right, elem_size will always be 4 in Metal since it only support 32-bit data?

taichi/backends/metal/shaders/ad_stack.metal.h
```diff
@@ -0,0 +1,76 @@
+#include "taichi/backends/metal/shaders/prolog.h"
+
+#ifdef TI_INSIDE_METAL_CODEGEN
+
+#ifndef TI_METAL_NESTED_INCLUDE
+#define METAL_BEGIN_RUNTIME_STRUCTS_DEF constexpr auto kMetalAdStackSourceCode =
+#define METAL_END_RUNTIME_STRUCTS_DEF ;
+#else
+#define METAL_BEGIN_RUNTIME_STRUCTS_DEF
+#define METAL_END_RUNTIME_STRUCTS_DEF
```
archibate, line 10: Really ad-hoc and cool guard!

taichi/backends/metal/shaders/ad_stack.metal.h
```diff
@@ -0,0 +1,76 @@
+#include "taichi/backends/metal/shaders/prolog.h"
+
+#ifdef TI_INSIDE_METAL_CODEGEN
+
+#ifndef TI_METAL_NESTED_INCLUDE
+#define METAL_BEGIN_RUNTIME_STRUCTS_DEF constexpr auto kMetalAdStackSourceCode =
```
xumingkuan, line 6: nit: Shall we use the `TI_` prefix?

taichi/backends/metal/codegen_metal.cpp
k-ye: I guess this will almost always be used for `f32`, so maybe supporting that is good enough?

taichi/backends/metal/codegen_metal.cpp
k-ye: Yeah, I'd say it's *almost* always, but it doesn't prevent us from using more complicated types (For example, one day we may support composite types like `complex {f32, f32}`?) 
Metal only supports 32-bit **primitive** data types, but it's fine to have composite types.

taichi/backends/metal/shaders/ad_stack.metal.h
k-ye: TLDR is no... This is a somewhat obscure mechanism to support both including these this code as C++ code, and stringifying them as real Metal kernel/shader code. Every code in `taichi/backends/metal/shaders` is following this style. The macros here are only meant to be used by the Metal backend, and not shared to a broader scope.

PR #1222, opened 2020-06-11
[IPython] Source inspection dispatcher for better IDLE compatibility

archibate:
Related issue = close #1221 close #1136

- [[Click here for the format server]](http://kun.csail.mit.edu:31415/)
- [[Click here for coverage report]](http://codecov.io/gh/taichi-dev/taichi/)

---

Inspired by https://stackoverflow.com/questions/60797338/get-ast-from-python-object-in-interpreter.

Working in `python` native shell:
```
(dill) [bate@archit taichi]$ python
Python 3.8.2 (default, Feb 26 2020, 22:21:03) 
[GCC 9.2.1 20200130] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import taichi as ti
[Taichi] mode=development
[Taichi] preparing sandbox at /tmp/taichi-440b8jj2
[Taichi] <dev mode>, supported archs: [cpu, cuda, opengl], commit fbdb7ece, python 3.8.2
>>> @ti.kernel
... def f(x: ti.f32) -> ti.f32: return x ** 2
... 
>>> f(1.414)
1.9993960857391357
>>> 
```

<del>Not working in `ipython` shell:</del> working now.

WANT_VER = v0.6.10

PR #1268, opened 2020-06-17
[CLI] Toggle torch import by env var: TI_ENABLE_TORCH

archibate:
Related issue = #

----
To reduce start up time by ~8x, and avoid possible compatibility issues.
I also find that `torch` also import `IPython` too???!! May cause compat issues in #1222.

----
Btw, I noticed that `torch` seems to be not installed in `misc/ci_setup.py`, this means **torch behavior is not tested** in CI, is this intentional?
TBH I install `torch` just for ensuring `ti test`.. my knowledge about torch is completely zero..


archibate:
Merging this given @k-ye's approvement, @yuanming-hu Feel free to tell me **why torch is not tested on CI and assign me an issue on that** later :)

PR #1272, opened 2020-06-18
[Opt] [ir] [refactor] Remove exceptions from demote_atomics pass

TH3CHARLie:
Related issue = #1059

----

### DIFF
--- a/taichi/ir/ir.cpp
+++ b/taichi/ir/ir.cpp
@@ -818,6 +818,7 @@ std::unique_ptr<Block> Block::clone() const {
 
 DelayedIRModifier::~DelayedIRModifier() {
   TI_ASSERT(to_insert_before.empty());
+  TI_ASSERT(to_insert_after.empty());
   TI_ASSERT(to_erase.empty());
   TI_ASSERT(to_replace_with.empty());
 }
--- a/taichi/ir/ir.cpp
+++ b/taichi/ir/ir.cpp
@@ -855,7 +855,8 @@ void DelayedIRModifier::replace_with(Stmt *stmt,
 }
 
 bool DelayedIRModifier::modify_ir() {
-  if (to_insert_before.empty() && to_insert_after.empty() && to_erase.empty() && to_replace_with.empty())
+  if (to_insert_before.empty() && to_insert_after.empty() && to_erase.empty() &&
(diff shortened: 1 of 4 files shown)


TH3CHARLie:
Modify `DelayedModifier` to support `replace_with`.

xumingkuan:
Note: `Block::replace_with(Stmt *old_statement, std::unique_ptr<Stmt> &&new_statement, bool replace_usages)` is safe even if we modify it directly and don't throw exceptions.

TH3CHARLie:
> Note: `Block::replace_with(Stmt *old_statement, std::unique_ptr<Stmt> &&new_statement, bool replace_usages)` is safe even if we modify it directly and don't throw exceptions. So is `Block::replace_with(Stmt *old_statement, VecStatement &&new_statements, bool replace_usages)` only when `new_statements->size() == 1`.

I that case, maybe we can add some flag to `DelayedModifier` and set it to denote that it has been changed. And when `modify_ir` is called, it set the flag back. I believe this mechanism will also work for the `foo=nullptr` change that happened in `simplify`.

xumingkuan:
I'm not sure about the order in `modify_ir()`: if there're only `insert_before` and `erase`, it's always safe to insert everything before erasing anything. Maybe the order doesn't matter if everything in `to_replace_with` doesn't appear in any other vectors?

TH3CHARLie:
> everything

I think(well, more likely guess) that `erase` and `replace_with` will have no conflicts, therefore we should ensure that `to_insert` happens before both `erase` and `replace_with`. In this PR, since no other modification happens in `demote_atomics` pass, the current order would work. But if in the future it fails, I'll try to fix that(and by then I'll have examples to help figure it out).

taichi/ir/ir.cpp
```diff
@@ -819,6 +819,7 @@ std::unique_ptr<Block> Block::clone() const {
 DelayedIRModifier::~DelayedIRModifier() {
   TI_ASSERT(to_insert_before.empty());
   TI_ASSERT(to_erase.empty());
+  TI_ASSERT(to_replace_with.empty());
```
xumingkuan, line 822: Also `to_insert_after` here :-) I forgot to add it when I added the member variable `to_insert_after`.

PR #1292, opened 2020-06-20
[Opt] [ir] [refactor] Remove exceptions from lower_access pass

TH3CHARLie:
Related issue = #1059

----


TH3CHARLie:
One wired thing is that, when I was testing this PR locally, `test_ad_for` failed due to cannot compile kernel on metal backend. Both master and this PR would fail while their IR generation(especially after `lower_access`) results are the same. So I believe this is not caused by my PR but some recent modification to ad/metal. Clearly CI reports no error, maybe another toolchain incompatibility? @k-ye

PR #1298, opened 2020-06-21
[metal] [autodiff] Fix StackLoadTopStmt codegen in Metal

k-ye:
I forgot to declare the variable inside the codegen.

Related issue = #1285 

----

### DIFF
--- a/tests/python/test_ad_for.py
+++ b/tests/python/test_ad_for.py
@@ -327,7 +327,7 @@ def test_double_for_loops_more_nests():
         assert b.grad[i] == total_grad_b
 
 
-@ti.require(ti.extension.adstack)
+@ti.require(ti.extension.adstack, ti.extension.data64)
 @ti.all_archs
 def test_complex_body():
     N = 5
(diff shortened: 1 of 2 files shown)


tests/python/test_ad_for.py
```diff
@@ -327,7 +327,7 @@ def double_for():
         assert b.grad[i] == total_grad_b
 
 
-@ti.require(ti.extension.adstack)
+@ti.require(ti.extension.adstack, ti.extension.data64)
```
TH3CHARLie, line 330: Just curious, why only this case needs the data64 extension?

k-ye: It's because of this `cast<f64>`. Unfortunately, Metal doesn't support `double` yet: https://github.com/taichi-dev/taichi/blob/df62174854af8f5d964b39f0239e0f909810110a/tests/python/test_ad_for.py#L351

TH3CHARLie: Thanks for the clarification!

PR #1310, opened 2020-06-23
[Metal] Add codegen/runtime support for print()

k-ye:
This PR completes the support for `print()` on Metal. Again, it's similar to OpenGL's implementation. I added a new Metal buffer to hold `PrintMsg`. We will need a sync when `print` is used.

Related issue = #1281 

----


taichi/backends/metal/codegen_metal.cpp
```diff
@@ -1043,6 +1086,11 @@ class KernelCodegen : public IRVisitor {
     return kernel_name + "_func";
   }
 
+  void mark_print_used() {
+    TI_ASSERT(current_kernel_attribs_ != nullptr);
+    current_kernel_attribs_->uses_print = true;
```
archibate, line 1091: nit: Use a `UsedFeatures` structure like OpenGL does.

tests/python/test_print.py
```diff
@@ -43,15 +43,33 @@ def func(x: ti.i32, y: ti.f32):
...
 
-@ti.archs_excluding(ti.metal)
+@ti.all_archs
 def test_print_matrix():
     x = ti.Matrix(2, 3, dt=ti.f32, shape=())
     y = ti.Vector(3, dt=ti.f32, shape=3)
 
     @ti.kernel
     def func(k: ti.f32):
+        x[None][0, 0] = -1.0
```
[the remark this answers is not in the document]
archibate: Btw, we may able to assert the print result after #1303?

taichi/backends/metal/codegen_metal.cpp
k-ye: Good point.. This is a sloppy design in Metal. I don't have a data structure that records the attributes of a **taichi** kernel. All i have is a vector of Metal kernels', see https://github.com/taichi-dev/taichi/blob/37976e750e1579ff9328f3d25f1a63cb3b3292e2/taichi/backends/metal/kernel_manager.h#L45.

I'd like to take this opportunity to clean up this flaw in another PR. Once I have a `TaichiKernelAttributes`, i can also add a `UsedFeatures` there. Then we don't even need to iterate over all the **metal** kernels to figure out if we need to flush the print buffers or not. (I tried adding something, but feel like it's too noisy for this one.) WDYT?

tests/python/test_print.py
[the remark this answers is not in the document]
k-ye: Yep, good point. I think we can compare `pop_python_print_buffer` with expected strs.

taichi/backends/metal/codegen_metal.cpp
archibate: Yes, ff2 do it iapr, you may also implement a `operator|=(UsedFeatures, UsedFeatures)` for recording all used feature among kernels :)

k-ye:
Note that code format presubmit failed, which is intentional. I guess the clang formatter is never gonna play well with these `.metal.h` files. The suggested code style added nested indentation for each kernel, and looks terrible.

```
+        [[maybe_unused]] inline int mtl_compute_num_print_msg_typemasks(
+            int num_entries) {
+          return (num_entries + kMetalNumPrintMsgTypePerI32 - 1) /
+                 kMetalNumPrintMsgTypePerI32;
+        }
+
+            [[maybe_unused]] inline int mtl_compute_print_msg_bytes(
+                int num_entries) {
+              // See PrintMsg's layout for how this is computed.
+              const int sz =
+                  sizeof(int32_t) *
+                  (1 + mtl_compute_num_print_msg_typemasks(num_entries) +
+                   num_entries);
+              return sz;
+            }
```

k-ye:
Merging so that i can further address https://github.com/taichi-dev/taichi/pull/1310#discussion_r444901120

PR #1311, opened 2020-06-23
[Lang] Support sep and end in print()

k-ye:
I found that `print` would add a ` ` between every two args. This could be very useful in cases like `print(x, y, z)` --> `1 2 3`. However, for cases like `print('x[', i, ']=', x[i])`, it results in `x[ i ]= 42`. I think the expected outcome should be `x[i]=42`.

Related issue = #

----

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -313,7 +313,7 @@ def layout(func):
 
 
 @taichi_scope
-def ti_print(*vars):
+def ti_print(*vars, sep=' '):
     def entry2content(var):
         if isinstance(var, str):
             return var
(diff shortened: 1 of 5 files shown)


archibate:
Maybe we could support the `sep=''` argument?

TH3CHARLie:
> Maybe we could support the `sep=''` argument?

I like this!

k-ye:
> Maybe we could support the sep='' argument?

SG, fixed.

k-ye:
> Is that possible to add end='\n' argument as well?

I believe the behavior of `print` is always to print the content on a new line. According to the documentation (https://docs.python.org/3/library/functions.html#print), the Python `print` doesn't support this `end` param. So I'd suggest not to. WDYT?

k-ye:
>  Is that possible to add end='\n' argument as well? 

Done. I had to tweak each backend's impl a bit to make it work as expected. PTAL

PR #1312, opened 2020-06-23
[doc] Fix some typos in our document export_results.rst

Rullec:
Fix some typos in the doc `export_results.rst` as @archibate mentioned.

### DIFF
--- a/docs/export_results.rst
+++ b/docs/export_results.rst
@@ -86,7 +86,7 @@ Export videos
 
     The video export utilities of Taichi depend on ``ffmpeg``. If ``ffmpeg`` is not installed on your machine, please follow the installation instructions of ``ffmpeg`` at the end of this page.
 
-- ``ti.VideoManger`` can help you export results in ``mp4`` or ``gif`` format. For example,
+- ``ti.VideoManager`` can help you export results in ``mp4`` or ``gif`` format. For example,
 
 .. code-block:: python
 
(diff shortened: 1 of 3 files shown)


archibate:
Thanks, could you also fix the ones in offset.rst, where `AssertionError` should be `TypeError`?

docs/export_results.rst
```diff
@@ -102,20 +102,20 @@ Export videos
             pixels[i, j, k] = ti.random() * 255
 
     result_dir = "./results"
-    video_manger = ti.VideoManager(output_dir=result_dir, framerate=24, automatic_build=False)
+    video_manager = ti.VideoManager(output_dir=result_dir, framerate=24, automatic_build=False)
```
archibate, line 105: Could you also fix the one in `examples/export_videos.py`?

Rullec:
> where `AssertionError` should be `TypeError`?

Umm, It seems just now I have responsed that problem [here](https://github.com/taichi-dev/taichi-docs-zh-cn/pull/95#discussion_r444229765)

IMO it's not a typo.

For example, If I say
```
import taichi as ti

ti.init()
c = ti.Matrix(2, 1, dt=ti.f32, shape=None, offset=(32,))             # AssertionError
```
then
```
[Taichi] mode=development
[Taichi] preparing sandbox at /tmp/taichi-60ypp77c
[Taichi] <dev mode>, supported archs: [cpu, cuda, opengl], commit 626afc81, python 3.6.10
  File "main.py", line 4, in <module>
    c = ti.Matrix(2, 1, dt=ti.f32, shape=None, offset=(32,))             # AssertionError
  File "/home/xudong/Projects/taichi/python/taichi/lang/matrix.py", line 127, in __init__
    assert offset is None, f"shape cannot be None when offset is being set"
AssertionError: shape cannot be None when offset is being set
```
It is `AssertionError` indeed. So the doc is correct, it matches the behavior.

Do you mean that `TypeError` is better than `AssertionError`

Rullec:
Thank @archibate  for pointing out a typo in `offset.rst`, It has been fixed now.

PR #1318, opened 2020-06-24
[Lang] [refactor] Deprecate x.shape() and x.dim(), use x.shape instead

archibate:
Related issue = https://github.com/taichi-dev/taichi/pull/1214#discussion_r444337338

----

- [x] deprecate `dim()`.
- [x] make `shape()` a `@property`.
- [ ] document `mat.n`, `mat.m` in `meta.rst`.

Will we make  `data_type()` and `snode()` as a `@property` as well?


archibate:
> LGTM. Do you know if there are examples using `shape()` or `dim()`?

I `grep`'ed to find them out, all are removed now, thank for the reminder!

PR #1320, opened 2020-06-24
[misc] Fix compilation warning

k-ye:
Related issue = #1304

----


k-ye:
I believe the CI just went flaky for this PR... Merging

xumingkuan:
So implicitly calling functions inherited from the base class will cause a warning when compiled by clang in Linux?

k-ye:
> So implicitly calling functions inherited from the base class will cause a warning when compiled by clang in Linux?

`clang` (the default compiler on Mac) seems to have a slightly different C++ name lookup rules from `gcc`. See https://stackoverflow.com/a/6727196/12003165

PR #1321, opened 2020-06-24
[GUI] Fix issue with triangles not showing up when they are drawn clockwisely

Rullec:
Hi all,

   One of our classmate in GAMES201 from Wechat found that ONLY when  [va, vb, vc] is in counterclockwise order can `ti.gui.triangle(va, vb, vc)` manage to display it. If the offered vertices are in clockwise, our `ti.GUI` simply ignore it.

   For example, if I say
```
    triangle=[(0.3,0.5),(0.4,0.7),(0.5,0.55)]
    gui.triangle(*triangle,color=0xff0000)
```
  This triangle won't be drawn.

   If it is a designed feature, maybe we can try to doc it?
   If not, here is one possible solution.

HDYT? Maybe we can have a better solution.

### DIFF
--- a/taichi/gui/gui.cpp
+++ b/taichi/gui/gui.cpp
@@ -61,7 +61,12 @@ void Canvas::triangle(Vector2 a, Vector2 b, Vector2 c, Vector4 color) {
       bool inside_a = cross(pixel - a, b - a) <= 0;
       bool inside_b = cross(pixel - b, c - b) <= 0;
       bool inside_c = cross(pixel - c, a - c) <= 0;
-      if (inside_a && inside_b && inside_c && img.inside(i, j)) {
+
+      // cover both clockwise and counterclockwise case for vertices [a, b, c]
+      bool inside_triangle =
+          (inside_a == inside_b) ? (inside_a == inside_c) : false;
+
+      if (inside_triangle && img.inside(i, j)) {
         img[i][j] = color;
       }
     }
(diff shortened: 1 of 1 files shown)


Rullec:
@Eydcao proposed that, it can also be implented by barycentric coordinate. Also a good solution!

taichi/gui/gui.cpp
```diff
@@ -61,7 +61,12 @@ void Canvas::triangle(Vector2 a, Vector2 b, Vector2 c, Vector4 color) {
       bool inside_a = cross(pixel - a, b - a) <= 0;
       bool inside_b = cross(pixel - b, c - b) <= 0;
       bool inside_c = cross(pixel - c, a - c) <= 0;
-      if (inside_a && inside_b && inside_c && img.inside(i, j)) {
+
+      // cover both clockwise and counterclockwise case for vertices [a, b, c]
+      bool inside_triangle =
+          (inside_a == inside_b) ? (inside_a == inside_c) : false;
```
archibate, line 67: nit:
```suggestion
          (inside_a == inside_b) && (inside_a == inside_c);
```

Eydcao:
> @Eydcao proposed that, it can also be implented by barycentric coordinate. Also a good solution!

Thanks~ I was in a meeting just now. I have reviewed your code and find the solution good enough. It turns out it does not matter whether the cross product is positive or negative (or say the particle is at the left side or the right side of the edge), as long as the it is **at the same side** for all edges, it can only be inside.

So this is good enough already. Good fix!

Rullec:
Thanks! You gave such a clear explanation.

So let's see whether there is something more we can improve on in this
issue...

Yadi Cao <notifications@github.com> 于2020年6月25日周四 上午12:18写道：

PR #1324, opened 2020-06-25
[Opt] Dead store and stack-related operation elimination by control-flow graph

xumingkuan:
Related issue = #926 #656 
Related PR = #932 #1248 
Replaces #859 #857 

Change in `build_cfg`: 
- Add an empty `final_node`. Then assume there are reads to all global pointers after the final node (just like there are writes to all global pointers before the start node), for more convenient analysis on global pointers.
- Tell each `CFGNode` if they are parallel executed (i.e. in an offloaded range_for/struct_for's body).

Change in `CFGNode`:
- Add `replace_with`, deprecate `erase_entire_node`.
- Use `contain_variable` for both `reach_kill_variable` (https://en.wikipedia.org/wiki/Reaching_definition) and `live_kill_variable` (https://en.wikipedia.org/wiki/Live_variable_analysis). Keep the function `reach_kill_variable` because `reach_kill` should contain **definition**s instead of **variable**s by definition, but I store variables in the `unordered_set` for convenience.

Change in `compile_to_offloads`:
- Remove the `variable_optimization` pass. (Faster compilation!)

**Benchmark**:

Compilation time (https://github.com/taichi-dev/taichi/issues/926#issuecomment-649839943):
6.872m -> 1.164m
(codegen_kernel_statements: 29087 -> 28573)

Number of statements:

benchmark20200625_2

The only case that becomes worse is:
```python
@ti.all_archs
def test_offload_with_cross_block_locals3():
    ret = ti.var(ti.f32, shape=())

    @ti.kernel
    def ker():
        s = 1
        t = s
        for i in range(10):
            s += i
        ret[None] = t

    ker()

    assert ret[None] == 1
```
Previously, we didn't consider global temp as global ptrs, so we could simplify the kernel to `ret[None] = 1`...

----

### DIFF
--- a/taichi/analysis/build_cfg.cpp
+++ b/taichi/analysis/build_cfg.cpp
@@ -15,18 +15,20 @@ class CFGBuilder : public IRVisitor {
   int current_stmt_id;
   int begin_location;
   std::vector<CFGNode *> prev_nodes;
+  bool in_offloaded_for;
 
  public:
   CFGBuilder()
       : current_block(nullptr),
         last_node_in_current_block(nullptr),
         current_stmt_id(-1),
-        begin_location(-1) {
+        begin_location(-1),
+        in_offloaded_for(false) {
     allow_undefined_visitor = true;
     invoke_default_visitor = true;
     graph = std::make_unique<ControlFlowGraph>();
     // Make an empty start node.
-    auto start_node = graph->push_back(nullptr, -1, -1, nullptr);
+    auto start_node = graph->push_back();
     prev_nodes.push_back(start_node);
   }
 
@@ -38,7 +40,7 @@ class CFGBuilder : public IRVisitor {
 
   CFGNode *new_node(int next_begin_location) {
     auto node = graph->push_back(current_block, begin_location, current_stmt_id,
-                                 last_node_in_current_block);
+                                 in_offloaded_for, last_node_in_current_block);
     for (auto &prev_node : prev_nodes) {
       CFGNode::add_edge(prev_node, node);
     }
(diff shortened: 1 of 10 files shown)


xumingkuan:
We can weaken an `AtomicOpStmt` to a load if there are no loads after it:
```
$1 = alloca
$2 = local store $1 <- const[10]
$3 = atomic add($1, const[5]) (can be replaced with local load $1)
$4 = print $3
```

We can do this optimization not only when `atomic->dest` is an alloca but also when it's a `global temp` -- as long as there are no loads after it before the next store.

However, we can't do this optimization if it's parallel executed:
```
$0 = offloaded range_for(0, 8) {
  $1 = global temp [offset=0]
  $2 = atomic add($1, const[5]) (cannot be replaced!)
  $3 = global ptr a
  $4 = global store $3 <- $2
}
```

xumingkuan:
How to deal with the "load" in offloaded range for?
```
kernel {
  $0 = offloaded
  body {
    <i32*x1> $1 = global ptr [S4place_i32], index [] activate=false
    <i32 x1> $2 = global load $1
    <i32*x1> $3 = global tmp var (offset = 0 B)
    <i32*x1> $4 : global store [$3 <- $2]
    <i32*x1> $5 = global ptr [S6place_i32], index [] activate=false
    <i32 x1> $6 = global load $5
    <i32*x1> $7 = global tmp var (offset = 4 B)
    <i32*x1> $8 : global store [$7 <- $6]
  }
  $9 = offloaded range_for(tmp(offset=0B), tmp(offset=4B)) block_dim=adaptive
  body {
    <i32 x1> $10 = loop $9 index 0
    <i32*x1> $11 = global ptr [S2place_i32], index [] activate=true
    <i32 x1> $12 = atomic add($11, $10)
  }
}
```
If we don't deal with it, the global stores will be eliminated.

We need something like this:
https://github.com/taichi-dev/taichi/blob/13462ff09560de75fc780de3ee6b0519086ee3b9/taichi/transforms/variable_optimization.cpp#L318-L331
@yuanming-hu Do you have any ideas? I think we probably need to traverse the IR to find a global temp of the same offset with `offload->begin_offset`, and add an `std::vector` for each `CFGNode` denoting the loads not representable in the node...

xumingkuan:
Are there better ways to support offloaded range fors without doing implicit loads...?

PR #1327, opened 2020-06-25
[Misc] Fix a bug in util.py

Rullec:
Hi all,
   When I am testing `ti.info`, I found that there is a literally typo in our `util.py`.
   It will crash all logging APIs. :(


Rullec:
> Thx!

With pleasure :)

PR #1330, opened 2020-06-25
[lang] Deprecate ti.raw_pow

archibate:
Related issue = https://github.com/taichi-dev/taichi/pull/1188#issuecomment-649578632

----
I removed ti.raw_pow and use pow directly instead, but seems the issue still exists? @k-ye Could you confirm that on your end?


k-ye:
Thanks!

As I mentioned, the real problem #1329 is that `random()` is not handled properly (I think). #1188 only exaggerated the problem a bit.

PR #1340, opened 2020-06-26
[Bug] [lang] Fix kernel argument not assignable

archibate:
Related issue = GAMES 201 WeChat Group

----

### DIFF
--- a/tests/python/test_function_parameter_by_value.py
+++ b/tests/python/test_function_parameter_by_value.py
@@ -21,8 +21,6 @@ def test_function_argument_pass_by_value():
 def test_kernel_argument_pass_by_value():
     @ti.kernel
     def task(x1: ti.i32, x2: ti.i32):
-        x1 = x1
-        x2 = x2
         if x1 > x2:
             x1 = x2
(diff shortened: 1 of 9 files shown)


tests/python/test_function_parameter_by_value.py
```diff
@@ -15,3 +15,21 @@ def task():
 
     task()
     assert ret[None] == 0
+
+
+@ti.all_archs
+def test_kernel_argument_pass_by_value():
+    @ti.kernel
+    def task(x1: ti.i32, x2: ti.i32):
+        x1 = x1
```
k-ye, line 24: I'm not exactly sure what the problem is. From the title it sounds like we cannot assign kernel arg `x1` to other vars in the kernel? If so, could you assign them to another name, e.g. `x3 = x1`? (Not sure if this suggestion makes sense, though) Or is this self assignment (rare to see) intentional?

archibate:
Seems diff kernels failed with `expr_init(arg)` attempt? Can we optimize out the `AllocaStmt` before verifying KernelSimplificationRule?

PR #1350, opened 2020-06-28
[ir] Fix ir_printer now that print() supports end

k-ye:
Now IR looks normal:

```
kernel {
  $0 = offloaded  
  body {
    <f32 x1> $1 = const [1.0]
    print "[", $1, ", ", $1, "]"
  }
}
```

Off-topic change: I switched Metal's print from `std::cout` to `py_cout`.

Related issue = https://github.com/taichi-dev/taichi/issues/1329#issuecomment-649633371

----


taichi/transforms/ir_printer.cpp
```diff
@@ -221,6 +236,7 @@ class IRPrinter : public IRVisitor {
         name = "\"" + std::get<std::string>(c) + "\"";
```
archibate, line 236: I think the ultimate solution is to make use of [std::quoted](https://en.cppreference.com/w/cpp/io/manip/quoted) that can take care of all escape sequences? i.e.:
```suggestion
        name = std::quoted(std::get<std::string>(c));
```
So that it will print `print "Hello, world!\n"` instead of `print "Hello, world!"`, which should be more clear.

taichi/backends/metal/kernel_manager.cpp
```diff
@@ -681,7 +681,7 @@ class KernelManager::Impl {
     }
 
     if (pa->next >= shaders::kMetalPrintBufferSize) {
-      std::cout << "...(maximum print buffer reached)\n";
+      py_cout << "...(maximum print buffer reached)\n";
```
archibate, line 684: Cool! FF2 merge this off-topic change if you did a careful test on Jupyter & IDLE.

tests/python/test_print.py
```diff
@@ -70,6 +70,8 @@ def func():
         print('chi', 233, 'world!')
         # hello42world!
         print('hello', 42, 'world!', sep='')
+        # '  '
+        print('  ', end='')
```
archibate, line 74:
```suggestion
        # '  ' (with no newline)
        print('  ', end='')
        # 'helloaswd42qwer'
        print('  ', 42, sep='aswd', end='qwer')
```

taichi/transforms/ir_printer.cpp
k-ye: I tried but `std::quoted` wasn't supposed to be used this way (AFAICT, it should be inserted to a stream). I kind you meant to escape `\n`, so i just added a function for that.

taichi/backends/metal/kernel_manager.cpp
k-ye: Ack, tested..

k-ye:
Now the IR looks like this:

```
[I 07/01/20 21:56:33.645] [compile_to_offloads.cpp:operator()@23] Simplified III:
kernel {
  $0 = offloaded  
  body {
    <i32 x1> $1 = const [42]
    print "hello ", $1, " world!\n"
    print "hello ", $1, " Tai"
    <i32 x1> $4 = const [233]
    print "chi ", $4, " world!\n"
    print "hello", $1, "world!\n"
    print "  "
    print "  aswd", $1, "qwer"
  }
}
```

taichi/transforms/ir_printer.cpp
```diff
@@ -34,6 +34,18 @@ std::string block_dim_info(int block_dim) {
...
+  for (char c : s) {
+    if (c == '\n') {
+      ss << '\\' << 'n';
+    } else {
+      ss << c;
+    }
+  }
+  return ss.str();
+}
+
```
archibate, line 48: Then what about '\\', '\r' and '\t'? Then won't display correctly too.

Just FYI:
https://github.com/taichi-dev/taichi/blob/ee0281f61278ebd429593391231e432b3c3fd1fc/taichi/backends/cc/cc_utils.h#L13-L39
Maybe we could make this a common utility in some common header?

k-ye: yeah, that SGTM

k-ye: Done

PR #1365, opened 2020-06-30
[Bug] [autodiff] Hotfix for #1359

xumingkuan:
Related issue = #1359

----


xumingkuan:
Closing because it may cause the result to be wrong...

PR #1371, opened 2020-07-01
[bug] [opt] Fix CFG ignoring local atomics after lower_access

xumingkuan:
Related issue = N/A

`stmt->is<AllocaStmt>() || stmt->is<LocalStoreStmt>()` is not enough -- it didn't cover the local atomics' case.

Also considers range_for and struct_for as parallel for-loops before the `offload` pass.

----


xumingkuan:
@yuanming-hu Could you please take a look at this? I need it for my next PR. Thanks!

PR #1372, opened 2020-07-02
[Opt] Do not flatten ifs by default

xumingkuan:
Related issue = #1367 

benchmark20200701_flatten

It indeed doesn't cause the number of statements to increase too much :)

----

### DIFF
--- a/taichi/program/compile_config.cpp
+++ b/taichi/program/compile_config.cpp
@@ -35,6 +35,7 @@ CompileConfig::CompileConfig() {
   verbose = true;
   fast_math = true;
   async = false;
+  flatten_if = false;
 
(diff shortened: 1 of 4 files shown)


archibate:
IIUC this will cause `if 0:` fail to optimize? Why removing this feature?

xumingkuan:
> IIUC this will cause `if 0:` fail to optimize? Why removing this feature?

`if 0:` is optimized in `unreachable_code_elimination` now.

xumingkuan:
> > IIUC this will cause `if 0:` fail to optimize? Why removing this feature?
> 
> `if 0:` is optimized in `unreachable_code_elimination` now.

I just realized that I didn't *flatten* the `if` in that pass. Will do it iapr.

PR #1376, opened 2020-07-02
[refactor] Move c_quoted into taichi/utils/

k-ye:
Moving this to a common utils so that IR printer can also use it.

@archibate I don't have a way to test the C backend on Mac. Could you confirm this doesn't break?

Related issue = https://github.com/taichi-dev/taichi/pull/1350#discussion_r448384579

----

### DIFF
--- a/taichi/backends/cc/cc_utils.h
+++ b/taichi/backends/cc/cc_utils.h
@@ -10,34 +10,6 @@
 TLANG_NAMESPACE_BEGIN
 namespace cccp {
 
-inline std::string c_quoted(std::string const &str) {
-  // https://zh.cppreference.com/w/cpp/language/escape
-  std::stringstream ss;
-  ss << '"';
-  for (auto const &c : str) {
-    switch (c) {
-#define REG_ESC(x, y) \
-  case x:             \
-    ss << "\\" y;     \
-    break;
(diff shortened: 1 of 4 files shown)


archibate:
But, how about to move this to `lang_util.h` and `lang_util.cpp`? I'm afraid adding a new C++ file just for this may increase compile time.. Or we do `lang_util.cpp` refactor IAPR?

k-ye:
> But, how about to move this to lang_util.h and lang_util.cpp? I'm afraid adding a new C++ file just for this may increase compile time.. 

Is there any evidence this might hurt compilation time? I'd suggest to keep `lang_util` as simple as possible, because it has been included by almost all the files to have this `TLANG_NAMESPACE_BEGIN|END` macros?

k-ye:
> Btw, would you help me set up C backend on OS X after #1354 gets merged?

Sure. I just did a local test with C backend enabled. Didn't encounter compilation error at least...

archibate:
> `lang_util`

Sounds reasonable, thank for the hard work!

> Didn't encounter compilation error

Would you run
```py
ti.init(ti.cc)

@ti.kernel
def hello():
  print('Hello, world!')

hello()
```
to confirm? TTYM in the C backend issue. Let's merge this for now.

PR #1391, opened 2020-07-03
[cli] Use "ti task" for run task (legacy), and "ti run" to run script

archibate:
Related issue = https://forum.taichi.graphics/t/internal-error-occurred/941

----
Some user consider `ti run` as running scripts, like `ti debug` does.
So I move it to `ti task` for runing task instead.

Also fixed the `FIXME` in `ti debug` as OFT.

### DIFF
--- a/taichi/analysis/verify.cpp
+++ b/taichi/analysis/verify.cpp
@@ -40,7 +40,7 @@ class IRVerifier : public BasicStmtVisitor {
           found,
           "IR broken: stmt {} cannot have operand {}."
           " Consider adding `ti.core.toggle_advance_optimization(False)`?"
-          " If that fix the problem, please report this bug by opening an"
+          " If that fixes the problem, please report this bug by opening an"
           " issue at https://github.com/taichi-dev/taichi to help us improve,"
           " many thanks!",
           stmt->id, op->id);
--- a/taichi/analysis/verify.cpp
+++ b/taichi/analysis/verify.cpp
@@ -36,11 +36,14 @@ class IRVerifier : public BasicStmtVisitor {
           break;
         }
       }
-      TI_ASSERT_INFO(found, "IR broken: stmt {} cannot have operand {}."
-                    " Consider adding `ti.core.toggle_advance_optimization(False)`?"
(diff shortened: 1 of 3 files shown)


taichi/analysis/verify.cpp
```diff
@@ -36,8 +36,14 @@ class IRVerifier : public BasicStmtVisitor {
           break;
         }
       }
-      TI_ASSERT_INFO(found, "stmt {} cannot have operand {}.", stmt->id,
-                     op->id);
+      TI_ASSERT_INFO(
+          found,
+          "IR broken: stmt {} cannot have operand {}."
+          " Consider adding `ti.core.toggle_advance_optimization(False)`?"
+          " If that fix the problem, please report this bug by opening an"
```
xumingkuan, line 43:
```suggestion
          " If that fixes the problem, please report this bug by opening an"
```

tests/python/test_cli.py
```diff
@@ -241,7 +241,14 @@ def test_cli_debug():
 
 
 def test_cli_run():
-    with patch_sys_argv_helper(["ti", "run", "test_task", "arg1",
+    with patch_sys_argv_helper(["ti", "run", "a.py"]) as custom_argv:
```
xumingkuan, line 244: What's `a.py`?

archibate: Ask @rexwangcc, I copied from `test_cli_debug`.

xumingkuan: Oh, this function looks exactly the same as `test_cli_debug`. I'm not sure about `test_mode=True` in the next line. Everything other than this function LGTM.

rexwangcc: Thanks for being late here, the `a.py` is just a bogus name for a random script (maybe `foo.py` is more appropriate here) for testing. 

`test_mode=True` is a short circuit flag to tell the CLI exit early so tests won't trigger any real side-effects (see https://github.com/taichi-dev/taichi/pull/1391/files#diff-54e9c326d68839f8cc6d23471881df32R889-R901). This was to make the testing easier, not sure if it's optimal tho.

PR #1382, opened 2020-07-02
[GUI] Support mouse wheel event on Win32

JYLeeLYJ:
Related issue = #1150 

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

## 1.Update 

[GUI] Support mouse wheel event on Win32
[doc] Update document about GUI event.
[Example] Update keyboard.py example , to show how to handle a wheel event . 

I have exported KeyEvent class and Type enum from taichi_core , and support mouse wheel event message on Win32 . 

## 2. Remaining work

1. handle wheel event on X11 
2. handle wheel event on cocoa

Issue #1150, opened 2020-06-05
[GUI] Add mouse wheel event


archibate:
**Concisely describe the proposed feature**
Mouse wheel should also be recognized by GUI as an event.
This could help students to zoom in/out their rendering result. e.g. `fractal.py`.

**Describe the solution you'd like (if any)**
Add `WHEELUP` and `WHEELDOWN` to event keys.
IIRC it's `WM_MOUSESCROLLUP` for win32.

**Additional comments**
More info about event keys at https://taichi.rtfd.io/en/latest/gui.html.

JYLeeLYJ:
1. WHEEL message on [win32](https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-mousewheel) is ```WM_MOUSEWHEEL``` . As for x11 , it seems to be mapped to [mouse button 4 and 5 ](https://stackoverflow.com/questions/15510472/scrollwheel-event-in-x11).

2. Usually , there is an extra parameter in WHEEL message represents the delta value of scrolling motion . So we might need to add an new data member for key event struct in ```gui.h```. Or we ignore this value and just send ```UP``` and ```down``` message to python , which only depends on whether ```delta``` value is negative.

3. How about sperate mouse event from key event ,and add another interface for mouse to python frontend ? It might lead to more modification.

archibate:
IMHO that would be not so easy to create another event queue..
Let's simply reuse `KeyEvent::pos.x` as `delta`, or use a union to represent?
Alternative solution: send `n` UP message when `delta = n`, and send `n` DOWN message when `delta = -n`, not sure how big `n` could be.

JYLeeLYJ:
> Alternative solution: send `n` UP message when `delta = n`, and send `n` DOWN message when `delta = -n`, not sure how big `n` could be.

Unfortunately , it's a short type value on Win32 ...

> Let's simply reuse `KeyEvent::pos.x` as `delta`, or use a union to represent?

I think union or std::variant will be a general solution . Pybind11 does support to export these type automatically.

An possible modification:

```
  struct KeyEvent {
    enum class Type { move, press, release };
    Type type;
    std::string key;
    union {int delta ; Vector2i pos;}; // or getting more type-safety with std::variant<int,Vector2i> ; 
  };
```
Further more, we can individually export the whole ```KeyEvent``` struct like :
```
py::class_<KeyEvent>(m,KeyEvent)
      ...
      .def("get_key" , ...)
      .def("get_type", ...)
[...]

PR #1411, opened 2020-07-05
[GUI] [linux] Support mouse wheel event on Linux/X11

archibate:
Related issue = #1150

----
This PR:
1. Given mouse movement event a key: `ti.GUI.MOVE`, to distinguish from `ti.GUI.WHEEL`.
2. Set up X11 wheel event handling, according to @JYLeeLYJ's advice in https://github.com/taichi-dev/taichi/pull/1382#issuecomment-653187109.

---
@k-ye Would you like add the Mac/Cocoa one later before next release? Thanks.

### DIFF
--- a/examples/keyboard.py
+++ b/examples/keyboard.py
@@ -7,7 +7,7 @@ radius = 8
 gui = ti.GUI("Keyboard", res=(400, 400))
 
 while gui.running:
-    while gui.get_event(ti.GUI.PRESS):
+    while gui.get_event(ti.GUI.PRESS, ti.GUI.MOTION):
         if gui.event.key == ti.GUI.ESCAPE:
             gui.running = False
         elif gui.event.key == ti.GUI.RMB:
--- a/examples/keyboard.py
+++ b/examples/keyboard.py
@@ -7,7 +7,7 @@ radius = 8
 gui = ti.GUI("Keyboard", res=(400, 400))
 
 while gui.running:
-    while gui.get_event(ti.GUI.PRESS, ti.GUI.MOTION):
+    while gui.get_event(ti.GUI.PRESS):
         if gui.event.key == ti.GUI.ESCAPE:
             gui.running = False
         elif gui.event.key == ti.GUI.RMB:
@@ -15,8 +15,8 @@ while gui.running:
         elif gui.event.key == ti.GUI.WHEEL:
             x, y = gui.event.pos
             dt = gui.event.delta
-            # delta is 2-dim vector (dx,dy)
-            # dx(dt[0]) denotes the horizontal direction , and dy(dt[1]) denotes the vertical direction
+            # delta is 2-dim vector (dx, dy)
+            # dt[0] denotes the horizontal direction, and dt[1] denotes the vertical direction
             if dt[1] > 0:
                 radius += 10
             elif dt[1] < 0:
--- a/python/taichi/misc/gui.py
+++ b/python/taichi/misc/gui.py
@@ -26,6 +26,7 @@ class GUI:
     RMB = 'RMB'
     EXIT = 'WMClose'
     WHEEL = 'Wheel'
+    MOVE = 'Motion'
 
     # Event types
     MOTION = ti_core.KeyEvent.EType.Move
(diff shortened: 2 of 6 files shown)


python/taichi/misc/gui.py
```diff
@@ -248,14 +247,18 @@ def get_key_event(self):
...
         if e.key == GUI.WHEEL:
             e.delta = event.delta
+        else:
+            e.delta = (0, 0)
 
         for mod in ['Shift', 'Alt', 'Control']:
             if self.is_pressed(mod):
                 e.modifier.append(mod)
+
         if e.type == GUI.PRESS:
```
JYLeeLYJ, line 257: IMHO , moving wheel event into press type might lead to a problem . Once this event is saved into ```key_pressed``` set , it will never be deleted . Cause wheel event does not have any corresponding 'Release' operation . I think it is better to treat it as a different type from 'Press'

python/taichi/misc/gui.py
```diff
@@ -248,14 +247,18 @@ def get_key_event(self):
...
             self.key_pressed.add(e.key)
         else:
             self.key_pressed.discard(e.key)
```
JYLeeLYJ, line 260: IMHO , moving wheel event into press type might lead to a problem . Once this event is saved into key_pressed set , it will never be deleted . Cause wheel event does not have any corresponding 'Release' operation . I think it is better to treat it as a different type from 'Press'.

taichi/gui/x11.cpp
```diff
@@ -89,8 +89,28 @@ void GUI::process_event() {
       case ButtonPress:
         set_mouse_pos(ev.xbutton.x, height - ev.xbutton.y - 1);
         mouse_event(MouseEvent{MouseEvent::Type::press, cursor_pos});
-        key_events.push_back(
-            KeyEvent{KeyEvent::Type::press, lookup_button(&ev), cursor_pos});
+        switch (ev.xbutton.button) {
+        case 4: // wheel up
+          key_events.push_back(
+              KeyEvent{KeyEvent::Type::press, "Wheel", cursor_pos, Vector2i{0, +120}});
```
k-ye, line 95: Can i ask why using a static value...?

python/taichi/misc/gui.py
JYLeeLYJ: Sorry , I got a wrong understanding with this code just now . It does not matter if we do not use ```gui.is_pressed(GUI.WHEEL)``` ... Just feel free to handle this _\(:3」∠)\_ @archibate

taichi/gui/x11.cpp
JYLeeLYJ: IIRC , X11 doesn't give out a concrete delta value within the wheel event struct .
And 120 is a general value on Win32 (see ```WHEEL_DELTA```)and Qt , which is set as 8 multiple of wheel angle in one step. We just reuse it on X11 .

JYLeeLYJ: > Can i ask why using a static value...?

IIRC , X11 doesn't give out a concrete delta value within the wheel event struct .
And 120 is a general value on Win32 (see WHEEL_DELTA)and Qt , which is set as 8 multiple of wheel angle in one step. We just reuse it on X11 .

taichi/gui/gui.h
```diff
@@ -503,7 +503,7 @@ class GUI : public GUIBase {
   };
 
   struct KeyEvent {
-    enum class Type { move, press, release };
+    enum class Type { move, press, release, scroll };
```
k-ye, line 506: is this used? Or is the plan to switch to `scroll` once wheel events are available on all OS?

archibate: Typo, sorry!
```suggestion
    enum class Type { move, press, release };
```

examples/keyboard.py
```diff
@@ -15,8 +15,8 @@
         elif gui.event.key == ti.GUI.WHEEL:
             x, y = gui.event.pos
             dt = gui.event.delta
```
TroyZhai, line 17: I have mixed feelings about naming this as dt. Normally, dt can be everything else but delta. Could you give it another name here?

PR #1414, opened 2020-07-05
[bug] Fix test_fibonacci

k-ye:
This is probably more of a workaround. I don't think the IR optimizer should change the semantics of the for loop. That means we need to preserve the top-level `if` if the compile time condition is true, or transform it to a serial kernel, like this PR does..

Maybe we should also have a dedicated method to instruct that the for loop is meant to be executed inside a serial kernel. Right now this `for _ in range(1):` seems hacky..

Related issue = https://github.com/taichi-dev/taichi/pull/1393#issuecomment-653854873

----

### DIFF
--- a/tests/python/test_tuple_assign.py
+++ b/tests/python/test_tuple_assign.py
@@ -6,7 +6,8 @@ def test_fibonacci():
     @ti.kernel
     def ti_fibonacci(n: ti.i32) -> ti.i32:
         a, b = 0, 1
-        if 1:
+        # This is to make the inner for loop serial on purpose...
+        for _ in range(1):
             for i in range(n):
                 a, b = b, a + b
         return b


k-ye:
> will the for loop inside function being parallelized?

Hmm, i don't think so? Why?

BTW, did this also break on OpenGL?

archibate:
> Hmm, i don't think so? Why?

Just to confirm...

> BTW, did this also break on OpenGL?

Don't know, my laptop is of OpenGL 3.3 which doesn't support compute shader..

xumingkuan:
IIRC some other unit tests also use `for _ in range(1):` to make the loop serial. It makes sense to me using only `for` (but not `if`) to make the loop serial for now.

PR #1418, opened 2020-07-05
[GUI] Support mouse wheel event on Mac Cocoa

k-ye:
Simpler than i thought..

Related issue = #1150

----

There's a possibility that the vertical scrolling direction of Mac's trackpad is the reverse of that of the mouse wheel.

### DIFF
--- a/taichi/gui/cocoa.cpp
+++ b/taichi/gui/cocoa.cpp
@@ -118,6 +118,7 @@ constexpr int NSApplicationActivationPolicyRegular = 0;
 constexpr int NSEventTypeKeyDown = 10;
 constexpr int NSEventTypeKeyUp = 11;
 constexpr int NSEventTypeFlagsChanged = 12;
+constexpr int NSEventTypeScrollWheel = 22;
 
 struct ModifierFlagsHandler {
   struct Result {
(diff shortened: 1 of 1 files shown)


taichi/gui/cocoa.cpp
```diff
@@ -372,6 +373,14 @@ void GUI::process_event() {
           }
           break;
         }
+        case NSEventTypeScrollWheel: {
+          set_mouse_pos(p.x, p.y);
+          const auto dx = (int)cast_call<CGFloat>(event, "scrollingDeltaX");
+          const auto dy = (int)cast_call<CGFloat>(event, "scrollingDeltaY");
```
archibate, line 379: Is the scale of scrollingDelta around 120?

taichi/gui/cocoa.cpp
```diff
@@ -372,6 +373,14 @@ void GUI::process_event() {
...
+          key_events.push_back(KeyEvent{KeyEvent::Type::move, "Wheel",
+                                        cursor_pos, Vector2i{dx, dy}});
```
archibate, line 381: Is the direction correct along y? iirc mac has a reversed scroll dir than win&x11, could you run examples/keyboard.py and confirm that 'scrolling up like win' scales up the dot?

k-ye: Good point.. I flipped delta Y so that, with Mac's default settings, when swipe **up**, the deltaY will be **positive**. See

zoom

Swiping up will increase the circle's radius. Is this consistent with Win/Linux?

taichi/gui/cocoa.cpp
k-ye: It depends on how hard you scroll, so I don't think we need to dictate a value (at my normal scroll strength it was around 50~70?). In addition, Mac has this ease-out animation effect so that the delta gradually decays to zero...

taichi/gui/cocoa.cpp
JYLeeLYJ: Correct √

PR #1422, opened 2020-07-06
[GUI] Get images from the current GUI using gui.get_image()

Rullec:
Related issue = From Wechat

Our classmate in GAMES201 said that, he want to fetch the current image array from the GUI. 

So here is a trivial implmention, please feel free to give me more suggestions.

Btw, do we need more test cases about this func?

### DIFF
--- a/docs/gui.rst
+++ b/docs/gui.rst
@@ -313,6 +313,12 @@ A *event filter* is a list combined of *key*, *type* and *(type, key)* tuple, e.
 Image I/O
 ---------
 
+.. function:: gui.get_image()
+
+    :return a ``np.ndarray`` which is the current image shown on the GUI.
+
+    Get the shown image from the current GUI system.
+
 .. function:: ti.imwrite(img, filename)
 
     :parameter img: (Matrix or Expr) the image you want to export
--- a/docs/gui.rst
+++ b/docs/gui.rst
@@ -317,7 +317,7 @@ Image I/O
 
     :return a ``np.ndarray`` which is the current image shown on the GUI.
 
-    Get the shown image from the current GUI system.
+    Get the RGBA shown image from the current GUI system which has four channels.
 
 .. function:: ti.imwrite(img, filename)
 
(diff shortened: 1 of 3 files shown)


python/taichi/misc/gui.py
```diff
@@ -75,6 +75,10 @@ def cook_image(self, img):
         assert res == self.res, "Image resolution does not match GUI resolution"
         return np.ascontiguousarray(img)
 
+    def get_image(self):
+        self.core.get_img(self.img.ctypes.data)
```
archibate, line 79: What if self.img is not a contigugous array? We may want to enforce this in set_image.

docs/gui.rst
```diff
@@ -313,6 +313,12 @@ A *event filter* is a list combined of *key*, *type* and *(type, key)* tuple, e.
 Image I/O
 ---------
 
+.. function:: gui.get_image()
+
+    :return a ``np.ndarray`` which is the current image shown on the GUI.
```
archibate, line 318: Can we emphersize that the returned array is a RGBA array?

Rullec: Yes, good.

RGBA is guranteed by the type of `canvas::img` `Array2d<Vector4d>`, right?

python/taichi/misc/gui.py
Rullec: Agree, WIP

Rullec:
How can I trigger the `Check PR Title` mannually after revising?

archibate:
> How can I trigger the `Check PR Title` mannually after revising?

Just update the doc with RGBA, and it wil be triggered again.

python/taichi/misc/gui.py
archibate: Line 115:
```py
            self.img = self.cook_image(img)
```
Could you check if `cook_image` return a RGBA array? I didn't remember..

Rullec: Just now I double check it. If the input of `cook_image` is a valid image (no matter graylike or rgb or rgba), the return will be an RGBA array.

The only exception is when `img`'shape is [x, y, 2], in this case our `cook_image` will fail to return RGBA array. But [x, y, 2] is an invalid image size, IMO.

archibate:
Thanks! Merging this for make a rush for v0.6.17 (#1436).

PR #1423, opened 2020-07-06
[Perf] Support TLS for GlobalTemporaryStmt

k-ye:
Related issue = #576 

----


archibate:
Cool works! But can we make this documented after it's mature? It seems that TLS/BLS doesn't work without user's `ti.cache_shared` instructions.

PR #1405, opened 2020-07-04
[Error] Show pretty & intuitive stack traceback when ti.enable_excepthook() used

archibate:
Related issue = #

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

----
Didn't add `ti.init(excepthook=True)` yet to prevent conflict with #1396.
@k-ye Could you double check that my changes doesn't break your cool ODOP functions? I didn't dear to use a decorator at all..

PR #1437, opened 2020-07-08
[error] Further improve traceback prettizer coverage

archibate:
Related issue = #1405

----


python/taichi/lang/impl.py
```diff
@@ -301,6 +310,8 @@ def var(dt, shape=None, offset=None, needs_grad=False):
             "right after 'import taichi as ti' if you are using Jupyter notebook."
         )
 
+    del _taichi_skip_traceback
```
k-ye, line 313: Just curious, what happens if we don't delete this?

python/taichi/lang/ops.py
```diff
@@ -439,17 +461,21 @@ def append(l, indices, val):
 
 
 def is_active(l, indices):
+    _taichi_skip_traceback = 1
```
k-ye, line 464: Personally i don't think this should be skipped, because these ops are public APIs (IIUC, originally this is to skip some internal implementation details like decorators?) But I don't feel strong about my opinion, your call..

misc/demo_excepthook.py
```diff
@@ -2,20 +2,17 @@
...
 ti.enable_excepthook()
 
-@ti.func
-def func3():
-    ti.static_assert(1 + 1 == 3)
 
 @ti.func
-def func2():
-    func3()
+def fun():
```
k-ye, line 7: nit: i thought the old code would better demonstrate this feature?

python/taichi/lang/impl.py
archibate: If delete: exceptions from codes below this line will be included in traceback.
If not delete: exceptions from codes below this line will not be included in traceback.

PR #1443, opened 2020-07-09
[Metal] Support fast_math, preps for saturating_grid_dim

k-ye:
* `fast_math` helped mostly with computation-bounded tasks. E.g. 
  * sdf_renderer: 25 -> 43 sps
  * cornell_box: 95 -> ~130 sps
* As you can tell, enabling `fast_math` broke a few tests. Given that these tests were doing a `3.0 / 3.0`, i'm not too surprised by that. But I had to turn off `fast_math`.
* @yuanming-hu IIUC, `saturating_grid_dims` can be used to limit the total number of threads? I'd like to do that once users can actually set it in `ti.init()`..

Related issue = #935

----

### DIFF
--- a/taichi/backends/metal/api.cpp
+++ b/taichi/backends/metal/api.cpp
@@ -64,13 +64,14 @@ nsobj_unique_ptr<MTLBlitCommandEncoder> new_blit_command_encoder(
 
 nsobj_unique_ptr<MTLLibrary> new_library_with_source(MTLDevice *device,
                                                      const std::string &source,
+                                                     bool fast_math,
                                                      int msl_version) {
   auto source_str = mac::wrap_string_as_ns_string(source);
 
   id options = clscall("MTLCompileOptions", "alloc");
   options = call(options, "init");
   auto options_cleanup = wrap_as_nsobj_unique_ptr(options);
-  call(options, "setFastMathEnabled:", false);
+  call(options, "setFastMathEnabled:", fast_math);
   if (msl_version != kMslVersionNone) {
     call(options, "setLanguageVersion:", msl_version);
   }
--- a/taichi/backends/metal/kernel_manager.cpp
+++ b/taichi/backends/metal/kernel_manager.cpp
@@ -115,7 +115,7 @@ class CompiledMtlKernelBase {
 
   void launch_if_not_empty(BindBuffers buffers,
                            MTLCommandBuffer *command_buffer) {
-    const int num_threads = get_total_num_threads();
+    const int num_threads = kernel_attribs_.num_threads;
     if (num_threads == 0) {
       return;
     }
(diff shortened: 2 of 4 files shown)


k-ye:
FYI, I think the ring artifacts were also due to fast math. Now it looks like this with `max_ray_depth=1`...

Screen Shot 2020-07-09 at 21 15 25

tests/python/test_element_wise.py
```diff
@@ -10,7 +10,7 @@ def _c_mod(a, b):
 
 @pytest.mark.parametrize('lhs_is_mat,rhs_is_mat', [(True, True), (True, False),
                                                    (False, True)])
-@ti.all_archs
+@ti.all_archs_with(fast_math=False)
```
archibate, line 13: Does `pytest.approx` with larger `rel=` and `abs=` solve your issue?

taichi/backends/metal/kernel_manager.cpp
```diff
@@ -130,40 +132,49 @@ class CompiledMtlKernelBase {
...
-    int num_threads_per_group = 0;
-    // Sometimes it is helpful to limit the maximum GPU block dim for the
-    // kernels. E.g., when you are generating iPhone shaders on a Mac.
-    const int prescribed_block_dim =
-        (std::size_t)get_current_program().config.max_block_dim;
-    if (prescribed_block_dim != 0) {
-      num_threads_per_group = std::min(native_block_dim, prescribed_block_dim);
-    } else {
-      num_threads_per_group = native_block_dim;
-    }
```
archibate, line 145: OFT?

taichi/backends/metal/api.cpp
```diff
@@ -64,13 +64,14 @@ nsobj_unique_ptr<MTLBlitCommandEncoder> new_blit_command_encoder(
...
                                                      const std::string &source,
+                                                     bool fast_math,
                                                      int msl_version) {
   auto source_str = mac::wrap_string_as_ns_string(source);
 
   id options = clscall("MTLCompileOptions", "alloc");
   options = call(options, "init");
   auto options_cleanup = wrap_as_nsobj_unique_ptr(options);
-  call(options, "setFastMathEnabled:", false);
+  call(options, "setFastMathEnabled:", fast_math);
```
archibate, line 74: Cool! Not sure about how `fast_math` works. IIUC is this the same as specifying [`precision mediump float;`](https://blog.csdn.net/weixin_45360983/article/details/100084990) in OpenGL?

k-ye: They are both for optimizing performance, but take different approaches. From what i can tell, `precision mediump float` reduces the bits to represent float. Metal also has a similar concept of `half` (16bits). Fast math, on the other hand, reduces the instructions in the computation to produce an approximated result.

taichi/backends/metal/kernel_manager.cpp
k-ye: Not really :) These are the necessary changes for supporting `saturing_grid_dim`.

tests/python/test_element_wise.py
k-ye: Nope, it failed at `//`, which translates to `floor(y / z)`. For `y[1][1]` and `z[1][1]`, they are both `3`, so I guess that the division becomes something like `0.99999`, which gets floored to `0.0` instead of `1.0`. I just tested by setting `y` at a slightly higher value, e.g. `3.0001`, and it worked..

k-ye:
> That's used to limit the maximum number of groups (blocks on CUDA). On CUDA if we use grid-stride loops, there's no need to allocate more than 16/32 blocks per streaming multiprocessor. Otherwise the blocks will be queued to bind to available SMs.

Ah I see. It seems like users aren't given the option to turn off grid-strided loop? That is, if `saturating_grid_dim = 0`, then it gets set to `num_SM * 32`... For some reason, `sdf_render.py` does run faster without grid-stride loop, so I guess the default params would work a bit differently on Metal.

> Note that although max larger max_ray_depth, you may seemingly rid of this artifact, this can still affect convergence/accuracy.

Ack. But if i turn off fast math, and still use `max_ray_depth=1`, the artifacts are gone.. See below:

Screen Shot 2020-07-11 at 0 32 40

yuanming-hu:
> Ah I see. It seems like users aren't given the option to turn off grid-strided loop? That is, if `saturating_grid_dim = 0`, then it gets set to `num_SM * 32`... For some reason, `sdf_render.py` does run faster without grid-stride loop, so I guess the default params would work a bit differently on Metal.

It's interesting that without grid-stride loops `sdf_renderer.py` runs faster. Maybe not using grid-stride loops gives you better load balancing? I'm not sure what happens if CUDA doesn't use grid-stride loops :-) The other motivation to use grid-stride loops, is that when the bounds of the loop depend on the previous kernel, using grid-stride loop allows CUDA to launch the kernel without waiting for the previous kernel to finish, since you don't need to know how many iterations are needed when you *launch* the kernel with grid-stride loop.

Feel free to do any specialization for Metal if that's beneficial to performance.



> Ack. But if i turn off fast math, and still use `max_ray_depth=1`, the artifacts are gone (maybe not completely, I still seem to see a round shape if zoomed in..) See below:

Interesting... I guess it's something related to numerical precision. The ring artifacts sometimes happen when you have `sqrt`/`1/sqrt` functions calls that need to be high-precision.

k-ye:
> using grid-stride loop allows CUDA to launch the kernel without waiting for the previous kernel to finish

Yep, that I remember and have done for Metal as well :)

https://github.com/taichi-dev/taichi/blob/0d2e34fbd1a08846fc08a79a15163efdbfae5486/taichi/backends/metal/codegen_metal.cpp#L846-L854

k-ye:
> Thanks! Didn't pass very carefully. Given @yuanming-hu's approvement, let's merge this for now to pvc with #1474.

Thanks! I've been working on #1480 this whole weekend and didn't get the time to re-check this. Let me verify if there's any example broken by fast math in the next few days...

PR #1447, opened 2020-07-10
[cli] Print some useful links in CLI banner

archibate:
Related issue = WeChat

----
A GAMES 201 student in WeChat group seems don't know our documentation yet, let's advertise it in our help message.
Or should this be placed in somewhere else for eye-catching?


python/taichi/main.py
```diff
@@ -88,12 +88,42 @@ def __call__(self):
...
+        """Show help message."""
+        parser = argparse.ArgumentParser(prog='ti help',
+                                         description=f"{self.help.__doc__}")
+        parser.add_argument(
+            '-s',
+            '--save',
+            required=False,
+            dest='save',
+            action='store_true',
+            help="Save source code to current directory instead of running it")
```
rexwangcc, line 109: this seems to be a mis copy-paste, `ti help` doesn't need any argument if I understand it correctly.

python/taichi/main.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, debug: bool = False, test_mode: bool = False):
     def __call__(self):
         # Print help if no command provided
         if len(sys.argv[1:2]) == 0:
-            self.main_parser.print_help()
+            self.help()
```
rexwangcc, line 81: I think this works in general, except `ti -h` won't print the links, which is fine to me.

archibate: Yes, I just confirmed that `ti -h` doesn't print the links. Any idea on how to solve this?

archibate: Solved by moving links to right-below of the banner.

archibate:
@yuanming-hu Merging this for now given that @k-ye&@rexwangcc's approve and you don't have time on these minor changes, feel free to revert me if you find these links are bad.

PR #1449, opened 2020-07-10
[misc] 'ti.init()' now supports all exported configurations in ti.cfg

archibate:
Related issue = WeChat talk with @yuanming-hu 

----


archibate:
> Very cool!
> 
> > Maybe CUDA specific configs like device_memory_* should be moved to somewhere like ti.cuda_cfg so that user don't get confused?
> 
> It's generally a good idea to add a `cuda` namespace. However in this case I guess it's better to keep the argument list simple?

Yes, what about `ti.init(cuda_device_memory_GB=1)` and `ti.cuda_cfg.device_memory_GB`? Let's move the talk iapr.

PR #1451, opened 2020-07-10
[doc] [example] Update faq.rst, add examples/revert_image_color.py

archibate:
Related issue = https://forum.taichi.graphics/t/pilimage-as-input/977

----

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -20,7 +20,11 @@ Frequently asked questions
         for i in range(100):  # This loop will not be parallelized
             ...
 
-**Q:** Can Taichi co-operate with **other Python packages** like ``matplotlib`` or ``PILImage``?
+**Q:** What's the most convinent way to load images / textures into Taichi tensors?
+
+**A:** Simply use ``tensor.from_numpy(ti.imread('filename.png'))``.
+
+**Q:** Can Taichi co-operate with **other Python packages** like ``matplotlib``?
 
 **A:** Yes, as long as that *package* provides an interface with ``numpy``, see :ref:`other_python_packages`.
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,7 +1,49 @@
 Frequently asked questions
 ==========================
 
+
+**Q:** Installing Taichi with ``pip``, complains ``package not found``.
+
+**A:** Is your Python version >= 3.6, and 64-bit? See :ref:`troubleshooting`.
+
+**Q:** Do we have something like ``ti.pi``?
+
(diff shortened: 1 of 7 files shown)


docs/faq.rst
```diff
@@ -1,7 +1,53 @@
...
 ==========================
 
+
+**Q:** Installing Taichi with ``pip``, complains ``package not found``.
+
+**A:** Is your Python version >= 3.6, and 64-bit? See :ref:`troubleshooting`.
+
+**Q:** Do we have something like ``ti.pi``?
+
+**A:** No, but you may use ``math.pi`` or ``numpy.pi`` instead.
```
k-ye, line 11:
```suggestion
**A:** No, but you may use ``math.pi`` or ``numpy.pi`` instead. Taichi is able to bake in these constants during JIT, so your kernels incur no runtime cost.
```

docs/faq.rst
```diff
@@ -1,7 +1,53 @@
...
+
+**Q:** How do I **force** a outer-most loop to be serial, i.e. **not parallelized**?
```
k-ye, line 13:
```suggestion
**Q:** How do I **force** an outermost loop to be serial, i.e. **not parallelized**?
```

docs/faq.rst
```diff
@@ -1,7 +1,53 @@
...
+**Q:** How can I **render 3D results** without writing a ray tracer myself?
+
+**A:** You may export it with :ref:`export_ply_files` so that you could view it in Houdini or Blender.
+       Or make use the extension library `Taichi THREE <https://github.com/taichi-dev/taichi_glsl>` to render images and update to GUI in real-time.
+
+**Q:** How do I declare a tensor with **dynamic length**?
+
+**A:** What you want may be the ``dynamic`` SNode, a kind of sparse tensor, see :ref:`dynamic`.
+       Or simply allocate a dense tensor large enough, and another 0-D tensor ``tensor_len[None]`` for length record.
+       But in fact, currently the ``dynamic`` SNode has poor performance compared the latter solution, we are now trying to improve it.
```
k-ye, line 48: I don't think it's poor performance. The latter solution is faster at the cost of more memory, and `dense` doesn't need to care about the sparsity runtime. How about we just rephrase it as "`dynamic` SNode could be slower than the latter solution, due to the cost of maintaining the sparsity information.`

archibate:
```suggestion
       But in fact, the ``dynamic`` SNode could be slower than the latter solution, due to the cost of maintaining the sparsity information.
```

PR #1466, opened 2020-07-11
[test] [bug] Fix master test failure by #1437

archibate:
Related issue = #

----
Sorry about breaking master test :(
But what's the point of using `np.ndarray` in Taichi-scope?

IMO `ti.ext_arr()` is being **edged** and can be easily buggy.. e.g.: https://forum.taichi.graphics/t/hw0-5-implicit-euler-method-on-mass-spring-system/961/13
If we don't maintain `ti.ext_arr()`, how about just deprecate it and do some walk around for `to_numpy`?

### DIFF
--- a/tests/python/test_numpy.py
+++ b/tests/python/test_numpy.py
@@ -134,8 +134,8 @@ def test_numpy_3d():
                 assert a[i, j, k] == i * j * (k + 1) + i + j + k * 2
 
 
-@ti.must_throw(AssertionError)
-def test_numpy_3d():
+@ti.must_throw(IndexError)
+def test_numpy_3d_error():
     val = ti.var(ti.i32)
 
     n = 4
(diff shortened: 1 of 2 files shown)


tests/python/test_numpy.py
```diff
@@ -134,8 +134,8 @@ def test_numpy(arr: ti.ext_arr()):
                 assert a[i, j, k] == i * j * (k + 1) + i + j + k * 2
 
 
-@ti.must_throw(AssertionError)
-def test_numpy_3d():
+@ti.must_throw(IndexError)
+def test_numpy_3d_error():
```
archibate, line 138: I'm wondering who wrote this test... Its name overrides the previous one.

archibate:
Merging this to prevent future failure, let's talk about ext_arr iapr.

PR #1473, opened 2020-07-12
[Bug] [opt] Hotfix for wrong parallel assignments

### LINKED ISSUE #1469: [Bug] Parallel assignment going wrong
The output for the following code should be 1, 2, 3 for each line. However, it occurs 0, 0, 0.
It seems that this only happens after I update v0.6.18.

```
import taichi as ti


ti.init(arch=ti.cpu)
real = ti.f32
mat = ti.var(real, shape=(3, 16))


@ti.kernel
def do_something():
    for r in range(4):
        c = r * 4
        for i in ti.static(range(4)):
            mat[0, c], mat[1, c], mat[2, c] = 1.0, 2.0, 3.0
            c += 1
    for c in range(16):
        print(mat[0, c], mat[1, c], mat[2, c])


if __name__ == "__main__":
    do_something()
```

### LINKED ISSUE #1471: [Bug] Parallel assignment going wrong 2
```
import taichi as ti


ti.init(arch=ti.cpu, print_preprocessed=True)
real = ti.f32
mat = ti.var(real, shape=(3, 16))
cnt = ti.var(dt=ti.i32, shape=())


@ti.kernel
def do_something():
    cnt[None] = 0
    for r in range(4):
        for i in ti.static(range(4)):
            c = ti.atomic_add(cnt[None], 1)
            mat[0, c], mat[1, c], mat[2, c] = 1.0, 2.0, 3.0
    for c in range(16):
        print(mat[0, c], mat[1, c], mat[2, c])


if __name__ == "__main__":
    do_something()
```

xumingkuan:
Related issue = fix #1469 fix #1471

Local loads of the same alloca can have different values.

----

### DIFF
--- a/tests/python/test_optimization.py
+++ b/tests/python/test_optimization.py
@@ -125,3 +125,20 @@ def test_advanced_store_elimination_in_loop():
 
     func()
     assert val[None] == 8
+
+
+@ti.all_archs
+def test_parallel_assignment():
+    mat = ti.var(ti.f32, shape=(3, 16))
+
+    @ti.kernel
+    def func():
+        for r in range(4):
+            c = r * 4
+            for i in ti.static(range(4)):
+                mat[0, c], mat[1, c], mat[2, c] = 1.0, 2.0, 3.0
+                c += 1
+        for c in range(16):
+            assert mat[0, c] == 1.0 and mat[1, c] == 2.0 and mat[2, c] == 3.0
+
+    func()
--- a/tests/python/test_optimization.py
+++ b/tests/python/test_optimization.py
@@ -129,16 +129,16 @@ def test_advanced_store_elimination_in_loop():
 
 @ti.all_archs
 def test_parallel_assignment():
-    mat = ti.var(ti.f32, shape=(3, 16))
+    mat = ti.var(ti.i32, shape=(3, 4))
 
     @ti.kernel
     def func():
-        for r in range(4):
-            c = r * 4
-            for i in ti.static(range(4)):
-                mat[0, c], mat[1, c], mat[2, c] = 1.0, 2.0, 3.0
-                c += 1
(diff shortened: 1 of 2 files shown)


archibate:
Would you add the sample code in #1469 as `test_edge_case.py` before you sys-fix it to prevent regression?

tests/python/test_optimization.py
```diff
@@ -125,3 +125,20 @@ def func():
...
+    def func():
+        for r in range(4):
+            c = r * 4
+            for i in ti.static(range(4)):
+                mat[0, c], mat[1, c], mat[2, c] = 1.0, 2.0, 3.0
+                c += 1
+        for c in range(16):
+            assert mat[0, c] == 1.0 and mat[1, c] == 2.0 and mat[2, c] == 3.0
+
+    func()
```
archibate, line 144: Use this instead?
```py
x = ti.var(ti.i32, shape=4)

@ti.kernel
def func():
    c = 0
    for i in ti.static(range(4)):
        x[c] = 1
        c += 1

func()
assert x.to_numpy() == np.array([[1, 2, 3]] * 16)
```

tests/python/test_optimization.py
```diff
@@ -125,3 +125,20 @@ def func():
...
+
+    @ti.kernel
+    def func():
+        for r in range(4):
+            c = r * 4
+            for i in ti.static(range(4)):
+                mat[0, c], mat[1, c], mat[2, c] = 1.0, 2.0, 3.0
+                c += 1
+        for c in range(16):
+            assert mat[0, c] == 1.0 and mat[1, c] == 2.0 and mat[2, c] == 3.0
```
archibate, line 142: No, `assert` won't work without `debug=True` and non-LLVM archs IIRC..

PR #1482, opened 2020-07-12
[metal] Fix active() on dynamic SNode

k-ye:
This has been fixed in #1480. But given it's not ready yet, I will fix it here first.

Related issue = #1480

----

### DIFF
--- a/taichi/backends/metal/codegen_metal.cpp
+++ b/taichi/backends/metal/codegen_metal.cpp
@@ -247,9 +247,17 @@ class KernelCodegen : public IRVisitor {
       current_appender().append_raw(
           make_sparse_snode_meta(stmt->snode, kSNodeMetaVarName));
       const bool is_dynamic = (stmt->snode->type == SNodeType::dynamic);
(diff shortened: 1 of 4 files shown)


k-ye:
BTW, i don't have OpenGL backend, but i think the GL impl is correct so it won't break..

archibate:
I don't have the GL backend either... @xumingkuan Do you have it? Would you mind do me a favor by `ti test -a opengl` and see if OpenGL is still functional these days?

xumingkuan:
I don't have it either...

archibate:
> I don't have it either...

Oh sorry! @yuanming-hu Would you do me this favor? IIRC you have a NV card with compute shader support?

k-ye:
Merging. I believe OpenGL has the correct implementation: https://github.com/taichi-dev/taichi/blob/d756cbd0c06d8f36ab67e0ad3a91c78192dc4b27/taichi/backends/opengl/codegen_opengl.cpp#L323-L325

PR #1489, opened 2020-07-13
[async] Put cloned offloaded tasks under a dummy shared block

k-ye:
`get_kernel()` wasn't 100% fixed in #1245. This PR takes @xumingkuan 's suggestion in https://github.com/taichi-dev/taichi/pull/1245#pullrequestreview-430242227, by creating a dummy `Block` for each kernel. (We could also link the cloned offloaded task's parent to that of the *clonee*. But then this feels like mixing ASTs up..)

This PR is still not enough to make async work again. We still need these:

* disable constant folding JIT. I got some LLVM errors, and didn't look further. My guess is that right now, we expect JIT kernels to finish synchronously in order to use their result. This no longer holds in async mode.
* disable kernel fusion for `mpm99.py`. 

Related issue = #1243

----

### DIFF
--- a/taichi/program/async_engine.cpp
+++ b/taichi/program/async_engine.cpp
@@ -12,6 +12,8 @@
 
 TLANG_NAMESPACE_BEGIN
 
+namespace {
+
 uint64 hash(IRNode *stmt) {
   TI_ASSERT(stmt);
   // TODO: upgrade this using IR comparisons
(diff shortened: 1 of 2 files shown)


taichi/program/async_engine.cpp
```diff
@@ -25,15 +27,29 @@ uint64 hash(IRNode *stmt) {
   return ret;
 }
 
+std::unique_ptr<IRNode> clone_offloaded_task(OffloadedStmt *from,
+                                             Kernel *kernel,
+                                             Block *dummy_root) {
+  auto new_ir = irpass::analysis::clone(from, kernel);
+  // This is not the ideal fix, because |new_ir|'s children blocks are NOT
+  // linked to |dummy_root|. However, if I manually do the linking, I got error
+  // during LLVM codegen.
```
xumingkuan, line 36: `irpass::fix_block_parents(dummy_root)` will result in an error?

k-ye: Yeah. FYI I don't think `fix_block_parents(dummy_root)` does the correct thing. Instead, we should still call `fix_block_parents(new_ir.get())`, but set `dummy_root` to https://github.com/taichi-dev/taichi/blob/faff7024bc09dc56e89f57c6cf68993e1c3e74b1/taichi/transforms/fix_block_parents.cpp#L16. However, that also didn't work, and I got errors when visiting some `GlobalStoreStmt*` in `mpm99.py`..

PR #1492, opened 2020-07-14
[Lang] Enhance warning system and support runtime assert message

archibate:
Related issue = #

----
I don't feel so like `warnings.warn` because it can't print the full stack traceback...


python/taichi/misc/util.py
```diff
@@ -132,14 +132,48 @@ def get_line_number(asc=0):
...
+    import traceback
+    import taichi as ti
+    use_spdlog = False
+    if use_spdlog:
+        s = traceback.extract_stack()[:-stacklevel]
+        raw = ''.join(traceback.format_list(s))
+        ti.warn(f'{type.__name__}: {msg}')
+        ti.warn(f'\n{raw}')
+    else:
+        warnings.warn(msg, type, stacklevel=stacklevel + 1)
```
archibate, line 146: WDYT?

tests/python/test_assert.py
```diff
@@ -26,6 +27,17 @@ def func():
     func()
 
 
+@ti.all_archs_with(debug=True, gdb_trigger=False)
```
archibate, line 30: #1488

archibate:
> The code itself LGTM. One thing to think about: there will be `ti.warn` and `ti.warning`, which could be a bit confusing?

Yes, renaming one of them make sense. How about `ti.warn(msg)` -> `ti.spdlog(ti.WARN, msg)`?

k-ye:
> How about ti.warn(msg) -> ti.spdlog(ti.WARN, msg)?

Maybe just drop `spd`, since SPD is kind of an implementation detail. 

But while I pointed out the confusion, I think it's fine to keep this as is for now. Once we have cleaned up the module relationships (e.g. your #1510) and can have nested modules, I think it's better to group these non-user facing methods in a deeper module (e.g. `ti.misc.warning` or something like that)

archibate:
> drop spd

In fact `ti.log` is already a math function...

> #1510

Yes, we may have `ti.misc.log` and `ti.lang.log`, please ff2 leave review when you've time :)

PR #1494, opened 2020-07-14
[cli] Remove the never used 'tid' command

archibate:
Related issue = #

----
1. Remove the `tid` command:
We already have `ti debug` and the `debug` argument in `main` seems useless...
In fact, it's already not functional at all:
```
root@archlinux ~/taichi (git)-[remove-tid-useless-bin-hook] # tid
[Taichi] mode=development
[Taichi] preparing sandbox at /tmp/taichi-6z2hiodm
[Taichi] <dev mode>, llvm 10.0.0, commit a93efc54, python 3.8.3
  File "/root/taichi/bin/tid", line 4, in <module>
    exit(taichi.main(debug=True))
TypeError: main() got an unexpected keyword argument 'debug'
```

2. Use `#!/usr/bin/env python3` instead of `#!${PYTHON_EXECUTABLE}`:
Well, what if one use anaconda python3?
The PYTHON_EXECUTABLE is determinated by the buildbots, I don't think it's always the same as end-users...
In contrast, `/usr/bin/env` will automatically search for `python3` in current `PATH`, which will make `ti` command work in anaconda python3.
@k-ye Just to confirm, is `/usr/bin/env` always available on Mac?

3. Make Linux users without CUDA able to use `ti dist`.
I thought our Linux buildbot have constant CUDA support? So no need for this confirm?

### DIFF
--- a/python/build.py
+++ b/python/build.py
@@ -49,10 +49,6 @@ if platform.system() == 'Linux':
         print('Only the wheel with clang will be released to PyPI.')
         sys.exit(-1)
 
-    if not gpu:
-        print('Linux release must ship with the CUDA backend.')
-        sys.exit(-1)
-
 with open('../setup.py') as fin:
     with open('setup.py', 'w') as fout:
         project_name = 'taichi'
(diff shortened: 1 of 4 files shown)


python/build.py
```diff
@@ -49,10 +49,6 @@ def get_python_executable():
         print('Only the wheel with clang will be released to PyPI.')
         sys.exit(-1)
 
-    if not gpu:
-        print('Linux release must ship with the CUDA backend.')
-        sys.exit(-1)
-
```
archibate, line 55: Just to confirm: @yuanming-hu Does this have fatal drawback?
I just want to run `ti dist` on my non-CUDA laptop.

archibate: Merging this given that v0.6.19 is already released, ff2 revert me if you don't like this.

PR #1496, opened 2020-07-14
[Lang] Support "break" and "continue" in static-fors

### LINKED ISSUE #1495: [bug or feature] "break/continue" not working in static loops
```py
import taichi as ti


@ti.kernel
def func():
    for i in ti.static(range(4)):
        print(i)
        if i == 2:
            break

func()
```
obtains:
```
RuntimeError: [detect_fors_with_break.cpp:visit@22] break statement out of loop scope
```
I guess this is because `break` is transformed into `insert_break_stmt`, and there's no real Taichi loop, just a Python static loop.

So there's no ways to break/continue a static for loop?

In `visit_Break` and `visit_Continue`, detect if it's in a static loop:
- if so, don't transform.
- if not, transform into `insert_break_stmt`.

**Additional comments**

archibate:
Related issue = close #1495

----
I added the concept of `control_scope` to detect if the current loop is a static loop, so that we can prevent translating the `break` there to be breaking the grandparent non-static loop.

### DIFF
--- a/python/taichi/lang/transformer.py
+++ b/python/taichi/lang/transformer.py
@@ -228,7 +228,7 @@ if 1:
   else:
     break
   ti.core.pop_scope()
-    '''
+'''
             cond = node.test
             t = ast.parse(template).body[0]
             t.body[1].value = cond
@@ -584,12 +584,12 @@ if 1:
         return node
 
     def visit_Module(self, node):
-        with self.variable_scope(), self.control_scope():
+        with self.variable_scope():
             self.generic_visit(node)
         return node
 
     def visit_Global(self, node):
-        with self.variable_scope(), self.control_scope():
+        with self.variable_scope():
             self.generic_visit(node)
         for name in node.names:
             self.create_variable(name)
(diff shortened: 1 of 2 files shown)


python/taichi/lang/transformer.py
```diff
@@ -32,18 +32,27 @@ def __init__(self,
...
         self.arg_features = arg_features
         self.returns = None
 
+    # e.g.: FunctionDef, Module, Global
     def variable_scope(self, *args):
-        return ScopeGuard(self, *args)
+        return ScopeGuard(self.local_scopes, *args)
+
+    # e.g.: For, While
+    def control_scope(self, *args):
```
k-ye, line 47: Nit: `ScopeGuard` is specialized for local variables, because its `__exit__` has this operation to delete the local variables in the scope, which doesn't make sense to control scope. I think the easiest thing to prevent errors is just to remove `*args` here, so that `stmt_block` will be `None` for control scopes.

python/taichi/lang/transformer.py
```diff
@@ -32,18 +32,27 @@ def __init__(self,
...
+        return ScopeGuard(self.control_scopes, *args)
```
archibate, line 48:
```suggestion
    def control_scope(self):
        return ScopeGuard(self.control_scopes)
```

Issue #1500, opened 2020-07-15
[Lang] [refactor] Use "field" to replace "var" in API, "tensor" in doc


archibate:
@Rullec @yuanming-hu Continue our discussion in: https://github.com/taichi-dev/taichi/issues/1156#issuecomment-658532635

# 1. Introduction about our new allocating API
The new naming mechanism can be formulated in these 3 rules:
- use ``ti.field(dtype, ...)`` to allocate a **global scalar**  field (tensors)
- use ``ti.Vector.field``/``ti.Matrix.field`` to allocate **global vector/matrix** fields (tensors)
- use `ti.Vector` and `ti.Matrix` for **local** variable

In one word, anything **involved in the name ``field``** must be a **global** stuff, and all others must be some **local** things.
The difference between old and new names are exposed below, in the following working plan.

# 2. Working plan

## 2.1 Add new API

 - [x] Support `ti.field`, `ti.Vector.field`, `ti.Matrix.field`

Simply renaming `ti.var` to `ti.field` should work.
In fact we already have `ti.Matrix.var`, so please rename that too :)

- [x] Use `dt` -> `dtype` as arguement by-the-way

Btw, the argument `dt` should be renamed to `dtype`, i.e. `ti.field(dtype=ti.f32, shape=())`.

## 2.2 Deprecate old API

 - [x] Deprecate `ti.var`, `ti.Vector.var`, `ti.Matrix.var`

We don't want to fully abandon `ti.var` support yet, so let's only raise a warning before the API-breaking v0.7.0.

Let's make a dummy `ti.var` with simply redirect to corresponding `field` function:
```
def var(...):
  return field(...)
```
And please use the `@deprecated` decorator, it will raise a warning when the decorated function is being called, see `ti.Matrix.transposed` for usage example.

 - [x] Deprecate `ti.Vector`, `ti.Matrix` usage for declaring global matrix fields.

We may not use `@deprecated` since `ti.Vector` is already for local matrix variables... So please use `ti.warning` in if-elif's instead.

## 2.3 Update docs, tests, examples for API change

- [x] Update examples for "ti.var" -> "ti.field"
- [x] Update tests for "ti.var" -> "ti.field"
- [x] Update docs for "ti.var" -> "ti.field"

Simply `s/\.var/\.field/g` should works. Relative trivial, compare to the next one:

## 2.4 Update docs for term change

- [x] Update docs for "tensor" -> "field"

[...]

PR #1502, opened 2020-07-15
[Refactor] [lang] Add "ti.field" and "ti.Matrix.field" for future deprecation of "ti.var"

archibate:
Related issue = #1500:2.1

----

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -291,7 +291,7 @@ def var(dt, shape=None, offset=None, needs_grad=False):
     return field(dt, shape, offset, needs_grad)
 
 @python_scope
-def field(dt, shape=None, offset=None, needs_grad=False):
+def field(dtype, shape=None, offset=None, needs_grad=False):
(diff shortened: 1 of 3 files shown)


python/taichi/lang/impl.py
```diff
@@ -285,8 +285,14 @@ def __repr__(self):
 root = Root()
 
 
-@python_scope
+#@deprecated('ti.var', 'ti.field')
```
k-ye, line 288: Are these `@deprecated` commented out on purpose?

archibate: Yes, we don't want to throw out tons of warnings yet, before it's fully deprecated from tests and examples. The `#@` is just a reminder, this is **to be deprecated in the future**.

archibate:
> Thanks and LGTM in general. Could you clarify a bit on why the `@deprecated`s are commented out?

Please see #1500, we **will only uncomment this after all docs, tests, examples, are adopt to `ti.field`**, for now they're just a reminder.

PR #1508, opened 2020-07-16
[Linux] [llvm] Fix LLVM symbol leakage to prevent conflict with other libs using LLVM like GLX

### LINKED ISSUE #958: Can't import taichi (from 0.5.15) using Ubuntu 20.04 in VM

archibate:
Related issue = fix #958

----
By using `libLLVM*.a`, we not only export `taichi::*` in our `libtaichi_core.so`, but also export `llvm::*`.
So that some symbols in `libLLVM-10.so` (also `llvm::*`) somehow gets replaced by those one in `libtaichi_core.so`, which is our **RTTI configured** LLVM, causing the conflict.
So, this PR uses a [LD version script](http://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_node/ld_25.html), to **prevent symbols in `llvm::*` being exported to global symbol table**, therefore fix the leakage conflict, ultimately.

### DIFF
--- a/cmake/TaichiCore.cmake
+++ b/cmake/TaichiCore.cmake
@@ -153,10 +153,11 @@ if (NOT WIN32)
         # Linux
         target_link_libraries(${CORE_LIBRARY_NAME} stdc++fs X11)
         target_link_libraries(${CORE_LIBRARY_NAME} -static-libgcc -static-libstdc++)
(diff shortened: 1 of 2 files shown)


archibate:
Just BTW, does this fix the Blender compat issue which IIRC is related OpenGL?

PR #1513, opened 2020-07-16
[test] Add test_fuse_dynamic.py to unit test

k-ye:
TODO: I'll see if I can add `test_mpm88` as well later..

Related issue = #1504 

----


archibate:
The CI will fail as expected. Please retrigger CI by `git rebase master; git push -f` after #1514.

tests/python/test_fuse_dynamic.py
```diff
@@ -0,0 +1,58 @@
+import taichi as ti
+import time
+import pytest
+
+
+def benchmark_fuse_dynamic_x2y2z(size=1024**2, repeat=10, first_n=100):
```
archibate, line 6: Why `benchmark`?

k-ye: Eventually, we should set up benchmarks and/or regression tests using this function as well.

PR #1522, opened 2020-07-17
[Example] [autodiff] Add ad_gravity.py (compute gradient from potential energy)

archibate:
Related issue = #

----
I :heart: AutoDiff! But currently our AD examples are only about ML... Let's add a CG-related example now.

### DIFF
--- a/examples/ad_gravity.py
+++ b/examples/ad_gravity.py
@@ -1,0 +1,39 @@
+import taichi as ti
+
+N = 8
+dt = 5e-5
+
+pos = ti.Vector.var(2, ti.f32, N, needs_grad=True)
+vel = ti.Vector.var(2, ti.f32, N)
+potential = ti.var(ti.f32, (), needs_grad=True)
+
+@ti.kernel
+def calc_potential():
+    for i, j in ti.ndrange(N, N):
+        disp = pos[i] - pos[j]
+        potential[None] += 1 / disp.norm(1e-3)
+
+@ti.kernel
+def init():
+    for i in pos:
+        pos[i] = [ti.random(), ti.random()]
+
+@ti.kernel
+def advance():
+    for i in pos:
+        vel[i] += dt * pos.grad[i]
+    for i in pos:
+        pos[i] += dt * vel[i]
+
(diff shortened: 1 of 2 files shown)


python/taichi/lang/__init__.py
```diff
@@ -265,7 +265,13 @@ def svd(A, dt=None):
 
 def Tape(loss, clear_gradients=True):
     get_runtime().materialize()
-    assert loss.snode().ptr.has_grad(), "gradient for loss not allocated"
+    if len(loss.shape) != 0:
+        raise RuntimeError(
+            'The loss tensor for `Tape` must be a 0D tensor, i.e. scalar')
```
Rullec, line 270:
```suggestion
            'The loss of `Tape` must be a 0D tensor, i.e. scalar')
```

python/taichi/lang/__init__.py
```diff
@@ -265,7 +265,13 @@ def svd(A, dt=None):
...
+    if not loss.snode().ptr.has_grad():
+        raise RuntimeError(
+            'Gradient for loss not allocated, please use ti.var(..., needs_grad=True)'
```
Rullec, line 273:
```suggestion
            'Gradients of loss are not allocated, please use ti.var(..., needs_grad=True)'
```

examples/ad_gravity.py
```diff
@@ -0,0 +1,44 @@
+import taichi as ti
+
```
Rullec, line 2: Just wonder: Dont we need to keep `ti.init` here anymore?

Rullec:
LGTM and few nits!
I don't know whether it's proper to add a reference in `differential_programming.rst`, guiding our reader to find this example.

examples/ad_gravity.py
archibate:
```suggestion
ti.init()

```
Yes, sometimes Taichi developments can break these examples without `ti.init()`.

archibate:
Great! Let's merge this to catch up with the next release. I'll try to complete the doc in AD and add a link to this example later.

PR #1531, opened 2020-07-18
[Refactor] 'ti.Matrix(n, m, dt, shape)' is deprecated, use 'ti.Matrix.var(n, m, dt, shape)' instead

archibate:
Related issue = #1500:2.2.2

----


python/taichi/lang/matrix.py
```diff
@@ -800,19 +772,71 @@ def identity(dt, n):
...
+        self.dt = dtype
+        assert not impl.inside_kernel()
+        for i in range(n * m):
+            self.entries.append(impl.field(dtype))
+        self.grad = self.make_grad()
+
+        if layout is not None:  # TODO(archibate): deprecate layout
+            assert shape is not None, 'layout is useless without shape'
+        if shape is None:
+            assert offset is None, f"shape cannot be None when offset is being set"
```
rexwangcc, line 805:
```suggestion
            assert offset is None, "shape cannot be None when offset is being set"
```

python/taichi/lang/matrix.py
```diff
@@ -800,19 +772,71 @@ def identity(dt, n):
...
         '''ti.Matrix.field'''
-        _taichi_skip_traceback = 1
-        # TODO: deprecate ad-hoc use ti.Matrix() as global (#1500:2.2/2)
-        return cls(n, m, dtype, *args, **kwargs)
+        self = cls.empty(n, m)
+        self.entries = []
+        self.n = n
+        self.m = m
+        self.dt = dtype
+        assert not impl.inside_kernel()
```
rexwangcc, line 797: nit: might be a bit better to move this line up so it stops eayly when assertion fails?

archibate: In fact, this can be removed since we are already using the decorator `@python_scope`, which do an assertion for us.
```suggestion
```

python/taichi/lang/matrix.py
```diff
@@ -76,57 +76,29 @@ def __init__(self,
...
-                assert offset is None, f"shape cannot be None when offset is being set"
+                # construct global matrix (deprecated)
+                #warning(  # TODO(archibate): uncomment this when #1500 is complete
+                #    "Declaring global matrices using `ti.Matrix(n, m, dt, shape)` is deprecated, "
+                #    "use `ti.Matrix.field(n, m, dt, shape)` instead",
+                #    DeprecationWarning,
+                #    stacklevel=2)
+                mat = Matrix.field(n=n,
+                                   m=m,
+                                   dtype=dt,
```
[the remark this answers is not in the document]
archibate: So here is you plan:
step 1: `ti.Matrix(dt)` -> `ti.Matrix(dtype)`.
step 2: `ti.Matrix(dtype)` -> `ti.Matrix.field(dtype)`.
And here is my plan:
directly: `ti.Matrix(dt)` ->  `ti.Matrix.field(dtype)` (two refactor done in once)

python/taichi/lang/matrix.py
```diff
@@ -76,57 +76,29 @@ def __init__(self,
...
-                        var_list.append(e)
-                    if needs_grad:
-                        for i, e in enumerate(self.entries):
-                            var_list.append(e.grad)
-                    ti.root.dense(ti.index_nd(dim),
-                                  shape).place(*tuple(var_list), offset=offset)
+                # create a local matrix with specific (n, m)
+                self.entries = [impl.expr_init(None) for i in range(n * m)]
+                self.n = n
+                self.m = m
```
k-ye, line 84: 1. Is `dt is None` case useful?
2a. If 1 is yes, note that the original logic still handles the layout/shape in this case (i.e. what you've put into `Matrix.field`). Yet the new logic won't do that anymore. Is this correct?
2b. Otherwise, shall we just enforce `dt` to never be `None`?

python/taichi/lang/matrix.py
```diff
@@ -800,19 +772,70 @@ def identity(dt, n):
...
         return Matrix([[ti.cos(alpha), -ti.sin(alpha)],
                        [ti.sin(alpha), ti.cos(alpha)]])
 
     @classmethod
     @python_scope
-    def field(cls, n, m, dtype, *args, **kwargs):
+    def field(cls,
+              n=1,
+              m=1,
+              dtype=None,
```
archibate, line 785: @k-ye's `2b`:
```suggestion
              n,
              m,
              dtype,
```

python/taichi/lang/matrix.py
archibate: 1. IMO `dt is None` is only used for local matrices.
2a. I think **the old logic is incorrect** at the first place @yuanming-hu, `ti.Matrix(n, m, dt=None, shape=233)`? How to allocate a matrix with `None` type ahead?
2b. SGTM, so we should use no default value `None` for dtype, see below.

k-ye: I see. So `ti.Matrix(3,4)` is a valid use case, i guess?

archibate:
> I see. So ti.Matrix(3,4) is a valid use case, i guess?

Yes, then `impl.expr_init(None)` will create 3x4 `AllocaStmt`.

PR #1536, opened 2020-07-19
[cc] Support "range for" and "while" statement on C backend

archibate:
Related issue = #1332

----

### DIFF
--- a/taichi/backends/cc/codegen_cc.cpp
+++ b/taichi/backends/cc/codegen_cc.cpp
@@ -279,13 +279,12 @@ class CCTransformer : public IRVisitor {
   }
 
   void generate_range_for_kernel(OffloadedStmt *stmt) {
-    //TI_ASSERT(stmt->const_begin && stmt->const_end);
+    // TI_ASSERT(stmt->const_begin && stmt->const_end);
     auto begin_value = stmt->begin_value;
     auto end_value = stmt->end_value;
     auto var = define_var("int", stmt->raw_name());
-    emit("for ({} = {}; {} < {}; {} += {}) {{",
-        var, begin_value, stmt->raw_name(), end_value,
-        stmt->raw_name(), 1 /* stmt->step? */);
+    emit("for ({} = {}; {} < {}; {} += {}) {{", var, begin_value,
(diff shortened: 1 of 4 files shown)


taichi/backends/cc/codegen_cc.cpp
```diff
@@ -273,10 +275,18 @@ class CCTransformer : public IRVisitor {
...
+    stmt->body->accept(this);
+  }
+
+  void generate_range_for_kernel(OffloadedStmt *stmt) {
+    // TI_ASSERT(stmt->const_begin && stmt->const_end);
+    auto begin_value = stmt->begin_value;
+    auto end_value = stmt->end_value;
+    auto var = define_var("int", stmt->raw_name());
+    emit("for ({} = {}; {} < {}; {} += {}) {{", var, begin_value,
+         stmt->raw_name(), end_value, stmt->raw_name(), 1 /* stmt->step? */);
```
k-ye, line 287: It seems to me that these are all the loop index?

```suggestion
         var, end_value, var, 1 /* stmt->step? */);
```

taichi/backends/cc/codegen_cc.cpp
```diff
@@ -287,13 +297,73 @@ class CCTransformer : public IRVisitor {
...
+      TI_NOT_IMPLEMENTED
+    }
+  }
+
+  void visit(RangeForStmt *stmt) override {
+    TI_ASSERT(stmt->width() == 1);
+    auto var = define_var("int", stmt->raw_name());
+    if (!stmt->reversed) {
+      emit("for ({} = {}; {} < {}; {} += {}) {{", var, stmt->begin->raw_name(),
+           stmt->raw_name(), stmt->end->raw_name(), stmt->raw_name(), 1);
```
k-ye, line 330: ditto

taichi/backends/cc/codegen_cc.cpp
archibate: No,
`var` is "int tmp233" when `C99_COMPAT = 0`.
`stmt->raw_name()` is "tmp233".

PR #1539, opened 2020-07-19
[Doc] Update all global tensor ti.Vector to ti.Vector.field

Rullec:
Related issue = #1500 

To keep a small change set, this PR simply **replaces all global tensor `ti.Vector` by `ti.Vector.field`.**

Though I try my best to replace all global tensors `ti.Vector`, there can still be some **slippery fish**. If you caught one, please tell me so that I can correct them.

In case of a waste of time, I want to double check these to-dos with you:

- [x] Do we need change `dt` to `dtype` in all `ti.Vector` and `ti.Vector.field`?
- [ ] In `vector.rst`, here are lots of vector operations. I guess they are also compatible to local vars so I kept them the same. If we need to do, please let me know.

----

### DIFF
--- a/docs/matrix.rst
+++ b/docs/matrix.rst
@@ -4,7 +4,7 @@ Matrices
 ========
 
 - ``ti.Matrix`` is for small matrices (e.g. `3x3`) only. If you have `64x64` matrices, you should consider using a 2D tensor of scalars.
-- ``ti.Vector.field`` is the same as ``ti.Matrix``, except that it has only one column.
+- ``ti.Vector`` is the same as ``ti.Matrix``, except that it has only one column.
 - Differentiate element-wise product ``*`` and matrix product ``@``.
 - ``ti.Vector.field(n, dt=ti.f32)`` or ``ti.Matrix(n, m, dt=ti.f32)`` to create tensors of vectors/matrices.
 - ``A.transpose()``
--- a/docs/matrix.rst
+++ b/docs/matrix.rst
@@ -4,9 +4,9 @@ Matrices
 ========
 
 - ``ti.Matrix`` is for small matrices (e.g. `3x3`) only. If you have `64x64` matrices, you should consider using a 2D tensor of scalars.
-- ``ti.Vector`` is the same as ``ti.Matrix``, except that it has only one column.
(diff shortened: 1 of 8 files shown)


Rullec:
@archibate request a review mannually and thanks in advance!

docs/matrix.rst
```diff
@@ -4,9 +4,9 @@ Matrices
 ========
 
 - ``ti.Matrix`` is for small matrices (e.g. `3x3`) only. If you have `64x64` matrices, you should consider using a 2D tensor of scalars.
-- ``ti.Vector`` is the same as ``ti.Matrix``, except that it has only one column.
+- ``ti.Vector.field`` is the same as ``ti.Matrix``, except that it has only one column.
```
archibate, line 7: r

docs/tensor_matrix.rst
```diff
@@ -36,7 +36,7 @@ Suppose you have a ``128 x 64`` tensor called ``A``, each element containing a `
 * If you want to get the matrix of grid node ``i, j``, please use ``mat = A[i, j]``. ``mat`` is simply a ``3 x 2`` matrix
 * To get the element on the first row and second column of that matrix, use ``mat[0, 1]`` or ``A[i, j][0, 1]``.
 * As you may have noticed, there are **two** indexing operators ``[]`` when you load an matrix element from a global tensor of matrices: the first is for tensor indexing, the second for matrix indexing.
-* ``ti.Vector`` is simply an alias of ``ti.Matrix``.
+* ``ti.Vector.field`` is simply an alias of ``ti.Matrix``.
```
archibate, line 39: r

archibate:
> Do we need change dt to dtype in all ti.Vector and ti.Vector.field?

Yes, thank in advance.

Rullec:
Thanks for your suggestions! Now revert has been done. I will update the left `dtype` part later.

Btw, what does "LGTMig" mean? I gets to know "LGTM", but yet the suffix "ig" is not very clear to me.Google also have no answer ;-)

archibate:
ig=in general
iapr=in another pr
ff2=feel free to
pvc=prevent conflict
Sorry, I invented these... hope they are benefits > cons...

Rullec:
Update: `dtype` has been replaced in this commit.

PR #1543, opened 2020-07-20
[OpenGL] [refactor] Ensure SSBO number <= 8 in most situations for better compatbility

archibate:
Related issue = depends #1540 depends #1542

----
@Eydcao IIUC you said that your computer only supports at most 8 SSBOs and therefore failed some test?
Before this PR, the test seems requires 11 SSBOs, and my mesa GL supports 12.
So I make some unused buffer undeclared to reduce down the maximum requirement down to 8, so that @Eydcao's machine can pass the test.

SSBO numbers on different situations:

minimal kernel:
root 2, runtime
3

normal kernel:
root 2, runtime, gtmp 2, args 2
7

struct-for kernel:
root 2, runtime, gtmp 2, args 2, listman
8

float64 struct-for kernel:
root 3, runtime, gtmp 3, args 3, listman
11 (but I guess users won't expect too much on float64 anyway...)

to/from_numpy kernel:
root 2, runtime, earg, extr 2
6

float64 to/from_numpy kernel:
root 3, runtime, earg, extr 3
8


Eydcao:
This is so helpful!

I could give a look. Thanks, genius. :>

Eydcao:
It is already working on my machine. When the PR is ready for review and merge, I could come back for the double-check :>

Great fix!

PR #1545, opened 2020-07-20
[bug] Update mpm_lagrangian_force and fix Matrix constructor

k-ye:
#1531 broke the example because it forgot to initialize `self.grad`. Adding the fix as well as updating `mpm_lagrangian_force.py` to use `field`.

Related issue = #1531

----

### DIFF
--- a/tests/python/test_matrix.py
+++ b/tests/python/test_matrix.py
@@ -186,3 +186,19 @@ def test_matrix_constant_index():
     func()
 
     assert np.allclose(m.to_numpy(), np.ones((5, 2, 2), np.int32) * 12)
+
+
+@ti.all_archs
+def test_matrix_needs_grad():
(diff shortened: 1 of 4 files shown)


archibate:
> #1531 broke the example because it forgot to initialize self.grad.

So we don't have a test on this?

tests/python/test_matrix.py
```diff
@@ -186,3 +186,19 @@ def func():
     func()
 
     assert np.allclose(m.to_numpy(), np.ones((5, 2, 2), np.int32) * 12)
+
+
+@ti.all_archs
+def test_matrix_needs_grad():
```
archibate, line 192: `test_scalar_needs_grad` as well?

k-ye: Done

PR #1552, opened 2020-07-21
[Doc] Replace ti.Matrix to ti.Matrix.field in the documents

Rullec:
Related issue = #1500 

Good news: this PR has a quite small changeset. 

I do the following stuff in it:
- [x] Replace all global tensor `ti.Matrix` to `ti.Matrix.field`, and keep the local allocation the same as before.
- [x] Replace `dt` by `dtype` for all `ti.Matrix`

----


Rullec:
Manually request review from @archibate 
Thanks for your suggestions in advance! ;-)

I want to begin:
1. replacing the allocating APIs accordly
2. checking all "tensor" text in our documens and do replacement.

in the **next 2 PRs** seperately. WDYT?

Rullec:
Thanks!

PR #1564, opened 2020-07-22
[Doc] Update new allocation APIs' definitions in the documents

Rullec:
Related issue = #1500 

In this issue: The definition of `ti.field`, `ti.Vector.field` and `ti.Matrix.field` are all maintained.

Also, I noticed that in `meta.rst` there are some new-added `ti.var`, so I replace them by `ti.field` as well. Hope this won't break the consistency.

----


Rullec:
@archibate mannually request the review, thanks in advance!

PR #1574, opened 2020-07-24
[OpenGL] [perf] Support TLS to improve reduction performance

archibate:
Related PR = #1336

----

Running on my poor Intel card:
fem99: 4.6 -> 40 (!!!!)
fem128: 18 -> 26
mpm_lagrangian_forces: 3.8 -> 15.7
misc/benchmark_reduction.py: `GL_OUT_OF_MEMORY` :(

No performance difference observed on the NV card, seems NV's compiler is already capable of doing TLS optimization for us?

(Partial) OFT:
[cli] [opengl] [metal] TI_MAKE_THREAD_LOCAL=0 to disable TLS on OpenGL/Metal.

### DIFF
--- a/taichi/backends/metal/codegen_metal.cpp
+++ b/taichi/backends/metal/codegen_metal.cpp
@@ -1232,7 +1232,7 @@ FunctionType CodeGen::compile() {
                               /*vectorize=*/false, kernel_->grad,
                               /*ad_use_stack=*/true, config.print_ir,
                               /*lower_global_access=*/true,
-                              /*make_thread_local=*/true);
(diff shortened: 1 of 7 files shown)


archibate:
> Thank you! LGTM. My only concerns is that `thls` seems to be used in place of `tls` in some places. Is that intentional?

Sorry, that's intentional. For now all the buffers, `root`, `args`, `gtmp`, ..., are 4 chars.
But I think it's OK to make `tls` to be 3 chars since it's **not a global buffer**, but a per-thread buffer. Thank for pointing out.

PR #1583, opened 2020-07-25
[Example] Add comet.py and fix to_numpy() on sparse matrix fields

archibate:
Related issue = #

----
video
Have you seen the NEOWISE comet yet :laughing:


examples/comet.py
```diff
@@ -0,0 +1,86 @@
...
+v = ti.Vector.field(dim, ti.f32)
+ti.root.bitmasked(ti.i, N).place(x, v, inv_m, color)
+xcnt = ti.field(ti.i32, ())
+img = ti.field(ti.f32, (res, res))
+
+
+@ti.kernel
+def substep():
+    ti.no_activate(x)
+    for i in x:
```
archibate, line 30: Can we automatically use `ti.no_activate(x)` when:
1. the following struct-for is on `x`.
2. within that struct-for, only `x[i]` is written.

e.g.:
```py
for i in x:
  x[i] = 1  # OK to optimize
  x[i // 2] = 1  # won't optimize
```

Thoughts? @xumingkuan @yuanming-hu

archibate:
Merging now given that this is a trivial change.

PR #1593, opened 2020-07-25
[async] Demote struct-fors in async compilation

k-ye:
1. I separated `compile_to_offloads()` into two functions: `compile_to_offloads()` and `compile_to_executable()`. The former stops right before struct-for demotion. This allows me to run async mpm99 without any patches now.
2. Note that after converting a `struct-for` to a `range-for`, we need to call `irpass::fix_root_block_kernel()` to fix the block kernels.

Related issue = #1541

----


taichi/backends/metal/codegen_metal.cpp
```diff
@@ -1228,11 +1228,11 @@ CodeGen::CodeGen(Kernel *kernel,
...
-  irpass::compile_to_offloads(kernel_->ir.get(), config,
-                              /*vectorize=*/false, kernel_->grad,
-                              /*ad_use_stack=*/true, config.print_ir,
-                              /*lower_global_access=*/true,
-                              /*make_thread_local=*/config.make_thread_local);
+  irpass::compile_to_executable(kernel_->ir.get(), config,
+                                /*vectorize=*/false, kernel_->grad,
+                                /*ad_use_stack=*/true, config.print_ir,
+                                /*lower_global_access=*/true,
+                                /*make_thread_local=*/config.make_thread_local));
```
archibate, line 1235:
```suggestion
                                /*make_thread_local=*/config.make_thread_local);
```

Issue #1601, opened 2020-07-27
[Discussion] Decouple compiler and runtime to support AOT & cross-compile?


archibate:
Continue our discussion in #1191.

**Concisely describe the proposed feature**
I'd like to decouple Taichi frontend from backends, i.e. decouple compiler from runtime.
This will enable commercial people to distribute their **compiled Taichi programs** in binary format (#920).
This will also enable AOT reduce the start up time when a Taichi program is huge (#1191).
This will also enable cross-compile which make it possible to run Taichi in non-Python runtime environment.

**Backgrounds**
The life of a Taichi program is:
1. Python `@ti.kernel`'s
V
2. Python frontend transformer (`lang/transformer.py`)
V
3. Taichi frontend IR in C++ (`frontend_ir.h`)
V
4. Taichi lowering & optimization passes (`compile_to_offloads.cpp`)
V
5. Taichi middle-end IR (`ir.h`)
V
6. Taichi backend codegens (`codegen_opengl.cpp`)
V
7. Backend sources (LLVM bitcode, GLSL, Metal, C)
V
8. Backend runtime (`opengl_api.cpp`)

**Current attempt**
Decouple at stage 7: 1~6 on compiler-side, 8 on runtime-side.
In codegen, we dump the kernel source code to external files (`kern.glsl`). Then, we distribute that backend specific source code.
We already have some progress on this direction in Metal and C backend using `ActionRecorder`.

Pros:
This can make the runtime-side as minimal as possible, very easy to embed into embed-systems like an ARM card.
Cons:
A Taichi program exported in GLSL is not able to run on Metal machines, i.e. exported Taichi programs are not portable.

**IMO attempt**
Decouple at stage 5: 1~4 on compiler-side, 6~8 on runtime-side.
Make Taichi IR portable, and export it into external files (`kern.ir`). Then, we distribute that binary Taichi IR.
On the runtime-side, we need to include the **backend codegens** that translate the Taichi IR read from `kern.ir` into backend codes, and then launch it.
This is a systematic solution for all backends.

Cons:
The runtime-side needs to contain a codegen, which is more heavy than the `Current attempt`.
Pros:
This can make exported Taichi programs portable too. Just like the Java-bitcode or SPIR-V, can run on any machine, no matter what machine it's compiled from.

**Additional comments**
Some refactor on compile_to_offloads may be applied to make cross-compiling IR possible.

k-ye:
[...]

PR #1602, opened 2020-07-27
[OpenGL] [perf] Support ti.block_dim as block size hint

archibate:
Related issue = #

----
Applying `ti.block_dim` to P2G & G2P for-loops in `mpm99.py` with `quality = 2`:
blk_dim: mesa, nvidia
128: 12.5 fps, 19.4 fps
1024: 11.1 fps, 18.9 fps
none: 10.9 fps, 18.4 fps
Very tiny performance difference that I had to run many times to confirm it...
But I guess it would be useful in some cases other than mpm, any idea?
@k-ye How does Metal decide block_dim now? Would you add `ti.block_dim` support to Metal later iapr?

### DIFF
--- a/tests/python/test_parallel_range_for.py
+++ b/tests/python/test_parallel_range_for.py
@@ -1,7 +1,8 @@
 import taichi as ti
 
 
-@ti.all_archs
+# such small block_dim will cause grid_dim too large for OpenGL...
+@ti.archs_excluding(ti.opengl)
 def test_parallel_range_for():
     n = 1024 * 1024
     val = ti.var(ti.i32, shape=(n))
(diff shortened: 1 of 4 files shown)


tests/python/test_parallel_range_for.py
```diff
@@ -1,7 +1,8 @@
 import taichi as ti
 
 
-@ti.all_archs
+# such small block_dim will cause grid_dim too large for OpenGL...
```
k-ye, line 4: IIRC, OpenGL supports grid stride loop, so is this still a problem?

archibate: You mean, when grid_dim is large, we 'share' some of the grid_dim into stride-loop-size? That would be non-trivial, since stride-loop-size is determined at compile-time, while grid_dim must be determined at runtime, for future cross-compile support #1601.

archibate: I think having `ti.stride_loop(n)` would be a good idea.

k-ye: I think grid stride loop is meant to be adaptive. For example, if we cap both `block_dim` and `grid_dim`, such that `block_dim * grid_dim < N` (`N` for number of items of an SNode), the kernel should still be able to handle all the `N` items?

PR #1606, opened 2020-07-28
[Error] Improve error message when using non-global variables in global operations

archibate:
Related issue = https://forum.taichi.graphics/t/runtimeerror-ir-cpp-get-attribute-293-attribute-dim-not-found/1045

The main reason for these confusing errors are originated by the fact that we use `Expr` for both local variables and global fields.

----
```py
x = 0
print(x[0])
```
Before:
`RuntimeError: [ir.cpp:get_attribute@293] Attribute dim not found`
After:
`TypeError: Cannot subscript a scalar expression`
```py
x = 0
for i in x:
```
Before:
`RuntimeError: [expr.cpp:snode@81] is<GlobalVariableExpression>()`
After:
`TypeError: can only iterate through global variables/fields`

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -61,7 +61,7 @@ def expr_init_func(rhs):  # temporary solution to allow passing in tensors as
 
 def begin_frontend_struct_for(group, loop_range):
     if not isinstance(loop_range, Expr) or not loop_range.is_global():
-        raise TypeError('can only iterate through global variables/fields')
+        raise TypeError('Can only iterate through global variables/fields')
     if group.size() != len(loop_range.shape):
         raise IndexError(
             'Number of struct-for indices does not match loop variable dimensionality '
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -60,6 +60,8 @@ def expr_init_func(rhs):  # temporary solution to allow passing in tensors as
 
 
 def begin_frontend_struct_for(group, loop_range):
+    if not isinstance(loop_range, Expr) or not loop_range.is_global():
+        raise TypeError('can only iterate through global variables/fields')
     if group.size() != len(loop_range.shape):
         raise IndexError(
             'Number of struct-for indices does not match loop variable dimensionality '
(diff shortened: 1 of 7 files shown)


taichi/common/core.h
```diff
@@ -141,7 +141,7 @@ static_assert(__cplusplus >= 201402L, "C++14 required.");
 #endif
 
 #define TI_STATIC_ASSERT(x) static_assert((x), #x);
-#define TI_ASSERT(x) TI_ASSERT_INFO((x), #x)
+#define TI_ASSERT(x) TI_ASSERT_INFO((x), "Assertion failure: " #x)
```
archibate, line 144: IMO this is more friendly than simply `#x`, e.g.:
```cpp
TI_ASSERT(p != nullptr);
```

Before:
```
RuntimeError: [expr.c@233] p != nullptr
```
This may make people think it's `p == nullptr` the expected value, while `p != nullptr` cause the error.

After:
```
RuntimeError: [expr.c@233] Assertion failure: p != nullptr
```

python/taichi/lang/impl.py
```diff
@@ -60,6 +60,8 @@ def expr_init_func(rhs):  # temporary solution to allow passing in tensors as
 
 
 def begin_frontend_struct_for(group, loop_range):
+    if not isinstance(loop_range, Expr) or not loop_range.is_global():
+        raise TypeError('can only iterate through global variables/fields')
```
xumingkuan, line 64:
```suggestion
        raise TypeError('Can only iterate through global variables/fields')
```

taichi/ir/expr.cpp
```diff
@@ -72,13 +72,15 @@ Expr &Expr::operator=(const Expr &o) {
 }
 
 Expr Expr::parent() const {
-  TI_ASSERT(is<GlobalVariableExpression>());
+  TI_ASSERT_INFO(is<GlobalVariableExpression>(),
+                 "cannot get snode parent of non-global variables.");
```
xumingkuan, line 76:
```suggestion
                 "Cannot get snode parent of non-global variables.");
```

taichi/ir/expr.cpp
```diff
@@ -72,13 +72,15 @@ Expr &Expr::operator=(const Expr &o) {
...
   return Expr::make<GlobalVariableExpression>(
       cast<GlobalVariableExpression>()->snode->parent);
 }
 
 SNode *Expr::snode() const {
-  TI_ASSERT(is<GlobalVariableExpression>());
+  TI_ASSERT_INFO(is<GlobalVariableExpression>(),
+                 "cannot get snode of non-global variables.");
```
xumingkuan, line 83:
```suggestion
                 "Cannot get snode of non-global variables.");
```

PR #1608, opened 2020-07-29
[async] Compute offloaded IR hash once and cache it

k-ye:
According to the profile, hash computation took 25.3% of the time in the main thread:

Screen Shot 2020-07-29 at 21 29 10

This PR caches the hash so that it is only done once. This is correct because: when we do fuse two kernels, the hash is re-computed on `task_a`, see https://github.com/taichi-dev/taichi/blob/2bfbca88b2d8cb1a070da9a40c5422c99b23fc2f/taichi/program/async_engine.cpp#L316

Because the major bottleneck is clone, the performance didn't improve much. FPS went from 5 -> 6 for `mpm88`.. Now the profile looked like this:

Screen Shot 2020-07-29 at 21 31 47

---

For cloning, I think we can do the similar thing, but it's a bit more difficult. We can clone it once at `launch()` as a template, but then when there is a fusion, we need to re-clone that task, so that it doesn't pollute the template.

Related issue = #742

----


archibate:
Btw, what `async_mode` is? What can I do for OpenGL?

taichi/program/async_engine.h
```diff
@@ -183,11 +179,26 @@ class AsyncEngine {
   void synchronize();
 
  private:
+  struct KernelMeta {
+    std::unique_ptr<Block> dummy_root;
+    std::vector<uint64> offloaded_hashes;
```
xumingkuan, line 184: This is a bit confusing to me: Why do offload hashes belong to the kernel meta? Is the same offload's hash different when the kernel is different?
I thought it should belong to the offloaded meta but the key of `offloaded_metas_` is the hash...

taichi/program/async_engine.cpp
```diff
@@ -130,25 +131,35 @@ void AsyncEngine::launch(Kernel *kernel) {
...
-    KernelLaunchRecord rec(
-        kernel->program.get_context(), kernel,
-        clone_offloaded_task(offload, kernel, dummy_root.get()));
+    auto cloned = clone_offloaded_task(offload, kernel, kmeta.dummy_root.get());
+    uint64 h;
+    if (kmeta_inited) {
+      h = kmeta.offloaded_hashes[i];
+    } else {
+      h = hash(cloned.get());
+      TI_ASSERT(kmeta.offloaded_hashes.size() == i);
```
xumingkuan, line 150: What if we have two same offloads in a kernel? And does it prevent two different offloads in two different kernels having the same hash?

xumingkuan: BTW if the hash really crashes, I think it should be an error rather than an assertion.

taichi/program/async_engine.h
k-ye: > Why do offload hashes belong to the kernel meta?

Each offloaded task in a kernel will generate a hash. What I'm trying to avoid is the re-computation of the hashes for the offloads inside a kernel...

> Is the same offload's hash different when the kernel is different?

I don't think so. From what I can tell, the `hash()` function computes the hash purely based on the input IR, not the kernel info. So same offloads will generate same hashes... 

However, this identical relationship is kind of shaky. The current impl of `hash()` uses `irpass::print()` to get a textual AST `std::string`, then computes the hash of that string. If the `print()` ever mixes in the kernel info (e.g. prints the kernel name), then same offloaded tasks would result in different hashes.

taichi/program/async_engine.cpp
k-ye: > BTW if the hash really crashes, I think it should be an error rather than an assertion.

Note that `offloaded_hashes` is a `std::vector`. It maps to an offloaded task by its *position* in that kernel... The assertion is checking for position (i.e. the hash of this offloaded task isn't computed yet), not hash collision.

> What if we have two same offloads in a kernel? And does it prevent two different offloads in two different kernels having the same hash?

I actually think it's fine for the same offloads (regardless of whether they are in the same kernel or not) to have the same hashes. If two offloads are indeed the same, then they should have the same `TaskMeta`s (i.e. input/output snodes, activation snodes)?

xumingkuan: Oh I see. I thought `offloaded_hashes` was an `std::unordered_map`, which would cause an assertion failure when there are two same offloads...

> I actually think it's fine for the same offloads to have the same hashes.

I agree.

k-ye:
> Btw, what `async_mode` is? What can I do for OpenGL?

Not for now. There are some necessary cleanups before this is useful.

PR #1611, opened 2020-07-30
[GUI] Support gui.fps_limit and reduce idle power consumption

archibate:
Related issue = #

----
Setting `gui.fps_limit = None` make `fractal.py` to be ~140 fps on CUDA!
Some tweaks in `system/timer.cpp` is also done for reducing CPU usage on idle (when actual FPS > our FPS limit).


taichi/system/timer.cpp
```diff
@@ -69,6 +69,24 @@ void Time::sleep(double s) {
   Time::usleep(s * 1e6_f64);
 }
```
JYLeeLYJ, line 70: Here is a mush more accurate milliseconds level sleep implementation instead of ```Sleep(DWORD(us * 1e-3))``` on windows . 
```
#ifdef _WIN64

#include <Windows.h>
#pragma comment(lib, "Winmm.lib")

void win_sleep(DWORD ms) {
  if (ms == 0)
    Sleep(0);
  else {
    HANDLE hEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
    timeSetEvent(ms, 1, (LPTIMECALLBACK)hEvent , 0, TIME_ONESHOT | TIME_CALLBACK_EVENT_SET);
    WaitForSingleObject(hEvent, INFINITE);
    CloseHandle(hEvent);
  }
}
#endif
```

taichi/system/timer.cpp
```diff
@@ -69,6 +69,24 @@ void Time::sleep(double s) {
...
+  double dt;
+  if (t < Time::get_time()) {
+    return;
+  }
+  do {  // use system-provided sleep for large scale sleeping:
+    dt = (t - Time::get_time()) * 0.5;
+    if (dt <= 0) {
+      return;
+    }
+    Time::sleep(dt * 1e-3);
```
JYLeeLYJ, line 82: We set the precision as 100us. But relatively , ```dt * 1e-3 ``` is too small . 
As a result , ```Sleep(0)``` is always called on windows , which will not give up the CPU time slice usually.
I think ```Time::sleep(dt)``` or ```Time::sleep(dt * 0.5)```  is enough.

taichi/system/timer.cpp
archibate: Sounds cool! But is that possible to sleep for microseconds (us) on Windows?

JYLeeLYJ: It's sad truth that sleep accuracy can not be guaranteed when delay time is lower than 1ms :(
The only possible approach is to use high resolution clock to continually counting in a for loop , but it 's unable to release CPU time slice ...

JYLeeLYJ: > Sounds cool! But is that possible to sleep for microseconds (us) on Windows?

It's sad truth that sleep accuracy can not be guaranteed when delay time is lower than 1ms :(
The only possible approach is to use high resolution clock to continually counting in a for loop , but it 's unable to release CPU time slice ...

JYLeeLYJ: Related problem : [is-there-a-windows-equivalent-of-nanosleep](https://stackoverflow.com/questions/7827062/is-there-a-windows-equivalent-of-nanosleep)

taichi/system/timer.cpp
archibate: Note that the unit of `(t - Time::get_time())` is seconds, while the unit of `Time::sleep` is milliseconds.

taichi/system/timer.cpp
archibate: Sleeping for us seems tricky on win, how about just:
```cpp
#ifndef _WIN64
} while (dt > 1e-4_f64);  // until dt <= 100us
#else
} while (dt > 1e-2_f64);  // until dt <= 10ms
#endif
```

JYLeeLYJ: 1ms is more accurate ( 1/60 = 16.6 ms  , 1/120 = 8.3ms , 1/40 = 25 ms )
···
#else
} while (dt > 1e-1_f64);  // until dt <= 1ms
#endif
···

taichi/system/timer.cpp
JYLeeLYJ: ```Time::sleep``` should be seconds .  See ```Time::usleep(s * 1e6_f64)``` .

archibate:
Hi, given that it's impossible to sleep for < 1ms on win, I simply abandoned reduce-idle-power-consumption there, WDYT about this change? Feel free to request change if you found a solution for this :)

taichi/system/timer.cpp
```diff
@@ -69,6 +69,27 @@ void Time::sleep(double s) {
...
+  double dt;
+  if (t < Time::get_time()) {
+    return;
+  }
+  do {  // use system-provided sleep for large scale sleeping:
+    dt = t - Time::get_time();
+    if (dt <= 0) {
+      return;
+    }
+    Time::sleep(dt * (dt < 4e-2_f64 ? 0.02 : 0.4));
```
JYLeeLYJ, line 84: Is this condition any different here ( using 0.02 or 0.4 )?
Why not just use :
```suggestion
    Time::sleep(dt * 0.4);
```

archibate: This ensures that the dt of sleep is more accurate when small; without this, I found the FPS to be around 59; with this, I found the FPS to be accurately 60.00.

archibate:
> In addition , for the frame rate accuracy , we just need to using that more accurate implementation win_sleep I suggested above (using WIN multimedia timer api and waitable event ), instead of Sleep which sometimes could lead to 1 ~ 10 ms error .
I can help u take an additional commit and test for this : )

That would be great! I invited you to `archibate/taichi` so that you could push to this branch, would you mind:
1. Implement a `win_sleep` of us precision in this PR?
2. In another PR and left `#ifndef _WIN64` in this PR?

JYLeeLYJ:
I have added ```win_usleep``` , ```win_msleep``` for different use cases on Windows .
None behavior changes on Linux. You can try to run on your machine to check if anything was different :)

taichi/system/timer.cpp
```diff
@@ -57,18 +57,82 @@ double Time::get_time() {
 }
 #endif
 
+#ifdef _WIN64
+#include <Windows.h>
+#pragma comment(lib, "Winmm.lib")
```
archibate, line 62: > I have added `win_usleep` , `win_msleep` for different use cases on Windows .
> None behavior changes on Linux. You can try to run on your machine to check if anything is different : )

Nice work! I can confirm that it's running happily on my machine!
One concern, what is `Winmm.lib`? Will it cause compat issue?
We'd better include it in `CMakeList.txt` instead of an inline `#pragma comment`, WDYT?
I'd like to provide more usage about linking libraries in CMake if you are not familiar to.

JYLeeLYJ: ```Winmm.lib``` is just a system multimedia API , which contains high resolution timer event functions .
```Winmm.dll``` can found in system path . And the minimum support client is winXP . This might not cause incompatibility problem.
See : [MSDN](https://docs.microsoft.com/en-us/previous-versions/dd757634(v=vs.85))

At the last commit  I have moved ```#pragma commit``` away and used ```target_link_library``` instead .

File: cmake/TaichiCore.cmake [hunk 5]
```diff
@@ -156,6 +156,9 @@ if (NOT WIN32)
         target_link_libraries(${CORE_LIBRARY_NAME} -Wl,--version-script,${CMAKE_CURRENT_SOURCE_DIR}/misc/linker.map)
         target_link_libraries(${CORE_LIBRARY_NAME} -Wl,--wrap=log2f) # Avoid glibc dependencies
     endif()
+else()
+    # windows
+    target_link_libraries(${CORE_LIBRARY_NAME} Winmm )
```
archibate, line 161:
```suggestion
    target_link_libraries(${CORE_LIBRARY_NAME} Winmm)
```

python/taichi/cc_compose.py
```diff
@@ -23,7 +23,7 @@ def run(self):
             func(e)
 
     def do_unknown(self, e):
-        self.emit(f"Unknown action type: {e['action']}")
+        pass
```
archibate, line 26:
```suggestion
        self.emit(f"Unknown action type: {e['action']}")
```

PR #1619, opened 2020-07-31
[async] Clone offloaded tasks lazily by caching the AST

k-ye:
To avoid having to clone the offloaded task upon every kernel launch, I've cached the cloned offloaded AST as well. The cached AST, along with its hash, is put inside a struct called `OffloadedCachedData`, and is used as a *read-only* template. This optimization works because for most of the times, we are just reading the offloaded task AST without any modification.

For places that do change the AST, the clone is done via a copy-on-write way. This clone is necessary so as to keep the template untouched. Please see `KernelLaunchRecord::clone_stmt_on_write()`.

For now, we only have two places that will modify the AST:

1. The first time an AST gets compiled in `compilation_workers`. Some IR passes will modify it.
2. When two ASTs are fused.

FPS improvement:

* `mpm88`: 6 -> 11
* `mpm99`: 12 -> 26

Profiling shows that clone is no longer a hotspot:

Screen Shot 2020-07-31 at 21 20 08

---

Another thing is that, I found that #1593 has actually prevented the fusion in the test cases in `test_fuse_dense.py` (probably `test_fuse_dynamic.py` as well, but i didn't test). This was because the kernel wasn't fully simplified, and contained an empty `serial` task:

```
kernel {
  $0 = offloaded  
  body {
  }
  $1 = offloaded struct_for(S1dense) grid_dim=0 block_dim=1024 bls= 
  body {
    <i32 x1> $2 = loop $1 index 0
    <i32*x1> $3 = global ptr [S2place_i32], index [$2] activate=false
    <i32 x1> $4 = global load $3
    <i32 x1> $5 = const [1]
    <i32 x1> $6 = add $4 $5
    <i32*x1> $7 : global store [$3 <- $6]
  }
}
```

I fixed this by adding another `fully_simplify()` pass in the end of `compile_to_offloads()`, but I wonder which specific pass can get rid of that empty offloaded..? @xumingkuan 

---

I'm not sure how much you like this template notion. Personally I think it's a not-very-intuitive but necessary workaround. Let me know your ideas :)

Related issue = #1608 

----

### DIFF
--- a/taichi/analysis/clone.cpp
+++ b/taichi/analysis/clone.cpp
@@ -123,9 +123,9 @@ class IRCloner : public IRVisitor {
     root->accept(&cloner);
 
     using namespace irpass;
-    typecheck(new_root.get());
     fix_block_parents(new_root.get());
     fix_root_block_kernel(new_root.get(), kernel);
+    typecheck(new_root.get());
     return new_root;
   }
 };
--- a/taichi/program/async_engine.cpp
+++ b/taichi/program/async_engine.cpp
@@ -27,39 +27,51 @@ uint64 hash(IRNode *stmt) {
   return ret;
 }
 
-std::unique_ptr<IRNode> clone_offloaded_task(OffloadedStmt *from,
-                                             Kernel *kernel,
-                                             Block *dummy_root) {
+std::unique_ptr<OffloadedStmt> clone_offloaded_task(OffloadedStmt *from,
+                                                    Kernel *kernel,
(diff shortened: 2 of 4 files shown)


xumingkuan:
> I wonder which specific pass can get rid of that empty offloaded..?

Quick answer:
https://github.com/taichi-dev/taichi/blob/6514e87f39d0748cf20fd22246808730c2c4c9dd/taichi/transforms/simplify.cpp#L791-L796

PR #1624, opened 2020-08-01
[GUI] Support images with 2 channels (RG) in GUI.set_image

archibate:
Related issue = #

----
This could be handy when visualizing 2D vector fields, e.g.: velocity field in `stable_fluid.py`.


Rullec:
LGTM!
Only curious: Is it a good way to define a literally "2D" image by only truncate the red and green channel? I'm curious about its performance.

archibate:
> LGTM!
> Only curious: Is it a good way to define a literally "2D" image by only truncate the red and green channel? I'm curious about its performance.

In fact we are always using RGBA as internal format storing in our GUI module, no matter grayscale, RG or RGB.

docs/gui.rst
```diff
@@ -92,6 +97,13 @@ Paint on a window
         ``img`` entries will be clipped into range ``[0, 1]`` for display.
 
 
+.. function:: gui.get_image()
+
+    :return: (np.array) the current image shown on the GUI
+
+    Get the RGBA shown image from the current GUI system which has four channels.
```
TroyZhai, line 104:
```suggestion
    Get the 4-channel (RGBA) image shown in the current GUI system.
```

examples/stable_fluid.py
```diff
@@ -8,11 +8,11 @@
...
 import time
 
-res = 600
+res = 512 #600
 dt = 0.03
-p_jacobi_iters = 30
+p_jacobi_iters = 100 #30
 f_strength = 10000.0
-curl_strength = 0
+curl_strength = 0 #3
```
TroyZhai, line 15: Just curious, is there a particular reason for these changes on the default parameters?

examples/stable_fluid.py
```diff
@@ -317,7 +317,7 @@ def reset():
         step(mouse_data)
 
     gui.set_image(dyes_pair.cur)
-    #gui.set_image(np.concatenate(
-    #        [np.abs(velocities_pair.cur.to_numpy() * 0.01 + 0.5),
-    #            np.zeros((res, res, 1))], axis=2))
+    #gui.set_image(velocities_pair.cur.to_numpy() * 0.01 + 0.5)
+    #divergence(velocities_pair.cur); gui.set_image(velocity_divs.to_numpy() * 0.1 + 0.5)
+    #vorticity(velocities_pair.cur); gui.set_image(velocity_curls.to_numpy() * 0.03 + 0.5)
```
TroyZhai, line 322: Are you keeping these for future use or do you expect the user to uncomment these lines? Might be a good idea to add some guidance in the comment. Also, doesn't the standard say that there should be a space after #?

archibate: Yes, I'm expecting them to uncomment one of these to see the different result, the space is **not left** purposely since it's code instead of human-text.

examples/stable_fluid.py
archibate: Setting to 3, for example, make curls strong for visual enhancement, but not really physics-based.

PR #1625, opened 2020-08-01
[GUI] Support gui.arrow() and gui.arrows()

archibate:
Related issue = #

----
This could be useful when visualizing a vector field, particle velocities and forces.
It's completely based on `gui.lines()`, so no C++ changes involved. It's just handy for users so they don't have to hand-write that much numpy ad-hoc operations on their own.

To test:
```py
import taichi as ti
import numpy as np

gui = ti.GUI()

orig = np.array([[0.5, 0.3], [0.3, 0.8]], dtype=np.float32)
dir = np.array([[-0.1, 0.4], [0.4, -0.2]], dtype=np.float32)

while gui.running and not gui.get_event(gui.ESCAPE):
    gui.arrows(orig, dir, radius=2, angle=25, tip_scale=0.4)
    #gui.arrow(orig[0], dir[0], radius=2)
    gui.show()
```

OFT: I assigned a default value `title='Taichi'` for `ti.GUI` since thinking about a title is often mind-taking, at least to me.

### DIFF
--- a/python/taichi/misc/gui.py
+++ b/python/taichi/misc/gui.py
@@ -34,7 +34,7 @@ class GUI:
     PRESS = ti_core.KeyEvent.EType.Press
     RELEASE = ti_core.KeyEvent.EType.Release
 
-    def __init__(self, name, res=512, background_color=0x0):
+    def __init__(self, name='Taichi', res=512, background_color=0x0):
         self.name = name
         if isinstance(res, numbers.Number):
             res = (res, res)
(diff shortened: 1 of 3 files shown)


python/taichi/misc/gui.py
```diff
@@ -286,6 +286,30 @@ def line(self, begin, end, radius=1, color=0xFFFFFF):
         self.canvas.path_single(begin[0], begin[1], end[0], end[1], color,
                                 radius)
 
+    @staticmethod
+    def _arrow_to_lines(orig, major, tip_scale=0.2, angle=45):
+        import math
+        angle = math.radians(180 - angle)
+        c, s = math.cos(angle), math.sin(angle)
+        minor1 = np.array([major[:, 0] * c - major[:, 1] * s,
+                           major[:, 0] * s + major[:, 1] * c]).swapaxes(0, 1)
```
k-ye, line 295: Is this equivalent to transpose?

archibate: `swapaxes`? Yes!

PR #1643, opened 2020-08-05
[cc] [workflow] Add C backend to Github CI

archibate:
Related issue = #1332

----


archibate:
Merging since:
All the CIs are passing.
I added the [`continue-on-error`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) option in case CC fails, it won't break CI completely.
This PR also emergency OFT fix `cc_compose.py`.

PR #1646, opened 2020-08-05
[Opt] Improve the binary ops simplify pass

xumingkuan:
Related issue = #1226 #656

- Do this optimization when `fast_math=False` and the operands are (return value is) of integral types.
- Add `(a - b) + c -> a - (b - c)`, `(a - b) - c -> a - (b + c)`, `(a / b) * c -> a / (b / c)`, `(a / b) / c -> a / (b * c)`.
- Disable `(a / b) * c -> a / (b / c)`, `(a * b) / c -> a * (b / c)` when the data type is integral.
- Let this pass return `true` when two operands are swapped but `modifier.modify_ir()` returns `false`.

Benchmark (#stmts): unchanged.

----


TH3CHARLie:
It's a nice improvement to the IR and optimization system.

>
> * Add `(a - b) + c -> a - (b - c)`, `(a - b) - c -> a - (b + c)`, `(a / b) * c -> a / (b / c)`, `(a / b) / c -> a / (b * c)`.
>

For the `(a / b) * c -> a / (b / c)` case, it can not be safely swapped due to the integer division roundoff. Consider a = 10, b = 3 and c = 3, the former produces 9 while the swapped one produces 10.

xumingkuan:
> It's a nice improvement to the IR and optimization system.
> 
> > * Add `(a - b) + c -> a - (b - c)`, `(a - b) - c -> a - (b + c)`, `(a / b) * c -> a / (b / c)`, `(a / b) / c -> a / (b * c)`.
> 
> For the `(a / b) * c -> a / (b / c)` case, it can not be safely swapped due to the integer division roundoff. Consider a = 10, b = 3 and c = 3, the former produces 9 while the swapped one produces 10.

Oh yes. I disabled this optimization when the return type is integral. And AFAIK, our `typecheck` guarantees that binary operations have the same type among operands and the return value.

PR #1658, opened 2020-08-07
[Misc] Add Windows Github CI

archibate:
Related issue = #1580

----


archibate:
Why GHCI windows doesn't have msbuild!!??

Issue #1483, opened 2020-07-13
IDLE still doesn't work for me


EaryChow:
**Describe the bug**
I saw this post today https://github.com/taichi-dev/taichi/pull/1222 and I realized this is already in 0.6.14 release, and I decided to try it out, but it still doesn't work for me.
**To Reproduce**
the codes are these:
```
import taichi as ti

ti.init()

a = 1


@ti.kernel
def test_kernel_print():
    print(a)


test_kernel_print()

```

**Log/Screenshots**
When I do it in IDLE concole, it still gives me this:
```
Traceback (most recent call last):
  File "<pyshell#6>", line 1, in <module>
    test_kernel_print()
  File "C:\Users\Eary\AppData\Local\Programs\Python\Python38\lib\site-packages\taichi\lang\kernel.py", line 559, in wrapped
    return primal(*args, **kwargs)
  File "C:\Users\Eary\AppData\Local\Programs\Python\Python38\lib\site-packages\taichi\lang\kernel.py", line 488, in __call__
    self.materialize(key=key, args=args, arg_features=arg_features)
  File "C:\Users\Eary\AppData\Local\Programs\Python\Python38\lib\site-packages\taichi\lang\kernel.py", line 300, in materialize
    src = remove_indent(oinspect.getsource(self.func))
  File "C:\Users\Eary\AppData\Local\Programs\Python\Python38\lib\inspect.py", line 985, in getsource
    lines, lnum = getsourcelines(object)
  File "C:\Users\Eary\AppData\Local\Programs\Python\Python38\lib\inspect.py", line 967, in getsourcelines
    lines, lnum = findsource(object)
  File "C:\Users\Eary\AppData\Local\Programs\Python\Python38\lib\inspect.py", line 798, in findsource
    raise OSError('could not get source code')
OSError: could not get source code
```
When I try it in IDLE with a file, it kind of runs but it prints nothing. 
```
[Taichi] mode=release
[Taichi] version 0.6.18, llvm 10.0.0, commit 35845482, python 3.8.1
[Taichi] Starting on arch=x64
>>> 
```

After reading the manual about https://taichi.readthedocs.io/en/stable/debugging.html?highlight=idle#run-time-print-in-kernels , I tried to change the `ti.init() ` to `ti.init(arch=ti.opengl)`
 but it doesn't work. 
```
[Taichi] mode=release
[Taichi] version 0.6.18, llvm 10.0.0, commit 35845482, python 3.8.1
[...]

PR #1664, opened 2020-08-08
[IPython] Use 'sourceinspect' to support Blender scripting module and IDLE shell

archibate:
Related issue = #1483 replace #1308

----
~Note that `sourceinspect` is not released on `PyPI` yet, you'll need to download it from https://github.com/archibate/sourceinspect for testing.~ Released `sourceinspect==0.0.1` just now!
Also note that `sourceinspect==0.0.1` won't work with Blender scripting module file mode on Windows for now, as it use `\\` as path separator.


archibate:
IDLE:
图片
Blender:
图片

python/taichi/lang/shell.py
```diff
@@ -1,123 +1,14 @@
...
-
+try:
+    import sourceinspect as oinspect
+except ImportError:
+    import taichi as ti
+    ti.warn('`sourceinspect` not installed!')
+    ti.warn('This may cause Taichi not functional in IDLE interactive shell, '
+            'Blender scripting module and Python native shell.')
+    ti.warn('Please run `python3 -m pip install sourceinspect` to install.')
+    import inspect as oinspect
```
isdanni, line 11: Thanks for adding this cool feature! 🚀  I only have few questions:
1. Why use `ti.warn` instead of `raise ImportError('any msgs...')` or directly printing the error plus the other suggestions?
2. Why add `import inspect as oinspect` in the exception block cuz I don't think this can really help with the error handling? 
3. Shall we add a `else` block to print sth like "Success!" ?

archibate: > 1. Why use ti.warn instead of raise ImportError('any msgs...') or directly printing the error plus the other suggestions?

In fact I did add `sourceinspect` to hard deps of `taichi` in `setup.py`. **Just in case** that `sourceinspect` didn't installed properly (somehow), users can still play with `taichi` via `inspect`, and a warning message.

> 2. Why add import inspect as oinspect in the exception block cuz I don't think this can really help with the error handling?

The required interface, `oinspect.getsource`, `oinspect.getsourcelines`, is exactly same among `sourceinspect` (ours) and `inspect` (Python builtin). So we can fallback to `inspect` for compatibility.

> 3. No, **ideally users will get `sourceinspect` installed properly thanks to `pip`'s dep resolver**, let's don't annoy them with any message.

isdanni: Ah I see. Thanks for the explanation. No more questions then!

misc/demo_import_time.py
```diff
@@ -0,0 +1,4 @@
+import time
+t = time.time()
+import taichi as ti
+print(time.time() - t)
```
yuanming-hu, line 4: Could you elaborate a bit on why including this file?

PR #1681, opened 2020-08-11
[test] Replace ti.var by ti.field in tests starting with a-g

Rullec:
Related issue = #1663 

This PR tries to replace all `ti.var` to `ti.field` in `tests/python/test_[a-g].*py`, except these four tests as shown below. Because they cannot be passed on my mac originally.
```
[error] error when testing test_bitmasked.py, return 256
[error] error when testing test_global_store_grad.py, return 1280
[error] error when testing test_ad_basics.py, return 256
[error] error when testing test_element_wise.py, return 256
```
**Is it a common case that some tests failed on my machine?**

Also, I ran the following script to verify that these test files are still valid:
```
import os
import re

files = os.listdir(".")
for name in files:
    cmd = f"pytest {name} 1>/tmp/delete/{name}.log 2>/tmp/delete/{name}.err"
    if re.search("^test_[a-g].*py$", name) is not None:
        print(f"[log] execute {cmd}")
        ret = os.system(cmd)
        if ret != 0:
            print(f"[error] error when testing {name}, return {ret}")
        else:
            print(f"[log] testing {name} succ")
```

### DIFF
--- a/tests/python/test_fill.py
+++ b/tests/python/test_fill.py
@@ -4,7 +4,6 @@ import taichi as ti
 @ti.all_archs
 def test_fill_scalar():
     val = ti.field(ti.i32)
-    
     n = 4
     m = 7
 
--- a/tests/python/test_fill.py
+++ b/tests/python/test_fill.py
@@ -4,6 +4,7 @@ import taichi as ti
 @ti.all_archs
 def test_fill_scalar():
     val = ti.field(ti.i32)
+    
     n = 4
     m = 7
 
--- a/tests/python/test_fill.py
+++ b/tests/python/test_fill.py
@@ -3,8 +3,7 @@ import taichi as ti
 
 @ti.all_archs
 def test_fill_scalar():
-    val = ti.var(ti.i32)
-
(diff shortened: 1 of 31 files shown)


Rullec:
I felt hard to make the PR title checker happy. 

What is extacly `github.event.pull_request.title` in `presubmit.yml`? I hope to find the exactly expression in order to get a correct PR title...

archibate:
Please do not use backquote \` in PR titie.

archibate:
@k-ye Could you help confirm if this is a local issue or a bug?

tests/python/test_fill.py
```diff
@@ -3,8 +3,8 @@
 
 @ti.all_archs
 def test_fill_scalar():
-    val = ti.var(ti.i32)
-
+    val = ti.field(ti.i32)
+
```
archibate, line 7:
```suggestion

```

archibate:
LGTMig! Please also update `ti.Matrix` and `ti.Matrix.var` respectively iapr.

PR #1693, opened 2020-08-13
[Perf] [opengl] Support 'ti.thread_dim(x)' to change invocations per thread for OpenGL

archibate:
Related issue = unrevert #1676

----

Now I realized I was totally misunderstood the term `stride`, I changed it to `invocation` now. Will do name nit in C++ iapr.

See #1691 for the new terms:

Taichi programs could be executed in parallel on GPU.
The level of structure of GPU is defined hierarchically.

From small to large, the computation units are:
**invocation** < **thread** < **block** < **grid**.

- **invocation**:
  Invocation is the **body of a for-loop**.
  Each invocation corresponding to a specific ``i`` value in for-loop.

- **thread**:
  Invocations are grouped into threads.
  Threads are the minimal unit that is parallelized.
  All invocations within a thread are executed in **serial**.
  We usually use 1 invocation per thread for maximizing parallel performance.

- **block**:
  Threads are grouped into blocks.
  All threads within a block are executed in **parallel**.
  Threads within the same block can share their **block local storage**.

- **grid**:
  Blocks are grouped into grids.
  Grid is the minimal unit that being **launched** from host.
  All blocks within a grid are executed in **parallel**.
  In Taichi, each **parallelized for-loop** is a grid.


archibate:
Closing, we need come to agree with a systematic solution for all these stuffs, block_dim, grid_dim, thread_dim, and make `DecoratorRecorder` more easy-to-add. FF2 put this into v0.8.0 roadmap.

PR #1697, opened 2020-08-14
[metal] Log errors if the compute kernel compilation failed

k-ye:
I guess it will be a lot easier to figure out the failure if we log the errors..

Related issue = #

----

### DIFF
--- a/taichi/platform/mac/objc_api.cpp
+++ b/taichi/platform/mac/objc_api.cpp
@@ -23,6 +23,12 @@ int ns_array_count(TI_NSArray *na) {
   return cast_call<int>(na, "count");
 }
 
+void ns_log_object(id obj) {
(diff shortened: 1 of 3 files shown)


taichi/platform/mac/objc_api.cpp
```diff
@@ -23,6 +23,12 @@ int ns_array_count(TI_NSArray *na) {
   return cast_call<int>(na, "count");
 }
 
+void ns_log_object(id obj) {
+  // https://github.com/halide/Halide/blob/bce3abe95cf3aef36c6eb8dafb8c85c197408b6d/src/runtime/objc_support.h#L88-L92
+  auto ns_str = wrap_string_as_ns_string("%@");
+  NSLog(reinterpret_cast<id>(ns_str.get()), obj);
```
archibate, line 29: Why not simply use `TI_ERROR`?

PR #1710, opened 2020-08-16
[example] Upgrade pbf2d.py to new syntax

archibate:
Related issue = WeChat (private)

----


examples/pbf2d.py
```diff
@@ -232,90 +215,60 @@ def compute_position_deltas():
...
-    find_particle_neighbors()
+    prologue()
     for _ in range(pbf_num_iters):
-        compute_lambdas()
-        compute_position_deltas()
-        apply_position_deltas()
-
-    confine_to_boundary()
-    update_velocities()
-    # no vorticity/xsph because we cannot do cross product in 2D...
```
archibate, line 274: We can do cross 2D now, so what should I do?

examples/pbf2d.py
```diff
@@ -46,21 +46,21 @@ def round_up(f, s):
...
+poly6_factor = 315.0 / 64.0 / math.pi
+spiky_grad_factor = -45.0 / math.pi
 
-old_positions = ti.Vector(dim, dt=ti.f32)
-positions = ti.Vector(dim, dt=ti.f32)
-velocities = ti.Vector(dim, dt=ti.f32)
+old_positions = ti.Vector.field(dim, float)
+positions = ti.Vector.field(dim, float)
+velocities = ti.Vector.field(dim, float)
 # Once taichi supports clear(), we can get rid of grid_num_particles
```
archibate, line 55: What do you mean by `clear()`?

Rullec: Maybe we need to ask the author of this example...

examples/pbf2d.py
k-ye: Nothing really. We need the direction, not the magnitude.

examples/pbf2d.py
```diff
@@ -194,7 +181,7 @@ def find_particle_neighbors():
 
 
 @ti.kernel
-def compute_lambdas():
+def substep():
```
k-ye, line 184: Actually, would you not merge them into a single kernel? I wrote the names to make the computation stages obvious.

examples/pbf2d.py
k-ye: I wanted to use `dynamic` initially, but `dense` seems to work OK. So I think we can remove this comment now.

```suggestion
```

examples/pbf2d.py
archibate: Why not add some comments per-offload instead? I just want to export them to [Taichi.js](https://taichi-dev.github.io/taichi.js) easier..

k-ye (approved):
Thank you for the help!

examples/pbf2d.py
k-ye: I see. Sorry i don't know much about js, is there a limitation for how many kernels we can have there?

archibate: Just hard to seek out the launch order..

PR #1712, opened 2020-08-16
[example] [refactor] Replace all ti.var->ti.field in our examples

Rullec:
Related issue = #1500 

This PR tries to **ONLY** replace all `ti.var` to `ti.field` in `examples`.

# Validation
I validate these changes by running:
```
import os
from multiprocessing import Pool

def run_py(file):
    cmd = f"python {file} 1>>sublog 2>>suberr &"
    print(cmd)
    os.popen(cmd)
    
modified_files = os.popen("git status | grep -i modified | awk '{print $2}'").read().strip().split("\n")
print(modified_files)
pool = Pool(8)
pool.map(run_py, modified_files)
```
Then check the log mannually
```
cat sub*  | grep -i "error\|fail\|warn"
```

What I got is like 
```
FileNotFoundError: [Errno 2] No such file or directory: 'mnist.pkl'
FileNotFoundError: mnist.pkl not found, please run examples/mnist_download_data.py first.
ModuleNotFoundError: No module named 'taichi_glsl'
EOFError: EOF when reading a line
RuntimeError: Window close button clicked, exiting... (use `while gui.running` to exit gracefully)
RuntimeError: Window close button clicked, exiting... (use `while gui.running` to exit gracefully)
RuntimeError: Window close button clicked, exiting... (use `while gui.running` to exit gracefully)
```

I resolved these errors by:
1.  running  `mnist_download_data`
2. `python -m pip install taichi_glsl`
3. rerun the example `rever_image_data.py`

**Then these errors can be fixed.**

After that, maybe I can confirm that **my changes is not harmful**?

----


archibate:
Nice script for testing examples! Does this test script need to shows up a GUI and manually close it? Do you verify the visual result manually? I think we can somehow make examples to be **tested** on CI in a later PR. And add a manual script for testing visual result in `misc/`.

Rullec:
> Nice work in general! Btw, would you like to advocate the `ti.field(dtype=float)` syntax in our examples iapr?

Good idea, I guess It can make our syntex more unified and readable!

> Nice script for testing examples! Does this test script need to shows up a GUI and manually close it? Do you verify the visual result manually?


Thanks for your confirmation! 

Yes, because `examples/` are more diverse than `tests/`, only when we saw the result can we confirm their correcteness. Opening a GUI, examining a GUI and close them manually is essential in this point of view.

> I think we can somehow make examples to be **tested** on CI in a later PR. And add a manual script for testing visual result in `misc/`.

Sounds good! Here is one of my possible solution:
1. Make each examples **output or return a result checksum** after running a speicified time of period.
2. **`CI` call a testing script in `misc/test_examples.py` which can check the result checksum w.r.t each example**, compare them with a predefined value in order to validate the correctness.
3. whether `CI` is passed would be determined by the return of `test_examples.py`

Do your like this idea? 

If you have a better / eaiser idea, please share it with us!

Btw, what does `misc` (literally `miscellaneous`) stand for? Can it contain everything?

PR #1716, opened 2020-08-17
[example] [refactor] Replace global field ti.Vector -> ti.Vector.field in examples

Rullec:
Related issue = #1500 

This PR tries to replace all global field(tensor) `ti.Vector` to `ti.Vector.field`

# Validation
1. Script testing is the same as #1712 
2. Then I watched the examples' result and played with them manually. They look good to me.

----

### DIFF
--- a/examples/cornell_box.py
+++ b/examples/cornell_box.py
@@ -6,7 +6,7 @@ from renderer_utils import ray_aabb_intersection, intersect_sphere, ray_plane_in
 
 ti.init(arch=ti.gpu)
 res = (800, 800)
-color_buffer = ti.Vector(3, dt=ti.f32, shape=res)
+color_buffer = ti.Vector.field(3, dtype=ti.f32, shape=res)
 count_var = ti.field(ti.i32, shape=(1, ))
 
 max_ray_depth = 10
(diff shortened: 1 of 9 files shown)


Rullec:
I posted some thinking about `CI` on `examples` in #1712 , But cuz I didn't noticed that the old PR #1712  has been merged in the morning, please let me post it again here:

***
# Old post in #1712 
> Nice work in general! Btw, would you like to advocate the `ti.field(dtype=float)` syntax in our examples iapr?

Good idea, I guess It can make our syntex more unified and readable!

> Nice script for testing examples! Does this test script need to shows up a GUI and manually close it? Do you verify the visual result manually?


Thanks for your confirmation! 

Yes, because `examples/` are more diverse than `tests/`, only when we saw the result can we confirm their correcteness. Opening a GUI, examining a GUI and close them manually is essential in this point of view.

> I think we can somehow make examples to be **tested** on CI in a later PR. And add a manual script for testing visual result in `misc/`.

Sounds good! Here is one of my possible solution:
1. Make each examples **output or return a result checksum** after running a speicified time of period.
2. **`CI` call a testing script in `misc/test_examples.py` which can check the result checksum w.r.t each example**, compare them with a predefined value in order to validate the correctness.
3. whether `CI` is passed would be determined by the return of `test_examples.py`

Do your like this idea? 

If you have a better / eaiser idea, please share it with us!

Btw, what does `misc` (literally `miscellaneous`) stand for? Can it contain everything?

archibate:
> Btw, what does misc (literally miscellaneous) stand for? Can it contain everything?

Yes, it literally can.

> Make each examples output or return a result checksum

The checksum method SGTM! Would you open an issue and demonstrate your planned workflow?

> after running a speicified time of period.

Thanks, we'd better run a **specified number of frames** instead, otherwise the checksum is dependent of FPS (machine performance).

PR #1724, opened 2020-08-18
[opengl] [refactor] Use result_buffer for return value

archibate:
Related issue = #1722

----
@k-ye Is this what you want? Note that I don't make a `GLBufId::Resu` but still reusing `GLBufId::Args` is because some OpenGL 'implementations' have **limited number of SSBOs**, the limit is 8 on @Eydcao's machine for example. Not to say we need `args_i32` and `args_f32` for simulating union. Not to say my ultra-secret [数据删除] project with @yuanming-hu. We actually can't turn on all 7 SSBOs at the same time on these machines at all.

### DIFF
--- a/taichi/backends/opengl/opengl_api.cpp
+++ b/taichi/backends/opengl/opengl_api.cpp
@@ -265,11 +265,15 @@ struct GLBuffer : GLSSBO {
   }
 
   void copy_back() {
+    copy_back(base);
+  }
+
+  void copy_back(void *ptr) {
     if (!size)
       return;
     void *mapped = this->map();
     TI_ASSERT(mapped);
-    std::memcpy(base, mapped, size);
+    std::memcpy(ptr, mapped, size);
     this->unmap();
   }
 };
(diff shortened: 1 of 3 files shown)


k-ye (approved):
Thank you, LGTM.

> Note that I don't make a GLBufId::Resu but still reusing GLBufId::Args is because some OpenGL 'implementations' have limited number of SSBOs, the limit is 8 on @Eydcao's machine for example.

Ack. I think this is fine because how OpenGL uses the buffer to pass value back and forth is a backend implementation detail. But copying the result back to `Program::result_buffer` means the host side can handle all the backends uniformly.

> Not to say we need args_i32 and args_f32 for simulating union. Not to say my ultra-secret [数据删除] project with @yuanming-hu. 

I don't get your point. Is switching to `result_buffer` causing other issues?

archibate:
> copying the result back to Program::result_buffer means the host side can handle all the backends uniformly.

So how do we pass args after your plan?

> result_buffer causing other issues?

TBH I'm maintaining two copy of `opengl_api.cpp`, in different langurages..

Will open another one for CC soon.

PR #1731, opened 2020-08-19
[Example] Make ODOP MGPCG class in mgpcg.py generic and reuse it in stable_fluid.py

archibate:
Related issue = #1713

----
图片


examples/mgpcg_advanced.py
```diff
@@ -1,41 +1,55 @@
...
+
+    $\\nabla^2 x = k r$
+
+Credits by `@KLozes <https://github.com/KLozes>`_.
+
+See `examples/stable_fluid.py <https://github.com/taichi-dev/taichi/blob/master/examples/stable_fluid.py>`_ for usage example.
+
+.. note::
+
+    This solver only runs on CPU and CUDA backend since it requires the ``pointer`` SNode. Consider replace ``pointer`` by ``bitmasked`` on Metal, by ``dense`` on OpenGL.
```
xumingkuan, line 19: nit: Consider separating these notes to several lines to make each line contain not too many characters?

xumingkuan (commented):
LGTMig but I'm not familiar with MGPCG :)

examples/stable_fluid.py
```diff
@@ -8,6 +8,7 @@
 import numpy as np
 import time
 
+use_mgpcg = True  # True for zero divergence fluid simulation
```
archibate, line 11:
```suggestion
use_mgpcg = False  # True for zero divergence fluid simulation
```

examples/mgpcg_advanced.py
```diff
@@ -138,24 +178,24 @@ def apply_preconditioner(self):
...
-    def paint(self):
-        if ti.static(self.dim == 3):
-            kk = self.N_tot * 3 // 8
-            for i, j in self.pixels:
-                ii = int(i * self.N / self.N_gui) + self.N_ext
-                jj = int(j * self.N / self.N_gui) + self.N_ext
-                self.pixels[i, j] = self.x[ii, jj, kk] / self.N_tot
+    def solve(self, steps=-1, eps=1e-12, tol=1e-12, rel=1e-12):
+        '''
+        Solve the possion equation.
```
yuanming-hu, line 183:
```suggestion
        Solve a Poisson problem.
```

examples/mgpcg_advanced.py
```diff
@@ -197,12 +237,48 @@ def run(self):
...
             self.reduce(self.z[0], self.r[0])
             new_zTr = self.sum[None]
-            self.beta[None] = new_zTr / old_zTr
+            self.beta[None] = new_zTr / (old_zTr + eps)
 
             # self.p = self.z + self.beta self.p
             self.update_p()
             old_zTr = new_zTr
 
+            yield rTr
```
yuanming-hu, line 246: Usually people just iterate until the residual is below a threshold. No need for a generator here.

examples/mgpcg_advanced.py
```diff
@@ -49,23 +64,48 @@ def __init__(self):
...
+            self.init_r(I, r[I] * k)
+
+    @ti.kernel
+    def get_result(self, x: ti.template()):
+        '''
+        Get the solution field after each solve step.
+
+        :parameter x: (ti.field) Specify where to store the solution field
+
+        $\\nabla^2 x = k r$
```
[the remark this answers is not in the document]
archibate: Do you mean:
```suggestion

        $\nabla^2 x = k r$
```
?

examples/mgpcg_advanced.py
```diff
@@ -49,23 +64,48 @@ def __init__(self):
...
+        '''
+        for I in ti.grouped(ti.ndrange(*[self.N] * self.dim)):
+            x[I] = self.get_x(I)
 
     @ti.func
     def neighbor_sum(self, x, I):
-        ret = 0.0
+        ret = x[I] * 0
```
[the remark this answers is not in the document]
archibate:
```suggestion
        ret = 0.0
```

examples/mgpcg_advanced.py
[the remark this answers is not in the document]
yuanming-hu: Sorry about the confusion - what does `k r` mean here?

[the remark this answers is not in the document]
archibate: See `def init(self, r: ti.template(), k: ti.template())`, r is the divergence field while k is the scale, usually 0.25 or 1.

examples/mgpcg_advanced.py
```diff
@@ -1,41 +1,56 @@
...
 @ti.data_oriented
 class MGPCG:
-    def __init__(self):
+    '''
+Grid-based MGPCG solver for the possion equation:
+
+    $\\nabla^2 x = k r$
+
+Credits by `@KLozes <https://github.com/KLozes>`_.
+
```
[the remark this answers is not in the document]
archibate: It would be great to let people know we respect creadits..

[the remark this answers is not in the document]
yuanming-hu: That's true at a high level, but for this script, credits are hard to assign. Better skip it instead..

examples/stable_fluid.py
```diff
@@ -8,6 +8,7 @@
 import numpy as np
 import time
 
+use_mgpcg = False  # True for zero divergence fluid simulation
```
yuanming-hu, line 11:
```suggestion
use_mgpcg = False  # True to use multigrid-preconditioned conjugate gradients
```

examples/mgpcg_advanced.py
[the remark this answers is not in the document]
yuanming-hu: I'm still a bit confused - usually, we don't pass `k` to MGPCG.

examples/mgpcg_advanced.py
```diff
@@ -138,24 +177,24 @@ def apply_preconditioner(self):
...
-    def run(self):
-        gui = ti.GUI("Multigrid Preconditioned Conjugate Gradients",
-                     res=(self.N_gui, self.N_gui))
+        :parameter steps: Specify the maximal steps, -1 for no limit.
+        :parameter eps: Specify a non-zero value to prevent ZeroDivisionError.
+        :parameter tol: Specify the absolute tolerance of loss.
+        :parameter rel: Specify the tolerance of loss relative to initial loss.
 
-        self.init()
+        :return: (generator) A sequence generator for the loss of each step.
```
yuanming-hu, line 190: please update this line?

examples/mgpcg_advanced.py
```diff
@@ -138,24 +177,24 @@ def apply_preconditioner(self):
...
 
-    @ti.kernel
-    def paint(self):
-        if ti.static(self.dim == 3):
-            kk = self.N_tot * 3 // 8
-            for i, j in self.pixels:
-                ii = int(i * self.N / self.N_gui) + self.N_ext
-                jj = int(j * self.N / self.N_gui) + self.N_ext
-                self.pixels[i, j] = self.x[ii, jj, kk] / self.N_tot
+    def solve(self, steps=-1, eps=1e-12, tol=1e-12, rel=1e-12):
```
yuanming-hu, line 180: `step` should probably be renamed to `max_iters`. `tol` and `rel` seem to be `abs_tolerance` and `rel_tolerance`? Please rename accordingly.

examples/mgpcg_advanced.py
```diff
@@ -138,24 +177,24 @@ def apply_preconditioner(self):
...
+        '''
+        Solve a Poisson problem.
+        Always use a ``for`` loop to iterate through me.
```
yuanming-hu, line 183: Update here?

yuanming-hu (commented):
Cool! Thanks for the updates. I think we are very close. Just a few minor naming/convention issues.

examples/mgpcg_advanced.py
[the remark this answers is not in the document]
archibate: But it's actually required for `stable_fluid.py`.. It has to be -1 for solving $v = -\nabla p$, i.e. $\nabla^2 p = -\nabla v$, where $p$ is x, and $\nabla v$ is r.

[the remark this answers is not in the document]
yuanming-hu: In that case, the right place for the equation is the `init` kernel. Here we should only doc "get the solution".

examples/mgpcg_advanced.py
yuanming-hu: Do we still need a `for` loop here?

examples/mgpcg_advanced.py
```diff
@@ -138,24 +177,22 @@ def apply_preconditioner(self):
...
 
-    def run(self):
-        gui = ti.GUI("Multigrid Preconditioned Conjugate Gradients",
-                     res=(self.N_gui, self.N_gui))
-
-        self.init()
+        :parameter max_iters: Specify the maximal iterations, -1 for no limit.
```
yuanming-hu, line 185:
```suggestion
        :parameter max_iters: Specify the maximal iterations. -1 for no limit.
```

examples/mgpcg_advanced.py
```diff
@@ -169,12 +206,12 @@ def run(self):
         old_zTr = self.sum[None]
 
         # CG
-        for i in range(400):
+        while max_iters != 0:
```
yuanming-hu, line 209: Is there a good reason to use a `while` loop here instead of `for iter in range(max_iters):` ?

examples/mgpcg_advanced.py
```diff
@@ -49,19 +63,44 @@ def __init__(self):
...
+
+    @ti.func
+    def get_x(self, I):
+        I = I + self.N_ext
+        return self.x[I]
+
+    @ti.kernel
+    def get_result(self, x: ti.template()):
+        '''
+        Get the solution field after each solve step.
```
yuanming-hu, line 96:
```suggestion
        Get the solution field.
```

examples/mgpcg_advanced.py
```diff
@@ -49,19 +63,44 @@ def __init__(self):
...
-            self.x[I] = 0.0
+    def init(self, r: ti.template(), k: ti.template()):
+        '''
+        Specify quantity field before each solve step.
+
+        :parameter r: (ti.field) Specify the quantity field
+        :parameter k: (scalar) Specify the scale multiply to the quantity field
+
+        $\\nabla^2 x = k r$
+        '''
```
yuanming-hu, line 84:
```suggestion
        '''
        Set up the solver for $\nabla^2 x = k r$, a scaled Poisson problem.

        :parameter k: (scalar) A scaling factor of the right-hand side.
        :parameter r: (ti.field) Unscaled right-hand side.
        '''
```

yuanming-hu (commented):
Thanks for the updates. I think we are very close! This PR can be merged once the issues are addressed.

examples/mgpcg_advanced.py
archibate: `max_iter = -1` for no limit.

examples/mgpcg_advanced.py
```diff
@@ -169,12 +200,12 @@ def run(self):
         old_zTr = self.sum[None]
 
         # CG
-        for i in range(400):
+        while max_iters != 0:
```
yuanming-hu, line 203:
```suggestion
        # We use a while loop here since max_iter = -1 means no limit on the number of iterations
        while max_iters != 0:
```

yuanming-hu: OK that makes sense. Let's add a line of comment for clarity. Thanks!

yuanming-hu (commented):
Again, I believe we are very close. This can be merged once
 - the two unresolved issues (on credits and on `float`) are resolved;
 - the PR is rebased on the latest master and we confirm everything still works.

Thanks for your patience and contribution.

examples/mgpcg_advanced.py
```diff
@@ -1,41 +1,56 @@
...
         self.bottom_smoothing = 50
-        self.dim = 3
+        self.dim = dim
 
         self.N_ext = self.N // 2  # number of ext cells set so that that total grid size is still power of 2
         self.N_tot = 2 * self.N
 
         # setup sparse simulation data arrays
-        self.r = [ti.field(dtype=real)
+        self.r = [ti.field(dtype=float)
```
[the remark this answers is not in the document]
archibate: I don't think it's a confusion at all. Python use 64bit types because they use CPU only. However it's not supported very well for OpenGL and Metal. And the 32bit types can reduce memory overhead. It's a good choice for compute graphics which doesn't need too much percsion.

yuanming-hu (approved):
Thank you so much! LGTM now. I'll take a look at MGPCG performance recently.

PR #1732, opened 2020-08-19
[Example] Upgrade mass_spring_3d.py to Taichi THREE v0.0.3 API

archibate:
Related issue = https://github.com/taichi-dev/taichi_three

----
Thank @victoriacity and @DSaurus for their contribution, we can now control the camera in better ways in Taichi THREE:
1. LMB for orbiting the target.
2. RMB or mouse wheel for scaling distance to target.
3. MMB for moving target position.

Since API breaking change is made in v0.0.2 -> v0.0.3, so I'd like to update the mass_spring_3d.py in Taichi main repo too.

### DIFF
--- a/examples/mass_spring_3d.py
+++ b/examples/mass_spring_3d.py
@@ -1,4 +1,5 @@
 import taichi as ti
+import numpy as np
 try:
     import taichi_glsl as tl
     import taichi_three as t3
@@ -6,9 +7,7 @@ except ImportError:
     print('This example needs the extension library Taichi THREE to work.'
           'Please run `pip install --user taichi_three` to install it.')
 
-import numpy as np
-import math
-ti.init(ti.gpu)
+ti.init(arch=ti.gpu)
 
 ### Parameters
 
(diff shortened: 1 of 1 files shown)


examples/mass_spring_3d.py
```diff
@@ -1,18 +1,18 @@
 import taichi as ti
+import numpy as np
 try:
     import taichi_glsl as tl
     import taichi_three as t3
 except ImportError:
     print('This example needs the extension library Taichi THREE to work.'
           'Please run `pip install --user taichi_three` to install it.')
+assert t3.__version__ == (0, 0, 3)
```
Rullec, line 9: Why do we need to assert the version of `taichi_three` here?

If the `taichi_three` updated this example maybe will be inavaliable ;)

Rullec:
Thanks for this nice work! LGTM in general!

examples/mass_spring_3d.py
archibate: I guess Taichi THREE v0.0.4 will break the API again, if not, we're OK to remove this before v0.0.4 release.

PR #1733, opened 2020-08-19
[blender] [example] Fix examples/waterwave.py memory leakage in Blender

archibate:
Related issue = https://github.com/taichi-dev/taichi/issues/1483#issuecomment-674958231

----
@EaryChow Does removing `exit()` fix the issue?
In fact, simply
```py
exit()
```
can cause Blender to crash.. see https://developer.blender.org/T44127.

### DIFF
--- a/examples/waterwave.py
+++ b/examples/waterwave.py
@@ -11,11 +11,11 @@ depth = 4
 dx = 0.02
 dt = 0.01
 shape = 512, 512
-pixels = ti.field(dtype=ti.f32, shape=shape)
-background = ti.field(dtype=ti.f32, shape=shape)
-height = ti.field(dtype=ti.f32, shape=shape)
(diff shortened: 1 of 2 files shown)


EaryChow:
>Does removing exit() fix the issue?

Yes, that seems to be the case. I also tried to Ctrl C in the terminal to forcefully interrupt it, then it stopped running, and it did not quit.
I have a guess... Could `exit()` actually be the code to quit Blender? Instead of quitting the script, `exit()` actually quit Blender entirely?

archibate:
> Could exit() actually be the code to quit Blender?

I think so, quit with a confusing error message.

EaryChow:
This is kind of a problem... It means all files that write `exit()` would need to be modified for Blender to run it... (ー_ー)

PR #1735, opened 2020-08-19
[GUI] Add handy ti.imscale(img, w, h) for scaling image

archibate:
Related issue = #

----
When I was debugging MGPCG, I realized that a small resolution is fast but hardly-visible. So it would be great if I could scale that small image up.

### DIFF
--- a/docs/gui.rst
+++ b/docs/gui.rst
@@ -522,3 +522,14 @@ Image I/O
     This function will create an instance of ``ti.GUI`` and show the input image on the screen.
 
     It has the same logic as ``ti.imwrite`` for different datatypes.
+
+.. function:: ti.imscale(img, w=512, h=None):
+
+    :parameter img: (np.array or ti.field) the input image.
+    :parameter w: (optional, scalar) the width after scale.
+    :parameter h: (optional, scalar) the height after scale.
+    :return: (np.array) the output image after scale.
+
+    If ``h`` is not specified, then it will be equal to ``w`` by default.
+
+    The output image shape is: ``(w, h, *img.shape[2:])``.
--- a/python/taichi/misc/image.py
+++ b/python/taichi/misc/image.py
@@ -34,7 +34,11 @@ def imdisplay(img):
     """
     Try to display image in interactive shell.
     """
-    if ti.lang.shell.oinspect.name == ti.lang.shell.ShellType.JUPYTER:
+    try:
+        get_ipython()
+    except:
+        ti.imshow(img)
+    else:
         import PIL.Image
         from io import BytesIO
         import IPython.display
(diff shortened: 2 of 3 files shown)


docs/gui.rst
```diff
@@ -522,3 +522,14 @@ Image I/O
     This function will create an instance of ``ti.GUI`` and show the input image on the screen.
 
     It has the same logic as ``ti.imwrite`` for different datatypes.
+
+.. function:: ti.imscale(img, w=512, h=None):
+
+    :parameter img: (np.array or ti.field) the input image.
+    :parameter w: (optional, scalar) the width after scale.
+    :parameter h: (optional, scalar) the height after scale.
```
Rullec, line 530:
```suggestion
    :parameter w: (optional, int) the width after scale.
    :parameter h: (optional, int) the height after scale.
```
Maybe we can enforce that the resolution if images should be integrator here. HDYT?

docs/gui.rst
```diff
@@ -522,3 +522,14 @@ Image I/O
     This function will create an instance of ``ti.GUI`` and show the input image on the screen.
 
     It has the same logic as ``ti.imwrite`` for different datatypes.
+
+.. function:: ti.imscale(img, w=512, h=None):
```
Rullec, line 526:
```suggestion
.. function:: ti.imscale(img, w=512, h=512):
```
I would suggest to use a meaningful number in the parameter `h`, because it can guide our users better.

docs/gui.rst
```diff
@@ -522,3 +522,14 @@ Image I/O
...
+
+    :parameter img: (np.array or ti.field) the input image.
+    :parameter w: (optional, scalar) the width after scale.
+    :parameter h: (optional, scalar) the height after scale.
+    :return: (np.array) the output image after scale.
```
Rullec, line 531:
```suggestion
    :return: (np.array) the output image after scaling.
```

python/taichi/misc/image.py
```diff
@@ -43,8 +47,20 @@ def imdisplay(img):
...
+
+def imscale(img, w=512, h=None):
+    """
+    Scale a image into specific size.
+    """
+    if not isinstance(img, np.ndarray):
+        img = img.to_numpy()
+    if h is None:
+        h = w
+    u, v = (img.shape[0] - 1) / (w - 1), (img.shape[1] - 1) / (h - 1)
```
Rullec, line 60: How about adding an safety assertion before we use these `w` and `h`? like
```
w = int(w)
h = int(h)
assert w >=1 and h >=1, "xxx"
```
It can makes the error message more obvious.

docs/gui.rst
archibate: How about `ti.imscale(img, w=512, h=w)`?

docs/gui.rst
archibate: In fact integrator means 积分器 :rofl:

python/taichi/misc/image.py
archibate:
```suggestion
    assert isinstance(w, int) and isinstance(h, int) and w > 1 and h > 1
    u, v = (img.shape[0] - 1) / (w - 1), (img.shape[1] - 1) / (h - 1)
```

docs/gui.rst
Rullec: Yes! I love that ;)

Issue #1715, opened 2020-08-17
[Perf] [CUDA] snode_reader/writer causes slowdown of following kernel


KLozes:
**Describe the bug**
I noticed that python scope field accesses can cause pretty big slowdowns with the cuda backend (and probably opengl and metal). I believe this is likely due to synchronize() calls associated with snode_readers and snode_writer. Is there some way we can get around this? I don't think synchronization is always necessary when using cuda with unified memory because it has page fault handling. This is particularly important for shapeless fields (ie. a = field(ti.f32, ()) since they're frequently read/written in python scope after global reductions. 

**To Reproduce**
This script hows how to setting ```c``` in python scope instead of a dedicated kernel can a big slowdown of the following  ```copy()```

``` python 
import taichi as ti

real = ti.f32
ti.init(arch=ti.cuda, default_fp=real, kernel_profiler=True)

fast = True
N = 1024

a = ti.field(real)
b = ti.field(real)
c = ti.field(real, ())

ti.root.dense(ti.ij, N).place(a)
ti.root.dense(ti.ij, N).place(b)

@ti.kernel
def copy():
    for I in ti.grouped(a):
        b[I] = a[I]

@ti.kernel
def copy2():
    for I in ti.grouped(a):
        b[I] = a[I]

@ti.kernel
def set_c():
    c[None] = 1.0

for i in range(10000):
    if fast:
        set_c()
    else:
        # this may cause a cuda sync that makes copy() slower
        c[None] = 1.0
    copy()
    copy2()

ti.kernel_profiler_print()

```

**Log/Screenshots**
using snode_writer
```
CUDA Profiler
=========================================================================
[      %     total   count |      min       avg       max   ] Kernel name
[ 78.98%   2.972 s  10000x |    0.115     0.297     1.696 ms] copy_c4_0_kernel_1_range_for
[ 21.02%   0.791 s  10000x |    0.074     0.079     0.294 ms] copy2_c6_0_kernel_2_range_for
-------------------------------------------------------------------------
[100.00%] Total kernel execution time:   3.763 s   number of records: 2
=========================================================================
```

using kernel
```
=========================================================================
[      %     total   count |      min       avg       max   ] Kernel name
[ 49.28%   0.796 s  10000x |    0.077     0.080     1.091 ms] copy2_c6_0_kernel_2_range_for
[ 49.05%   0.792 s  10000x |    0.077     0.079     0.824 ms] copy_c4_0_kernel_1_range_for
[...]

PR #1736, opened 2020-08-19
[cc] [refactor] C backend use result_buffer for return value too

archibate:
Related issue = #1715

----

### DIFF
--- a/taichi/backends/cc/cc_program.cpp
+++ b/taichi/backends/cc/cc_program.cpp
@@ -49,8 +49,9 @@ void CCKernel::launch(Context *ctx) {
   TI_TRACE("[cc] entering kernel [{}]", name);
   auto entry = program->load_kernel(name);
   TI_ASSERT(entry);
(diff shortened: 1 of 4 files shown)


archibate:
I found a bug of Taichi middle-end:
The `snode_writer` actually need 9 arguments. 1 for data value, 8 for indicing.
And `tensor_dimensionality` fails on CC just because I memcpy `ctx->args` for `8 * sizeof(uint64)` instead of zero-copy.
I.e. **we are already memory leaking in `ctx->args` for 8 bytes**.
So we'd better set `taichi_max_num_args` to `taichi_max_num_indices * 2` iapr.

archibate:
@k-ye Can we deprecate `ctx->args` and make result_buffer read-write? I still feel like to reuse ret buffer as arg buffer.

archibate:
What's the point of separating them actually.. The name `context` is already a hint that it can be both read and write.

archibate:
Replaced by #1739.

Issue #1757, opened 2020-08-23
[Bug][example][Jacobi solver] stable fluid


Eydcao:
**Describe the bug**
Credits to @yucrazing @yuanming-hu  for finding the bug in `stable_fluid.py` example. Artifacts in `stable_fluid.py` example.

**To Reproduce**
Press the left mouse button and drag it to the left-down corner of the window, you will see a strange acceleration of the vortex and extremely white dye.

One bug about the wrong symbol of adding the correction term (by `shi` in Games 201 WeChat group#2, TODO credit his GitHub id here)

I also found that the bug exists in Jacobi_dual pressure passion solver if turned off, the bug goes away.

* Side note: I would love to fix this one if not urgent, maybe 1~2 days will suffice.

Eydcao:
Hi I am going to close the issue since changing the symbol for correction term in bfecc and commenting the jacobi-dual is very simple.

I could help doing so after #1731 which affects this example file to avoid conflict. Or this can be done alongside @archibate working progress in that PR. Thanks for the patience and the rule of simplicity, which :> Saves a lot of bother

PR #1764, opened 2020-08-24
[bug] Fix stable fluid BFECC advection

Eydcao:
Related issue = #1757 

In this PR I plan to fix the BFECC advection in `stable_fluid.py`. Seems that there are several problems tangled inside this one. So let's try to solve them one-by-one. I may need your @yuanming-hu  and @archibate help to review if the errors are from the BFECC or the PPE solving. Thanks a lot,

- [x] Fixed the correction term's symbol.
- [x] Reset Poisson solver to `jacobi_single`.
- [x] Fixed the whole `advect_bfecc` function.
- [x] Fixed the `int()` (with `math.floor()`) which irritates when input is negative

----

### DIFF
--- a/examples/stable_fluid.py
+++ b/examples/stable_fluid.py
@@ -314,7 +314,7 @@ class MouseDataGen(object):
         # [4:7]: color
         mouse_data = np.zeros(8, dtype=np.float32)
         if gui.is_pressed(ti.GUI.LMB) or frame < 20:
-            if frame < 20:
+            if gui.frame < 20:
                 # print(frame)
                 mxy = np.array([0 + frame*0.01, 0 + frame*0.01],
                                dtype=np.float32) * res
@@ -346,7 +346,6 @@ def reset():
 
 gui = ti.GUI('Stable Fluid', (res, res))
 md_gen = MouseDataGen()
-frame = 0
 while gui.running:
     if gui.get_event(ti.GUI.PRESS):
         e = gui.event
@@ -360,13 +359,12 @@ while gui.running:
         elif e.key == 'd':
             debug = not debug
 
-    frame += 1
 
-    if frame > 4500:
+    if gui.frame > 4500:
         break
 
     if not paused:
-        mouse_data = md_gen(gui, frame)
+        mouse_data = md_gen(gui, gui.frame)
         # if frame < 40:
         #     print(mouse_data)
         step(mouse_data)
--- a/examples/stable_fluid.py
+++ b/examples/stable_fluid.py
@@ -68,9 +68,9 @@ def bilerp(vf, p):
     # fract
     fu, fv = s - iu, t - iv
     a = sample(vf, iu, iv)
-    b = sample(vf, iu+1, iv)
-    c = sample(vf, iu, iv+1)
-    d = sample(vf, iu+1, iv+1)
+    b = sample(vf, iu + 1, iv)
+    c = sample(vf, iu, iv + 1)
+    d = sample(vf, iu + 1, iv + 1)
     return lerp(lerp(a, b, fu), lerp(c, d, fu), fv)
 
 
@@ -81,9 +81,9 @@ def sample_minmax(vf, p):
     # floor
     iu, iv = ti.floor(s), ti.floor(t)
     a = sample(vf, iu, iv)
-    b = sample(vf, iu+1, iv)
-    c = sample(vf, iu, iv+1)
-    d = sample(vf, iu+1, iv+1)
+    b = sample(vf, iu + 1, iv)
+    c = sample(vf, iu, iv + 1)
+    d = sample(vf, iu + 1, iv + 1)
     return min(a, b, c, d), max(a, b, c, d)
 
 
@@ -115,8 +115,8 @@ backtrace = backtrace_rk3
 
 
 @ti.kernel
-def advect_semilag(vf: ti.template(), qf: ti.template(),
-                   new_qf: ti.template(), intermedia_qf: ti.template()):
+def advect_semilag(vf: ti.template(), qf: ti.template(), new_qf: ti.template(),
+                   intermedia_qf: ti.template()):
     ti.cache_read_only(qf, vf)
     for i, j in vf:
         p = ti.Vector([i, j]) + 0.5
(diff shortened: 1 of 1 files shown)


examples/stable_fluid.py
```diff
@@ -341,8 +360,15 @@ def reset():
...
             debug = not debug
 
+    frame += 1
+
+    if frame > 4500:
+        break
+
     if not paused:
-        mouse_data = md_gen(gui)
+        mouse_data = md_gen(gui, frame)
```
archibate, line 369:
```suggestion
        mouse_data = md_gen(gui, gui.frame)
```

examples/stable_fluid.py
```diff
@@ -296,17 +308,23 @@ def __init__(self):
...
-    def __call__(self, gui):
+    def __call__(self, gui, frame):
         # [0:2]: normalized delta direction
         # [2:4]: current mouse xy
         # [4:7]: color
         mouse_data = np.zeros(8, dtype=np.float32)
-        if gui.is_pressed(ti.GUI.LMB):
-            mxy = np.array(gui.get_cursor_pos(), dtype=np.float32) * res
+        if gui.is_pressed(ti.GUI.LMB) or frame < 20:
+            if frame < 20:
```
archibate, line 317:
```suggestion
            if gui.frame < 20:
```

examples/stable_fluid.py
```diff
@@ -123,21 +125,29 @@ def advect_semilag(vf: ti.template(), qf: ti.template(),
...
-        p_mid = backtrace(vf, p, dt)
-        q_mid = bilerp(qf, p_mid)
-        p_fin = backtrace(vf, p_mid, -dt)
-        q_fin = bilerp(qf, p_fin)
-        new_qf[i, j] = q_mid + 0.5 * (q_fin - qf[i, j])
+        p = backtrace(vf, p, dt)
+        intermedia_qf[i, j] = bilerp(qf, p)
+
+    ti.cache_read_only(intermedia_qf, qf, vf)
+    for i, j in vf:
```
archibate, line 136: Is the key problem about separating bfecc to two offload?

Eydcao: Yes. To be more precise, we need a forward advection (backtrack + interpolate) to get an inter-media quantity field first, then a backward advection to (backtrace in -dt direction + interpolate) to get the field at the original time stamp (assuming there is no error).

Since for the second one (backward advection) we need the whole inter-media field ready, it's necessary to run it in a separate another loop and store it already.

examples/stable_fluid.py
Eydcao: Hi, thanks for the convenient hint~

This part is just for reproducing the same phenomena for test. I will comment out and restore to old, interactive way when finished.

examples/stable_fluid.py
Eydcao: But to be noticed, `jacobbi_dual` caused serious artifacts with the new `bfecc`.

I will post detailed comparisons between combination of different methods when I get a minute.

examples/stable_fluid.py
```diff
@@ -328,6 +346,7 @@ def reset():
 
 gui = ti.GUI('Stable Fluid', (res, res))
 md_gen = MouseDataGen()
+frame = 0
```
Eydcao, line 349:
```suggestion
```

examples/stable_fluid.py
```diff
@@ -341,8 +360,15 @@ def reset():
         elif e.key == 'd':
             debug = not debug
 
+    frame += 1
```
Eydcao, line 363:
```suggestion
```

examples/stable_fluid.py
```diff
@@ -341,8 +360,15 @@ def reset():
...
+
+    if frame > 4500:
```
Eydcao, line 365:
```suggestion
    if gui.frame > 4500:
```

Eydcao:
Comparison List:

1. Old `bfecc`

old

2. New `bfecc` with `jacobi_dual`

video

3. New `bfecc` with `jacobi_single`

video

Eydcao:
4. New `bfecc` with `mgpcg`

video

This is super interesting~@archibate

archibate:
Thanks for all the hard work here! I think I found the reason now.

Eydcao:
After fixing the irrigating `int()` issue in interpolation, the boundary artifact is gone. And the new `bfecc` works nicely with `jacobi_single`. As the `jacobi_double` or `mgpcg` concerns other issues in PPE solving, I will leave them there to solve :)

Will organize and make it ready for review

examples/stable_fluid.py
```diff
@@ -123,21 +125,29 @@ def advect_semilag(vf: ti.template(), qf: ti.template(),
...
+    ti.cache_read_only(intermedia_qf, qf, vf)
+    for i, j in vf:
+        p = ti.Vector([i, j]) + 0.5
+        p_two_star = backtrace(vf, p, -dt)
+        p_star = backtrace(vf, p, dt)
+        q_star = intermedia_qf[i, j]
+        new_qf[i, j] = bilerp(intermedia_qf, p_two_star)
 
-        min_val, max_val = sample_minmax(qf, p_mid)
+        # new
```
Eydcao, line 143:
```suggestion
```

examples/stable_fluid.py
```diff
@@ -326,6 +338,9 @@ def reset():
     dyes_pair.cur.fill(0)
 
 
+v_mngr = ti.VideoManager('./', framerate=24, automatic_build=False)
+
+
```
Eydcao, line 343:
```suggestion
```

examples/stable_fluid.py
```diff
@@ -326,6 +338,9 @@ def reset():
     dyes_pair.cur.fill(0)
 
 
+v_mngr = ti.VideoManager('./', framerate=24, automatic_build=False)
+
```
Eydcao, line 342:
```suggestion
```

examples/stable_fluid.py
Eydcao, line 343:
```suggestion
```

examples/stable_fluid.py
```diff
@@ -123,21 +125,29 @@ def advect_semilag(vf: ti.template(), qf: ti.template(),
...
-        p_fin = backtrace(vf, p_mid, -dt)
-        q_fin = bilerp(qf, p_fin)
-        new_qf[i, j] = q_mid + 0.5 * (q_fin - qf[i, j])
+        p = backtrace(vf, p, dt)
+        intermedia_qf[i, j] = bilerp(qf, p)
+
+    ti.cache_read_only(intermedia_qf, qf, vf)
+    for i, j in vf:
+        p = ti.Vector([i, j]) + 0.5
+        p_two_star = backtrace(vf, p, -dt)
```
Eydcao, line 138:
```suggestion
        # star means the temp value after a back tracing (forward advection)
        # two star means the temp value after a forward tracing (reverse advection)
        p_two_star = backtrace(vf, p, -dt)
```

archibate:
Please do not use `apply suggestions` for large modification, always keep in mind that all our actions on GitHub will be sent into 
322 people's mailbox. See https://github.com/taichi-dev/taichi/watchers.

yuanming-hu (approved):

PR #1771, opened 2020-08-24
[ir] [opt] Unify ifloordiv implementation over backends in demote_operations

archibate:
Related issue = #

----
Currently we have 4 duplicate implementation of floordiv, each is different.
It would be great to convert `floordiv` during the IR lowering process, so that we could remove the per-backend ones.

This PR aims to `floordiv` on integer types.
Will do `floordiv` on float types, also `mod` and other Python-vs-C-stuffs later iapr.

----
State: `test_div.py` passed on all my archs but `mpm128.py` failed.


archibate:
@xumingkuan It seems that we're using `alg_simp` for `scalar op constant` optimization? Then where should I put `floordiv -> floor + div`?

archibate:
> @xumingkuan It seems that we're using `alg_simp` for `scalar op constant` optimization? Then where should I put `floordiv -> floor + div`?

I put them into a fresh new pass: `demote_operations`. Will add `mod` and other demotions to there too later.

taichi/transforms/demote_operations.cpp
```diff
@@ -0,0 +1,85 @@
...
+    auto rhs = stmt->rhs;
+    if (stmt->op_type == BinaryOpType::floordiv) {
+      if (is_integral(rhs->element_type()) &&
+          is_integral(lhs->element_type())) {
+        // @ti.func
+        // def ifloordiv(a, b):
+        //     r = ti.raw_div(a, b)
+        //     if (a < 0) != (b < 0) and a and b * r != a:
+        //         r = r - 1
+        //     return r
```
xumingkuan, line 29: I would suggest not involving `b * r != a` since I don't think we can optimize it out easily.

What we need is:
```
a >= 0, b > 0: a / b
a >= 0, b < 0: (a - b + 1) / b
a <= 0, b > 0: (a - b + 1) / b
a <= 0, b < 0: a / b
```
So `((a < 0) == (b < 0) ? a : (a - b + 1)) / b` should be fine.

archibate: It doesn't work for `a = 2, b = -3`:

```py
import taichi as ti


@ti.func
def ifloordiv(a, b):
    return (a if (a < 0) == (b < 0) else (a - b + 1)) / b


@ti.kernel
def func(a: int, b: int) -> int:
    return ifloordiv(a, b)


a, b = 2, -3
print(func(a, b), a // b)
```

This is non-trivial, ff2 change it iapr on your own.

xumingkuan: Oh, it should be
```
a >= 0, b > 0: a / b
a >= 0, b < 0: (a - b - 1) / b
a <= 0, b > 0: (a - b + 1) / b
a <= 0, b < 0: a / b
```
......
BTW when do we encounter `floordiv` at the frontend? Only when `ti.floordiv(a, b)`?

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -499,6 +499,7 @@ class KernelGen : public IRVisitor {
     if (bin->op_type == BinaryOpType::floordiv) {
       if (is_integral(bin->lhs->element_type()) &&
           is_integral(bin->rhs->element_type())) {
+        TI_WARN("Integer floordiv called! It should be taken care by alg_simp");
```
xumingkuan, line 502:
```suggestion
        TI_WARN("Integer floordiv called! It should be taken care by demote_operations");
```

taichi/analysis/verify.cpp
```diff
@@ -41,14 +41,12 @@ class IRVerifier : public BasicStmtVisitor {
...
-          found,
-          "IR broken: stmt {} cannot have operand {}."
-          " Consider adding `ti.core.toggle_advanced_optimization(False)`."
-          " If that fixes the problem, please report this bug by opening an"
-          " issue at https://github.com/taichi-dev/taichi to help us improve."
-          " Thanks!",
-          stmt->id, op->id);
+      TI_ASSERT_INFO(found,
+                     "IR broken: stmt {} cannot have operand {}."
+                     " Please report this bug by opening an issue at"
```
xumingkuan, line 46:
```suggestion
                     " If you are using autodiff, please check"
                     " https://taichi.readthedocs.io/en/stable/differentiable_programming.html#kernel-simplicity-rule."
                     " If it doesn't help, please report this bug by opening an issue at"
```

Not sure if we want to add this... Programs violating these rules often result in this error message.

Note that the URL is so long that @taichi-gardener may complain...

xumingkuan (commented):
LGTMig. I wonder if there is a way to write something in Python equivalent to or as *fast* as `a / b` in C++...

archibate:
> LGTMig. I wonder if there is a way to write something in Python equivalent to or as _fast_ as `a / b` in C++...

Python didn't care about fast, the C style can sometime cause hard-to-debug issues when writing physics engines.

taichi/transforms/compile_to_offloads.cpp
```diff
@@ -48,6 +48,9 @@ void compile_to_offloads(IRNode *ir,
   print("Typechecked");
   irpass::analysis::verify(ir);
 
+  irpass::demote_operations(ir);
+  print("Operations Demoted");
+
```
xumingkuan, line 53: Let's move this to just before `Simplified IV` as it can impede other optimizations.

taichi/backends/opengl/codegen_opengl.cpp
xumingkuan: The pass' name is `demote_operations`.

xumingkuan (commented):
Just a few nits.

taichi/transforms/compile_to_offloads.cpp
archibate: So, we only demote_operations in `compile_to_executable`, not `compile_to_offloads`?

xumingkuan (approved):
LGTM now. Thanks!

taichi/transforms/compile_to_offloads.cpp
xumingkuan: Yes. We may perform some optimizations after `compile_to_offloads` in async mode.

PR #1784, opened 2020-08-26
[Example] Fix stable_fluid.py boundary broken: int -> ti.floor, and make mpm3d.py 3x faster on CUDA

### LINKED ISSUE #1757: [Bug][example][Jacobi solver] stable fluid

archibate:
Related issue = close #1757 replace #1764 break_down #1691

----

### DIFF
--- a/examples/stable_fluid.py
+++ b/examples/stable_fluid.py
@@ -62,7 +62,8 @@ def bilerp(vf, p):
     u, v = p
     s, t = u - 0.5, v - 0.5
     # floor
-    iu, iv = int(s), int(t)
+    iu, iv = ti.floor(s), ti.floor(t)
+    iu, iv = int(s + 1) - 1, int(t + 1) - 1
     # fract
     fu, fv = s - iu, t - iv
     a = sample(vf, iu + 0.5, iv + 0.5)
--- a/examples/stable_fluid.py
+++ b/examples/stable_fluid.py
@@ -63,7 +63,6 @@ def bilerp(vf, p):
     s, t = u - 0.5, v - 0.5
     # floor
     iu, iv = ti.floor(s), ti.floor(t)
-    iu, iv = int(s + 1) - 1, int(t + 1) - 1
     # fract
     fu, fv = s - iu, t - iv
     a = sample(vf, iu + 0.5, iv + 0.5)
--- a/examples/mpm3d.py
+++ b/examples/mpm3d.py
@@ -1,15 +1,18 @@
-export_file = '/tmp/mpm3d.ply'
+export_file = ''  # use '/tmp/mpm3d.ply' for exporting result to disk
 
 import taichi as ti
 import numpy as np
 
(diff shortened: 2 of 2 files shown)


examples/stable_fluid.py
```diff
@@ -62,7 +62,8 @@ def bilerp(vf, p):
     u, v = p
     s, t = u - 0.5, v - 0.5
     # floor
-    iu, iv = int(s), int(t)
+    iu, iv = ti.floor(s), ti.floor(t)
+    iu, iv = int(s + 1) - 1, int(t + 1) - 1
```
archibate, line 66:
```suggestion
```

examples/stable_fluid.py
```diff
@@ -62,7 +62,7 @@ def bilerp(vf, p):
     u, v = p
     s, t = u - 0.5, v - 0.5
     # floor
-    iu, iv = int(s), int(t)
+    iu, iv = ti.floor(s), ti.floor(t)
```
archibate, line 65:
```py
>>> import math
>>> math.floor(-1.1)
-2
>>> int(-1.1)
-1
```
So `int` is quite different from `floor` when things turn negative. The same to Taichi-scope.
This causes the -x and -y side having funny effects when non-zero velocity is encountered.
Hi, dear @Eydcao:
Did you assumed `int` to be exactly same with `ti.floor`, return the largest integer below the input argument?
It warns that we should never do assumptions when debugging. Hope we all have fun and learnt a lot from this vivid lesson :)

Eydcao: Interesting!

but in the old code, we used `int()` instead of `math.floor()`, so in python these two are the same thing?

archibate: Both in Python and Taichi, they are different thing :)

Eydcao: Great! This one is good to merge.

I have got the idea why `int()` is irrigating here as it leads to the wrong coefficients in `lerp`. So when I fixed these two lines in #1764.

**The strange boundary artifact is gone, great!**

The new `bfecc` works perfec with `jacobi_single` now, **but still has problems with `jacobi_dual` and `mgpcg`**.

Eydcao (approved):
LGTM!

Please see the detailed comments about the problem this `int()` thing solved. (Still some problems exist)

PR #1793, opened 2020-08-28
[workflow] Travis CI only test Python 3.6 unless release tag is used

archibate:
Related issue = #1580 

----

### DIFF
--- a/.github/workflows/persubmit.yml
+++ b/.github/workflows/persubmit.yml
@@ -9,18 +9,20 @@ jobs:
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'skip ci') && github.event.sender.login != 'taichi-gardener' }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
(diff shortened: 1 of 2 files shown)


.github/workflows/persubmit.yml
```diff
@@ -9,18 +9,20 @@ jobs:
...
     strategy:
       matrix:
-        os: [ubuntu-latest]
-        python: [3.6, 3.8]
-        with_cc: [OFF]
         include:
+          - os: ubuntu-latest
+            python: 3.6
+            with_cc: OFF
           - os: macos-latest
```
k-ye, line 16: Does this mean both Github actions and travis are testing OSX/Linux?

k-ye (approved):
Cool!

.github/workflows/persubmit.yml
archibate: Yeah both GitHub actions an Travis are testing OS X, only GitHub actions are testing Linux.

PR #1779, opened 2020-08-25
[test] Expand the tolerance for testing on metal backend

Rullec:
Related issue = Close #1679 

As @archibate adviced, I expand the approx value on metal platform.

Now `test_ad_basic.py` can work properly on my machine! 

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

----

PR #1798, opened 2020-08-28
[Lang] [test] Add ti.get_rel_eps() for maximal relative error tolerance on current backend

k-ye:
It turns out that #1779 doesn't play well on my device. Adding this to make the `allclose` tests pass again on Metal.

Related issue = #1779

----

### DIFF
--- a/python/taichi/testing.py
+++ b/python/taichi/testing.py
@@ -2,6 +2,17 @@ import taichi as ti
 
 
 # Helper functions
+def default_epsilon_of_arch(arch):
+    if arch == ti.opengl:
+        return 1e-3
+    elif arch == ti.metal:
+        # Debatable, different hardware could yield different precisions
+        # On AMD Radeon Pro 5500M, 1e-6 works fine...
+        # https://github.com/taichi-dev/taichi/pull/1779
+        return 1e-4
+    return 1e-6
+
+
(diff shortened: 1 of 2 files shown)


python/taichi/testing.py
```diff
@@ -2,6 +2,17 @@
 
 
 # Helper functions
+def default_epsilon_of_arch(arch):
+    if arch == ti.opengl:
```
archibate, line 6:
```suggestion
def get_rel_eps():
    if ti.cfg.arch == ti.opengl:
```
We can actually expose this function to end-users? E.g., in `examples/renderer_util.py`.

k-ye: Done

archibate (approved):
LGTM now. We may also find a proper `ti.get_abs_eps` for each backend.

k-ye:
> We may also find a proper ti.get_abs_eps for each backend.

Actually, what's the difference between `rel` and `abs` in this case?

archibate:
> > We may also find a proper ti.get_abs_eps for each backend.
> 
> Actually, what's the difference between `rel` and `abs` in this case?

IIUC it's roughly:
```py
def erf(x):
  return max(abs_eps, rel_eps * x)
```

PR #1800, opened 2020-08-28
[Doc] Add how to set environment variables for fish users

chengluyu:
🙋‍♂️ Hi, I'm a [fish](https://fishshell.com/) user and encountered a small problem when I was adding environment variables. So I added this part to the documentation. Hope it helps other fish users.

### DIFF
--- a/docs/dev_install.rst
+++ b/docs/dev_install.rst
@@ -116,7 +116,9 @@ Setting up Taichi for development
 
     Then execute ``source ~/.bashrc`` to reload shell config.
 
-    If you're using fish, use ``set -x NAME VALUES``, otherwise it won't be loaded by child processes.
+    .. note::
+
+        If you're using fish, use ``set -x NAME VALUES``, otherwise it won't be loaded by child processes.
 
   * On Windows, please add these variables by accessing your system settings:
 
--- a/docs/dev_install.rst
+++ b/docs/dev_install.rst
@@ -116,6 +116,17 @@ Setting up Taichi for development
 
     Then execute ``source ~/.bashrc`` to reload shell config.
 
+    If you're using fish. Use ``set -x`` to set environment variables. Execute following commands
+    or add them to ``~/.config/fish/config.fish`` (create if it doesn't exist).
+
+    .. code-block:: fish
+
+      set -x TAICHI_REPO_DIR /path/to/taichi
+      set -x PYTHONPATH $TAICHI_REPO_DIR/python $PYTHONPATH
+      set -x PATH $TAICHI_REPO_DIR/bin $PATH
+      # set -x CXX /path/to/clang # Uncomment if you encounter issue about compiler in the next step.
+      # set -x PATH /opt/llvm/bin $PATH # Uncomment if your llvm or clang is at somewhere else.
+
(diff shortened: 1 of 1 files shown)


docs/dev_install.rst
```diff
@@ -116,6 +116,17 @@ Setting up Taichi for development
...
+    If you're using fish. Use ``set -x`` to set environment variables. Execute following commands
+    or add them to ``~/.config/fish/config.fish`` (create if it doesn't exist).
+
+    .. code-block:: fish
+
+      set -x TAICHI_REPO_DIR /path/to/taichi
+      set -x PYTHONPATH $TAICHI_REPO_DIR/python $PYTHONPATH
+      set -x PATH $TAICHI_REPO_DIR/bin $PATH
+      # set -x CXX /path/to/clang # Uncomment if you encounter issue about compiler in the next step.
+      # set -x PATH /opt/llvm/bin $PATH # Uncomment if your llvm or clang is at somewhere else.
```
archibate, line 128:
```suggestion
      # set -x PATH /opt/llvm/bin:$PATH # Uncomment if your llvm or clang is at somewhere else.
```
Not super sure about `fish`, is this intentional?

archibate (approved):
LGTMig!
Plus 1 concern: will we add a explaination for each kind of shell including `csh`, `pwsh`, `pysh`? My opinion is `fish` users should know how to set environment variables (`set -x`). If it's not used by majority, we'd better use some generic advices, e.g.: `set TAICHI_REPO_DIR to the path to Taichi repo, no matter bash, zsh, fish, or even windows`.

docs/dev_install.rst
chengluyu: Yes! This is intentional. See https://github.com/fish-shell/fish-shell/issues/527#issuecomment-12435368 for an example.

chengluyu:
> LGTMig!
> Plus 1 concern: will we add a explaination for each kind of shell including `csh`, `pwsh`, `pysh`? My opinion is `fish` users should know how to set environment variables (`set -x`). If it's not used by majority, we'd better use some generic advices, e.g.: `set TAICHI_REPO_DIR to the path to Taichi repo, no matter bash, zsh, fish, or even windows`.

Your concern is right. I've read some documentations (e.g. [nvm](https://github.com/nvm-sh/nvm#fish)), which include variations of different shells for users’ convenience. But this practice will make our documentation long-winded. So I shorten the description, including the possible pitfall only.

docs/dev_install.rst
```diff
@@ -116,6 +116,8 @@ Setting up Taichi for development
 
     Then execute ``source ~/.bashrc`` to reload shell config.
 
+    If you're using fish, use ``set -x NAME VALUES``, otherwise it won't be loaded by child processes.
```
archibate, line 119: Sounds great! I only have 1 nit:
```suggestion
    .. note::

        If you're using fish, use ``set -x NAME VALUES``, otherwise it won't be loaded by child processes.
```

chengluyu: Great! I'd adopt this.

archibate (approved):
LGTM now.

PR #1802, opened 2020-08-29
[workflow] Import missing package sys in code_format.py

chengluyu:
Related issue = N/A

----

In code_format.py, `sys` is used but not imported. This PR aims to fix this. The error happens when clang-format executable is not found.

https://github.com/taichi-dev/taichi/blob/1662fc52fa5eadeac1abd46e52f1303f6aa11e36/python/taichi/code_format.py#L50-L55

### DIFF
--- a/python/taichi/code_format.py
+++ b/python/taichi/code_format.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python3
 
-import os, re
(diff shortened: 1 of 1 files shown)


archibate (approved):
Nice catch :+1:

Issue #1820, opened 2020-09-02
ti.einsum


parsarahimi:
Einsum is used for matrix multiplications where we use index notation to describe dot/tensor/outer products, it would be a nice feature to have for taichi

https://numpy.org/doc/stable/reference/generated/numpy.einsum.html

PR #1826, opened 2020-09-03
[Lang] Support x.fill(n) and x.copy_from(y) in top-level of Taichi-scope

archibate:
Related issue = to make a way for #1820

----
Previously we can only call `fill` and `copy_from` in Python-scope, because the meta functions in `lang/meta.py` are defined to be `@ti.kernel`.
But actually if we change it to `@ti.func`, they will still functional when invoked in top-level of Taichi-scope (just can't be nested in parallel for loops).
So I'd like to add a `@ti.kernelfunc` that automatically detect the scope:
- call `@ti.func` when `impl.inside_kernel()`, otherwise call `@ti.kernel`.

I think ultimately we will rename `@ti.kernelfunc` to `@ti.kernel` and deprecate the old `@ti.kernel`, i.e., **allowing kernels to call other kernels**! (may also rename `@ti.pyfunc` to `@ti.func` in the future)
After these two renames, the only difference between `@ti.func` and `@ti.kernel` will be: the latter needs type-hints and the first doesn't.

@k-ye Could you help me with the ODOP inspector for our new `@ti.kernelfunc`? I can't make it work in this PR..
Of course, we won't add this to doc so it's not a public API yet, like `@ti.pyfunc` did.

### DIFF
--- a/examples/tree_gravity.py
+++ b/examples/tree_gravity.py
@@ -2,7 +2,7 @@
 # Author: archibate <1931127624@qq.com>, all left reserved
 import taichi as ti
 import taichi_glsl as tl
-ti.init()
+ti.init(ti.cpu)
 if not hasattr(ti, 'jkl'):
     ti.jkl = ti.indices(1, 2, 3)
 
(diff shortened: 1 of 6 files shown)


feisuzhu:
Closing ancient PR.

PR #1827, opened 2020-09-03
[refactor] Drop ODOP support temporary (rullecized)

archibate:
Related issue = #

----
As @yuanming-hu said in his final lecture in GAMES 201, *simplicity matters*.
I noticed that our ODOP implementation is getting too complicated that developers can hardly maintain them.
Although not found in Python Zen, I think I will add *serialized is better than parallelized* and *don't put things too tight together*.
Currently `is_classfunc` is embed into `Kernel` and `Func`, which is lossening the focus IMO.
The module `re` is actually completely unrelated to the rest part of `kernel.py`, but linked by our ODOP implementation.
I'm thinking if it's possible to move that outside of `Kernel` and `Func`, or even implement class inspection via an external package?
So I'd like to remove these codes for now, and completely reimplement it in a much simpler way, in a following PR.
Using the same workflow as @Rullec did in `rullec-field`, I merge this PR into a temp branch called `archibate-odop`, so `master` won't suffer regressions on ODOP, what a smart idea! And will only merge `archibate-odop` to `master` after everything is done. I'd like to call this cool workflow `rullecized PRs` as it's first used by @Rullec, WDYT about the name?
@k-ye I **promise** I'll add it back in another PR, except for increasing maintainability, so please don't feel upset :) Your work is still valuable to me. So please approve if you think the change doesn't break non-ODOP behaviors, will add it back iapr and your implementation is an important reference for implementing a decoupled version ODOP.

### DIFF
--- a/python/taichi/lang/kernel.py
+++ b/python/taichi/lang/kernel.py
@@ -31,10 +31,8 @@ def remove_indent(lines):
 
 # The ti.func decorator
 def func(foo):
-    is_classfunc = _inside_class(level_of_class_stackframe=3)
-
     _taichi_skip_traceback = 1
-    fun = Func(foo, classfunc=is_classfunc)
+    fun = Func(foo)
 
     @functools.wraps(foo)
     def decorated(*args):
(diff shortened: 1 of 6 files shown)


k-ye:
> So please don't feel upset :) Your work is still valuable to me.

Thanks. Most of ODOP is @yuanming-hu 's work, so the major credit goes to him... 

> So please approve if you think the change doesn't break non-ODOP behaviors

While I don't know what the refactoring plan is heading towards, I'm fine with experimental features if they are not directly merged into `master`, and I don't think you need my (or anyone's) approval to merge to experimental branches?

k-ye (approved):

archibate:
> I don't know why you need to remove these first, instead of just having new impls

I have a much better idea that is complete different from the current approach, will show you in later PR.

> and I don't think you need my (or anyone's) approval to merge to experimental branches?

Just to confirm if I'm doing something totally wrong.

PR #1831, opened 2020-09-04
[refactor] Use promise/future to retrieve the function compiled asynchronously

k-ye:
Mostly a code cleanup. I saw no improvement nor regression on my end...

Related issue = #

----


yuanming-hu (approved):
Cool! Thanks for making this part much cleaner!

PR #1837, opened 2020-09-05
[misc] Check the previous Github workflow run status if CI is skipped

k-ye:
Test test test

---

Adds a `check_previous_run` job to our presubmit workflow, so that when `skip ci` is enabled, the job will use the previous workflow run status. If the previous run hasn't finished, this job will poll the status every 15s, with a deadline of 1 hour.

The idea is described in https://github.com/taichi-dev/taichi/issues/1809#issuecomment-686538091. Basically:

1. Find the previous commit of this one in the current PR
2. Find the workflow run corresponding to the previous commit
3. Check the status of the run.

**Future TODOs**

* Now that we have the workflow run ID of the previous commit, we can actually instruct to cancel the run ([`cancel` doc](https://docs.github.com/en/rest/reference/actions#cancel-a-workflow-run)).
* Unfortunately, `[skip ci]` in the commit message is actually not working in Github Actions (see https://github.community/t/github-actions-does-not-respect-skip-ci/17325). Instead, the commit message is only useful to skip Travis and Appveyor. 

  We can add a step to extract the head commit's message, and use that to decide whether to trigger the subsequent steps.

See also https://github.com/k-ye/test_gh_action/pull/4

Screen Shot 2020-09-05 at 17 14 21

Related issue = #1809

----

### DIFF
--- a/.github/workflows/persubmit.yml
+++ b/.github/workflows/persubmit.yml
@@ -58,9 +58,9 @@ jobs:
           ti test -vr2 -t2
 
   check_previous_run:
-    name: Checks the Workflow Run Conclusion of the Previous Commit
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'skip ci') || github.event.sender.login == 'taichi-gardener' }}
+    name: Checks the Workflow Run of the Previous Commit
     runs-on: ubuntu-latest
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'skip ci') || github.event.sender.login == 'taichi-gardener' }}
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
--- a/.github/workflows/persubmit.yml
+++ b/.github/workflows/persubmit.yml
@@ -57,6 +57,21 @@ jobs:
           ti diagnose
           ti test -vr2 -t2
 
+  check_previous_run:
+    name: Checks the Workflow Run Conclusion of the Previous Commit
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'skip ci') || github.event.sender.login == 'taichi-gardener' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Check the previous run
+        run: |
+          PR=${{ github.event.pull_request.number }}
+          SHA=${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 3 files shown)


archibate (approved):
LGTM! Have you tested @taichi-gardener's `[skip ci] enforce code format` yet?

k-ye:
> Have you tested @taichi-gardener's [skip ci] enforce code format yet?

Good point, let me do some test before we declare ready to use.

k-ye:
> Have you tested @taichi-gardener's [skip ci] enforce code format yet?

I think it's working now :) See

Screen Shot 2020-09-06 at 10 09 37

Here's the action run page for the second commit from @taichi-gardener : https://github.com/taichi-dev/taichi/actions/runs/241166594:

```
...
2020-09-06 01:08:57 INFO:Waiting to get the status of run=241166332 (url=https://api.github.com/repos/taichi-dev/taichi/actions/runs/241166332). retries=23
2020-09-06 01:09:12 DEBUG:request=https://api.github.com/repos/taichi-dev/taichi/actions/runs/241166332
2020-09-06 01:09:12 DEBUG:run=241166332 conclusion=failure
2020-09-06 01:09:12 INFO:workflow_run_id=241166332 ok=False
##[error]Process completed with exit code 1.
```

It points to run https://github.com/taichi-dev/taichi/actions/runs/241166332, which was triggered by the first commit from myself.

yuanming-hu (approved):
Looks great to me! This will prevent a lot of misoperations in the future. Thank you so much!

k-ye:
As a side note, it seems that we have selected `Persubmit Checks / Build and Test (ubuntu-latest, 3.8, ON)` as [required status checks on `master` branch](https://docs.github.com/en/enterprise/2.16/admin/developer-workflow/configuring-protected-branches-and-required-status-checks#enabling-required-status-checks). This means even if checking the previous status has succeeded, because the entire `Build and Test` job is skipped, the check will always be in a pending status and block the PR from being merged. 

1. I think this is a nice thing, because it requires at least one successful build and test being done on the PR. On the other hand, this could be a bit too restrictive on certain kinds of PRs, like improving docs or fixing typos... 
2. In the future, we could have yet another job, e.g. `Summarize Build Status`, which collects the result from either the `Build and Test` job, or the ` Checks the Workflow Run of the Previous Commit` job. Then we can make this summarizing job the one whose OK status is required?

At last, I'm not sure if `persubmit` is a typo of `presubmit`, or an intentional expression for `trigger per submit`...

PR #1839, opened 2020-09-05
[async] Implement IRBank and IRHandle

xumingkuan:
Related issue = #742

This PR replaces some `std::unordered_map<uint64 (hash), ...>`'s key from the IR's hash to `IRHandle`, while two `IRHandle` are considered the same iff their hashes are the same.

IR's hash is cached in `IRBank::hash_bank_`. For each `IRNode *`, we only calculate its hash once when calling `IRBank::get_hash()` multiple times.

For now, task fusion is not cached -- we still fuse the same tasks multiple times.

----


taichi/analysis/clone.cpp
```diff
@@ -134,6 +134,9 @@ class IRCloner : public IRVisitor {
 
     if (kernel != nullptr) {
       new_root->kernel = kernel;
+    } else if (new_root->get_parent() == nullptr &&
+               new_root->get_kernel() == nullptr) {
+      new_root->kernel = root->get_kernel();
```
xumingkuan, line 139: I think with these lines, we won't need to pass the `Kernel *` argument. The reason why we passed it is the `OffloadedStmt` is usually under a block initially (so it doesn't store the `Kernel *`) but will become an IR root later.

taichi/program/async_engine.cpp
```diff
@@ -188,15 +210,13 @@ void ExecutionQueue::enqueue(TaskLaunchRecord &&ker) {
       auto func = codegen->codegen();
       async_func->set(func);
     });
+    trash_bin.push_back(std::move(cloned_stmt));
```
xumingkuan, line 213: This is `ExecutionQueue::trash_bin`. Ideally we should have only one trash bin in `IRBank`, but `ExecutionQueue` can't access `IRBank` directly...

taichi/program/async_engine.cpp
```diff
@@ -21,20 +21,53 @@ uint64 hash(IRNode *stmt) {
...
+  // TODO: remove get_kernel() here
+  return irpass::analysis::clone(const_cast<IRNode *>(ir_), ir_->get_kernel());
 }
 
-}  // namespace
+uint64 IRBank::get_hash(IRNode *ir) {
+  auto result_iterator = hash_bank_.find(ir);
+  if (result_iterator == hash_bank_.end()) {
+    auto result = hash(ir);
+    hash_bank_.insert(hash_bank_.end(), std::make_pair(ir, result));
```
k-ye, line 44: nit:
```suggestion
    set_hash(ir, result);
```
?

taichi/program/async_engine.cpp
```diff
@@ -21,20 +21,53 @@ uint64 hash(IRNode *stmt) {
...
+
+void IRBank::set_hash(IRNode *ir, uint64 hash) {
+  hash_bank_[ir] = hash;
+}
+
+bool IRBank::insert(std::unique_ptr<IRNode> &&ir, uint64 hash) {
+  IRHandle handle(ir.get(), hash);
+  auto insert_place = ir_bank_.find(handle);
+  if (insert_place == ir_bank_.end()) {
+    ir_bank_.insert(ir_bank_.end(), std::make_pair(handle, std::move(ir)));
```
k-ye, line 58: nit: 
```suggestion
    ir_bank_.emplace(handle, std::move(ir));
```

---

Also, I wonder if `insert()` should always call `set_hash`?

taichi/program/async_engine.cpp
```diff
@@ -216,28 +236,23 @@ void AsyncEngine::launch(Kernel *kernel, Context &context) {
...
-  const bool kmeta_inited = kmeta.initialized;
   for (std::size_t i = 0; i < offloads.size(); i++) {
     auto *offload = offloads[i]->as<OffloadedStmt>();
-    uint64 h;
-    OffloadedStmt *offl_template = nullptr;
-    if (kmeta_inited) {
-      auto &oc = kmeta.offloaded_cached[i];
-      h = oc.get_hash();
-      offl_template = oc.get_template();
+    uint64 h = ir_bank_.get_hash(offload);
```
k-ye, line 241: Note that previously, we compute the hash on the *cloned* IR, not the original one in the kernel. I'm not sure if this change could result in any problem, though..

taichi/program/async_engine.cpp
k-ye: Hmm, could you then pass IRBank from `AsyncEngine` to `ExecutionQueue`?

taichi/program/async_engine.h
```diff
@@ -19,6 +19,62 @@ TLANG_NAMESPACE_BEGIN
...
+    return ir_handle.hash();
+  }
+};
+}  // namespace std
+
+TLANG_NAMESPACE_BEGIN
+
+class IRBank {
+ public:
+  std::vector<std::unique_ptr<IRNode>> trash_bin;
```
k-ye, line 63: Nit: I'd suggest to make `trashi_bin` private, and add a `insert_to_trash_bin()` method for these reasons:

1. `insert()` could use `insert_to_trash_bin()`
2. In `ExecutionQueue::enqueue`, we can also use this method to collect the cloned `stmt` (provided that ExecutionQueue has access to IRBank..)

WDYT?

taichi/program/async_engine.cpp
```diff
@@ -414,30 +431,29 @@ bool AsyncEngine::fuse() {
...
       irpass::full_simplify(task_a, /*after_lower_access=*/false, kernel);
-      task_queue[i].h = hash(task_a);
+      // For now, re_id is necessary for the hash to be correct.
+      irpass::re_id(task_a);
+
+      auto h = ir_bank_.get_hash(task_a);
+      task_queue[i].ir_handle = IRHandle(task_a, h);
+      ir_bank_.insert(std::move(cloned_task_a), h);
+      task_queue[i + 1].ir_handle = IRHandle(nullptr, 0);
+      ir_bank_.trash_bin.push_back(std::move(cloned_task_b));
```
k-ye, line 444: Not sure if we really need this, since `task_b`'s body has been fully moved into `task_a`... But maybe there could be unexpected references to `b`, could you add a TODO to see if we can remove this?

k-ye (commented):
Great!! I do have a few nits + questions..

taichi/program/async_engine.cpp
xumingkuan: Good point. I just realized that the `stmt->id`s may be different -- the original ones in the kernel can be
```
kernel {
  $0 = offload {
    $1 ~ $9
  }
  $10 = offload {
    $11 ~ $20
  }
}
```
but the cloned ones will have `stmt->id` relabeled. I think we should reuse `KernelMeta` and compute the hash on the cloned IR...

yuanming-hu (approved):
Looks great!! Let's see how the performance goes :-)

taichi/program/async_engine.h
```diff
@@ -19,6 +19,62 @@ TLANG_NAMESPACE_BEGIN
 
 // TODO(yuanming-hu): split into multiple files
 
+class IRHandle {
+ public:
+  IRHandle(IRNode const *ir, uint64 hash) : ir_(ir), hash_(hash) {
```
k-ye, line 24: nit: I think the majority of the codebase put `const` before type..
```suggestion
  IRHandle(const IRNode *ir, uint64 hash) : ir_(ir), hash_(hash) {
```

taichi/program/async_engine.h
```diff
@@ -19,6 +19,62 @@ TLANG_NAMESPACE_BEGIN
...
+  }
+
+  std::unique_ptr<IRNode> clone() const;
+
+  IRNode const *ir() const {
```
k-ye, line 29:
```suggestion
  const IRNode *ir() const {
```

taichi/program/async_engine.h
```diff
@@ -19,6 +19,62 @@ TLANG_NAMESPACE_BEGIN
...
+    return hash_;
+  }
+
+  // Two IRHandles are considered the same iff their hash values are the same.
+  bool operator==(const IRHandle &other_ir_handle) const {
+    return hash_ == other_ir_handle.hash_;
+  }
+
+ private:
+  IRNode const *ir_;  // not owned
```
k-ye, line 43:
```suggestion
  const IRNode *ir_;  // not owned
```

taichi/program/async_engine.h
```diff
@@ -19,6 +19,62 @@ TLANG_NAMESPACE_BEGIN
...
+  uint64 hash_;
+};
+
+TLANG_NAMESPACE_END
+
+namespace std {
+template <>
+struct hash<taichi::lang::IRHandle> {
+  std::size_t operator()(taichi::lang::IRHandle const &ir_handle) const
```
k-ye, line 52:
```suggestion
  std::size_t operator()(const taichi::lang::IRHandle &ir_handle) const
```

k-ye (approved):
LGTM (w/ nits)!

PR #1846, opened 2020-09-06
[cli] Setup pylint: run 'ti lint' to obtain linting results

archibate:
Related issue = #677

----

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,2 @@
+[MESSAGES CONTROL]
+disable=invalid-name, missing-function-docstring, missing-class-docstring, missing-module-docstring, unused-wildcard-import, cyclic-import, blacklisted-name, import-outside-toplevel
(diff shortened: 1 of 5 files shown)


Rullec:
Sorry for my outdated infomation, I didn't notice this has been implemented.

archibate:
> Sorry for my outdated infomation, I didn't notice this has been implemented.

Not completely, this PR is just a setup for the infrasructure. Still many things to do like:

1. pylint can't treat Taichi-scope very well, whose syntax is sightly different from Python-scope.
2. We should be able to lint only some specific files, like `ti test` does. And perfectly only lint the `git diff`.
3. We may even intergrate pylint into our CI system, which should fail when lint 'ERROR' occurred.

yuanming-hu (approved):
Thank you so much! LGTM. Sorry about my delayed review.

PR #1849, opened 2020-09-06
[misc] Fix check previous run by using GITHUB_TOKEN to authenticate APIs

k-ye:
* For authenticated requests on Github Actions, its rate limit is 1000 per hour: https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits
* Slowed down the periodical poll from 15s to 1min.

Intentionally mess up the cpp code to make taichi-gardener work.

Related issue = https://github.com/taichi-dev/taichi/pull/1844#issuecomment-687720574, #1837 

----

### DIFF
--- a/.github/workflows/persubmit.yml
+++ b/.github/workflows/persubmit.yml
@@ -67,10 +67,15 @@ jobs:
         with:
           python-version: 3.8
       - name: Check the previous run
+        env:
+          PR: ${{ github.event.pull_request.number }}
+          SHA: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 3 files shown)


k-ye:
@archibate  You can monitor the run status of the last commit by @taichi-gardener 's in https://github.com/taichi-dev/taichi/runs/1077882106. Note that rate limit is `1000` now.

k-ye:
I'm merging this directly because it imposes no risk to Taichi itself. Also, we need this to unblock the workflow runs from being too flaky.

PR #1850, opened 2020-09-07
[Example] Physarum simulation example

znah:
Code example: A toy model of [Physarum Polycephalum](https://sagejenson.com/physarum).

### DIFF
--- a/examples/physarum.py
+++ b/examples/physarum.py
@@ -1,0 +1,76 @@
+'''Physarum simulation example.
+
+See https://sagejenson.com/physarum for the details.'''
+
+import numpy as np
+import taichi as ti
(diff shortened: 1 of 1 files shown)


znah:
Screenshot:
image

yuanming-hu (approved):
Wow, this is so cool! I have a lot of fun playing with this demo. It's incredible that we can achieve this with ~75 lines of code - this is perhaps the charm of self-organizing systems,

Btw, how far do you think we are from [this](https://payload.cargocollective.com/1/18/598881/13800048/explosion.gif)? I tried to increase particle count/resolution but seems to end up with a different pattern. Maybe it's different parameters :-)

PR #1855, opened 2020-09-08
[async] Stop using std::unordered_multimap as node edges type

k-ye:
This was https://bugs.llvm.org/show_bug.cgi?id=16747#c1

---

It seems like `std::unordered_multimap`, introduced in #1843, was way slower than expected, and became a major bottleneck in SFG (at least with clang's libc++ impl..). Profiling showed that ~74% of the time was spent here:

Screen Shot 2020-09-08 at 21 58 15

This PR switches to `std::unordered_map<AsyncState, std::unordered_set<Node*>>`, which restores the performance to its prior level on my device.

Off-the-topic change: since I removed `print_edges()`, I used `<-` when printing input nodes...

Related issue = #1843 

----


yuanming-hu (approved):
Cool! Thank you so much for fixing this! It's surprising that `std::unordered_multimap` becomes the bottleneck...

xumingkuan (approved):
LGTM!

PR #1857, opened 2020-09-08
[Lang] Support ti.zero(x) and ti.one(x) for making zero/one value for the same type with x

archibate:
Related issue = http://www.jlhub.com/julia/manual/en/function/zero

----
Yeah it's still using `x * 0` since we seems can't get the data type of an expr from python-scope.. maybe iapr, let's set up the API here imutating the Julia zero/one.

### DIFF
--- a/tests/python/test_lang.py
+++ b/tests/python/test_lang.py
@@ -1,4 +1,5 @@
 import taichi as ti
+import pytest
 
 
 @ti.all_archs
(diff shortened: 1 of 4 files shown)


tests/python/test_lang.py
```diff
@@ -131,3 +132,29 @@ def test():
         print(i)
 
     test()
+
+
+@ti.test(require=ti.extension.data64)
```
k-ye, line 137: IIUC, this applies to all the test cases, so platforms like Metal cannot be covered, even for 32-bit types? If so, could you split this into two cases, one for 64-bit and the other for non-64?

tests/python/test_lang.py
```diff
@@ -131,3 +132,29 @@ def test():
...
+def test_meta_zero(dt):
+    @ti.kernel
+    def func(x: dt) -> dt:
+        return ti.zero(x)
+
+    for x in [-1, -2.3, 1.2, 2, 3]:
+        if ti.core.is_integral(dt):
+            x = int(x)
+        assert func(x) == 0  # intentionally not using ti.approx here
+
```
k-ye, line 148: nit: Test for `ti.Matrix` as well?

k-ye (approved):
LGTM + nits!

PR #1859, opened 2020-09-09
[async] Migrate AsyncEngine::fuse() to StateFlowGraph

xumingkuan:
Related issue = #742

Also implements `bit::Bitset`.

TODO: element-wise check; iteratively update taskmeta to iteratively fuse tasks (currently 10 `inc()`s are fused to 5 tasks rather than 1).

----


yuanming-hu (approved):
Thank you so much! I took a quick look and it LGTM.

PR #1873, opened 2020-09-15
[Async] Fix ClearListStmt metadata

yuanming-hu:
Related issue = #742

@xumingkuan just in case you need this for making SFG figures...

----


xumingkuan (approved):
Thanks!

PR #1874, opened 2020-09-15
[refactor] Create class TypePromotionMapping for refactoring function promoted_type in lang_util.cpp

Hanke98:
Related issue = #1750

Hi, I tried refactoring function promoted_type by creating a class named TypePromotionMapping and use a global variable for initializing the type map. But I am not sure whether it is good method. Should I use a singleton mode？

### DIFF
--- a/taichi/lang_util.cpp
+++ b/taichi/lang_util.cpp
@@ -324,7 +324,44 @@ bool command_exist(const std::string &command) {
 #endif
 }
 
+namespace {
+class TypePromotionMapping {
+ public:
+  TypePromotionMapping() {
+#define TRY_SECOND(x, y)                                            \
+  mapping[std::make_pair(get_data_type<x>(), get_data_type<y>())] = \
+      get_data_type<decltype(std::declval<x>() + std::declval<y>())>();
+#define TRY_FIRST(x)      \
+  TRY_SECOND(x, float32); \
+  TRY_SECOND(x, float64); \
+  TRY_SECOND(x, int8);    \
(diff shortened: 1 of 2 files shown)


yuanming-hu (commented):
> But I am not sure whether it is good method. Should I use a singleton mode？

You are going great! The implementation mostly LGTM. Just one nit: it would be nice to move the class to `lang_util.cpp` and put it in an anonymous namespace. By doing this we 1) avoid unnecessary compilation of the class 2) avoid exposing the class symbol to other files (symbols in anonymous namespaces will not be exported in C++).

See https://github.com/taichi-dev/taichi/blob/953694161007cc60f447a211cafe3fe33836b1ad/taichi/util/short_name.cpp for an example.

Hanke98:
> > But I am not sure whether it is good method. Should I use a singleton mode？
> 
> You are going great! The implementation mostly LGTM. Just one nit: it would be nice to move the class to `lang_util.cpp` and put it in an anonymous namespace. By doing this we 1) avoid unnecessary re-compilations of the class 2) avoid exposing the class symbol to other files (symbols in anonymous namespaces will not be exported in C++).
> 
> See
> 
> https://github.com/taichi-dev/taichi/blob/953694161007cc60f447a211cafe3fe33836b1ad/taichi/util/short_name.cpp#L5
> 
> for an example.

Thanks a lot for your advice!! I moved the class into a anonymous namespace in `lang_util.cpp`.

yuanming-hu (approved):
Thank you so much! LGTM.

PR #1875, opened 2020-09-15
[async] Cache fusion results

xumingkuan:
Related issue = #742 

Move a part of `StateFlowGraph::fuse()::do_fuse()` to `IRBank::fuse()`.

----


yuanming-hu (approved):
Thanks! In the future, we may want to avoid having `IRBank` do too many things but for now we are good.

PR #1884, opened 2020-09-20
[async] Add element-wise info into TaskMeta for fusion

xumingkuan:
Related issue = #742 

Changes: 
- Add `std::unordered_map<SNode *, bool> TaskMeta::element_wise`
- Refactor `AsyncStateHash` into `std::hash<AsyncState>` because I think we don't need other hashes of `AsyncState`
- Fix typo "fusable" -> "fusible"

Question:
- How does the element-wise info affect mask/list states?

----

### DIFF
--- a/taichi/ir/ir.cpp
+++ b/taichi/ir/ir.cpp
@@ -406,6 +406,27 @@ GlobalPtrStmt::GlobalPtrStmt(const LaneAttribute<SNode *> &snodes,
   TI_STMT_REG_FIELDS;
 }
 
+bool GlobalPtrStmt::is_element_wise(SNode *snode) const {
+  if (snode == nullptr) {
+    // check all snodes
+    for (const auto &snode_i : snodes.data) {
+      if (!is_element_wise(snode_i)) {
+        return false;
+      }
+    }
+    return true;
+  }
+  // check if this statement is element-wise on snode
+  for (int i = 0; i < (int)indices.size(); i++) {
+    if (auto loop_index_i = indices[i]->cast<LoopIndexStmt>();
+        !(loop_index_i && loop_index_i->loop->is<OffloadedStmt>() &&
+          loop_index_i->index == snode->physical_index_position[i])) {
+      return false;
+    }
+  }
+  return true;
+}
+
 std::string GlobalPtrExpression::serialize() {
   std::string s = fmt::format("{}[", var.serialize());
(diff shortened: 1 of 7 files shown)


yuanming-hu (approved):
Awesome!! LGTM.

> How does the element-wise info affect mask/list states?

Great question. The short answer is we don't need to worry about it. 
If a write access is element-wise, the activation will be demoted here: https://github.com/taichi-dev/taichi/blob/f2a798d185543c7d35cdb0860c98315286bf6d3f/taichi/transforms/flag_access.cpp#L114

So by the time you are doing the `is_element_wise` test, whether a task needs list generation/gc/mask state output is already determined. We don't need to handle that once more. 

(Sorry I didn't notice we already have a similar implementation of `is_element_wise` there - maybe we can unify two implementations sometime in the future.)

xumingkuan:
> If a write access is element-wise, the activation will be demoted here:
> 
> https://github.com/taichi-dev/taichi/blob/f2a798d185543c7d35cdb0860c98315286bf6d3f/taichi/transforms/flag_access.cpp#L114
> 
> (Sorry I didn't notice we already have a similar implementation of `is_element_wise` there - maybe we can unify two implementations sometime in the future.)

Interesting! It looks like the element-wise check in `flag_access` also includes a sparsity check. For example, if `x` and `y` have different shapes on the parent sparse SNode (`x` is smaller than `y`) or they have the same shape but their parent sparse SNodes are different, `y` will not be considered element-wise in
```python
for I in ti.grouped(x):
    y[I] = 1
```
(`y[I]` needs activation here). However, for fusion, we **can** consider the task as element-wise on `y`.

A fusible example:
```python
x = ti.field(ti.i32)
y = ti.field(ti.i32)
ti.root.pointer(ti.i, 4).place(x)
ti.root.pointer(ti.i, 8).place(y)

@ti.kernel
def foo():
    for i in x:
        y[i] = 1

@ti.kernel
def bar():
    for i in x:
        x[i] = y[i]

foo()
bar()

ti.sync()
```

Now I think the mask/list states have the same criteria as value states here -- if task a has an edge to task b (struct for or range for tasks), no matter which type the edge's state is, only if both tasks are element-wise on the edge's SNode, can a and b be fused.

xumingkuan:
Why `Some checks haven’t completed yet` -- `Build and Test (CPU) (ubuntu-latest, 3.6, OFF) Expected — Waiting for status to be reported`?

PR #1893, opened 2020-09-26
[export] Support exporting custom config variables

archibate:
Related issue = #394

----
This would allows user to specify configuration direct-in-place in Python.
I need this for TaichiHub, an online playground, for customizable particle numbers & screen size.
Please merge this soon, idealy before next release.

### DIFF
--- a/python/taichi/cc_compose.py
+++ b/python/taichi/cc_compose.py
@@ -124,8 +124,7 @@ class ComposerCC(ComposerBase):
         if isinstance(value, str):
             declaration = f'const char {variable}[]'
             value = repr(value)
-            value[0] = '"'
-            value[-1] = '"'
+            value = '\"' + value[1:-1] + '\"'
         elif isinstance(value, int):
             declaration = f'const int {variable}'
         else:
--- a/python/taichi/cc_compose.py
+++ b/python/taichi/cc_compose.py
@@ -117,6 +117,23 @@ class ComposerCC(ComposerBase):
         declaration = source.split('{', 1)[0].strip()
         self.emit_header(f'extern {declaration};')
 
+    def do_config(self, e):
+        key = e['key']
+        value = e['value']
+        variable = f'Ti_cfg_{key}'
+        if isinstance(value, str):
+            declaration = f'const char {variable}[]'
+            value = repr(value)
+            value[0] = '"'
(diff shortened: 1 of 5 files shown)


Hanke98 (changes requested):
Hi Yubin, I think it's a really good idea to log some custom configurations and most of your code LGTM.

When I was testing the code, I wrote a simple test script:
`
import taichi as ti
ti.start_recording('test_record.yml')
ti.init(arch=ti.cc)
ti.record_action_config('num_particles', 8192)
n_particles = 8192

x = ti.Vector.field(2, float, n_particles)


@ti.kernel
def init():
    for i in range(n_particles):
        x[i] = [ti.random() * 0.4 + 0.2, ti.random() * 0.4 + 0.2]


init()
ti.stop_recording()
`

And I got:


[I 09/26/20 11:51:25.204] [action_recorder.cpp:start_recording@27] ActionRecorder: start recording to [test_record.yml]
[Taichi] Starting on arch=cc
  File "test_record.py", line 4, in <module>
    ti.record_action_config('num_particles', 8192)
  File "/home/ljf/playground/taichi/python/taichi/core/record.py", line 16, in record_action_config
    record_action_entry('config', {'key': key, 'value': value})
  File "/home/ljf/playground/taichi/python/taichi/core/record.py", line 6, in record_action_entry
    ti_core.record_action_entry(name, list(contents.items()))
RuntimeError: Unexpected index


So, could you please have a look at that? Thank you very much!

python/taichi/cc_compose.py
```diff
@@ -117,6 +117,23 @@ def do_compile_kernel(self, e):
...
 
+    def do_config(self, e):
+        key = e['key']
+        value = e['value']
+        variable = f'Ti_cfg_{key}'
+        if isinstance(value, str):
+            declaration = f'const char {variable}[]'
+            value = repr(value)
+            value[0] = '"'
+            value[-1] = '"'
```
Hanke98, line 128: Please note that `str` object cannot be assigned in this way. Maybe using slice is a better way. : )
```suggestion
            value = '\"' + value[1:-1] + '\"'
```

Hanke98 (approved):
Great! Thanks again! Learnt a lot from your code!

PR #1896, opened 2020-09-26
[refactor] Remove DFS in DOT generation

k-ye:
No point doing DFS if simply iterating over `nodes_` can get the work done..

----


xumingkuan (approved):
LGTM!

Issue #933, opened 2020-05-09
[refactor] Reduce compilation time by splitting and shortening headers


archibate:
**Concisely describe the proposed feature**
Currently compliling taichi takes too long time can be a great resistance for contributors especially when your work involving common headers like `ir.h` or `lang_util.h` (I guess @xumingkuan suffers this most).

**Describe the solution you'd like (if any)**
1. reduce length of these common headers by removing unused functions and classes.
2. move some of the inline functions in `.h` to `.c`, and then `.h` only holds function declaration.
3. split a large header like `ir.h` into multiple headers, since only part of the functions are used.
4. some common utils may use macros instead of inline functions to reduce c++ lexer's pain.

**Additional comments**
It's slowing down recently. Not sure what added to these headers caused this.

TODO: add compile-time test/comparison result here.

Update: I also found that `clang-8` is 5x slower than `clang-7`.

xumingkuan:
Thanks for proposing this! I think we can try to remove this as the first task:

https://github.com/taichi-dev/taichi/blob/d4a9c18ce3fd39a3c7e448825c9c8ff38a342295/taichi/ir/ir.h#L1555-L1556

It's virtually increasing the length of ir.h.

k-ye:
Covered in #2203

PR #1898, opened 2020-09-26
[Refactor] [ir] Move statements from ir.h to statements.h

xumingkuan:
Related issue = #933

Changes:
- Move all `XxxStmt` from `ir.h` to `statements.h`;
- Move all implementations of frontend statements from `ir.cpp` to `frontend_ir.cpp`;
- Remove `#include "statements.h"` from `visitors.h`.

Something that may be useful to reduce the compilation time of the Taichi compiler but I don't know how to do:
- We need `OffloadedStmt::TaskType` in some files (`async_utils.h` for example). For now, we can only include `statements.h`, which includes `ir.h`. `statements.h` needs `ir.h` because it needs the implementation of `Stmt`, which also needs the implementation of `IRNode`. However, `ir.h` may become too short if even `IRNode` is in `statements.h`. So shall we separate `OffloadedStmt::TaskType` in a light-weighted file which doesn't need `statements.h`?

----


k-ye (approved):
LGTM!

> So shall we separate OffloadedStmt::TaskType in a light-weighted file which doesn't need statements.h

+N on this

yuanming-hu (approved):
Awesome!!

> > So shall we separate OffloadedStmt::TaskType in a light-weighted file which doesn't need statements.h
> 
> +N on this

Maybe we can create an `ir_fwd.h` file that contains all the existing forward declarations https://github.com/taichi-dev/taichi/blob/d266c0faab7231b2d0a7925c6a9f77b2e10c4f93/taichi/ir/ir.h#L20-L34

and a new `OffloadedTaskType` (which used to be `OffloadedStmt::TaskType`)?

k-ye:
> and a new OffloadedTaskType (which used to be OffloadedStmt::TaskType)?

SGTM! (But i'd suggest to separate `OffloadedTaskType` from `ir_fwd.h`... I think it's even fine to have a header just for this type)

PR #1899, opened 2020-09-26
[misc] Make check previous CI run only check for Build-and-Test jobs

k-ye:
When the previous CI run's code format check fails, two thing happens:

1. @taichi-gardener will do a `[skip ci] enforce code format` to automatically correct the things
2. The second commit would still fail because it runs "Check the previous CI workflow" job, which finds that the previous workflow... has failed --> this time it fails again :-/

"Persubmit Checks / Code Format (pull_request)" is triggered unconditionally now, because it is a required check. (Previously it is disabled for @taichi-gardener 's commits)

Related issue = #1809

----

### DIFF
--- a/.github/workflows/persubmit.yml
+++ b/.github/workflows/persubmit.yml
@@ -109,7 +109,8 @@ jobs:
   code_format:
     name: Code Format
     runs-on: ubuntu-latest
-    if: ${{ !contains(github.event.pull_request.requested_reviewers.*.login, 'taichi-gardener') }}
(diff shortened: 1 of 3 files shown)


xumingkuan (approved):
Cool, thanks!

k-ye:
Seems like we have one more thing to be addressed: "Build and Test (CPU) (ubuntu-latest, 3.6, OFF)" is a required check for merging, but it's skipped for @taichi-gardener 's commits. I probably need to add another job to gather the test result, and make that a required check instead...

PR #1902, opened 2020-09-27
[perf] Wrap clear loss/grad into a kernel

k-ye:
Let me submit this part first. With this simple tweak MPM lagrangian force on my end already runs from 15fps to 24fps...

Related issue = #N/A

----


archibate (approved):
:+1: for make Taichi.js export process a bit easier!

PR #1908, opened 2020-09-29
[Example] Fix rounding error for cornell_box.py

lyd405121:
[fix]
fix rounding error artifact for cornell_box

Related issue = #

----

### DIFF
--- a/examples/cornell_box.py
+++ b/examples/cornell_box.py
@@ -335,6 +335,7 @@ def sample_direct_light(hit_pos, hit_normal, hit_color):
     direct_li = ti.Vector([0.0, 0.0, 0.0])
     fl = lambertian_brdf * hit_color * light_color
     light_pdf, brdf_pdf = 0.0, 0.0
+    
     # sample area light
     to_light_dir = sample_area_light(hit_pos, hit_normal)
     if to_light_dir.dot(hit_normal) > 0:
@@ -346,6 +347,20 @@ def sample_direct_light(hit_pos, hit_normal, hit_color):
                 w = mis_power_heuristic(light_pdf, brdf_pdf)
                 nl = dot_or_zero(to_light_dir, hit_normal)
                 direct_li += fl * w * nl / light_pdf
+
+
+    # sample brdf
+    brdf_dir = sample_brdf(hit_normal)
+    brdf_pdf = compute_brdf_pdf(hit_normal, brdf_dir)
+    if brdf_pdf > 0:
+        light_pdf = compute_area_light_pdf(hit_pos, brdf_dir)
+        if light_pdf > 0:
+            l_visible = visible_to_light(hit_pos, brdf_dir)
+            if l_visible:
+                w = mis_power_heuristic(brdf_pdf, light_pdf)
+                nl = dot_or_zero(brdf_dir, hit_normal)
+                direct_li += fl * w * nl / brdf_pdf
+
     return direct_li
 
 
--- a/examples/cornell_box.py
+++ b/examples/cornell_box.py
@@ -347,7 +347,7 @@ def sample_direct_light(hit_pos, hit_normal, hit_color):
                 nl = dot_or_zero(to_light_dir, hit_normal)
                 direct_li += fl * w * nl / light_pdf
     '''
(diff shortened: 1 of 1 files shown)


examples/cornell_box.py
```diff
@@ -354,6 +358,7 @@ def sample_direct_light(hit_pos, hit_normal, hit_color):
                 w = mis_power_heuristic(brdf_pdf, light_pdf)
                 nl = dot_or_zero(brdf_dir, hit_normal)
                 direct_li += fl * w * nl / brdf_pdf
+    '''
```
archibate, line 361: Why is this part commented?

examples/cornell_box.py
```diff
@@ -229,7 +229,10 @@ def intersect_scene(pos, ray_dir):
 
 @ti.func
 def visible_to_light(pos, ray_dir):
-    a, b, c, mat = intersect_scene(pos, ray_dir)
+    #eps*ray_dir is easy way to prevent rounding error
+    #here is best way to check the float precision:
+    #http://www.pbr-book.org/3ed-2018/Shapes/Managing_Rounding_Error.html
```
archibate, line 234:
```suggestion
    # eps*ray_dir is easy way to prevent rounding error
    # here is best way to check the float precision:
    # http://www.pbr-book.org/3ed-2018/Shapes/Managing_Rounding_Error.html
```
Sounds cool! Just some format nits.

examples/cornell_box.py
lyd405121: for my opinion:
[1] this part sample a reflect direction, so that is indirect lighting, not direct lighting
[2] next iter will count direct lighting as previous indirect lighting, so the radiance is over counted

examples/cornell_box.py
lyd405121: Here is  where the origin idea from in chinese:
https://forum.taichi.graphics/t/cornell-box-bug/1196/6

I am not familiar with git

examples/cornell_box.py
k-ye: Now thinking it back, this is actually doing multiple importance sampling (MIS). See http://www.pbr-book.org/3ed-2018/Light_Transport_I_Surface_Reflection/Direct_Lighting.html#fragment-SampleBSDFwithmultipleimportancesampling-0, so it's not really over counted.

I cannot say that the MIS impl was correct, but could you keep this part, and just tweak the epsilon to see if the dark ring problem goes away?

lyd405121: There is no doubt that you are right!
This implementation is the same with pbrt-v3 ,totally right!

---------------------------advice -----------------------
I have advice for efficience:

change the calculation when hitting a light for depth > 0 and always record the brdfPdf
if hit_light:
    acc_color += throughput * light_color * powerHeuristic(brdfPdf, lightPdf)  / brdfPdf 
    break
Then comment the brdf mis part

--------------------compare------------------------------
[current method]  there are two sample direction, three brdf color eval , one intersect detection for light in one iter 
[anthoer method] one sample dir, two brdf color eval, none for light intersection
where I learn from:
https://github.com/knightcrawler25/Optix-PathTracer/blob/master/src/optixPathTracer/hit_program.cu
https://github.com/knightcrawler25/Optix-PathTracer/blob/master/src/optixPathTracer/light_hit_program.cu

lyd405121: I am so sorry that I commit 5 times for such minor bug······
And seems the 5th commits still fails：
Error: Process completed with exit code 1.

k-ye: >  am so sorry that I commit 5 times for such minor bug
No worries && no need to apologize at all!

Just FYI, the check failed at Code Format https://github.com/taichi-dev/taichi/pull/1908/checks?check_run_id=1187328749. Usually you can add @taichi-gardener  for review, which is a bot that does the code format automatically :)

> I have advice for efficience:

Ah I see. Do I understand correctly that, if the direct light sampling actual hits the light source, then we no longer need to do the BRDF sampling (Sorry I'm no expert on this, and it's been a while since i wrote this example... )?

k-ye (approved):
LGTM thank you!

PR #1912, opened 2020-09-30
[Lang] Add @ti.materialize_callback for functions to be called right after root materialization

archibate:
Related issue = https://github.com/taichi-dev/taichi_three

----
This would allow extension libraries initialize a field on declaration. For example, I have a `Camera` class, I want to initialize the transform matrix with identity matrix by default.

### DIFF
--- a/tests/python/test_runtime.py
+++ b/tests/python/test_runtime.py
@@ -122,13 +122,12 @@ def test_init_arch(arch):
         assert ti.cfg.arch == arch
 
 
-@ti.must_throw(KeyError)
 def test_init_bad_arg():
-    ti.init(_test_mode=True, debug=True, foo_bar=233)
+    with pytest.raises(KeyError):
+        ti.init(_test_mode=True, debug=True, foo_bar=233)
 
 
-@ti.all_archs
-@ti.must_throw(RuntimeError)
+@ti.test(arch=ti.cpu)
 def test_materialization_after_kernel():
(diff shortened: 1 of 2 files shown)


tests/python/test_runtime.py
```diff
@@ -138,27 +137,45 @@ def func():
...
 
     print(x.shape)
 
-    y = ti.field(ti.f32, (2, 3))
+    with pytest.raises(RuntimeError, match='declared after'):
+        y = ti.field(ti.f32, (2, 3))
     # ERROR: No new variable should be declared after Python-scope field access!
+
+
+@ti.test(arch=ti.cpu)
```
k-ye, line 167: nit: `ti.all_archs`?

k-ye (approved):
> I have a Camera class, I want to initialize the transform matrix with identity matrix by default.

Can you be more specific? Why `ti.Matrix.rows(...)` cannot solve the problem?

I think this could be a handy feature, although I don't expect that most users know (or should know) what materialization is -- it's an internal implementation of Taichi.

archibate:
Can you be more specific? Why ti.Matrix.rows(...) cannot solve the problem?

I mean I want to initialize a field on start up.
I want sth like:
```
x = ti.field_from_numpy(...)
y = ti.field_from_numpy(...)
```
instead of:
```
x = ti.field()
y = ti.field()
... (other fields)

x.from_numpy(...)
y.from_numpy(...)
```

PR #1920, opened 2020-10-04
[async] [metal] Support async mode on Metal

k-ye:
Added `Program::compile_to_backend_executable()` to dispatch to each backend's codegen, so that it can be shared between sync/async mode.

While we should only benchmark on CUDA, let me include Metal so that I get a sense of what the performance looks like...

Related issue = #742

----

### DIFF
--- a/taichi/backends/metal/codegen_metal.cpp
+++ b/taichi/backends/metal/codegen_metal.cpp
@@ -3,13 +3,14 @@
 #include <functional>
 #include <string>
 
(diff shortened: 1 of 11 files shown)


xumingkuan (approved):
Cool, LGTM!

xumingkuan (approved):
Cool, LGTM!

yuanming-hu (approved):
Awesome! LGTM. Thank you so much!

PR #1930, opened 2020-10-08
[ir] Add the case of bit_shr to binary_op_type_symbol

xumingkuan:
Related issue = #1871 

`binary_op_type_symbol` is used when printing initial (frontend) IR, and an IR with `bit_shr` will crash if that case is not supported :-)

----


yuanming-hu (approved):
Cool! Thanks.

PR #1943, opened 2020-10-12
[refactor] VectorType -> LegacyVectorType

taichi-gardener:
Related issue = #1905 

Making way for the real `VectorType` in the new type system.

----


Hanke98 (approved):
Great! LGTM! Thanks !

PR #1945, opened 2020-10-12
[async] Do not keep latest state readers in SFG

xumingkuan:
Related issue = #742 

There can be thousands of tasks reading a state.

IIRC we don't need the latest state readers in any of the SFG optimizations?

----


k-ye (approved):
LGTM thanks!

yuanming-hu (approved):
Awesome! LGTM.

PR #1949, opened 2020-10-13
[metal] Support TLS for struct-for tasks

k-ye:
This makes test_bitmasked.py pass..

Related issue = #1941

----


yuanming-hu (approved):
LGTM. Thanks!!

PR #1922, opened 2020-10-05
[GUI] [linux] Support "ti.GUI(fast_gui=True)" for zero-copy while gui.set_image(img)

archibate:
Related issue = depends #1921

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

----
We can now obtain 60fps even with 1920x1080!
This would allow me to implement a real-time soft renderer in [Taichi THREE](https://github.com/taichi-dev/taichi_three)!

PR #1953, opened 2020-10-14
[windows] [gui] Support ti.GUI(fast_gui=True) on Win32

archibate:
Related issue = #1922

----
With no way to confirm if this works as I don't have a win dev env..


k-ye (approved):
Thanks!

PR #1955, opened 2020-10-14
[refactor] Add format_error_message() method

k-ye:
This is just pulling out the error message formatting part used by LLVM backend into a separate function. This way we can share it on other backends :)

Related issue = #677

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

----

PR #1959, opened 2020-10-15
[Metal] Support assert()

k-ye:
Similar to how we've supported `print()`, I reserved a dedicated area in the Metal buffer to store assertion failure messages. The assertion message and print messages are placed in the same buffer for simplicity. Here's the buffer memory layout:

```cpp
// MetalPrintAssertBuffer memory view:
//
// +------------------------+ \
// | AssertRecorderData     | |
// +------------------------+ | -> for assert(), kMetalAssertBufferSize
// | PrintMsg for assert()  | |
// +------------------------+ /
// | PrintMsgAllocator      | \
// +------------------------+ |
// |                        | |
// |                        | |
// | a queue of PrintMsgs   | | -> for print()
// | for print()            | |
// | ... ...                | |
// |                        | |
// +------------------------+ /
```

We only store the error message of the first assertion failure. This is done by using `AssertRecorder::mark_first_failure()`. Under the hood it's just an atomic flag.

Sample generated code:

```py
@ti.kernel
def assert_formatted():
    for i in x:
        assert x[i] == 0, 'x[%d] expect=%d got=%d' % (i, 0, x[i])

```

```cpp
if (!(tmp22)) {
    if (assert_rec_.mark_first_failure()) {
      assert_rec_.set_num_args(4);
      PrintMsg tmp23_msg_(assert_rec_.msg_buf_addr(), 4);
      tmp23_msg_.pm_set_str(/*i=*/0, 2);  // 2 is the index in the string table. We don't store the actual string on GPU.
      tmp23_msg_.pm_set_i32(1, tmp2);
      tmp23_msg_.pm_set_i32(2, tmp11);
      tmp23_msg_.pm_set_i32(3, tmp20);
    }
    return;
  }
```

To answer @archibate question in https://github.com/taichi-dev/taichi/pull/1955#pullrequestreview-508423133, hopefully it's now clear? These backends can all share `format_error_message()`, which does the formatting of `%`.

---

This PR is not urgent at all. Please take your time and consider it low priority if you have other things to deal with...

Related issue = #1955

----

### DIFF
--- a/taichi/backends/metal/codegen_metal.cpp
+++ b/taichi/backends/metal/codegen_metal.cpp
@@ -41,8 +41,9 @@ constexpr char kContextBufferName[] = "ctx_addr";
 constexpr char kContextVarName[] = "kernel_ctx_";
 constexpr char kRuntimeBufferName[] = "runtime_addr";
 constexpr char kRuntimeVarName[] = "runtime_";
-constexpr char kPrintBufferName[] = "print_addr";
+constexpr char kPrintAssertBufferName[] = "print_assert_addr";
 constexpr char kPrintAllocVarName[] = "print_alloc_";
+constexpr char kAssertRecorderVarName[] = "assert_rec_";
 constexpr char kLinearLoopIndexName[] = "linear_loop_idx_";
 constexpr char kListgenElemVarName[] = "listgen_elem_";
 constexpr char kRandStateVarName[] = "rand_state_";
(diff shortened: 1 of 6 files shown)


yuanming-hu (approved):
Great! LGTM. Hopefully we can handle assertions uniformly on all backends.

I guess for printing, on CPU and CUDA the built-in *synchronous* printing support is still helpful. On CUDA the `printf` support is kind of limited, so we may want to use the print buffer as well in the future. We may also want to support Python-style (or C++20 `std::format`-style) formatting on the backends :-)

PR #1964, opened 2020-10-16
[misc] Set CMAKE_CXX_STANDARD to 17

xumingkuan:
This is to let CLion recognize C++17 features... Without this line (`set(CMAKE_CXX_STANDARD 17)`), CLion keeps complaining about stuff like `std::optional` does not name a type.

See also https://stackoverflow.com/questions/53505844/clion-does-not-recognise-c17-features.

----


yuanming-hu (approved):
Cool! I guess it may make more sense to move this line to the very beginning of the root `CMakeLists.txt`, since this option applies to the entire project?

As a CLion user on Linux, I didn't run into the warning. It seems to be platform-specific.

PR #1967, opened 2020-10-17
[Type] [refactor] Remove LegacyVectorType

Hanke98:
Related issue = #1905 

In this pr, we remove the class `LegacyVectorType` and use a static function `TypeFactory::create_vector_or_scalar_type` instead.

----


taichi/ir/type_factory.cpp
```diff
@@ -36,4 +36,16 @@ Type *TypeFactory::get_pointer_type(Type *element) {
 TypeFactory::TypeFactory() {
 }
 
+DataType TypeFactory::create_vector_or_scalar_type(int width,
+                                                   DataType element,
+                                                   bool element_is_pointer) {
```
Hanke98, line 41: This is the core change, we use this function to replace the `LegacyVectorType`

yuanming-hu (approved):
Thanks! LGTM.

PR #1969, opened 2020-10-17
[Error] [ipython] Use print directly instead of builtin warnings module to prevent being supressed

archibate:
Related issue = https://github.com/taichi-dev/taichi/issues/677#issuecomment-710742766

----
It seems that if I use IPython (`ti repl`), the warning is depressed. If I use raw Python shell (`python`), the warning is shown.
So IPython is setting `warnings.filterwarnings('ignore')` on their own? Anyway, the built-in warning system seems unreliable.
Let's switch to standard `print` for warning instead.

### DIFF
--- a/python/taichi/misc/util.py
+++ b/python/taichi/misc/util.py
@@ -110,18 +110,18 @@ def get_line_number(asc=0):
     return inspect.stack()[1 + asc][2]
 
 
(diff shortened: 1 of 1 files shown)


yuanming-hu (approved):
Thanks! I don't exactly know what happened with IPython, but it works on my end.

PR #1973, opened 2020-10-18
[async] Add allocator async state

k-ye:
https://github.com/taichi-dev/taichi/blob/ceea597141bd1f0f0b2a380e5dabd50fb6ed04a6/taichi/lang_util.cpp#L156-L158

Before this PR, `test_block_async()` would actually fail. Here's the SFG. Note that all GC tasks are floating (as expected):

no-alloc

With this PR, SFG:

alloc

Highlighting the `allocator` state:

alloc-single

Related issue = #742

----

### DIFF
--- a/tests/python/test_gc.py
+++ b/tests/python/test_gc.py
@@ -2,7 +2,7 @@ import taichi as ti
 
 
 @ti.test(require=ti.extension.sparse)
-def test_block():
+def _test_block_gc():
     N = 100000
 
     dx = 1 / 128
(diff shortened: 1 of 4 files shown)


xumingkuan (commented):
LGTM in general! I think we should also change some code in task fusion:
https://github.com/taichi-dev/taichi/blob/b8a1bc4db092bd471a32cc4c3f75103030330b8a/taichi/program/state_flow_graph.cpp#L455-L458

IIUC allocator states should be treated like value states here: if A->B is an allocator edge on SNode `s`, only if both A and B are element-wise (or loop-unique on the same value in the future), can A and B be fused. For example, these two tasks cannot be fused:
```python
@ti.kernel()
def foo():
  for i in x:
    ti.activate(y, i + 1)

@ti.kernel()
def bar():
  for i in x:
    ti.deactivate(y, i)
```
BTW, do these `SNodeOpStmt`s change mask states?

k-ye:
> LGTM in general! I think we should also change some code in task fusion:
> https://github.com/taichi-dev/taichi/blob/b8a1bc4db092bd471a32cc4c3f75103030330b8a/taichi/program/state_flow_graph.cpp#L455-L458
>
> IIUC allocator states should be treated like value states here: if A->B is an allocator edge on SNode `s`, only if both A and B are element-wise (or loop-unique on the same value in the future), can A and B be fused. For example, these two tasks cannot be fused:
> 
> ```python
> @ti.kernel()
> def foo():
>   for i in x:
>     ti.activate(y, i + 1)
> 
> @ti.kernel()
> def bar():
>   for i in x:
>     ti.deactivate(y, i)
> ```

Great catch! I've updated the code accordingly. However, note that I checked for `mask` state instead of `allocator`. IMHO, `mask` is directly related to sparsity, element_wise, etc. WDYT?

> BTW, do these `SNodeOpStmt`s change mask states?

Yes, thanks!

tests/python/test_gc.py
```diff
@@ -79,3 +94,38 @@ def test_pointer_gc():
...
+    @ti.kernel
+    def deactivate_y():
+        for i in x:
+            ti.deactivate(y_parent, i)
+
+    activate_y()
+    deactivate_y()
+    ti.sync()
+
+    # TODO: assert that activate_y and deactivate_y are not fused.
```
k-ye, line 126: I've confirmed this with the dumped dot graph. But task fusion is currently not recorded in the stats yet.

xumingkuan (approved):
Cool, thanks!

PR #1975, opened 2020-10-20
[Type] [refactor] Implement is_primitive and refactor primitive type equality check

TH3CHARLie:
Related issue = #1905

This PR implements `Type::is_primitive`, a member function to check if a type is of given PrimitiveType. This PR also replaces all `dt == PrimitiveType::xx` with `dt->is_primitive(PrimitiveTypeID::xx`, a step further towards obsoleting `DataType`

----


taichi/ir/type.cpp
```diff
@@ -75,4 +75,12 @@ int Type::vector_width() const {
...
 }
 
+bool Type::is_primitive(PrimitiveTypeID type) const {
+  if (auto p = cast<PrimitiveType>()) {
+    return p->type == type;
+  } else {
+    return false;
+  }
+}
+
```
TH3CHARLie, line 85: core change.

Besides this function, everything else is just a simple replacement. Though I've checked every replacement, I would like the reviewer to do it again and thanks in advance!

Hanke98 (approved):
Great Job! I did a search with regex again and went over all the changes here. The result shows that you refactoring is exactly right! Thanks very much for the careful work !

yuanming-hu (approved):
Thanks! LGTM.

PR #1991, opened 2020-10-23
[Example] Interpolate vertices for mciso_advanced.py to make it smoother

archibate:
Related issue = #

----
References: https://www.cnblogs.com/shushen/p/5542131.html

image
image

We can even use `taichi_three` for 3D visualization iapr.

Btw, is that possible to avoid re-computing the same vertex shared by four triangles? I.e. skip 'Weld' in Blender?

### DIFF
--- a/examples/mciso_advanced.py
+++ b/examples/mciso_advanced.py
@@ -295,6 +295,7 @@ class MCISO:
 
         et = [self.et2, self.et3][dim - 2]
         self.et = ti.Vector.field(dim, int, et.shape[:2])
+
         @ti.materialize_callback
         def init_et():
             self.et.from_numpy(et)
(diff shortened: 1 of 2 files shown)


examples/mciso_advanced.py
```diff
@@ -344,16 +347,91 @@ def march(self) -> int:
...
+                            p = (1 - self.m[i + 1, j]) / (
+                                self.m[i + 1, j + 1] - self.m[i + 1, j])
+                            R = [i + 1, j + p]
+                        elif e == 3:
+                            p = (1 - self.m[i, j + 1]) / (
+                                self.m[i + 1, j + 1] - self.m[i, j + 1])
+                            R = [i + p, j + 1]
+                    else:
+                        i, j, k = I
+                        # (.5, 0, 0), (1, .5, 0), (.5, 1, 0), (0, .5, 0)
```
k-ye, line 380: Just an idea, it might be easier for you to maintain if these coordinate offsets are put into a field. You can then index them with `e`?

examples/mciso_advanced.py
```diff
@@ -408,20 +488,22 @@ def main(self, export_result=None):
...
-                    writer.export_frame(gui.frame, export_result)
+                    writer.export('mciso_output.ply')
+                    print('Mesh saved to mciso_output.ply')
             gui.show()
 
 
 if __name__ == '__main__':
-    ti.init(arch=ti.gpu)
-    main = MCISO_Example()
+    ti.init(arch=ti.cpu)
```
k-ye, line 507: Is this change intentional?

k-ye (approved):
Great thanks!

PR #1992, opened 2020-10-23
[blender] Fix inspect.stack() not working in Blender with sourceinspect 0.0.4

archibate:
Related issue = #

----
I found that ODOP is not working properly in Blender.


k-ye (approved):
Thanks!

PR #1999, opened 2020-10-26
[Example] Fix euler.py example

Eydcao:
Related issue = #

----

Fix the misuse of `atomic_max` in the compressible flow example


yuanming-hu (approved):
Awesome! Thank you so much for fixing this! Finally `euler.py` is working again!

PR #2000, opened 2020-10-26
[metal] Add SNodeRep_pointer

k-ye:
`SNodeRep_pointer` is functionally equivalent to LLVM's https://github.com/taichi-dev/taichi/blob/master/taichi/runtime/llvm/node_pointer.h. There are a few implementation details specific to Metal, though:

* Like the LLVM backend, each cell of a `pointer` SNode stores only the *pointer* to the data, not the data itself. However, unlike LLVM, this is not a real pointer, but merely an `int32_t` that indexes into somewhere in the Metal buffer memory. The `NodeManager` instance passed to this Rep's `init()` knows how to recover the data address from this `int32_t`. See `child_or_ambient_addr()`.
* In #1994, I added `NodeManager::ElemIndex`, which has a notion of `raw` vs `index` value. They differ by `kIndexOffset`. 
https://github.com/taichi-dev/taichi/blob/16af509ad5d6d95a19ff217105e22f3cace30a0b/taichi/backends/metal/shaders/runtime_structs.metal.h#L89-L99
This shift is necessary for `ElemIndex` to know if its `raw_` is a valid index in the memory or not. A valid `raw_` value is always greater than or equal to `kIndexOffset`, while values below that represents special values that are necessary for the memory allocation (#1174) to work (`0`: `nullptr`; `1`: allocation in progress).

* A `pointer` SNode cell stores the **raw** value of its `ElemIndex`, which is initialized to `0`. Inside `activate()`, if the allocation succeeds, it will store `nm_idx_raw = nm_.allocate().raw()` into the cell. As a result, in `to_nodemgr_idx()`, we read the cell's `int32_t`, and uses `ElemIndex::from_raw()`, instead of `ElemIndex::from_index()`. Same for `deactivate()`.

Related issue = #1740 

----

### DIFF
--- a/taichi/backends/metal/shaders/runtime_utils.metal.h
+++ b/taichi/backends/metal/shaders/runtime_utils.metal.h
@@ -308,6 +308,79 @@ STR(
       int32_t meta_offset_ = 0;
     };
 
+    class SNodeRep_pointer {
+     public:
+      using ElemIndex = NodeManagerData::ElemIndex;
+
+      void init(device byte * addr, NodeManager nm, ElemIndex ambient_idx) {
+        addr_ = addr;
+        nm_ = nm;
+        ambient_idx_ = ambient_idx;
+      }
+
+      device byte *child_or_ambient_addr(int i) {
+        auto nm_idx = to_nodemgr_idx(addr_, i);
+        nm_idx = nm_idx.is_valid() ? nm_idx : ambient_idx_;
+        return nm_.get(nm_idx);
+      }
+
+      inline bool is_active(int i) { return is_active(addr_, i); }
+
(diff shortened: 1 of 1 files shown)


yuanming-hu (approved):
LGTM! 

Just curious: it seems that Metal doesn't support 64-bit data types, but there are indeed pointers in the code. Are all pointers 32-bit in Metal?

k-ye:
> Just curious: it seems that Metal doesn't support 64-bit data types, but there are indeed pointers in the code. Are all pointers 32-bit in Metal?

I believe the pointers are still 64bit, we just don't have 64bit data support (at least not in the Metal buffer :-/ )

PR #2002, opened 2020-10-27
[async] [bug] Fix a bug in fusion and add some debug info

xumingkuan:
Related issue = #742 

Sorry for introducing the subtle bug... 
(It shouldn't cause undefined behavior as `has_path` is a square matrix. But it can make SFG malformed.)

Also adds some assert_info that may be useful during debugging.

----


k-ye (approved):
Great thanks!

Issue #1931, opened 2020-10-08
[opengl] [test] "test_bit_operations::test_bit_shr" fails on opengl


xumingkuan:
**Describe the bug**
`ti.bit_shr` is demoted to `bit_sar` on opengl, but the data type becomes `u32`, which is not supported on opengl...

**To Reproduce**
`ti test bit_operations -v -t 1`.

```py
@ti.test()
def test_bit_shr():
    @ti.kernel
    def shr(a: ti.i32, b: ti.i32) -> ti.i32:
        return ti.bit_shr(a, b)

    n = 8
    test_num = 2**n
    neg_test_num = -test_num
    for i in range(n):
        assert shr(test_num, i) == 2**(n - i)
    for i in range(n):
        offset = 0x100000000 if i > 0 else 0
        assert shr(neg_test_num, i) == (neg_test_num + offset) >> i
```

**Log/Screenshots**
```
..\tests\python\test_bit_operations.py::test_bit_shr[opengl] [Taichi] Starting
 on arch=opengl
[Taichi] materializing...
[I 10/08/20 18:27:05.793] [compile_to_offloads.cpp:taichi::lang::irpass::`anon
ymous-namespace'::make_pass_printer::<lambda_fe1d620add3df83d4ee306f9c0ab10ca>
::operator ()@18] [shr_c14_0] Initial IR:
kernel {
  <i32 x1> $0 : kernel return (cast_value<int32> (arg[0] (dt=int32) shr arg[1]
 (dt=int32)))
}
[I 10/08/20 18:27:05.794] [compile_to_offloads.cpp:taichi::lang::irpass::`anon
ymous-namespace'::make_pass_printer::<lambda_fe1d620add3df83d4ee306f9c0ab10ca>
::operator ()@18] [shr_c14_0] Lowered:
kernel {
  <i32 x1> $0 = arg[0]
  <i32 x1> $1 = arg[1]
  $2 = bit_shr $0 $1
  $3 = cast_value<i32> $2
  <i32 x1> $4 : kernel return $3
}
[I 10/08/20 18:27:05.794] [compile_to_offloads.cpp:taichi::lang::irpass::`anon
ymous-namespace'::make_pass_printer::<lambda_fe1d620add3df83d4ee306f9c0ab10ca>
::operator ()@18] [shr_c14_0] Typechecked:
kernel {
  <i32 x1> $0 = arg[0]
  <i32 x1> $1 = arg[1]
[...]

PR #2007, opened 2020-10-30
[OpenGL] [perf] Utilize glDispatchComputeIndirect to prevent sync when dynamic ranges are used

archibate:
Related issue = [数据删除]

----
## What this PR do?

In `master` we sync back the gtmp buffer to compute the `num_groups`, and invoke `glDispatchCompute` with it.
In this PR we use `glDispatchComputeIndirect` to let OpenGL directly fetch `num_groups` from GPU memory.
This prevents a force sync when dynamic range-for is used, resulting in a better performance.

## Why I use indirect-evaluator instead of utilizing grid-stride-loop?

OpenGL CS is completely different from CUDA and Metal.
So the term **work group** they claimed can be implemented differently from the CUDA grid.
**Using a fixed work group size + grid-stride loop would harm performance**. 
Instead, OpenGL provides another API for dynamic work group size, which CUDA don't have IMO:
https://www.khronos.org/opengl/wiki/GLAPI/glDispatchComputeIndirect
This is already a sign that OpenGL officially recommend us not to make iteration body small, and scale work group dynamically.
I'll merge this as soon as CI passed to push our ultra-secret project forward.

## Test status

All tests except for the `test_bit_shl` (#1931) passed on my NVIDIA card, which is already a known issue in latest `master`.

## Example

```py
import taichi as ti
import timeit

ti.init(ti.opengl)

N = 2**14

x = ti.field(int, N)

@ti.kernel
def func():
    for i in x:
        x[i] = i
    for i in range(x[2]):
        x[i] = i

stmt = lambda: func()
print(timeit.timeit(stmt, stmt, number=10000))
```
master: 1.0291837109998596
this PR: 0.1320957290008664


k-ye (commented):
LGTM, thank you!

> This is already a sign that OpenGL officially recommend us not to make iteration body small, and scale work group dynamically.

Interesting. IIUC, the saving still comes most from the removal of the extra sync rather than the matched size. So even if the work group size matches perfectly, and is significantly larger than the static grid-strided-loop size, it's still bounded by the hardware to divide the work into batches..

k-ye (approved):

yuanming-hu (approved):
Cool, thanks! LGTM.

PR #2010, opened 2020-10-31
[sparse] Fix allocator initialization

yuanming-hu:
The wrong allocators may be initialized.

----


xumingkuan (approved):
LGTM, thanks!

PR #2015, opened 2020-11-01
[metal] Support pointer SNode in codegen

k-ye:
This teaches the Metal codegen how to use `SNodeRep_pointer` in the generated classes. For example:

```cpp
struct S1 {
  // pointer
  constant static constexpr int n = 4096;
  constant static constexpr int elem_stride = /*pointer ElemIndex=*/4;
  constant static constexpr int stride = elem_stride * n;

  S1(device byte *addr, device Runtime *rtm, device MemoryAllocator *ma) {
    NodeManager nm;
    nm.nm_data = (rtm->snode_allocators + 1);  // 1 is the SNode ID
    nm.mem_alloc = ma;
    const auto amb_idx = rtm->ambient_indices[1];  // 1 is the SNode ID
    rep_.init(addr, nm, amb_idx);
  }

  S1_ch children(int i) {
    return {rep_.child_or_ambient_addr(i)};
  }

  inline bool is_active(int i) {
    return rep_.is_active(i);
  }

  inline void activate(int i) {
    rep_.activate(i);
  }

  inline void deactivate(int i) {
    rep_.deactivate(i);
  }

  SNodeRep_pointer rep_;
};
```

It also handles the `listgen` tasks properly for `pointer` SNodes now. I've inlined some explanation/docs in the Metal shader files themselves...

Things that will be added later:

* Initializing the node allocators, ambient elements are runtime
* GC

Related issue = #1740, #1174 

----

### DIFF
--- a/taichi/backends/metal/shaders/runtime_structs.metal.h
+++ b/taichi/backends/metal/shaders/runtime_structs.metal.h
@@ -145,7 +145,9 @@ STR(
       };
       BelongedNodeManager belonged_nodemgr;
 
-      inline bool in_root_buffer() const { return belonged_nodemgr.id < 0; }
+      inline bool in_root_buffer() const {
+        return belonged_nodemgr.id < 0;
+      }
     };
     // clang-format off
 )
--- a/taichi/backends/metal/shaders/runtime_structs.metal.h
+++ b/taichi/backends/metal/shaders/runtime_structs.metal.h
@@ -99,7 +99,13 @@ STR(
 
     // This class is very similar to metal::SNodeDescriptor
     struct SNodeMeta {
-      enum Type { Root = 0, Dense = 1, Bitmasked = 2, Dynamic = 3 };
+      enum Type {
+        Root = 0,
+        Dense = 1,
+        Bitmasked = 2,
+        Dynamic = 3,
+        Pointer = 4,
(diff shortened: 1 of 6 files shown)


taichi/backends/metal/shaders/runtime_structs.metal.h
```diff
@@ -126,10 +132,20 @@ STR(
...
+        // If -1, the memory where this cell lives isn't in a particular
+        // NodeManager's dynamically allocated memory. Instead, it is at a fixed
+        // location in the root buffer.
+        //
+        // For {dense, bitmasked}, this should always be -1.
+        int32_t id = -1;
+        // Index of the element within the NodeManager.
+        NodeManagerData::ElemIndex elem_idx;
+      };
+      BelongedNodeManager belonged_nodemgr;
```
k-ye, line 146: This encodes the info of whether the cell represented by this `ListgenElement` is in some NodeManager's dynamically allocated memory or not (i.e. the `root` buffer).

yuanming-hu (approved):
Thanks!  (TBH I don't believe I'm capable of pointing out any real issues within a limited amount of time, but as always I trust your code :-)

k-ye:
kun server is back... 🎉

PR #2020, opened 2020-11-03
[opengl] [refactor] Reduce SSBO numbers: merge earg with args

archibate:
Related issue = #

----
The limit of SSBOs can be down to 8 for OpenGL core, or 4 for OpenGL ES.
So we want to reduce the number of SSBO used to improve compatibility on low-end machines.
This PR merges the `earg` SSBO with `args` SSBO.


yuanming-hu (approved):
Cool! LGTM!

PR #2021, opened 2020-11-03
[opengl] [refactor] Move rand_state from runtime to gtmp to reduce SSBO numbers

archibate:
Related issue = #

----
This further reduce 1 SSBO when `print` is not used.
I wonder if we could even demote `RandStmt` into `GtmpStmt` and a series of bit operations in the middle end so that random  generators are backend-independent?

### DIFF
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -125,8 +125,10 @@ class KernelGen : public IRVisitor {
     emit("}}");
 
     // clang-format off
+    std::string kernel_header;
 #define __GLSL__
-    std::string kernel_header = (
+    if (used.print)  // the runtime buffer is only used for print now..
+      kernel_header += (
 #include "taichi/backends/opengl/shaders/runtime.h"
         );
     if (used.listman)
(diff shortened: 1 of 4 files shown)


taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -142,14 +144,14 @@ class KernelGen : public IRVisitor {
     if (used.int64)
       kernel_header += "layout(std430, binding = 0) buffer data_i64 { int64_t _data_i64_[]; };\n";
 
-    if (used.buf_gtmp) {
+    if (used.buf_gtmp || used.random) {
       kernel_header +=
-          "layout(std430, binding = 1) buffer gtmp_i32 { int _gtmp_i32_[]; };\n"
-          "layout(std430, binding = 1) buffer gtmp_f32 { float _gtmp_f32_[]; };\n";
+          "layout(std430, binding = 1) buffer gtmp_i32 { int _rand_state_[4]; int _gtmp_i32_[]; };\n"
```
k-ye, line 149: why `4`?

k-ye (approved):
LGTM!

taichi/backends/opengl/codegen_opengl.cpp
archibate: Make `double` and `int64_t` alignment happy.

PR #2034, opened 2020-11-08
[opt] [async] Fix constant folding in async mode

xumingkuan:
Related issue = #954 #2032

`Kernel::Kernel()` does not support multi-threads unless we lock the following part with a mutex:
```C++
  taichi::lang::context = std::make_unique<FrontendContext>();
  ir = taichi::lang::context->get_root();

  {
    CurrentKernelGuard _(program, this);
    program.start_function_definition(this);
    func();
    program.end_function_definition();
    ir->as<Block>()->kernel = this;
  }
```

This PR simply locks `get_jit_evaluator_kernel()` monolithically to prevent two threads from calling `Kernel::Kernel()` concurrently.

----

### DIFF
--- a/taichi/program/async_engine.cpp
+++ b/taichi/program/async_engine.cpp
@@ -210,10 +210,17 @@ void AsyncEngine::enqueue(const TaskLaunchRecord &t) {
 void AsyncEngine::synchronize() {
   TI_AUTO_PROF
   bool modified = true;
-  TI_TRACE("Synchronizing SFG of {} nodes", sfg->size());
-  debug_sfg("initial");
   sfg->reid_nodes();
   sfg->reid_pending_nodes();
+  TI_TRACE("Synchronizing SFG of {} nodes ({} pending)", sfg->size(),
+           sfg->num_pending_tasks());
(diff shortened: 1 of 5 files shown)


taichi/program/async_engine.cpp
```diff
@@ -210,10 +210,11 @@ void AsyncEngine::enqueue(const TaskLaunchRecord &t) {
 void AsyncEngine::synchronize() {
   TI_AUTO_PROF
   bool modified = true;
-  TI_TRACE("Synchronizing SFG of {} nodes", sfg->size());
-  debug_sfg("initial");
   sfg->reid_nodes();
   sfg->reid_pending_nodes();
+  TI_TRACE("Synchronizing SFG of {} nodes ({} pending)", sfg->size(),
+           sfg->num_pending_tasks());
+  debug_sfg("initial");
```
xumingkuan, line 217: Off-topic change: `num_pending_tasks()` matches the following `Ended up with {} nodes` better.

k-ye (approved):
Great, thanks!

PR #2035, opened 2020-11-08
[async] More fixes of constant folding in async mode

k-ye:
This is a follow-up of #2034. Besides the `Kernel` constructor, running kernel in multiple threads could cause problems on device synchronization (at least on Metal, where `metal::KernelManager` isn't thread safe).

`jit_evaluator_cache_mut` also guards the `fetch_result` call. This is because we have to make sure the content of the global Program's result buffer is not written by two threads, too.

Related issue = #

----


xumingkuan (approved):
LGTM!

archibate (approved):
LGTM! Btw, why we have two `(*ker)(launch_ctx);`? According to the DRY principle, shall we move them (including the mutex guard) into a separate function (like `jit_launch_and_fetch_result`)?

PR #2028, opened 2020-11-06
[type] Refactor bit pointers

yuanming-hu:
Related issue = #1905 

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

----

PR #2046, opened 2020-11-12
[type] Refactor bit pointers for SNodeLookupStmt in type check transform

TH3CHARLie:
Related issue = #1905, follow-up of #2028

----


Hanke98 (approved):
Great job! I think we are on the right way. Thanks~

yuanming-hu (approved):
LGTM. Thanks!

PR #2051, opened 2020-11-19
[OpenGL] [bug] Fix OpenGL for global fields with unsigned type

archibate:
Related issue = #

----


Eydcao:
By using the size of bytes it's indeed more general than using two specific types. The fix doesn't violate the existing judgement, so IMO, this fix looks good!

Eydcao (approved):
LGTM! See comment above.

TH3CHARLie (approved):
LGTM! Thanks!

Issue #2060, opened 2020-11-22
Taichi crashes when trying to initialize a u32 


medakk:
**Describe the bug**
When writing a valid unsigned 32 bit number to a `ti.u32` vector, taichi crashes

**To Reproduce**

```py
import taichi as ti

N = 1000
c = ti.Vector.field(1, dtype=ti.u32, shape=N)

@ti.kernel
def init():
    for i in range(N):
        c[i] = [0xffffffff]

init()
```

**Log/Screenshots**

```
[Taichi] mode=release
[Taichi] preparing sandbox at /tmp/taichi-mn0kbcbg
[Taichi] version 0.7.1, llvm 10.0.0, commit fc0f8be7, linux, python 3.6.9
[Taichi] materializing...
Traceback (most recent call last):
  File "test.py", line 11, in <module>
    init()
  File "/scratch/karthik/.virtualenvs/taichi-playground/lib/python3.6/site-packages/taichi/lang/kernel.py", line 574, in wrapped
    return primal(*args, **kwargs)
  File "/scratch/karthik/.virtualenvs/taichi-playground/lib/python3.6/site-packages/taichi/lang/kernel.py", line 501, in __call__
    self.materialize(key=key, args=args, arg_features=arg_features)
  File "/scratch/karthik/.virtualenvs/taichi-playground/lib/python3.6/site-packages/taichi/lang/kernel.py", line 370, in materialize
    taichi_kernel = taichi_kernel.define(taichi_ast_generator)
  File "/scratch/karthik/.virtualenvs/taichi-playground/lib/python3.6/site-packages/taichi/lang/kernel.py", line 367, in taichi_ast_generator
    compiled()
  File "test.py", line 9, in init
    c2[i] = [(0xff<<24) + (0xff<<16) + (0x00<<8) + (0xff)]
  File "/scratch/karthik/.virtualenvs/taichi-playground/lib/python3.6/site-packages/taichi/lang/common_ops.py", line 272, in assign
    return ti.assign(self, other)
  File "/scratch/karthik/.virtualenvs/taichi-playground/lib/python3.6/site-packages/taichi/lang/ops.py", line 136, in wrapped
    return a.element_wise_writeback_binary(imp_foo, b)
  File "/scratch/karthik/.virtualenvs/taichi-playground/lib/python3.6/site-packages/taichi/lang/matrix.py", line 162, in element_wise_writeback_binary
    other = Matrix(other)
[...]

PR #2083, opened 2020-12-02
[lang] [bug] Clamp int constant to its native range

k-ye:
I believe the problem is that `pybind11` has strict checks on its input args. Without the constant value knowing the type of the LHS, let's just clamp the constant to its valid range (while keeping the bits same)

Related issue = #2060 

----


xumingkuan (approved):
LGTM!

PR #2085, opened 2020-12-03
[type] Support atomic add on CustomFloatType

Hanke98:
Related issue = #1905 #2078

We try to support atomic add on `CustomFloatType` following the implementation of `CustomIntType` in this pr. 

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

----

PR #2078, opened 2020-12-01
[type] Support atomic add on CustomIntType

yuanming-hu:
Related issue = #1905

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

----

PR #2091, opened 2020-12-07
[type] [bug] Support atomic add negative numbers for custom types

Hanke98:
Related issue = #1905 #2078 #2085 

**Atomic add may crash when adding negative numbers.** It is because we didn't partially set the value to memory using mask after we do the addition. Adding negative numbers may cause the change of the higher bits. 

For example: if we pack `a` (ci3) and `b`(ci5) in a bit_struct(num=8), and set them all zero at beginning(`a` is the lower 3-bits, and `b` is the higher 5-bits). Then we add `-1` to `a`, and we will find that `b` will be changed too.

So, a possible solution might be partially setting value to memory with mask in `runtime.cpp` just like `set_partial_bits_b` which only affects the bits of `a` in the former example .

----

### DIFF
--- a/taichi/runtime/llvm/runtime.cpp
+++ b/taichi/runtime/llvm/runtime.cpp
@@ -1553,6 +1553,7 @@ void stack_push(Ptr stack, size_t max_num_elements, std::size_t element_size) {
 
 #include "internal_functions.h"
 
+// TODO: make here less repetitious.
 #define DEFINE_SET_PARTIAL_BITS(N)                                            \
   void set_partial_bits_b##N(u##N *ptr, u32 offset, u32 bits, u##N value) {   \
     u##N mask = ((((u##N)1 << bits) - 1) << offset);                          \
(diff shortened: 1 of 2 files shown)


taichi/runtime/llvm/runtime.cpp
```diff
@@ -1569,11 +1570,13 @@ void stack_push(Ptr stack, size_t max_num_elements, std::size_t element_size) {
                                                                               \
   u##N atomic_add_partial_bits_b##N(u##N *ptr, u32 offset, u32 bits,          \
                                     u##N value) {                             \
+    u##N mask = ((((u##N)1 << bits) - 1) << offset);                          \
     u##N new_value = 0;                                                       \
     u##N old_value = *ptr;                                                    \
     do {                                                                      \
       old_value = *ptr;                                                       \
       new_value = old_value + (value << offset);                              \
+      new_value = (old_value & (~mask)) | (new_value & mask);                 \
```
Hanke98, line 1579: The key change here. Use `mask` to ensure the irrelevant bits will not be affected.

TH3CHARLie (approved):
Nice catch and smart change, thanks!

yuanming-hu (approved):

PR #2096, opened 2020-12-15
[type] [bug] Correct mask used in setting bits partially

Hanke98:
Related issue = #1905 #2091 

The current way we create mask in `set_partial_bits_b` may cause the problem of `shift count >= width of type`. For example, we will do a left shifting for 32 times when we want to set a 32-bits `CustomIntType` which is exactly placed in a 32-bits `BitStruct`.

Please refer to the test case in this pr to see more details.

----

### DIFF
--- a/taichi/runtime/llvm/runtime.cpp
+++ b/taichi/runtime/llvm/runtime.cpp
@@ -1554,6 +1554,11 @@ void stack_push(Ptr stack, size_t max_num_elements, std::size_t element_size) {
 #include "internal_functions.h"
 
 // TODO: make here less repetitious.
+// Original implementation is
+// u##N mask = ((((u##N)1 << bits) - 1) << offset);
+// When N equals bits equals 32, 32 times of left shifting will be carried on
+// which is an undefined behavior.
+// see #2096 for more details
 #define DEFINE_SET_PARTIAL_BITS(N)                                            \
   void set_partial_bits_b##N(u##N *ptr, u32 offset, u32 bits, u##N value) {   \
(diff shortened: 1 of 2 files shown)


tests/python/test_bit_struct.py
```diff
@@ -77,6 +77,34 @@ def verify_val(idx: ti.i32):
...
+        x[0] = 12
+
+    @ti.kernel
+    def varify_val2():
+        assert x[0] == 12
+
+    set_val()
+    varify_val1()
+    set_val2()
+    varify_val2()
```
Hanke98, line 105: This case will fail in current implementation. Because 32 times of left shifting of a 32-bits `1` will be carried on.

taichi/runtime/llvm/runtime.cpp
```diff
@@ -1556,7 +1556,7 @@ void stack_push(Ptr stack, size_t max_num_elements, std::size_t element_size) {
 // TODO: make here less repetitious.
 #define DEFINE_SET_PARTIAL_BITS(N)                                            \
   void set_partial_bits_b##N(u##N *ptr, u32 offset, u32 bits, u##N value) {   \
-    u##N mask = ((((u##N)1 << bits) - 1) << offset);                          \
+    u##N mask = ((~(u##N)0) << (N - bits)) >> (N - offset - bits);            \
```
xumingkuan, line 1559: Maybe add some comment here like the comment in this PR or add a link to this PR here?

xumingkuan (approved):
LGTM!

PR #2098, opened 2020-12-15
[type] [bug] Add rounding for atomic adding of CustomFloatType

Hanke98:
Related issue = #1905 #2093 #2085 

The current implementation of atomic add of `CustomFloatType` miss the rounding step which may cause some problems.

Thanks to @yuanming-hu 's implementation of `CustomFloatType` rounding in `GlobalStoreStmt`, I can reuse it here and make code clean.

----

### DIFF
--- a/taichi/codegen/codegen_llvm.cpp
+++ b/taichi/codegen/codegen_llvm.cpp
@@ -1047,13 +1047,21 @@ llvm::Value *CodeGenLLVM::atomic_add_custom_float(AtomicOpStmt *stmt,
                             llvm::APFloat(1.0 / cft->get_scale())),
       llvm_type(compute_type));
   auto val_scaled = builder->CreateFMul(llvm_val[stmt->val], s);
-  auto val_scaled_int =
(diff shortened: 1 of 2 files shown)


taichi/codegen/codegen_llvm.cpp
```diff
@@ -1168,30 +1186,7 @@ void CodeGenLLVM::visit(GlobalStoreStmt *stmt) {
...
-          fmt::format("rounding_prepare_f{}", data_type_bits(compute_type)),
-          {scaled});
-
-      if (cit->get_is_signed()) {
-        store_value =
-            builder->CreateFPToSI(scaled, llvm_type(cit->get_compute_type()));
-      } else {
-        store_value =
-            builder->CreateFPToUI(scaled, llvm_type(cit->get_compute_type()));
-      }
```
Hanke98, line 1194: Moved to a separate function `CodeGenLLVM ::float_to_custom_int` to be reused.

TH3CHARLie (approved):
LGTM! Do we have any verified failing cases on this?(I saw you use "may" in the description)

PR #2099, opened 2020-12-16
[Lang] [test] Support dict as expression initializer

archibate:
Related issue = N/A

----
Also add some more information when import failed as oft.


k-ye (approved):
LGTM, thanks!

PR #2116, opened 2020-12-23
[async] Record the kernel's name in the hash only if it has arguments

xumingkuan:
Related issue = #742

This can reduce the number of uncached IR operations.

----


yuanming-hu (approved):
LGTM, thanks!

PR #2119, opened 2020-12-25
[async] Disable profiler macros in release mode

k-ye:
Related issue = #742

It has turned out that the scoped profilers have a huge overhead. E2E time when benchmarking MGPCG at 512 resolution:

1. Before this PR: 0.49s
2. With this PR: **0.27s**

(A slightly off the topic change is to tweak the color, the default blue are green colors are a bit hard to read with Ubuntu's purple background... That said, these stats won't be helpful now, and we should resort to the lower-overhead perf solutions like `perf` and vtune..)

Screenshot from 2020-12-25 21-47-31

Screenshot from 2020-12-25 21-47-17

----

### DIFF
--- a/taichi/program/async_engine.cpp
+++ b/taichi/program/async_engine.cpp
@@ -12,6 +12,9 @@
 #include "taichi/ir/transforms.h"
 #include "taichi/program/extension.h"
 
+// Keep this include in the end!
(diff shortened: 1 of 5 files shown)


xumingkuan (approved):
LGTM!

yuanming-hu (approved):
LGTM, thanks!

PR #2123, opened 2020-12-27
[async] Add explicit flush API and async_flush_every

k-ye:
Related issue = #742

By flushing more frequently, we can have wider CPU-GPU execution overlap. MGPCG (res=512) async E2E execution has been reduced from ~0.25s to ~0.15s when flushing every 50 tasks. 

----

### DIFF
--- a/taichi/program/async_engine.cpp
+++ b/taichi/program/async_engine.cpp
@@ -275,6 +275,7 @@ void AsyncEngine::flush() {
   for (auto &task : tasks) {
     queue.enqueue(task);
   }
+  flush_counter_++;
 }
 
 void AsyncEngine::debug_sfg(const std::string &stage) {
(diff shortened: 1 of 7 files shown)


taichi/program/async_engine.cpp
```diff
@@ -205,11 +205,29 @@ void AsyncEngine::launch(Kernel *kernel, Context &context) {
...
 
 void AsyncEngine::synchronize() {
   TI_AUTO_PROF;
+  flush();
+  queue.synchronize();
+
+  sync_counter_++;
+  // Clear SFG debug stats
+  cur_sync_sfg_debug_counter_ = 0;
+  cur_sync_sfg_debug_per_stage_counts_.clear();
```
xumingkuan, line 225: nit: When debugging SFG, I think the counter of #flushes is more useful than #syncs. Maybe include both #flushes and #syncs in the generated .dot file name?

xumingkuan (approved):
LGTM!

taichi/program/async_engine.cpp
k-ye: Yep, done

PR #2132, opened 2020-12-30
[GUI] Fix ti.GUI.MOVE not working on win32 and cocoa

archibate:
Related issue = A windows Taichi THREE user reports

----
To reproduce:
```
import taichi as ti

gui = ti.GUI()
while gui.running:
    for e in gui.get_events():
        print(e.type, e.key)
    gui.show()
```
You should see no `Motion` event when your mouse is moving above the window. (Didn't have a win env to reproduce yet, but I let that win user to help me confirmed that this is reproduced)
But this works fine on X11:
```
EType.Move Motion
EType.Move Motion
EType.Move Motion
EType.Move Motion
EType.Move Motion
...
```
Please merge this fix soon so that windows and mac users could enjoy orbiting camera around in Taichi THREE :smiley: 

@k-ye Would you mind help me confirm the cocoa part works (by using the reproduce code)? Thank in advance!

### DIFF
--- a/taichi/gui/cocoa.cpp
+++ b/taichi/gui/cocoa.cpp
@@ -360,6 +360,8 @@ void GUI::process_event() {
         case 7:   // NSRightMouseDragged
         case 27:  // NSNSOtherMouseDragged
           set_mouse_pos(p.x, p.y);
+          key_events.push_back(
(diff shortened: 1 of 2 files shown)


JYLeeLYJ (commented):
Hi , this code already works well on Win after my confirmation.

k-ye (approved):
Thanks LGTM!

PR #2136, opened 2021-01-01
[type] Local adder structure

TH3CHARLie:
Related issue = #1905

----


TH3CHARLie:
some performance numbers comparing two kinds of loops:
```
[ 41.96%   0.105 s      1x |  105.085   105.085   105.085 ms] evolve_naive_c8_0_kernel_10_range_for
[  0.70%   0.002 s      1x |    1.756     1.756     1.756 ms] evolve_vectorized_c6_0_kernel_7_struct_for
[  0.51%   0.001 s      1x |    1.272     1.272     1.272 ms] evolve_vectorized_c6_0_kernel_6_listgen_S2dense
[  0.37%   0.001 s      1x |    0.926     0.926     0.926 ms] evolve_vectorized_c6_0_kernel_4_listgen_S1pointer
[  0.00%   0.000 s      1x |    0.001     0.001     0.001 ms] evolve_vectorized_c6_0_kernel_5_serial
[  0.00%   0.000 s      1x |    0.001     0.001     0.001 ms] evolve_vectorized_c6_0_kernel_3_serial
```

taichi/ir/statements.h
```diff
@@ -141,9 +141,16 @@ class BinaryOpStmt : public Stmt {
...
-  BinaryOpStmt(BinaryOpType op_type, Stmt *lhs, Stmt *rhs)
-      : op_type(op_type), lhs(lhs), rhs(rhs) {
+  BinaryOpStmt(BinaryOpType op_type,
+               Stmt *lhs,
+               Stmt *rhs,
+               bool is_bit_vectorized = false)
+      : op_type(op_type),
+        lhs(lhs),
+        rhs(rhs),
+        is_bit_vectorized(is_bit_vectorized) {
```
yuanming-hu, line 153: This is a little too much intrusion into the existing system. It seems to me that `is_bit_vectorized` is only used in the `bit_loop_vectorize` pass - maybe you can use an `std::unordered_map<Stmt *, bool>` member variable in `class BitLoopVectorize`, instead of adding a new field in `class BinaryOpStmt`? (Just like `llvm_val` in LLVM codegens.)

tests/python/test_bit_array_vectorization.py
```diff
@@ -93,3 +93,85 @@ def verify(dx: ti.template(), dy: ti.template()):
...
+    def init():
+        for i, j in ti.ndrange((boundary_offset, N - boundary_offset),
+                               (boundary_offset, N - boundary_offset)):
+            x[i, j] = ti.random(dtype=ti.i32) % 2
+
+    @ti.kernel
+    def evolve_vectorized(x: ti.template(), y: ti.template()):
+        ti.bit_vectorize(32)
+        for i, j in x:
+            num_active_neighbors = 0
```
yuanming-hu, line 138:
```suggestion
```

tests/python/test_bit_array_vectorization.py
```diff
@@ -93,3 +93,85 @@ def verify(dx: ti.template(), dy: ti.template()):
...
+    verify(1, 1)
+    assign_vectorized(1, -1)
+    verify(1, -1)
+    assign_vectorized(-1, -1)
+    verify(-1, -1)
+    assign_vectorized(-1, 1)
+    verify(-1, 1)
+
+
+@ti.test(require=ti.extension.quant)
```
yuanming-hu, line 106:
```suggestion
@ti.test(require=ti.extension.quant, debug=True)
```
Debug mode needed for `assert` to work.

taichi/transforms/bit_loop_vectorize.cpp
```diff
@@ -155,17 +156,180 @@ class BitLoopVectorize : public IRVisitor {
...
+            auto and_a_b = std::make_unique<BinaryOpStmt>(
+                BinaryOpType::bit_and, not_a.get(), load_b.get());
+            auto and_b_c = std::make_unique<BinaryOpStmt>(
+                BinaryOpType::bit_and, and_a_b.get(),
+                rhs_val == 2 ? (Stmt *)(not_c.get()) : (Stmt *)(load_c.get()));
+            // mark the last stmt as vectorized
+            and_b_c->is_bit_vectorized = true;
+            // modify IR
+            auto and_a_b_p = and_a_b.get();
+            stmt->insert_before_me(std::move(load_a));
```
yuanming-hu, line 229: A more elegant way to do this: https://github.com/taichi-dev/taichi/blob/b13abfaf8cb7c8b7a03d03ec4422795bf50dcda6/taichi/transforms/demote_operations.cpp#L22

Use `VecStatement::push_back<...>` to create the statements and `stmt->insert_before_me(vec_stmt)` to insert. (You don't need the `DelayedIRModifier` part.)

taichi/transforms/bit_loop_vectorize.cpp
```diff
@@ -155,17 +156,180 @@ class BitLoopVectorize : public IRVisitor {
...
+    auto sum_a =
+        std::make_unique<AtomicOpStmt>(AtomicOpType::bit_xor, a, carry_b.get());
+    // modify IR
+    stmt->insert_before_me(std::move(load_c));
+    stmt->insert_before_me(std::move(carry_c));
+    stmt->insert_before_me(std::move(sum_c));
+    stmt->insert_before_me(std::move(load_b));
+    stmt->insert_before_me(std::move(carry_b));
+    stmt->insert_before_me(std::move(sum_b));
+    stmt->insert_before_me(std::move(sum_a));
```
yuanming-hu, line 305: Ditto.

yuanming-hu (approved):
Awesome!! Just a few nits. Thanks!

taichi/ir/statements.h
TH3CHARLie: `is_bit_vectorized` is used only in the `bit_loop_ vectorize` pass when tagged on `BinaryOpStmt` and this part should be replaced with some pass-scope data structure, just as you suggested. But for `GlobalPtrStmt` and `GetChStmt`, they need the tag to pass later passes including `lower_access` and `type_check`

tests/python/test_bit_array_vectorization.py
```diff
@@ -93,3 +93,85 @@ def verify(dx: ti.template(), dy: ti.template()):
...
+            num_active_neighbors += ti.cast(x[i - 1, j - 1], ti.u32)
+            num_active_neighbors += ti.cast(x[i - 1, j], ti.u32)
+            num_active_neighbors += ti.cast(x[i - 1, j + 1], ti.u32)
+            num_active_neighbors += ti.cast(x[i, j - 1], ti.u32)
+            num_active_neighbors += ti.cast(x[i, j + 1], ti.u32)
+            num_active_neighbors += ti.cast(x[i + 1, j - 1], ti.u32)
+            num_active_neighbors += ti.cast(x[i + 1, j], ti.u32)
+            num_active_neighbors += ti.cast(x[i + 1, j + 1], ti.u32)
+            y[i, j] = (num_active_neighbors == 3) or (num_active_neighbors == 2
+                                                      and x[i, j] == 1)
```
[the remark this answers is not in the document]
TH3CHARLie: ummm, I find it extracting it into a `ti.func` and calling it produces incorrect results while inline it does not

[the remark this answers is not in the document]
yuanming-hu: Oh OK. Maybe passing arguments leads to some issues. Nevermind.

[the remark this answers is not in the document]
TH3CHARLie: yes, passing arguments would cause the index offset computation different from the inlined version. I'd say we stick to this version for now.

taichi/ir/statements.h
yuanming-hu: Right, that's what I meant: we can use a pass-scope data structure just for `BinaryOpStmt::is_bit_vectorized`. Given we are rushing for the deadline it's fine that we don't do it now.

taichi/transforms/bit_loop_vectorize.cpp
TH3CHARLie: I find to do it the elegant way we may still need `DelayedIRModifier` and there are other places that require changes as well(e.g. see the visitor for `GlobalLoadStmt`), therefore I think we should do it in a separate PR later and refactoring all this together. This also applies to the changes for `BinaryOpStmt::is_bit_vectorized`.

PR #2149, opened 2021-01-11
[misc] Add a Github Action workflow to trigger on publishing a release

k-ye:
This PR sets up a GitHub actions workflow upon publishing a new release. Right now it just forks  the "Build and Test (CPU)" job. Later on we can add these:

* [ ] PyPI upload
* [ ] [`macos-11.0`](https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources) (I briefly tested this platform. However, it seems to me that this particular resource is limited. The job queued for ~0.5h...)

Related issue = #1580

----

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,23 +6,19 @@ on:
 
 jobs:
   build_and_test_cpu:
-    name: Build and Test (CPU)
-    if: ${{ !contains(github.event.pull_request.labels.*.name, 'skip ci') && github.event.sender.login != 'taichi-gardener' }}
+    name: Build and Test (macOS only)
     strategy:
       matrix:
         include:
-          - os: ubuntu-latest
+          - os: macos-latest
             python: 3.6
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
+name: Publishing Release
+on:
+  release:
+    # https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release
+    types: [published]
```
k-ye, line 5: The real change

yuanming-hu (approved):
Awesome! LGTM. Hopefully, we can save 100$+ per month by adding this...

k-ye:
I made this  macOS only for now, so that we don't conflict with Jenkins on LInux. Also, before we can use `python build.py try_upload`, we need to setup the pypi password in the secrets first...

PR #2150, opened 2021-01-11
[async] [bug] Fix missing memory access options in async mode

xumingkuan:
Related issue = #742

Fixes `OffloadedStmt::clone()`, and merges `OffloadedStmt::mem_access_opt`s when fusing.

Off-topic: I'm not sure if two struct-fors with different index offsets can be fused together.

----

### DIFF
--- a/taichi/program/async_utils.cpp
+++ b/taichi/program/async_utils.cpp
@@ -372,6 +372,7 @@ TaskFusionMeta get_task_fusion_meta(IRBank *bank, const TaskLaunchRecord &t) {
   if (task->task_type == OffloadedTaskType::struct_for) {
     meta.snode = task->snode;
     meta.block_dim = task->block_dim;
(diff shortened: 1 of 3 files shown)


archibate:
IIUC coordinate offsets can be completely implemented in Python side to reduce maintainance cost?

taichi/program/async_utils.cpp
```diff
@@ -372,6 +372,7 @@ TaskFusionMeta get_task_fusion_meta(IRBank *bank, const TaskLaunchRecord &t) {
   if (task->task_type == OffloadedTaskType::struct_for) {
     meta.snode = task->snode;
     meta.block_dim = task->block_dim;
+    // TODO: What about index_offsets?
```
xumingkuan, line 375:
```suggestion
    // We don't need to record index_offsets because it's not used anymore.
```

PR #2152, opened 2021-01-12
[opt] Simplify bit_cast of bit_cast

xumingkuan:
Related issue = #656 

Also adds `DelayedIRModifier::mark_as_modified()` because this pattern (modifying IR, not through DelayedIRModifier, but need to record if the IR is modified) repeats in some other passes.

----


yuanming-hu (approved):
Great! LGTM.

Issue #2146, opened 2021-01-09
Seg faulting on index that shouldn't seg fault


JoshWolper:
**Describe the bug**
Getting a seg fault on my data structure when using an in-range index. This is an attempt to implement a 2D/3D spatial hash following a very similar approach to the one here: https://github.com/taichi-dev/taichi/blob/master/examples/pbf2d.py
Also worth noting that this bug does not occur for 2D, only 3D!

**To Reproduce**

```py
import taichi as ti
import numpy as np
import triangle as tr

#ti.init(default_fp=ti.f64, arch=ti.gpu) # Try to run on GPU    #GPU, parallel
ti.init(default_fp=ti.f64, arch=ti.cpu, cpu_max_num_threads=1)  #CPU, sequential

#Sparse Grids
#---Params
dim = 3
dx = 0.003
rp = (3*(dx**2))**0.5
invDx = 1.0 / 0.003
nGrid = ti.ceil(invDx)
grid_size = 4096
grid_block_size = 128
leaf_block_size = 16 if dim == 2 else 8
indices = ti.ij if dim == 2 else ti.ijk
offset = tuple(0 for _ in range(dim))
#---Grid Shapes for PID
grid = ti.root.pointer(indices, grid_size // grid_block_size) # 32
block = grid.pointer(indices, grid_block_size // leaf_block_size) # 8
pid = ti.field(int)
block.dynamic(ti.indices(dim), 1024 * 1024, chunk_size=leaf_block_size**dim * 8).place(pid, offset=offset + (0, ))
#---Grid Shapes for Rest of Grid Structures
grid2 = ti.root.pointer(indices, grid_size // grid_block_size) # 32
block2 = grid2.pointer(indices, grid_block_size // leaf_block_size) # 8
def block_component(c):
    block2.dense(indices, leaf_block_size).place(c, offset=offset) # 16 in 3D, 8 in 2D (-2048, 2048) or (0, 4096) w/o offset

#Grid Structures
gridNumParticles = ti.field(dtype=int)      #track number of particles in each cell using cell index
maxPPC = 2**10
block_component(gridNumParticles) #keep track of how many particles are at each cell of backGrid
backGrid = ti.field(int)              #background grid to map grid cells to a list of particles they contain
backGridIndeces = ti.ijk if dim == 2 else ti.ijkl
backGridShape = (nGrid, nGrid, maxPPC) if dim == 2 else (nGrid, nGrid, nGrid, maxPPC)
[...]

PR #2154, opened 2021-01-13
[misc] Change SNode max_num_elements to use int64

k-ye:
Related issue = #2146

Hi,

I'm changing the return type of `SNode::max_num_elements()` to `int64` to avoid overflow in #2146. @TH3CHARLie please let me know if this also fixes your problem, thx!

Note that this is NOT a thorough fix for #2146, as I still prefer that we dump a huge warning when the requested size is impractically large. (WIth this patch, the reproducer still segfaulted without dumping more information.) I will take a look tmr...

----


yuanming-hu (approved):
Thank you so much! Looks great.

PR #2157, opened 2021-01-14
[Misc] Add experimental Python 3.9 support

archibate:
Related issue = #

----
The only issue seems to be the deprecated `ast.Index`, fixed in this Pr - now Python 3.9 can pass all the tests!
We may discuss if we should officially release Python 3.9 wheel on PyPI, FYI Arch Linux has officially make Python 3.9 the default version after a full system upgrade, it seems the Blender bundled Python is also 3.9 now.. I know you are busy with a interesting research projects of SIGGRAPH (let me know more details after that!), so please let me take some responsbilities if you'd like to :smile:

### DIFF
--- a/.github/workflows/persubmit.yml
+++ b/.github/workflows/persubmit.yml
@@ -16,8 +16,8 @@ jobs:
           - os: macos-latest
             python: 3.7
             with_cc: OFF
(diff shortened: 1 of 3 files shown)


k-ye (approved):
LGTM, thanks!

PR #2158, opened 2021-01-15
[misc] Deleted 3 debug messages in codegen_cc.cpp

Leonz5288:
Related issue = #

----
There were 3 emit operations in this file that emit print operations to the compiled c file when dumping some gradient kernels. These debug messages won't be helpful to developers that works on gradient kernels.
@yuanming-hu


Leonz5288:
No performance of code would be affected.
Code format will not be changed since just deleted 3 lines of code.

archibate (approved):
Nice catch! Sorry it's me wrote these insane code :rofl: thank for fixing it!

PR #2165, opened 2021-01-19
[async] Add config flag async_max_fuse_per_task

k-ye:
Related issue = #742

This allows us to control how many fusions can happen per task.

----

### DIFF
--- a/taichi/program/state_flow_graph.cpp
+++ b/taichi/program/state_flow_graph.cpp
@@ -727,6 +727,10 @@ std::unordered_set<int> StateFlowGraph::fuse_range(int begin, int end) {
   for (int i = 0; i < n; i++) {
     fused[i] = nodes[i]->rec.empty();
   }
+  auto get_initial_remaining_fuses = [n, this]() {
+    const auto configured_max = program_->config.async_max_fuse_per_task;
+    return configured_max > 0 ? configured_max : n;
+  };
   // The case without an edge: O(sum(size * min(size, n / 64))) = O(n^2 / 64)
   const int kLargeFusionSetThreshold = std::max(n / 16, 16);
   for (auto &fusion_map : task_fusion_map) {
@@ -744,13 +748,19 @@ std::unordered_set<int> StateFlowGraph::fuse_range(int begin, int end) {
           // Fuse no more than one task into task a,
           // otherwise do_fuse may be very slow
           Bitset current_mask = (mask & ~(has_path[a] | has_path_reverse[a]));
-          int b = current_mask.lower_bound(a + 1);
-          if (b == -1) {
-            mask[a] = false;  // a can't be fused in this iteration
-          } else {
-            do_fuse(a, b);
-            mask[a] = false;
-            mask[b] = false;
+          int b = a + 1;
+          int remaining_fuses = get_initial_remaining_fuses();
+          while ((b != -1) && (remaining_fuses > 0)) {
+            int b = current_mask.lower_bound(b);
+            if (b == -1) {
+              mask[a] = false;  // a can't be fused in this iteration
+            } else {
+              do_fuse(a, b);
+              mask[a] = false;
+              mask[b] = false;
+              b++;  // for computing the next lower bound;
+              remaining_fuses--;
+            }
           }
         }
       }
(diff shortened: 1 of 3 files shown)


taichi/program/state_flow_graph.cpp
```diff
@@ -790,14 +801,18 @@ std::unordered_set<int> StateFlowGraph::fuse_range(int begin, int end) {
...
       // Fuse no more than one task into task i
       bool i_updated = false;
+      int remaining_fuses = get_initial_remaining_fuses();
       for (auto &edge : nodes[i]->output_edges.get_all_edges()) {
         if (edge.second->pending()) {
           const int j = edge.second->pending_node_id - begin;
           if (j >= 0 && j < nodes.size() && edge_fusible(i, j)) {
             do_fuse(i, j);
             // Iterators of nodes[i]->output_edges may be invalidated
             i_updated = true;
```
k-ye, line 811: IIUC, `i_updated` isn't useful, and we can remove L819-821?

taichi/program/state_flow_graph.cpp
```diff
@@ -790,14 +801,18 @@ std::unordered_set<int> StateFlowGraph::fuse_range(int begin, int end) {
...
-            break;
+            remaining_fuses--;
+            if (remaining_fuses == 0) {
```
k-ye, line 813: Intentionally not writing this as `if ((--remaining_fuses) == 0)` to reduce confusion..

taichi/program/state_flow_graph.cpp
xumingkuan: Oh yes. I thought there were nested for-loops when I wrote this but there weren't.

yuanming-hu (approved):
LGTM in general. I'm not the one with the most expertise in this part, so let's see what @xumingkuan says.

taichi/program/state_flow_graph.cpp
```diff
@@ -790,14 +801,18 @@ std::unordered_set<int> StateFlowGraph::fuse_range(int begin, int end) {
     if (!fused[i]) {
       // Fuse no more than one task into task i
       bool i_updated = false;
+      int remaining_fuses = get_initial_remaining_fuses();
       for (auto &edge : nodes[i]->output_edges.get_all_edges()) {
```
xumingkuan, line 805: Note that "Iterators of nodes[i]->output_edges may be invalidated" upon `do_fuse`, so we may need to break this loop or rewrite the loop and handle the iterators carefully.

taichi/program/state_flow_graph.cpp
```diff
@@ -727,6 +727,10 @@ std::unordered_set<int> StateFlowGraph::fuse_range(int begin, int end) {
   for (int i = 0; i < n; i++) {
     fused[i] = nodes[i]->rec.empty();
   }
+  auto get_initial_remaining_fuses = [n, this]() {
+    const auto configured_max = program_->config.async_max_fuse_per_task;
+    return configured_max > 0 ? configured_max : n;
+  };
```
xumingkuan, line 733: I suggest appending `_per_task` to this variable name to reduce confusion.

BTW what's the benefit to write it as a lambda expression instead of a const int (since `n` never changes in this function)?

xumingkuan (approved):
Looks great, but we may need to handle the iterators carefully.

taichi/program/state_flow_graph.cpp
xumingkuan: Because of `replace_reference` in `do_fuse`, it's hard to handle the iterators here. Shall we loop over all edges for `remaining_fuses` times?

k-ye: Ah you are right, and I believe the latest decision is not to change this part?

taichi/program/state_flow_graph.cpp
k-ye: >  what's the benefit to write it as a lambda expression

I started writing this with the intention of passing in the upper bound, then noticed that we can just use `n`, so no real benefit.. Fixed

taichi/program/state_flow_graph.cpp
xumingkuan: Yes, we decide to not change this part.

taichi/program/state_flow_graph.cpp
```diff
@@ -744,13 +748,19 @@ std::unordered_set<int> StateFlowGraph::fuse_range(int begin, int end) {
...
-          if (b == -1) {
-            mask[a] = false;  // a can't be fused in this iteration
-          } else {
-            do_fuse(a, b);
-            mask[a] = false;
-            mask[b] = false;
+          int b = a + 1;
+          int remaining_fuses = get_initial_remaining_fuses();
+          while ((b != -1) && (remaining_fuses > 0)) {
+            int b = current_mask.lower_bound(b);
```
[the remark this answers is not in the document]
k-ye: good catch..

PR #2172, opened 2021-01-24
[doc] Better CUDA out of memory messages

archibate:
Related issue = https://github.com/taichi-dev/taichi_blend/issues/8

----
I find this note in `docs/hello.rst`:

>  When used with the CUDA backend on Windows or ARM devices (e.g. NVIDIA Jetson),
> Taichi by default allocates 1 GB GPU memory for field storage.

But I find this issue occur on Linux too? Maybe the true issue is not about Windows or ARM, but all CUDA devices that doesn't support `Unified Memory`?


archibate:
Seems our OpenGL CI is broken on `zhen`?
```
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  151 (GLX)
  Minor opcode of failed request:  3 (X_GLXCreateContext)
  Value in failed request:  0x0
  Serial number of failed request:  30
  Current serial number in output stream:  31
```

yuanming-hu (approved):
Thanks for improving this and sorry about my delayed review. I have been AFK in the past 6 months due to many personal issues (traveling internationally during COVID, finishing my Ph.D. defense, etc.) I think this PR is helpful to Taichi users.

You are very right that this is about **unified memory**. Note that Taichi's CUDA backend requires unified memory to *signal* the host (during kernel execution) to dynamically allocate memory on GPU, so on platforms without unified memory support (e.g., Windows) we have to fall back to a pre-allocated fixed-size memory pool.

The memory pool is used for **dynamic** memory allocation only. Your original example, 

```
ti.init(ti.cuda)     
x = ti.field(ti.u8, 1024*1024*1024)  # 1 GB
```

does not really trigger the out-of-memory issue, since the compiler knows how large the `x` field is and allocates the buffer ahead of time. There is actually no dynamic allocation in this process. I included another example that can trigger what you reported. I also took a pass on the text.

I think this PR LGTM now and is ready to merge :-) Thanks again!

(Btw, we have also replaced `zhen` with a more stable machine!)

PR #2174, opened 2021-01-24
[type] Atomic demotion for bit struct stores

xumingkuan:
Related issue = #1905

Test case(0.75s -> 0.42s)
```python
import taichi as ti

ti.init(arch=ti.cpu, kernel_profiler=True, print_ir=True, quant_opt_atomic_demotion=True)

quant = True

n = 1024 * 1024 * 256

if quant:
    ci16 = ti.quant.int(16, True)

    x = ti.field(dtype=ci16)
    y = ti.field(dtype=ci16)

    ti.root.dense(ti.i, n).bit_struct(num_bits=32).place(x, y)
else:
    x = ti.field(dtype=ti.i16)
    y = ti.field(dtype=ti.i16)

    ti.root.dense(ti.i, n).place(x, y)


@ti.kernel
def foo():
    for i in range(n):
        x[i] = 1


for i in range(10):
    foo()

ti.kernel_profiler_print()
```

----

### DIFF
--- a/taichi/analysis/gather_uniquely_accessed_pointers.cpp
+++ b/taichi/analysis/gather_uniquely_accessed_pointers.cpp
@@ -163,11 +163,70 @@ class UniquelyAccessedSNodeSearcher : public BasicStmtVisitor {
   }
 };
 
+class UniquelyAccessedBitStructGatherer : public BasicStmtVisitor {
+ private:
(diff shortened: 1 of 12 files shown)


yuanming-hu (approved):
Awesome!!! LGTM.

PR #2175, opened 2021-01-25
[type] Improve bit struct store fusion and atomic demotion

xumingkuan:
Related issue = #1905 #2171 #2174

Test case (there are no atomic operations when either one of the optimizations is enabled):
fusion=False, atomic_demotion=False: 1.171s
fusion=False, atomic_demotion=True: 0.355s
fusion=True, atomic_demotion=False: 0.356s
fusion=True, atomic_demotion=True: 0.358s
```python
import taichi as ti

ti.init(arch=ti.cpu, kernel_profiler=True, print_ir=False,
        quant_opt_store_fusion=True, quant_opt_atomic_demotion=True)

quant = True

n = 1024 * 1024 * 256

if quant:
    ci16 = ti.quant.int(16, True)

    x = ti.field(dtype=ci16)
    y = ti.field(dtype=ci16)

    ti.root.dense(ti.i, n).bit_struct(num_bits=32).place(x, y)
else:
    x = ti.field(dtype=ti.i16)
    y = ti.field(dtype=ti.i16)

    ti.root.dense(ti.i, n).place(x, y)


@ti.kernel
def foo():
    for i in range(n):
        x[i] = i & 1023
        y[i] = i & 15


for i in range(10):
    foo()

ti.kernel_profiler_print()
```

----

### DIFF
--- a/taichi/analysis/gather_uniquely_accessed_pointers.cpp
+++ b/taichi/analysis/gather_uniquely_accessed_pointers.cpp
@@ -185,7 +185,7 @@ class UniquelyAccessedBitStructGatherer : public BasicStmtVisitor {
           irpass::analysis::gather_uniquely_accessed_pointers(stmt);
       for (auto &it : loop_unique_ptr) {
         auto *snode = it.first;
-        auto *ptr = it.second;
+        auto *ptr1 = it.second;
         if (snode->is_bit_level) {
           // Find the nearest non-bit-level ancestor
           while (snode->is_bit_level) {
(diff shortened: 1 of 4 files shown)


taichi/analysis/gather_uniquely_accessed_pointers.cpp
```diff
@@ -194,11 +194,15 @@ class UniquelyAccessedBitStructGatherer : public BasicStmtVisitor {
...
-                                                           ptr)) {
-              accessed_ptr->second = nullptr;  // not uniquely accessed
+            auto *ptr2 = accessed_ptr->second;
+            TI_ASSERT(ptr1->indices.size() == ptr2->indices.size());
+            for (int id = 0; id < (int)ptr1->indices.size(); id++) {
+              if (!irpass::analysis::same_value(ptr1->indices[id],
+                                                ptr2->indices[id])) {
+                accessed_ptr->second = nullptr;  // not uniquely accessed
+              }
             }
```
xumingkuan, line 206: Since they are different components of a bit struct, the SNodes are definitely different, and what we want to check are the indices.

yuanming-hu (approved):
Looks great!!

Issue #2185, opened 2021-02-11
Unblock the mac release


k-ye:
We have recently switched to use Github Actions (GHA) for Mac release. The first run wasn't successful, due to the missing env vars (https://github.com/taichi-dev/taichi/runs/1839080159?check_suite_focus=true). I will do some local experiments with `build.py` to see if there are other problems. One workaround will be to let GHA retain the wheel artifacts, then upload them manually.

k-ye:
Fixed by #2188

PR #2186, opened 2021-02-11
[misc] Refactor release workflow to upload the wheel artifacts

k-ye:
Related issue = #2185 

I did some cleanup to `python/build.py`. The change is actually not as large as it appears, because I was mostly grouping lots of the global statements into functions. A new mode, `build`, is added to this script, which only builds the wheels into `python/dist`, but doesn't do the upload. Also added an option `--skip_build`.

The release job now becomes:

1. `python build.py build`: builds the wheels without uploading.
2. `Archive Wheel Artifacts`: make the wheel files the artifacts of this job. (Unfortunately, I haven't found a way to include the release tag in the artifacts.) In case the final PyPI upload step failed, we still have a way to access these wheels and upload them manually.
3. `python build.py try_upload --skip_build`: Try uploading the wheels built from step 1.

Tested in https://github.com/k-ye/test_gh_action/pull/10/checks?check_run_id=1879776553

----

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,7 +52,24 @@ jobs:
           ti diagnose
           ti test -vr2 -t2
 
-      # TODO(#1580): Upload to PyPI. Need to set up PYPI_PWD in the secrets first.
+      - name: Create Python Wheel
+        run: |
+          export TAICHI_REPO_DIR=`pwd`
+          export PATH=$TAICHI_REPO_DIR/bin:$PATH
+          export PATH=$TAICHI_REPO_DIR/taichi-llvm/bin/:$PATH
+          export PYTHONPATH=$TAICHI_REPO_DIR/python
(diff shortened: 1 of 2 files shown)


yuanming-hu (approved):
Awesome!

> In case the final PyPI upload step failed, we still have a way to access these wheels and upload them manually.

Not sure if this is related: sometimes the wheels fail just because PyPI runs out of space. Since we need to ship binaries the wheels take much larger space than a normal Python project. I manually deleted a few unimportant releases to save space, but we should try to systematically fix this when we have time (e.g., email PyPI and ask for more space).

k-ye:
> I manually deleted a few unimportant releases to save space, but we should try to systematically fix this when we have time (e.g., email PyPI and ask for more space).

Ah, I was mostly afraid if there is still some problem in the Github Action workflow :)

PR #2081, opened 2020-12-02
[Lang] Fix ndrange symbol leakage in AST transformer

archibate:
Related issue = WeChat group by @Eydcao 

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

----

PR #2187, opened 2021-02-12
[Bug] [lang] Fix AST not being transformed inside ti.ndrange

archibate:
Related issue = #

----
To reproduce:
```py
import taichi as ti

a = ti.field(int, ())
b = ti.field(int, ())

@ti.kernel
def func_ok():
    for i in range(a[None]):
        print(i)

@ti.kernel
def func_err():
    for i, j in ti.ndrange(a[None], b[None]):
        print(i, j)

a[None] = 3
b[None] = 3
func_err()
```
Running it will raises `AssertionError: __getitem__ cannot be called in Taichi-scope`.

This is because `a[None]` not being transformed into `ti.subscript(a, None)`.
This is an unexpected error introduced since #2081, which generates:
```py
__ndrange = ti.static(ti.ndrange(a[None], b[None]))
```
Here I was mean to assign statically to prevent symbol leakage, but `ti.static` leads to another side effect that 'do not transform the AST inside'. And `a[None]` no longer being transformed into `ti.subscript(a, None)`.
Sorry for not considering enough test cases! So in this PR I removed `ti.static` and use
```py
__ndrange{id(node)} = ti.ndrange(a[None], b[None])
```
instead, where `id(node)` is the unique id for each node, so that `a[None]` will be further transformed into `ti.subscript(a, None)` to prevent the above issue, while still allow nested ndrange as #2081 does.

---
BTW how do you think if we deprecate `ti.subscript`? We can always distinguish when `Expr.subscript` or `Expr.__getitem__` should be called by detecting `ti.inside_kernel()`, so why not just merge them into `Expr.__getitem__`?

### DIFF
--- a/python/taichi/lang/transformer.py
+++ b/python/taichi/lang/transformer.py
@@ -428,7 +428,8 @@ if ti.static(1):
         t = ast.parse(template).body[0]
         t.body[0].value = node.iter
         t_loop = t.body[1]
-        t_loop.iter.args[0] = self.parse_expr(f'__ndrange{id(node)}.acc_dimensions[0]')
+        t_loop.iter.args[0] = self.parse_expr(
+            f'__ndrange{id(node)}.acc_dimensions[0]')
         targets = self.get_targets(node)
         targets_tmp = ['__' + name for name in targets]
         loop_body = t_loop.body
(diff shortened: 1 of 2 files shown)


xumingkuan (approved):
LGTM, thanks!

PR #2191, opened 2021-02-25
[test] [opengl] Avoid floor division cornor cases by adjusting test data

BillXu2000:
Three python tests related to floor division in test_element_wise.py fail on my laptop under OpenGL (Ubuntu 20.04, AMD Ryzen 7 3700U with Radeon Vega Mobile Gfx).
In these tests, 3 / 3 gives 0.99999997 and 3 // 3 gives 0.
The difference between 0.99999997 and 1 is acceptable when doing float divisions, but the floor operation makes the answer totally wrong and makes these three tests fail on my laptop. (3 // 3 should be 1 instead of 0)


yuanming-hu (approved):
Awesome, LGTM!

PR #2199, opened 2021-03-03
[ir] [transforms] Added assertion that indices won't cause overflow under debug mode

Leonz5288:
Related issue = #2146

With large indices such as variable[300][300][300][300], when performing IR transform and simplifying this to an integer, this integer will overflow to negative number, which would cause SegFault. I added assertion to IR under debug mode which would assert that the simplified index must NOT be less than 0. One test is provided named "test_indices_assert.py".

For modification in docs, there is one typo that added an 's' after 'exclude'. I simply deleted this 's'.
@k-ye 

----


taichi/transforms/simplify.cpp
```diff
@@ -275,8 +275,24 @@ class BasicBlockSimplify : public IRVisitor {
...
     }
-    stmt->replace_with(sum.get());
-    stmt->insert_before_me(std::move(sum));
+    // Compare the result with 0 to make sure no overflow occurs under Debug Mode.
+    bool debug = stmt->get_kernel()->program.config.debug;
+    if (debug) {
+      auto zero = Stmt::make<ConstStmt>(LaneAttribute<TypedConstant>(0));
+      auto check_sum = Stmt::make<BinaryOpStmt>(BinaryOpType::cmp_ge, sum.get(), zero.get());
+      auto assert = Stmt::make<AssertStmt>(check_sum.get(), "The indices provided are too big!", std::vector<Stmt *>());
+      auto select = Stmt::make<TernaryOpStmt>(TernaryOpType::select, check_sum.get(), sum.get(), zero.get());
```
k-ye, line 284: Let's add this comment to explain why we need `select`:

Because Taichi's assertion is checked only after the execution of the whole kernel, when the linear index overflows and goes negative, we have to replace that with 0 to make sure that the rest of the kernel can still complete. Otherwise, Taichi would crash due to the illegal memory address.

tests/python/test_indices_assert.py
```diff
@@ -0,0 +1,15 @@
+import taichi as ti
+import pytest
+
+@ti.test(debug=True, gdb_trigger=False, exclude=[ti.metal])
```
k-ye, line 4: Let's just make sure it runs on CPU for now.

```suggestion
@ti.test(debug=True, gdb_trigger=False, arch=[ti.cpu])
```

tests/python/test_indices_assert.py
```diff
@@ -0,0 +1,15 @@
...
+def test_indices_assert():
+    
+    overflow = ti.field(int, (334, 334, 334, 2*10))
```
k-ye, line 7: I saw the CI failed because of `std::bad_alloc`. Maybe create a smaller `field`, since the assertion you added doesn't care about the actual size of the field.

k-ye (commented):
Thanks for the fix!

k-ye (approved):
LGTM

Issue #2195, opened 2021-03-03
[cleanup] Add googletest unit test framework


k-ye:
**Concisely describe the proposed feature**

We will use `googletest` as our unit test framework.

PR #2201, opened 2021-03-07
[test] Add a basic unit test using googletest

k-ye:
Related issue = #2195 

This PR adds `tests/cpp_new/common/core_test.cpp`. It is just an illustration of how we can write C++ unit tests, with no real valuable test in it. In the future, the directory tree of `tests/cpp_new` can mirror that of `taichi/` (Within Google the tests are usually directly placed along with the source files. Go is also designed like so. But it seems more common to have a separate directory for tests in other places..)

TODO:

Right now we have to manually trigger the test with `./taichi_cpp_tests`. Integrate this into the `ti` CLI.

----

### DIFF
--- a/tests/cpp_new/common/core_test.cpp
+++ b/tests/cpp_new/common/core_test.cpp
@@ -4,7 +4,7 @@
 
 namespace taichi {
 
-TEST(StatementsTest, Basic) {
+TEST(CoreTest, Basic) {
   EXPECT_EQ(trim_string("hello taichi  "), "hello taichi");
 }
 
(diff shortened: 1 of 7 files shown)


tests/cpp_new/common/core_test.cpp
```diff
@@ -0,0 +1,11 @@
+#include "gtest/gtest.h"
+
+#include "taichi/common/core.h"
+
+namespace taichi {
+
+TEST(StatementsTest, Basic) {
```
k-ye, line 7:
```suggestion
TEST(CoreTest, Basic) {
```

yuanming-hu (approved):
Awesome, LGTM! (Sorry about the delayed review...)

Issue #2196, opened 2021-03-03
[cleanup] Untangle the dependencies between the Taichi submodules


k-ye:
Clean up the organization of the code in CMake, such that `taichi_core.so` can be composed from a few smaller libraries. 

The near-term goal is to isolate the part that do not need to interact with `pybind11` into its own library (call it `liba`). `liba` can be linked into both `taichi_core.so` and the unit test executable (#2195 ). This is mainly to make Taichi code unit-testable. Currently, linking `taichi_core.so` into the unit tests doesn't work, and it still requires the full python dependency. (I tried linking the python lib as well, but then got GIL issue and felt like it was not worth the efforts digging into it)

PR #2205, opened 2021-03-08
[refactor] Separate SNode read/write kernels into a dedicated class

k-ye:
Related issue = #2196 

### Purpose

This is one of the steps to decouple `SNode` from `taichi/program`. Because many files inside `taichi/ir` depend on SNode, they transitively depend on `taichi/program` as well. `taichi/program` is very high level, and almost depends on everything in the codebase. So without breaking the tie, it is very hard for us to separate out IR.

Note that this PR alone is not enough (actually, far from enough) for modularizing the IR. I'm thinking of moving out `SNode::expr` later (we may have to have a similar mapping from `SNode` to `expr`). The reason is similar:

1. `expr` is an `Expr`, which depends on the expression and the frontend IR
2. Frontend IR depends on `taichi/program`.

As a short-term goal, I think it will be good enough if we can make `ir.h/cpp`, `statements.h/cpp`, `snode.h/cpp` self contained.

----

### DIFF
--- a/taichi/ir/snode.cpp
+++ b/taichi/ir/snode.cpp
@@ -7,18 +7,6 @@
 
 TLANG_NAMESPACE_BEGIN
 
-// namespace {
-
-// void set_kernel_args(const std::vector<int> &I,
-//                      int num_active_indices,
-//                      Kernel::LaunchContextBuilder *launch_ctx) {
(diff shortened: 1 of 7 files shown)


yuanming-hu (approved):
Cool! LGTM!

xumingkuan (approved):
LGTM!

PR #2206, opened 2021-03-08
[refactor] Move type related utils away from lang_util.h

k-ye:
Related issue = #2196 

The reason for this PR is similar to that of #2205: Because `taichi/lang_util.cpp` depends on `taichi/program`, if any source file in the testable library includes `taichi/lang_util.h`, we will be forced to include the entire `taichi/program` part as well. As a result, I will break down `lang_util` into a few smaller pieces.

----


yuanming-hu (approved):
LGTM, thanks!

PR #2209, opened 2021-03-13
[refactor] Move code away from lang_utils

k-ye:
Related issue = #2196 

This PR is literally moving code around, so that files in `taichi/ir` do not need to include `taichi/lang_util.h`.

----


taichi/ir/ir.cpp
```diff
@@ -92,14 +88,6 @@ IRNode *IRNode::get_ir_root() {
   return node;
 }
 
-Kernel *IRNode::get_kernel() const {
```
k-ye, line 95: These two methods are moved to `taichi/program/ir_node_extended_impl.cpp` because they need the definition of `Kernel` and `Program`.

yuanming-hu (approved):
Awesome, looks great! It's nice to see the code is getting more organized!

PR #2214, opened 2021-03-15
[test] [ir] Add a simple unit test for IRBuilder

k-ye:
Related issue = #2196, #2193

Note that I've switched from `OBJECT`  to `STATIC` now. It could be too laborious to refactor the whole code base to a point where they are layered nicely. Instead, we should just create a (static) library that is unit testable by excludeing `taichi/python` and being python-free. Most of Taichi's dependencies will be merged into this lib, including LLVM, GLEW, etc. ( `target_link_librariess()` doesn't seem to be effective with an `OBJECT` library)

----

### DIFF
--- a/tests/cpp_new/ir/ir_builder_test.cpp
+++ b/tests/cpp_new/ir/ir_builder_test.cpp
@@ -1,0 +1,22 @@
+#include "gtest/gtest.h"
+
+#include "taichi/ir/ir_builder.h"
+#include "taichi/ir/statements.h"
+
+namespace taichi {
+namespace lang {
+
+TEST(IRBuilder, Basic) {
+  IRBuilder builder;
+  auto *lhs = builder.get_int32(40);
+  auto *rhs = builder.get_int32(2);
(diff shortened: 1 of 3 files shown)


tests/cpp_new/ir/ir_builder_test.cpp
```diff
@@ -0,0 +1,22 @@
+#include "gtest/gtest.h"
+
+#include "taichi/ir/ir_builder.h"
+#include "taichi/ir/statements.h"
+
+namespace taichi {
+namespace lang {
+
+TEST(IRBuilder, Basic) {
```
k-ye, line 9: googletest is pretty easy to learn: https://google.github.io/googletest/primer.html. By having these files unit tested, we should be able to finalize our IR/passes API, while still iterating and improving the implementations.

yuanming-hu (approved):
LGTM!

xumingkuan (approved):
LGTM, thanks!

PR #2225, opened 2021-03-20
[ir] Add more statements and tests to IR Builder

xumingkuan:
Related issue = #2193 

This PR adds these statements to the IR Builder:
- All UnaryOpStmts
- All BinaryOpStmts
- TernaryOpStmt
- RangeForStmt, StructForStmt, WhileStmt, IfStmt
- WhileControlStmt, ContinueStmt
- AllocaStmt, LocalLoadStmt, LocalStoreStmt
- LoopIndexStmt

And add the following tests:
- PrintStmt test
- RangeFor test

----

### DIFF
--- a/taichi/ir/ir_builder.cpp
+++ b/taichi/ir/ir_builder.cpp
@@ -49,6 +49,86 @@ Stmt *IRBuilder::create_return(Stmt *value) {
   return insert(Stmt::make<KernelReturnStmt>(value));
 }
 
+Stmt *IRBuilder::create_cast(Stmt *value, DataType output_type) {
+  auto &&result = Stmt::make_typed<UnaryOpStmt>(UnaryOpType::cast_value, value);
+  result->cast_type = output_type;
+  return insert(std::move(result));
+}
+
+Stmt *IRBuilder::create_bit_cast(Stmt *value, DataType output_type) {
+  auto &&result = Stmt::make_typed<UnaryOpStmt>(UnaryOpType::cast_bits, value);
+  result->cast_type = output_type;
+  return insert(std::move(result));
+}
+
+Stmt *IRBuilder::create_neg(Stmt *value) {
+  return insert(Stmt::make<UnaryOpStmt>(UnaryOpType::neg, value));
+}
+
+Stmt *IRBuilder::create_not(Stmt *value) {
+  return insert(Stmt::make<UnaryOpStmt>(UnaryOpType::bit_not, value));
+}
+
+Stmt *IRBuilder::create_logical_not(Stmt *value) {
+  return insert(Stmt::make<UnaryOpStmt>(UnaryOpType::logic_not, value));
+}
+
+Stmt *IRBuilder::create_floor(Stmt *value) {
+  return insert(Stmt::make<UnaryOpStmt>(UnaryOpType::floor, value));
+}
+
+Stmt *IRBuilder::create_ceil(Stmt *value) {
+  return insert(Stmt::make<UnaryOpStmt>(UnaryOpType::ceil, value));
+}
+
+Stmt *IRBuilder::create_abs(Stmt *value) {
+  return insert(Stmt::make<UnaryOpStmt>(UnaryOpType::abs, value));
+}
+
+Stmt *IRBuilder::create_sgn(Stmt *value) {
+  return insert(Stmt::make<UnaryOpStmt>(UnaryOpType::sgn, value));
+}
+
+Stmt *IRBuilder::create_sqrt(Stmt *value) {
+  return insert(Stmt::make<UnaryOpStmt>(UnaryOpType::sqrt, value));
+}
+
+Stmt *IRBuilder::create_rsqrt(Stmt *value) {
+  return insert(Stmt::make<UnaryOpStmt>(UnaryOpType::rsqrt, value));
+}
+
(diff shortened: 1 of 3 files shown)


k-ye (approved):
LGTM!

xumingkuan:
@k-ye Sorry for adding a commit after approving... Could you please re-review this PR?

BTW, it seems that the format server is not working.

taichi/ir/ir_builder.h
```diff
@@ -5,12 +5,13 @@
 TLANG_NAMESPACE_BEGIN
 
 class IRBuilder {
- private:
+ public:
   struct InsertPoint {
     Block *block;
```
k-ye, line 10:
```suggestion
    Block *block{nullptr};
```

taichi/ir/ir_builder.h
```diff
@@ -5,12 +5,13 @@
...
     int position;
```
k-ye, line 11:
```suggestion
    int position{0};
```

taichi/ir/ir_builder.h
```diff
@@ -5,12 +5,13 @@
...
   };
 
+ private:
   std::unique_ptr<IRNode> root_;
```
k-ye, line 15:
```suggestion
  std::unique_ptr<IRNode> root_{nullptr};
```

taichi/ir/ir_builder.cpp
```diff
@@ -10,7 +11,87 @@ IRBuilder::IRBuilder() {
...
+    TI_ERROR("Statement {} is not a loop.", loop->name());
+  }
+}
+
+void IRBuilder::set_insertion_point_to_true_branch(Stmt *if_stmt) {
+  TI_ASSERT(if_stmt->is<IfStmt>());
+  set_insertion_point({if_stmt->as<IfStmt>()->true_statements.get(), 0});
+}
+
+void IRBuilder::set_insertion_point_to_false_branch(Stmt *if_stmt) {
```
k-ye, line 51: ditto

k-ye (approved):
LGTM with nits

taichi/ir/ir_builder.cpp
```diff
@@ -10,7 +11,87 @@ IRBuilder::IRBuilder() {
...
+    set_insertion_point({struct_for->body.get(), 0});
+  } else if (auto while_stmt = loop->cast<WhileStmt>()) {
+    set_insertion_point({while_stmt->body.get(), 0});
+  } else {
+    TI_ERROR("Statement {} is not a loop.", loop->name());
+  }
+}
+
+void IRBuilder::set_insertion_point_to_true_branch(Stmt *if_stmt) {
+  TI_ASSERT(if_stmt->is<IfStmt>());
```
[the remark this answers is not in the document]
xumingkuan: Then we need to include each `XxxStmt`'s name in `ir.h`. Since we actually do that now, maybe it's fine with that?

And for convenience, maybe we want to set the return values of `IRBuilder::create_xxx` to the corresponding `XxxStmt` so that we can write `builder.set_insertion_point_to_true_branch(builder.create_if(...))` without more type casts.

taichi/ir/ir_builder.h
xumingkuan: This is actually set in the constructor... Do we need to initialize it here?

k-ye: Yeah, let's keep the style consistent

taichi/ir/ir_builder.cpp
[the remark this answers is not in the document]
k-ye: >  Since we actually do that now, maybe it's fine with that?

Hmm, could you share how long it takes for you to re-compile when `ir.h` is changed? (Also, how many threads are you using typically?) If it is too slow for you, would you take a look at `ir.h` and try to split it into smaller pieces? 

That said, for IRBuilder, by definition it will need to know all the stmts, so I'm not sure if/how we can avoid including the contents in `ir.h`.

> And for convenience, maybe we want to set the return values of IRBuilder::create_xxx to the corresponding XxxStmt so that we can write builder.set_insertion_point_to_true_branch(builder.create_if(...)) without more type casts.

Yeah, that sounds like a good idea :)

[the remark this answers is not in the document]
xumingkuan: > Hmm, could you share how long it takes for you to re-compile when ir.h is changed? (Also, how many threads are you using typically?)

It took me 4m17s just now with 8 threads. I think it's acceptable...

BillXu2000 (commented):
LGTM

taichi/ir/ir_builder.cpp
```diff
@@ -10,7 +11,87 @@ IRBuilder::IRBuilder() {
...
+
+void IRBuilder::set_insertion_point_to_after(Stmt *stmt) {
+  set_insertion_point({stmt->parent, stmt->parent->locate(stmt) + 1});
+}
+
+void IRBuilder::set_insertion_point_to_before(Stmt *stmt) {
+  set_insertion_point({stmt->parent, stmt->parent->locate(stmt)});
+}
+
+void IRBuilder::set_insertion_point_to_loop_begin(Stmt *loop) {
```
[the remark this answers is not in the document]
xumingkuan: Done

PR #2227, opened 2021-03-21
[Example] Reduce c calculation from pixel loop

arisliang:
value `c` is not dependent on pixels, moving it outside of pixel loop to reduce calculation needed.

### DIFF
--- a/examples/fractal.py
+++ b/examples/fractal.py
@@ -13,8 +13,8 @@ def complex_sqr(z):
 
 @ti.kernel
 def paint(t: float):
+    c = ti.Vector([-0.8, ti.cos(t) * 0.2])
     for i, j in pixels:  # Parallized over all pixels
-        c = ti.Vector([-0.8, ti.cos(t) * 0.2])
         z = ti.Vector([i / n - 1, j / n - 0.5]) * 2
         iterations = 0
         while z.norm() < 20 and iterations < 50:


arisliang:
This shouldn't change the outermost loop and therefore parallelization of the pixel loop, is it?

k-ye:
Thanks for the contribution! There are actually quite a few details for this change:

1. A variable that is global to the scope of @ti.kernel (i.e., outside the parallel for-loop, but still within the kernel body) is actually stored in a device buffer, for it to be accessible by each GPU thread. This change would result in a read to that global buffer, instead of the thread local register or memory. 
2. Internally, each parallel for-loop translates to a CUDA kernel, the serial part in between also translates to a single-thread CUDA kernel. So this change would result in two CUDA kernel launches instead of one. The overhead might very well be negligible for such a simple example, but it is something to be aware of.

arisliang:
Yeah, the low-level details is needed to be aware of. That's why this PR. If we need to do something global in ti.kernel scope, potentially more resource intensive, will it be better to put it in the global scope to be accessible by local thread, or repeat the calculation in each thread?

Also a question on your 2nd comment `two CUDA kernel launches instead of one`. Does this mean the number of CUDA kernel will be doubled (2 * N), not just plus 1 (N + 1), for the global variable in ti.kernel scope?

k-ye:
> will it be better to put it in the global scope to be accessible by local thread, or repeat the calculation in each thread?

I think this is a case-by-case question. Ultimately it is the computation resource vs memory bandwidth resource. But I agree that the more intensive the computation gets, the better to compute it once and reuse it among threads.

> Does this mean the number of CUDA kernel will be doubled (2 * N), not just plus 1 (N + 1), for the global variable in ti.kernel scope?

Good point, it is (N + 1). Sorry for not clarify this. Say we have this kernel:

```python
@ti.kernel
def foo():
  for i in x:  # 1
    do_something_with(x[i])
  for i in y:  # 2
    do_something_with(y[i])
```

Taichi translates each for-loop `1` and `2` into a parallelized CUDA kernel. (In some settings where `x` and `y` are of the same shape, it can be fused. But that's a more advanced topic)

Now if we add some statements in between:

```python
@ti.kernel
def foo():
  for i in x:  # 1
    do_something_with(x[i])
  do_other_things()  # 3
  for i in y:  # 2
    do_something_with(y[i])
```

Because `3`'s grid layout is different (it must be executed on a single CUDA thread), so Taichi will need to emit 3 CUDA kernels.

arisliang:
By turning on `ti.init(arch=ti.gpu, print_preprocessed=True)`, the final AST shows 
the passed-in parameter `t = ti.decl_scalar_arg(float)` is also at the outermost scope. Does the passed-in parameters trigger the launch of a single-thread CUDA kernel? Or not, because there's no calculation to the parameter?

arisliang:
Think I'm clear for now, thanks for the clarification!

Issue #2219, opened 2021-03-18
Make taichi-gardener's CI trigger the required workflows


k-ye:
While we have made @taichi-gardener 's CI inherit from the previous run's result, the coverage this CI job is not enough to merge the PR. The problem is that it doesn't trigger the workflows required for merging into `master`. We should make @taichi-gardener 's CI trigger these required workflows (Build on Linux with Python 3.6, and Code Format) as well. It takes a bit longer to run, but at least we will have a way to merge PR.

Close #1809, which is superseded by this.

k-ye:
Fixed by #2229

PR #2229, opened 2021-03-22
[misc] Make taichi-gardener also run the required CI jobs

k-ye:
Related issue = #2219

* Before this change, @taichi-gardener  won't trigger the *required* "Ubuntu 3.6 Build and Test (CPU)" --> We have to do another push.
* After this change, @taichi-gardener 's CI will be slower (maybe by another 20 minutes). But we can at least merge PRs if there are no other issue.

Long term speaking, I see that TiDB's workflow is to manually trigger the CI by replying some magic keywords (e.g. "/run-all-tests"), which the bots will listen to. Maybe that's a better way to utilize CI resources.

----

### DIFF
--- a/.github/workflows/persubmit.yml
+++ b/.github/workflows/persubmit.yml
@@ -4,16 +4,57 @@ on:
     types: [opened, synchronize, reopened]
 
 jobs:
+  build_and_test_cpu_required:
+    # This job will be required to pass before merging to master branch.
+    name: Required Build and Test (CPU)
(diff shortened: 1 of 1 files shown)


yuanming-hu (approved):
Awesome! Hopefully, the workflow is less intimidating/confusing to new contributors now.

k-ye:
Not sure how long it takes for `Required Build and Test (CPU) (ubuntu-latest, 3.6) (pull_request)` to propagate to the settings..

PR #2231, opened 2021-03-23

k-ye:
Related issue = #2223 

----

### DIFF
--- a/python/taichi/core/__init__.py
+++ b/python/taichi/core/__init__.py
@@ -2,6 +2,7 @@ from taichi.core.util import *
 from taichi.core.settings import *
 from taichi.core.record import *
 from taichi.core.logging import *
+from taichi.core.primitive_types import *
 
 ti_core.build = build
 ti_core.load_module = load_module
--- a/python/taichi/core/primitive_types.py
+++ b/python/taichi/core/primitive_types.py
@@ -1,0 +1,60 @@
+from taichi.core.util import ti_core
+
+# Real types
+
+float32 = ti_core.DataType_f32
+f32 = float32
+float64 = ti_core.DataType_f64
+f64 = float64
+
+real_types = [f32, f64, float]
+real_type_ids = [id(t) for t in real_types]
+
+# Integer types
+
+int8 = ti_core.DataType_i8
+i8 = int8
+int16 = ti_core.DataType_i16
+i16 = int16
+int32 = ti_core.DataType_i32
+i32 = int32
+int64 = ti_core.DataType_i64
+i64 = int64
+
+uint8 = ti_core.DataType_u8
+u8 = uint8
+uint16 = ti_core.DataType_u16
+u16 = uint16
+uint32 = ti_core.DataType_u32
+u32 = uint32
+uint64 = ti_core.DataType_u64
+u64 = uint64
+
+integer_types = [i8, i16, i32, i64, u8, u16, u32, u64, int]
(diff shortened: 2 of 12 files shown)


python/taichi/core/primitive_types.py
```diff
@@ -0,0 +1,60 @@
+from taichi.core.util import ti_core
```
k-ye, line 1: This is probably a special case where we need to import `ti_core` directly, because the rest of the type definitions are directly executed during the import. It doesn't make a difference if we use `util.ti_core` here.

python/taichi/core/__init__.py
```diff
@@ -2,6 +2,7 @@
 from taichi.core.settings import *
 from taichi.core.record import *
 from taichi.core.logging import *
+from taichi.core.primitive_types import *
```
k-ye, line 5: Not sure how I feel about this... It's bad to do wildcard import. On the other hand, there are arguments saying that doing this in `__init__.py` is an exception, in order to lift the symbols into the package level. If not written this way, we would have to write `from taichi.x import float32, f32, float64, f64, ..., u64` twice, which is quite repetitive (and equally hard to maintain). What's your opinion :) ?

python/taichi/lang/__init__.py
```diff
@@ -3,13 +3,17 @@
...
+                                     data_oriented, func, kernel, pyfunc)
 from taichi.lang.matrix import Matrix, Vector
 from taichi.lang.ndrange import GroupedNDRange, ndrange
+from taichi.lang.ops import *
 from taichi.lang.quant_impl import quant
 from taichi.lang.runtime_ops import async_flush, sync
 from taichi.lang.transformer import TaichiSyntaxError
 from taichi.lang.type_factory_impl import type_factory
-from taichi.lang.util import deprecated
+from taichi.lang.util import *
```
k-ye, line 16: Will fix later. The reason we don't have do to this wildcard import previously is because it was imported inside `taichi/lang/impl.py`.

python/taichi/lang/impl.py
```diff
@@ -589,29 +587,23 @@ def static(x, *xs):
...
 def current_cfg():
-    return taichi_lang_core.current_compile_config()
+    return cutil.ti_core.current_compile_config()
 
 
 def default_cfg():
-    return taichi_lang_core.default_compile_config()
-
-
-from .kernel import *
```
k-ye, line 611: Ouch..

python/taichi/lang/matrix.py
```diff
@@ -1,13 +1,17 @@
...
 import numbers
-import numpy as np
-from .util import taichi_scope, python_scope, deprecated, to_numpy_type, to_pytorch_type, in_python_scope, is_taichi_class, warning
-from .common_ops import TaichiOperations
-from .exception import TaichiSyntaxError
 from collections.abc import Iterable
 
+import numpy as np
+from taichi.lang import expr, impl
+from taichi.lang import ops as ops_mod
```
k-ye, line 7: There are (local) variables named `ops`, therefore the alias

python/taichi/core/__init__.py
yuanming-hu: I think the current solution should be good, given `taichi.core.primitive_types` already clearly defined the `__all__` variable, listing the names to export :-)

yuanming-hu (approved):
Awesome! LGTM. It's so nice to see the imports are now much more organized than the previous version!

PR #2234, opened 2021-03-25
[refactor] Add ASTResolver to properly resolve AST nodes to functions

k-ye:
Related issue = #2223 

The idea is to consult the `globals()` scope to make sure functions like `ti.static` actually resolve to Taichi functions during AST transformation.

----

### DIFF
--- a/python/taichi/lang/ops.py
+++ b/python/taichi/lang/ops.py
@@ -2,7 +2,7 @@ import builtins
 import ctypes
 import functools
 import math
-import operator as ops
+import operator as _bt_ops_mod  # bt for builtin
 import traceback
 
 from taichi.lang import impl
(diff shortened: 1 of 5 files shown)


python/taichi/lang/ops.py
```diff
@@ -2,7 +2,7 @@
 import ctypes
 import functools
 import math
-import operator as ops
+import operator as _bt_ops_mod  # bt for builtin
```
k-ye, line 5: Previously this has overridden our `ops.py` module, renaming this here for unittests.

python/taichi/lang/transformer.py
```diff
@@ -543,6 +538,7 @@ def visit_For(self, node):
             if decorator != '' and len(node.iter.args) == 1:
                 double_decorator = self.get_decorator(node.iter.args[0])
             ast.fix_missing_locations(node)
+            print(f'decorator={decorator} double_decorator={double_decorator}')
```
k-ye, line 541:
```suggestion
```

yuanming-hu (approved):
Looks great! It's so nice to see that the previous string-based hacky keyword matching is now done much more systematically :-)

PR #2245, opened 2021-04-03
[lang] Replaced usage of taichi.lang.core with taichi.core.util.ti_core.

Leonz5288:
Issue=#2223

The way we use ti_core is not clear and a bit confusing. Multiple repeat imports can make code redundant and hard to understand. This PR refactor the part in python/taichi/lang by:
Deleted core.py in under python/taichi/lang and replaced all usage of taichi_lang_core with _ti_core imported from taichi.core.util.ti_core. 
@k-ye 

----


k-ye (approved):
This is great, thank you!

PR #2246, opened 2021-04-04
[test] [ir] Add tests for the alias_analysis pass

k-ye:
Related issue = #2193 

----

### DIFF
--- a/tests/cpp_new/analysis/alias_analysis_test.cpp
+++ b/tests/cpp_new/analysis/alias_analysis_test.cpp
@@ -1,0 +1,119 @@
+#include <algorithm>
+#include <memory>
+#include <vector>
+
+#include "gtest/gtest.h"
+#include "taichi/ir/analysis.h"
+#include "taichi/ir/ir_builder.h"
+#include "taichi/ir/snode.h"
+#include "taichi/ir/statements.h"
+
+namespace taichi {
+namespace lang {
+namespace irpass {
+namespace analysis {
+namespace {
+
+std::vector<Stmt *> make_const_indices(const std::vector<int> indices,
+                                       IRBuilder *builder) {
+  std::vector<Stmt *> result;
+  result.reserve(indices.size());
+  std::for_each(indices.begin(), indices.end(), [builder, &result](int i) {
+    result.push_back(builder->get_int32(i));
+  });
+  return result;
+}
+
+}  // namespace
+
+TEST(AliasAnalysis, Alloca) {
+  IRBuilder builder;
+
+  auto *alloca1 = builder.create_local_var(PrimitiveType::f32);
+  auto *alloca2 = builder.create_local_var(PrimitiveType::f32);
+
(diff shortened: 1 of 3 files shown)


tests/cpp_new/analysis/alias_analysis_test.cpp
```diff
@@ -0,0 +1,119 @@
...
+  for (int i = 0; i < 2; ++i) {
+    snode->physical_index_position[i] = i;
+  }
+  snode->extractors[snode->physical_index_position[2]].trailing_bits = 2;
+
+  IRBuilder builder;
+  // The difference along the 2-dimension is (6 -3) = 3, which is smaller than
+  // (1 << trailing_bits) = 4.
+  const auto indices1 = make_const_indices({1, 2, 3}, &builder);
+  const auto indices2 = make_const_indices({1, 2, 6}, &builder);
```
xumingkuan, line 77: Well... IIUC `a[1, 2, 3]` and `a[1, 2, 6]` are **different** elements when `trailing_bits == 2`.

Maybe we should change
```C++
      if (!diff.first) {
        uncertain = true;
      } else if (std::abs(diff.second) >=
                 (1 << snode->extractors[snode->physical_index_position[i]]
                           .trailing_bits)) {
        return AliasResult::different;
      }
```
to
```C++
      if (!diff.first || diff.second != 0) {
        uncertain = true;
      }
      if (std::abs(diff.second) >=
          (1 << snode->extractors[snode->physical_index_position[i]]
                           .trailing_bits)) {
        return AliasResult::different;
      }
```

xumingkuan (commented):
Thanks!

tests/cpp_new/analysis/alias_analysis_test.cpp
k-ye: Hmm, could you explain the original intention when writing this in https://github.com/taichi-dev/taichi/pull/1657?

xumingkuan:
Well... For `trailing_bits == 2`, indices `0, 1, 2, 3` point to the same address, and so do `4, 5, 6, 7`, `8, 9, 10, 11`, etc.


So when two indices are different by >= 4, we are sure that they point to the different address.


However, I forgot to add the condition that if two indices are different by < 4, we are not sure if they point to the same address...







在 2021-04-06 08:44:04，"Ye Kuang" ***@***.***> 写道：

@k-ye commented on this pull request.

In tests/cpp_new/analysis/alias_analysis_test.cpp:

> +  // (1 << trailing_bits) = 4.
+  const auto indices1 = make_const_indices({1, 2, 3}, &builder);
+  const auto indices2 = make_const_indices({1, 2, 6}, &builder);


Hmm, could you explain the original intention when writing this in #1657?

tests/cpp_new/analysis/alias_analysis_test.cpp
k-ye: Thanks, fixed

xumingkuan (approved):
LGTM, thanks!

PR #2247, opened 2021-04-04
[misc] Sort python import order

taichi-gardener:
Related issue = #2223 

This PR adds `isort` to the python code formatter, which helps sort the import orders. Right now it is not a required package, as we may need to update the format server first.

Also sorted the contents in `requirements.txt`

----


k-ye (approved):

yuanming-hu (approved):
LGTM!

Issue #2177, opened 2021-01-26
[Bug] incorrect loop indices with mutlilevel structs


KLozes:
**Describe the bug**
Struct for loop indices seem to be wrong when a multilevel struct is used. note that b is placed at the block level, while a is at the leaf level.

**To Reproduce**
```py
import taichi as ti
ti.init(arch=ti.cpu)

a =  ti.field(ti.f32);
b = ti.field(ti.f32);

blocks = ti.root.dense(ti.i, 4)
blocks.place(b)
blocks.dense(ti.i, 4).place(a)

@ti.kernel
def init():
    for i in a:
        a[i] = 1.0
        print(i)

init()

for i in range(4*4):
    assert(a[i] == 1.0)
```

**Log/Screenshots**
I would expect this to print 0 through 15, but it does not. If field b is removed the struct for does print 0 through 15.  Oddly, though, the assert does not trip even though indices look messed up. 
```
[Taichi] mode=development
[Taichi] preparing sandbox at /tmp/taichi-ia11g84x
[Taichi] <dev mode>, llvm 10.0.0, commit 368157a3, linux, python 3.6.9
[Taichi] Starting on arch=x64
[Taichi] materializing...
0
2
3
4
5
1
2
3
1
2
3
4
3
4
5
6

```

There are some differences in the ir when placing b, compared to when it is not placed. 

When placing b:
```
[I 01/27/21 20:44:18.351] [compile_to_offloads.cpp:operator()@19] [init_c4_0] Bit struct stores optimized:
kernel {
  $0 = offloaded range_for(0, 16) grid_dim=0 block_dim=4 
  body {
    <i32> $1 = loop $0 index 0
    <i32> $2 = const [2]
    <i32> $3 = bit_sar $1 $2
    <i32> $4 = const [3]
    <i32> $5 = bit_and $3 $4
    <i32> $6 = bit_and $1 $4
    <i32> $7 = add $5 $6
    <f32> $8 = const [1.0]
    <*gen> $9 = get root
    <i32> $10 = const [0]
    <*gen> $11 = [S0root][root]::lookup($9, $10) activate = false
    <*gen> $12 = get child [S0root->S1dense] $11
    <i32> $13 = bit_and $7 $4
    <*gen> $14 = [S1dense][dense]::lookup($12, $13) activate = false
    <*gen> $15 = get child [S1dense->S3dense] $14
    <*gen> $16 = [S3dense][dense]::lookup($15, $13) activate = false
[...]

Issue #1682, opened 2020-08-11
Problems with SNodes


florianHofherr:
First, thanks for the awesome work! I've just started to use Taichi, so I am not entirely sure if this might be a bug or just not enough understanding from my side. I hope I don't bother you with a completely silly question

**Describe the bug**
If I try to reproduce the example for `snode.shape()` from the documentation of the Structural nodes I do not get the expected dimensions. The child node seems to influence the parent node

A second issue is, that the example for `snode.place()` throws an assertion error, which from my understanding I feel it shouldn't

**To Reproduce**
```
w = ti.field(ti.f32)
x = ti.field(ti.f32)
y = ti.field(ti.f32)
z = ti.field(ti.f32)

blk1 = ti.root
blk2 = blk1.dense(ti.i,  3)
blk3 = blk2.dense(ti.jk, (5, 2))
blk4 = blk3.dense(ti.k,  2)

blk1.place(w)
blk2.place(x)
blk3.place(y)
blk4.place(z)

print(w.shape)
print(x.shape)
print(y.shape)
print(z.shape)
````
Output
```
[Taichi] Starting on arch=x64
[Taichi] materializing...
()
(3,)
(3, 5, 4)
(3, 5, 4)
```
From my understanding `y.shape` should yield `(3,5,2)` since the child nodes inherit the structure from the parents, but do not influence them

Regarding the second point: The following code throws an assertion error on my machine (example from the documentation)
```
x = ti.field(dtype=ti.i32)
y = ti.field(dtype=ti.f32)
ti.root.place(x, y)
assert x.snode == y.snode
```

**Comment by archibate** [comment]
Hi, thank for reporting. The first one is likely because of the recently added #1558. Not sure if this is a bug or a feature. @yuanming-hu Is this intentional?

x = ti.field(dtype=ti.i32)
y = ti.field(dtype=ti.f32)
ti.root.place(x, y)
assert x.snode == y.snode

It seems to me that the doc is wrong. This is because `x.snode` and `y.snode` is of `SNodeType::place`.
We should use `assert x.snode.parent == y.snode.parent` instead. Where the parent is of `SNodeType::root`. Thank for pointing out!

**Comment by aryansoman** [comment]
Hi, I don't know if this has been fixed, but 
```
x.snode.parent == y.snode.parent 
```
[...]

PR #2248, opened 2021-04-06
[Bug] [llvm] [ir] Complex multilevel structs. Consecutive indices for all snodes Draft

KLozes:
Related issue = #2177 

I think this is pretty important PR. It opens the door for much more complicated structures!

This PR addresses the peculiar behavior seen in listgen and accessing fields with more complex multilevel structs, see the issue above. As @k-ye discovered (thanks!), it is because we assume the first child of an snode when we compute the 'start' member of the extractors. 

So I've done away with the 'start' member entirely, and instead they are now calculated on-the-fly where they are needed in demote_struct_for and lower_access IR passes. This was necessary because the list of start values is unique to the leaf-to-root path of the filed being accessed.
 
I also was able to get rid of the trailing_bits member, since every field can now be accessed independently of the shape of other fields. Which I think is more intuitive.

This PR also changes the definition of indices for parents of leaf snodes. for example, the structure

``` python 

root.dense(ti.i, 4).dense(ti.i, 4).place(x)
```
used to have the indices below. I believe 0,1,2 or 3 could be used the access the first element of dense0.
```
dense1:  0 1 2 3   4 5 6 7   8 9 10 11  12 13 14 15 

dense0:  0         4         8          12

root:    0
```

with this PR, dense0 has unique indices. see the changes to test_sparse_deactivate.py
```
dense1:  0 1 2 3   4 5 6 7   8 9 10 11  12 13 14 15 

dense0:  0         1         2           3

root:    0
```

If someone can figure out a fix for bls that would be really helpful. My attempt did not work, but I think it is on the right track. 

----


k-ye:
Hi @KLozes ,

Thanks for proposing this change! It's not a trivial change so I will list my understanding and thoughts here.

The trailing bits feature was introduced back in https://github.com/taichi-dev/taichi/pull/1558. It adds two things:

1. It requires that the shapes of all the children at a given dimension (index) are the same. This invalidates such kind of usage: https://github.com/taichi-dev/taichi/blob/0035ec061a000e77f8507e6240ed8be6c8d59eb8/tests/python/test_trailing_bits.py#L47-L50
2. Shapes from other indices of other children are "attached" to a child SNode. Effectively, the other children's shapes span the shape of a given child. This is done in https://github.com/taichi-dev/taichi/blob/0035ec061a000e77f8507e6240ed8be6c8d59eb8/taichi/struct/struct.cpp#L152-L163. Say if we have

```py
blk = ti.root.dense(ti.i, 2)
a = blk.dense(ti.i, 4)
b = blk.dense(ti.j, 4)
```
The `i`-th dimension of `a` gets attached to `b`, so `b`'s shape is `(2x4, 4) = (8, 4)`.

I can see how these would result in some confusion, such as #1682 . @yuanming-hu probably knows better about the use case. 

> it is because we assume the first child of an snode when we compute the 'start' member of the extractors.

Right. But this was a bug where Taichi didn't properly enforce `1`. I.e., if Taichi fixed this bug, the example in #2177 would result in Taichi throwing a "Inconsistent trailing bit configuration" error. As you've pointed out, whether to keep this feature is another question.

> I also was able to get rid of the trailing_bits member, since every field can now be accessed independently of the shape of other fields. 

That's the question to discuss...

> This PR also changes the definition of indices during for parents of leaf snodes. 

Right, I can see this is more intuitive. But one property Taichi needs to keep is that, when accessing some intermediary SNode with the leaf SNode's index, it will go to the same ancestor (if the leaf SNodes are from the same ancestor). In your example,  `dense0[0]`, `dense0[1]`, `dense0[2]`, `dense0[3]` will all point to the same cell. (We can see if there is a way to remove the necessity of this property to make the implementation more intuitive)

---

Appendix: Some notes on `IndexExtractor` for the future me... This struct records the information about a particular dimension (index).

* num_bits: Number of bits required to store the index. E.g. if a SNode is of shape `(ti.ij, (4, 8))`. Then it needs 2 bits for the i-th dimension and 3 bits for the j-th.
* num_elements: Max number of elements (not padded to the power-of-two)
* start: This is the starting bit index of the snode for a coordinate value at the given dimension . It characterizes the relationship between a parent and a child snode.  `parent.start = child.start + child.num_bits`. For example, if there are two 1D snodes `a` and `b`, where `b = a.dense(ti.i, 8)`, we have `a.start = b.start + 3` for the `i`-th dimension. When accessing `b[15]`, Taichi first goes through `a` with `(15 >> 3) = 1`.     https://github.com/taichi-dev/taichi/blob/0035ec061a000e77f8507e6240ed8be6c8d59eb8/taichi/struct/struct.cpp#L65-L67 
* acc_offset: Accumulated offset. This is the starting bit index within a flattened loop index value. It describes the relationship of the bit offsets between different dimensions for a given snode. `extractor[i].acc_offset = extractor[i + 1].acc_offset + num_bits`. For example, for a 3D snode `a` of `(ti.ijk, (2, 4, 8))` and a flattened loop index `li`, bits [0, 3) in `li` are for the `k`-th dim, bits [3, 5) are for the j-th, and bits [5, 6) the `i`-th.  
https://github.com/taichi-dev/taichi/blob/0035ec061a000e77f8507e6240ed8be6c8d59eb8/taichi/struct/struct.cpp#L68-L78 
* trailing_bits: This is quite similar to `start`, i.e. `parent.trailing_bits = child.trailing_bits + child.num_bits`. However, after this bottom-up compute, it goes through another top-down pass. https://github.com/taichi-dev/taichi/blob/0035ec061a000e77f8507e6240ed8be6c8d59eb8/taichi/struct/struct.cpp#L152-L163 This is useful when a parent has multiple children that have used different dimensions. The effect can be seen from this test case:

https://github.com/taichi-dev/taichi/blob/0035ec061a000e77f8507e6240ed8be6c8d59eb8/tests/python/test_trailing_bits.py#L8-L16

KLozes:
HI @k-ye 

So my idea here is that trailing bits concept no longer exists. Every field will have the same indices as if it were the only field placed in the the program.  So the structure in trailing bit test
``` python
    x = ti.field(ti.f32)
    z = ti.field(ti.f32)

    block = ti.root.pointer(ti.i, 8)
    block.dense(ti.i, 32).place(x)
    block.dense(ti.i, 16).place(z)
```
 is now a valid data structure. The x field will have indices [0,1...256) and z will have indices [0,1,...128). The key was to compute the 'start' value on the fly during access_lowering so that we can correctly access the data structure based on the leaf to root path.

k-ye:
Right, this PR basically reverts what #1558 has introduced. IIRC, it was added mainly to support this use case:

```py
parent = # some SNode
parent.dense(ti.ijk, 8).place(blk)
parent.dynamic(ti.l, 256).place(lst)

for I in ti.grouped(blk):
  # Every 8x8x8 dense `blk` cells share the same dynamic `lst`. 
  # So we can use `I` to access the sibling `lst` cell. 
```

This pattern is now used in the taichi MPM solvers.

Without trailing bits, sharing the same `lst` cell among the 8x8x8 `blk` cells requires one to manually divide the index by `(8, 8, 8)`, which is kind of repetitive. Ideally, we should only need to define the shape of the SNodes once in the program. In another  word, this violates the design goal that the computation is decoupled from the data structure.

I think what we probably need here is an API, `rescale_index(index, from_snode, to_snode)`, that returns `index * shape(to_snode) / shape(from_snode)`. The API itself may need to have additional preconditions, e.g. one snode must be the ancestor of the other, etc. Hopefully this strikes a balance between keeping taichi's hierarchical data structure intuitive and satisfying the need of the MPM solvers.

As of this PR, because the MPM solvers already depend on the trailing bits, a more reasonable approach would be to first add that API, migrate the solvers to that and then deprecate the trailing bits.

k-ye:
Hi @KLozes , there are a few updates.

We are trying to finalize the definition of this behavior in the eventual v1.0 release. This includes the following items:

1. Do we need trailing bits that result in the orthogonal shapes from the sibling snode being attached to this snode (and its ancestors)? If not, what's the alternative to the existing use case?
1. Do we want to access the parent snode using consecutive index (`0, 1, 2...`) or the child index (`0, 4, 8, ...`). If the former, what are the solutions to migrate the existing code? One possible solution is some API like `rescale_index()`. Alternatively, the index itself can record which Snode it comes from. And when using this index to access some parent SNode, the compiler knows how to convert the index automatically. (This works for `struct-for` indices, what about `range-for`? It seems like an API that takes in both the child and the ancestor SNode will be more universal.) 
1. Do we want to enforce that the shapes of the same dimension of two sibling snodes must be the same? This is necessary in order to access the parent block using the child index. But if 2 uses the consecutive indexing, this is no longer a requirement.
1. The above decisions should also handle field offsets correctly.
2. Provide a concrete and formal documentation of how the index of an ancestor snode is computed with respect to the leaf snode.
2. Avoid padding the shape to power-of-two as much as possible.

KLozes:
Hi @k-ye

I personally think having consecutive indices for all snodes is the most intuitive. It also naturally avoids any limits on the shape snodes, which expands the applicability of Taichi greatly. 

I think an API to rescale the indices is a good idea. Perhaps the API could be similar to the 'offset' feature? I think this would basically set the trailing bits of dynamic to [3,3,3,0]. And all ancestors would have 0 trailing bits. The trailing bits would then be added to numbits during refine_cordinates, access_lowering and demote_dense_struct_for.

``` python
parent = # some SNode
parent.dense(ti.ijk, 8).place(blk)
parent.dynamic(ti.l, 256).place(lst, rescale=[8,8,8,1])
```

Without this API though, porting existing code applications to the new indexing scheme should be easy. I already fixed the tests. And I think only quadtree.py, taichi_sparse.py, and taichi_elements are left to be fixed.

k-ye:
Thanks! Yeah, I agree and think the following design could make the entire SNode structure more intuitive:

1) Use consecutive indexing at each layer of the SNode chain.
2) An SNode's shape is purely determined by the path from root to itself. I.e. the shape of an SNode will not play a role in deciding that of its sibling or ancestors.

> Perhaps the API could be similar to the 'offset' feature? This would probably be achieved by setting the trailing bits of dynamic to [3,3,3,0]. 

Personally I think it's better to have a function that does this rescaling purely on the index, rather than making it another argument:

1. If we want to re-design this part, maybe it makes sense to consider remove the trailing bits all together (which is also where this PR starts). 
2. `rescale=[8,8,8,1]` means that the users will still need to repeat the definition of `blk`'s shape. It's also unlikely that there will be a use case where this rescaling vector is something other than some SNode's shape (i.e., while `(5, 6, 7, 1)` is technically valid in the 8x8x8 block case, I doubt anyone will need this).
3. Given that this feature is not widely used, I think it's more reasonable to group this complexity into a special API, rather than making it part of the SNode definition API.

If we have a API that takes in an index and two SNodes, Taichi will know how to generate code to convert the index back and forth.

---

IIUC, the index rescaling feature is used in these two places in the MPM solver:

1. https://github.com/taichi-dev/taichi_elements/blob/089ac09501e633eaa0f6da17513007a68c0b8c95/engine/mpm_solver.py#L201-L204

```py
for p in self.x:
    base = int(ti.floor(self.x[p] * self.inv_dx - 0.5))
    ti.append(self.pid.parent(), base - ti.Vector(list(self.offset)),
              p)
```

Without the trailing bits feature, the rescaling requires us to divide `base` by `(8,8,8,1)`.

2. https://github.com/taichi-dev/taichi_elements/blob/089ac09501e633eaa0f6da17513007a68c0b8c95/engine/mpm_solver.py#L212-L216

```py
for I in ti.grouped(self.pid):
    p = self.pid[I]
    base = ti.floor(self.x[p] * self.inv_dx - 0.5).cast(int)
    for D in ti.static(range(self.dim)):
        base[D] = ti.assume_in_range(base[D], I[D], 0, 1)
```

Without the trailing bits feature, the rescaling requires us to multiply `I` by `(8,8,8,1)`.

KLozes:
> 1. Use consecutive indexing at each layer of the SNode chain.
  > 2. An SNode's shape is purely determined by the path from root to itself. I.e. the shape of an SNode will not play a role in deciding that of its sibling or ancestors.

Yes! I'm glad we agree. This is exactly what I'm trying to achieve with this PR. 

---

You're right. It is probably a better idea to only allow re-scaling indices to match some other snode. I think this re-scaling only needs to be done once per program. So we could still use trailing bits to do it. How about an API like this?

``` python
parent = # some SNode
parent.dense(ti.ijk, 8).place(blk)
parent.dynamic(ti.l, 256).place(lst)
lst.rescale_indices(blk)
```
This would essentially tell taichi to add the difference in num_bits between lst and blk to lst's trailing_bits. 

---

So I was trying to make the change to get block local storage to work with the new consecutive indexing. and I ran into another bug. see #2265

KLozes:
This PR is finally ready to be reviewed @k-ye @xumingkuan @yuanming-hu! Every test passes. 

It does not include an API for rescaling indices. I think that should be a  separate PR.

taichi/ir/analysis.h
```diff
@@ -42,7 +42,7 @@ class DiffRange {
   }
 
   bool linear_related() const {
-    return related && coeff == 1;
+    return related && coeff > 0;
```
xumingkuan, line 45: The term `linear_related()` is confusing here in that it's actually used to check whether `related && coeff == 1` in the codebase. Actually, as long as `related` is true, the relation is linear... So is there a better name for the function `related && coeff == 1`?

taichi/transforms/insert_scratch_pad.cpp
```diff
@@ -82,14 +82,17 @@ class BLSAnalysis : public BasicStmtVisitor {
...
       offsets.resize(ptr->indices.size());
+      coeffs.resize(ptr->indices.size());
       int num_indices = (int)ptr->indices.size();
       for (int i = 0; i < num_indices; i++) {
         auto diff = irpass::analysis::value_diff_loop_index(ptr->indices[i],
                                                             for_stmt, i);
         if (diff.linear_related()) {
           offsets[i].first = diff.low;
           offsets[i].second = diff.high;
+          coeffs[i] = diff.coeff;
```
xumingkuan, line 95: Yeah, we should support the `coeff != 1` case here! Maybe we need to check other usages of `linear_related()` if they can also support the `coeff != 1` case?

taichi/transforms/demote_dense_struct_fors.cpp
```diff
@@ -15,8 +15,12 @@ void convert_to_range_for(OffloadedStmt *offloaded) {
   std::vector<SNode *> snodes;
   auto *snode = offloaded->snode;
   int total_bits = 0;
+  int start0[taichi_max_num_indices] = {0};
```
xumingkuan, line 18: I'm not quite sure what `start0` means here -- maybe some other names like `start_bits`?

taichi/ir/scratch_pad.h
```diff
@@ -32,6 +32,7 @@ class ScratchPad {
   SNode *snode;
   using AccessFlag = taichi::lang::AccessFlag;
 
+  std::vector<int> coeff;
```
xumingkuan, line 35: IIUC the work of multiplying coefficients is in make_block_local.cpp. I wonder if it's better to multiply the `coeff` to `bounds` in insert_scratch_pad.cpp? (See also #2272: insert_scratch_pad.cpp actually doesn't need recursions to generate the indices because we only need `bounds` in the `ScratchPad`. I'm removing one of the two recursions in that PR.)

xumingkuan (commented):
Thanks for the great work! I have a few concerns here.

taichi/transforms/demote_dense_struct_fors.cpp
KLozes: This is actually the start bits at the root. I think I'll name it start_root, or maybe start_bits_root.

taichi/ir/scratch_pad.h
KLozes: Yes, it is actually the block_corner index (which is related to the loop index) that is multiplied by the coefficients. So I don't think we can get away with multiplying the bounds by coeff.

xumingkuan: IIUC it is the loop index that is multiplied by the coefficients, for example,
```python
for i in a:
  a[3 * i + 4] ...
  a[3 * i + 2] ...
```
For `a`, we should have `coeff = 3, low = 2, high = 5`. Suppose `i` ranges [16, 32) in a block, the BLS should store [3 * 16 + 2, 3 * 31 + 4] = [50, 98), i.e., 48 elements. I think the BlockCornerStmt should have value 50 (i.e., the lowest loop index multiplied by 3 and then added 2), and when we want to access the index `3 * i + 4`, we just compute `3 * i + 4` and subtract the BlockCornerStmt from it, to get the index in the BLS. Please correct me if I'm wrong :)

KLozes: Yep! That sounds right. You can see it here:
https://github.com/taichi-dev/taichi/blob/f1e5819bf71715a852fa9eb96d9933eb5c439bc5/taichi/transforms/make_block_local.cpp#L221-L230

taichi/ir/analysis.h
KLozes: Hmm, Do we actually need to check if coeff == 1 though? All tests pass at the moment with coeff > 0.

taichi/analysis/alias_analysis.cpp
```diff
@@ -82,16 +82,7 @@ AliasResult alias_analysis(Stmt *var1, Stmt *var2) {
       auto diff = value_diff_ptr_index(ptr1->indices[i], ptr2->indices[i]);
       if (!diff.is_diff_certain || (diff.diff_range != 0)) {
```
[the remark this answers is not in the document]
KLozes: Thanks! I'll go ahead and commit this

taichi/ir/scratch_pad.h
xumingkuan: I see! It makes sense to me now.

> Suppose `i` ranges [16, 32) in a block, ...

So BlockCornerIndexStmt should be 16 in this case, and so it's necessary for make_block_local.cpp to know `coeff`.

taichi/analysis/value_diff.cpp
```diff
@@ -18,6 +18,18 @@ DiffRange operator-(const DiffRange &a, const DiffRange &b) {
                    a.low - b.high + 1, a.high - b.low);
 }
 
+DiffRange operator*(const DiffRange &a, const DiffRange &b) {
+  return DiffRange(
+      a.related_() && b.related_(), fmax(a.low * b.coeff, a.coeff * b.low),
```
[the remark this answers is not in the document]
KLozes: Right, so would it be enough to check that a.coeff*b.coeff==0 as well? 
```
 return DiffRange(
      a.related_() && b.related_() && a.coeff*b.coeff == 0,
```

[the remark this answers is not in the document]
xumingkuan: Cool! I think this is enough. Although it doesn't cover the "zero * anything" case, that case should be already handled in the `alg_simp` pass.

taichi/ir/analysis.h
xumingkuan: I think we don't have any tests such that `related && coeff != 1` now... Maybe we should not change the logic of other places using this function in this PR.

KLozes: So the only reason I needed related && coeff > 1 was to make it into this if statement. I could make the it just check if diff.related_? But I'm not sure if thats coeff = 0 would be well behaved. Should I just make a separate function for coeff > 0? Trying to think of a a good name for it.  

https://github.com/taichi-dev/taichi/blob/df0ee9e7861abeec1322b9a7411991798fa56596/taichi/transforms/insert_scratch_pad.cpp#L90-L98

KLozes: I went ahead and reverted linear related back to related & coeff == 1. changed the if statment there to check diff.related_() && diff.coeff > 0

KLozes:
Thanks for the review @xumingkuan! I think all your suggestions have been implemented. We might want to wait for another reviewer, considering this PR is a pretty big change to Taichi

taichi/transforms/insert_scratch_pad.cpp
```diff
@@ -82,14 +82,17 @@ class BLSAnalysis : public BasicStmtVisitor {
...
       std::vector<std::pair<int, int>> offsets;
+      std::vector<int> coeffs;
       offsets.resize(ptr->indices.size());
+      coeffs.resize(ptr->indices.size());
       int num_indices = (int)ptr->indices.size();
       for (int i = 0; i < num_indices; i++) {
         auto diff = irpass::analysis::value_diff_loop_index(ptr->indices[i],
                                                             for_stmt, i);
-        if (diff.linear_related()) {
+        if (diff.related_() and diff.coeff > 0) {
```
xumingkuan, line 92: nit :)
```suggestion
        if (diff.related_() && diff.coeff > 0) {
```

xumingkuan (approved):
LGTM! We might also update the code in other repos such as https://github.com/taichi-dev/taichi_elements to make sure these codes are not broken.

taichi/codegen/codegen_llvm.cpp
```diff
@@ -1552,6 +1552,14 @@ void CodeGenLLVM::create_offload_struct_for(OffloadedStmt *stmt, bool spmd) {
...
 
     parent_coordinates = element.get_ptr("pcoord");
+    block_corner_coordinates =
+        create_entry_block_alloca(physical_coordinate_ty);
+
+    auto refine =
+        get_runtime_function(leaf_block->refine_coordinates_func_name());
+
+    create_call(refine, {parent_coordinates, block_corner_coordinates,
+                         tlctx->get_constant(0)});
```
k-ye, line 1562: Why is this `0`?

taichi/codegen/codegen_llvm.cpp
```diff
@@ -1605,13 +1613,24 @@ void CodeGenLLVM::create_offload_struct_for(OffloadedStmt *stmt, bool spmd) {
...
 
     // initialize the coordinates
-    auto refine =
-        get_runtime_function(leaf_block->refine_coordinates_func_name());
     auto new_coordinates = create_entry_block_alloca(physical_coordinate_ty);
 
     create_call(refine, {parent_coordinates, new_coordinates,
                          builder->CreateLoad(loop_index)});
 
+    if (stmt->snode->type == SNodeType::bit_array && stmt->snode->parent) {
```
k-ye, line 1621: Could you explain why is this needed?

k-ye (commented):
This is great work, thank you so much for the update! However, I still have to point out a few things:

1. We need to first add the index rescaling API, so that we don't manually compute that in the kernel. We will also need to this API to update other repos before accepting the changes here.
2. How do we handle the parent coordinates if the leaf SNode has offsets? Does it start from 0, or the rescaled offset?
3. Did we verify that the BLS still worked? I.e. the compiler can still be able to infer the BLS storage size and convert the global index <-> block index. I think one possible approach is to add some CPP unit tests around `initialize_scratch_pad()` and `make_block_local()`. Given that these two functions are so large, it will be great if the implementation of these two passes can be broken down into smaller, testable functions (possibly in a series of PRs)

taichi/codegen/codegen_llvm.cpp
KLozes: The block corner has the same coordinates as the leaf element of the first (0th) loop_index . And it is the same for any loop_index. The refine function here basically just left shifts the parent coordinates by num_bits.

taichi/codegen/codegen_llvm.cpp
KLozes: One more refine step was needed for bit_arrays to make the loop coordinates non-consecutive since the operation is going to be vectorized over multiple coordinates per loop_index. 
For example, a 1-d block of 64 leaf elements and vectorization of 32, only two loop coordinates are needed. Without this extra refine step, their coordinates would be a consecutive 0, 1 instead of non-consecutive 0, 32. 
This step was also not needed before since coordinates were always non-consecutive.

KLozes:
> This is great work, thank you so much for the update! However, I still have to point out a few things:

Thanks! I've learned a ton too!

> 1. We need to first add the index rescaling API, so that we don't manually compute that in the kernel. We will also need to check in this API to update other repos before accepting the changes here.

Alright, I'll start think about an API. 

> 2. How do we handle the parent coordinates if the leaf SNode has offsets? Does it start from 0, or the rescaled offset?

I am not sure about this, but offset tests seem to be working fine. One just has to make sure to that offsets are applied to blocks correctly on the front end when necessary. see the changes to bls_test_template.py:
https://github.com/taichi-dev/taichi/pull/2248/commits/f1e5819bf71715a852fa9eb96d9933eb5c439bc5#diff-2ea5f113062438816406fbdb85d44d0b44fde412eed818ea1c06c9584c887277R138-R154

> 3. Did we verify that the BLS still worked? I.e. the compiler can still be able to infer the BLS storage size and convert the global index <-> block index. I think one possible approach is to add some CPP unit tests around initialize_scratch_pad() and make_block_local(). Given that these two functions are so large, it will be great if the implementation of these two passes can be broken down into smaller, testable functions (possibly in a series of PRs)

I did verify that the tests in test_bls.py and test_bls_assume_in_range.py work, and they look pretty thorough. It definitely took some figuring out. See the changes to initialize_scratch_pad.cpp, make_block_local.cpp, value_diff.cpp, and bls_test_template.py.

k-ye:
> I am not sure about this, but offset tests seem to be working fine. One just has to make sure to that offsets are applied to blocks correctly on the front end when necessary. see the changes to bls_test_template.py:
f1e5819#diff-2ea5f113062438816406fbdb85d44d0b44fde412eed818ea1c06c9584c887277R138-R154

Thanks! After taking another look, I'm not entirely sure if this is correct in the first place:

https://github.com/taichi-dev/taichi/blob/939e1d3b8f4cc9ff6210aa0b7bcdcfd4af7f96f2/tests/python/bls_test_template.py#L170-L177

The question is do we really need to subtract `offset` when defining `base`?

---

@xumingkuan and I had a discussion to go over how offsets are handled in Taichi right now. Here's the summary:

For a `place` SNode with offset, the offset is useful only when we iterate it using a `struct-for`. 

1. When flattening a `GlobalPtrExpression` (.e.g. `x[i, j]`), the indices are first subtracted by `offsets`:

https://github.com/taichi-dev/taichi/blob/f34203fe2efa2f834526f14490041c1d89857407/taichi/ir/frontend_ir.cpp#L196-L207

2. Later, the loop index of the struct-for is transformed into `BinOp(+, loop_index, offset)` in `lower_ast()`:

https://github.com/taichi-dev/taichi/blob/f34203fe2efa2f834526f14490041c1d89857407/taichi/transforms/lower_ast.cpp#L302-L306

A concrete example:

```py
x = ...place(..., offset=(-16,))  # ``x`` has an offset of -16
y = ...place(...)  # ``y`` has no offset

for I in x:
  x[I]
  y[I]
```

GlobalPtrExpression::flatten --->

```py
for I in x:
  x[I - (-16)]  # ``x`` has an offset of -16, so its GlobalPtrExpression's indices are subtracted by (-16)
  y[I]  # ``y`` does not have any offset, so its GlobalPtrExpression's indices are not offseted
```

`lower_ast()` --->

```py
for (I - 16) in x:  # ``I`` still starts from 0
  x[(I - 16) - (-16)]  # i.e., x[I]
  y[I - 16]
```

So, when we are accessing a leaf SNode with offset, the accessing indices should still begin from 0. It's just that the compiler replaces the loop index `I` statement with a binary op of `I + offset`.

---

Therefore I don't think it makes sense to subtract `offset` in that test case. To further support this point, consider this simpler example:

```py
import taichi as ti

ti.init(arch=ti.cpu, print_ir=True)

y = ti.field(ti.i32)

offset = (-2,)

grid = ti.root.dense(ti.i, 4)
dyn = grid.dynamic(ti.j, 4)
dyn.place(y, offset=offset +(0,))

@ti.kernel
def foo():
  ti.append(dyn, [-2], 0)

foo()
```

The final IR is shown as below:

```
kernel {
  $0 = offloaded  
  body {
    <i32> $1 = const [0]
    <*gen> $2 = get root
    <*gen> $3 = [S0root][root]::lookup($2, $1) activate = false
    <*gen> $4 = get child [S0root->S1dense] $3
    <i32> $5 = const [2]
    <*gen> $6 = [S1dense][dense]::lookup($4, $5) activate = false
    <*gen> $7 = get child [S1dense->S2dynamic] $6
    <i32> $8 = append [S2dynamic] ptr = $7, val = $1
  }
}
```

`$5 = const [2]` is what ``[-2]` is translated to, which should have been a const of `0` (the offset is -2). Moreover, different offset here actually doesn't make a difference, and the const `2` is simply derived from `-2 & 3`:

```
    <i32> $1 = const [-2]
    <i32> $2 = const [0]
    <i32> $3 = shuffle $1[0]  # same as $1
    ...
    <i32> $8 = const [0]
    <i32> $9 = bit_sar $3 $8  # same as $1
    <i32> $10 = const [3]
    <i32> $11 = bit_and $9 $10  <-- equiv. to ($1 & $10) = (-2 & 3) = 2
```

---

> I did verify that the tests in test_bls.py and test_bls_assume_in_range.py work, and they look pretty extensive.

Unfortunately, I don't think that these tests passing would mean anything (sorry I didn't mean to blame). IIUC, if the compiler cannot auto deduct the BLS size, it would simplify not use the shared memory. So the test would suffer a slower performance, but there won't be any correctness issue. Therefore a better approach would be to test the BLS-related passes in the C++ layer. Let me think about how to refactor the code so that we can achieve that :-)

k-ye:
Also, could you break down this PR into several smaller ones, so that we can make progress much quicker? E.g.

1. rescale_index API
2. changes to deal with DiffRange *multiplication* and scratch pad initialisation
3. changes to remove `start`
4. changes to remove the trailing bits (might need further breaking down)

KLozes:
Hi @k-ye 

> $5 = const [2] is what [-2] gets translated to, which should have been a const of 0 (the offset is -2). Moreover, different offset here actually doesn't make a difference, and the const 2 is simply derived from -2 & 3

I see what you mean! The grid_offset definitely should not be required to make this work. this seems to be a bug with how offsets are handled with flattening the append function call.  

> Unfortunately, I don't think that these tests passing would mean anything (sorry I didn't mean to blame). IIUC, if the compiler cannot deduce the BLS size, it would simplify not use the shared memory. So the test would suffer a slower performance, but there won't be any correctness issue. Therefore a better approach would be to test the BLS-related passes in the C++ layer. Let me think about how to refactor the code so that we can achieve that :-)

Ahh. Thanks for picking up this task!

KLozes:
> Also, could you break down this PR into several smaller ones, so that we can make progress much quicker? E.g.
    1. rescale_index API
    2. changes to deal with DiffRange multiplication and scratch pad initialisation
    3. changes to remove start
    4. changes to remove the trailing bits (might need further breaking down)

Sure! I can try.

I think it will be best to keep 2 and 3 together. Since the diffrange multiplication was needed to make bls tests work after changing indices to consecutive. 1 and 4 also go together. I actually put trailing bits back in. Since attaching trailing bits to the "place" snode is a good way to implement the index re-scaling. 

So I can split this PR into 2. 
1. Consecutive indices for all snodes
2. API for index re-scaling

k-ye:
> I think it will be best to keep 2 and 3 together. Since the diffrange multiplication was needed to make bls tests work after changing indices to consecutive.

Hmm, I thought that "3. remove `start`" was purely a cleanup that is not strongly tied to other changes here. Maybe I missed something. Still, in general I prefer breaking down a big PR containing a series of related changes into individual ones. It should save both of us some efforts to discuss back and forth :-)

I think this PR serves as a great baseline to demonstrate our end goal. What we need is just to figure out the path to upstream it piece by piece, without breaking the existing usages or to introduce new bugs. Personally I find these `git` tips useful:

* https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
* https://stackoverflow.com/questions/1085162/commit-only-part-of-a-file-in-git  

> 1 and 4 also go together. I actually put trailing bits back in. Since attaching trailing bits to the "place" snode is a good way to implement the index re-scaling.

But we still need the following path to migrate the existing examples:

1. add the rescaling index API to taichi
2. migrate the existing code using the trailing bits feature to this new API
3. remove the trailing bits from taichi

If 1 and 3 are done in a single PR, there will be no way for us to do 2.

KLozes:
> Hmm, I thought that "3. remove start" was purely a cleanup that is not strongly tied to other changes here

Yes, I suppose actually remove the 'start' property from the extractor was just cleanup. Since I now compute 'start' on-the-fly when it is needed. So it is no longer a property of an snode. 

> Ah right! I used to think that we need to follow these steps:
> 1. add the rescaling index API to taichi
> 2. migrate the existing code using the trailing bits feature to this new API
> 3. remove the trailing bits from taichi

> But so long as the trailing bits are not removed, using the API will produce the wrong result.

So, I should be more clear here. I'm using trailing bits to implement this API on the backend. I have totally replaced the old way of computing trailing bits. Now trailing bits are attached to the place snode to account for the difference in bits between an snode and its assigned "rescaling_snode". See my new commit. commit https://github.com/taichi-dev/taichi/pull/2248/commits/abbc4b9ea607e30dd482bed02d939ab07c9b0be3
Maybe I should rename trailing_bits to rescaling_bits? 

This API doesn't quite work yet. I believe lower_access and alias analysis work. 

I still have some work to do on on demote_dense_struct_for and listgen. Currently, the StructForStmt and OffloadStmt classes do not have access to the leaf (place) snode to grab the trailing_bits from. The 'snode' member is actually the leaf_block. So I added a new 'leaf_snode' member to them. But for some reason this causes async tests to fail. Any idea why??

KLozes:
> I think this PR serves as a great baseline to demonstrate our end goal. What we need is just to figure out the path to upstream it piece by piece, without breaking the existing usages or to introduce new bugs. Personally I find these git tips useful:

Thanks!

I think I can split it into 3 fresh PRs. But they need to pulled in this order
1. diffrange multiplication
2. consecutive indices. 
3. rescaling API

k-ye:
Thanks for the update!

> I'm using trailing bits to implement this API on the backend. I have totally replaced the old way of computing trailing bits. Now trailing bits are attached to the place snode to account for the difference in bits between an snode and its assigned "rescaling_snode". 

Hmm, doesn't this work against the original goal to drop the trailing bits feature completely? In my (maybe oversimplified) thought, `rescale_index(from_snode, to_snode, index)` can simply look at the shape of the two snodes, then divide one by the other to get the rescaling coefficient. (We may need additional preconditions on this method, such as the indices used of these two SNodes must be the same, i.e., both using `ti.ijk`. They must be siblings, etc)

KLozes:
@k-ye 
Thanks for the discussion!

> Hmm, doesn't this work against the original goal to drop the trailing bits feature completely?

So yes, I want to drop _automatically_ computing trailing bits based on sibling snode shapes. Instead, trailing bits would be non-zero only when the rescaled API is called. So with this PR, the rescale_indices_to() function

``` python
a = ti.field(ti.f32)
b = ti.field(ti.f32)
c = ti.field(ti.f32)
d = ti.field(ti.f32)

blk = ti.root.pointer(ti.ij, 128)
blk.dense(ti.ij, 4).place(a)
blk.dense(ti.ij, 4).place(b)
blk.dynamic(ti.l, 128).place(c)
blk.dynamic(ti.l, 128).place(d)

c.rescale_indices_to(a)
d.rescale_indices_to(b)
```
would give the place(c) and place(d) trailing bits to rescale their ij indices from [0,1,2...,128) to [0,4,8,...512). Which would lead to the same behavior we see in Taichi before this PR. All struct-fors and accesses of c and d will use indices [0,4,8, ...512). 

My issue with the per-kernel rescaling that you proposed in another comment is that you can only rescale one field in each kernel. I think would cause problems if there are two fields that need to be rescaled, like the example above. 


> In my (maybe oversimplified) thought, rescale_index(from_snode, to_snode, index) can simply look at the shape of the two snodes, then divide one by the other to get the rescaling coefficient. 

Yep! This basically how I'm computing the trailing bits when this API is called. The trailing bits are just the difference in the total number of bits from root to leaf. The two snodes actually don't even have to be siblings or have the same ijkl indices.

yuanming-hu:
Hi @KLozes, thank you for being an awesome Taichi contributor! We really appreciate your contribution. We have a Taichi cup souvenir for every Taichi developer. Given that now we have a company around Taichi, we are finally able to ship the souvenir across the world (previously only within mainland China, unfortunately, due to COVID and lack of ￥). If you don't mind, could you share your address with yuanming@taichi.graphics and linan@taichi.graphics? We will then FedEx the cup to you :-)

(Sent the message on LinkedIn too a few days ago, you can reply there if you'd like to.)

taichicup

(Sorry about the off-topic post here!)

KLozes:
Hi @yuanming-hu 

No problem! I've learned a ton and had fun contributing to Taichi! Thanks, I'd love to have one of those souvenirs :). Sorry about missing your message on LinkedIn!

k-ye:
Thanks! Maybe it's better to list what we want to fix (https://github.com/taichi-dev/taichi/pull/2248#issuecomment-818387847):

1. Remove the trailing bits, otherwise it complicates how a SNode's shape is computed
2. Make indexing consecutive.

> would give the place(c) and place(d) trailing bits to rescale their ij indices from [0,1,2...,128) to [0,4,8,...512).

This means the field index is still non-consecutive, so taichi still has to keep the implementation to support that :-(

>  My issue with the per-kernel rescaling that you proposed in another comment is that you can only rescale one field in each kernel.

Hmm, I guess you mean that we can keep the field index rescaled in all the kernels iterating over it? We should be able to rescale the index for any number of fields inside one kernel :-)

```py
for I in a:
  Ib = ti.rescale_index(a, b, I) # Ib = I // 4
  # use `lb` to access `b`
  Ic = ti.rescale_index(a, c, I)  # Ic = I // 4
  # use `lc` to access `c`
```

Also, the fact that we had to introduce a `rescale_snode` in `SNode` itself means we're still paying the price of changing a pretty generic data structure to support a narrow use case. As mentioned before, I hope that we can agree that this complexity should be scoped into a special API, rather than making it part of the SNode data structure.

KLozes:
@k-ye 

Ahh, I was thinking we wanted an API to optionally maintain the old non-consecutive index behavior. And that can only be done with trailing bits. But if you don't think that is necessary, I can certainly implement this (hopefully) simpler API! Looks like it should only require python code changes too.

k-ye:
>  I can certainly implement this (hopefully) simpler API! Looks like it should only require python code changes too.

Sounds great, thank you!!

KLozes:
implemented in #2327

PR #2252, opened 2021-04-08
[IR] [refactor] Remove all get_kernel() and get_config() usages in IR passes

xumingkuan:
Related issue = #2193

<del>I just rewrote the `alg_simp` pass using the pass manager as an example, and I wonder if we can design the `Pass` and other classes better. WDYT?</del>

This PR removes all `Stmt::get_kernel()` and `Stmt::get_config()` usages in IR passes (there are still some usages in async-related stuff). It also unifies the interface of some IR passes to `bool foo(IRNode *root, const CompileConfig &config, const FooPass::Args &args);`.

The `AnalysisManager` together with `AnalysisResult`s are not yet used now. Maybe the interface of IR passes will be
`bool foo(IRNode *root, const CompileConfig &config, AnalysisManager *amgr, const FooPass::Args &args);`
in the future?

----


taichi/transforms/alg_simp.cpp
```diff
@@ -327,6 +328,22 @@ class AlgSimp : public BasicStmtVisitor {
...
+    bool modified;
+    explicit Result(bool modified) : modified(modified) {
+    }
+  };
+  static void run(const PassContext &ctx, IRNode *module, PassManager *mgr) {
+    auto result = AlgSimp::run(module, ctx.get_config().fast_math);
+    mgr->put_pass_result<AlgSimpPass>(std::make_unique<Result>(result));
+  }
+};
+const PassID AlgSimpPass::id = "AlgSimpPass";
```
xumingkuan, line 345: I don't like this `PassID` design since it needs writing two separate lines for each pass. What about using `std::typeindex(typeid(XPass))` directly?

taichi/transforms/alg_simp.cpp
```diff
@@ -341,7 +358,15 @@ bool use_fast_math(IRNode *root) {
...
   TI_AUTO_PROF;
-  return AlgSimp::run(root, hack::use_fast_math(root));
+  PassContext ctx;
+  if (root->get_kernel())
+    ctx.config_ = root->get_config();
+  else
+    ctx.config_.fast_math = false;
+  PassManager mgr;
+  AlgSimpPass::run(ctx, root, &mgr);
+  return mgr.get_pass_result<AlgSimpPass>().modified;
```
xumingkuan, line 368: I will rewrite this when we deprecate the API `bool alg_simp(IRNode *root)`. Will `TI_AUTO_PROF` still look good if we move it to `static void AlgSimpPass::run`?

k-ye (commented):
Thanks. I think there are a few issues:

1. `PassManager` is really `AnalysisManager`
2. `run()` is an instance method instead of a class method (otherwise there's no point inheriting from `Pass`...)
3. We need to put the pass classes in the header files. Otherwise `get_pass_result()` doesn't work across different passes.

taichi/transforms/alg_simp.cpp
xumingkuan: `TI_AUTO_PROF` uses `__FUNCTION__` as the profiled function name, so maybe we'll have a lot of `run()` in the profiler's output...

taichi/transforms/alg_simp.cpp
xumingkuan: > Could you use `static constexpr char *` as the ID type? This should allow you to write within the class?

`std::string` doesn't have a constexpr constructor, so I can't write it within the class.

> Pass ID may get used elsewhere. When that grows, this could cause us to write `std::type_index(...)` everywhere. (IIUC, your point is to avoid having an `ID` member? In general it is necessary to have an identifier for each pass) It's also less searchable than writing out the ID explicitly.

My point is to minimize the repeated part of code (or, similar part with only the class name different) that each pass (class) shares. I thought there may be an easier way to implement an id for each pass than this:
```C++
static const PassID id;
...
const PassID AlgSimpPass::id = "AlgSimpPass";
```

xumingkuan: To use `std::type_index`, I mean something like this:
```C++
class Pass {
  ...
  std::type_index get_pass_id() const {
    return std::type_index(typeid(*this));
  }
};
```

xumingkuan:
Thanks for the review! Here are my thoughts:

> 1. `PassManager` is really `AnalysisManager`

We also return something per *transform*, i.e., whether the IR is modified. Shall they also be called "analysis"?

> 2. `run()` is an instance method instead of a class method (otherwise there's no point inheriting from `Pass`...)

I wonder if there will be anything that needs to be stored in the instance, i.e., `int AlgSimpPass::xxx;`. I think if we need them, it's OK to store them in the `run()` function.

Also, when do we need to instantiate an instance and run the pass more than once? I think it makes sense if we put the pattern `while (modified) run_the_pass_again();` outside `AlgSimp`(not the pass class) in `alg_simp.cpp` and let the users decide the number of times to run.

> 3. We need to put the pass classes in the header files. Otherwise `get_pass_result()` doesn't work across different passes.

You are right... But I'm afraid the header files will become very large if we need to write tens of lines of common code per pass class.

If we store the necessary fields in the pass' instance, the header file will contain all passes' fields, which are unrelated to each other. Maybe it's fine since `statements.h` does similar work?

taichi/transforms/alg_simp.cpp
k-ye: Then just use `TI_PROFILE` with a more explicit name?

taichi/transforms/alg_simp.cpp
k-ye: I see. I think this is still a class property than an instance one. Pass ID might get used elsewhere as an ID. For example, we might have pass dependencies that are specified through the names/IDs. Using `type_index` gives no information of what the pass name will be.

k-ye:
> We also return something per transform, i.e., whether the IR is modified. Shall they also be called "analysis"?

This information applies to many passes, so we can just return whether an IR is modified as part of the interface. We have this special use case where a subportion of the passes are executed repeatedly. For other compiler infras, they probably run the entire pass sequence multiple times.

SPIR-V's passes also return a status indicating whether the IR module is modified: https://github.com/KhronosGroup/SPIRV-Tools/blob/212895d4c2cd9f4900adba4514aa5ea2c8256f9f/source/opt/pass.h#L42-L46

> I wonder if there will be anything that needs to be stored in the instance, i.e., int AlgSimpPass::xxx;. I think if we need them, it's OK to store them in the run() function.

The idea is simply to make concrete passes inherit from a common `Pass` interface. That offers us a way to build higer-level utilities to run these passes in a more flexible way. For example, right now the `compile_to_executable` hardcodes the passes. As a first step, we can keep this behavior by doing something like this:

```cpp
// inside compile_to_executable
std::vector<std::unqiue_ptr<Pass>> passes;
passes.push_back(std::make_unique<ReverseSegmentPass>());
passes.push_back(std::make_unique<OffloadPass>());
...
passes.push_back(std::make_unique<TypeCheckPass>());

for (auto& p : passes) p->run();
```

Later on, we can turn this simple `std::vector` into a more complicated data structure (i.e., the actual pass manager).

> I'm afraid the header files will become very large if we need to write tens of lines of common code per pass class.

We can have a separate header for each pass. This is standard C++ practices. Since the concerns around the header file has come up several times, my opinion is that we should follow the common practices most of the time, instead of designing solutions just for the sake of avoiding introducing new headers.

taichi/transforms/alg_simp.cpp
xumingkuan: Right, I agree that we may need the pass name somewhere. If there are no convenient macros, maybe we can only repeat the pass name in the pass ID.

taichi/transforms/alg_simp.cpp
xumingkuan: Yeah, this is doable.

taichi/analysis/clone.cpp
```diff
@@ -136,7 +136,9 @@ class IRCloner : public IRVisitor {
     if (kernel != nullptr) {
       new_root->kernel = kernel;
     }
-    irpass::type_check(new_root.get());  // probably unnecessary
+    irpass::type_check(new_root.get(),
+                       kernel ? kernel->program.config
+                              : CompileConfig());  // TODO(#2193): remove this
```
xumingkuan, line 141: Will be removed in the next PR since it doesn't make sense to use `CompileConfig()`.

taichi/transforms/offload.cpp
```diff
@@ -567,13 +572,14 @@ class FixCrossOffloadReferences : public BasicStmtVisitor {
   }
 
  private:
+  CompileConfig config;
```
k-ye, line 575: ditto?

k-ye (approved):
LGTM thanks!

PR #2220, opened 2021-03-18
[test] Migrate test_alg_simp to googletest

k-ye:
Related issue = #2221 

This PR demonstrates how we can migrate the existing CPP tests to googletest.

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

----

PR #2256, opened 2021-04-10
[test] Rename algorithm simplification test to algebraic simp. test

xumingkuan:
Related issue = #2220

IIUC "alg" means algebraic here...

----


k-ye (approved):
LGTM!

PR #2261, opened 2021-04-13
[Doc] Update install.rst, Change installer for archlinux from yaourt to yay

liberix:
yaourt is no longer developed, yay is a relatively more popular aur helper.

Related issue = #

----


k-ye (approved):
Sorry for the delay, thanks!

PR #2264, opened 2021-04-13
[ir] Move uniquely_accessed_bit_structs from compile_to_offloads to AnalysisManager

xumingkuan:
Related issue = #2193 #2252

It seems that `std::unique_ptr` cannot be `static_cast` in `get_pass_result()` so I cast the raw pointer instead.

----


k-ye (approved):
LGTM!

PR #2268, opened 2021-04-16
[metal] Add 3-stage GC Metal kernels

k-ye:
Related issue = #1174

This PR adds the GC kernels on Metal. It mostly follows the CUDA's three stage GC implementation.

----


k-ye:
Will just merge given this is metal only.

PR #2272, opened 2021-04-17
[opt] Avoid recursively generating indices twice for BLS

xumingkuan:
Related issue = #1394

`generate_block_indices()` generates indices which actually form a cube, so we can use the lower/upper bounds directly without generating the indices.

Also slightly improves value_diff.cpp in the `LoopIndexStmt` case.

----

### DIFF
--- a/taichi/analysis/value_diff.cpp
+++ b/taichi/analysis/value_diff.cpp
@@ -44,8 +44,17 @@ class ValueDiffLoopIndex : public IRVisitor {
   }
 
   void visit(LoopIndexStmt *stmt) override {
-    results[stmt->instance_id] =
-        DiffRange(stmt->loop == loop && stmt->index == loop_index, 1, 0);
+    results[stmt->instance_id] = DiffRange();
+    if (stmt->loop == loop && stmt->index == loop_index) {
+      results[stmt->instance_id] = DiffRange(true, 1, 0);
+    } else if (auto range_for = stmt->loop->cast<RangeForStmt>()) {
+      if (range_for->begin->is<ConstStmt>() &&
+          range_for->end->is<ConstStmt>()) {
+        results[stmt->instance_id] = DiffRange(
+            true, 0, range_for->begin->as<ConstStmt>()->val[0].val_int(),
+            range_for->end->as<ConstStmt>()->val[0].val_int());
+      }
+    }
   }
 
   void visit(ElementShuffleStmt *stmt) override {
--- a/taichi/analysis/value_diff.cpp
+++ b/taichi/analysis/value_diff.cpp
@@ -46,13 +46,15 @@ class ValueDiffLoopIndex : public IRVisitor {
   void visit(LoopIndexStmt *stmt) override {
     results[stmt->instance_id] = DiffRange();
     if (stmt->loop == loop && stmt->index == loop_index) {
-      results[stmt->instance_id] = DiffRange(/*related=*/true, /*coeff=*/1, /*low=*/0);
+      results[stmt->instance_id] =
+          DiffRange(/*related=*/true, /*coeff=*/1, /*low=*/0);
     } else if (auto range_for = stmt->loop->cast<RangeForStmt>()) {
       if (range_for->begin->is<ConstStmt>() &&
           range_for->end->is<ConstStmt>()) {
         results[stmt->instance_id] = DiffRange(
-            true, 0, range_for->begin->as<ConstStmt>()->val[0].val_int(),
-            range_for->end->as<ConstStmt>()->val[0].val_int());
+            /*related=*/true, /*coeff=*/0,
+            /*low=*/range_for->begin->as<ConstStmt>()->val[0].val_int(),
+            /*high=*/range_for->end->as<ConstStmt>()->val[0].val_int());
       }
     }
   }
--- a/taichi/analysis/value_diff.cpp
+++ b/taichi/analysis/value_diff.cpp
@@ -46,7 +46,7 @@ class ValueDiffLoopIndex : public IRVisitor {
   void visit(LoopIndexStmt *stmt) override {
     results[stmt->instance_id] = DiffRange();
     if (stmt->loop == loop && stmt->index == loop_index) {
-      results[stmt->instance_id] = DiffRange(true, 1, 0);
+      results[stmt->instance_id] = DiffRange(/*related=*/true, /*coeff=*/1, /*low=*/0);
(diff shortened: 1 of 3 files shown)


tests/python/test_bls.py
```diff
@@ -143,6 +143,33 @@ def copy(bls: ti.template(), w: ti.template()):
             assert w[i, j] == w2[i, j]
 
 
+@ti.require(ti.extension.bls)
+@ti.all_archs
+def test_bls_large_block():
+    n = 2**10
+    block_size = 2**5
+    stencil_length = 2**5
```
xumingkuan, line 151: Before this PR, `initialize_scratch_pad` in this test runs in $O(block_size^2 * stencil_length^2)$ time (which is 2**20, still super fast...).

After this PR, `initialize_scratch_pad` in this test runs in $O((block_size + stencil_length)^2)$ time (which is 2**12).

When I set either one of `block_size` or `stencil_length` to a value greater than `2**6`, I got cudaErrors. Is it the case that cuda doesn't support a BLS that is too large (i.e., > 2**12 f32s)?

k-ye: Yeah, but I think the BLS size is `(2x(2^6)) ^ 2 = 2^14`, since the length of each dimension is `block_size + stencil_length`. It seems that the standard shared memory limit is [48Kb per threadblock](https://www.google.com/search?q=cuda+shared+memory+max+size&rlz=1C5CHFA_enKR947KR947&oq=cuda+shared+memory+max+size&aqs=chrome..69i57.4356j0j7&sourceid=chrome&ie=UTF-8), so `2^14 * 4 = 64Kb` exceeds that.

taichi/analysis/value_diff.cpp
```diff
@@ -44,8 +44,17 @@ class ValueDiffLoopIndex : public IRVisitor {
   }
 
   void visit(LoopIndexStmt *stmt) override {
-    results[stmt->instance_id] =
-        DiffRange(stmt->loop == loop && stmt->index == loop_index, 1, 0);
+    results[stmt->instance_id] = DiffRange();
+    if (stmt->loop == loop && stmt->index == loop_index) {
+      results[stmt->instance_id] = DiffRange(true, 1, 0);
```
k-ye, line 49:
```suggestion
      results[stmt->instance_id] = DiffRange(/*related=*/true, /*coeff=*/1, /*low=*/0);
```

taichi/analysis/value_diff.cpp
```diff
@@ -44,8 +44,17 @@ class ValueDiffLoopIndex : public IRVisitor {
...
+    } else if (auto range_for = stmt->loop->cast<RangeForStmt>()) {
+      if (range_for->begin->is<ConstStmt>() &&
+          range_for->end->is<ConstStmt>()) {
+        results[stmt->instance_id] = DiffRange(
```
k-ye, line 53: ditto

taichi/transforms/insert_scratch_pad.cpp
```diff
@@ -44,27 +45,13 @@ class BLSAnalysis : public BasicStmtVisitor {
...
-  // Recursively (dimension by dimension) generate the indices in a SNode
-  // (block). E.g., a dense(ti.ij, (2, 4)) SNode has indices
-  // [(0, 0), (0, 1), (0, 2), (0, 3), (1, 0), (1, 1), (1, 2), (1, 3)]
-  void generate_block_indices(SNode *snode,
-                              BlockIndices &block_indices,
-                              std::vector<int> index,
-                              int index_id) {
+  // Generate the index bounds in a SNode (block). E.g., a dense(ti.ij, (2, 4))
+  // SNode has index bounds [[0, 1], [0, 3]].
+  static void generate_block_indices(SNode *snode, BlockIndices &indices) {
```
k-ye, line 50:
```suggestion
  static void generate_block_indices(SNode *snode, BlockIndices *indices) {
```

taichi/transforms/insert_scratch_pad.cpp
```diff
@@ -96,21 +83,20 @@ class BLSAnalysis : public BasicStmtVisitor {
...
-                     i++) {
-                  ind[depth] = bind[depth] + i;
-                  visit(ind, depth + 1);
-                }
-              };
-          visit(bind, 0);
-        }
+        const auto &index_bounds = block_indices[block];
+        std::vector<int> index(num_indices, 0);
+        std::function<void(int)> visit = [&](int depth) {
```
k-ye, line 88: How about using `dimension` instead of `depth`?

k-ye (approved):
Thanks for the cleanup LGTM!

tests/python/test_bls.py
xumingkuan: I see! I'm already using 16KB shared memory in this test. Maybe I should make it a little bit smaller to prevent using up all shared memory in the 16KB-per-block case?

k-ye: Right, I think it's fine so long as we don't exceed the 48Kb shared memory size.

taichi/analysis/value_diff.cpp
```diff
@@ -44,8 +44,19 @@ class ValueDiffLoopIndex : public IRVisitor {
...
+    results[stmt->instance_id] = DiffRange();
+    if (stmt->loop == loop && stmt->index == loop_index) {
+      results[stmt->instance_id] =
+          DiffRange(/*related=*/true, /*coeff=*/1, /*low=*/0);
+    } else if (auto range_for = stmt->loop->cast<RangeForStmt>()) {
+      if (range_for->begin->is<ConstStmt>() &&
+          range_for->end->is<ConstStmt>()) {
+        results[stmt->instance_id] = DiffRange(
+            /*related=*/true, /*coeff=*/0,
+            /*low=*/range_for->begin->as<ConstStmt>()->val[0].val_int(),
```
k-ye, line 56: BTW, this is incorrect if we iterate over a range in the decreasing order?

xumingkuan: Good catch! I'll fix it in another PR.

xumingkuan: Oh, even in decreasing order, we have `begin < end`... Nevertheless, I'll use min(begin, end - 1) to reduce confusion here.

xumingkuan: See also https://github.com/taichi-dev/taichi/pull/2276/files#diff-917d9436dcaafa0f1e41ae9bad90273a303f036f00da94e417788a7fa1dc5260R640-R643

PR #2273, opened 2021-04-18
[cuda] Add memfence to dynamic allocate and append

k-ye:
Related issue = #2179

`dynamic` uses double checked locking pattern (DCLP) in allocate and append operations. This pattern requires memfence being inserted after the load or before the store. Here are some references for correctly implementing DCLP before C++11:

* https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_C++11
* https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/
* https://www.aristeia.com/Papers/DDJ_Jul_Aug_2004_revised.pdf

I don't find a concrete DCLP example in CUDA, but the memfence is required to form a synchronize-with relationship, regardless of the programming language.

----


ljcc0930 (approved):
LGTM!

yuanming-hu (approved):
LGTM! Sorry about my very delayed review. I'm not sure adding more `grid_memfence` here may lead to a performance drop, so let's wait until @yolo2themoon add a `dynamic` SNode population test case.

feisuzhu:
Closing ancient PR

PR #2277, opened 2021-04-20
[ir] Rename is_np_array to is_external_array except the frontend

xumingkuan:
Related issue = #2193 

At the CHI level, the name `is_np_array` doesn't make sense since we don't know and don't care if the external array is from numpy. However, at the frontend, the external array is from numpy.

----


k-ye (approved):
Thank you!

PR #2279, opened 2021-04-21
[test] Add a test for range analysis of indices of reversed loops

xumingkuan:
Related issue = #2272

Reversed loops only appear in AutoDiff for now, and BLS uses `value_diff` to analyze the BLS range, so I write a test for AutoDiff + BLS.

----


k-ye (approved):
Thanks LGTM!

PR #2283, opened 2021-04-22
[ir] [refactor] Rename StackXStmt to AdStackXStmt

xumingkuan:
Related issue = #2193

"Stack" is a very generic term, so it's better to add "Ad" to denote "AutoDiff" in the statement name.

----


k-ye (approved):
lgtm

PR #2297, opened 2021-04-25
[LLVM] Add random seed support

### LINKED ISSUE #2251: Random number generator improvements

victoriacity:
Related issue = close #2251

This is the last of 3 PRs regarding #2251, which allows for specification of random seed when initializing Taichi runtime. With different random seeds, two Taichi runtime contexts on the same machine (presumably) and backend will produce different random number sequences, and vice versa. The corresponding `test_random_seed_per_program` test fails before this PR and passes after.

Due to the parallel nature of Taichi runtime, the random seed is multiplied by 2^20 to obtain the initial random state for the RNG, assuming that number of threads is well below 2^20 so that no two threads share the same random state.

As the title suggests, specification of random seed only works for CPU and CUDA backends. For OpenGL and Metal backends the random seed always varies from run to run.

----


ljcc0930 (approved):
LGTM!

Issue #2298, opened 2021-04-26
Add Vulkan backend


k-ye:
**Concisely describe the proposed feature**

This is a collaborated efforts between @g1n0st and I. We have finished the implementation locally and will upstream the code back. It is based on SPIR-V. As always, the Vulkan backend starts simple and only supports `dense`.

**Describe the solution you'd like (if any)**

**Additional comments**

bobcao3:
#1641 #992 & GL backend. Given there are multiple on-going backend effort, would it possible to build a code generator that targets the "Logical" memory model in SPIR-V so that GL, Vulkan, Metal, and DX backends can all use SPIR-V? It should be possible to directly translate CHI to SPIR-V, or translate LLVM IR to SPIR-V (the problem here is probably the memory model).

k-ye:
Thanks for the comments!

> would it possible to build a code generator that targets the "Logical" memory model in SPIR-V so that GL, Vulkan, Metal, and DX backends can all use SPIR-V?

Good point. Being able to unify the codegens is definitely a great thing to have. If it were to only support `dense`, SPIR-V is likely the solution for all. But as soon as we want to support sparse data structures (particularly `pointer`, which requires some sort of device-side memory management), the different levels of expressiveness of these backend APIs could make the  unification rather difficult.

1. As you've pointed out, memory model is a major issue here. The SPIR-V codegen needs to know how to generate instructions depending on whether the physical memory model is supported or not. 
2. Sparse data structure requires not only a codegen, but also a small runtime system to track various states. For example, https://github.com/taichi-dev/taichi/blob/master/taichi/runtime/llvm/runtime.cpp is the LLVM's sparse runtime implementation. I doubt if it is possible to re-write this in SPIR-V using only the logical memory model. But even if it is, it is not going to be very maintainable.

> It should be possible to directly translate CHI to SPIR-V, or translate LLVM IR to SPIR-V 

Precisely! I'm not too sure how mature the LLVM -> SPIR-V path is, but it's not too difficult to directly translate CHI to SPIR-V. Actually, @g1n0st has already implemented a huge amount of works for SPIR-V :-)

I think once @g1n0st 's SPIR-V codegen gets merged, we can try to get it working on other backends by supporting just `dense`.  The situation may be a bit different for Metal, though, since Metal backend does have a sparse runtime implementation as well. However, if all the SNodes are `dense`, we can see if the SPIR-V cross compiled MSL has an advantage over the existing Metal source code codegen :)

[...]

PR #2299, opened 2021-04-26
[vulkan] Add Vulkan API

k-ye:
Related issue = #2298

@g1n0st could you help take a look, thanks :-)

----


cmake/TaichiCore.cmake
```diff
@@ -16,6 +23,10 @@ if (APPLE)
         set(TI_WITH_CC OFF)
         message(WARNING "C backend not supported on OS X. Setting TI_WITH_CC to OFF.")
     endif()
+    if (TI_WITH_VULKAN)
+        set(TI_WITH_VULKAN OFF)
+        message(WARNING "Vulkan backend not supported on OS X. Setting TI_WITH_VULKAN to OFF.")
+    endif()
```
g1n0st, line 29: Why not support Vulkan backend on OS X?

k-ye: Fixed

g1n0st (approved):

Issue #2208, opened 2021-03-11
To use linalg.eig(a) inside taichi scope


yaolyu:
Hi,

I know I couldn't visit python scope in taichi scope. I need to use linalg.eig() to compute eigenvalues and eigenvectors of a stress tensor inside a taichi kernel function. Howver, it doesn't allow to do so. Using ti.avd() to compute sigma, U, V  could somehow convert output to eigenvalues and eigenvectors. But it's not accurate and not exactly what I need. Does anyone meet similar issues  that to use numpy functions inside taichi scope. Could you please share some ideas? I am really appreciated!!

Yao

k-ye:
Hi Yao,

Thanks for reporting this. As you've pointed out, it's not possible to call python code from taichi scope. One workaround will be to use taichi's external functiona call. See https://github.com/taichi-dev/taichi/blob/master/misc/demo_external_func.py. The drawback is that 1) you'll still have to implement eigen solver in C++ and compile it to a shared library, and 2) it's only available in CPU and `ti.cc` backend.

BTW, what's the shape of your matrix?

k-ye:
> maybe we could add ti.eig as a utility function, like ti.svd does?

Yeah I agree

yaolyu:
> Hi Yao,
> 
> Thanks for reporting this. As you've pointed out, it's not possible to call python code from taichi scope. One workaround will be to use taichi's external functiona call. See https://github.com/taichi-dev/taichi/blob/master/misc/demo_external_func.py. The drawback is that 1) you'll still have to implement eigen solver in C++ and compile it to a shared library, and 2) it's only available in CPU and `ti.cc` backend.
> 
> BTW, what's the shape of your matrix?

yaolyu:
To k-ye,

Thanks for asking. It is a normal 3*3 stress tensor matrix. I am looking forward to ti.eig()!!

Yao

k-ye:
Let's keep it open to track this feature :)

k-ye:
Some notes:

1. We should write a solver manually for small matrices (3x3, 2x2)
2. For generics ones, we should consider integrate with BLAS/CUBLAS

k-ye:
[...]

PR #2303, opened 2021-04-29
[Lang] Add ti.eig for 2x2 matrices

### LINKED ISSUE #2293: [Lang] Add ti.eig() as a utility function
I'm planning to implement a utility function `ti.eig()` discussed in #2208, which computes the eigenvalues and right eigenvectors for a matrix. For the convenience of users, the interface of `ti.eig()` is designed similar to `numpy.linalg.eig()`, which is showed below:
```python
def eig(A):
    """
    Parameters:
    ----------------
    A: ti.Matrix
        Matrix for which the eigenvalues and right eigenvectors will be computed

    Returns:
    -----------
    w: ti.Vector
        The eigenvalues, each repeated according to its multiplicity. The eigenvalues 
        are not necessarily ordered.

    v: ti.Matrix
        The normalized (unit “length”) eigenvectors, such that the column v[:,i] is the 
        eigenvector corresponding to the eigenvalue w[i]
    """
```

As `ti.eig()` aims to compute the eigenvalues and right eigenvectors of all general matrices, different classes of matrices should adopt different implementations considering performance. In brief, three implementations are proposed below.

1. Direct calculation for 2x2 matrices

- [ ] Implement a `@ti.func` named `eig2x2()`. The algorithm to directly calculate eigenvalues and eigenvectors of 2x2 matrices is described in https://en.wikipedia.org/wiki/Eigenvalue_algorithm#2%C3%972_matrices.

2. Direct calculation for symmetric 3x3 matrices

- [ ] Implement a `@ti.func` named `eigs3x3()`. The algorithm to directly calculate eigenvalues and eigenvectors of 3x3 matrices is described in https://en.wikipedia.org/wiki/Eigenvalue_algorithm#3%C3%973_matrices.

3. Calculation for other matrices (Need to be further discussed)

- [ ] Use the implementation in third-party libraries (eg. BLAS/cuBLAS, etc.) for asymmetric 3x3 matrices and other generic matrices.

**Additional comments**
Here I want to discuss some details about integrating with BLAS/cuBLAS to compute eigenvalues and eigenvectors of generic matrices. I would suggest using [taichi external function calls](https://github.com/taichi-dev/taichi/blob/master/misc/demo_external_func.py), for example by simply loading libcublas.so and call the function api as external functions. I am not sure whether this solution will work, so I hope I can get some advice experienced developer.

Robslhc:
Related issue = #2208, #2293 

----
This PR aims to implement utility functions `ti.eig()` for 2x2 matrices as described in #2293.

### DIFF
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -303,7 +303,7 @@ def eig(A, dt=None):
     -------
     eigenvalues: ti.Matrix(n, 2)
         The eigenvalues in complex form. Each row stores one eigenvalue. The first number
-        of the eigenvalue represents the real part and the second number represents the 
+        of the eigenvalue represents the real part and the second number represents the
         imaginary part.
     eigenvectors: ti.Matrix(n*2, n)
         The eigenvectors in complex form. Each column stores one eigenvector. Each eigenvector
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -289,6 +289,13 @@ def svd(A, dt=None):
     return svd(A, dt)
 
 
+def eig(A, dt=None):
+    if dt is None:
+        dt = impl.get_runtime().default_fp
+    from .linalg import eig
+    return eig(A, dt)
+
+
 def randn(dt=None):
     if dt is None:
         dt = impl.get_runtime().default_fp
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -290,10 +290,32 @@ def svd(A, dt=None):
 
 
 def eig(A, dt=None):
+    """Compute the eigenvalues and right eigenvectors of a real matrix.
+
+    Parameters
+    ----------
+    A: ti.Matrix(n, 2)
+        2D Matrix for which the eigenvalues and right eigenvectors will be computed.
+    dt: Optional[DataType]
+        The datatype for the eigenvalues and right eigenvectors
+
+    Returns
+    -------
+    eigenvalues: ti.Matrix(n, 2)
+        The eigenvalues in complex form. Each row stores one eigenvalue. The first number
+        of the eigenvalue represents the real part and the second number represents the 
+        imaginary part.
+    eigenvectors: ti.Matrix(n*2, n)
+        The eigenvectors in complex form. Each column stores one eigenvector. Each eigenvector
+        consists of n entries, each of which is represented by two numbers for its real part
+        and imaginary part.
+    """
     if dt is None:
         dt = impl.get_runtime().default_fp
-    from .linalg import eig
-    return eig(A, dt)
+    from taichi.lang import linalg
+    if A.n == 2:
+        return linalg.eig2x2(A, dt)
+    raise Exception("Eigen solver only supports 2D matrices.")
 
 
 def randn(dt=None):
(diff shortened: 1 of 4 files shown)


python/taichi/lang/linalg.py
```diff
@@ -81,6 +81,46 @@ def svd3d(A, dt, iters=None):
...
+        A1r = A - lambda1[0] * ti.Matrix.identity(dt, 2)
+        A1i = -lambda1[1] * ti.Matrix.identity(dt, 2)
+        A2r = A - lambda2[0] * ti.Matrix.identity(dt, 2)
+        A2i = -lambda2[1] * ti.Matrix.identity(dt, 2)
+        v1 = ti.Vector([A2r[0, 0], A2i[0, 0], A2r[1, 0],
+                        A2i[1, 0]]).cast(dt).normalized()
+        v2 = ti.Vector([A1r[0, 0], A1i[0, 0], A1r[1, 0],
+                        A1i[1, 0]]).cast(dt).normalized()
+    eigenvalues = ti.Matrix.rows([lambda1, lambda2])
+    eigenvectors = ti.Matrix.cols([v1, v2])
```
Robslhc, line 119: I think there is one thing that needs to be discussed. From my knowledge, taichi doesn't have such a datatype to represent complex numbers, which could be generated by eigensolvers. 

Currently, I use 2 numbers to store the real part and imaginary part of a complex number. That is to say, for a 2x2 matrix, function `ti.eig()` returns a 2x2 matrix v and a 4x2 matrix w. Each row of matrix v stores one eigenvalue represented by a 2d vector for its real and imaginary part. Each column of w stores one eigenvector with 2 entries each represented by two numbers for its real and imaginary part. I'm not sure whether this is the best way to handle this problem. Is there any advice on this? @k-ye

python/taichi/lang/__init__.py
```diff
@@ -289,6 +289,13 @@ def svd(A, dt=None):
     return svd(A, dt)
 
 
+def eig(A, dt=None):
+    if dt is None:
+        dt = impl.get_runtime().default_fp
+    from .linalg import eig
+    return eig(A, dt)
```
k-ye, line 296: While this style is probably following how `svd()` is implemented, it could be a bit confusing to have multiple `eig()` symbols. I'd suggest to change to this:

```py
def eig(A, dt=None):
    if dt is None:
        dt = impl.get_runtime().default_fp
    from taichi.lang import linalg
    if A.n == 2:
      return linalg.eig2x2(A, dt)
    raise Exception("Eigen solver only supports 2D matrices.")
```

python/taichi/lang/linalg.py
k-ye: I see. Yeah this sounds good, thanks! Could you write this as the docstring?

k-ye (approved):
Thank you! LGTM + nits

python/taichi/lang/__init__.py
k-ye: Actually, I think it's better to split into two PRs. This makes the review process easier ;-)

Robslhc: Sure. I'll fix it.

Robslhc:
The CI of GPU failed but I have no idea about the error. It seems the error is not related to my changes, is it?

k-ye:
No, it's an issue on the CI bot itself #2307 ..

k-ye (approved):

PR #2305, opened 2021-04-30
[misc] Set up buildbot for gpu tests on linux

BillXu2000:

----

An self-hosted machine is set up for build and run gpu tests on linux.


k-ye (approved):
LGTM, thx!

PR #2306, opened 2021-04-30
[IR] Experimental real function support (Stage 1)

xumingkuan:
Related issue = #2193

This PR provides a new compile option `experimental_real_function`, which delays inlining all functions at the frontend (when generating the C++ AST) to after `lower_ast` and `simplify`. The PR itself accelerates the Python part, and provides the infrastructure to support real functions in the future.

For now, only **scalar/template** arguments and **scalar** return values are supported.

`return` is handled by storing the return values into a local variable, and jumping to the end of the function is not implemented. Therefore, the original behavior (only allowing one `return` statement at the end of the function) is supported, and the case of multiple return statements with each one at the end of the control flow (e.g. 
```python3
@ti.func
def foo():
  ...
  if ...:
    ...
    return ...
  else:
    ...
    return ...
```
) is also supported. However, other cases of multiple return statements are not correctly handled.

The compiled functions are cached in `Func.compiled` in Python, and autodiff is not yet supported. The `FunctionKey` (in C++) consists of the function's unique id and the instance's (which corresponds to template instantiations) id, and the `FunctionKey::func_name` is used only for debugging.

`FrontendExprStmt` is added to the frontend IR (AST). This handles a statement with only a single expression, which is a common case when calling a function with no return values.

Example:
```python
import taichi as ti

ti.init(print_ir=True, print_preprocessed=True, experimental_real_function=True)

@ti.func
def foo(x: ti.i32) -> ti.i32:
    print('foo called!')
    return x

@ti.kernel
def run():
    a = foo(1)
    foo(2)
    print(a)


if __name__ == '__main__':
    run()
```
Output:
```
...
def run():
    import taichi as ti
    a = ti.expr_init(foo(1))
    ti.func_call_with_check(foo, 2)
    ti.ti_print(a)
...
def foo():
    import taichi as ti
    ti.lang.kernel_arguments.decl_scalar_ret(ti.i32)
    x = ti.lang.kernel_arguments.decl_scalar_arg(ti.i32)
    ti.ti_print('foo called!')
    ti.core.create_kernel_return(ti.cast(ti.Expr(x), ti.i32).ptr)
...
[I 05/06/21 16:43:51.884] [compile_to_offloads.cpp:operator()@23] [foo_13_0] Simplified:
kernel {
  print "foo called!\n"
  <i32> $1 = arg[0]
  <i32> $2 : kernel return $1
}
...
kernel {
  $0 = offloaded  
  body {
    <i32> $1 = const [1]
    print "foo called!\n"
    print "foo called!\n"
    print $1, "\n"
  }
}
[T 05/06/21 16:38:32.320] [codegen_llvm.cpp:operator()@1920] Launching kernel run_c4_0_kernel
foo called!
foo called!
1
```

----


taichi/program/function.h
```diff
@@ -0,0 +1,35 @@
...
+#include "taichi/lang_util.h"
+#include "taichi/ir/ir.h"
+#include "taichi/program/kernel.h"
+
+namespace taichi {
+namespace lang {
+
+class Program;
+
+// TODO: Let Function and Kernel inherit from some class like "Callable"
```
k-ye, line 12: Sounds great!

taichi/transforms/inlining.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "taichi/ir/ir.h"
+#include "taichi/ir/statements.h"
+#include "taichi/ir/transforms.h"
+#include "taichi/ir/visitors.h"
+
+namespace taichi {
+namespace lang {
+
+// Inline all functions.
+class Inlining : public BasicStmtVisitor {
```
k-ye, line 12: super nit: `Inliner`?

taichi/transforms/inlining.cpp
```diff
@@ -0,0 +1,96 @@
...
+ public:
+  using BasicStmtVisitor::visit;
+  DelayedIRModifier modifier;
```
k-ye, line 15: Sorry we haven't enforced this before, but let's keep the fields private as much as we can.

taichi/transforms/statement_replace.cpp
```diff
@@ -5,16 +5,81 @@
 
 TLANG_NAMESPACE_BEGIN
 
+// TODO: rewrite and simplify these classes
+// Replace all usages and remove the statements themselves
+class StatementReplaceAndRemove : public IRVisitor {
```
k-ye, line 10: Can we unify this class and `StatementReplace`?

tests/python/test_function.py
```diff
@@ -0,0 +1,111 @@
+import taichi as ti
+
+
+@ti.test(experimental_real_function=True)
+def test_function_without_return():
```
k-ye, line 5: These tests are great to have! However, the problem with them is that they don't fully cover what we want to test. That is, even if  `experimental_real_function=False`, it's going to pass anyway.

We can add more tests on the C++ side and verify at the CHI IR level. Here's how it may look like:

1. Construct a CHI IR tree containing 'FuncCallStmt'
2. Run all the passes, including the inlining one.
3. Check that the IR is processed correctly (e.g. no more `FuncCallStmt`, types are correct, etc.). 
4. * Bonus point, feed this transformed IR into an actual program and see if it works as expected.

tests/python/test_function.py
```diff
@@ -0,0 +1,111 @@
...
+            return val * 2
+        else:
+            x[None] += val
+            return val
+
+    @ti.kernel
+    def run():
+        a = foo(40)
+        foo(1)
+        if ti.cfg.arch != ti.opengl:
```
k-ye, line 59: Use `ti.test(..., exclude=[ti.opengl])`

k-ye (commented):
Thank you very much for the great progress! It seems that some design are still debatable, and we should try split this into smaller pieces. But I think we are on the right track :-)

A problem uncovered by this PR is that we should figure out how to make frontend expressions as part of the statements. Function call (being on the RHS or not) is an expression, and  many languages treat expressions as statements as well (E.g. in C/C++, an expression + `;` is a statement).

taichi/transforms/statement_replace.cpp
xumingkuan: Maybe I'll rewrite it in the next PR.

python/taichi/lang/impl.py
```diff
@@ -170,6 +172,13 @@ def chain_compare(comparators, ops):
     return ret
 
 
+@taichi_scope
+def func_call_rvalue(func, *args):
+    assert get_runtime().experimental_real_function
+    args = make_expr_group(args)
+    return ti.Expr(_ti_core.make_func_call_expr(func.__name__, args))
```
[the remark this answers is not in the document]
xumingkuan: fixed

python/taichi/lang/impl.py
```diff
@@ -186,7 +195,13 @@ def func_call_with_check(func, *args, **kwargs):
             UserWarning,
             stacklevel=2)
 
-    return func(*args, **kwargs)
+    if get_runtime().experimental_real_function:
```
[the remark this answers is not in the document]
xumingkuan: fixed (probably)

python/taichi/lang/transformer.py
```diff
@@ -922,3 +929,13 @@ def visit_Return(self, node):
                 ' Consider using a local variable to walk around.')
 
         return node
+
+
+class HackFunctionCall(ASTTransformerBase):
```
k-ye, line 934: Give this a more meaningful name, e.g. `TransformFunctionCallAsStmt`? Also explain that the purpose of this pass is to transform a standalone taichi function call expression into a statement?

python/taichi/lang/transformer.py
```diff
@@ -922,3 +929,13 @@ def visit_Return(self, node):
...
+    def __init__(self, func):
+        super().__init__(func)
+
+    def visit_Call(self, node):
+        node.args = [node.func] + node.args
+        node.func = self.parse_expr('ti.func_call_with_check')
```
k-ye, line 940: Consider  renaming this method, e.g. `maybe_transform_ti_func_call_to_stmt`?

taichi/ir/statements.h
```diff
@@ -751,13 +753,12 @@ class FuncBodyStmt : public Stmt {
  */
 class FuncCallStmt : public Stmt {
  public:
-  std::string funcid;
+  FunctionKey funcid;
```
k-ye, line 756: nit: `func_key`

taichi/ir/transforms.h
```diff
@@ -75,9 +75,13 @@ bool constant_fold(IRNode *root,
                    const CompileConfig &config,
                    const ConstantFoldPass::Args &args);
 void offload(IRNode *root, const CompileConfig &config);
-void replace_statements_with(IRNode *root,
+void replace_statements_with(
+    IRNode *root,
+    std::function<bool(Stmt *)> filter,
+    std::function<std::unique_ptr<Stmt>(Stmt *)> generator);
+bool replace_statements_with(IRNode *root,
```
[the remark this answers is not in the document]
k-ye: Hmm, why one of them returns `bool`?

taichi/program/function_key.h
```diff
@@ -0,0 +1,32 @@
...
+#include <functional>
+#include <string>
+
+namespace taichi {
+namespace lang {
+
+class FunctionKey {
+ public:
+  std::string func_name;
+  int func_id;
```
k-ye, line 12: Explain that:

1. `func_id` maps to a `@ti.func`
2. `instance_id` is because a `@ti.func` may be instantiated multiple times
3. `func_name` is mostly for debugging/visualizing purpose, and doesn't need to participate in the hash computation?

python/taichi/lang/kernel_impl.py
```diff
@@ -67,19 +68,30 @@ def decorated(*args):
...
-        self.arguments = []
+        self.argument_annotations = []
         self.argument_names = []
         _taichi_skip_traceback = 1
         self.extract_arguments()
+        self.template_slot_locations = []
+        for i in range(len(self.argument_annotations)):
+            if isinstance(self.argument_annotations[i], template):
+                self.template_slot_locations.append(i)
+        self.mapper = KernelTemplateMapper(self.argument_annotations,
```
k-ye, line 93: TODO: rename this to `TaichiCallableTemplateMapper` (in a next PR?)

k-ye (commented):
Thanks for the cleanup! I'll go over it more carefully tmr. A few directions that we can consider:

1. Support passing in a `field` (which will be a counted as a template)
2. Instead of having an `experimental_real_function` flag, we should have a method that "tries to see if the function inlining can be delayed", and if so, delay it.

taichi/ir/transforms.h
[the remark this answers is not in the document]
xumingkuan: Well... both can return `bool` actually. I just didn't modify the original one.

xumingkuan:
> Thanks for the cleanup! I'll go over it more carefully tmr. A few directions that we can consider:
> 
>     1. Support passing in a `field` (which will be a counted as a template)

Actually this is already supported (we need `ti.template()` annotation for this).

>     2. Instead of having an `experimental_real_function` flag, we should have a method that "tries to see if the function inlining can be delayed", and if so, delay it.

For now only autodiff is not yet supported, and I think we can support it soon. Maybe we can delay all function inlinings in the future.

python/taichi/lang/kernel_impl.py
```diff
@@ -90,12 +102,32 @@ def __call__(self, *args):
...
+                self.do_compile(key=key, args=args)
+            return self.func_call_rvalue(key=key, args=args)
+        else:
+            if self.compiled is None:
+                self.do_compile(key=None, args=args)
+            ret = self.compiled(*args)
+            return ret
+
+    def func_call_rvalue(self, key, args):
+        assert impl.get_runtime().experimental_real_function
```
xumingkuan, line 122:
```suggestion
        # Skip the template args, e.g., |self|
        assert impl.get_runtime().experimental_real_function
```

python/taichi/lang/transformer.py
```diff
@@ -49,8 +50,10 @@ def print_ast(tree, title=None):
     def visit(self, tree):
         self.print_ast(tree, 'Initial AST')
         self.pass_Preprocess.visit(tree)
+        ast.fix_missing_locations(tree)
         self.print_ast(tree, 'Preprocessed')
         self.pass_Checks.visit(tree)
+        self.pass_hack_function_call.visit(tree)
         self.print_ast(tree, 'Checked')
```
xumingkuan, line 57:
```suggestion
        self.print_ast(tree, 'Checked')
        self.pass_transform_function_call.visit(tree)
```

python/taichi/lang/transformer.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, func=None, *args, **kwargs):
                                                         *args,
                                                         **kwargs)
         self.pass_Checks = ASTTransformerChecks(func=func)
+        self.pass_hack_function_call = TransformFunctionCallAsStmt(func=func)
```
xumingkuan, line 39:
```suggestion
        self.pass_transform_function_call = TransformFunctionCallAsStmt(func=func)
```

python/taichi/lang/transformer.py
```diff
@@ -705,27 +704,41 @@ def visit_FunctionDef(self, node):
...
-                # so that they are passed by value.
-                arg_init = self.parse_stmt('x = ti.expr_init_func(0)')
-                arg_init.targets[0].id = arg.arg
-                self.create_variable(arg.arg)
-                arg_init.value.args[0] = self.parse_expr(arg.arg +
-                                                         '_by_value__')
-                args.args[i].arg += '_by_value__'
-                arg_decls.append(arg_init)
+            if impl.get_runtime().experimental_real_function:
+                if self.is_kernel and self.func.is_grad:
```
xumingkuan, line 720: self.is_kernel always false?

taichi/ir/frontend_ir.cpp
```diff
@@ -363,6 +363,16 @@ void ExternalTensorShapeAlongAxisExpression::flatten(FlattenContext *ctx) {
   stmt = ctx->back_stmt();
 }
 
+void FuncCallExpression::flatten(FlattenContext *ctx) {
+  std::vector<Stmt *> stmt_args;
+  for (auto &arg : args.exprs) {
+    arg->flatten(ctx);
+    stmt_args.push_back(arg->stmt);
+  }
+  ctx->push_back<FuncCallStmt>(funcid, stmt_args);
```
xumingkuan, line 372: funcid->func_key

python/taichi/lang/kernel_impl.py
```diff
@@ -363,8 +413,10 @@ def taichi_ast_generator():
                     "Kernels cannot call other kernels. I.e., nested kernels are not allowed. Please check if you have direct/indirect invocation of kernels within kernels. Note that some methods provided by the Taichi standard library may invoke kernels, and please move their invocations to Python-scope."
                 )
             self.runtime.inside_kernel = True
+            self.runtime.current_kernel_func = self.func
```
xumingkuan, line 416:
```suggestion
            self.runtime.current_kernel = self
```

python/taichi/lang/kernel_impl.py
```diff
@@ -363,8 +413,10 @@ def taichi_ast_generator():
...
             compiled()
             self.runtime.inside_kernel = False
+            self.runtime.current_kernel_func = None
```
xumingkuan, line 419:
```suggestion
            self.runtime.current_kernel = None
```

python/taichi/lang/impl.py
```diff
@@ -199,8 +211,10 @@ def __init__(self, kernels=None):
         self.compiled_grad_functions = {}
         self.scope_stack = []
         self.inside_kernel = False
+        self.current_kernel_func = None
```
xumingkuan, line 214:
```suggestion
        self.current_kernel = None
```

k-ye (approved):
Great work, LGTM!

A few TODOs:

* [ ] split transformer for expr and stmt
* [ ] In python `Func`, ideally we should have different fields for `compiled`. Right now it could be either a `dict[int, function]` or a python function. As a fix, we can have a small tuple to store both the python transformed function and C++ `Function`, which will be the values of `taichi_functions`.
* [ ] Have a `Callable` class for `Function` and `Kernel`.
* [ ] `CurrentFunctionGuard` and `CurrentKernelGuard` can also be unified.
* [ ] Stop `using namespace llvm` in several places.
* [ ] (Ambitious) get rid of the GLOBAL `FrontendContext`.

There is yet another problem. A `@ti.kernel` could call into a *python function*, which could then invoke a `@ti.func`. In that case, the `ti.func` cannot be properly transformed into a statement.

tests/python/test_function.py
```diff
@@ -0,0 +1,186 @@
...
+        y[None] = 20
+        kernel_inc(x)
+        assert x[None] == 11
+        assert y[None] == 20
+        kernel_inc(y)
+        assert x[None] == 11
+        assert y[None] == 21
+
+    @ti.func
+    def inc(x: ti.template()):
```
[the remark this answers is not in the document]
xumingkuan: Yes!

xumingkuan:
> There is yet another problem. A `@ti.kernel` could call into a _python function_, which could then invoke a `@ti.func`. In that case, the `ti.func` cannot be properly transformed into a statement.

Nice catch...

I wonder if there are cases that users need to call Python functions from Taichi kernels? If not, maybe we can just disable calling Python functions from Taichi kernels (and just throw a TaichiSyntaxError).

k-ye:
> I wonder if there are cases that users need to call Python functions from Taichi kernels? If not, maybe we can just disable calling Python functions from Taichi kernels (and just throw a TaichiSyntaxError).

Unfortunately there are a few such cases. For example: https://github.com/taichi-dev/taichi/blob/a07b5a5b20a3c2687025a167a422655c4a3b3724/python/taichi/lang/__init__.py#L325-L348

A Taichi kernel could call `sym_eig()`, which is itself a python function. `sym_eig()` calls `linalg.sym2x2`, which is a taichi function.

xumingkuan:
> > I wonder if there are cases that users need to call Python functions from Taichi kernels? If not, maybe we can just disable calling Python functions from Taichi kernels (and just throw a TaichiSyntaxError).
> 
> Unfortunately there are a few such cases. For example:
> 
> https://github.com/taichi-dev/taichi/blob/a07b5a5b20a3c2687025a167a422655c4a3b3724/python/taichi/lang/__init__.py#L325-L348
> 
> A Taichi kernel could call `sym_eig()`, which is itself a python function. `sym_eig()` calls `linalg.sym2x2`, which is a taichi function.

IIUC, `sym_eig()` can be rewritten as a Taichi function. And Python functions that cannot be rewritten as a Taichi function sounds like undefined behavior itself when called by Taichi kernels?

k-ye:
> sym_eig() can be rewritten as a Taichi function.

Almost. But `dt = impl.get_runtime().default_fp` needs to be evaluated in the Python context. And `sym_eig()` is one example. The larger problem here is that we have to always be aware of whether the code is executed in Python or will become part of CHI. The boundary here is really blurry..

> And Python functions that cannot be rewritten as a Taichi function sounds like undefined behavior itself when called by Taichi kernels?

Right. I think the TorchScript (TS) design will help achieve this. But TS itself may bring other problems. For example, we may have to move a large portion of the Python code to CPP...

PR #2308, opened 2021-04-30
[lang] Rescale_indix api and test

KLozes:
Related issue = #2177 
A simple index rescaling API that could be useful once we switch to consecutive indices.  #2248 

----

### DIFF
--- a/python/taichi/lang/ops.py
+++ b/python/taichi/lang/ops.py
@@ -550,6 +550,23 @@ def length(l, indices):
 
 
 def rescale_index(a, b, I):
+    """Rescales the index 'I' of field 'a' the match the shape of field 'b'
+
+    Parameters
+    ----------
+    a: ti.field(), ti.Vector.field, ti.Matrix.field()
+        input taichi field
+    b: ti.field(), ti.Vector.field, ti.Matrix.field()
+        output taichi field
+    I: ti.Vector()
+        grouped loop index
+
+    Returns
+    -------
+    Ib: ti.Vector()
+        rescaled grouped loop index
+
+    """
     assert isinstance(a, Expr) and a.is_global(), \
             f"first arguement must be a field"
     assert isinstance(b, Expr) and b.is_global(), \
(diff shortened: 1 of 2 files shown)


python/taichi/lang/ops.py
```diff
@@ -546,3 +547,24 @@ def deactivate(l, indices):
...
+            f"first arguement must be a field"
+    assert isinstance(b, Expr) and b.is_global(), \
+            f"second arguement must be a field"
+    assert isinstance(I, matrix.Matrix) and not I.is_global(), \
+            f"third arguement must be a grouped index"
+    assert len(a.shape) >= len(b.shape)
+    Ib = I.copy()
+    for n in range(len(b.shape)):
+        if a.shape[n] > b.shape[n]:
+            assert (a.shape[n] // b.shape[n]) % 2 == 0, \
```
k-ye, line 563: Do we have to enforce `% 2 == 0`?

python/taichi/lang/ops.py
```diff
@@ -546,3 +547,24 @@ def deactivate(l, indices):
 
 def length(l, indices):
     return Expr(_ti_core.insert_len(l.snode.ptr, make_expr_group(indices)))
+
+
+def rescale_index(a, b, I):
```
k-ye, line 552: nit: Could you add a docstring like https://github.com/taichi-dev/taichi/blob/abd063c92fa24616118c9feddcc13efc4b1c75fb/python/taichi/lang/__init__.py#L293-L312? Thanks!

k-ye (approved):
LGTM, thanks!

---

It seems that the first commit is not properly associated with your github account? Could you fix that issue so that CLA bot stops complaining?

python/taichi/lang/ops.py
KLozes: Good question! Probably no.

python/taichi/lang/ops.py
KLozes: Sure thing.

KLozes:
> It seems that the first commit is not properly associated with your github account? Could you fix that issue so that CLA bot stops complaining?

Sorry about this! I'm not really sure there's a good way to fix this though. I could just create a whole new PR if you would like?

PR #2313, opened 2021-05-03
[lang] Add rescale_index() and test

KLozes:
Related issue = #2177

Same as the previous one, but CLA should work!

A simple index rescaling API that could be useful once we switch to consecutive indices. #2248


k-ye (approved):
Thanks for re-submitting this :-)!

PR #2314, opened 2021-05-03
[doc] Wrap `for` to avoid confusion

Ravenclaw-OIer:
Wrapped "for" (as in for loop) in double ` marks so they render to monospace font. It may be mistaken for the *English word* "for" (as used grammatically). 

I think "range-for loops" can be left as is because they won't lead to confusion.

Related issue = N/A

----


k-ye:
Sorry, could you reopen one and request to merge it into [`taichi-dev:master`](https://github.com/taichi-dev/taichi/tree/master) please, thanks!

Issue #2317, opened 2021-05-03
[LLVM] compile_runtime_bitcode does not work when file path includes space


bobcao3:
**Describe the bug**
When `void compile_runtime_bitcode(Arch arch)` is generating commands to run clang or llvm-as, if the path contains a space, the path will be considered as multiple arguments to the programs. 

**To Reproduce**
Use a taichi source installation from a folder containing spaces (e.g. `C:\Users\Cheng Cao\source\repos\taichi\`) and run any example.

**Additional comments**
Fix: https://github.com/bobcao3/taichi/commit/afdefbe76c7764cbc58f28700d1c9d23c4b5cfce

bobcao3:
PR: https://github.com/taichi-dev/taichi/pull/2318

k-ye:
Fixed by #2318

PR #2318, opened 2021-05-03
[bug] Protect against path with space

bobcao3:
Protect against path with space.

e.g. "C:\Users\Cheng Cao\source\repos\taichi" will be interpreted as "C:\Users\Cheng" and "Cao\source\repos\taichi", and causing the runtime bitcode compilation to fail (llvm_context.cpp line 152 `void compile_runtime_bitcode(Arch arch)`)

Related issue = #2317


yuanming-hu (approved):
Thank you, LGTM! Welcome to the Taichi community. Let's see if all the tests pass on all platforms :-)

PR #2319, opened 2021-05-03
[doc] Windows uses backslash for filepath

bobcao3:
When setting up Windows Path related environment variables, it may reject forward slash as invalid file path.


k-ye (approved):
LGTM thanks!

---

Could you re-target this PR to merge into the `master` branch? Also could you rebase your commit to the head please (I'm afraid this is causing Code Format CI failure)?

Issue #2316, opened 2021-05-03
[GUI] Support HiDPI / toggle native resolution


bobcao3:
**Concisely describe the proposed feature**
Support HiDPI in the GUI or have a toggle to enable native resolution (client side scaling) so the output can be pixel-perfect on fractional scaling displays.

**Describe the solution you'd like (if any)**
This is relatively simple to implement on Windows and probably non-problem on Linux due to its quirks on X11 scaling & DPI protocols. (todo: attach a commit here) However it may be bad to just use client side scaling and ignoring the DPI, therefore some API changes will be needed to enable DPI aware GUI.

**Additional comments**
Reference the design of GLFW, SDL, and ImGUI

bobcao3:
Just checked the behavior of the OpenGL backend, it is using native resolution but it does not seem like the GUI is using the DPI information. Therefore having the WIN32 backend to be pixel perfect should be fine for now.

PR #2320, opened 2021-05-03
[GUI] Use native resolution (client scaling) on WIN32

bobcao3:
Related issue = #2316

### DIFF
--- a/taichi/gui/win32.cpp
+++ b/taichi/gui/win32.cpp
@@ -155,6 +155,8 @@ void GUI::process_event() {
 void GUI::create_window() {
   auto CLASS_NAME = L"Taichi Win32 Window";
 
+  SetThreadDpiAwarenessContext(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE);
+
   WNDCLASS wc = {};
 
   wc.lpfnWndProc = WindowProc;


bobcao3:
Two of the commits are included accidentally. The actual change is the one-liner in commit 23efb42. This PR probably needs some sanitation.

k-ye:
Thank you for improving the GUI! As you've pointed out, this PR has a large overlap with #2315, could you separate the two pls :-) ?

k-ye (approved):
LGTM, thanks!

k-ye:
Some user reported that Taichi no longer worked on their Win7 since https://github.com/taichi-dev/taichi/releases/tag/v0.7.19. 

```
ImportError: DLL load failed while importing taichi_core
```

I took a look and this PR seems to be the most suspicious one. Could you help take a look whether this change has introduced  some new DLL dependency? Thanks!

bobcao3:
I just checked and it seems that this DPI awareness mode only exists post windows 8.1. This might need some detection scheme, but I'm not sure whether this can be a runtime variant or it must be compile time.

bobcao3:
Ref https://docs.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows

k-ye:
Thanks. Can we use https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setprocessdpiaware?redirectedfrom=MSDN for Win7 compatibility?

bobcao3:
#2450

PR #2322, opened 2021-05-05
[doc] REOPEN #2314: wrap 'for' to avoid confusion

Ravenclaw-OIer:
@k-ye  **This is a reopen of #2314, merging to master**

Wrapped "for" (as in for loop) in double ` marks so they render to monospace font. It may be mistaken for the *English word* "for" (as used grammatically). 

I think "range-for loops" can be left as is because they won't lead to confusion.

Related issue = #

----


k-ye (approved):
Thank you! LGTM

PR #2323, opened 2021-05-05
[perf] Loop-invariant code motion

bobcao3:
Performance Observations:
Complicated. Check the related issue.

Structural Observations:
1. Many programs contain loops and certainly graphis programs contain a lot of them. It is very often that the loop is going over the same data structure and the same computations, just with different data
2. If the operands of a statement that has no side effects are all not from the current scope, it can only be from all the previous scopes. (If we assume the IR is well formed, and we have access to those operands)
3. Moving a statement in a loop, that does not depend on anything within the loop, one layer up, will result in less redundant computations. The more iterations a loop has, the more redundancy is removed.
4. Doing so may or may not increase register pressure. In the case of all operands are still in use after the moved statement, we have increased register pressure in the loop up until  the moved statement. If the operands are not used later, we have in fact decreased or maintained the same register pressure.

Related issue = #2324

### DIFF
--- a/taichi/python/export_lang.cpp
+++ b/taichi/python/export_lang.cpp
@@ -95,8 +95,9 @@ void export_lang(py::module &m) {
       .def(py::self == py::self)
       .def("__hash__", &DataType::hash)
       .def("to_string", &DataType::to_string)
-      .def("get_ptr", [](DataType *dtype) -> Type * { return *dtype; },
-           py::return_value_policy::reference)
+      .def(
+          "get_ptr", [](DataType *dtype) -> Type * { return *dtype; },
+          py::return_value_policy::reference)
       .def(py::pickle(
           [](const DataType &dt) {
             // Note: this only works for primitive types, which is fine for now.
@@ -139,6 +140,8 @@ void export_lang(py::module &m) {
       .def_readwrite("simplify_after_lower_access",
                      &CompileConfig::simplify_after_lower_access)
       .def_readwrite("lower_access", &CompileConfig::lower_access)
+      .def_readwrite("move_loop_invariant_outside_if",
+                     &CompileConfig::move_loop_invariant_outside_if)
       .def_readwrite("default_cpu_block_dim",
                      &CompileConfig::default_cpu_block_dim)
       .def_readwrite("default_gpu_block_dim",
@@ -195,9 +198,10 @@ void export_lang(py::module &m) {
   m.def("reset_default_compile_config",
         [&]() { default_compile_config = CompileConfig(); });
 
-  m.def("default_compile_config",
-        [&]() -> CompileConfig & { return default_compile_config; },
-        py::return_value_policy::reference);
+  m.def(
+      "default_compile_config",
+      [&]() -> CompileConfig & { return default_compile_config; },
+      py::return_value_policy::reference);
 
   py::class_<Program>(m, "Program")
       .def(py::init<>())
@@ -214,11 +218,12 @@ void export_lang(py::module &m) {
            })
       .def("print_memory_profiler_info", &Program::print_memory_profiler_info)
       .def("finalize", &Program::finalize)
-      .def("get_root",
-           [&](Program *program) -> SNode * {
-             return program->snode_root.get();
-           },
-           py::return_value_policy::reference)
+      .def(
+          "get_root",
+          [&](Program *program) -> SNode * {
+            return program->snode_root.get();
+          },
+          py::return_value_policy::reference)
       .def("get_total_compilation_time", &Program::get_total_compilation_time)
       .def("print_snode_tree", &Program::print_snode_tree)
       .def("get_snode_num_dynamically_allocated",
@@ -233,9 +238,10 @@ void export_lang(py::module &m) {
   m.def("get_current_program", get_current_program,
         py::return_value_policy::reference);
 
-  m.def("current_compile_config",
-        [&]() -> CompileConfig & { return get_current_program().config; },
-        py::return_value_policy::reference);
+  m.def(
+      "current_compile_config",
+      [&]() -> CompileConfig & { return get_current_program().config; },
+      py::return_value_policy::reference);
 
   py::class_<Index>(m, "Index").def(py::init<int>());
   py::class_<SNode>(m, "SNode")
@@ -270,9 +276,10 @@ void export_lang(py::module &m) {
       .def("data_type", [](SNode *snode) { return snode->dt; })
       .def("get_num_ch",
            [](SNode *snode) -> int { return (int)snode->ch.size(); })
-      .def("get_ch",
-           [](SNode *snode, int i) -> SNode * { return snode->ch[i].get(); },
-           py::return_value_policy::reference)
+      .def(
+          "get_ch",
+          [](SNode *snode, int i) -> SNode * { return snode->ch[i].get(); },
+          py::return_value_policy::reference)
       .def("lazy_grad",
            [](SNode *snode) {
              make_lazy_grad(snode,
@@ -372,13 +379,14 @@ void export_lang(py::module &m) {
 
   py::class_<Stmt>(m, "Stmt");
   py::class_<Program::KernelProxy>(m, "KernelProxy")
-      .def("define",
-           [](Program::KernelProxy *ker,
-              const std::function<void()> &func) -> Kernel & {
-             py::gil_scoped_release release;
-             return ker->def(func);
-           },
-           py::return_value_policy::reference);
+      .def(
+          "define",
+          [](Program::KernelProxy *ker,
+             const std::function<void()> &func) -> Kernel & {
+            py::gil_scoped_release release;
+            return ker->def(func);
+          },
+          py::return_value_policy::reference);
 
   m.def("insert_deactivate", [](SNode *snode, const ExprGroup &indices) {
     return Deactivate(snode, indices);
--- a/taichi/transforms/loop_invariant_code_motion.cpp
+++ b/taichi/transforms/loop_invariant_code_motion.cpp
@@ -63,9 +63,6 @@ class LoopInvariantCodeMotion : public BasicStmtVisitor {
       }
     }
 
-    //if (can_be_moved)
-    //  TI_WARN("stmt can be moved");
-
     return can_be_moved;
   }
 
--- a/taichi/transforms/loop_invariant_code_motion.cpp
+++ b/taichi/transforms/loop_invariant_code_motion.cpp
@@ -39,7 +39,8 @@ class LoopInvariantCodeMotion : public BasicStmtVisitor {
       if (config.move_loop_invariant_outside_if &&
           stmt->parent != loop_blocks.top()) {
         // If we enable moving code from a nested if block, we need to check
-        // visibility Example: for i in range(10):
+        // visibility. Example:
+        // for i in range(10):
         //   a = x[0]
         //   if b:
         //     c = a + 1
--- a/taichi/transforms/loop_invariant_code_motion.cpp
+++ b/taichi/transforms/loop_invariant_code_motion.cpp
@@ -1,0 +1,160 @@
+#include "taichi/ir/ir.h"
+#include "taichi/ir/statements.h"
+#include "taichi/ir/transforms.h"
+#include "taichi/ir/visitors.h"
+#include "taichi/system/profiler.h"
+
+#include <stack>
+
+TLANG_NAMESPACE_BEGIN
+
+class LoopInvariantCodeMotion : public BasicStmtVisitor {
+ public:
+  std::stack<Block *> loop_blocks;
+
+  const CompileConfig &config;
+
+  DelayedIRModifier modifier;
+
+  LoopInvariantCodeMotion(const CompileConfig &config) : config(config) {
+    allow_undefined_visitor = true;
+  }
+
+  bool stmt_can_be_moved(Stmt *stmt) {
+    if (loop_blocks.size() <= 1 || (!config.move_loop_invariant_outside_if &&
+                                    stmt->parent != loop_blocks.top()))
+      return false;
+
+    bool can_be_moved = true;
+
+    Block *current_scope = stmt->parent;
+
+    for (Stmt *operand : stmt->get_operands()) {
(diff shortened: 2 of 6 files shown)


xumingkuan:
Hi @bobcao3 , thanks for the contribution and welcome to the Taichi developer community!

There are a few issues here:
1. The pass `full_simplify` is invoked both before and after `offload`. The *top-level `RangeForStmt` and `StructForStmt`* before `offload` will be offloaded into *parallel* tasks after that pass. After `offload`, there will not be any `RangeForStmt` and `StructForStmt`s at the top-level because the top level must be an `OffloadedStmt`. There can be serial `RangeForStmt`s inside the  `OffloadedStmt`. So we should not move the loop-invariant statements in the top-level `RangeForStmt` and `StructForStmt` because it may not only incur some correctness issues (although I think it should also be correct) but also increase the overhead of one more possible serial offloaded task.
2. The scope (`loop_block` in the PR) is not properly handled when there are nested loops -- when exiting the inner one, the outer one's information is lost.

To solve these 2 issues, I think we can maintain a stack for loops, and also consider `OffloadedStmt`s with range-for or struct-for types as loops, and do not perform the optimization for the top-level loop (i.e., when the stack's size is 1).

3. The function `stmt->has_global_side_effect()` is misused in some places and probably need to be refactored in the future. For now, it's better to reduce introducing new usages of them. And actually, `BinaryOpStmt::has_global_side_effect()` and `UnaryOpStmt::has_global_side_effect()` are always `false`, so we can remove that check.
4. The return value (`modified`) shows if the pass modifies the IR. You can refer to this part as an example of using `modify_ir()` and the return value: https://github.com/taichi-dev/taichi/blob/6586d6b53a2d39e6b274c4ec952215adb71896be/taichi/transforms/alg_simp.cpp#L316-L327

bobcao3:
Introduced a stack & only applying this when stack size > 1.

bobcao3:
One question remains, consider this structure:
```
a, b = ...
for i in range():
  ...
  if i < N:
    c = a + b
```

Should this `c` be moved outside? Without allowing this, many programs generated by taichi will not have anything to be moved due to most computations being wrapped within these ifs (seems to be an result of promoting container sizes or some range checking?)

xumingkuan:
> One question remains, consider this structure:
> 
> ```
> a, b = ...
> for i in range():
>   ...
>   if i < N:
>     c = a + b
> ```
> 
> Should this `c` be moved outside? Without allowing this, many programs generated by taichi will not have anything to be moved due to most computations being wrapped within these ifs (seems to be an result of promoting container sizes or some range checking?)

Good question! This can be an optimization, but can also lead to more computations for if branches which are seldomly selected. Maybe you can do some benchmarks to see if moving statements inside ifs to outside loops can accelerate Taichi programs.

bobcao3:
I have some numbers in the issue related to this. #2324 only one case sees
an improvement (CUDA for Cornell box), for all other cases including stuff
in the if destroys performance. I think if this is to be added, either it
will be an user enabled special flag, or we need some sort of prediction
and heuristic on how often the branch is taken, how large is the loop
containing it, and how much penalty would be involved (e.g. division vs
multiplication adds)

taichi/transforms/redundancy_in_loop_elimination.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include "taichi/ir/statements.h"
+#include "taichi/ir/transforms.h"
+#include "taichi/ir/visitors.h"
+#include "taichi/system/profiler.h"
+
+#include <stack>
+
+TLANG_NAMESPACE_BEGIN
+
+//#define LOOP_INVARIANT_ONLY_ALLOWING_LOOPS
```
xumingkuan, line 11: I think this can be a `CompileConfig` -- something like `move_loop_invariant_outside_if`, with default value `False`. See compile_config.h for other examples.

taichi/transforms/redundancy_in_loop_elimination.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+class RedundancyInLoopElimination : public IRVisitor {
+ public:
+  bool modified = false;
```
xumingkuan, line 15: This field is unnecessary -- we can just use `DelayedIRModifier::modify_ir()`.

taichi/transforms/redundancy_in_loop_elimination.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  void visit(OffloadedStmt *stmt) override {
+    if (stmt->tls_prologue)
+      stmt->tls_prologue->accept(this);
+
+    if (stmt->bls_prologue)
+      stmt->bls_prologue->accept(this);
+
+    if (stmt->body)
+      stmt->body->accept(this);
```
xumingkuan, line 110: This one should also be a loop when the task type is range_for or struct_for.

taichi/transforms/redundancy_in_loop_elimination.cpp
```diff
@@ -0,0 +1,146 @@
...
+    if (stmt->tls_epilogue)
+      stmt->tls_epilogue->accept(this);
+  }
+
+  void visit(IfStmt *if_stmt) override {
+    if (if_stmt->true_statements)
+      if_stmt->true_statements->accept(this);
+    if (if_stmt->false_statements)
+      if_stmt->false_statements->accept(this);
+  }
```
xumingkuan, line 124: nit: If `BasicStmtVisitor` (which automatically dives into container statements) is used instead of `IRVisitor`, this part can be removed.

taichi/transforms/redundancy_in_loop_elimination.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  static bool run(IRNode *node) {
```
xumingkuan, line 126: Please modify this function like https://github.com/taichi-dev/taichi/blob/6586d6b53a2d39e6b274c4ec952215adb71896be/taichi/transforms/alg_simp.cpp#L316-L327 so that the return value is well-defined.

taichi/transforms/redundancy_in_loop_elimination.cpp
```diff
@@ -0,0 +1,146 @@
...
+      node->accept(&eliminator);
+      modified = eliminator.modifier.modify_ir();
+    } while (modified);
+
+    return modified;
+  }
+};
+
+namespace irpass {
+bool redundancy_in_loop_elimination(IRNode *root) {
```
xumingkuan, line 140: I think the standard name of this optimization is "Loop-invariant code motion". Let's use the standard name :)

taichi/transforms/redundancy_in_loop_elimination.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  std::stack<Block *> loop_blocks;
+
+  DelayedIRModifier modifier;
+
+  RedundancyInLoopElimination() {
+    allow_undefined_visitor = true;
+  }
+
+  bool visit_arithmetic(Stmt *stmt) {
```
xumingkuan, line 25: nit: I think `can_be_moved` is a more intuitive name than `visit_arithmetic`.

xumingkuan (changes requested):

taichi/transforms/redundancy_in_loop_elimination.cpp
bobcao3: Thanks! With my super limited knowledge I can't even come up with the word "loop-invariant" haha

taichi/transforms/loop_invariant_code_motion.cpp
```diff
@@ -0,0 +1,137 @@
...
+      return false;
+
+    bool can_be_moved = true;
+
+    Block *current_scope = stmt->parent;
+
+    for (Stmt *operand : stmt->get_operands()) {
+      if (operand->parent == current_scope) {
+        // Has an input that is in the current scope
+        // Operand can either be in the current scope or a outer scope
```
xumingkuan, line 35:
```suggestion
        // This statement has an operand that is in the current scope,
        // so it can not be moved out of the scope.
```

I just thought about a case:
```python
for i in range(10):
  a = x[0]
  if b:
    c = a + x[1]  # a is not an operand in the current scope, but we cannot move it out.
```
(`x` is a global field preventing other optimizations such as constant fold)

Would you please add some tests like this? I'm not sure if we need something like
https://github.com/taichi-dev/taichi/blob/a4279bd8e1aa6b7d0d269ff909773d333fab5daa/taichi/analysis/verify.cpp#L42
to check if the operand is visible outside the loop.

taichi/transforms/loop_invariant_code_motion.cpp
```diff
@@ -0,0 +1,137 @@
...
+        can_be_moved = false;
+        break;
+      }
+    }
+
+    if (can_be_moved) {
+      TI_TRACE("Stmt can be moved: {} {}", stmt->name(), typeid(*stmt).name());
+    }
```
xumingkuan, line 43: nit: Remove this debug output?

taichi/python/export_lang.cpp
```diff
@@ -139,6 +139,8 @@ void export_lang(py::module &m) {
       .def_readwrite("simplify_after_lower_access",
                      &CompileConfig::simplify_after_lower_access)
       .def_readwrite("lower_access", &CompileConfig::lower_access)
+      .def_readwrite("simplify_before_lower_access",
+                     &CompileConfig::simplify_before_lower_access)
```
xumingkuan, line 143: Should this be changed to `move_loop_invariant_outside_if`?

xumingkuan (commented):
LGTM with some nits!

taichi/transforms/loop_invariant_code_motion.cpp
bobcao3: For this case I thought about it but I think `x[1]` should be an global load operation right? Since we don't move load store operations, then `c = a + x[1]` can not be moved as `x[1]` is in the `if` scope.

taichi/python/export_lang.cpp
bobcao3: Forget to change this haha

taichi/transforms/loop_invariant_code_motion.cpp
xumingkuan: You’re right, but what about when the statement is `c = a + 1`, and the constant 1 is outside the if clause?

bobcao3: I don't quite know why the statement can't be moved if that's the case?

xumingkuan:
```python
for i in range(10):
  a = x[0]
  if b:
    c = a + 1
    print(c)
```
If we move `c = a + 1` out of the loop:
```python
c = a + 1  # a is not visible here!
for i in range(10):
  a = x[0]
  if b:
    print(c)
```

bobcao3: Oh yeah you are right, this does need a check. This is only for statements within a if block nested in a loop block right?

xumingkuan: Right, we need to flag loads from read-only data and move those as well. I think we can add it in another PR.

taichi/transforms/loop_invariant_code_motion.cpp
```diff
@@ -0,0 +1,160 @@
...
+            break;
+          }
+        }
+        if (!can_be_moved)
+          break;
+      }
+    }
+
+    //if (can_be_moved)
+    //  TI_WARN("stmt can be moved");
```
xumingkuan, line 66: nit: remove this debug info

taichi/transforms/loop_invariant_code_motion.cpp
```diff
@@ -0,0 +1,160 @@
...
+        // This statement has an operand that is in the current scope,
+        // so it can not be moved out of the scope.
+        can_be_moved = false;
+        break;
+      }
+      if (config.move_loop_invariant_outside_if &&
+          stmt->parent != loop_blocks.top()) {
+        // If we enable moving code from a nested if block, we need to check
+        // visibility Example: for i in range(10):
+        //   a = x[0]
```
xumingkuan, line 43: nit:
```suggestion
        // visibility. Example:
        // for i in range(10):
        //   a = x[0]
```

xumingkuan (approved):
LGTM, thanks!

taichi/transforms/loop_invariant_code_motion.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    return can_be_moved;
+  }
+
+  void visit(BinaryOpStmt *stmt) override {
+    if (stmt_can_be_moved(stmt)) {
+      auto replacement = stmt->clone();
+      stmt->replace_with(replacement.get());
+
+      modifier.insert_before(stmt->parent->parent_stmt, std::move(replacement));
```
lin-hitonami, line 74: Shall `stmt->parent->parent_stmt` here be `loop_blocks.top()->parent_stmt`? If a statement is inside an `if` inside a `for`, it is moved out of the `if` but it is still inside the `for`. This may also be the reason why the performance is low when you enabled `move_loop_invariant_outside_if`.

lin-hitonami: I'll fix it in #6072 and test again.

bobcao3: Ah right makes sense, thanks for noticing it!

xumingkuan: Good catch! However, I think the result will be the same. For example, in this code,
```python
a = x[0]
for i in range(10):
  if b:
    c = a + 1
```
we will move `c = a + 1` like this:
```python
a = x[0]
for i in range(10):
  c = a + 1
  if b:
    pass
```
In the next iteration of this pass, we will have:
```python
a = x[0]
c = a + 1
for i in range(10):
  if b:
    pass
```
So this will only increase the number of iterations by 1 and only if this is the bottleneck (the statement that will be moved the most times).

Nevertheless, thanks for catching this!

bobcao3: We might need a cost function to properly address this

PR #2327, opened 2021-05-07
[Lang] [ir] Consecutive indices for all snodes

KLozes:
Related issue = #2177
This is main feature of the draft PR #2248. It makes indices for all snodes consecutive. After this I can make a PR to fully deprecate trailing bits.  

----


k-ye (commented):
Sorry , just some update to let you know that I'm still reviewing this PR.  I've spent quite some time reviewing the refine coordinates part. From what I understand, the end result of running the `refine_coordinates` from root to leaf is still the same, it's the intermediary results that are changed. If so, would it be possible to separate this part of the change to a smaller PR?

taichi/ir/scratch_pad.h
```diff
@@ -32,6 +32,7 @@ class ScratchPad {
   SNode *snode;
   using AccessFlag = taichi::lang::AccessFlag;
 
+  std::vector<int> coeff;
```
k-ye, line 35: nit: `coefficients`?

taichi/runtime/llvm/runtime.cpp
```diff
@@ -1093,7 +1089,7 @@ void element_listgen_root(LLVMRuntime *runtime,
     elem.element = ch_element;
     elem.loop_bounds[0] = c * ch_element_size;
     elem.loop_bounds[1] = std::min((c + 1) * ch_element_size, ch_num_elements);
-    elem.pcoord = refined_coord;
+    elem.pcoord = element.pcoord;
```
k-ye, line 1092: Could you add a comment explaining that we don't have to refine coordinates, because root snode's num_bits is always 0?

taichi/transforms/make_block_local.cpp
```diff
@@ -139,14 +139,20 @@ void make_block_local_offload(OffloadedStmt *offload,
...
+                  BinaryOpType::add, bls_coord,
                   element_block->push_back<ConstStmt>(
-                      TypedConstant(pad.second.bounds[0][i])),
-                  bls_coord);
+                      TypedConstant(pad.second.bounds[0][i])));
+
+              auto block_corner =
+                  element_block->push_back<BlockCornerIndexStmt>(offload, i);
+
+              block_corner = element_block->push_back<BinaryOpStmt>(
```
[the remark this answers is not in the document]
KLozes: Alright, I can throw an 'if' there to check. I imagine it would get optimized out by the final compiler anyways though, right?

taichi/codegen/codegen_llvm.cpp
```diff
@@ -1552,6 +1552,14 @@ void CodeGenLLVM::create_offload_struct_for(OffloadedStmt *stmt, bool spmd) {
     auto upper_bound = get_arg(4);
 
     parent_coordinates = element.get_ptr("pcoord");
+    block_corner_coordinates =
```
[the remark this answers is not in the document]
KLozes: Should I set the rest of the coordinates to null as well: parent_coordinates, current_coordinates, new_coordinates?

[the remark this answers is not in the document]
k-ye: Ah yeah, that will be great. We don't have to do this for `new_coordinates` though, because it is a local variable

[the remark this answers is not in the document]
k-ye: Ah yeah, that will be great. We don't have to do this for `new_coordinates` though, because it is a local variable

tests/python/test_sparse_deactivate.py
```diff
@@ -25,7 +25,7 @@ def func():
 
     @ti.kernel
     def deactivate():
-        ti.deactivate(ptr, 4)
+        ti.deactivate(ptr, 1)
```
k-ye, line 28: Hmm, I don't think this is relevant with this PR. That is, even before this change, changing this to `1` still passes the test.

IIUC, changing to consecutive indices requires us to change how `GlobalPtr` gets lowered to a chain of `GetCh` and `SNodeLookup` stmts. By having `print_ir=True`, you can see that there is a `bit_extract($2) bit_range=[4, 8)` stmt, where `$2` is the indices to be deactivated.  The `bit_range[4,8)` is what is actually causing the non-consecutive indices. @xumingkuan to confirm.

KLozes: So before this PR `ti.deactivate(ptr, 4)` would deactivate the first block. In fact, any index 0-15 would deactivate the first block. So to maintain the same exact behavior with with consecutive indices, I should have changed it to `ti.deactivate(ptr, 0)`. I just got lucky that `ti.deactivate(ptr, 1)` worked actually.  With this PR `ti.deactivate(ptr, 4)` defiently will not pass. Since it would deactivate the 5th block, which was never activated in the first place!

KLozes: will change to `ti.deactivate(ptr, 0)`

taichi/struct/struct.cpp
```diff
@@ -52,48 +52,30 @@ void StructCompiler::infer_snode_properties(SNode &snode) {
...
-          SNode *p = snode.parent;
-          while (p) {
-            TI_ASSERT_INFO(
-                p->extractors[i].num_bits == 0,
-                "Dynamic SNode must have a standalone dimensionality.");
-            p = p->parent;
-          }
+  // infer extractors
+  int acc_offsets = 0;
+  for (int i = taichi_max_num_indices - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
KLozes: We shouldn't have to check that anymore! The shape of all snodes are now totally independent of their siblings. See the complex struct test:
``` python
   blk = ti.root.pointer(ti.ij, 4)
   blk.place(w)
   blk.pointer(ti.ij, 2).dense(ti.ij, 2).place(x)
   blk.dense(ti.i, 4).place(y)
   blk.dense(ti.k, 4).place(z)
```
Structures like this are now totally valid

tests/python/test_sparse_deactivate.py
xumingkuan: Right, `bit_range=[4, 8)` will become `bit_range=[0, 4)` after this PR. `ti.deactivate(ptr, 0)` SGTM!

taichi/transforms/make_block_local.cpp
```diff
@@ -212,7 +218,11 @@ void make_block_local_offload(OffloadedStmt *offload,
           //   bls_stride_i * (gbl_idx_i - loop_base_i - bls_lower_bound_i)
           // Note that when index offsets are used, the offset contributions are
           // already included in bls_lower_bound_i.
-          auto block_corner = bls.push_back<BlockCornerIndexStmt>(offload, i);
+
+          auto block_corner = bls.push_back<BinaryOpStmt>(
```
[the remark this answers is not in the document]
KLozes: Thanks for the tip!

k-ye (commented):
Thank you for moving this forward! I only have  a few nits and test suggestions now

tests/python/bls_test_template.py
```diff
@@ -183,13 +186,13 @@ def p2g(use_shared: ti.template(), m: ti.template()):
         ti.block_dim(256)
         if ti.static(use_shared):
             ti.block_local(m)
-        for i, j, l in pid:
-            p = pid[i, j, l]
+        for I in ti.grouped(pid):
```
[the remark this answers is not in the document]
KLozes: How do I run a cpp test btw?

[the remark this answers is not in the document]
k-ye: After `make`, there will be a `taichi_cpp_tests` executable. You can just run that and see the results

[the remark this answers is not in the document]
KLozes: Im sorry, where is the `taichi_cpp_tests` located? I don't see a Makefile over in tests, so I'm guessing you mean after making the project.

taichi/transforms/lower_access.cpp
```diff
@@ -82,8 +82,13 @@ class LowerAccess : public IRVisitor {
     }
 
     std::deque<SNode *> snodes;
-    for (auto s = leaf_snode; s != nullptr; s = s->parent)
+    int start_bits[taichi_max_num_indices] = {0};
```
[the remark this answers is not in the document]
KLozes: Sounds good to me!

tests/python/bls_test_template.py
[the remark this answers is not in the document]
KLozes: Right.. I do not see taichi_cpp_tests exec anywhere! This all I have after make
```
klozes@system76-pc:~/Documents/software/taichi/build$ ls -a
.  ..  CMakeCache.txt  CMakeFiles  cmake_install.cmake  external  libtaichi_core.so  Makefile
```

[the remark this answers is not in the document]
k-ye: My bad! You need to invoke `cmake` with `-DTI_BUILD_TESTS=ON`...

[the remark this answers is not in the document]
KLozes: thanks!

KLozes:
@k-ye and @xumingkuan,

Thank you for the valuable comments and reviews! I think I've implemented all the suggestions. I think were just about home free.

xumingkuan (approved):
Great work, LGTM!

xumingkuan:
Well... I wonder if `ScratchPad::pad_size_linear()` should be multiplied with the coefficients?

k-ye:
> Thank you for the valuable comments and reviews! I think I've implemented all the suggestions. I think were just about home free.

Sorry, I need to go over the entire process of `make_block_local()` again. I believe there are a few issues because of the coefficients.

k-ye (approved):
> Well... I wonder if `ScratchPad::pad_size_linear()` should be multiplied with the coefficients?

Giving it more thoughts, I believe this is not necessary, and what is done by this PR looks correct. The tricky part is that `ScratchPad` corresponds to the SNode that the BLS wants to cache, not the one being iterated over in the struct-for. Hence, the `pad_size` will be correctly scaled by the coefficients already. As a comparison, `BlockCornerStmt` corresponds to the SNode being iterated over, so we have to manually scale this one by the coefficients.

@KLozes it was such a great pleasure working with you on this PR! This fundamentally touches and changes so many critical parts of Taichi. Learnt a great deal from the work. And thanks @xumingkuan @yuanming-hu  for many rounds of offline discussion!

One last thing: I'm working on a local PR to make `make_block_local()` testable. Could you hold off on that a bit?

taichi/struct/struct_llvm.cpp
```diff
@@ -262,7 +262,7 @@ void StructCompilerLLVM::run(SNode &root, bool host) {
 
   if (host) {
     infer_snode_properties(root);
-    compute_trailing_bits(root);
+    // compute_trailing_bits(root);
```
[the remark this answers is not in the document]
KLozes: ****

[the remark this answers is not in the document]
KLozes: Yes, trailing bits should always be zero now. I'm planning to totally remove trailing_bits in a subsequent PR.

KLozes:
> @KLozes` it was such a great pleasure working with you on this PR! This fundamentally touches and changes so many critical parts of Taichi. Learnt a great deal from the work. And thanks @xumingkuan @yuanming-hu for many rounds of offline discussion!

@k-ye It was pleasure working with you too! I really enjoy following the great work you guys do on this project. I've learned a ton working on this PR too. It ended up being more complicated than I expected. 

> One last thing: I'm working on a local PR to make make_block_local() testable. Could you hold off on that a bit?

Sure thing, I'll try to keep this PR update to date with the trunk in the mean time.

KLozes:
uhoh @xumingkuan  @k-ye, 

This new optimization pass #2332 broke block local storage. DiffRange cannot handle the left shift operator. I think we realistically only need to support two cases: 
index << const
const << const

So something like this could work
``` cpp
DiffRange operator<<(const DiffRange &a, const DiffRange &b) {
  return DiffRange(
      a.related_() && b.related_() && b.coeff == 0 && b.high - b.low == 1,
      a.coeff << b.low, a.low << b.low, ((a.high - 1) << b.low) + 1);
}
```

k-ye:
> uhoh @xumingkuan @k-ye,
> 
> This new optimization pass #2332 broke block local storage. DiffRange cannot handle the left shift operator. I think we realistically only need to support two cases:
> index << const
> const << const
> 
> So something like this could work
> 
> ```c++
> DiffRange operator<<(const DiffRange &a, const DiffRange &b) {
>   return DiffRange(
>       a.related_() && b.related_() && b.coeff == 0 && b.high - b.low == 1,
>       a.coeff << b.low, a.low << b.low, ((a.high - 1) << b.low) + 1);
> }
> ```

I see. Feel free to add this in a new PR, thanks!

xumingkuan:
> uhoh @xumingkuan @k-ye,
> 
> This new optimization pass #2332 broke block local storage. DiffRange cannot handle the left shift operator. I think we realistically only need to support two cases:
> index << const
> const << const
> 
> So something like this could work
> 
> ```c++
> DiffRange operator<<(const DiffRange &a, const DiffRange &b) {
>   return DiffRange(
>       a.related_() && b.related_() && b.coeff == 0 && b.high - b.low == 1,
>       a.coeff << b.low, a.low << b.low, ((a.high - 1) << b.low) + 1);
> }
> ```

Thanks for pointing this out! `DiffRange operator<<` sounds great to me.

Also, this shows that our tests don't cover this case 😅

KLozes:
I'll make a cpp test for it. :)

tests/cpp/transforms/make_block_local_test.cpp
```diff
@@ -130,7 +131,10 @@ TEST_F(MakeBlockLocalTest, Basic) {
   // the generated BLS offset is correct for the given global loop indices.
   for (int idx0 = 0; idx0 < loop_shape_at(0); ++idx0) {
     for (int idx1 = 0; idx1 < loop_shape_at(1); ++idx1) {
-      const int loop_indices_vals[] = {idx0, idx1};
+      const int loop_indices_vals[] = {
+          get_block_size(0) * idx0,
```
k-ye, line 135: IIUC, without the trailing_bits, the `loop_indices_vals` will not be scaled by block_size? (I.e., the looping range is the same as the shape of `struct_for_snode_`)? The block corner itself is not scaled by block_size, either. Instead, `make_bls()` will generate a statement to scale that by coefficients.

KLozes: Yep. exactly right :)

k-ye: Then neither L135-136 nor  `get_block_corner()` needs to multiply by `get_block_size(x)`? I guess we also need to update how `expected_bls_offset_in_bytes` are calculated..

KLozes: Oh sorry I misread your comment. So the coefficients of the scratch pad come directly from the user multiplying the loop index with the block size. This is why in the mpm example, we have to manually rescale the loop index to make it a block_corner before passing it to assume_in_range. 
Also, This test is not a perfect copy of the of mpm example since we are not using the position of the particles to get the central index of the neighborhood to do the scatter/gather. Instead we are just using the loop index, which is just a block index of the grid snode. So we need to rescale this one too, making it also simply a block corner.

k-ye: > So the coefficients of the scratch pad come directly from the user multiplying the loop index with the block size. This is why in the mpm example, we have to manually rescale the loop index to make it a block_corner before passing it to assume_in_range.

Right, I should have better explained the test code: the `loop_indices_vals` array is meant to mock the value of `LoopIndexStmt` for each axis in the `ArithmeticInterpretor`. Similarly, `block_corner_vals` is to mock each `BlockCornerIndexStmt` (see https://github.com/taichi-dev/taichi/blob/22e33a605b5caccc62a02be491a3dc33c2f31c91/tests/cpp/transforms/make_block_local_test.cpp#L170-L177). As you've pointed out, it is the user that is manually multiplying loop index by the block size (L105-108). So `LoopIndexStmt`/`BlockCornerIndexStmt` themselves should not be scaled yet.

We happen to use `loop_indices_vals` and `block_corner_vals` to compute `expected_bls_offset_in_bytes`, which is probably causing lots of confusion. Since the formula of BLS index <--> global indices  has changed, I think we can update how `expected_bls_offset_in_bytes` gets computed. I.e.:

```
expected_bls_offset_in_bytes = sum_i inc_i
inc_i = ( loop_index_i * block_size   # multiplication done by the user
            - block_corner_i * block_size  # multiplication generated by the compiler
            - bls_lower_bounds_i)
```

WDYT?

KLozes: Ahh yes, I think I see what you mean. the BlockCornerIndexStmt _should_ be returning consecutive indices in this cases since we are looping over a dynamic snode which has zero num_bits for i and j. So it does not get refined to non conecutive indices.

KLozes: I have implemented your suggestion

tests/cpp/analysis/bls_analyzer_test.cpp
```diff
@@ -8,12 +8,25 @@
 #include "taichi/ir/snode.h"
 #include "taichi/ir/transforms.h"
 #include "taichi/struct/struct.h"
-#include "tests/cpp/struct/fake_struct_compiler.h"
 
 namespace taichi {
 namespace lang {
 namespace {
 
+class TestStructCompiler : public StructCompiler {
```
k-ye, line 16: Remove?

tests/cpp/analysis/bls_analyzer_test.cpp
```diff
@@ -74,6 +88,68 @@ TEST_F(BLSAnalyzerTest, Basic) {
   EXPECT_EQ(pad.bounds[1].high, kBlockSize);
 }
 
+TEST_F(BLSAnalyzerTest, Mul) {
```
k-ye, line 91: nit: Since we mostly want to test `DiffRange`'s multiplication and SHL, we can create a new test just for `DiffRange`. But I think these are fine to have :-)

k-ye (approved):
Thank you again for the great works and patience, LGTM!

tests/cpp/analysis/bls_analyzer_test.cpp
KLozes: Oops. not sure how that happened. reverted now

tests/cpp/analysis/bls_analyzer_test.cpp
KLozes: Alright. yeah we don't seem to have any tests specifically for DiffRange yet. I'll figure out how to make some.

KLozes: @k-ye tests add!

k-ye: Awesome. FYI one of our interns will test this PR with https://github.com/taichi-dev/taichi_elements next week. Let's see how well it works (or what we are missing...)!

KLozes:
The format server is not working on my end. Is anyone else having the same problem?

KLozes:
@k-ye. Would you be able to merge this for me? I cant get the code format and CI to pass for some reason no matter what I do.

k-ye:
> @k-ye. Would you be able to merge this for me? I cant get the code format and CI to pass for some reason no matter what I do.

Sure, I'll be happy to do so. FYI @Jack12xl is helping carry out some final verification.

Jack12xl:
Hi, guys

I test the `rescale_index` with the [Taichi_elements](https://github.com/taichi-dev/taichi_elements/). To clarify, the `taichi`  is compiled with `0.7.20` merged with [this](https://github.com/taichi-dev/taichi/pull/2327) pull request. The [taichi elements](https://github.com/Jack12xl/taichi_elements/tree/final-merge-cons) is at this [commit ](https://github.com/Jack12xl/taichi_elements/tree/final-merge-cons). 

I basically test with the `demo_2d` and `demo_3d`, which is running smoothly after merging the [70](https://github.com/taichi-dev/taichi_elements/pull/70) PR with some fix.

I believe the consecutive index is ready to go !

KLozes:
> I test the rescale_index with the Taichi_elements. To clarify, the taichi is compiled with 0.7.20 merged with this pull request. The taichi elements is at this commit .

Yep that sounds right! Thanks for the review @Jack12xl :).

k-ye:
Sorry I forgot to loop you in. https://github.com/taichi-dev/taichi/pull/2366 has tweaked `BLSAnalyzer` to introduce a `run()` method. Could you update the corresponding tests? Thanks!

taichi/analysis/bls_analyzer.cpp
```diff
@@ -44,14 +44,17 @@ void BLSAnalyzer::record_access(Stmt *stmt, AccessFlag flag) {
...
     std::vector<IndexRange> offsets;
+    std::vector<int> coeffs;
     offsets.resize(ptr->indices.size());
+    coeffs.resize(ptr->indices.size());
     const int num_indices = (int)ptr->indices.size();
     for (int i = 0; i < num_indices; i++) {
       auto diff = irpass::analysis::value_diff_loop_index(ptr->indices[i],
                                                           for_stmt_, i);
-      if (diff.linear_related()) {
+      if (diff.related_() and diff.coeff > 0) {
```
k-ye, line 54: s/and/&&, msvc doesn't seem to support `and` 😂

k-ye (approved):

taichi/analysis/bls_analyzer.cpp
KLozes: Oops! I had no idea that was even valid syntax with any compiler! will fix

k-ye:
I think it's fine to temporarily ignore the format issue in this PR. The clang-format on the format server has not played well when there's a change to `export_lang.cpp` for a long time.. We're upgrading the automatic code formatting CI process.

KLozes:
> I think it's fine to temporarily ignore the format issue in this PR

Sounds good to me! Feel free to merge whenever :).

k-ye:
Merged :-) Thank you ALL !! @KLozes @xumingkuan @Jack12xl @yuanming-hu

PR #2332, opened 2021-05-11
[Opt] Simplify multiplying/dividing POT

xumingkuan:
Related issue = #944 #656 #2177

This PR may help with some optimizations after #2327 is merged. When we want to access a leaf (place) SNode and (the same address of) its parent or some ancestor, simple optimizations such as CSE can remove the duplicated part of the access paths after lower_access when nonconsecutive indices are used (now). However, when consecutive indices are used, we may have some code like this:
```python
a = ti.field(ti.i32)
block = root.dense(ti.i, 64)
block.dense(ti.i, 3).place(a)

@ti.kernel
def foo():
  for i in ...:
    ti.activate(block, i / 8)
    a[i]...
```
And we want the common part of  `block[i / 8]` and `a[i]` is reused in the two accesses. After expanding the `BitExtractStmt`, the former one will be `(i / 8) & 63` and the latter one will be `(i >> 3) & 63`. This PR replaces `i / 8` with `i >> 3`, and further replaces `(a >> b) >> c` with `a >> (b + c)` when both `b` and `c` are constants, so that CSE will be able to remove the duplicated access paths after lower_access.

This PR also enables constant folding for unsigned data types.

----

### DIFF
--- a/cmake/TaichiTests.cmake
+++ b/cmake/TaichiTests.cmake
@@ -10,7 +10,8 @@ endif()
 # TODO(#2195):
 # 1. "cpp" -> "cpp_legacy", "cpp_new" -> "cpp"
 # 2. Re-implement the legacy CPP tests using googletest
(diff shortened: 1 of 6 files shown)


k-ye (approved):
LGTM!

PR #2334, opened 2021-05-13
[Example] cornell_box: Remove unused sphere

jammm:
Hello ! Was playing around with this and found the second sphere position/radius unused, so thought it can be removed. Let me know if it's still needed however, in which case we don't need to merge.

Cheers,
Aaryaman


k-ye (commented):
Thank you LGTM!

PR #2337, opened 2021-05-13
[opt] Slightly simplify algebraic simplification

xumingkuan:
Related issue = a continuation of #2332

`val_cast_to_uint64()` looks cleaner than two if-branches with the same body.

----

### DIFF
--- a/taichi/ir/type.cpp
+++ b/taichi/ir/type.cpp
@@ -355,8 +355,10 @@ float64 TypedConstant::val_float() const {
 }
 
 int64 TypedConstant::val_cast_to_int64() const {
-  if (is_real(dt))
+  if (is_real(dt)) {
+    TI_WARN("Casting floating point type {} to i64.", dt->to_string());
     return val_float();
+  }
   else if (is_signed(dt))
     return val_int();
   else if (is_unsigned(dt))
@@ -367,8 +369,10 @@ int64 TypedConstant::val_cast_to_int64() const {
 }
 
 uint64 TypedConstant::val_cast_to_uint64() const {
-  if (is_real(dt))
+  if (is_real(dt)) {
+    TI_WARN("Casting floating point type {} to u64.", dt->to_string());
     return val_float();
+  }
   else if (is_signed(dt))
     return val_int();
   else if (is_unsigned(dt))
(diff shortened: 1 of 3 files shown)


xumingkuan:
I'm not sure if we need a function to cast integer values to int64/uint64 (i.e., handle signed/unsigned cases together)...

k-ye:
> I'm not sure if we need a function to cast integer values to int64/uint64 (i.e., handle signed/unsigned cases together)...

We already have `val_int()` and `val_uint()`?

xumingkuan:
> > I'm not sure if we need a function to cast integer values to int64/uint64 (i.e., handle signed/unsigned cases together)...
> 
> We already have `val_int()` and `val_uint()`?

These 2 functions only handles signed/unsigned versions separately. I wonder if we want to use only one function when we don't care if the value of the `TypedConstant` is signed or not.

> We do need to add the `const` version of these functions. For now they return a mutable reference to the underlying data..

They are already `const`, I think.

k-ye:
> > > I'm not sure if we need a function to cast integer values to int64/uint64 (i.e., handle signed/unsigned cases together)...
> > 
> > 
> > We already have `val_int()` and `val_uint()`?
> 
> These 2 functions only handles signed/unsigned versions separately. I wonder if we want to use only one function when we don't care if the value of the `TypedConstant` is signed or not.

Right, sounds like it could be useful

> > We do need to add the `const` version of these functions. For now they return a mutable reference to the underlying data..
> 
> They are already `const`, I think.

Ah, i was talking about these:

https://github.com/taichi-dev/taichi/blob/c824a8b145c2d09d52dc7c0a65b91a102e451cc9/taichi/ir/type.h#L402-L411

xumingkuan:
> > > We do need to add the `const` version of these functions. For now they return a mutable reference to the underlying data..
> > 
> > 
> > They are already `const`, I think.
> 
> Ah, i was talking about these:
> 
> https://github.com/taichi-dev/taichi/blob/c824a8b145c2d09d52dc7c0a65b91a102e451cc9/taichi/ir/type.h#L402-L411

Right, I agree that we need these, but maybe copying these functions will make this file too long?

k-ye:
> Right, I agree that we need these, but maybe copying these functions will make this file too long?

That shouldn't be a concern...

xumingkuan:
This PR should be ready for review now :)

k-ye (approved):

PR #2338, opened 2021-05-13
[Refactor] Add a class Callable to unify Kernel and Function

xumingkuan:
Related issue = #2193 

A continuation of #2306.

This PR unifies the common fields of `Kernel` and `Function` into `Callable`, including `Program *program;`, `std::unique_ptr<IRNode> ir;`, `std::vector<Arg> args;`, `std::vector<Ret> rets;`.

So this PR changes `Program &Kernel::program` into `Program *`, which is a part of the somewhat large change set.

This PR also changes `current_function_or_kernel` introduced in #2306 to `current_callable`, unifies `CurrentFunctionGuard` and `CurrentKernelGuard` to `CurrentCallableGuard`, adds `Callable::get_name()`, and renames `KernelTemplateMapper` to `TaichiCallableTemplateMapper`.

BTW, I found that `program.h` is included in so many files that modifying it causes a long build time, similar to `ir.h`...

----


taichi/program/callable.h
```diff
@@ -0,0 +1,58 @@
...
+  std::vector<Arg> args;
+  std::vector<Ret> rets;
+
+  virtual ~Callable() = default;
+
+  int insert_arg(const DataType &dt, bool is_external_array);
+
+  int insert_ret(const DataType &dt);
+
+  [[nodiscard]] virtual std::string get_name() const;
```
k-ye, line 44:
```suggestion
  [[nodiscard]] virtual std::string get_name() const = 0;
```

k-ye (approved):
LGTM thanks!

taichi/program/callable.h
xumingkuan: Interesting -- a pure virtual function can still have an implementation.

PR #2341, opened 2021-05-13
[opt] Better encapsulate BLS bounds

k-ye:
Related issue = #2177, #2327 

----

### DIFF
--- a/taichi/ir/scratch_pad.h
+++ b/taichi/ir/scratch_pad.h
@@ -29,11 +29,23 @@ inline AccessFlag operator|=(AccessFlag &a, AccessFlag &b) {
 
 class ScratchPad {
  public:
-  SNode *snode;
+  // The lowest and highest index in each dimension.
+  struct BoundRange {
+    int low{0};
+    int high{0};
+
+    int range() const {
+      return high - low;
+    }
+  };
+
+  SNode *snode{nullptr};
   using AccessFlag = taichi::lang::AccessFlag;
 
-  std::vector<int> bounds[2];
+  std::vector<BoundRange> bounds;
+  // pad_size[i] := bounds[i].high - bounds[i].low
   std::vector<int> pad_size;
+  // block_size[i] := (1 << snode.extractor[i].num_bits)
   std::vector<int> block_size;
   bool finalized;
   int dim;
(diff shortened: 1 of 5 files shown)


taichi/ir/scratch_pad.h
```diff
@@ -224,8 +234,6 @@ class ScratchPads {
   void print() {
     for (auto &it : pads) {
       TI_P(it.first->node_type_name);
-      TI_P(it.second.bounds[0]);
-      TI_P(it.second.bounds[1]);
```
xumingkuan, line 228: nit: If this function is merely used for debugging, why not printing bounds as well?

taichi/transforms/make_block_local.cpp
```diff
@@ -67,45 +68,58 @@ void make_block_local_offload(OffloadedStmt *offload,
...
             i, j, k = bls_to_global(bls_element_id)
             bls[bls_element_id] = x[i, j, k]
             // or x[i, j, k] = bls[bls_element_id]
             bls_element_id += block_dim;
           }
 
+          func bls_to_global(bls_element_id):
+            partial = bls_element_id
+            global_indices = []  // "i, j, k"
+            for i in [(dim - 1), 0]:
```
xumingkuan, line 91: nit:
```suggestion
            for i in reversed(range(0, dim)):
```
`[(dim - 1), 0]` itself has a different meaning in Python so I think we should avoid it, and `range(dim - 1, -1, -1)` (if we replace it with this one) is a little bit hard to understand.

taichi/transforms/make_block_local.cpp
```diff
@@ -67,45 +68,58 @@ void make_block_local_offload(OffloadedStmt *offload,
...
+              pad_size = pad.pad_size[i]  // a.k.a. bounds[i].range()
+              bls_coord = partial % pad_size
+              partial = partial / pad_size
+              global_index_at_i = BlockerCorner[i] + bls_coord
```
xumingkuan, line 95:
```suggestion
              global_index_at_i = BlockCorner[i] + bls_coord
```

xumingkuan (approved):
LGTM in general!

taichi/ir/scratch_pad.h
k-ye: Fixed :)

PR #2342, opened 2021-05-14
[ir] Add ArithmeticInterpretor to evaluate a subset of CHI IR

k-ye:
Related issue = #2177, #2327

As a start, `ArithmeticInterpretor` is used to evaluate a piece of BLS code. We can gradually make it support more IRs, or even grow it to a  full-blown VM for CHI IR.

----

### DIFF
--- a/taichi/analysis/arithmetic_interpretor.cpp
+++ b/taichi/analysis/arithmetic_interpretor.cpp
@@ -1,0 +1,150 @@
+#include "taichi/analysis/arithmetic_interpretor.h"
+
+#include <algorithm>
+#include <type_traits>
+#include <vector>
+
+#include "taichi/ir/type_utils.h"
+#include "taichi/ir/visitors.h"
+
+namespace taichi {
+namespace lang {
+namespace {
+
+using CodeRegion = ArithmeticInterpretor::CodeRegion;
+using EvalContext = ArithmeticInterpretor::EvalContext;
+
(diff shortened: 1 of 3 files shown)


taichi/analysis/arithmetic_interpretor.cpp
```diff
@@ -0,0 +1,150 @@
...
+    auto rhs = rhs_opt.value();
+    if (lhs.dt != rhs.dt) {
+      failed_ = true;
+      return;
+    }
+
+    const auto op = stmt->op_type;
+    const auto dt = lhs.dt;
+    // TODO: Consider using macros to avoid duplication
+    if (is_signed(dt)) {
```
xumingkuan, line 83: Caveat here: `is_signed` asserts `is_integral`. Consider moving `is_real` to the first...

xumingkuan (approved):
LGTM! Maybe we can replace constant_fold with this in the future (when fast_math=True)?

taichi/analysis/arithmetic_interpretor.cpp
k-ye: Nice catch!

k-ye:
> Maybe we can replace constant_fold with this in the future (when fast_math=True)?

That's a good point! I think @yuanming-hu and I talked about this before, and the conclusion was that each backend may yield a different precision, so using this interpretor could result in different results.

PR #2348, opened 2021-05-15
[Docker] Fix the docker build by bumping CMake version in Dockerfile

rexwangcc:
Related issue = #

----

It seems the docker build trigger has been failing on Docker Hub for a couple of weeks... This PR fixes the docker build by using the latest CMake (v3.20.2). A few points I want to flag:

- The `ti test` passed the second time I ran them, while it failed the first time on `test_fuse_dense_x2y2z[x64]` reporting `[snode_types.cpp:snode_type_name@18] Not supported`, I assumed it's transient.
- Ubuntu LTS's `apt` does not ship the latest packages such CMake, which is the root cause of this issue, I switched to use a pinned version of CMake and build fro source, but not sure if the latest (v3.20.2) is too aggressively new?
- I just discovered `python/taichi/core/util.py` has this `in_docker` predicate function which relies on `TI_IN_DOCKER`, so I added it to the Dockerfile. I'm not sure what it was used for (maybe no one is using it which can explain why we never ran into issues).
- I guess, unfortunately, in order to test whether this fixes the post-ci, we have to merge it?

I have tested the newly built docker image works on Ubuntu with CUDA support, following https://taichi.readthedocs.io/en/latest/dev_install.html?highlight=docker#use-docker-image-on-ubuntu-with-cuda-support

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,6 @@ RUN apt-get update && \
                        python3-pip \
                        libtinfo-dev \
                        clang-8 \
-                       cmake \
                        wget \
                        git \
                        libx11-dev \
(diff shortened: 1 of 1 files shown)


k-ye (commented):
Thank you LGTM!

k-ye (approved):

PR #2349, opened 2021-05-17
[ir] Rename KernelReturnStmt to ReturnStmt

xumingkuan:
Related issue = #2193 

Since #2306 uses `KernelReturnStmt` for function returns, I think this statement should be renamed without the word "kernel".

----


k-ye (approved):
Thanks!

PR #2350, opened 2021-05-17
[lang] Support std::optional and enum classes in the serializer

k-ye:
Related issue = #N/A

Could be useful for AOT

----


xumingkuan (approved):
LGTM!

PR #2358, opened 2021-05-20
[lang] Remove serializer's variadic template API

k-ye:
Related issue = Follow up of #2350 

Previous implementation of the serializer's variadic template API:

https://github.com/taichi-dev/taichi/blob/284f75eddd0e172b17b144e4f60d443f32a5520e/taichi/common/serialization.h#L545-L550

https://github.com/taichi-dev/taichi/blob/284f75eddd0e172b17b144e4f60d443f32a5520e/taichi/common/serialization.h#L790-L799

Caused a problems where a struct that doesn't satisfy any of the parameter type known by the serializer *can still be accepted by the serializers*. When none of the known parameter type matches, the C++ compiler will just fallback to this template API. Through a series of template instantiation, this would lead to unterminated recursive calls and stackoverflow at runtime. This is bad as we are supposed to catch such errors at compile time.

This PR refactors how `TI_IO(...)` is implemented, by splitting the stringified `__VA_ARGS__` by `','` at compile time, then invoking  `serializer::operatpr()("arg", arg)` on each args. This way we can remove the variadic template API from the serializers, and detect the aforementioned problem at compile time.

Another fix is that I've completely removed the set of `TI_IO_DECL_INST` macros. These macros define `io()` methods that accepts `Binary{Input|Output}Serializer`, and are (indirectly) called by most CHI IR stmts via `TI_STMT_DEF_FIELDS`.  It has turned out that these stmts are not serializable, because they contain unserializable types like `DataType` (which contains a pointer to the polymorphic `Type` API). Serializing/Deserializing a polymorphic type is not a trivial thing in C++. So let's just remove this functionality since nobody is using it. If we ever need to restore this, we should consider using 3rd party libraries.

----

### DIFF
--- a/taichi/common/serialization.h
+++ b/taichi/common/serialization.h
@@ -5,6 +5,7 @@
 
 #pragma once
 
+#include <array>
 #include <cassert>
 #include <cstring>
 #include <fstream>
@@ -13,6 +14,7 @@
 #include <memory>
 #include <optional>
 #include <sstream>
+#include <string_view>
 #include <type_traits>
 #include <unordered_map>
 #include <vector>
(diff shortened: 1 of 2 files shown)


taichi/common/serialization.h
```diff
@@ -54,71 +56,93 @@ using is_unit = typename std::is_base_of<Unit, remove_cvref_t<T>>;
...
+    size_t si = 0;
+    size_t cur_head_i = 0;
+    size_t ri = 0;
+    while (si < StrsN) {
+      if (str[si] != delim) {
+        si += 1;
+      } else {
+        res[ri] = {head, (si - cur_head_i)};
+        ++ri;
+        si += 2;  // skip ", "
```
xumingkuan, line 92: What if there is no " " after ","? Can we write something like `.strip()` in Python here?

taichi/common/serialization.h
```diff
@@ -54,71 +56,93 @@ using is_unit = typename std::is_base_of<Unit, remove_cvref_t<T>>;
...
-#define TI_IO_DECL_VIRT \
-  TI_IO_DECL_INST_VIRT  \
-  template <typename S> \
-  void io(S &serializer) const
+template <typename SER, size_t N, typename T, typename... Args>
+void serialize_kv_impl(SER &ser,
+                       const std::array<std::string_view, N> &keys,
+                       T &&head,
+                       Args &&... rest) {
+  constexpr auto i = (N - sizeof...(Args));
```
xumingkuan, line 116: Looks like this is reversing the keys -- why not set `i` as a parameter or a template parameter? Or, why do the resulting strings look not reversed?

taichi/common/serialization.h
```diff
@@ -313,25 +337,36 @@ class BinarySerializer : public Serializer {
     }
   }
 
+  template <typename T>
+  void operator()(const char *, const T &val) {
+    this->process(val);
+  }
```
xumingkuan, line 343: Is this one ever used anywhere? And why doesn't it call `this->process(const char *key, const std::string &val)`?

xumingkuan (commented):
Cool!

taichi/common/serialization.h
k-ye: This is mostly handling the string created via macros (e.g., `#__VA_ARGS__`), so it should be fine.

taichi/common/serialization.h
k-ye: > Looks like this is reversing the keys 
Not really, `sizeof...(Args)` decrements after each call, so `i` is actually incrementing.

taichi/common/serialization.h
k-ye: There is no `process(const char* key, T val)` interface. Note that the binary serializers don't need the key :-)

taichi/common/serialization.h
xumingkuan: I see!

xumingkuan (approved):
LGTM, thanks!

PR #2361, opened 2021-05-25
[Example] Improve stable fluids

Scoutydren:
Related issue = #

fix boundary condition
toggle display modes with keys
simply code
add gravity for the impulse calculation
make decay in correspondence with velocity field and dye

----


k-ye (approved):
LGTM thanks!

PR #2362, opened 2021-05-25
[Misc] Add test flag when releasing taichi

Leonz5288:
Enabled release test by additional flag in build.py.
Other changes are due to python formatter.
@k-ye 

----

### DIFF
--- a/python/build.py
+++ b/python/build.py
@@ -131,12 +131,18 @@ def parse_args():
     parser.add_argument('--skip_build',
                         action='store_true',
                         help=('Skip the build process if this is enabled'))
+    parser.add_argument(
+        '--testpypi',
+        action='store_true',
+        help='Upload to test package server if this is enabled')
     return parser.parse_args()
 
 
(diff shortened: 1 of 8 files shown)


python/build.py
```diff
@@ -149,16 +155,19 @@ def main():
...
         build()
 
     if mode == 'build':
         return
     elif mode == 'upload':
-        os.system(
-            '{} -m twine upload dist/* --verbose -u yuanming-hu -p {}'.format(
-                get_python_executable(),
-                '%PYPI_PWD%' if get_os_name() == 'win' else '$PYPI_PWD'))
+        os.system('{} -m twine upload {}dist/* --verbose -u {} -p {}'.format(
```
k-ye, line 168: nit: Just create different command lines depending on whether `args.testpypi` if True. `test_pypi` itself isn't a reasonable name in every case :-)

k-ye (approved):
LGTM!

PR #2366, opened 2021-05-27
[Opt] Add error if bls analysis failed

Jack12xl:
Related issue = #2327 
Using `ti.block_local` could be hard for new Taichi Users.  So we add an error when `ti.block_local` could not infer the scratch pad range.

----


k-ye (approved):
LGTM thanks!

PR #2372, opened 2021-05-28
[metal] Add AOT module builder for metal

k-ye:
Related issue = LAN-4

----


Scoutydren (approved):
reviewed

PR #2390, opened 2021-06-01
[Opt] Optimize nested type casts

### LINKED ISSUE #2375: Performance issue: Weird performance change with type casting
**Performance Issue**
Type casting sometimes seems to take a lot of overhead but sometimes it is not.

**Additional Context**
When I was doing some interpolations on "textures", I need to use a lot type casting between `float` and `int`, then I found some "interesting" performance changes after some changes to my code that is about type casting.

**To Reproduce**
My code is a bit long(100+ lines), so I pick up the changes I've made to my code and explain a bit my findings.
Originally, my code is the below. `sample_volume_trilinear` just samples a 3D scalar `ti.field` tri-linearly, and it takes float arguments.
```python
delta=0.001
@ti.func
def get_normal(x: int, y: int, z: float):
    x = float(x)  # 1
    y = float(y)  # 1
    dx = sample_volume_trilinear(x + delta, y, z) - sample_volume_trilinear(x - delta, y, z)  # 2
    dy = sample_volume_trilinear(x, y + delta, z) - sample_volume_trilinear(x, y - delta, z)  # 2
    dz = sample_volume_trilinear(x, y, z + delta) - sample_volume_trilinear(x, y, z - delta)  # 2
    n = ti.Vector([dx, dy, dz])
    return n / n.norm()
```
With that function, I got some runtime warnings, which basically state that on lines makred `1`, I tried to store float numbers in int variables, and on lines marked `2` I tried to convert implicitly convert ints to floats. But, that was not my intention. I, in a Pythonic lazy way, intended to make `x` and `y` hold float values. 

Then, I thought, okay fine, this function must be converted to C-like code so I cannot do such lazy conversions, so I change the code to be like this.
```python
delta = 0.001
@ti.func
def get_normal(x: int, y: int, z: float):
    xf = float(x) 
    yf = float(y)
    dx = sample_volume_trilinear(xf + delta, yf, z) - sample_volume_trilinear(xf - delta, y, z)
    dy = sample_volume_trilinear(xf, yf + delta, z) - sample_volume_trilinear(xf, yf - delta, z)
    dz = sample_volume_trilinear(xf, yf, z + delta) - sample_volume_trilinear(xf, yf, z - delta)
    n = ti.Vector([dx, dy, dz])
    return n / n.norm()
```
Now, I got rid of all runtime warnings, but I got ~40% fps decrease from ~58fps to ~36fps, which is weird for me, because in the original function, the number of casting operations should be at least the same as the number of the second version, but the first one runs much faster.

For the full code, please see my simplest trial on implementing Direct Volume Rendering with Taichi [here](https://github.com/ifsheldon/Differender/blob/DVR/src/hello.py) and checkout the differences between two commits(`9ffc3945` and `9aa7b64c`)

**Additional comments**
I know that I should not have used so many casting operations because I can optimize it by prestore cast float values, so I did it in my commits to my repo and my final version is the below. 
But, I think such performance changes are weird and you should look into it.
```python
@ti.func
def get_normal(x: float, y: float, z: float): # this runs at least as fast as the original one
    dx = sample_volume_trilinear(x + delta, y, z) - sample_volume_trilinear(x - delta, y, z)
    dy = sample_volume_trilinear(x, y + delta, z) - sample_volume_trilinear(x, y - delta, z)
    dz = sample_volume_trilinear(x, y, z + delta) - sample_volume_trilinear(x, y, z - delta)
    n = ti.Vector([dx, dy, dz])
    return n / n.norm()
```

xumingkuan:
Related issue = fix #2375

Test case:
```python
import taichi as ti

ti.init(print_ir=True)


@ti.func
def low_high_frac(x: float):
    low = int(x)
    high = low + 1
    frac = x - low
    return low, high, frac


@ti.func
def sample_volume_trilinear(x: float):
    x_low, x_high, x_frac = low_high_frac(x)
    return x_frac


@ti.kernel
def get_normal(x: int):
    xf = float(x)
    dy = sample_volume_trilinear(xf)
    print(dy)


get_normal(2)
```
Before:
```
  $0 = offloaded
  body {
    <i32> $1 = arg[0]
    <f32> $2 = cast_value<f32> $1
    <i32> $3 = cast_value<i32> $2
    <f32> $4 = cast_value<f32> $3
    <f32> $5 = sub $2 $4
    print $5, "\n"
  }
```
After:
```
  $0 = offloaded  
  body {
    <f32> $1 = const [0.0]
    print $1, "\n"
  }
```

----

### DIFF
--- a/taichi/transforms/alg_simp.cpp
+++ b/taichi/transforms/alg_simp.cpp
@@ -57,21 +57,22 @@ class AlgSimp : public BasicStmtVisitor {
       return false;
     }
     if (is_real(second_cast)) {
+      // float(...(a))
       return is_real(first_cast) &&
              data_type_bits(second_cast) <= data_type_bits(first_cast);
+    }
+    if (is_integral(first_cast)) {
+      // int(int(a))
+      return data_type_bits(second_cast) <= data_type_bits(first_cast);
+    }
+    // int(float(a))
+    if (data_type_bits(second_cast) <= data_type_bits(first_cast) * 2) {
+      // f64 can hold any i32 values.
+      return true;
     } else {
-      if (is_integral(first_cast)) {
-        return data_type_bits(second_cast) <= data_type_bits(first_cast);
-      } else {
-        if (data_type_bits(second_cast) <= data_type_bits(first_cast) * 2) {
-          // f64 can hold any i32 values.
-          return true;
-        } else {
-          // Assume a floating point type can hold any integer values when
-          // fast_math=True.
-          return fast_math;
-        }
-      }
+      // Assume a floating point type can hold any integer values when
+      // fast_math=True.
+      return fast_math;
     }
   }
 
--- a/taichi/transforms/alg_simp.cpp
+++ b/taichi/transforms/alg_simp.cpp
@@ -48,6 +48,33 @@ class AlgSimp : public BasicStmtVisitor {
       : BasicStmtVisitor(), fast_math(fast_math_) {
   }
 
+  [[nodiscard]] bool is_redundant_cast(const DataType &first_cast,
+                                       const DataType &second_cast) const {
+    // Tests if second_cast(first_cast(a)) is guaranteed to be equivalent to
+    // second_cast(a).
+    if (!first_cast->is<PrimitiveType>() || !second_cast->is<PrimitiveType>()) {
+      // TODO(type): handle this case
+      return false;
+    }
+    if (is_real(second_cast)) {
+      return is_real(first_cast) &&
+             data_type_bits(second_cast) <= data_type_bits(first_cast);
+    } else {
+      if (is_integral(first_cast)) {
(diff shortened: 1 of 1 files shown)


taichi/transforms/alg_simp.cpp
```diff
@@ -48,6 +48,33 @@ class AlgSimp : public BasicStmtVisitor {
...
+    // Tests if second_cast(first_cast(a)) is guaranteed to be equivalent to
+    // second_cast(a).
+    if (!first_cast->is<PrimitiveType>() || !second_cast->is<PrimitiveType>()) {
+      // TODO(type): handle this case
+      return false;
+    }
+    if (is_real(second_cast)) {
+      return is_real(first_cast) &&
+             data_type_bits(second_cast) <= data_type_bits(first_cast);
+    } else {
```
k-ye, line 62: nit: remove these `else`s...

k-ye (approved):
LGTM + nits!

Issue #2369, opened 2021-05-27
Fix CI auto formatter


k-ye:
**Concisely describe the proposed feature**

Trigger heavy CI tests only when specific keyworks are used in a PR

**Describe the solution you'd like (if any)**

Use existing Github Actions to detect specific keyword

k-ye:
We can now trigger auto format via `/format`

xumingkuan:
https://github.com/taichi-dev/taichi/pull/2438/commits/65e8ec4edeacc9eeb1b779595a09636b479b5b6c looks somewhat correct but horrible...

PR #2394, opened 2021-06-02
[CI] Recovered format check, add dependancy

Leonz5288:
Related issue = #2369

----

### DIFF
--- a/.github/workflows/persubmit.yml
+++ b/.github/workflows/persubmit.yml
@@ -19,7 +19,6 @@ jobs:
           git config user.email "taichigardener@gmail.com"
           git config user.name "Taichi Gardener"
           git checkout -b _fake_squash
-          git commit -am "compare code format"
           git remote add upstream https://github.com/taichi-dev/taichi.git
           git fetch upstream master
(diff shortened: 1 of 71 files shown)


.github/workflows/persubmit.yml
```diff
@@ -138,6 +169,7 @@ jobs:
 
   check_previous_run:
     name: Checks the Workflow Run of the Previous Commit
+    needs: check_code_format
```
k-ye, line 172: Actually, we don't need this `needs` here

.github/workflows/persubmit.yml
```diff
@@ -4,9 +4,38 @@ on:
...
+    # This job will be required to pass before merging to master branch.
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Check code format
+        run: |
+          git fetch https://github.com/taichi-dev/taichi.git refs/heads/master
```
k-ye, line 19: Simplify?

Issue #2381, opened 2021-05-29
mass_spring_3d.py should exit on import error


peterjpxie:
**Describe the bug**
On import error, it should sys.exit(0) instead of continuing.

```py 
try:
    import taichi_glsl as tl
    import taichi_three as t3
except ImportError:
    print('This example needs the extension library Taichi THREE to work.'
          'Please run `pip install taichi_three==0.0.5` to install it.')
   # [To add] sys.exit(0)
if t3.__version__ != (0, 0, 5):
```

**To Reproduce**
pip uninstall taichi_three
python examples\mass_spring_3d.py

**Log/Screenshots**
```
[Taichi] mode=release
[Taichi] version 0.7.20, llvm 10.0.0, commit 284f75ed, win, python 3.9.3
This example needs the extension library Taichi THREE to work.Please run `pip install taichi_three==0.0.5` to install it.
Traceback (most recent call last):
  File "c:\Users\xieji\AppData\Roaming\Python\Python39\site-packages\taichi\examples\mass_spring_3d.py", line 9, in <module>
    if t3.__version__ != (0, 0, 5):
NameError: name 't3' is not defined
```

PR #2400, opened 2021-06-03
[Misc] Fixed that mass_spring_3d.py do not exit on import error

Leonz5288:
Related issue = #2381

----

### DIFF
--- a/examples/mass_spring_3d.py
+++ b/examples/mass_spring_3d.py
@@ -1,5 +1,5 @@
 import numpy as np
-
+import sys
 import taichi as ti
 
 try:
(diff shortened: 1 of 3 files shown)


k-ye (approved):

peterjpxie:
sys.exit() is a better than exit().  
sys.exit(): Use in a script
exit(): Should use only in interactive python, i.e. when you run python without script file argument.
Reference: https://www.geeksforgeeks.org/python-exit-commands-quit-exit-sys-exit-and-os-_exit/

Leonz5288:
@peterjpxie Thanks for the information!

PR #2402, opened 2021-06-04
[Metal] Make file names dumped from aot configurable

Scoutydren:
Related issue = #
Make file names dumped from aot configurable

----


k-ye (approved):
Thanks LGTM!

PR #2403, opened 2021-06-04
[Example] Add example test_reduction.py

Frenchfries135:
Related issue = #

----


k-ye:
Pls sign the CLA. Also could you add a few comments to your methods? :-)

PR #2382, opened 2021-05-30
[lang] Add tests for refine_coordinates

k-ye:
Related issue = #2327

A follow-up of #2327. Add tests for the JIT generated `refine_coordinates()`. Hopefully this servers as a better *documentation* for how the list-gen process works.

[[Click here for the format server]](http://kun.csail.mit.edu:31415/)

----

PR #2407, opened 2021-06-06
[ir] Make lower_scalar_pointer testable

k-ye:
Related issue = #2382

`lower_scalar_pointer()` used to do three difference things:

1. transform the indices into a series of extract_bits and linearized statements 
2. test if the SNode is linear related to the struct_for
3. generate SNode access ops (e.g. GetCh)

This PR pulls out step 1 into a separate class `ScalarPointerLowerer` to make it testable. Later we can also grow the logic around 1 separately (e.g. generate index decomposition statements for non power-of-two SNodes).

Steps 2 and 3 still remain in the original `lower_acesss` pass.

----

### DIFF
--- a/taichi/analysis/arithmetic_interpretor.h
+++ b/taichi/analysis/arithmetic_interpretor.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <optional>
+#include <unordered_set>
 #include <unordered_map>
 
 #include "taichi/ir/statements.h"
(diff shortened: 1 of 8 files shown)


taichi/analysis/arithmetic_interpretor.h
```diff
@@ -33,8 +46,30 @@ class ArithmeticInterpretor {
       return itr->second;
     }
 
+    /**
+     * Tells the interpretor to ignore statement @param s.
+     *
+     * This is effective only for statements that are not supported by
+     * ArithmeticInterpretor.
+     *
+     * @param s: Statemet to ignore
```
xumingkuan, line 55:
```suggestion
     * @param s: Statement to ignore
```

xumingkuan (approved):
LGTM!

PR #2412, opened 2021-06-08
[ir] Support BLS in control-flow graph and add docstrings in build_cfg

xumingkuan:
Related issue = #2193 

Hopefully these comments can help others understand how a control-flow graph is built.

----

### DIFF
--- a/taichi/analysis/build_cfg.cpp
+++ b/taichi/analysis/build_cfg.cpp
@@ -73,6 +73,8 @@ class CFGBuilder : public IRVisitor {
         /*is_parallel_executed=*/in_parallel_for,
         /*prev_node_in_same_block=*/last_node_in_current_block);
     for (auto &prev_node : prev_nodes) {
+      // Now that the "(next node)" is created, we should insert edges
+      // "node... -> (next node)" here.
       CFGNode::add_edge(prev_node, node);
     }
     prev_nodes.clear();
@@ -93,6 +95,8 @@ class CFGBuilder : public IRVisitor {
    *     ...
    *   }
    * }
+   *
+   * Note that the edges are inserted in visit_loop().
    */
   void visit(ContinueStmt *stmt) override {
     // Don't put ContinueStmt in any CFGNodes.
@@ -111,6 +115,8 @@ class CFGBuilder : public IRVisitor {
    *     ...
    *   }
    * }
+   *
+   * Note that the edges are inserted in visit_loop().
    */
   void visit(WhileControlStmt *stmt) override {
     // Don't put WhileControlStmt in any CFGNodes.
--- a/taichi/analysis/build_cfg.cpp
+++ b/taichi/analysis/build_cfg.cpp
@@ -4,7 +4,24 @@
 
 TLANG_NAMESPACE_BEGIN
 
-// Build a control-flow graph
+/**
+ * Build a control-flow graph. The resulting graph is guaranteed to have an
+ * empty start node and an empty final node.
+ *
+ * In the following docstrings, node... means a CFGNode's corresponding
+ * statements in the CHI IR. Other blocks are just Blocks in the CHI IR.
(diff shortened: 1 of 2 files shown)


taichi/analysis/build_cfg.cpp
```diff
@@ -52,18 +64,69 @@ class CFGBuilder : public IRVisitor {
     return node;
   }
 
+  /**
+   * Structure:
+   *
+   * block {
+   *   node {
+   *     ...
+   *   } -> (node_loop_begin), (node_loop_end);
```
k-ye, line 73: What is the relationship between `node`, `node_loop_begin` and `node_loop_begin`? Also, why are the later two quoted with `()`? (Is it that they are not in the graph?)

k-ye (commented):
Can you also write up a doc to summarize how CFG works, what kind of optimization it does, etc? Thanks!

taichi/analysis/build_cfg.cpp
xumingkuan: The nodes with `()` are not handled in this function (`void visit(ContinueStmt *stmt)` in this case). See the definition of `node_loop_begin` and `node_loop_end` in the docstring of `void visit_loop(Block *body, CFGNode *before_loop, bool is_while_true)`.

The node `(next node)` is not yet created, and probably doesn't exist. `(next node)` just shows that the next node will be at that place in the block.

xumingkuan: Oh, that shouldn't be `(node_loop_end)`...

Also, `node_loop_begin` is already created, so maybe it's clearer to remove the parentheses here.

k-ye: Ah OK.

> The nodes with () are not handled in this function

Maybe put such info to the beginning of the documentation. It will also be great to explain how to interpret the "structures", as well as giving a short example :-)

For example, it's not really obvious what this means:

>  -> node_loop_begin, (the next node after the loop);

I guess it means `continue` brings the control flow back to `node_loop_begin`, but how exactly?

xumingkuan: Right, but I think maybe it's too long to put them in the docstring. 

> I guess it means `continue` brings the control flow back to `node_loop_begin`, but how exactly?

We have a vector `continues_in_current_loop`, and we insert the control-flow edges in `visit_loop()`.

k-ye: >  I think maybe it's too long to put them in the docstring.

Which part would be too long? CFG is a rather important piece in the IR optimization passes, so it would make sense to document this part well. In addition, it's not the full CFG algorithm that we need to document, just some words to explain how the structure should be interpreted:

As an example:

* `node_a -> node_b, node_c`: does this mean `a` connects to both `b` and `c`?
* `node -> node_loop_begin, (next node) if !is_while_true;`: Does this mean `a` connects to either `node_loop_begin` or `(next node)`; or `a` always connects to `node_loop_begin`, and could connect to `(next node)` if `!is_while_true`?

xumingkuan: I see! I'll add some more documentations.

taichi/analysis/build_cfg.cpp
```diff
@@ -4,7 +4,24 @@
...
+ * In the following docstrings, node... means a CFGNode's corresponding
+ * statements in the CHI IR. Other blocks are just Blocks in the CHI IR.
+ * Nodes denoted with "()" mean not yet created when visiting the Stmt/Block.
+ *
+ * Structures like
+ * node_a {
+ *   ...
+ * } -> node_b, node_c;
+ * means node_a has edges to node_b and node_c, or equivalently, node_b and
+ * node_c appear in the |next| field of node_a.
```
k-ye, line 20: Also explain `[]`?

k-ye (approved):
Thanks LGTM!

PR #2413, opened 2021-06-08
[Bug] Fixed internal lagrangian forces

saullocastro:
Related issue = #

----


k-ye (approved):
Thanks!

Issue #2417, opened 2021-06-09
libtinfo6 required when installing from pip (v. 0.7.21)


marekjg:
**Describe the bug**
When installing default version on pip (0.7.21), libtaichi_core.so is linked with libtinfo6 not libtinfo5 ausing problems on systems with libtinfo5 only. 

**To Reproduce**
Install taichi on ubuntu 18.04 and try to load the library. 

**Log/Screenshots**
```
ldd libtaichi_core.so
	linux-vdso.so.1 (0x00007fff74ff5000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fb673241000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fb6730f2000)
	libX11.so.6 => /lib/x86_64-linux-gnu/libX11.so.6 (0x00007fb672fb5000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fb672f99000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fb672f93000)
	libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007fb672f63000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fb672f3e000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fb672d5d000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb672b6b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fb6766bb000)
	libxcb.so.1 => /lib/x86_64-linux-gnu/libxcb.so.1 (0x00007fb672b41000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fb672b26000)
	libXau.so.6 => /lib/x86_64-linux-gnu/libXau.so.6 (0x00007fb672b1e000)
	libXdmcp.so.6 => /lib/x86_64-linux-gnu/libXdmcp.so.6 (0x00007fb672b16000)
	libbsd.so.0 => /lib/x86_64-linux-gnu/libbsd.so.0 (0x00007fb672afc000)
```

k-ye:
* A hack fix: https://github.com/taichi-dev/taichi/issues/2347
[...]

PR #2419, opened 2021-06-10
[Doc] Changed llvm-10 link and installation instruction

Leonz5288:
Related issue = #2417
The download link of llvm-10 in docs has changed to the version without libtinfo.
LLVM installation instruction also changed to disable libtinfo.


k-ye (approved):
lgtm!

PR #2421, opened 2021-06-11
[Example] Added mpm99 example with cubic interpolation

saullocastro:
Related issue = #

### DIFF
--- a/examples/mpm99_cubic.py
+++ b/examples/mpm99_cubic.py
@@ -1,0 +1,99 @@
+import taichi as ti
+import numpy as np
+ti.init(arch=ti.gpu) # Try to run on GPU
+quality = 1 # Use a larger value for higher-res simulations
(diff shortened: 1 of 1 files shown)


k-ye:
Thanks for doing this! But I'd suggest to just keep the original example (we're also trying to reduce the number of examples so that it's more friendly to new users)

saullocastro:
@k-ye it's done. I'm not sure if I agree with the decision to remove the examples, because they provide so much input on how Taichi works and how different physical problems can be solved with Taichi. With this added example, it will be easy for someone to use the cubic interpolation and compare with the quadratic one.

k-ye:
Sorry, I didn't mean to remove all the examples, just to control how many of them. There are a few examples that are over-complicated at this point, which no longer serve as the tutorial purpose. Could you consider moving this one to https://github.com/taichi-dev/advanced_examples ? Thanks!

PR #2422, opened 2021-06-11
[IR] Fully deprecate trailing_bits

KLozes:
Related issue = #2177


k-ye (approved):
Thanks! I think you also need to fix `alias_analysis_test.cpp`

KLozes:
> Thanks! I think you also need to fix alias_analysis_test.cpp
Good catch, thanks!

Issue #2418, opened 2021-06-10
[RFC] Support adding new fields dynamically


k-ye:
**Concisely describe the proposed feature**

Right now all the `ti.field`s must be defined first before the first `@ti.kernel` call. This is not quite flexible, so we should consider removing this limitation. Ideally we should be able to run the following program:

```py
x = ti.field(...)

@ti.kernel
def uses_x():
  # uses `x`

y = ti.field(...)

@ti.kernel
def uses_y():
  # uses `y`
```

On top of that, we also need to make sure the change decouples the compiled kernel from the actual SNodes. That is, for two SNodes of the same topology, the generated kernel shouldn't really care which SNode it binds to.

**Describe the solution you'd like (if any)**

Note that SNodes immediately under `ti.root` are completely isolated from each other. They don't necessarily have to reside inside the same memory buffer. Therefore, we can introduce a concept called `SNodeTree`, which is equivalent to a `ti.root`. Effectively, this means there can be multiple roots (`SNodeTree`s) in a Taichi program, with each `SNodeTree` stored in its own chunk of memory.

Currently, all the SNodes are JIT compiled *implicitly* upon the first invocation of a `ti.kernel`. We need to control when JIT happens more precisely, and to allow compiling SNodes more than once. One possible API is to provide a `ti.fields_builder`:

```py
x = ti.field(...)
y = ti.field(...)

fb = ti.fields_builder()
block = fb.pointer(...)
block.dense(...).place(x)
block.dense(...).place(y)
fb.finalize()  # JIT compiles the SNodeTree, with `x` and `y` placed
```

We need a new IR pass to analyze the kernel and to find out all the SNodes being used. From that it would then find out the `SNodeTree`s these SNodes belonging to. This also means that `GetRoot` will be replaced by (extended to) `GetSNodeTree(tree_id)`. Below shows an example:

```cpp
// kernel "foo" {
//   $0 = offloaded  
//   body {
//     $1 = get_snode_tree id=0
//     $2 = get_ch $1, 1
//     ...
//   }
// }

// pseudo code for the generated kernel ``foo``
void foo(Context* ctx, SNodeTree* trees) {
  auto tree = trees[0];
  auto ch = tree.get_ch(1);
  // ...
}
```

Later when `foo` gets launched, by binding to different `SNodeTree` instances **of the same topology**, `foo` is decoupled from the actual taichi field.

[...]

PR #2424, opened 2021-06-14
[Lang] [refactor] Separate runtime and root initialization

k-ye:
Related issue = #2418

This PR refactors `Program` so that the LLVM runtime and SNode related initialization are separated into two functions: `initialize_llvm_runtime_system()` and `initialize_llvm_runtime_snodes()`. In the future, we should only call `initialize_llvm_runtime_system()` once globally, but call `initialize_llvm_runtime_snodes()` for each new SNode tree.

Several notes:

* PTX `.entry` vs `.func`: CUDA kernels are marked as `.entry` + `__glb__` in the PTX module. If not properly marked, `cuModuleGetFunction` won't be able to find this kernel. This is done around https://github.com/taichi-dev/taichi/blob/de3b2262de944cbf04e41fe0679e7e18eb242099/taichi/llvm/llvm_context.cpp#L544-L559
* Unified memory: #2427 
 
Also fully deprecate `ti.layout()` (FYI @yuanming-hu ) .

### DIFF
--- a/examples/chi_examples/main.cpp
+++ b/examples/chi_examples/main.cpp
@@ -51,7 +51,7 @@ void run_snode() {
   auto *place = &pointer->insert_children(SNodeType::place);
   place->dt = PrimitiveType::i32;
 
-  program.materialize_layout();
+  program.materialize_runtime();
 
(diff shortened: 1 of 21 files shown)


Leonz5288 (approved):
L G T M !!!

PR #2426, opened 2021-06-15
[Lang] Add get_addr() to retrieve the buffer addr of an SNode

AmesingFlank:
Add get_addr() to retrieve the pointer of an SNode. This only works on LLVM backends for now.


k-ye (commented):
I think we need to document this PR more clearly -- Originally I thought that `get_addr` was intended to be used on the host side. Also there seems to be a few unrelated examples, such as `demo_sparse*`, `mpm88_sparse`, `performance_cacheline`.

examples/print_offset.py
```diff
@@ -0,0 +1,41 @@
+import taichi as ti
+
+ti.init(arch=ti.cpu, print_ir=True)
+
+n = 4
+m = 8
+
+# a = ti.field(dtype=ti.u8, shape=N)
+a = ti.field(dtype=ti.i32)
+# ti.root.dense(ti.ij, (4, 8)).place(a)
```
k-ye, line 10: Clean up?

AmesingFlank: Sure!

AmesingFlank:
> I think we need to document this PR more clearly -- Originally I thought that `get_addr` was intended to be used on the host side. Also there seems to be a few unrelated examples, such as `demo_sparse*`, `mpm88_sparse`, `performance_cacheline`.

Sure. I'll remove add some comments and remove the unrelated examples from this PR.

python/taichi/lang/ops.py
```diff
@@ -579,3 +579,22 @@ def rescale_index(a, b, I):
         if a.shape[n] < b.shape[n]:
             Ib.entries[n] = I.entries[n] * (b.shape[n] // a.shape[n])
     return Ib
+
+
+def get_addr(f, indices):
```
k-ye, line 584: Sorry my bad, we actually follow the Google docstring style... See https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

AmesingFlank: Understood. will fix.

python/taichi/lang/ops.py
```diff
@@ -579,3 +579,17 @@ def rescale_index(a, b, I):
...
+    """Returns the address (for CUDA/x64) for index 'indices' of field 'f'. Currently, this function can only be called inside a kernel.
```
k-ye, line 585: super nits: indices and f should be quoted with a single backquote "`"

k-ye (approved):
Awesome LGTM!

python/taichi/lang/ops.py
AmesingFlank: Ah right. will fix.

Issue #2428, opened 2021-06-15
Latest Appveyor broke Windows CI


k-ye:
This started in https://ci.appveyor.com/project/yuanming-hu/taichi/builds/39600030, where the latest windows image has been upgraded to use `MSVC 14.29.30037` (https://www.appveyor.com/updates/). This requires `clang11`.

```sh
[Taichi] mode=development
[Taichi] <dev mode>, llvm 10.0.0, commit 93c97796, win, python 3.6.8
%PYTHON% examples/laplace.py
In file included from C:\taichi/taichi/runtime/llvm/runtime.cpp:6:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\atomic:9:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\yvals.h:9:
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\yvals_core.h:537:2: error: STL1000: Unexpected compiler version, expected Clang 11.0.0 or newer.
#error STL1000: Unexpected compiler version, expected Clang 11.0.0 or newer.
 ^
1 error generated.
[E 06/15/21 02:02:52.091] [llvm_context.cpp:taichi::lang::compile_runtime_bitcode@204] LLVMRuntime compilation failed.
***********************************
```

The succeeded build proceeding this one (https://ci.appveyor.com/project/yuanming-hu/taichi/builds/39584809) has `MSVC 14.28.29910`.

k-ye:
Now that #2436 is in, I've disabled appveyor for `pull_request` events

PR #2432, opened 2021-06-16
[CI] Fixed windows build & test

Leonz5288:
Related issue = #2428

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -161,6 +161,57 @@ jobs:
           ti diagnose
           ti test -vr2 -t2
 
+  build_and_test_windows:
+    name: Build and Test (Windows)
+    needs: check_code_format
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        include:
+          python-version: ['3.6', '3,7', '3,8', '3.9']
+    steps:
+      - name: Install 7Zip PowerShell
+        shell: powershell
+        run: Install-Module 7Zip4PowerShell -Force -Verbose
+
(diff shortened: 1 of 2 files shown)


taichi/runtime/llvm/runtime.cpp
```diff
@@ -2,6 +2,10 @@
 // The generated bitcode will likely get inlined for performance.
 
 #if !defined(TI_INCLUDED) || !defined(_WIN32)
+// The latest MSVC uses llvm-11 as requirements. However, we use llvm-10
```
k-ye, line 5: A few nits:

* Add "NOTE(#2428)"
* Let's be explicit about "latest": Visual Studio 2019 version 16.10.1 (MSVC 14.29.30037)
* Add this link: https://github.com/microsoft/STL/blob/1866b848f0175c3361a916680a4318e7f0cc5482/stl/inc/yvals_core.h#L550-L561
* guard it with `#if defined(_WIN32) || defined(_WIN64)`

.github/workflows/presubmit.yml
```diff
@@ -161,6 +161,56 @@ jobs:
           ti diagnose
           ti test -vr2 -t2
 
+  build_and_test_windows:
+    name: Build and Test (Windows)
+    needs: check_code_format
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        python-version: ['3.6', '3.7', '3.8', '3.9']
```
k-ye, line 170: Let's just use 3.6 for speed consideration.

PR #2441, opened 2021-06-19
[Metal] Support pointer SNode on Metal

k-ye:
Related issue =  #1174


k-ye:
Hopefully this is in a stable enough point now:

* [x] All tests involving `pointer` have passed on metal
* [x] `mgpcg.py`, `mgpcg_advanced.py` produced the same result as CPU
* [x] Got a `small_particle_renderer.py` working. The original one took too much memory. Because of the 32-bit data-type limitation, right now the Metal backend only supports a maximum of ~4GB memory pool.
 
[small_particle_renderer.txt](https://github.com/taichi-dev/taichi/files/6734302/small_particle_renderer.txt)

PR #2450, opened 2021-06-22
[GUI] Win 32 DPI awareness stuff

bobcao3:
Set process DPI awareness when detecting system version less than 6.3 (Windows 8.1)

Related #2320


k-ye (approved):
Thanks for the prompt fix!

Issue #2439, opened 2021-06-19
taichi version 0.7.21 requires libm.so.6: version `GLIBC_2.29' which is unavailable on Ubuntu 18.04.5


Rashfu:
After installing latest taichi through `pip install --upgrade taichi` on my Ubuntu 18.04 machine, I got the following error in 'import taichi_core as core':

```
ImportError: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by  home/dell/anaconda3/envs/physics/lib/python3.8/site-packages/taichi/core/../lib/taichi_core.so)
Consider installing libtinfo5: sudo apt-get install libtinfo5
```
But I have already installed libtinfo5.
I use `strings /lib/x86_64-linux-gnu/libm.so.6 | grep GLIBC_` and apparently Ubuntu 18.04 supports glibc up to version 2.27.

Reverting to the previous version `pip install taichi==0.7.16` fixed the issue.

k-ye:
Thanks for reporting this. `libtinfo` will be removed in the release next week.

dhl2011dj:
I meet this error also. I updated the glibc to 2.29 on my ubuntu 18.04 but found no help, and the update will cause problems on the ubuntu 18.04.

k-ye:
Hmm, sorry I misread the issue. Right now we only have a 20.04 buildbot. Let us investigate some docker solution.

cloudhan:
@k-ye I met the same issue.  
`strings /lib/x86_64-linux-gnu/libc.so.6 | grep GLIBC_` gives
```
GLIBC_2.2.5
GLIBC_2.2.6
GLIBC_2.3
GLIBC_2.3.2
GLIBC_2.3.3
GLIBC_2.3.4
GLIBC_2.4
GLIBC_2.5
GLIBC_2.6
GLIBC_2.7
GLIBC_2.8
GLIBC_2.9
GLIBC_2.10
GLIBC_2.11
GLIBC_2.12
GLIBC_2.13
GLIBC_2.14
GLIBC_2.15
GLIBC_2.16
GLIBC_2.17
GLIBC_2.18
GLIBC_2.22
GLIBC_2.23
GLIBC_2.24
GLIBC_2.25
GLIBC_2.26
GLIBC_2.27
GLIBC_PRIVATE
```
so the pip package cannot work on my machine. To my knowledge,  you should follow PEP 513/571/599 to build you package.

dhl2011dj:
[...]

PR #2457, opened 2021-06-23
[ci] [docker] Build release with docker in ubuntu18.04

Leonz5288:
Related issue = #2439

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,11 @@
 # Taichi Dockerfile for development
-FROM nvidia/cuda:10.0-devel-ubuntu18.04
-
+ARG UBUNTU
+FROM nvidia/cuda:${UBUNTU}
+ARG PYTHON
+ARG TEST_OPTION
+ARG PYPI_PWD
+ENV PYPI_PWD=$PYPI_PWD
+ENV DEBIAN_FRONTEND=noninteractive
 LABEL maintainer="https://github.com/taichi-dev"
 
 # This installs Python 3.6.9 by default. Once the
@@ -8,9 +13,11 @@ LABEL maintainer="https://github.com/taichi-dev"
 # install Python 3.8 by default
 RUN apt-get update && \
     apt-get install -y software-properties-common \
+                       $PYTHON \
                        python3-pip \
+                       ${PYTHON}-dev\
                        libtinfo-dev \
-                       clang-8 \
+                       clang-10 \
                        wget \
                        git \
                        libx11-dev \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,16 +1,23 @@
 # Taichi Dockerfile for development
-FROM nvidia/cuda:10.0-devel-ubuntu18.04
-
+ARG UBUNTU
```
k-ye, line 2: Let's create a separate Dockerfile for CI purpose. This one is used for building an image with taichi, IIUC. @rexwangcc 

That said, it seems we can create a script so that the two Dockerfiles can share lots of stuff

Dockerfile
```diff
@@ -21,45 +28,37 @@ RUN apt-get update && \
...
+RUN unzip taichi-llvm-10.0.0-linux.zip
+ENV PATH="/taichi-llvm-10.0.0-linux/bin:$PATH"
 
 # Install Taichi from source
+ENV CC="clang-10"
+ENV CXX="clang++-10"
 WORKDIR /taichi-dev
 RUN git clone https://github.com/taichi-dev/taichi --depth=1 --branch=master
+# Install Taichi's Python dependencies
+RUN $PYTHON -m pip install --user -r taichi/requirements.txt
```
k-ye, line 55: Do we need to update `requirements.txt` :) ?

Jenkinsfile
```diff
@@ -5,7 +5,7 @@ pipeline {
         PATH = "/opt/taichi-llvm-10.0.0/bin:/usr/local/cuda/bin/:$PATH"
         CC = "clang-10"
         CXX = "clang++-10"
-        PYTHON_EXECUTABLE = "python3"
+        // PYTHON_EXECUTABLE = "python3"
```
k-ye, line 8:
```suggestion
```

k-ye (approved):

ci/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+ARG PYTHON
+ARG TEST_OPTION
+ARG PYPI_PWD
+ENV PYPI_PWD=$PYPI_PWD
+ENV DEBIAN_FRONTEND=noninteractive
+LABEL maintainer="https://github.com/taichi-dev"
+
+# This installs Python 3.6.9 by default. Once the
+# docker image is upgraded to Ubuntu 20.04 this will
+# install Python 3.8 by default
```
rexwangcc, line 13:
```suggestion
```

ci/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+WORKDIR /taichi-dev
+RUN git clone https://github.com/taichi-dev/taichi --depth=1 --branch=master
+# Install Taichi's Python dependencies
+RUN $PYTHON -m pip install --user -r taichi/requirements.txt
+# Build Taichi from source
+RUN cd taichi && \
+    git submodule update --init --recursive --depth=1 && \
+    mkdir build
+WORKDIR /taichi-dev/taichi/build
+RUN cmake .. -DPYTHON_EXECUTABLE=$PYTHON -DCUDA_VERSION='10.0' -DTI_WITH_CUDA:BOOL=True -DTI_WITH_CC:BOOL=ON -DCMAKE_CXX_COMPILER=$CXX
```
rexwangcc, line 61: `-DCUDA_VERSION='10.0'`is related to the docker tag, maybe parameterize it too?

rexwangcc (commented):
LGTM in general and this is a great start towards containerization! I left one comment.

In the future, we could potentially also:
- tune the LLVM version with docker.
- simplify by let images share the same scripts (as Ye mentioned)
- extract all docker manifests into a separate repo just like https://github.com/nodejs/docker-node or https://github.com/docker-library/redis, and separate runtime images from the CI images.

ci/Dockerfile
Leonz5288: I believe this is just the hacked version we pass in order to avoid build errors.

k-ye: Right, we kind of have to pass in 10.0 for `-DCUDA_VERSION` (even if the actual CUDA version is 11.0) :-/

rexwangcc: Ah gotcha, thanks for the explanation!

rexwangcc (approved):

PR #2473, opened 2021-06-28
[misc] Add checking of gitpython in cmake

xndcn:
generate_commit_hash.py relies on `gitpython` to generate commit hash,
so we should check it in cmake and show the way to install it.


k-ye (approved):
LGTM, thanks!

frostming:
Hi, I can't find generate_commit_hash.sh in the current codebase, so can we remove the gitpython check in cmake? this will remove gitpython from the build dependencies.

PR #2479, opened 2021-06-30
[opengl] Dump compute shader source code when print_kernel_llvm_ir is true

xndcn:
Dump compute shader source code when print_kernel_llvm_ir is true


archibate (approved):
great! so prinr_llvm_ir actually should be print_backend_code..

k-ye (approved):
Thanks!

PR #2486, opened 2021-07-02
[opengl] [refactor] Use macro to generate atomic float functions

xndcn:
So we don't have to maintain 3 similar shader header files any more.


k-ye (commented):
Awesome!

Since you are helping clean up this part, could you take a step further and define a `constexpr char *` in the shader files to hold the string constants? E.g.

https://github.com/taichi-dev/taichi/blob/fb8a0b405bbc53d98e1e16853ff02487b59ff92a/taichi/backends/metal/shaders/runtime_structs.metal.h#L6-L30

Or in a simplified version:

```cpp
#define METAL_BEGIN_RUNTIME_STRUCTS_DEF \ 
   constexpr auto kMetalRuntimeStructsSourceCode = 
METAL_BEGIN_RUNTIME_STRUCTS_DEF
STR(
// actual code
)
;
```

This way we can stop using the weird `kernel_header += (#include "foo")` syntax. Thanks!

xndcn:
@k-ye Thanks! I'll try this later today.

k-ye (approved):
LGTM thanks!

PR #2490, opened 2021-07-05
[Refactor] [cuda] Cleanup CUDA AtomicOpStmt codegen

AmesingFlank:
Avoid redundant if-elses in `visit(AtomicOpStmt *stmt)` using a look-up table.

### DIFF
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -246,9 +246,7 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
     if (!stmt->is_reduction) {
       return nullptr;
     }
-    if (!stmt->val->ret_type->is<PrimitiveType>()) {
-      return nullptr;
-    }
+    TI_ASSERT(stmt->val->ret_type->is<PrimitiveType>());
     PrimitiveTypeID prim_type =
         stmt->val->ret_type->cast<PrimitiveType>()->type;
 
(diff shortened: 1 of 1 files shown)


taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -240,6 +240,48 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
 #undef UNARY_STD
   }
 
+  // Not all reduction statements can be optimized.
+  // If the operation cannot be optimized, this function returns nullptr.
+  llvm::Value *optimized_reduction(AtomicOpStmt *stmt) {
+    if (!stmt->is_reduction) {
+      return nullptr;
+    }
+    if (!stmt->val->ret_type->is<PrimitiveType>()) {
```
k-ye, line 249: nit: I believe this deserves a `TI_ASSERT`, because if this is not a primitive, it's a bug in our code.

taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -240,6 +240,48 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
...
+    fast_reductions[PrimitiveTypeID::i32][AtomicOpType::bit_or] =
+        "reduce_or_i32";
+    fast_reductions[PrimitiveTypeID::i32][AtomicOpType::bit_xor] =
+        "reduce_xor_i32";
+
+    AtomicOpType op = stmt->op_type;
+    if (fast_reductions.find(prim_type) == fast_reductions.end()) {
+      return nullptr;
+    }
+    if (fast_reductions.at(prim_type).find(op) ==
```
k-ye, line 277: Similarly, if `op` is not in {add, min, max}, it will be a bug in our code. So we can just remove this?

k-ye (commented):
Awesome! I think we can take a further step here:

1. For CustomInt/FloatType, we can create a similar helper, which only handles `add`
2. For float, we handle `add`, `min` and `max`
3. We handle all sorts of opts for integral types.

taichi/backends/cuda/codegen_cuda.cpp
AmesingFlank: Sure. That makes sense.

taichi/backends/cuda/codegen_cuda.cpp
AmesingFlank: Yes, I think we can replace it with a TI_ASSERT for now.

k-ye: Thanks. But it would still be a bug in our code if we generate a pair of `f32/xor`

AmesingFlank: Shall I do a `TI_ASSERT(fast_reductions.at(prim_type).find(op) !=fast_reductions.at(prim_type).end())` then?

k-ye: Sounds good!

k-ye (approved):
LGTM. Thanks for doing this refactor!

taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -240,6 +240,109 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
...
+      return nullptr;
+    }
+    TI_ASSERT(fast_reductions.at(prim_type).find(op) !=
+              fast_reductions.at(prim_type).end());
+    return create_call(fast_reductions.at(prim_type).at(op),
+                       {llvm_val[stmt->dest], llvm_val[stmt->val]});
+  }
+
+  llvm::Value *custom_type_atomic(AtomicOpStmt *stmt) {
+    if (stmt->op_type == AtomicOpType::add) {
```
k-ye, line 282: A few more nits: I'd prefer early returns, e.g. `if (!stmt->op_type == AtomicOpType::add) return;`

taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -240,6 +240,109 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
...
+        return atomic_add_custom_int(stmt, cit);
+      } else if (auto cft = dst_type->cast<CustomFloatType>()) {
+        return atomic_add_custom_float(stmt, cft);
+      }
+    }
+    return nullptr;
+  }
+
+  llvm::Value *integral_type_atomic(AtomicOpStmt *stmt) {
+    if (is_integral(stmt->val->ret_type)) {
```
k-ye, line 295: ditto

taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -240,6 +240,109 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
...
+  llvm::Value *real_type_atomic(AtomicOpStmt *stmt) {
+    if (!stmt->val->ret_type->is<PrimitiveType>()) {
+      return nullptr;
+    }
+    AtomicOpType op = stmt->op_type;
+    if (is_real(stmt->val->ret_type) && op == AtomicOpType::add) {
+      return builder->CreateAtomicRMW(llvm::AtomicRMWInst::FAdd,
+                                      llvm_val[stmt->dest], llvm_val[stmt->val],
+                                      AtomicOrdering::SequentiallyConsistent);
+    } else {
```
k-ye, line 321: no need for the `else ` branch?

k-ye (approved):

PR #2493, opened 2021-07-06
[lang] Support fake FieldsBuilder() with same memory location

ljcc0930:
Related issue = #2418


python/taichi/snode/fields_builder.py
```diff
@@ -110,3 +110,6 @@ def finalize(self):
     def _check_not_finalized(self):
         if self._finalized:
             raise InvalidOperationError('FieldsBuilder finalized')
+
+    def get_finalized(self):
```
k-ye, line 114:
```suggestion
    @property
    def finalized(self):
```

taichi/program/program.cpp
```diff
@@ -428,11 +428,12 @@ void Program::initialize_llvm_runtime_snodes(const SNodeTree *tree,
 }
 
 int Program::add_snode_tree(std::unique_ptr<SNode> root) {
+  /*
```
k-ye, line 431: Just delete this?

taichi/struct/struct.h
```diff
@@ -31,7 +31,9 @@ class StructCompiler {
 
   virtual void run(SNode &node) = 0;
 
-  static std::unique_ptr<StructCompiler> make(Program *prog, Arch arch);
+  static std::unique_ptr<StructCompiler> make(Program *prog, Arch arch) {
```
k-ye, line 34: Can we remove this now?

taichi/struct/struct_llvm.h
```diff
@@ -10,9 +10,10 @@ class StructCompilerLLVM : public StructCompiler, public LLVMModuleBuilder {
  public:
   StructCompilerLLVM(Arch arch,
                      const CompileConfig *config,
-                     TaichiLLVMContext *tlctx);
+                     TaichiLLVMContext *tlctx,
+                     std::unique_ptr<llvm::Module> &&module = nullptr);
```
k-ye, line 14: nit: Can we avoid using default values here?

```suggestion
                     std::unique_ptr<llvm::Module> &&module);
```

k-ye (commented):
Great work + a few nits!

taichi/program/program.cpp
```diff
@@ -448,8 +442,11 @@ SNode *Program::get_snode_root(int tree_id) {
 void Program::materialize_snode_tree(SNodeTree *tree) {
   auto *const root = tree->root();
   // always use host_arch() for host accessors
+  std::unique_ptr<llvm::Module> host_module = nullptr;
+  if (!snode_trees_.empty())
+    host_module = llvm_context_host->clone_struct_module();
```
k-ye, line 447: It was a bit confusing to me how come this didn't crash on the first call, then I realized that this might be calling `clone_runtime_module()` inside the constructor.

nit: I believe we can make this a bit more obvious ;-) E.g.

```cpp
... host_module = nullptr;
if (!...)
  host_module = llvm...->clone_struct_module();
if (!host_module)
  host_module = llvm...->clone_runtime_module();
```
Then we can get rid of the `module == nullptr ? ... : ...` inside `StructCompilerLLVM`.

k-ye (approved):
LGTM!

PR #2495, opened 2021-07-06
[Refactor] Split transformer.py into StmtBuilder and ExprBuilder (Stage 1)

xumingkuan:
Related issue = #2193

This PR replaces `ASTTransformerPreprocess` with `StmtBuilder` and `ExprBuilder`.

This PR doesn't modify the behavior (and the code) of the other 2 passes in transformer.py.

Detailed changeset:
- Move all members of the class `ASTTransformerPreprocess` to `BuilderContext`. These members should be context instead of members of a pass.
- Move `ScopeGuard`, `parse_stmt`, `parse_expr` to ast_builder_utils.py.
- Rename all functions in `ASTTransformerPreprocess` from `visit_Xxx` to `build_Xxx` and move them to expr_builder.py and stmt_builder.py.
- Remove all `generic_visit` in `ASTTransformerPreprocess` in transformer.py. Since we need to know if it's a statement or an expr before building the subtree of AST, we cannot use a general `generic_visit`. The invocations of this function is rewritten as `build_expr`, `build_exprs`, `build_stmt`, `build_stmts`.
- Note that `build_exprs` and `build_stmts` create a `variable_scope`. Also note that I do not use a list comprehension in these two functions because `with` will not work properly otherwise.
- Note that `build_Module` does not use `build_stmts`, just like its not using `generic_visit` before this PR. This is because if we create a `variable_scope` associated with the statement list (Python list here), the **parameters** passed into the Taichi kernel will be **deleted** when destructing the variable scope.
- Fix some missing fields required by the Python AST but not required when executing it (compiling Taichi). (Otherwise the AST may not be printed properly.)
- Throw exception if we encounter an AST node without corresponding builder (e.g., if `build_Name` is not implemented in `ExprBuilder` and we invoke `build_expr` with a Name node as the parameter).
- Throw TaichiSyntaxError if the Python keyword `global`/`nonlocal` is used or when a Python set (including set comprehension) is used.
- Add many builder functions which look trivial and in fact unnecessary if we use `visit_Xxx` before this PR. I think there's a benefit of writing these functions explicitly -- they show which features in Python are supported in Taichi, and throw exceptions when users use features in Python that are not considered by Taichi developers, which may or may not be supported.

Note:
- Not supported both before and after this PR:
  - `a = f'aaa{123}'` (ast.JoinedStr, together with ast.FormattedValue)
  - `b = a[1:2]` (ast.Slice)
  - `b = (n**2 for n in aa if n>5 if n<10)` (ast.GeneratorExp)
  - `a = lambda x, y: x + y` (ast.Lambda)
- **(Probably incorrectly) supported before this PR but unsupported after this PR:**
  - `(b): ti.i32 = 1` (ast.AnnAssign)
  - `yield a` and `yield from a` (ast.Yield and ast.YieldFrom)
  - `{1, 2, 3}` and `{x for x in numbers}` (ast.Set and ast.SetComp)
  - `class Foo: ...` (ast.ClassDef)
  - `def foo()` (ast.FunctionDef) in a kernel
- *Not supported but did not throw syntax error before this PR:*
  - ast.AsyncFunctionDef, ast.Await, ast.AsyncFor, ast.AsyncWith do not throw syntax error before this PR. They are not considered in this PR.
  - `global` and `nonlocal` throw TaichiSyntaxError after this PR. They may crash later before this PR.


python/taichi/lang/stmt_builder.py
```diff
@@ -0,0 +1,675 @@
...
+
+    @staticmethod
+    def build_Expr(ctx, node):
+        if isinstance(node.value, ast.Call):
+            # A function call.
+            node.value = build_expr(ctx, node.value)
+            # note that we can only return an ast.Expr instead of an ast.Call.
+            return node
+        # A statement with a single expression.
+        # TODO(#2495): Deprecate maybe_transform_ti_func_call_to_stmt
```
k-ye, line 652: I wonder what's the situation with `maybe_transform_ti_func_call_to_stmt` after this PR? Also does this method generate a ExprStmt?

k-ye (commented):
Thanks for this huge refactoring!

1. How many stages do you expect there will be?
1. Could you write some comments to guide  the review process?

xumingkuan:
> Thanks for this huge refactoring!
> 
> 1. How many stages do you expect there will be?
> 2. Could you write some comments to guide  the review process?

1. I would expect that there will be only 2 stages -- the other one is to remove `maybe_transform_ti_func_call_to_stmt` and merge the multiple-return check into the new ast builders. I think there will be only one pass in Python after this refactoring.
2. Sure!

python/taichi/lang/stmt_builder.py
xumingkuan: No. This PR should not modify the behavior of the Python frontend (the whole transformer.py together with new files introduced in this PR).

python/taichi/lang/expr_builder.py
```diff
@@ -0,0 +1,228 @@
...
+            exit(0)
+
+        new_node = node.values[0]
+        for i in range(1, len(node.values)):
+            new_node = make_node(new_node, node.values[i], token)
+
+        return new_node
+
+    @staticmethod
+    def build_BinOp(ctx, node):
```
xumingkuan, line 160: New functions (not included in transformer.py before) start form here.

python/taichi/lang/stmt_builder.py
```diff
@@ -0,0 +1,675 @@
...
+            node.body = [build_stmt(ctx, stmt) for stmt in list(node.body)]
+        return node
+
+    @staticmethod
+    def build_Global(ctx, node):
+        raise TaichiSyntaxError(
+            "Keyword 'global' not supported in Taichi kernels")
+
+    @staticmethod
+    def build_Raise(ctx, node):
```
xumingkuan, line 640: New functions (not included in transformer.py before) start form here.

python/taichi/lang/stmt_builder.py
```diff
@@ -0,0 +1,675 @@
...
+    @staticmethod
+    def build_Module(ctx, node):
+        with ctx.variable_scope():
+            # Do NOT use |build_stmts| which inserts 'del' statements to the
+            # end and deletes parameters passed into the module
+            node.body = [build_stmt(ctx, stmt) for stmt in list(node.body)]
+        return node
+
+    @staticmethod
+    def build_Global(ctx, node):
```
xumingkuan, line 635: Note that the behavior of `global` is different from transformer.py -- If we use `global`, the program may crash later before this PR.

tests/python/test_ptr_assign.py
```diff
@@ -104,23 +104,3 @@ def func(self):
     obj = MyClass()
     obj.func()
     assert obj.a[None] == 5.0
-
-
-@ti.test(ti.cpu)
-def test_expr_list():
```
xumingkuan, line 110: Moved to separate files.

xumingkuan:
Do we support Python 3.9 now?

k-ye:
> Do we support Python 3.9 now?

Yep

xumingkuan:
Is NamedExpr unsupported in Python 3.6?

python/taichi/lang/expr_builder.py
```diff
@@ -0,0 +1,255 @@
...
+    def build_NameConstant(ctx, node):
+        return node
+
+
+build_expr = ExprBuilder()
+
+
+def build_exprs(ctx, exprs):
+    result = []
+    with ctx.variable_scope(result):
```
xumingkuan, line 252:
```suggestion
    # TODO(#2495): check if we really need this variable scope
    with ctx.variable_scope(result):
```

python/taichi/lang/stmt_builder.py
```diff
@@ -0,0 +1,684 @@
...
+
+    @staticmethod
+    def make_constant(value):
+        # Do not use ast.Constant which does not exist in python3.5
+        node = parse_expr('0')
+        node.value = value
+        return node
+
+    @staticmethod
+    def get_targets(node):
```
xumingkuan, line 38: get_for_loop_targets

python/taichi/lang/stmt_builder.py
```diff
@@ -0,0 +1,684 @@
...
+        new_node.value.args[1] = parse_expr("'{}'".format(msg.strip()))
+        new_node.value.args[2] = extra_args
+        new_node = ast.copy_location(new_node, node)
+        return new_node
+
+    @staticmethod
+    def build_Assign(ctx, node):
+        assert (len(node.targets) == 1)
+        node.value = build_expr(ctx, node.value)
+        node.targets = [build_expr(ctx, t) for t in list(node.targets)]
```
xumingkuan, line 123: build_exprs?

k-ye (approved):
LGTM!!

ljcc0930 (approved):
LGTM!

k-ye:
FYI you don't have to `[skip ci]` now.... New commit will automatically cancel the previous CI 😆

PR #2497, opened 2021-07-07
[lang] Add _RootFieldsBuilder class for root

ljcc0930:
Related issue = #2418 
https://github.com/taichi-dev/taichi/pull/2493#discussion_r664321743

### DIFF
--- a/python/taichi/snode/fields_builder.py
+++ b/python/taichi/snode/fields_builder.py
@@ -130,6 +130,7 @@ class _RootFieldsBuilder(FieldsBuilder):
     def finalize(self):
         if not self._finalized:
             super().finalize()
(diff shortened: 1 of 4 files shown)


python/taichi/snode/fields_builder.py
```diff
@@ -111,6 +122,18 @@ def _check_not_finalized(self):
         if self._finalized:
             raise InvalidOperationError('FieldsBuilder finalized')
 
-    @property
-    def finalized(self):
-        return self._finalized
+class _RootFieldsBuilder(FieldsBuilder):
+    def __init__(self):
+        self._old_roots = []
```
k-ye, line 127: If you need to recreate this object, shall we place it inside `_Root`?

PR #2501, opened 2021-07-08
[lang] Support ti.FieldsBuilder()

ljcc0930:
Related issue = #2418

### DIFF
--- a/taichi/ir/statements.h
+++ b/taichi/ir/statements.h
@@ -912,7 +912,7 @@ class BitExtractStmt : public Stmt {
  */
 class GetRootStmt : public Stmt {
  public:
-  GetRootStmt(SNode *root) : root_(root) {
+  GetRootStmt(SNode *root = nullptr) : root_(root) {
     while (this->root_->parent) {
       this->root_ = this->root_->parent;
     }
--- a/taichi/ir/statements.h
+++ b/taichi/ir/statements.h
@@ -912,8 +912,8 @@ class BitExtractStmt : public Stmt {
  */
 class GetRootStmt : public Stmt {
  public:
-  GetRootStmt(SNode *root): root_(root) {
-    while(this->root_->parent) {
+  GetRootStmt(SNode *root) : root_(root) {
+    while (this->root_->parent) {
       this->root_ = this->root_->parent;
     }
     TI_STMT_REG_FIELDS;
@@ -929,6 +929,7 @@ class GetRootStmt : public Stmt {
   SNode *root() {
     return root_;
   }
+
  private:
   SNode *root_;
 };
--- a/taichi/ir/statements.h
+++ b/taichi/ir/statements.h
@@ -912,10 +912,9 @@ class BitExtractStmt : public Stmt {
  */
 class GetRootStmt : public Stmt {
  public:
-  SNode *snode;
-  GetRootStmt(SNode *snode): snode(snode) {
-    while(this->snode->parent) {
-      this->snode = this->snode->parent;
+  GetRootStmt(SNode *root): root_(root) {
+    while(this->root_->parent) {
+      this->root_ = this->root_->parent;
     }
     TI_STMT_REG_FIELDS;
   }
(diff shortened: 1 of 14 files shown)


taichi/ir/statements.h
```diff
@@ -912,15 +912,19 @@ class BitExtractStmt : public Stmt {
  */
 class GetRootStmt : public Stmt {
  public:
-  GetRootStmt() {
+  SNode *snode;
```
k-ye, line 915: nit: How about we call this `root`, so that it's more easy to be distinguished from the ctor parameter name.

Another personal note: it's better to privatize member variables, and provide a getter to access that (i.e., Python's `@property`). However, I know that most of Taichi's codebase is not following this style, so it's fine either way.

taichi/runtime/llvm/runtime.cpp
```diff
@@ -525,8 +525,11 @@ struct LLVMRuntime {
   host_printf_type host_printf;
   host_vsnprintf_type host_vsnprintf;
   Ptr program;
-  Ptr root;
-  size_t root_mem_size;
+  
+  Ptr roots[taichi_max_num_snodes];
```
k-ye, line 529: nit: Let's make this `kTaichiMaxNumRoots`? (We can also reduce that number a bit..)

taichi/runtime/llvm/runtime.cpp
```diff
@@ -525,8 +525,11 @@ struct LLVMRuntime {
...
+  size_t root_mem_sizes[taichi_max_num_snodes];
+  i32 root_num{0};
```
k-ye, line 531:
```suggestion
  i32 num_roots{0};
```

taichi/runtime/llvm/runtime.cpp
```diff
@@ -894,10 +897,14 @@ void runtime_initialize_snodes(LLVMRuntime *runtime,
                                int num_snodes) {
   // For Metal runtime, we have to make sure that both the beginning address
   // and the size of the root buffer memory are aligned to page size.
-  runtime->root_mem_size =
+  // TODO: use current_root as argument.
+  i32 current_root = runtime->root_num;
```
k-ye, line 901:
```suggestion
  const i32 current_root = runtime->num_roots;
```

taichi/transforms/scalar_pointer_lowerer.cpp
```diff
@@ -46,35 +46,37 @@ void ScalarPointerLowerer::run() {
...
-      for (int k = 0; k < taichi_max_num_indices; k++) {
-        if (snode->physical_index_position[k_] == k) {
-          start_bits[k] -= snode->extractors[k].num_bits;
-          const int begin = start_bits[k];
-          const int end = begin + snode->extractors[k].num_bits;
-          auto extracted = Stmt::make<BitExtractStmt>(indices_[k_], begin, end);
-          lowered_indices.push_back(extracted.get());
-          lowered_->push_back(std::move(extracted));
-          strides.push_back(1 << snode->extractors[k].num_bits);
+  if(path_length_ > 0) {
```
k-ye, line 49: nit: early return? I.e., `if (path_length_ == 0) return;`

taichi/runtime/llvm/runtime.cpp
```diff
@@ -1743,9 +1750,10 @@ i32 wasm_materialize(Context *context) {
       (RandState *)((size_t)context->runtime + sizeof(LLVMRuntime));
   // set random seed to (1, 0, 0, 0)
   context->runtime->rand_states[0].x = 1;
-  context->runtime->root =
+  // TODO: remove hard coding on root id 0(SNodeTree::kFirstID)
+  context->runtime->roots[0] =
```
k-ye, line 1754: @squarefk : This could potentially break the current WASM runtime...

k-ye (commented):
Great! Left a few nits

taichi/runtime/llvm/runtime.cpp
ljcc0930: I thought that it works if there's only one SNodeTree.

k-ye (approved):
LGTM!

PR #2503, opened 2021-07-08
[lang] Support ti.fields with shape after materialized

### LINKED ISSUE #2418: [RFC] Support adding new fields dynamically

ljcc0930:
Related issue = close #2418

### DIFF
--- a/python/taichi/snode/fields_builder.py
+++ b/python/taichi/snode/fields_builder.py
@@ -11,7 +11,6 @@ _snode_registry = _ti_core.SNodeRegistry()
 
 _Axis = _ti_core.Index
 
-
 class FieldsBuilder:
     """A builder that constructs a SNodeTree instance.
 
@@ -34,11 +33,21 @@ class FieldsBuilder:
         #  +-- pointer +-- dense +-- place(y)
         fb.finalize()
     """
+    _finalized_fbs = []
+
     def __init__(self):
         self._ptr = _snode_registry.create_root()
         self._root = snode.SNode(self._ptr)
         self._finalized = False
         self._empty = True
+
+    @classmethod
+    def finalized_fbs(cls):
+        return cls._finalized_fbs
+
+    @classmethod
+    def clear_finalized_fbs(cls):
+        cls._finalized_fbs = []
 
     @property
     def ptr(self):
@@ -115,6 +124,7 @@ class FieldsBuilder:
         """Same as :func:`taichi.SNode.lazy_grad`"""
         # TODO: This complicates the implementation. Figure out why we need this
         self._check_not_finalized()
+        self._empty = False
         self._root.lazy_grad()
 
     def finalize(self):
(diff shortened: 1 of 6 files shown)


k-ye (approved):
LGTM!!

python/taichi/snode/fields_builder.py
```diff
@@ -103,14 +120,12 @@ def lazy_grad(self):
     def finalize(self):
         """Constructs the SNodeTree and finalizes this builder."""
         self._check_not_finalized()
+        if self._empty:
+            warning("Finalized a empty FieldsBuilder!")
```
k-ye, line 124:
```suggestion
            warning("Finalizing an empty FieldsBuilder!")
```

Issue #2518, opened 2021-07-13
[taichi 0.7.26] Vector does not accumulate properly inside loops


AmosChenZixuan:
**Describe the bug**
Updating to this newer version of taichi, my previous code stopped working; when using a local accumulating variable to store the sum of some values, it only stores the first value.
P.S. the code is running on cuda.

**To Reproduce**

```py
# inside some kernel
@ti.func
def calc_CM(x):
    # center of mass
    cm = ti.Vector([0.0, 0.0])
    total_mass = 0.0
    for i in x:
        cm += x[i] * M[i]
        total_mass += M[i]
    print(cm, x[0], total_mass)
    cm /= total_mass
    return cm
```
It prints that cm is the same as x[0], instead of the sum of all x. However, total_mass is accumulated correctly.
If I move the print statement inside the loop, then everything is working again.
```py
# inside some kernel
@ti.func
def calc_CM(x):
    # center of mass
    cm = ti.Vector([0.0, 0.0])
    total_mass = 0.0
    for i in x:
        cm += x[i] * M[i]
        total_mass += M[i]
        print(cm, x[0], total_mass)
    cm /= total_mass
    return cm
```

**Log/Screenshots**
print statement is outside the loop 
```
[Taichi] mode=release
[Taichi] preparing sandbox at /tmp/taichi-ujlv54ih
[Taichi] version 0.7.26, llvm 10.0.0, commit e37bdb5e, linux, python 3.8.8
[Taichi] Starting on arch=cuda
[0.000000, 0.600000] [0.000000, 0.600000] 20.000000
```
print statement is inside the loop (20 lines of the same content)
```
[Taichi] mode=release
[Taichi] preparing sandbox at /tmp/taichi-ujlv54ih
[Taichi] version 0.7.26, llvm 10.0.0, commit e37bdb5e, linux, python 3.8.8
[Taichi] Starting on arch=cuda
[0.000000, 9.999999] [0.000000, 0.600000] 20.000000
[0.000000, 9.999999] [0.000000, 0.600000] 20.000000
[0.000000, 9.999999] [0.000000, 0.600000] 20.000000
...
```

k-ye:
Suspicious PR: https://github.com/taichi-dev/taichi/pull/2487. @AmesingFlank @BillXu2000 would you help take a look, thanks!

AmesingFlank:
Hi! 

Thank you for submitting this issue.

[...]

PR #2519, opened 2021-07-13
[Bug] Warp reduction bug fix

AmesingFlank:
Related issue = #2518


AmesingFlank:
I am unsure of why this bug wasn't detected by `test_reduction.py`. It appears that shuffling a `float32` using  `cuda_shfl_down_sync_i32`  gives undefined behavior: it works most of the time, but not always.

k-ye (approved):
LGTM! Thank you for the prompt fix

Issue #2520, opened 2021-07-13
[windows] [bug] Pointer fields with 32-bit data types crash on CUDA on Windows


xumingkuan:
**Describe the bug**
If a `pointer` has a direct `place` child of data type `i32` or `f32` (a data type with <= 4 Bytes), and that child is not the only field in the whole program, the program will crash on CUDA on Windows.

**To Reproduce**
```py
import taichi as ti

ti.init(arch=ti.cuda)

x, y = ti.field(ti.i32), ti.field(ti.i32)

N = 1
ti.root.pointer(ti.i, N).place(x)
ti.root.pointer(ti.i, N).place(y)


@ti.kernel
def foo():
    pass


foo()
```
Note: it doesn't crash if:
both fields are of `i64`;
one of the fields is removed;
both `ti.root.pointer(ti.i, N)` are replaced with `ti.root.pointer(ti.i, N).dense(ti.i, 2)`;
either `pointer` is replaced with `dense`;
`foo()` is not called;
`arch` is `ti.cpu`;
OS is not Windows.

**Log/Screenshots**
```
D:\Anaconda3\python.exe C:/Users/xmk/PycharmProjects/pythontest/main.py
[Taichi] mode=development
[Taichi] <dev mode>, llvm 10.0.0, commit d2a139a4, win, python 3.7.3
[T 07/13/21 20:34:33.821 3956] [cuda_driver.cpp:taichi::lang::CUDADriver::CUDADriver@48] CUDA driver API (v11.4) loaded.
[Taichi] Starting on arch=cuda
[T 07/13/21 20:34:33.821 3956] [program.cpp:taichi::lang::Program::Program@72] Program initializing...
[T 07/13/21 20:34:33.846 3956] [cuda_context.cpp:taichi::lang::CUDAContext::CUDAContext@25] Using CUDA device [id=0]: NVIDIA GeForce RTX 3060 Laptop GPU
[T 07/13/21 20:34:33.846 3956] [cuda_context.cpp:taichi::lang::CUDAContext::CUDAContext@33] CUDA Device Compute Capability: 8.6
[T 07/13/21 20:34:33.931 3956] [cuda_context.cpp:taichi::lang::CUDAContext::CUDAContext@38] Total memory 6.00 GB; free memory 5.04 GB
[T 07/13/21 20:34:33.931 3956] [cuda_context.cpp:taichi::lang::CUDAContext::CUDAContext@51] Emitting CUDA code for sm_75
[...]

PR #2521, opened 2021-07-13
[bug] Set use_unified_memory=True for tests with pointer fields with 32-bit data types

xumingkuan:
Related issue = #2520 

Also adds a test for the issue.


k-ye (approved):
Thank you!

PR #2524, opened 2021-07-14
[Refactor] Move llvm runtime build for archs from C++ to cmake.

ailzhang:
When we build python packages, we used to call `ti.core.compile_runtimes()` to build llvm runtime for archs, but this can be conveniently done in cmake instead of invoking `clang/llvm-as` from C++. 

This PR follows the C++ logic that we always build for host arch, and also cuda if it's available. 

Our C++ code is a bit confusing in where to put the final `runtime_arm64.bc` (it's copied over a few times). AFAIK the hardcoded path in cmake is sufficient for now.


k-ye (approved):
Thank you!

PR #2528, opened 2021-07-14
[bug] Fix python detection on windows

### LINKED ISSUE #2527: ti.exe fail to run on windows for develop build
**Describe the bug**
When there is multiple python install on Windows, the command `where.exe python` will generate a multiline output, which can cause a failure running ti.exe.

```
> where.exe python
C:\Users\xq114\AppData\Local\.xmake\packages\p\python\3.9.6\7f012ab6adf64a85be8cf4480051f442\bin\python.exe
C:\Users\xq114\AppData\Local\Microsoft\WindowsApps\python.exe
> mkdir build && cd build && cmake .. && cmake --build . --config Release
> ..\bin\ti.exe
  File "ti", line 2
    C:\Users\xq114\AppData\Local\Microsoft\WindowsApps\python.exe
                                                                ^
SyntaxError: unexpected character after line continuation character
```

**To Reproduce**
1. Install multiple python in parallel;
2. Build taichi from source;
3. Run ti.exe

**Log/Screenshots**

image

**Additional comments**
The solution is straight forward and I've presented a pr for this

xq114:
Related issue = fixes #2527


k-ye:
Thanks! Seems like this causes CI to fail:

```
MSBUILD : error MSB1009: Project file does not exist.
Switch: taichi.sln
```

k-ye (approved):
Sorry for the late reply, LGTM!

Issue #2513, opened 2021-07-12
Remove FieldsBuilder._finalized_fbs


ljcc0930:
Class `FieldsBuilder` maintained all ofthe builders' information here:
https://github.com/taichi-dev/taichi/blob/e37bdb5e7c688004e6cd34bdcfc14aefdad65cf1/python/taichi/snode/fields_builder.py#L37-L51
It's a redundancy design. Taichi maintained the same information in class `Program` in C++.
https://github.com/taichi-dev/taichi/blob/e37bdb5e7c688004e6cd34bdcfc14aefdad65cf1/taichi/program/program.h#L359

PR #2530, opened 2021-07-15
[refactor] Move snode tree maintainance out of python

Leonz5288:
Related issue = #2513

### DIFF
--- a/python/taichi/snode/fields_builder.py
+++ b/python/taichi/snode/fields_builder.py
@@ -34,32 +34,20 @@ class FieldsBuilder:
         #  +-- pointer +-- dense +-- place(y)
         fb.finalize()
     """
-    _finalized_fbs = []
-
-    def __init__(self, ptr=None, root=None):
-        if ptr is None:
-            self._ptr = _snode_registry.create_root()
-        else:
-            self._ptr = ptr
-        if root is None:
-            self._root = snode.SNode(self._ptr)
-        else:
-            self._root = root
+    def __init__(self):
+        self._ptr = _snode_registry.create_root()
+        self._root = snode.SNode(self._ptr)
         self._finalized = False
         self._empty = True
 
     @classmethod
-    def finalized_fbs(cls):
-        fbs = []
+    def finalized_roots(cls):
+        roots_ptr = []
         size = impl.get_runtime().prog.get_snode_tree_size()
         for i in range(size):
             res = impl.get_runtime().prog.get_snode_root(i)
-            fbs.append(FieldsBuilder(res, snode.SNode(res)))
-        return fbs
-
-    @classmethod
-    def clear_finalized_fbs(cls):
-        cls._finalized_fbs = []
+            roots_ptr.append(snode.SNode(res))
+        return roots_ptr
 
     @property
     def ptr(self):
@@ -147,7 +135,6 @@ class FieldsBuilder:
         _ti_core.finalize_snode_tree(_snode_registry, self._ptr,
                                      impl.get_runtime().prog)
         self._finalized = True
-        self._finalized_fbs.append(self)
 
     def _check_not_finalized(self):
         if self._finalized:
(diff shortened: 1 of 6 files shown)


Leonz5288:
I spent a whole day debugging a non-existed bug... I think my files are somewhat corrupted, I guess I'll build from source again.

python/taichi/snode/fields_builder.py
```diff
@@ -36,15 +36,26 @@ class FieldsBuilder:
...
-        self._root = snode.SNode(self._ptr)
+    def __init__(self, ptr=None, root=None):
+        if ptr is None:
+            self._ptr = _snode_registry.create_root()
+        else:
+            self._ptr = ptr
+        if root is None:
+            self._root = snode.SNode(self._ptr)
+        else:
+            self._root = root
```
ljcc0930, line 47:
```suggestion
    def __init__(self):
        self._ptr = _snode_registry.create_root()
        self._root = snode.SNode(self._ptr)
```

python/taichi/snode/fields_builder.py
```diff
@@ -36,15 +36,26 @@ class FieldsBuilder:
...
         self._finalized = False
         self._empty = True
 
     @classmethod
     def finalized_fbs(cls):
```
ljcc0930, line 52: Maybe we can change finalized_fbs() to finalized_roots()?

python/taichi/snode/fields_builder.py
```diff
@@ -36,15 +36,26 @@ class FieldsBuilder:
...
-        return cls._finalized_fbs
+        fbs = []
+        size = impl.get_runtime().prog.get_snode_tree_size()
+        for i in range(size):
+            res = impl.get_runtime().prog.get_snode_root(i)
+            fbs.append(FieldsBuilder(res, snode.SNode(res)))
```
ljcc0930, line 57:
```suggestion
            fbs.append(snode.SNode(res))
```

ljcc0930:
Good works!

python/taichi/snode/fields_builder.py
ljcc0930: There are only two places that are using finalized_fbs():
https://github.com/taichi-dev/taichi/blob/cd96133e9e45df588c53042099a0b8b37b6c75e1/tests/python/test_tensor_reflection.py#L62
https://github.com/taichi-dev/taichi/blob/cd96133e9e45df588c53042099a0b8b37b6c75e1/python/taichi/lang/__init__.py#L408-L409

In fact, in these places, all we need is the root of SNodeTrees. So, it will have the same effect if we only save the SNodes.

python/taichi/snode/fields_builder.py
```diff
@@ -36,15 +36,26 @@ class FieldsBuilder:
...
+        return fbs
 
     @classmethod
     def clear_finalized_fbs(cls):
```
ljcc0930, line 61:
```suggestion
```

python/taichi/snode/fields_builder.py
Leonz5288: That's how I thought about it in the beginning. However, there are also places that uses self._ptr.

python/taichi/snode/fields_builder.py
Leonz5288: I thought we can get root directly from cpp, instead of creating one in python?

ljcc0930: But it's not necessary to create a FieldsBuilder? I thought we keep a SNode is enough to achieve the effects we need.

ljcc0930: I mean `finalized_fbs` returns a list consists of `SNode`s, instead of a list consists of `FieldsBuilder`s.

ljcc0930 (approved):
LGTM!

PR #2531, opened 2021-07-15
[Refactor] Use setup.py from root dir.

ailzhang:
* **#2531**

A few things included in this PR:
- Move `setup.py` up to root dir so that we can run `python setup.py
  install/develop`.
- Remove dependency of `taichi_core` in build process. We used to call
  `ti.core.get_runtime_dir()/get_version_string()/libdevice_path()` from
C++, but these information are already available in build process.
- Simplify `python/build.py` to call `setup.py` in the root. Tested
  `build/test` commands still work as expected.
- Update github workflows to use `python setup.py install` in linux/mac instead of
  setting env `TAICHI_REPO_DIR`.
- Developers used to customize taichi_core.so build using args passed
  into cmake. Under this new mechanism they can put everything inside
`TAICHI_CMAKE_ARGS` and run `setup.py` using: `TAICHI_CMAKE_ARGS="-DXXXX -DXXXX" python setup.py ..`
- Move `tests/conftest.py` into `tests/python` folder. I think it was a
  bug that we skipped it when we copied `tests` folder when building wheels.

To explain how build works after this PR:
In short, c++ build is no longer a separate step. It's built as an
extension when you build `taichi` python package.

Developers can use
- `python setup.py develop`(or `pip install -e .`) in root dir to
  install taichi into their current python environment. Note in
`develop` mode any local changes to python files take effect
immediately. You only need to rerun this command when you change C++
code.
- `python setup.py install`(or `pip install .`) to install taichi into
  your their current python environment. Note in `install` mode files
are copied over to python's `site-packages` so your local changes won't
take effect until you rerun the command.
- No more manipulation with `TAICHI_REPO_DIR` and `PYTHONPATH`.

A few known followups:
- I left windows job to remain using `TAICHI_REPO_DIR` since I don't
  have a windows machine. But it should be straightfoward to update if
  someone else can help.
- We can start removing `TAICHI_REPO_DIR` in the codebase after this PR since our
  development mode has changed.
- Update developer install documentation.
- I'm pretty sure the handling for windows in `setup.py` might have
  minor problems about relative paths but will leave it to debug in CI.
   It should be a simple fix.
- We probably need to reconsider if we want to include assets/examples
  and a few other .py files under `python/taichi` in the release.


k-ye:
This is great! Sorry it's a big change so I will take some time to look into it later. Thanks!!

setup.py
```diff
@@ -16,35 +24,187 @@
...
+            f'-DTI_VERSION_PATCH={TI_VERSION_PATCH}',
+        ]
+
+        cfg = 'Debug' if self.debug else 'Release'
+        build_args = ['--config', cfg]
+
+        cmake_args += ['-DCMAKE_BUILD_TYPE=' + cfg]
+
+        # Assuming Makefiles
+        build_args += ['--', '-j8']
```
k-ye, line 129: super nit: `multiprocessing.cpu_count() * 2`?

k-ye (commented):
Thank you for the big improvement to the build workflow 🤩 ! I don't have too much to say, will merge once the Win CI gets sorted out :-D

ailzhang:
Btw this PR is ready for a final review, please don't click merge, it requires landing from ghstack instead ;)

.github/workflows/presubmit.yml
```diff
@@ -126,22 +129,30 @@ jobs:
...
         run: |
-          export TAICHI_REPO_DIR=`pwd`
+          TAICHI_REPO_DIR=`pwd`
           export PATH=$TAICHI_REPO_DIR/taichi-llvm/bin/:$PATH
           export CXX=clang++
           python misc/ci_setup.py ci
         env:
           CI_SETUP_CMAKE_ARGS: -DTI_WITH_OPENGL:BOOL=OFF -DTI_WITH_CC:BOOL=${{ matrix.with_cc }} -DTI_BUILD_TESTS:BOOL=${{ matrix.with_cpp_tests }}
 
+      # [DEBUG] Copy this step around to enable debugging inside Github Action instances.
```
k-ye, line 139: Awesome!

k-ye (approved):
LGTM!

PR #2532, opened 2021-07-15
[Perf] Reduce GL backend's overhead (driver overhead & copy overhead)

bobcao3:
This is a (sort of) WIP PR.

In the OpenGL backend there are quite some overhead on kernel launch. These overheads can be categorized into two types:
1. Driver overhead
2. Synchronization / copy overhead

For the driver overhead, the main problem is the (unpredictable) cost of GL functions. Currently every frame the code creates a series of buffers, and then populate them with the host side data we want to pass in. This process not only involves C++'s RAII and related allocation overhead, it also calls `glGenBuffers` and `glDeleteBuffers` which do not have a predictable cost. The driver will do a lot of things behind, and therefore it is better to not allocating and reallocating resources every frame. This process should be managed internally since we know the details of these buffers, the drivers do not, not until we request to upload data.

For the synchronization overhead, after a kernel's invocation all buffers are transferred back to host indiscriminately. This transfer should not be needed if say the kernel does not have a return argument, or it only performs read on a certain buffer. Currently this PR has not utilized the read write access insight from codegen yet, but simply avoiding a copy if there is no returns is already yielding performance improvements. (Also in GL a device to host copy causes an implicit synchronization which flushes the GPU pipeline, so the cost is higher than what we may expect.)

Currently I observed in `cornell_box.py` an increase in `270~280 samples` to `300~330 samples` per second.

For the future iterations of this PR (or maybe making a new PR) I'm aiming to revise the device resource management in the GL backend. Currently it is doing a few things that are not quite optimal, and some internal terms / APIs in the code are slightly confusing.

----

7.16.2021, commit `1be6fa4`
- Added a transform that detects the read write operations to external ptrs (perhaps this should be in analysis instead of transforms?)
- Use this knowledge to reduce host/device transfers. (If kernel does not read from extptr, we don't need to transfer the current content in.)
- Potential issue here: What if the kernel is expected to write to a few entries in the extptr but not the entire one, where old data should be preserved? Is this a valid case in taichi? If that is the case then this transfer becomes more or less unavoidable, unless we eliminate the use of this external buffer on the host side and let it stay in GPU.
- I propose we should add a unified device memory allocator / manager, and all backends or frontends use this allocator to create buffers. Hopefully with this we can use advanced features such as host-visible device memory and persistently mapped resources. This will also help later API interop work.

### DIFF
--- a/taichi/backends/opengl/opengl_api.cpp
+++ b/taichi/backends/opengl/opengl_api.cpp
@@ -672,11 +672,9 @@ struct CompiledProgram::Impl {
     }
     for (auto &[idx, buf] : launcher->impl->user_bufs.bufs) {
       if (buf->index == GLBufId::Args) {
-        if (ret_count > 0) {
-          // Copying of data between host and device cause implicit
-          // synchronization Should avoid any synchronization if not needed
-          buf->copy_back(launcher->result_buffer, ret_count * sizeof(uint64_t));
-        }
+        // Copying of data between host and device cause implicit
+        // synchronization Should avoid any synchronization if not needed
+        buf->copy_back(launcher->result_buffer, ret_count * sizeof(uint64_t));
       } else {
         // TODO: Use the analysis information built during codegen to figure out
         // whether a buffer is read-only or not If a buffer is read-only there
--- a/taichi/backends/opengl/opengl_api.cpp
+++ b/taichi/backends/opengl/opengl_api.cpp
@@ -155,147 +155,75 @@ struct GLProgram {
 // https://blog.csdn.net/ylbs110/article/details/52074826
 // https://www.khronos.org/opengl/wiki/Shader_Storage_Buffer_Object
 // This is Shader Storage Buffer, we use it to share data between CPU & GPU
-struct GLSSBO {
-  GLuint id_;
+struct GLBuffer {
+  GLuint id;
+  GLenum type;
+  size_t size;
 
-  GLSSBO() {
-    glGenBuffers(1, &id_);
-  }
-
-  ~GLSSBO() {
-    glDeleteBuffers(1, &id_);
-  }
-
-  /***
-   GL_{frequency}_{nature}:
-
-
-   STREAM
-       The data store contents will be modified once and used at most a few
-   times.
-
-   STATIC
-       The data store contents will be modified once and used many times.
-
-   DYNAMIC
-       The data store contents will be modified repeatedly and used many times.
-
-
-   DRAW
-       The data store contents are modified by the application, and used as the
-   source for GL drawing and image specification commands.
-
-   READ
-       The data store contents are modified by reading data from the GL, and
-   used to return that data when queried by the application.
-
-   COPY
-       The data store contents are modified by reading data from the GL, and
-   used as the source for GL drawing and image specification commands.
-   ***/
-
-  void bind_data(void *data,
-                 size_t size,
-                 GLuint usage = GL_DYNAMIC_READ) const {
-    glBindBuffer(GL_SHADER_STORAGE_BUFFER, id_);
+  GLBuffer(size_t size,
+           void *initial_data = nullptr,
+           GLenum type = GL_SHADER_STORAGE_BUFFER,
+           GLbitfield access = GL_MAP_READ_BIT | GL_MAP_WRITE_BIT)
+      : size(size), type(type) {
+    glGenBuffers(1, &id);
+    check_opengl_error("glGenBuffers");
+    glBindBuffer(type, id);
     check_opengl_error("glBindBuffer");
-    glBufferData(GL_SHADER_STORAGE_BUFFER, size, data, usage);
-    check_opengl_error("glBufferData");
+
+    if (size) {
+      glBufferStorage(type, size, initial_data, access);
+      check_opengl_error("glBufferStorage");
+    }
   }
 
-  void bind_index(size_t index) const {
-    // SSBO index, is `layout(std430, binding = <index>)` in shader.
-    // We use only one SSBO though...
-    glBindBufferBase(GL_SHADER_STORAGE_BUFFER, index, id_);
-    check_opengl_error("glBindBufferBase");
+  void bind_to_index(GLuint location, size_t offset, size_t size) {
+    if (size) {
+      glBindBufferRange(type, location, id, GLintptr(offset), GLsizeiptr(size));
+      check_opengl_error("glBindBufferRange");
+    }
   }
 
-  void bind_range(size_t index, size_t offset, size_t size) const {
-    glBindBufferRange(GL_SHADER_STORAGE_BUFFER, index, id_, offset, size);
-    check_opengl_error("glBindBufferRange");
+  void bind_to_index(GLuint location) {
+    bind_to_index(location, 0, size);
   }
 
-  void as_indirect_buffer() {
-    glBindBuffer(GL_DISPATCH_INDIRECT_BUFFER, id_);
+  void *map(GLenum access) {
+    return map_region(0, size, access);
   }
 
-  void *map(size_t offset,
-            size_t length,
-            GLbitfield access = GL_READ_ONLY) const {
-    // map GPU memory to CPU address space, offset within SSBO data
-    glBindBuffer(GL_SHADER_STORAGE_BUFFER, id_);
+  void *map_region(size_t offset, size_t size, GLenum access) {
+    glBindBuffer(type, id);
     check_opengl_error("glBindBuffer");
-    void *p =
-        glMapBufferRange(GL_SHADER_STORAGE_BUFFER, offset, length, access);
+    void *mapped =
+        glMapBufferRange(type, GLintptr(offset), GLsizeiptr(size), access);
     check_opengl_error("glMapBufferRange");
-    TI_ASSERT_INFO(p, "glMapBufferRange returned NULL");
-    return p;
+    TI_ASSERT(mapped);
+    return mapped;
   }
 
-  void *map(GLbitfield access = GL_READ_ONLY) const {
-    glBindBuffer(GL_SHADER_STORAGE_BUFFER, id_);
+  void unmap() {
+    glBindBuffer(type, id);
     check_opengl_error("glBindBuffer");
-    void *p = glMapBuffer(GL_SHADER_STORAGE_BUFFER, access);
-    check_opengl_error("glMapBuffer");
-    TI_ASSERT_INFO(p, "glMapBuffer returned NULL");
-    return p;
-  }
-
-  void unmap() const {
-    glBindBuffer(GL_SHADER_STORAGE_BUFFER, id_);
-    check_opengl_error("glBindBuffer");
-    glUnmapBuffer(GL_SHADER_STORAGE_BUFFER);
+    glUnmapBuffer(type);
     check_opengl_error("glUnmapBuffer");
   }
 
-  void upload_data(void *data, size_t offset, size_t size) const {
-    glBindBuffer(GL_SHADER_STORAGE_BUFFER, id_);
+  void flush_mapped_region(size_t offset, size_t size) {
+    glBindBuffer(type, id);
     check_opengl_error("glBindBuffer");
-    glBufferSubData(GL_SHADER_STORAGE_BUFFER, offset, size, data);
-    check_opengl_error("glBufferData");
+    glFlushMappedBufferRange(type, offset, size);
+    check_opengl_error("glFlushMappedBufferRange");
   }
 
-  void read_back_data(void *data, size_t offset, size_t size) const {
-    glBindBuffer(GL_SHADER_STORAGE_BUFFER, id_);
+  void read_back(void *buffer, size_t offset, size_t size) {
+    glBindBuffer(type, id);
     check_opengl_error("glBindBuffer");
-    glGetBufferSubData(GL_SHADER_STORAGE_BUFFER, offset, size, data);
-    check_opengl_error("glBufferData");
-  }
-};
-
-struct GLBuffer : GLSSBO {
-  GLBufId index;
-  void *base;
-  size_t size;
-
-  GLBuffer(GLBufId index, void *base, size_t size)
-      : index(index), base(base), size(size) {
-    bind_data(base, size);
-    bind_index((int)index);
+    glGetBufferSubData(type, offset, size, buffer);
+    check_opengl_error("glGetBufferSubData");
   }
 
-  GLBuffer(GLBufId index) : index(index), base(nullptr), size(0) {
-    bind_index((int)index);
-  }
-
-  void copy_forward() {
-    bind_data(base, size);
-  }
-
-  void rebind(void *new_base, size_t new_size) {
-    base = new_base;
-    size = new_size;
-    bind_data(base, size);
-  }
-
-  void copy_back() {
-    copy_back(base, size);
-  }
-
-  void copy_back(void *ptr, size_t len) {
-    if (!len)
-      return;
-    this->read_back_data(ptr, 0, len);
+  void read_back(void *buffer) {
+    read_back(buffer, 0, size);
   }
 };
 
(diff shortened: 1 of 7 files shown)


bobcao3:
To be clear this is not a common case in taichi sine this is purely CPU-side overhead. Most examples / applications in taichi are heavy kernels where the majority of time is spent on the kernel. The cornell box example has a relatively fast kernel, therefore currently its performance is dominated by these overheads. (I'm testing on a RTX3080)

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -285,26 +287,106 @@ struct GLBuffer : GLSSBO {
   }
 };
 
+struct GLBufferAllocator {
+ private:
+  const size_t max_free_resident_size = 64 << 20;
+
+  std::list<std::unique_ptr<GLBuffer>> buffers_storage;
+
+  std::map<std::pair<GLBufId, size_t>, GLBuffer *> buffers_mapping;
```
k-ye, line 296: can we use unordered_map instead?

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -285,26 +287,106 @@ struct GLBuffer : GLSSBO {
...
+
+      // TI_WARN("New buffer {}, {}, {}", int(index), size, (void*)buffer);
+    } else {
+      // Reuse
+      buffer = buffer_iter->second;
+      free_mapping.erase(buffer_iter);
+
+      buffer->rebind(base, size);
+      buffer->bind_index(int(index));
+
```
k-ye, line 338: Update `buffers_mapping` as well?

k-ye (commented):
Thanks for helping improve the perf!

taichi/backends/opengl/opengl_api.cpp
bobcao3: Since `size` and `index` are kept constant during reuse, the entry within `buffers_mapping` should be constant as well

k-ye: Then we should clean up `buffers_mapping` in `new_frame()`.

taichi/backends/opengl/opengl_api.cpp
k-ye: Hmm, I guess it's because `unordered_map` can't really hash the key of type `std::pair<GlBufId, size_t>`. I think we can have a very small struct here, both for being the key type (with its hasher), and for making the semantics more clear (i.e. we don't have to write  `first.second` and guess what `second` is...)

taichi/backends/opengl/opengl_api.cpp
bobcao3: Ah right

taichi/backends/opengl/opengl_api.cpp
bobcao3: Actually reading the code a bit I think index is not even needed. Since all buffers are created with essentially the same access hint, and GL does not have the concept of memory pool types (i think), we only need to have the size to be the same to reuse.

bobcao3: In Vulkan we will need to only reuse buffers from the same pool / type of pool, but not in GL.

bobcao3:
@k-ye The latest commit added something that detects the usage of external buffer. Using this knowledge we saved a solid 1~2ms on the CPU side doing transfers. However this has a potential issue, outlined in the updated comment, I'm not sure whether this will be correct anymore. Potential solution for that issue said without just copying everything is to have external buffer binding be explicit about whether the original content should be preserved. (Kind of like in Vulkan RenderPass you have a LoadOp and a StoreOp, where LoadOp can be `don't care` or `load`)

bobcao3:
Also it seems it would be possible to put argument buffers into OpenGL's Uniform Buffer. Uniform Buffer can be preloaded to shader invocations, making access to them local compared to SSBOs. This also avoids the awkward issue when the return value size is much larger than the argument buffer, and some the arguments must be duplicated and padded on the host.

k-ye:
> Added a transform that detects the read write operations to external ptrs (perhaps this should be in analysis instead of transforms?)

Great! Yeah agreed that this is an analysis.

> Use this knowledge to reduce host/device transfers. (If kernel does not read from extptr, we don't need to transfer the current content in.)

Great. BTW there seems to be a parity. If the kernel doesn't write to the extptr, then we don't have to transfer it back to the CPU.

> Potential issue here: What if the kernel is expected to write to a few entries in the extptr but not the entire one, where old data should be preserved? Is this a valid case in taichi? If that is the case then this transfer becomes more or less unavoidable, unless we eliminate the use of this external buffer on the host side and let it stay in GPU.

This is a valid case, although in practice it's rarely used. When there is a extptr wrtite, it's mostly for transferring the entire Taichi field.

One way to resolve this problem is probably to introduce a finer-grained type annotation. E.g.

```
ti.ext_arr(dir=in)       # read-only
ti.ext_arr(dir=out)      # write-only
ti.ext_arr(dir=inout)    # read-write
```

But I feel this to be not very intuitive from the user's perspective.

> I propose we should add a unified device memory allocator / manager, and all backends or frontends use this allocator to create buffers. Hopefully with this we can use advanced features such as host-visible device memory and persistently mapped resources. This will also help later API interop work.

Strongly agreed on this! We'd actually wish to push this one step further and unify the entire backend runtime interface. A lot of the logic for how to sync/to transfer data has to be re-implemented on each backend. (E.g. right now the host-side synchronization invocation is done on multiple layers)

bobcao3:
Update on putting argument buffer in Uniform Buffer Objects, testing shows no performance change. In addition, due to the oddity of std140 layout requirement of UBOs, array elements are padded to 16 bytes therefore wasting significant space (this might be why the performance is bad). This will be a good idea in Vulkan (with push constants or uniform buffer descriptors), not in GL sadly.

taichi/backends/opengl/opengl_api.cpp
bobcao3: An update to this reuse topic: Since I switched to persistent storage (glShaderStorage), there is (kind of) memory pool in GL. For example, a buffer created without the MAP_WRITE_BIT can not be mapped and written to by the host.

bobcao3:
> 
> 
> > Added a transform that detects the read write operations to external ptrs (perhaps this should be in analysis instead of transforms?)
> 
> Great! Yeah agreed that this is an analysis.
> 
> > Use this knowledge to reduce host/device transfers. (If kernel does not read from extptr, we don't need to transfer the current content in.)
> 
> Great. BTW there seems to be a parity. If the kernel doesn't write to the extptr, then we don't have to transfer it back to the CPU.
> 
> > Potential issue here: What if the kernel is expected to write to a few entries in the extptr but not the entire one, where old data should be preserved? Is this a valid case in taichi? If that is the case then this transfer becomes more or less unavoidable, unless we eliminate the use of this external buffer on the host side and let it stay in GPU.
> 
> This is a valid case, although in practice it's rarely used. When there is a extptr wrtite, it's mostly for transferring the entire Taichi field.
> 
> One way to resolve this problem is probably to introduce a finer-grained type annotation. E.g.
> 
> ```
> ti.ext_arr(dir=in)       # read-only
> ti.ext_arr(dir=out)      # write-only
> ti.ext_arr(dir=inout)    # read-write
> ```
> 
> But I feel this to be not very intuitive from the user's perspective.
> 
> > I propose we should add a unified device memory allocator / manager, and all backends or frontends use this allocator to create buffers. Hopefully with this we can use advanced features such as host-visible device memory and persistently mapped resources. This will also help later API interop work.
> 
> Strongly agreed on this! We'd actually wish to push this one step further and unify the entire backend runtime interface. A lot of the logic for how to sync/to transfer data has to be re-implemented on each backend. (E.g. right now the host-side synchronization invocation is done on multiple layers)
> 
> > to have external buffer binding be explicit about whether the original content should be preserved.
> 
> +1 on this. Sorry I missed your later comments...
> 
> > Also it seems it would be possible to put argument buffers into OpenGL's Uniform Buffer.
> 
> SGTM!

For the external array, it seems in GL there's a bit flag that controls whether the previous content of a buffer is discarded before being mapped on the host. Maybe we can use a similar concept here, by default if the kernel only writes to a buffer the original content will be discarded. If the kernel reads from the kernel or there's an explicit flag saying content should be preserved, then the external array will be protected. (However this feels a bit gimmicky due to the reliance on kernel read / write)

taichi/transforms/detect_read_only.cpp
```diff
@@ -21,6 +22,43 @@ void detect_read_only_in_task(OffloadedStmt *offload) {
   }
 }
 
+class ExternalPtrAccessVisitor : public BasicStmtVisitor {
+ private:
+  std::unordered_map<int, int> &map;
```
k-ye, line 27: nit: please name these private members with a `_` suffix. Also, for consistency, let's place the member variables in the end of theclass definition.

taichi/transforms/detect_read_only.cpp
```diff
@@ -21,6 +22,43 @@ void detect_read_only_in_task(OffloadedStmt *offload) {
...
+
+ public:
+  using BasicStmtVisitor::visit;
+
+  ExternalPtrAccessVisitor(std::unordered_map<int, int> &map)
+      : map(map), BasicStmtVisitor() {
+  }
+
+  void visit(GlobalLoadStmt *stmt) {
+    if (stmt->src && stmt->src->is<ExternalPtrStmt>()) {
```
k-ye, line 37: nit: I'd suggest to do early return here, i.e.

```cpp
if (!(...)) return;
```

taichi/transforms/detect_read_only.cpp
```diff
@@ -21,6 +22,43 @@ void detect_read_only_in_task(OffloadedStmt *offload) {
...
+      if (map.find(arg->arg_id) != map.end()) {
+        map[arg->arg_id] = int(map[arg->arg_id]) | int(ExternalPtrAccess::READ);
+      } else {
+        map[arg->arg_id] = int(ExternalPtrAccess::READ);
+      }
+    }
+  }
+
+  void visit(GlobalStoreStmt *stmt) {
+    if (stmt->dest && stmt->dest->is<ExternalPtrStmt>()) {
```
k-ye, line 49: ditto

taichi/ir/transforms.h
```diff
@@ -114,6 +114,11 @@ void optimize_bit_struct_stores(IRNode *root,
                                 const CompileConfig &config,
                                 AnalysisManager *amgr);
 
+enum class ExternalPtrAccess : int { NONE = 0, READ = 1, WRITE = 2 };
+
+std::unordered_map<int, int> detect_external_ptr_access_in_task(
```
k-ye, line 119: Since this counts as an API, let's be explicit about the return types. Also would you add a docstring to briefly explain this (something like https://github.com/taichi-dev/taichi/blob/c2bdc82f8f62b74be93ba7df8e9c871b389fc477/taichi/ir/analysis.h#L66-L78) ? 

```suggestion
std::unordered_map<int, ExternqalPtrAccess> detect_external_ptr_access_in_task(
```

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -525,23 +572,33 @@ struct CompiledProgram::Impl {
   }
 
   void launch(Context &ctx, GLSLLauncher *launcher) const {
-    GLBufferTable &bufs = launcher->impl->user_bufs;
+    launcher->impl->gl_allocator.new_frame();
```
k-ye, line 575: nit: could you break this function into smaller pieces?

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -152,161 +155,194 @@ struct GLProgram {
...
+    }
   }
 
-  void bind_index(size_t index) const {
-    // SSBO index, is `layout(std430, binding = <index>)` in shader.
-    // We use only one SSBO though...
-    glBindBufferBase(GL_SHADER_STORAGE_BUFFER, index, id_);
-    check_opengl_error("glBindBufferBase");
+  void bind_to_index(GLuint location) {
+    bind_to_index(location, 0, size);
```
k-ye, line 187:
```suggestion
    bind_to_index(location, /*offset=*/0, size);
```

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -152,161 +155,194 @@ struct GLProgram {
...
   }
 
-  void bind_range(size_t index, size_t offset, size_t size) const {
-    glBindBufferRange(GL_SHADER_STORAGE_BUFFER, index, id_, offset, size);
-    check_opengl_error("glBindBufferRange");
+  void *map(GLenum access) {
+    return map_region(0, size, access);
```
k-ye, line 191:
```suggestion
    return map_region(/*offset=*/0, size, access);
```

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -152,161 +155,194 @@ struct GLProgram {
...
   }
 };
 
-struct GLBuffer : GLSSBO {
-  GLBufId index;
-  void *base;
-  size_t size;
+struct GLBufferAllocator {
+ private:
+  static constexpr size_t max_free_resident_size = 64 << 20;
```
k-ye, line 232: nit:
```suggestion
  static constexpr size_t kMaxFreeResidentSize = 64 << 20;
```

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -152,161 +155,194 @@ struct GLProgram {
...
 
-  GLBuffer(GLBufId index, void *base, size_t size)
-      : index(index), base(base), size(size) {
-    bind_data(base, size);
-    bind_index((int)index);
-  }
+  struct BufferKey {
+    GLbitfield access;
```
k-ye, line 235:
```suggestion
    GLbitfield access{0};
```

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -152,161 +155,194 @@ struct GLProgram {
...
+    size_t size;
```
k-ye, line 236:
```suggestion
    size_t size{0};
```

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -152,161 +155,194 @@ struct GLProgram {
...
+
+      if (free_resident_size > max_free_resident_size) {
+        GLBuffer *buf = pair.second;
+
+        // TI_WARN("Release {}", (void *)buf);
 
-  GLBuffer *get(GLBufId id) {
-    return bufs.at(id).get();
+        {
+          auto buffer_iter =
```
k-ye, line 270: nit: https://en.cppreference.com/w/cpp/algorithm/remove and https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom

k-ye (commented):
Thanks! I only have coding style suggestions

taichi/backends/opengl/opengl_api.cpp
bobcao3: It's kind of hard, I removed the special case handling for the one external ptr case (the logic are unified), but overall there's just a lot of code here to sync buffers in and out

k-ye (approved):
LGTM! Please help fix the nits, otherwise thanks for the huge contribution!

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -524,79 +578,151 @@ struct CompiledProgram::Impl {
     runtime->unmap();
   }
 
+  GLbitfield get_ext_arr_access(size_t &total_ext_arr_size) const {
+    GLbitfield access = 0;
+    for (const auto &[i, size] : ext_arr_map) {
+      total_ext_arr_size += size;
+      if ((ext_arr_access.at(i) & irpass::ExternalPtrAccess::READ) !=
+          irpass::ExternalPtrAccess::NONE) {
+        access |= GL_MAP_WRITE_BIT;
```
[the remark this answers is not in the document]
bobcao3: Because kernel reads external pointer which comes from the host, host will need to write to the buffer through mapping

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -524,79 +578,151 @@ struct CompiledProgram::Impl {
...
-        for (const auto &[i, size] : ext_arr_map) {
-          auto ptr = (void *)ctx.args[i];
-          saved_ctx_ptrs.push_back(ptr);
+      size_t total_ext_arr_size = 0;
+      GLbitfield access = get_ext_arr_access(total_ext_arr_size);
+
+      extr_buf = launcher->impl->gl_allocator.alloc_buffer(
+          total_ext_arr_size, nullptr, GL_SHADER_STORAGE_BUFFER, access);
+
+      void *baseptr = nullptr;
```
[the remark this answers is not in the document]
bobcao3: Ah this might explain the seg faults, kind of

bobcao3:
Can't seem to understand why CI is cancelled

k-ye (approved):
LGTM!!

PR #2534, opened 2021-07-16
[lang] Add a name parameter to fields

### LINKED ISSUE #2006: [lang] Add a "name" parameter to fields
Currently, if we write something like
```python
a = ti.field(ti.i32)
ti.root.dense(ti.i, 8).place(a)
```
, Taichi doesn't know that the field's name is `a`, and doesn't know that the `place` SNode's name is `a`.

So when users encounter some compilation error or developers want to print out some field's name, the Taichi compiler can only print out something like `S2place_i32`, which is not intuitive for debugging.

We can add a `name` parameter to fields (including `ti.field, ti.Vector.field, ti.Matrix.field`), and pass it into `SNode`s. Then we can acquire the field's name through a `SNode *` in the codebase. The code above can be modified into
```python
a = ti.field(ti.i32, name='a')
ti.root.dense(ti.i, 8).place(a)
```
for printing out the field's name.

**Additional comments**
This feature is especially useful when a Taichi program has too many fields and a developer wants to locate a specific one.

beginlner:
Related issue = close #2006

### DIFF
--- a/python/taichi/lang/expr.py
+++ b/python/taichi/lang/expr.py
@@ -142,6 +142,10 @@ class Expr(TaichiOperations):
         return self.ptr.get_raw_address()
 
     @property
+    def name(self):
+        return self.snode.name
+
+    @property
     def shape(self):
         if self.ptr.is_external_var():
             dim = impl.get_external_tensor_dim(self.ptr)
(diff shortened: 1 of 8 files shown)


ljcc0930:
Good work!

beginlner:
API: `<field>.name` or <field>.snode.name`

ljcc0930:
LGTM & thanks for your contribution! It would be great if you can add some tests for the new feature!

ljcc0930 (approved):

ljcc0930 (approved):
LGTM!

PR #2535, opened 2021-07-17
[Example] Use Taichi instead of numpy to perform tonemapping & significantly lower CPU's work

bobcao3:
cornell_box is thought to have significant CPU overhead issue. This is true, but only a part of the problem. The larger problem resides in the tonemapping code where a taichi field is transferred to a numpy field, where the mean and tonemapping is computed, and transferred into the GUI to display. Reducing all this chain into a taichi kernel brings significant performance improvement.

On RTX3080 + Ryzen 7 3700X, this specific example improved from 300 samples/s to 1000 samples/s

### DIFF
--- a/examples/rendering/cornell_box.py
+++ b/examples/rendering/cornell_box.py
@@ -1,6 +1,7 @@
 import time
 
 import numpy as np
+from numpy.lib.function_base import average
 
 import taichi as ti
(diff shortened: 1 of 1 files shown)


k-ye (approved):
LGTM!

PR #2536, opened 2021-07-17
[Example] Fix an oversight on the variance computation

bobcao3:
I computed the variance of the accumulation buffer without dividing by the number of frames, causing variance to only grow over time. (This is a very dumb commit, I'm sorry that the last one has this oversight)

Related issue = #


k-ye (approved):
Thanks!

Issue #2506, opened 2021-07-09
SNodeTree support destruction and memory recycling


ljcc0930:
Related issue = #2418

ljcc0930:
@beginlner good work!

PR #2558, opened 2021-07-20
[Lang] Support to destory a Snode tree and release memory

beginlner:
Related issue = #2506
API: `<FieldsBuilder>.destroy()`
To achieve it, I allocate a separate memory for a large size SNode tree. We can set the critical size manually by `ti.init(memory_allocate_critical_size=x)`.


beginlner:
\format

python/taichi/snode/fields_builder.py
```diff
@@ -139,3 +139,8 @@ def finalize(self):
     def _check_not_finalized(self):
         if self._finalized:
             raise InvalidOperationError('FieldsBuilder finalized')
+
+    def destroy(self):
```
k-ye, line 143: Consider adding this method to `SNodeTree` itself: The builder is for building the tree. What's the semantics of calling `destroy()` while it's not finalized?

taichi/program/program.cpp
```diff
@@ -509,6 +510,10 @@ void Program::materialize_snode_tree(SNodeTree *tree) {
   }
 }
 
+void Program::destroy_snode_tree(const int snode_tree_id) {
+  memory_pool->destroy_snode_tree(snode_tree_id);
```
k-ye, line 514: This is not the correct level of abstraction. Memory pool should only care about memory allocation. I'd suggest to have a  separate class to manage SNodeTree's memory.

k-ye (commented):
Thanks for the quick work! I think the abstraction level needs some improvenment. E.g. you can have a dedicated `SNodeTreeBufferManager` that uses `MemoryPool` and manages the allocated buffers. Let's keep the classes single responsibility.

taichi/program/program.cpp
beginlner: Thanks for your suggestions. I'll rewrite this part of code.

ljcc0930:
Moved to #2564

PR #2563, opened 2021-07-21
[ci] Allow build specific commit when uploading, used new building procedure.

Leonz5288:
Related issue = #

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -54,6 +54,7 @@ WORKDIR /taichi-dev
 ADD https://api.github.com/repos/taichi-dev/taichi/git/refs/heads/master version.json
 RUN git clone https://github.com/taichi-dev/taichi --branch=master
 WORKDIR ./taichi
+RUN git cat-file -t $COMMIT_SHA || exit 1
 RUN git checkout $COMMIT_SHA
 # Install Taichi's Python dependencies
 RUN $PYTHON -m pip install --user -r requirements.txt
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -73,4 +73,4 @@ RUN strings build/libtaichi_core.so | grep GLIBC
 RUN $PYTHON -m pip install --user twine
 RUN ti test -t 2
 WORKDIR python/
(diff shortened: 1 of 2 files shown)


ci/Dockerfile
```diff
@@ -51,29 +52,25 @@ WORKDIR /taichi-dev
 
 # Prevent docker caching when head changes
 ADD https://api.github.com/repos/taichi-dev/taichi/git/refs/heads/master version.json
-RUN git clone https://github.com/taichi-dev/taichi --depth=1 --branch=master
+RUN git clone https://github.com/taichi-dev/taichi --branch=master
+WORKDIR ./taichi
+RUN git checkout $COMMIT_SHA
```
k-ye, line 57: Could you do a validation on the SHA to see if its not empty? Otherwise abort with an error message

k-ye (commented):
This is awesome! I have one suggestion

ci/Dockerfile
```diff
@@ -51,29 +52,25 @@ WORKDIR /taichi-dev
...
 # Build Taichi from source
-RUN cd taichi && \
-    git submodule update --init --recursive --depth=1 && \
-    mkdir build
-WORKDIR /taichi-dev/taichi/build
-RUN cmake .. -DPYTHON_EXECUTABLE=$PYTHON -DCUDA_VERSION='10.0' -DTI_WITH_CUDA:BOOL=True -DTI_WITH_CC:BOOL=ON -DCMAKE_CXX_COMPILER=$CXX
-RUN make -j 8
+RUN git submodule update --init --recursive --depth=1
+WORKDIR /taichi-dev/taichi
+RUN $PYTHON setup.py develop
```
ailzhang, line 63: where do we use this dockerfile btw?  I'd hope we can switch it to `install` mode if it's user facing/for testing (instead of developer facing) :P But that can be a followup for sure

k-ye: It's used inside our Jenkins release pipeline :-) See https://github.com/taichi-dev/taichi/blob/e6093bd667737aabc2606294a353bcda3c3a29c8/Jenkinsfile#L89-L90

k-ye (approved):

PR #2574, opened 2021-07-23
[IR] Update CHI example to match SNode API

BillXu2000:
Update SNode API in CHI example code and fix array of variable length.


k-ye (approved):
LGTM!

PR #2575, opened 2021-07-23
[Lang] Change the data layout for vector fields with that needs grad

AmesingFlank:
Change the data layout for vector fields with `needs_grad = True`, so that the primal field always occupies a continuous region of memory.


k-ye (approved):
Thanks LGTM!

PR #2576, opened 2021-07-23
[ci] Correct windows build and test in CI

Leonz5288:
Related issue = #

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -220,7 +220,7 @@ jobs:
           python misc/ci_setup.py ci
         env:
           PYTHON: C:\hostedtoolcache\windows\Python\3.7.9\x64\python.exe
-          CI_SETUP_CMAKE_ARGS: -A x64 -DLLVM_DIR='C:\taichi_llvm\lib\cmake\llvm'
+          CI_SETUP_CMAKE_ARGS: -G "Visual Studio 16 2019" -A x64 -DLLVM_DIR=C:\taichi_llvm\lib\cmake\llvm
 
       - name: Test
         shell: powershell
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -217,10 +217,10 @@ jobs:
           clang --version
           cd D:\a\taichi\taichi
           python -m pip install --user numpy pybind11 gitpython
-          $env:CI_SETUP_CMAKE_ARGS = "-G'Visual Studio 16 2019' -A x64 -DPYTHON_EXECUTABLE='$env:PYTHON' -DLLVM_DIR='C:\taichi_llvm\lib\cmake\llvm'"
           python misc/ci_setup.py ci
         env:
           PYTHON: C:\hostedtoolcache\windows\Python\3.7.9\x64\python.exe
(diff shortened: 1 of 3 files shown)


misc/ci_setup.py
```diff
@@ -255,23 +255,25 @@ def run(self):
...
             execute_command('echo $PYTHONPATH')
-        elif get_os_name() != 'win':
+        else:
             # compile ..
             arg = environ.get('CI_SETUP_CMAKE_ARGS', '')
             os.makedirs('build', exist_ok=True)
-            execute_command(
-                f'TAICHI_CMAKE_ARGS="{arg}" {sys.executable} setup.py install --user'
-            )
+            if get_os_name() != 'win':
```
k-ye, line 262: set `TAICHI_CMAKE_ARGS={args}` as a new env variable, then we can unify this part?

.github/workflows/presubmit.yml
```diff
@@ -217,10 +217,10 @@ jobs:
           clang --version
           cd D:\a\taichi\taichi
           python -m pip install --user numpy pybind11 gitpython
-          $env:CI_SETUP_CMAKE_ARGS = "-G'Visual Studio 16 2019' -A x64 -DPYTHON_EXECUTABLE='$env:PYTHON' -DLLVM_DIR='C:\taichi_llvm\lib\cmake\llvm'"
           python misc/ci_setup.py ci
         env:
           PYTHON: C:\hostedtoolcache\windows\Python\3.7.9\x64\python.exe
+          CI_SETUP_CMAKE_ARGS: -A x64 -DLLVM_DIR='C:\taichi_llvm\lib\cmake\llvm'
```
k-ye, line 223: Did you drop `-G'Visual Studio 16 2019'` intentionally?

Leonz5288: Correct, this is done by default with the msbuild path added. Also, passing this with spaces leads to confusion when parsing cmake_args.

setup.py
```diff
@@ -127,7 +127,8 @@ def run(self):
         cmake_args += ['-DCMAKE_BUILD_TYPE=' + cfg]
 
         # Assuming Makefiles
-        build_args += ['--', f'-j{multiprocessing.cpu_count()}']
+        if get_os_name() != 'win':
```
ailzhang, line 130: sorry I'm not familiar with windows, why do we want to specialize this? ;)

Leonz5288: This is simply because MSBuild does not support "-j{number}" switch for multiprocessor build.

ailzhang (approved):
Looks great! Thanks @Leonz5288 !

setup.py
```diff
@@ -94,8 +94,9 @@ def run(self):
 class CMakeBuild(build_ext):
     def parse_cmake_args_from_env(self):
         # Source: TAICHI_CMAKE_ARGS=... python setup.py ...
+        import shlex
```
k-ye, line 97: This is great :-)

k-ye (approved):

k-ye:
Hmm, need to update Appveyor as well? (https://ci.appveyor.com/project/yuanming-hu/taichi/build/job/8j39ds90h03fli5r)

https://github.com/taichi-dev/taichi/blob/master/appveyor.yml

Issue #2579, opened 2021-07-25
Add Python docstring for public APIs


k-ye:
We will use the Google Python docstring style.

k-ye:
Thank you all for the contribution!

PR #2580, opened 2021-07-25
[Doc] Add Python docstring for SNode

k-ye:
Related issue = #2579

In addition to the docstring itself, this PR serves as an example for how to do class/method referencing.

### DIFF
--- a/python/taichi/lang/snode.py
+++ b/python/taichi/lang/snode.py
@@ -46,6 +46,7 @@ class SNode:
 
     def hash(self, axes, dimensions):
         """Not supported."""
+        raise RuntimeError('hash not yet supported')
         if isinstance(dimensions, int):
             dimensions = [dimensions] * len(axes)
         return SNode(self.ptr.hash(axes, dimensions))
@@ -119,14 +120,14 @@ class SNode:
 
         Args:
             *args (List[ti.field]): A list of Taichi fields to place.
-            offsest (tupe[numbers.Number]): Offset of the field domain.
+            offsest (Union[Number, tuple[Number]]): Offset of the field domain.
             shared_exponent (bool): Only useful for quant types.
 
         Returns:
             The `self` container.
         """
         if offset is None:
-            offset = []
+            offset = ()
         if isinstance(offset, numbers.Number):
             offset = (offset, )
         if shared_exponent:
--- a/python/taichi/lang/snode.py
+++ b/python/taichi/lang/snode.py
@@ -115,7 +115,7 @@ class SNode:
         return SNode(self.ptr.bit_array(axes, dimensions, num_bits))
 
     def place(self, *args, offset=None, shared_exponent=False):
-        """Places a list of Taichi fields under the `self` container. 
+        """Places a list of Taichi fields under the `self` container.
 
         Args:
             *args (List[ti.field]): A list of Taichi fields to place.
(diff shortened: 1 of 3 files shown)


python/taichi/lang/snode.py
```diff
@@ -12,52 +12,119 @@
...
-        return SNode(self.ptr.bit_array(indices, dimensions, num_bits))
+            dimensions = [dimensions] * len(axes)
+        return SNode(self.ptr.bit_array(axes, dimensions, num_bits))
 
     def place(self, *args, offset=None, shared_exponent=False):
+        """Places a list of Taichi fields under the `self` container.
+
+        Args:
+            *args (List[ti.field]): A list of Taichi fields to place.
+            offsest (tupe[numbers.Number]): Offset of the field domain.
```
rexwangcc, line 122: It seems `offset` as an argument is `numbers.Number` than a `Tuple`?

python/taichi/lang/snode.py
```diff
@@ -12,52 +12,119 @@
...
+        """
         if isinstance(dimensions, int):
-            dimensions = [dimensions] * len(indices)
-        return SNode(self.ptr.pointer(indices, dimensions))
+            dimensions = [dimensions] * len(axes)
+        return SNode(self.ptr.pointer(axes, dimensions))
 
-    def hash(self, indices, dimensions):
+    def hash(self, axes, dimensions):
+        """Not supported."""
```
rexwangcc, line 48: maybe we could add a deprecation warning later if this is not supported anymore?

rexwangcc (commented):
I left a few comments but in general looks good!

One minor question: is the reason we don't use type annotations in `python/taichi/lang/snode.py` that it requires extra `import List, ... from typing`, or that it simply hasn't happened yet?

k-ye:
> One minor question: is the reason we don't use type annotations in python/taichi/lang/snode.py that it requires extra import List, ... from typing, or that it simply hasn't happened yet?

It's a combination of me being lazy and the code hierarchy: I was planning to use the `Axis` type, but then ran into circular import issue...

rexwangcc:
> > One minor question: is the reason we don't use type annotations in python/taichi/lang/snode.py that it requires extra import List, ... from typing, or that it simply hasn't happened yet?
> 
> It's a combination of me being lazy and the code hierarchy: I was planning to use the `Axis` type, but then ran into circular import issue...

Ah, gotcha. I'm fine with leaving them as is then! (we could use forward references to bypass the circular import, but I guess we can solve it once and for all by refactoring the code later?)

python/taichi/lang/snode.py
k-ye: It's not exactly deprecated... Just stuck in a not supported mode when taichi switched to the LLVM backend

python/taichi/lang/snode.py
k-ye: Actually it's a `Union[Number, tuple[Number]]` :)

python/taichi/lang/snode.py
rexwangcc: I see!

rexwangcc (approved):
LGTM

k-ye:
> (we could use forward references to bypass the circular import, but I guess we can solve it once and for all by refactoring the code later?)

+1, I feel like the code needs some refactor before we have all the types we need for annotation

PR #2581, opened 2021-07-25
[opengl] Use OpenGL window visible hint instead of hiding after creation.

bobcao3:
This avoids a short popup everytime GL is initialized.

Tiny PR, no related issue


k-ye (approved):
Thanks!

PR #2582, opened 2021-07-26
[Vulkan] Add SPIR-V codegen on Vulkan backend

g1n0st:
Related issue = #2298


taichi/backends/vulkan/spirv.hpp
```diff
@@ -0,0 +1,2203 @@
+// Copyright (c) 2014-2020 The Khronos Group Inc.
```
k-ye, line 1: Let's remove this header and ask Vulkan devs to install it instead... https://github.com/KhronosGroup/SPIRV-Headers#how-to-install-the-headers

taichi/backends/vulkan/spirv_ir_builder.cpp
```diff
@@ -0,0 +1,940 @@
+#include "taichi/backends/vulkan/spirv_ir_builder.h"
+
+TLANG_NAMESPACE_BEGIN
```
k-ye, line 3: nit: Please use `namespace taichi { namespace lang {` instead of this macro. (same for `TLANG_NAMESPACE_END`)

k-ye (commented):
Amazing!

bobcao3:
This headers should be in Vulkan SDK already

taichi/backends/vulkan/spirv_ir_builder.cpp
```diff
@@ -0,0 +1,940 @@
...
+      ib_.begin(spv::OpLabel).add(merge_label).commit(&func_);
+      Value tmp14 = load_var(old_val, t_int32_);
+      Value tmp15 = new_value(t_fp32_, ValueKind::kNormal);
+      ib_.begin(spv::OpBitcast).add_seq(t_fp32_, tmp15, tmp14).commit(&func_);
+      ib_.begin(spv::OpReturnValue).add(tmp15).commit(&func_);
+      ib_.begin(spv::OpFunctionEnd).commit(&func_);
+      // function end
+    }
+
+    return float_atomic_add_;
```
bobcao3, line 674: This is a bit crazy, maybe have a few glsl runtime files that are pre-comiled to SPV? (VulkanSDK has glslc built-in, so this should be possible to do for taichi devs)

taichi/backends/vulkan/spirv_ir_builder.cpp
```diff
@@ -0,0 +1,940 @@
...
+    Value tmp2 = make_value(spv::OpShiftRightLogical, t_uint32_, tmp_w, _19u);
+    Value tmp3 = make_value(spv::OpBitwiseXor, t_uint32_, tmp_w, tmp2);
+    Value tmp4 = make_value(spv::OpShiftRightLogical, t_uint32_, tmp_t, _8u);
+    Value tmp5 = make_value(spv::OpBitwiseXor, t_uint32_, tmp_t, tmp4);
+    Value new_w = make_value(spv::OpBitwiseXor, t_uint32_, tmp3, tmp5);
+    store_variable(_rand_w_, new_w);
+    Value val = make_value(spv::OpIMul, t_uint32_, new_w, _1000000007u);
+    
+    return val;
+  }
```
bobcao3, line 702: ditto

taichi/backends/vulkan/spirv_ir_builder.cpp
```diff
@@ -0,0 +1,940 @@
...
+      init_random_function(global_tmp_);
+    }
+    
+    Value _1_4294967296f = float_immediate_number(t_fp32_, 1.0f / 4294967296.0f);    
+    Value tmp0 = rand_u32(global_tmp_);
+    Value tmp1 = cast(t_fp32_, tmp0);
+    Value val = mul(tmp1, _1_4294967296f);
+ 
+    return val;
+  }
```
bobcao3, line 715: ditto

taichi/backends/vulkan/spirv_ir_builder.cpp
```diff
@@ -0,0 +1,940 @@
...
+
+  Value IRBuilder::rand_i32(Value global_tmp_) {
+    if (!init_rand_) {
+      init_random_function(global_tmp_);
+    }
+  
+    Value tmp0 = rand_u32(global_tmp_);
+    Value val = cast(t_int32_, tmp0);
+    return val;
+  }
```
bobcao3, line 725: ditto

taichi/backends/vulkan/spirv_ir_builder.cpp
```diff
@@ -0,0 +1,940 @@
...
+    // how RAND_STATE changes, `gl_GlobalInvocationID.x` can still help
+    // us to set different seeds for different threads.
+    // Discussion: https://github.com/taichi-dev/taichi/pull/912#discussion_r419021918
+    Value tmp9 = load_var(rand_gtmp_, t_int32_);
+    Value tmp10 = new_value(t_int32_, ValueKind::kNormal);
+    ib_.begin(spv::OpIAdd).add_seq(t_int32_, tmp10, tmp9, _1).commit(&func_header_);
+    store_var(rand_gtmp_, tmp10);
+
+    init_rand_ = true;
+  }
```
bobcao3, line 935: same thing here (runtime should be in a glsl & precompiled)

taichi/backends/vulkan/spirv_ir_builder.cpp
g1n0st: One reason is SPIR-V is SSA form, we can't determine which SSA values would be compiled by shaderc so may have conflict with the values generated by our ir builder

taichi/backends/vulkan/codegen_vulkan.cpp
```diff
@@ -257,208 +312,365 @@ class TaskCodegen : public IRVisitor {
...
+    spirv::Value val;
+    if (ptr_to_buffers_.at(stmt->src) == BuffersEnum::Root) {
+      buffer_ptr = ir_->query_value(stmt->src->raw_name());
+      buffer_ptr.flag = spirv::ValueKind::kVariablePtr; // make this value could store/load
+    } else {
+      buffer_ptr = at_buffer(stmt->src, dt);
+    }
+    val = ir_->make_value(spv::OpBitcast, ir_->get_primitive_type(dt),
+          ir_->load_variable(buffer_ptr, ir_->i32_type()));
+    ir_->register_value(stmt->raw_name(), val);
```
bobcao3, line 339: If I'm understanding this correctly (loading a i32 value and cast it to the target primitive type), avoid using BitCast. GPU raw loads from structured buffers have built-in type conversion, so it is better to have multiple typed buffers binding to the same location (like the current GL backend)

taichi/backends/vulkan/codegen_vulkan.cpp
```diff
@@ -257,208 +312,365 @@ class TaskCodegen : public IRVisitor {
...
   void visit(ArgLoadStmt *stmt) override {
     const auto arg_id = stmt->arg_id;
     const auto &arg_attribs = ctx_attribs_->args()[arg_id];
     const auto offset_in_mem = arg_attribs.offset_in_mem;
     if (stmt->is_ptr) {
-      emit("// Pointer arg: id={} offset_in_mem={}", arg_id, offset_in_mem);
       // Do not shift! We are indexing the buffers at byte granularity.
-      emit("const int {} = {};", stmt->raw_name(), offset_in_mem);
+      spirv::Value val = ir_->int_immediate_number(ir_->i32_type(), offset_in_mem);
+      ir_->register_value(stmt->raw_name(), val);
```
bobcao3, line 349: nit: Maybe look into Vulkan's unique feature: PushConstants. PushConstants should be perfect for arguments (that is written to / updated almost every dispatch call, and small in size).

bobcao3:
We can have our SSA values in a "namespace" by setting a few of the id's
top bits, or we can have a pass to relocate the SSA values from the
precompiled runtime. There also might be capabilities of using multiple
binaries, but I'm not sure, I need to check.

bobcao3:
Right, there is an in-development SPIRV Linker in SPIR-V Tools. We can look
into that

taichi/backends/vulkan/codegen_vulkan.cpp
```diff
@@ -467,89 +679,187 @@ class TaskCodegen : public IRVisitor {
...
+      ir_->uint_immediate_number(ir_->u32_type(), 1),
+      ir_->uint_immediate_number(ir_->u32_type(), 0),
+      data);
+    } else {
       TI_ERROR("Vulkan only supports 32-bit atomic data types");
     }
-    // const dt stmt = atomicAdd(mem, val);
-    emit("const {} {} = {}({}, {});", opengl_data_type_name(dt),
-         stmt->raw_name(), func, mem, stmt->val->raw_name());
+    ir_->register_value(stmt->raw_name(), val);
```
bobcao3, line 703: nit: work-group parallel reduction should be possible in all compute shaders, maybe look into that as an potential optimization?

taichi/backends/vulkan/spirv_ir_builder.cpp
bobcao3: There is an in-development SPIRV Linker in SPIR-V Tools. We can look
into that to link our runtime SPV library to the kernels.

taichi/backends/vulkan/codegen_vulkan.cpp
bobcao3: We can leave a issue & comment here

taichi/backends/vulkan/spirv_ir_builder.h
```diff
@@ -0,0 +1,473 @@
...
+  Value const_i32_one_;
+
+  // Use float_atomic_add
+  Value float_atomic_add();
+  Value rand_u32(Value global_tmp_);
+  Value rand_f32(Value global_tmp_);
+  Value rand_i32(Value global_tmp_);
+
+  // Build Snode structure in SPIR-V
+  SType get_snode_struct(const CompiledSNodeStructs* compiled_structs, const SNodeDescriptor& sn_desc);
```
k-ye, line 388: nit: These are one level above the SPIR-V builder. Ideally we can have a SNode compiler that uses this IR-builder.

taichi/backends/vulkan/spirv_ir_builder.cpp
k-ye: In the GL backend, we used to have to "bake" in the buffer into the atomic helper, resulting in a few duplicates. (This is now mitigated by https://github.com/taichi-dev/taichi/pull/2486). One nice thing about this SPIR-V based helper is that we can pass in an Int32 pointer directly, so it only needs to be written once. 

IIRC, I once wrote the atomic helper first in GLSL, then compiled it to SPIR-V and replaced the embedded buffer with an Int32 pointer parameter, which worked just fine. Wonder if we can achieve this automatically?

taichi/backends/vulkan/codegen_vulkan.cpp
k-ye: Yeah, that requires the backend to support Taichi's Thread Local Storage optimization pass.

taichi/backends/vulkan/codegen_vulkan.cpp
k-ye: This is to make the atomics work for floating point numbers. Since we have generated a struct for each SNode (instead of manually computing the position in the global buffer), not using the bitcast means that we would have to re-generate the OpAccessChain twice, once for the target type and the other for i32.

That said, in the fture we can add another impl  that follows the GL's approach to see how it performs.

bobcao3: I think OpAccessChain may not actually compile to an instruction, same with bit cast. However it is more likely that bitcast is compiled into separate instruction instead of using the load store unit's free modifier.

taichi/backends/vulkan/spirv_ir_builder.h
g1n0st: Fixed

taichi/backends/vulkan/spirv_ir_builder.cpp
k-ye: Ack. Let's start with this, then migrate to the GLSL-based helpers.

k-ye (approved):
LGTM! Thank you for this great works!

A few things we should improve in the future:

1. I'd suggest to create a small interface for the result of the SNode struct compiler. This interface will know how to generate code for `SNodeLookupStmt` and `GetChStmt`. We can
then have two impls: the current struct-based one, and the other that does pointer arithmetic on a primitive arrays (much like the OpenGL backend). This should help us test if we can work around the bit-casts.
2. We should implement the runtime helpers in GLSL, then convert it to SPIR-V, so as to keep the code maintainable.

k-ye (approved):

PR #2591, opened 2021-07-27
[bug] recursively include examples in distribution script

### LINKED ISSUE #2588: taichi example missing
[Taichi] version 0.7.26, llvm 10.0.0, commit e37bdb5e, win, python 3.9.6

taichi examples only have {minimal}

victoriacity:
Related issue = close #2588 

Note: this will still not include CHI examples in PyPI distribution.


victoriacity:
Ailing,

If this one-liner change looks good then we can merge it to fix #2588. Thanks!

ailzhang (approved):
LGTM! Thanks @victoriacity !

ailzhang:
Btw since now we are testing in install mode, we should be able to test another example other than minimal so that similar bug can be caught in the future. I'm okay with adding the test in this PR or a followup :D

victoriacity:
> Btw since now we are testing in install mode, we should be able to test another example other than minimal so that similar bug can be caught in the future. I'm okay with adding the test in this PR or a followup :D

Sounds good! I would be inclined to directly adding the test into this PR.

PR #2592, opened 2021-07-27
[Doc] Add Python docstring in several places

strongoier:
Related issue = #2579

### DIFF
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -91,7 +91,6 @@ def is_extension_supported(arch, ext):
     return _ti_core.is_extension_supported(arch, ext)
 
 
-
 def reset():
     impl.reset()
     global runtime
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -76,16 +76,20 @@ def memory_profiler_print():
 
 
 extension = _ti_core.Extension
-is_extension_supported = _ti_core.is_extension_supported
-"""Checks whether an extension is supported on an arch.
 
-Args:
-    arg0: Specified arch.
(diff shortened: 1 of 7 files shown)


python/taichi/lang/__init__.py
```diff
@@ -77,6 +77,15 @@ def memory_profiler_print():
 
 extension = _ti_core.Extension
 is_extension_supported = _ti_core.is_extension_supported
+"""Checks whether an extension is supported on an arch.
```
k-ye, line 80: Let's use `def is_extension_supported(arch, ext)`...

python/taichi/lang/type_factory_impl.py
```diff
@@ -3,6 +3,10 @@
 
 
 class TypeFactory:
+    """Custom type factory for quantization.
```
k-ye, line 6: ditto

python/taichi/lang/quant_impl.py
```diff
@@ -5,6 +5,10 @@
 
 
 class Quant:
+    """Generator of quantized types.
```
k-ye, line 8: Could you also document the methods within this class?

python/taichi/lang/__init__.py
strongoier: done

python/taichi/lang/type_factory_impl.py
strongoier: done

python/taichi/lang/quant_impl.py
strongoier: done

k-ye (approved):
LGTM~

PR #2595, opened 2021-07-27
[opengl] Work group reduction

bobcao3:
- Reduces traffic to L2
- (In theory) better performance
- Did not use vendor specific warp/subgroup extensions
- Did not use the subgroup ballot extension
- Used a shared memory implementation (which is in L1)

### DIFF
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -671,6 +671,8 @@ class KernelGen : public IRVisitor {
            opengl_data_type_name(stmt->val->element_type()), val_name, op_name,
            opengl_data_type_name(stmt->val->element_type()),
            stmt->val->short_name());
+
+      emit("if (gl_LocalInvocationIndex == 0)");
     }
 
     if (dt->is_primitive(PrimitiveTypeID::i32) ||
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -645,6 +645,9 @@ class KernelGen : public IRVisitor {
 
     auto val_name = stmt->val->short_name();
 
+    emit("{} {};", opengl_data_type_name(stmt->val->element_type()),
+         stmt->short_name());
+
     if (stmt->is_reduction &&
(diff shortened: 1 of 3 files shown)


bobcao3:
Related PR #2487

bobcao3:
Actually a way easier option is this: https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_shader_subgroup.txt

That requires an extension, but it should be as simple as replacing the call to our own workgroup reduction function to the one from subgroup extension, and replacing the `if` check after it to check against `SubgroupLocalInvocationId`

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -627,6 +642,39 @@ class KernelGen : public IRVisitor {
...
+      }
+
+      emit("{} {} = reduction_workgroup_{}_{}({});",
+           opengl_data_type_name(stmt->val->element_type()), val_name, op_name,
+           opengl_data_type_name(stmt->val->element_type()),
+           stmt->val->short_name());
+
+      emit("if (gl_LocalInvocationIndex == 0)");
+    }
+
```
k-ye, line 677: nit: I think it's better to emit an additional pair of `{}` between L678-707. That way if we ever emit more than one line, it won't suddenly break  when workgroup reduction is enabled.

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -627,6 +642,39 @@ class KernelGen : public IRVisitor {
...
+          stmt->op_type == AtomicOpType::add) {
+        op_name = "add";
+      } else if (stmt->op_type == AtomicOpType::max) {
+        op_name = "max";
+      } else if (stmt->op_type == AtomicOpType::min) {
+        op_name = "min";
+      }
+
+      emit("{} {} = reduction_workgroup_{}_{}({});",
+           opengl_data_type_name(stmt->val->element_type()), val_name, op_name,
```
k-ye, line 671: I believe that by the time `val_name` is used, it is already defined due to the SSA form. Therefore this would be a re-definition. Consider using a different variable name, e.g. `auto val_name = stmt->val->short_name() + "_tmp";`?

k-ye (approved):
LGTM!

PR #2605, opened 2021-07-28
[vulkan] Draft: Remove shaderc on Windows & Add spirv-tools's optimization

bobcao3:
1. Removed shaderc dependency & the Glsl to SPIRV compiler on Windows. (shaderc binaries are linked with different windows runtime libraries and it seems we don't have a easy way to link it into taichi core)
2. Introduced SPIRV-Tools dependency (existing in Vulkan SDK, but the build script where it detects the library needs some work). We can now use its optimizers and validation tools to process our generated SPIR-V kernels.
3. Setup for adding the spirv linker for separate runtime routines written in GLSL

I can only test on Windows, I'm sure this will work fine on other platforms, but I'm not so sure about the cmake scripts for this. Can anyone with access to Linux & Metal (MoltenVK) test this? SPIR-V Tools should be a part of provided pre-built libraries within LunarG Vulkan SDK

### DIFF
--- a/taichi/backends/vulkan/runtime.cpp
+++ b/taichi/backends/vulkan/runtime.cpp
@@ -266,8 +266,8 @@ class ClearBufferCommandBuilder : private VulkanCommandBuilder {
 
 class VkRuntime ::Impl {
  private:
-  std::unique_ptr<spvtools::SpirvTools> _spirv_tools;
-  std::unique_ptr<spvtools::Optimizer> _spirv_opt;
+  std::unique_ptr<spvtools::SpirvTools> spirv_tools_;
+  std::unique_ptr<spvtools::Optimizer> spirv_opt_;
 
   spvtools::OptimizerOptions _spirv_opt_options;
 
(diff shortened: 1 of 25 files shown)


taichi/backends/vulkan/runtime.cpp
```diff
@@ -262,6 +265,31 @@ class ClearBufferCommandBuilder : private VulkanCommandBuilder {
 }  // namespace
 
 class VkRuntime ::Impl {
+ private:
+  std::unique_ptr<spvtools::SpirvTools> _spirv_tools;
```
k-ye, line 269: nit: s/_spirv_tools/spirv_tools_

taichi/backends/vulkan/runtime.cpp
```diff
@@ -262,6 +265,31 @@ class ClearBufferCommandBuilder : private VulkanCommandBuilder {
...
+  std::unique_ptr<spvtools::Optimizer> _spirv_opt;
```
k-ye, line 270: s/_spirv_opt/spirv_opt_

taichi/backends/vulkan/runtime.cpp
```diff
@@ -262,6 +265,31 @@ class ClearBufferCommandBuilder : private VulkanCommandBuilder {
...
+
+  spvtools::OptimizerOptions _spirv_opt_options;
+
+  static void spriv_message_consumer(spv_message_level_t level,
+                                     const char *source,
+                                     const spv_position_t &position,
+                                     const char *message) {
+    if (level <= SPV_MSG_FATAL) {
```
k-ye, line 278: We can add a macro, e.g. `TI_LOG_AT_LEVEL(lv, ...)` in the future

k-ye (approved):
LGTM!

PR #2608, opened 2021-07-29
[vulkan] Add SPIRV-Headers as a submodule

k-ye:
Related issue = #2605

### DIFF
--- a/taichi/backends/vulkan/spirv_header.h
+++ b/taichi/backends/vulkan/spirv_header.h
@@ -2,6 +2,6 @@
 
 #ifdef TI_WITH_VULKAN
 
-#include <SPIRV-Headers/spirv.hpp>
+#include <spirv/1.0/spirv.hpp>
 
 #endif  // TI_WITH_VULKAN
--- a/taichi/backends/vulkan/spirv_header.h
+++ b/taichi/backends/vulkan/spirv_header.h
@@ -2,6 +2,6 @@
 
 #ifdef TI_WITH_VULKAN
 
-#include <spirv/1.0/spirv.hpp>
+#include <spirv/unified1/spirv.hpp>
 
 #endif  // TI_WITH_VULKAN
--- a/taichi/backends/vulkan/vulkan_utils.h
+++ b/taichi/backends/vulkan/vulkan_utils.h
@@ -14,7 +14,7 @@ namespace vulkan {
 class VulkanEnvSettings {
  public:
   static constexpr uint32_t kApiVersion() {
-    return VK_API_VERSION_1_0;
+    return VK_API_VERSION_1_2;
   }
 };
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 	path = external/googletest
 	url = https://github.com/google/googletest
 	shallow = true
+[submodule "external/SPIRV-Headers"]
+	path = external/SPIRV-Headers
+	url = https://github.com/KhronosGroup/SPIRV-Headers
--- a/cmake/TaichiCore.cmake
+++ b/cmake/TaichiCore.cmake
@@ -212,17 +212,21 @@ if (TI_WITH_VULKAN)
     include_directories(${Vulkan_INCLUDE_DIR})
     target_link_libraries(${CORE_LIBRARY_NAME} ${Vulkan_LIBRARY})
 
+    # Is this the best way to include the SPIRV-Headers?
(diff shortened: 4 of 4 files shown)


taichi/backends/vulkan/vulkan_utils.h
```diff
@@ -15,30 +14,8 @@ namespace vulkan {
 class VulkanEnvSettings {
  public:
   static constexpr uint32_t kApiVersion() {
-    return VK_API_VERSION_1_2;
+    return VK_API_VERSION_1_0;
```
bobcao3, line 17: Isn't this bit a part of fix with SPV version? #2603

taichi/backends/vulkan/spirv_header.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#ifdef TI_WITH_VULKAN
+
+#include <spirv/1.0/spirv.hpp>
```
bobcao3, line 5: Is this a header from the SDK?

taichi/backends/vulkan/vulkan_utils.h
k-ye: That was an accidental change. We have some clients preferring to use the 1.0 version. 

Just to confirm, does VK 1.0 work for you?

taichi/backends/vulkan/spirv_header.h
k-ye: It's from https://github.com/KhronosGroup/SPIRV-Headers. According to https://github.com/KhronosGroup/glslang

> If people are only using this location to get spirv.hpp, I recommend they get that from SPIRV-Headers instead.

taichi/backends/vulkan/vulkan_utils.h
bobcao3: It does not if the SPIRV code gen spits out SPV 1.5. The target SPV version for VK 1.0 should be SPV 1.2 iirc. (They need to match otherwise the shader will not load)

taichi/backends/vulkan/spirv_header.h
bobcao3: I think the include path on Windows' SDK is `SPIRV-Headers/spirv.hpp`, same would be true if it's a git submodule

taichi/backends/vulkan/vulkan_utils.h
g1n0st: https://github.com/taichi-dev/taichi/blob/70566a74f27766d73ea16e723182765d92118846/taichi/backends/vulkan/spirv_ir_builder.cpp#L16-L22
Now include a `spirv.hpp` header with `SPV_VERSION=0x10000` may work properly, but the following solution should be better:
https://github.com/apache/tvm/commit/019275075151a300a8d3aaa15db16be62da3327c#diff-a5248d0972171c59770042a098d0097a90452de88efacc8be7d49013ed34a322

g1n0st: I can try if not in a hurry

g1n0st: Related issue=https://github.com/taichi-dev/taichi/issues/2612

taichi/backends/vulkan/spirv_header.h
k-ye: I don't think `spirv.hpp` is exposed as a  top-level header. Also if you install it to the default header search dir on Linux, it will be `/usr/local/include/spirv/1.0/spirv.hpp`.

bobcao3:
Maybe we should clone the SPIRV-Headers as submodule

g1n0st (approved):

k-ye:
> Maybe we should clone the SPIRV-Headers as submodule

Done

taichi/backends/vulkan/vulkan_utils.h
```diff
@@ -14,7 +14,7 @@ namespace vulkan {
 class VulkanEnvSettings {
  public:
   static constexpr uint32_t kApiVersion() {
-    return VK_API_VERSION_1_0;
+    return VK_API_VERSION_1_2;
```
g1n0st, line 17: Why 1.2 after added submodule?

k-ye: Sorry for the back and forth lol... After https://github.com/taichi-dev/taichi/pull/2615 we can set Taichi's default Vulkan version to 1.2

g1n0st (approved):

PR #2609, opened 2021-07-29
[Doc] Fix return type of Python Docstring in ops.py

Scoutydren:
Related issue = #

Fix return type of Python Docstring in ops.py


ljcc0930 (approved):
LGTM!

Issue #2612, opened 2021-07-30
[vulkan] [refactor] Add SPIRVSupport to query device requirements and configure extensions


g1n0st:
We have had a macro-based solution to check `SPV_VERSION` and generate the corresponding SPIR-V code on the Vulkan backend. But it is not platform-independent plus we cannot configure its behaviors. The next steps for this development are:
(1) to query driver/device support information from the target platform
(2) to pass these query parameters to Vulkan backend, and collect them in `SPIRVSupport` structure 
(3) to ensure correct SPIR-V shader generation even when some features are not supported

A verified solution is here: https://github.com/apache/tvm/commit/019275075151a300a8d3aaa15db16be62da3327c#diff-a5248d0972171c59770042a098d0097a90452de88efacc8be7d49013ed34a322

PR #2615, opened 2021-07-30
[vulkan] Detect Vulkan / SPIR-V version & enabling required features

bobcao3:
Issue = #2612

### DIFF
--- a/taichi/backends/vulkan/codegen_vulkan.cpp
+++ b/taichi/backends/vulkan/codegen_vulkan.cpp
@@ -49,6 +49,7 @@ class TaskCodegen : public IRVisitor {
  public:
   struct Params {
     OffloadedStmt *task_ir;
+    VkRuntime *runtime;
     const CompiledSNodeStructs *compiled_structs;
     const KernelContextAttributes *ctx_attribs;
     std::string ti_kernel_name;
@@ -61,11 +62,13 @@ class TaskCodegen : public IRVisitor {
         ctx_attribs_(params.ctx_attribs),
         task_name_(fmt::format("{}_t{:02d}",
                                params.ti_kernel_name,
-                               params.task_id_in_kernel)) {
+                               params.task_id_in_kernel)),
+        runtime_(params.runtime) {
     allow_undefined_visitor = true;
     invoke_default_visitor = true;
 
-    ir_ = std::make_shared<spirv::IRBuilder>();
+    ir_ =
+        std::make_shared<spirv::IRBuilder>(params.runtime->get_capabilities());
   }
 
   struct Result {
@@ -902,7 +905,16 @@ class TaskCodegen : public IRVisitor {
     std::array<int, 3> group_size = {
         task_attribs_.advisory_num_threads_per_group, 1, 1};
     ir_->set_work_group_size(group_size);
-    ir_->commit_kernel_function(kernel_function_, "main", {},
+    std::vector<spirv::Value> buffers;
+    if (runtime_->get_capabilities().spirv_version > 0x10300) {
+      for (const auto &bb : task_attribs_.buffer_binds) {
+        const auto it = buffer_value_map_.find(bb.type);
+        if (it != buffer_value_map_.end()) {
+          buffers.push_back(it->second);
+        }
+      }
+    }
+    ir_->commit_kernel_function(kernel_function_, "main", buffers,
                                 group_size);  // kernel entry
   }
 
@@ -1136,6 +1148,8 @@ class TaskCodegen : public IRVisitor {
     return continue_label_stack_.front();
   }
 
+  VkRuntime *runtime_;
+
   std::shared_ptr<spirv::IRBuilder> ir_;  // spirv binary code builder
   std::unordered_map<TaskAttributes::Buffers, spirv::Value> buffer_value_map_;
   std::unordered_map<TaskAttributes::Buffers, uint32_t> buffer_binding_map_;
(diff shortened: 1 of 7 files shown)


taichi/backends/vulkan/codegen_vulkan.cpp
```diff
@@ -49,6 +49,7 @@ class TaskCodegen : public IRVisitor {
  public:
   struct Params {
     OffloadedStmt *task_ir;
+    VkRuntime *runtime;
```
k-ye, line 52: Is it possible to pass in just the `VulkanCapabilities`? The reason is that we should have a clean separation between the runtime stuff and codegen stuff. (In the future I'd like to completely separate them into different folders, e.g. `runtime/` and `codegen/`

taichi/backends/vulkan/spirv_ir_builder.cpp
```diff
@@ -34,6 +25,10 @@ void IRBuilder::init_header(bool support_int8,
 
   // capability
   ib_.begin(spv::OpCapability).add(spv::CapabilityShader).commit(&header_);
+  // FIXME: What about devices don't support this?
+  ib_.begin(spv::OpCapability)
```
k-ye, line 29: That's a good question. Can we guard this somehow (at least by SPIR-V version?) We do need to put mobile platforms in mind here.

taichi/backends/vulkan/vulkan_api.cpp
```diff
@@ -191,7 +192,8 @@ void ManagedVulkanDevice::create_instance(const Params &params) {
   app_info.applicationVersion = VK_MAKE_VERSION(1, 0, 0);
   app_info.pEngineName = "No Engine";
   app_info.engineVersion = VK_MAKE_VERSION(1, 0, 0);
-  app_info.apiVersion = params.api_version;  // important
+  app_info.apiVersion =
+      VK_API_VERSION_1_2;  // The highest version designed to use
```
k-ye, line 196: Please still use `params.api_version` and update the API version there. I'd like to keep `vulkan_api` itself as platform-neutral as possible.

k-ye (commented):
Thank!!

taichi/backends/vulkan/vulkan_api.cpp
bobcao3: Actually what does params.api_version do? Is it the max API version to use or is it the exact API version used?

k-ye: >  is it the exact API version used

This

bobcao3:
So we are targeting Vulkan 1.0 and use extensions for everything else?

k-ye:
> So we are targeting Vulkan 1.0 and use extensions for everything else?

It's fine for Taichi to default at 1.2,  but ideally we should be able to support from 1.0 to 1.2. IIUC, https://github.com/taichi-dev/taichi/issues/2612 is a solution to help us achieve such support. @g1n0st ?

g1n0st:
> > So we are targeting Vulkan 1.0 and use extensions for everything else?
> 
> It's fine for Taichi to default at 1.2, but ideally we should be able to support from 1.0 to 1.2. IIUC, #2612 is a solution to help us achieve such support. @g1n0st ?

I think this PR is exactly this solution. But Taichi may need a compile option or a runtime option to determine the Vulkan runtime and SPIR-V version?

bobcao3:
The code in this PR does that. Vulkan 1.2 is requested, and according to
the standard either 1.1 or 1.2 will be reported back during detection. If
1.2 failed to init (which will only happen if only 1.0 is supported by the
loader), we will request 1.0 and the capabilities struct will have version
& related SPV version set to 1.0 & SPV 1.0. Variable pointer extension is
also detected in there, just that the codegen backend don't have a path to
generate code without that capability

bobcao3:
Actually there's a slight problem with this PR rn but I don't see it really solved (or needs to be solved) soon. Under Vulkan 1.0 API, all devices that support Vulkan will support Vulkan 1.0, with various extensions turned on or off. Under Vulkan 1.1+ API, an individual physical device can report its API support, and this support is recorded into the VulkanCapabilities structure. So this means this structure is per-device, and this means the codegen will need to know the device it's on. I don't see an elegant solution (or need for it) for decoupling them until we have an unified device API.

k-ye:
Thanks for the quick reply! I think we're on the same page now. I see why https://github.com/taichi-dev/taichi/pull/2615#discussion_r680293563 is confusing to you now. Here's what I think we can do:

1. rename that to `required_api_version`. However, make it an `std::optional`
2. When `required_api_version` is empty, do the auto-detection
3. Otherwise just use the `required_api_version`. This would allow our customers to ensure that Taichi does what they want.

g1n0st (approved):

taichi/backends/vulkan/codegen_vulkan.cpp
k-ye: Or at least leave a TODO here:

```suggestion
    // TODO: Pass in a VulkanCapabilities directly.
    VkRuntime *runtime;
```

taichi/backends/vulkan/runtime.h
```diff
@@ -46,6 +48,10 @@ class VkRuntime {
 
   void synchronize();
 
+#ifdef TI_WITH_VULKAN
```
k-ye, line 51: Shouldn't really need to guard this: `VkRuntime` is implemented using the C++ Pimpl pattern.

k-ye (approved):
LGTM, thx!

PR #2621, opened 2021-07-31
[vulkan] Use native struct types when supported & fix performance

bobcao3:
The current performance issue is entirely from the host/device transfer.

Originally, the `host_visible` memory pool is allocated actually on host & is cached. This means this chunk of memory is very fast when mapping (since no mapping is really there, it's host memory), but GPU has a hard time read / write on it.

Now the arguments and the return buffer are now on the `host_visible` and at the same time `device_local` memory. This is a special memory heap for sending arguments where host can directly write to it (and pay a small cost), and GPU can directly read/write off it. However this memory type is very slow when reading from host since it is not cached. (This is essentially the PCI-E BAR, and recently improved as ReBAR or SAM). So when we read back the return value, the context buffer is transferred from the device memory to a host memory, then mapped & copied out.

In cornell_box, when changing the interval to 5000 (making it almost entirely compute dominated), the Vulkan backend massively out-performed both the OpenGL and CUDA backend. (The CUDA side performance is a bit of a mystery, the advantage to OpenGL is perhaps we are compiling SNodes directly to structs)

On my RTX3080 + Ryzen 3700X + ReBAR enabled, the performance on `cornell_box.py` is roughly
OpenGL: 1500 samples/s, Vulkan 1900 samples/s, CUDA 1000 samples/s

This PR is dependent on #2615 (It is branched off there), we should merge the other one first and rebase this one to master later.

### DIFF
--- a/taichi/backends/vulkan/runtime.cpp
+++ b/taichi/backends/vulkan/runtime.cpp
@@ -436,8 +436,8 @@ class VkRuntime ::Impl {
     LinearVkMemoryPool::Params mp_params;
     mp_params.physical_device = managed_device_->physical_device();
     mp_params.device = managed_device_->device()->device();
-#pragma message("Vulkan memory pool size hardcoded to 64MB")
-    mp_params.pool_size = 64 * 1024 * 1024;
+#pragma message("Vulkan memory pool size hardcoded to 256MB")
+    mp_params.pool_size = 256 * 1024 * 1024;
     mp_params.required_properties = VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
     mp_params.compute_queue_family_index =
         managed_device_->queue_family_indices().compute_family.value();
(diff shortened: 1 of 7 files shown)


g1n0st (approved):

taichi/backends/vulkan/runtime.cpp
```diff
@@ -409,8 +434,8 @@ class VkRuntime ::Impl {
     LinearVkMemoryPool::Params mp_params;
     mp_params.physical_device = managed_device_->physical_device();
     mp_params.device = managed_device_->device()->device();
-#pragma message("Vulkan memory pool size hardcoded to 64MB")
-    mp_params.pool_size = 64 * 1024 * 1024;
+#pragma message("Vulkan memory pool size hardcoded to 256MB")
```
k-ye, line 437: 😂

taichi/backends/vulkan/vulkan_api.h
```diff
@@ -199,24 +212,19 @@ class VulkanCommandBuilder {
...
-  using VulkanCommandBuilder::VulkanCommandBuilder;
-
-  void append(const VulkanPipeline &pipeline, int group_count_x);
-};
-
-enum class VulkanCopyBufferDirection {
-  H2D,
-  D2H,
-  // D2D does not have a use case yet
+  VkDevice device;  // do not own
```
k-ye, line 227: We should probably get a linter..

```suggestion
  VkDevice device_;  // do not own
```

k-ye (approved):

PR #2623, opened 2021-08-01
[vulkan] Dynamically load Vulkan

AmesingFlank:
This PR uses [volk](https://github.com/zeux/volk) to dynamically load vulkan functions. This ensures that users who do not have vulkan installed can still use taichi.

### DIFF
--- a/taichi/backends/vulkan/loader.cpp
+++ b/taichi/backends/vulkan/loader.cpp
@@ -13,7 +13,7 @@ VulkanLoader::VulkanLoader() {
 }
 
 bool VulkanLoader::init() {
-  std::call_once(init_flag_, [&](){ 
+  std::call_once(init_flag_, [&]() {
     if (initialized) {
       return;
     }
--- a/taichi/backends/vulkan/loader.cpp
+++ b/taichi/backends/vulkan/loader.cpp
@@ -10,24 +10,25 @@ namespace lang {
 namespace vulkan {
 
 VulkanLoader::VulkanLoader() {
-  initialized = false;
 }
 
 bool VulkanLoader::init() {
-  if (initialized) {
-    return true;
-  }
-  VkResult result = volkInitialize();
-  initialized = result == VK_SUCCESS;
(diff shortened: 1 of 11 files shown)


k-ye (approved):
Great LGTM!

taichi/backends/vulkan/loader.h
```diff
@@ -0,0 +1,34 @@
...
+ public:
+  VulkanLoader(VulkanLoader const &) = delete;
+  void operator=(VulkanLoader const &) = delete;
+
+  void load_instance(VkInstance instance_);
+  void load_device(VkDevice device_);
+  bool init();
+
+ private:
+  bool initialized;
```
k-ye, line 24: nit here and below: add `_` suffix to these variable names

taichi/backends/vulkan/loader.cpp
```diff
@@ -0,0 +1,36 @@
...
+namespace taichi {
+namespace lang {
+namespace vulkan {
+
+VulkanLoader::VulkanLoader() {
+  initialized = false;
+}
+
+bool VulkanLoader::init() {
+  if (initialized) {
```
k-ye, line 17: Use https://en.cppreference.com/w/cpp/thread/call_once ?

AmesingFlank: Hmm, is this mainly because we want to ensure thread safety?

k-ye: exactly :-)

taichi/backends/vulkan/loader.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace taichi {
+namespace lang {
+namespace vulkan {
+
+VulkanLoader::VulkanLoader() {
+}
+
+bool VulkanLoader::init() {
+  std::call_once(init_flag_, [&]() {
```
k-ye, line 16: Do we still need `initialized`?

Issue #285, opened 2019-11-27
External arrays with types other than float32


yuanming-hu:

yuanming-hu:
Done.

PR #2628, opened 2021-08-02
[vulkan] Move SPIR-V type capabilities into VulkanCapabilities & fix compilation errors

g1n0st:
Related issue = #2298

### DIFF
--- a/taichi/backends/vulkan/spirv_ir_builder.cpp
+++ b/taichi/backends/vulkan/spirv_ir_builder.cpp
@@ -5,10 +5,7 @@ namespace lang {
 namespace vulkan {
 
 namespace spirv {
-void IRBuilder::init_header(bool support_int8,
-                            bool support_int16,
-                            bool support_int64,
-                            bool support_fp64) {
+void IRBuilder::init_header() {
   TI_ASSERT(header_.size() == 0U);
   header_.push_back(spv::MagicNumber);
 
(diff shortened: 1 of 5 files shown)


taichi/backends/vulkan/vulkan_api.h
```diff
@@ -81,21 +83,29 @@ class VulkanDevice {
...
   VulkanDeviceDebugStruct *debug_struct_{nullptr};
+#endif
   Params rep_;
 };
 
 struct VulkanCapabilities {
   uint32_t api_version;
   uint32_t spirv_version;
 
+  bool has_int16;
```
k-ye, line 105: default initializers

g1n0st: Fixed

k-ye (approved):
LGTM!

taichi/backends/vulkan/spirv_ir_builder.cpp
```diff
@@ -39,20 +36,13 @@ void IRBuilder::init_header(bool support_int8,
         .commit(&header_);
   }
 
-  this->support_int8_ = support_int8;
-  if (support_int8) {
-    ib_.begin(spv::OpCapability).add(spv::CapabilityInt8).commit(&header_);
-  }
-  this->support_int16_ = support_int16;
-  if (support_int16) {
+  if (vulkan_cap_.has_int16) {
```
bobcao3, line 39: Maybe keep int8 types? GPUs like AMD's RDNA2 has a very good int8 performance.

g1n0st: `Int8` should enable the `Kernel` capability in SPIR-V

bobcao3: We can still have 8 bit storage and pointers to 8 bit elements through `VK_KHR_8bit_storage` and `SPV_KHR_8bit_storage`. I'm not sure whether we should mark it as support int8 tho since its sole purpose is storage not registers. I guess this would be fine.

bobcao3 (approved):

taichi/backends/vulkan/spirv_ir_builder.cpp
bobcao3: Ah also Kernel capability is not a requirement now. According to SPIR-V spec:
"Khronos SPIR-V Issue #285: Remove Kernel as a capability implicitly declared by Int8."

As long as the hardware support it we should enable it

g1n0st: It makes sense. I'll fix it.

g1n0st: Fixed

PR #2633, opened 2021-08-03
[Bug] Fix osx release workflow.

ailzhang:
This PR does 2 things: 
- Fix path issues in osx release workflow and stop using `TAICHI_REPO_DIR`
- Install from the wheel and run tests (instead of running tests in dev mode). 

There's a minor followup:
- Our changelog requires installing taichi but building taichi wheel requires making changelog. So currently we keep installing taichi using `python setup.py install` to first build the wheel and then uninstall it. Ideally this can be removed so that we just build a wheel -> install the wheel -> test taichi installed from wheel.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,31 +38,29 @@ jobs:
         env:
           CI_PLATFORM: ${{ matrix.os }}
 
-      - name: Build
+      - name: Create Python Wheel
         run: |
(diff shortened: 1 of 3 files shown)


k-ye:
> Our changelog requires installing taichi 

You're right. Several commands should just be moved out from `ti` ...

k-ye (approved):
Thank you LGTM!

I think it's time for us to think about how we can better test these release scripts.

PR #2630, opened 2021-08-02
[Doc] Add docstring for AutoDiff part

ljcc0930:
Related issue = close #2618 
#2579

PR #2635, opened 2021-08-04
[Refactor] Split transformer.py into StmtBuilder and ExprBuilder (Stage 2)

xumingkuan:
Related issue = #2193

Changeset:
- Deprecate `maybe_transform_ti_func_call_to_stmt` and split its logic into StmtBuilder and ExprBuilder
- Remove the pass transforming function calls to `maybe_transform_ti_func_call_to_stmt`
- Add a method `insert_expr_stmt_if_ti_func` to check if a function is Taichi function at Python runtime (we cannot check it in StmtBuilder). This method is called only when `experimental_real_function=True` and the Python statement contains only one expression which is the function call. So there will be no redundant calls like `maybe_transform_ti_func_call_to_stmt` before this PR.
- Remove an unused field `is_classfunc`
- Rename `ASTTransformer` to `ASTTransformerTotal`

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -188,6 +188,32 @@ def chain_compare(comparators, ops):
     return ret
 
 
+@taichi_scope
+def insert_expr_stmt_if_ti_func(func, *args, **kwargs):
+    """
+    This method is used only for real functions. It inserts a FrontendExprStmt
+    to the C++ AST to hold the function call if |func| is a Taichi function.
+    :param func: The function to be called.
+    :param args: The arguments of the function call.
+    :param kwargs: The keyword arguments of the function call.
+    :return: The return value of the function call if it's a non-Taichi
+    function. Returns None if it's a Taichi function.
+    """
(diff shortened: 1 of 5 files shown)


xumingkuan:
Tests related to experimental_real_function failing. Debugging...

python/taichi/lang/impl.py
```diff
@@ -189,32 +189,29 @@ def chain_compare(comparators, ops):
...
-            f'Calling into non-Taichi function {ti_func.__name__}.'
-            ' This means that scope inside that function will not be processed'
-            ' by the Taichi transformer. Proceed with caution! '
-            ' Maybe you want to decorate it with @ti.func?',
-            UserWarning,
-            stacklevel=2)
-
-    is_taichi_function = getattr(ti_func, '_is_taichi_function', False)
+def insert_expr_stmt_if_ti_func(func, *args, **kwargs):
+    """
```
k-ye, line 193: Great comment! But please use Google's docstring style. E.g. https://github.com/taichi-dev/taichi/pull/2630/files

k-ye (approved):
LGTM + 1 nit!

PR #2636, opened 2021-08-04
[bug] Fix missing ti.template() in rand_vector(n) in examples

### LINKED ISSUE #2617: examples math_utils: rand_vector(n) broken, 'Expr' object cannot be interpreted as an integer
**Describe the bug**
`tree_gravity.py` example seems broken, I get an error here:
https://github.com/taichi-dev/taichi/blob/7a2d5d11d4d03d0d306e1a1c4a131ef65b8e5232/examples/simulation/tree_gravity.py#L132-L134

**To Reproduce**
```sh
$ cd taichi
$ git log -n 1 --oneline --no-decorate
7a2d5d11 [Lang] [refactor] Move SNode property calculation from materialization time to construction time (#2600)
$ python3 -m venv ./venv
$ source ./venv/bin/active
$ pip install -U pip wheel
$ pip install taichi==0.7.26  # I think this is sufficient
$ python3 examples/simulation/tree_gravity.py
[Taichi] mode=release
[Taichi] preparing sandbox at /tmp/taichi-u3n9wd_j
[Taichi] version 0.7.26, llvm 10.0.0, commit e37bdb5e, linux, python 3.6.9
[Taichi] Starting on arch=x64
...
# Click on GUI
...
  File "./tree_gravity.py", line 135, in add_particle_at
    rand_vec = rand_vector(2)
  File ".../taichi/venv_prebuilt/lib/python3.6/site-packages/taichi/lang/kernel_impl.py", line 50, in decorated
    return fun.__call__(*args)
  File ".../taichi/venv_prebuilt/lib/python3.6/site-packages/taichi/lang/kernel_impl.py", line 122, in __call__
    ret = self.compiled(*args)
  File ".../taichi/examples/simulation/math_utils.py", line 11, in rand_vector
    return ti.Vector([ti.random() for _ in range(n)])
TypeError: 'Expr' object cannot be interpreted as an integer
```

**Log/Screenshots**
See above

**Additional comments**
It works if I just inline the logic for `rand_vector()`, or manually expand, i.e.:
```py
# rand_vec = rand_vector(2)  # Does not work (see above)
rand_vec = ti.Vector([ti.random() for _ in range(2)])  # Works
rand_vec = ti.Vector([ti.random(), ti.random()]  # Works
```

xumingkuan:
Related issue = fix #2617

I think we should also update the docs about the restrictions on list comprehensions and dictionary comprehensions but I don't know which page to put them...

### DIFF
--- a/examples/simulation/math_utils.py
+++ b/examples/simulation/math_utils.py
@@ -4,7 +4,7 @@ import taichi as ti
 
 
 @ti.func
-def rand_vector(n):
+def rand_vector(n: ti.template()):
     '''
     Samples a n-dimensional random uniform vector.
     '''


victoriacity (commented):
LGTM! Perhaps we'd also like to add this type of `ti.template()` usage into the documentation into the [Kernels and Functions](https://docs.taichi.graphics/docs/lang/articles/basic/syntax#advanced-arguments-1) page.

For usage of list and dictionary comprehensions I would suggest stating in the doc page that list and dictionary comprehensions are strictly python-scoped, i.e., they need to be either outside Taichi kernels or be compile-time Python-scope constants within the kernel.

victoriacity (approved):

PR #2638, opened 2021-08-05
[Lang] [refactor] Add Field classes for ti.field/ti.Vector.field/ti.Matrix.field

### LINKED ISSUE #2499: ti.field() returns a "field" class instead of taichi.lang.expr.Expr

strongoier:
Related issue = close #2499 

Previously, a `ti.field` is a `Expr` with `is_global()=True`, while a `ti.Matrix.field` is a `Matrix` with `is_global()=True`. This is not intuitive. After this PR, `ti.field` will become a `ScalarField`, and `ti.Matrix.field` will become a `MatrixField`, which both inherit from a super class `Field`.

The `Field` class keeps the original implementation way of using SNodes. A field is constructed by a list of field members. For example, a scalar field has 1 field member, while a 3x3 matrix field has 9 field members. A field member is a Python `Expr` wrapping a C++ `GlobalVariableExpression`. A C++ `GlobalVariableExpression` wraps the corresponding SNode.

All methods belonging to `ti.field`/`ti.Matrix.field` have been migrated from `Expr`/`Matrix` classes to `ScalarField`/`MatrixField` classes. If a method is implemented exactly the same in both classes, it will be put into `Field` class. Otherwise common methods will be marked as `NotImplemented` in `Field` class to avoid misuse.

During migration, the property methods and meta (fill, to/from numpy/torch, copy) methods are mainly kept the same with slight cleaning in both implementation and docstrings. The biggest change is around subscript (both in Python scope and Taichi scope) - the main idea is that for `MatrixField`, after first subscript with coordinates, we should get a `Matrix`, where each element is under the same state as if we are accessing a corresponding `ScalarField`. It acts just like a `Matrix` so no more `Proxy` classes are needed. By the way, `Matrix.with_entries()` is just added for this purpose.

Note:
- Previously, when `x` is a field with 0-D shape, although deprecated, it can still be accessed via `x` instead of `x[None]`. After this PR this invalid behavior will not work any more.
- Deprecated interfaces of these two classes are removed.


python/taichi/lang/field.py
```diff
@@ -0,0 +1,459 @@
...
+    def __getitem__(self, key):
+        self.initialize_host_accessors()
+        return self.host_accessors[0].getter(*self.pad_key(key))
+
+    def __repr__(self):
+        # make interactive shell happy, prevent materialization
+        return '<ti.field>'
+
+
+class MatrixField(Field):
```
k-ye, line 275: nit: I feel like it's better to put this inside `matrix.py`. WDYT?

python/taichi/lang/field.py
```diff
@@ -0,0 +1,459 @@
...
+        n (Int): Number of rows.
+        m (Int): Number of columns.
+    """
+    def __init__(self, vars, n, m):
+        assert len(vars) == n * m
+        super().__init__(vars)
+        self.n = n
+        self.m = m
+
+    def get_scalar_field(self, *indices):
```
k-ye, line 289: Hmm, wehave to keep `__call__()` for a while to help ppl migrate, if necessary. Consider adding `__call__()` and mark it as deprecated?

k-ye (commented):
Awesome! Could you briefly write up what's been added/refactored to ease the review process? An excellent example is https://github.com/taichi-dev/taichi/pull/2495

strongoier:
> Awesome! Could you briefly write up what's been added/refactored to ease the review process? An excellent example is #2495 (which is not that "brief"...)

Updated.

python/taichi/lang/field.py
strongoier: Maybe? Then people can add more types and add a corresponding field class in the same file?

python/taichi/lang/field.py
strongoier: What about the test cases where I replaced the `__call__()`?

python/taichi/lang/field.py
k-ye: I think it's natural to put different field impls in separate files.

strongoier: Yes it makes sense. Then is it clearer to have a `field.py`, a `scalar_field.py`, and `matrix_field.py`?

k-ye (approved):
LGTM! Thanks for the huge efforts!

python/taichi/lang/field.py
k-ye: Fine to update the test cases

python/taichi/lang/field.py
victoriacity: I'd also inclined to putting each type of field in a separate file.

victoriacity (approved):
LGTM! This will make #2627 much clearer than what it is now. It will also facilitate the implementation of recursive custom structs.

python/taichi/lang/field.py
strongoier: `MatrixField` is now put in `matrix.py` to avoid potential circular import.

victoriacity:
P.S. It looks like [taichi_glsl defines extensions of the current Matrix/Expr classes](https://github.com/taichi-dev/taichi_glsl/blob/master/taichi_glsl/classes.py). I'm not sure if the package will still be fully functional after the changes are merged.

k-ye:
>  I'm not sure if the package will still be fully functional after the changes are merged.

Thanks for catching this! Let's get this in first to move other things forward :-)

Issue #2634, opened 2021-08-04
Unify the Windows CI platform


k-ye:
Let's aim for deprecating Appveyor. Also we should trigger the nightly release more often to ensure the entire release pipeline works well.

Leonz5288:
Now windows & macos both release through github action. I guess there is still one more place that uses appveyor?

k-ye:
> I guess there is still one more place that uses appveyor?

Yeah, every PR-merge triggers a appveyor build. Once we have a cron job to release nightly, we can deprecate that

Leonz5288:
> > I guess there is still one more place that uses appveyor?
> 
> Yeah, every PR-merge triggers a appveyor build. Once we have a cron job to release nightly, we can deprecate that

Great! I think we are very close to this!

PR #2641, opened 2021-08-06
[ci] Fix mac release and integrate windows release into github

Leonz5288:
Related issue = #2634

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,8 +5,8 @@ on:
     types: [published]
 
 jobs:
-  build_and_test_cpu:
-    name: Build and Test (macOS only)
+  build_and_upload_mac:
+    name: Build and Upload (macOS only)
     strategy:
       matrix:
         include:
@@ -25,6 +25,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python }}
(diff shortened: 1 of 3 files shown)


python/build.py
```diff
@@ -42,9 +42,8 @@ def build(project_name):
             f'cd ..; PROJECT_NAME={project_name} {get_python_executable()} setup.py bdist_wheel -p manylinux1_x86_64'
         )
     else:
-        os.system(
-            f'cd ..; PROJECT_NAME={project_name} {get_python_executable()} setup.py bdist_wheel'
-        )
+        os.environ['PROJECT_NAME'] = f'{project_name}'
```
k-ye, line 45:
```suggestion
        os.environ['PROJECT_NAME'] = project_name
```

k-ye (approved):
LGTM!

.github/workflows/release.yml
```diff
@@ -5,8 +5,8 @@ on:
     types: [published]
```
k-ye, line 5: I hope we can trigger the nightly path for each merge-to-`master` event, and the official build path for a release. Is it possible to achieve that?

Leonz5288: Yes, I'm working on this. Seems that time-based nightly release and merge-based nightly release can both work.

.github/workflows/release.yml
```diff
@@ -45,38 +48,90 @@ jobs:
...
+          python setup.py develop
 
-      - name: Install Wheel and Test
+      - name: Test
         run: |
-          TAICHI_REPO_DIR=`pwd`
-          export PATH=$TAICHI_REPO_DIR/taichi-llvm/bin/:$PATH
-          NUM_WHL=`ls dist/*.whl | wc -l`
-          if [ $NUM_WHL -ne 1 ]; then echo 'ERROR: created more than 1 whl.' && exit 1 ; fi
-          pip install dist/*.whl
```
ailzhang, line 60: Oh sorry for the late review, we actually want to preserve this logic :P Since we want to test a taichi installed from our built wheel (python setup.py develop is local dev mode which might slightly differ from the content inside the wheel). So we need another PR to bring this logic back. 
In fact it's indeed better to unify all the CI to first build the wheel and test from there. (compared to testing in dev mode).

PR #2644, opened 2021-08-07
[vulkan] Use VulkanMemoryAllocator for memory allocation

bobcao3:
Ditched our own linear memory pool. VulkanMemoryAllocator will handle the edge cases much better than our own solution.


k-ye (approved):
Great, LGTM!

PR #2645, opened 2021-08-09
[gui] GGUI 1/n: Add necessary cuda structs/enums/functions

AmesingFlank:
Related Issue = #2646

This is the 1st of a series of PRs that adds a GPU-based GUI to taichi. This PR adds the necessary cuda structs, enums, and functions.

### DIFF
--- a/taichi/common/serialization.h
+++ b/taichi/common/serialization.h
@@ -203,7 +203,7 @@ class Serializer {
   };
 };
 
-inline std::vector<uint8> read_data_from_file(const std::string &fn) {
+inline std::vector<uint8_t> read_data_from_file(const std::string &fn) {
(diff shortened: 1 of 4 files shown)


taichi/common/serialization.h
```diff
@@ -203,7 +203,7 @@ class Serializer {
   };
 };
 
-inline std::vector<uint8> read_data_from_file(const std::string &fn) {
```
AmesingFlank, line 206: I think this was a typo in the original code, because `uint8_t` is used everywhere else. This will start to cause compile errors once we `enable_language(CUDA)` in cmake.

k-ye (approved):
LGTM!

Issue #2646, opened 2021-08-09
GPU based GUI system


k-ye:
**Concisely describe the proposed feature**

Taichi's existing `ti.GUI` system is slow and a bit hard to maintain. Let's upgrade it to Vulkan :-)

**Describe the solution you'd like (if any)**

See https://github.com/taichi-dev/docs.taichi.graphics/pull/68 for the complete design

PR #2648, opened 2021-08-09
[gui] GGUI 2/n: Add optional graphics queue, compute queue, and surface to EmbeddedVulkanDevice

AmesingFlank:
Related issue = #2646 

This is the 2nd of a series of PRs that adds a GPU-based GUI to taichi. This PR adds an optional graphics queue, compute queue, and surface to EmbeddedVulkanDevice.

### DIFF
--- a/taichi/backends/vulkan/loader.cpp
+++ b/taichi/backends/vulkan/loader.cpp
@@ -4,6 +4,7 @@
 #include <volk.h>
 
 #include "taichi/backends/vulkan/loader.h"
+#include "taichi/common/logging.h"
 
 namespace taichi {
 namespace lang {
@@ -35,10 +36,7 @@ void VulkanLoader::load_device(VkDevice device) {
 PFN_vkVoidFunction VulkanLoader::load_function(const char *name) {
   auto result =
       vkGetInstanceProcAddr(VulkanLoader::instance().vulkan_instance_, name);
-  // printf("loading %s \n",name);
-  if (result == nullptr) {
-    printf("%s is nullptr\n", name);
-  }
+  TI_WARN_IF(result == nullptr, "loaded vulkan function {} is nullptr", name);
   return result;
 }
 
--- a/taichi/backends/vulkan/loader.cpp
+++ b/taichi/backends/vulkan/loader.cpp
@@ -32,6 +32,16 @@ void VulkanLoader::load_device(VkDevice device) {
   volkLoadDevice(device);
 }
 
+PFN_vkVoidFunction VulkanLoader::load_function(const char *name) {
+  auto result =
+      vkGetInstanceProcAddr(VulkanLoader::instance().vulkan_instance_, name);
+  // printf("loading %s \n",name);
+  if (result == nullptr) {
+    printf("%s is nullptr\n", name);
+  }
+  return result;
+}
+
 }  // namespace vulkan
 }  // namespace lang
 }  // namespace taichi
--- a/taichi/backends/vulkan/vulkan_api.cpp
+++ b/taichi/backends/vulkan/vulkan_api.cpp
@@ -156,7 +156,10 @@ VulkanQueueFamilyIndices find_queue_families(VkPhysicalDevice device,
 bool is_device_suitable(VkPhysicalDevice device, VkSurfaceKHR surface) {
   auto indices = find_queue_families(device, surface);
   if (surface != VK_NULL_HANDLE) {
-    return indices.is_complete_for_ui();
+    // this means we need ui
+    VkPhysicalDeviceFeatures features{};
+    vkGetPhysicalDeviceFeatures(device, &features);
+    return indices.is_complete_for_ui() && features.wideLines == VK_TRUE;
   } else {
     return indices.is_complete();
   }
--- a/taichi/backends/vulkan/vulkan_api.cpp
+++ b/taichi/backends/vulkan/vulkan_api.cpp
@@ -450,8 +450,6 @@ void EmbeddedVulkanDevice::create_logical_device() {
   capability_.has_int64 = true;
   capability_.has_float64 = true;
   if (params_.is_for_ui) {
-    device_features.samplerAnisotropy = VK_TRUE;
-    device_features.geometryShader = VK_TRUE;
     device_features.wideLines = VK_TRUE;
   }
 
--- a/taichi/backends/vulkan/vulkan_api.cpp
+++ b/taichi/backends/vulkan/vulkan_api.cpp
@@ -99,7 +99,8 @@ std::vector<const char *> get_required_extensions() {
   return extensions;
 }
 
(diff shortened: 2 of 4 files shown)


taichi/backends/vulkan/vulkan_api.cpp
```diff
@@ -400,6 +449,11 @@ void EmbeddedVulkanDevice::create_logical_device() {
   capability_.has_int16 = false;
   capability_.has_int64 = true;
   capability_.has_float64 = true;
+  if (params_.is_for_ui) {
+    device_features.samplerAnisotropy = VK_TRUE;
+    device_features.geometryShader = VK_TRUE;
```
bobcao3, line 454: Don't enable geometry shaders. They are a legacy pipeline stage.

bobcao3:
Can you somehow incorporate the debug swapchain into this? We don't need a debug swapchain / window / surface if we have the Vulkan GUI.

taichi/backends/vulkan/vulkan_api.cpp
bobcao3: Also why do we enable anisotropic filtering?

AmesingFlank: We will need geometry shaders to generate face normals when the user are not providing them. And not every user's machine will have mesh shader support (my own laptop, for example..)

bobcao3: Compute shaders are the way to go for those, or we can generate them on the CPU. To be honest, geometry shaders are garbage, never use them. In addtion lot of devices don't support geometry shaders (here is a list https://vulkan.gpuinfo.org/listdevicescoverage.php?feature=geometryShader&platform=all&option=not)

taichi/backends/vulkan/vulkan_api.cpp
```diff
@@ -400,6 +449,11 @@ void EmbeddedVulkanDevice::create_logical_device() {
...
+    device_features.wideLines = VK_TRUE;
```
bobcao3, line 455: Only enable a feature when physical device support of them are deteced.

taichi/backends/vulkan/vulkan_api.cpp
AmesingFlank: Sampler anisotropy is not necessary. And i'm fine with generating normals elsewhere.

AmesingFlank: Out of curiosity tho, why are geom shaders garbage lol

AmesingFlank:
> Can you somehow incorporate the debug swapchain into this? We don't need a debug swapchain / window / surface if we have the Vulkan GUI.

Am I correct to understand that what's needed is to just always create a swapcahin and a surface when `TI_VULKAN_DEBUG` is on?

bobcao3:
> 
> 
> > Can you somehow incorporate the debug swapchain into this? We don't need a debug swapchain / window / surface if we have the Vulkan GUI.
> 
> Am I correct to understand that what's needed is to just always create a swapcahin and a surface when `TI_VULKAN_DEBUG` is on?

that sounds good

taichi/backends/vulkan/vulkan_api.cpp
bobcao3: They don't fit in the pipeline nicely. Vertex shaders and fragment shaders are on different parallelization scheme (per vertex vs. per fragment), geometry shaders add another stage in between and therefore requires the GPU to reschedule. Also since vertex shaders may output not certain amount of primitives, it's very hard for the GPU to decide how to parallelize them efficiently. On some devices geometry shaders run fine, on some else, they run terribly. The performance is therefore not consistent and usually worse than a pipeline without it.

Fun fact BTW, mesh shaders can also be slower than normal pipelines even if they do the same thing. They allow more flexibility, but it's hard to compete with fixed function hardware.

taichi/backends/vulkan/loader.cpp
```diff
@@ -32,6 +32,16 @@ void VulkanLoader::load_device(VkDevice device) {
   volkLoadDevice(device);
 }
 
+PFN_vkVoidFunction VulkanLoader::load_function(const char *name) {
+  auto result =
+      vkGetInstanceProcAddr(VulkanLoader::instance().vulkan_instance_, name);
+  // printf("loading %s \n",name);
+  if (result == nullptr) {
+    printf("%s is nullptr\n", name);
```
k-ye, line 40: TI_WARN?

taichi/backends/vulkan/vulkan_api.h
```diff
@@ -41,6 +44,10 @@ struct VulkanQueueFamilyIndices {
   bool is_complete() const {
     return compute_family.has_value();
   }
+
+  bool is_complete_for_ui() {
```
k-ye, line 48: nit: Later we can change this to take in bitmask, e.g. `VK_QUEUE_GRAPHICS_BIT`, and check if `VulkanQueueFamilyIndices` satisfies that

k-ye (commented):
Great!

AmesingFlank:
Update: since taichi now supports [dynamic field creation](https://github.com/taichi-dev/taichi/issues/2418), I tried to replace the geometry shader with a taichi kernel that generate normals. This works and is pretty fast. So I will disable geometry shaders.

And regarding debug swap chain: I will do that in a future PR, after the GGUI related work is complete.

taichi/backends/vulkan/vulkan_api.h
AmesingFlank: Later as in after we only set `VK_QUEUE_COMPUTE_BIT` for our compute queue?  IIUC, the reason why we currently set both `VK_QUEUE_COMPUTE_BIT` and `VK_QUEUE_GRAPHICS_BIT` is that we need them for debugging? @bobcao3 

If thats the case, after GGUI, I can create a separate PR that

1. removes the debug swapchain
2. only use `VK_QUEUE_COMPUTE_BIT` for the compute queue
3. fix this nits.

k-ye: Sounds good by me :-)

taichi/backends/vulkan/vulkan_api.cpp
AmesingFlank: Cool stuff.

bobcao3:
Yes, the graphics bit is only for debugging right now. Async compute queue
would be more ideal (only if GUI is somehow made to be async). If we sync
everything then there's no benefit of using the compute queue. This could
be quite hard to design tho

k-ye (approved):
LGTM!

bobcao3 (approved):
Add a TODO on the wide line thing and have alternative ways to do that, maybe like in SVG renderers that uses a rectangle. (Lots of devices don't support wide lines)

PR #2649, opened 2021-08-09
[ci] Added changelog.py that does not depend on taichi

Leonz5288:
Related issue = #


python/changelog.py
```diff
@@ -0,0 +1,108 @@
+import argparse
+import json
+import os
+import sys
+
+from git import Repo
+
+
+def changelog(arguments: list = sys.argv[2:]):
```
k-ye, line 9: Can we re-use https://github.com/taichi-dev/taichi/blob/master/python/taichi/make_changelog.py ?

Leonz5288: done

python/changelog.py
```diff
@@ -2,6 +2,8 @@
 import json
 import os
 import sys
+sys.path.insert(0, './taichi')
```
k-ye, line 5: I still don't know why we need this instead of directly using https://github.com/taichi-dev/taichi/blob/master/python/taichi/make_changelog.py ..

Leonz5288: Because if we do `from taichi import changelog`, it would use taichi again?

k-ye: I mean the CI just uses `python python/taichi/make_changelog.py`. This should be OK?

python/taichi/main.py
```diff
@@ -224,37 +224,31 @@ def example(self, arguments: list = sys.argv[2:]):
...
         """Display changelog of current version"""
         parser = argparse.ArgumentParser(
-            prog='ti changelog', description=f"{self.changelog.__doc__}")
-        import taichi as ti
-        if ti.is_release():
-            args = parser.parse_args(arguments)
-            changelog_md = os.path.join(ti.package_root(), 'CHANGELOG.md')
-            with open(changelog_md) as f:
-                print(f.read())
+            prog='ti changelog',
```
ailzhang, line 227: Note this is likely no longer needed, if you type `ti changelog` as a user, you just want to read the changelog installed in the package. You don't really want to write a changelog as user. 
In other words we might safely move the writing part in this function to a helper script (e.g. make_changelog.py). It's even better if we just move make_changelog.py out of `python/taichi` folder to e.g. `misc` folder (so that we don't release it to users).

python/changelog.py
ailzhang: Yea it's better that we move make_changelog.py out of python folder :P It doesn't have to be released. ;)

Leonz5288: Sure, I was just preserving the saving function of this script. In this case, I'll modify the make_changelog.py to allow save.

ailzhang (commented):
Looks great! Nits:
- It's good to try calling `ti changelog` in the ci test (.github/workflow/presubmit.yaml) ;) It's currently not covered ;) 
- Also not related to your PR, it seems our `ti changelog` uses `master` as default branch, but I think it makes more sense to use the **current branch** as default value? cc: @k-ye for more context here.

ailzhang (approved):
LGTM! As discussed offline, we can merge this PR first without testing in CI (since it's not installing from a whl). @Leonz5288 will add a followup PR to move making changelog into `setup.py` so that it can be tested in CI.

PR #2655, opened 2021-08-10
[ci] Add cuda bin folder to PATH

AmesingFlank:
Add cuda bin folder to PATH so that we can compiler .cu files.


k-ye (approved):
LGTM! (We should really move these mess to docker asap..)

PR #2660, opened 2021-08-11
[vulkan] More capabilities detection and enabling

bobcao3:
Enable float 16, int 8, and a bunch of other capabilities / device extension / device property related fixes


g1n0st (approved):

AmesingFlank:
Have you tried using taichi fields with `u8` after this? Does it work?

bobcao3:
> 
> 
> Have you tried using taichi fields with `u8` after this? Does it work?

That may not. Those will require separate extensions that adds support for uint8 in storage buffers (VK_KHR_8bit_storage). I have not added that yet. This only adds support for computations using int8 or int16 or float16

AmesingFlank (approved):

k-ye (approved):
LGTM!

PR #2665, opened 2021-08-11
[ir] Improve ExternalTensorShapeAlongAxisStmt IR print result

g1n0st:
`ExternalTensorShapeAlongAxisStmt` has explicit return value but has not be shown by IR print yet, which would confuse developers. This PR fixed this.
The previous IR print result:
```
external_tensor_shape_along_axis 0, arg_id 0
```
Now:
```
<i32> $1 = external_tensor_shape_along_axis 0, arg_id 0
```


ailzhang (approved):
LGTM!

k-ye (approved):

PR #2670, opened 2021-08-13
[ci] Added taichi nightly auto release to github action

Leonz5288:
Build and release nightly per day!!!
```
PROD_PWD: ${{ secrets.PYPI_PWD_PROD }}
NIGHT_PWD: ${{ secrets.PYPI_PWD_NIGHTLY }}
```
We will need two new secrets in our github repo for these two @k-ye .

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,7 +52,8 @@ jobs:
           cd python
           git fetch origin master
           export TAICHI_CMAKE_ARGS=$CI_SETUP_CMAKE_ARGS
-          [ -z "${{ github.event.action }}" ] && python build.py build --project_name taichi_nightly
+          export COMMIT_SHA=$(git rev-parse --short "$GITHUB_SHA")
+          [ -z "${{ github.event.action }}" ] && python build.py build --project_name taichi-nightly
           [ -z "${{ github.event.action }}" ] || python build.py build
           cd ..
           NUM_WHL=`ls dist/*.whl | wc -l`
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -49,7 +51,9 @@ jobs:
           python -m pip install -r requirements_dev.txt
           cd python
           git fetch origin master
-          TAICHI_CMAKE_ARGS=$CI_SETUP_CMAKE_ARGS python build.py build
+          export TAICHI_CMAKE_ARGS=$CI_SETUP_CMAKE_ARGS
+          [ -z "${{ github.event.action }}" ] && python build.py build --project_name taichi_nightly
```
ailzhang, line 55: Nit: could you use if-else here instead of &&/||? IIUC here you just want to make sure there's only one wheel built, but it's less straightforward writing this way :P

.github/workflows/release.yml
```diff
@@ -66,10 +70,14 @@ jobs:
       - name: Upload PyPI
         env:
           # https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow
-          PYPI_PWD: ${{ secrets.PYPI_PWD }}
+          PROD_PWD: ${{ secrets.PYPI_PWD_PROD }}
+          NIGHT_PWD: ${{ secrets.PYPI_PWD_NIGHTLY }}
         run: |
           cd python
-          python build.py upload --skip_build
+          [ -z "${{ github.event.action }}" ] && export PYPI_PWD="$NIGHT_PWD"
```
ailzhang, line 77: Same as above.

ailzhang (commented):
Overall looks good! Only had minor inline suggestions to make it easier to understand. ;) 
Also double check: did you try testing it on test_actions?

.github/workflows/release.yml
Leonz5288: I tested using this syntax. And I am not sure about using if-else syntax. I mean this syntax looks a lot more simple and tidy? I can add a few comments to make this clearer?

ailzhang: haha sure, adding a comment sounds good to me.

ailzhang (approved):

k-ye:
> We will need two new secrets in our github repo for these two @k-ye .
Done!

PR #2671, opened 2021-08-13
[vulkan] Querying features like a mad man

bobcao3:
- Added the very first initial piece of Device API and moving capability queries to that. (So now the codegen is decoupled from the runtime)
- Query & enable features properly (previously we don't query the feature support within an extension, e.g. a device may support f16 but not i8)


bobcao3:
Build is finally passing!

AmesingFlank (approved):
LGTM! and can confirm that this fixed the crash on my win machine.

PR #2673, opened 2021-08-13
[Opt] [ir] Optimize offload

squarefk:
Related issue = #


taichi/transforms/offload.cpp
```diff
@@ -617,7 +538,6 @@ class FixCrossOffloadReferences : public BasicStmtVisitor {
         stmt_to_offloaded[copy.get()] = stmt_to_offloaded[stmt];
         stmt->set_operand(index, copy.get());
         stmt->insert_before_me(std::move(copy));
-        return true;
       }
     }
     if (op->is<GlobalPtrStmt>()) {
```
k-ye, line 543: return?

k-ye (approved):
Great! We likely need to split `visit_operand()` into smaller functions

taichi/transforms/offload.cpp
```diff
@@ -626,40 +546,32 @@ class FixCrossOffloadReferences : public BasicStmtVisitor {
...
-      stmt->insert_before_me(std::move(copy_stmt));
-      stmt->insert_before_me(std::move(global_temporary_stmt));
-      stmt->insert_before_me(std::move(ptr_offset_stmt));
-      return true;
+    if (local_to_global_offset.find(op) == local_to_global_offset.end()) {
+      // For cases like ConstStmt
+      auto copy = op->clone();
+      stmt_to_offloaded[copy.get()] = stmt_to_offloaded[stmt];
+      stmt->set_operand(index, copy.get());
+      stmt->insert_before_me(std::move(copy));
```
xumingkuan, line 556: If this part can handle ConstStmt correctly, we should remove the code fragment `if (config.advanced_optimization) {  if (op->is<ConstStmt>()) ...` to avoid inserting the statement twice.
Also, it's better to give an example other than ConstStmt, otherwise `if (op->is<ConstStmt>())` is more intuitive to me.

k-ye: We had a discussion on this and feel like this could be a better/simpler approach: Regardless of `advanced_optimization`, we should *not* put Const into GlobalTmps. Instead, if OffloadedTask 2 references a Const defined in OffloadedTask 1, we can just clone that Const.

xumingkuan: Yeah but when `advanced_optimization=True`, the current code clone the const twice?

squarefk: https://github.com/taichi-dev/taichi/pull/2673/commits/87be94aee909d74b44319cbe1df7096e8286b05c

taichi/transforms/offload.cpp
```diff
@@ -441,166 +457,71 @@ class FixCrossOffloadReferences : public BasicStmtVisitor {
     local_to_global_vector_type[stmt] = ret_type;
     auto ptr = replacement.push_back<GlobalTemporaryStmt>(
         local_to_global_offset[stmt], ret_type);
+    stmt_to_offloaded[ptr] = stmt_to_offloaded[stmt];
```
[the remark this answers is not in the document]
squarefk: https://github.com/taichi-dev/taichi/pull/2673/commits/9f9214ca1fdd51c782574ca4dbf5dc4ab19895e8

taichi/transforms/offload.cpp
```diff
@@ -441,166 +457,71 @@ class FixCrossOffloadReferences : public BasicStmtVisitor {
...
-    auto offset = stmt->offset;
-
-    if (stmt_to_offloaded[offset] != stmt_to_offloaded[stmt]) {
+  // Replace local LD/ST with global LD/ST
+  void visit(LocalLoadStmt *stmt) override {
+    generic_visit(stmt);
+    TI_ASSERT(stmt->width() == 1)
+    auto ptr = stmt->src[0].var;
+    auto top_level_ptr = ptr;
+    while (top_level_ptr->is<PtrOffsetStmt>())
```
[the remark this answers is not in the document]
squarefk: https://github.com/taichi-dev/taichi/pull/2673/commits/9707bb04abc3c1c5a43f71a4608dff4daf824360

PR #2674, opened 2021-08-13
[Test] Unify tests decorators with @ti.test()

squarefk:
Some options need backend to support specific extension, like:
- `default_fp=ti.f64` and `ti.extension.data64`
- `dynamic_index=True` and `ti.extension.dynamic_index`
Currently, the former one is implemented outside @ti.test() and the latter one is implemented inside @ti.test(). We may want to figure out a more elegant way to deal with such cases.

We don't want to sneakily change the user's init configurations. How should we respond with dynamic_index=True, when the target backend doesn't support dynamic_index? 

Clean up tests without no ti.init().

Figure out how to print to the terminal while running pytest.


tests/python/grad_test.py
```diff
@@ -3,6 +3,7 @@
 
 
 def grad_test(tifunc, npfunc=None):
```
strongoier, line 5: I believe this test will not be executed and should be removed. The same function already exists as a helper function in `test_ad_basics.py`.

taichi/transforms/offload.cpp
```diff
@@ -14,6 +14,28 @@ TLANG_NAMESPACE_BEGIN
 namespace irpass {
 namespace {
 
+class SquashPtrOffset : public IRVisitor {
```
strongoier, line 17: Changes to this file should not appear in this PR.

strongoier (approved):
LGTM! Thanks!

PR #2676, opened 2021-08-13
[Lang] Add a Ndarray class to serve as an alternative to dense scalar fields

strongoier:
Related issue = #

Usage:
```
a = ti.ndarray(ti.i32, shape=(n, m)) // by default torch tensor is used
// arbitrary initialization using numpy arrays and torch tensors is also supported
b = ti.Ndarray(np.empty(shape=(n, m), dtype=np.int32))
c = ti.Ndarray(torch.empty((n, m), dtype=torch.int32))
```

When passing it to Taichi kernels, you need to annotate it as `ti.ext_arr()` for now.


k-ye (approved):
LGTM!

PR #2678, opened 2021-08-13
[Refactor] Re-enable gdb attach on crash.

ailzhang:
* __->__ #2678


ailzhang:
Thought that I'd use this PR to test ghstack again to make sure we cannot accidentally push to master :P sorry for the noise, the real PR is at https://github.com/taichi-dev/taichi/pull/2681

PR #2686, opened 2021-08-15
[Refactor] Allow taichi_cpp_tests run in release mode as well.

ailzhang:
taichi_cpp_tests used to depend on `TAICHI_REPO_DIR` and thus only runs
in dev mode. This PR decouples them, instead we explicitly ask users to
pass in `TI_LIB_DIR` if they want to run the cpp tests.

Note this PR will unblock `TAICHI_REPO_DIR` and is_release check removal in the codebase.

Also instead of a crash report, we'll properly throw an error
message to remind setting `TI_LIB_DIR` when running cpp tests.

```
[E 08/14/21 14:34:16.990 2077742] [lang_util.cpp:runtime_lib_dir@25]
If you are running the taichi_cpp_tests please set TI_LIB_DIR to
$TAICHI_INSTALL_DIR/lib where TAICHI_INSTALL_DIR can be retrieved from
taichi.__path__ in python
```

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -99,9 +99,9 @@ jobs:
           python examples/algorithm/laplace.py
           ti diagnose
           ti changelog
-          # TODO: make cpp tests work in both release & dev mode.
-          # Currently cpp tests only works in dev mode since it depends on the path set there.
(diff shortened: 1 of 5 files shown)


k-ye (approved):
LGTM!

PR #2687, opened 2021-08-15
[Refactor] Re-enable gdb attach on crash.

ailzhang:
update: thought that I'd merge this first and look into ghstack permission issue later. sorry for the noise. 


k-ye (approved):
LGTM!

PR #2688, opened 2021-08-15
[CUDA] Save an extra host to device copy if arg_buffer is already on device.

ailzhang:
This PR is pretty self contained and came from a discussion with @k-ye. 

I added a new test in test_torch_io.py since we only test cpu arch there, so I want to make sure cuda arch is tested. 

Note to self: I had indeterministic crash when I first tested this PR and crash also happened on master branch. After some time I was no longer able to reproduce the crash so I guess it was due to my GPU in a weird state but not related to this PR. 

Leaving a note here just in case if happens again in the future.

### DIFF
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -46,36 +46,46 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
 
     return [offloaded_local, cuda_module,
             kernel = this->kernel](Context &context) {
-      // copy data to GRAM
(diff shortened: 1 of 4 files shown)


strongoier:
The original tests are meant to test all archs including cuda. There's some issue with the `torch_test` decorator and I fixed it in #2676. So I think if you pull master you'll get cuda tested.

ailzhang:
@strongoier Nice I'll remove the extra test then. No the crash is not related to OOM, it happened at some LLVM code IIRC.

k-ye (approved):
LGTM, thanks!

PR #2689, opened 2021-08-15
[Refactor] Make is_release always True and delete runtime dep on TAICHI_REPO_DIR

ailzhang:
This PR removes runtime dependency on `TAICHI_REPO_DIR` in the codebase. 

Known followups:
- There's a use of `TAICHI_REPO_DIR` in `python/ti.cpp` which is used by
  windows. I'll skip it for now.
- Need to cleanup `TAICHI_REPO_DIR` in docs and examples.

### DIFF
--- a/python/taichi/core/util.py
+++ b/python/taichi/core/util.py
@@ -76,7 +76,7 @@ def package_root():
 
 
 def is_release():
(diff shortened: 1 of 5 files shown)


k-ye (approved):
Thanks!

PR #2691, opened 2021-08-15
[Refactor] Remove is_release() in the codebase.

ailzhang:
This PR removes all the logic depending on `is_release` in the codebase. (It already returned True before this PR so it's safe to do so).


k-ye (approved):
LGTM, thanks for the cleanup!

PR #2692, opened 2021-08-15
[vulkan] [test] Enable Vulkan backend on OS X

g1n0st:
Related issue = #2607

In my OS X, the Vulkan backend has better performance than the Metal backend (+5fps on average in mpm128 example). We already have a feature querier on the Vulkan backend to handle different capabilities (e.g., float_atomic_add). Based on these reasons, this PR enables the Vulkan backend on OS X. It also makes the Vulkan backend passed all desired unit tests on OS X.


k-ye (approved):
Thanks!

PR #2695, opened 2021-08-16
[Vulkan] Move vulkan to device API

bobcao3:
1. Memory allocation has been moved to Device API
2. Vulkan codegen don't have dependency on Vulkan or its runtime
3. `runtime.cpp` is mostly free of Vulkan things. A few SPIR-V and kernel generation stuff are still operating on Vulkan / SPIRV-tools
4. Descriptor layout & allocation management through the resource binder API

These are things that I have not done yet:
1. Change `vulkan_api.cpp` to purely provide the necessary Vulkan constructs to initialize a VulkanDevice. (So that it is clear how a 3rd party may want to integrate the runtime)
2. Entirely isolate `runtime.cpp` from any Vulkan stuff, making it multi-platform. (e.g. if we want, we can use the same code to run DX backend or OpenGL backend as long as there's full Device API support)

At that point the runtime.cpp can be moved out and be a unified "Logical" addressing model runtime. Other APIs that has "Physical" addressing model can use another set of different runtime & codegen.

Some designs of the Device API are yet to be desirable, such as:
1. How would we support opaque types like images or acceleration structures?
2. For graphics use (using Device API as a RHI), how would we create pipelines? (GLSL/HLSL and cross-compiling? SPIR-V cross compiling?)

Finally, another goal of supporting graphics on the Device API is that GGUI and other renderer we will build can reuse these code and enabling multi-platform without additional code.


taichi/backends/vulkan/vulkan_device.cpp
```diff
@@ -0,0 +1,602 @@
...
+
+  if (free_cmdbuffers_.size()) {
+    buffer = free_cmdbuffers_.back();
+    free_cmdbuffers_.pop_back();
+  } else {
+    // FIXME: Proper multi queue support, instead of defaulting to compute
+    // command pool
+    VkCommandBufferAllocateInfo alloc_info{};
+    alloc_info.sType = VK_STRUCTURE_TYPE_COMMAND_BUFFER_ALLOCATE_INFO;
+    alloc_info.commandPool = compute_cmd_pool();
```
AmesingFlank, line 437: Can we add an argument to `new_command_list` (e.g. a `struct CommandListParams`) so that we can optionally use the graphics cmd pool?

taichi/backends/vulkan/vulkan_device.h
```diff
@@ -0,0 +1,221 @@
...
+  VkPipeline pipeline_{VK_NULL_HANDLE};
+  VkDescriptorPool descriptor_pool_{VK_NULL_HANDLE};
+  VkDescriptorSet descriptor_set_{VK_NULL_HANDLE};
+};
+
+class VulkanCommandList : public CommandList {
+ public:
+  VulkanCommandList(VulkanDevice *ti_device,
+                    VkDevice device,
+                    VkCommandBuffer buffer);
```
AmesingFlank, line 96: nit: do we still need the `VkDevice` here since we have the `VulkanDevice`?

taichi/backends/device.cpp
```diff
@@ -0,0 +1,25 @@
+#include <taichi/backends/device.h>
```
g1n0st, line 1: This PR influences out of the Vulkan scope, maybe @k-ye should check it?

taichi/backends/vulkan/vulkan_device.cpp
bobcao3: Sounds about right. Should we also assume all graphics queue will carry compute capability? Benefit of that being not bothering with cross-queue synchronization.

AmesingFlank: Hmm. I would slightly prefer not making that assumption. At some point we will need to properly handle graphics-compute synchronization anyways, since the compute backend isn't always Vulkan. And at that point it might be easier to use two queues even though both backends are vulkan.

bobcao3: If the compute backend isn't Vulkan the compute should be on another different Device, maybe a CudaDevice class. Data sharing through those should somehow be achieved with the `share_to()` call. I think for all the APIs I can think of, unless it's old OpenGL, a graphics queue always contain compute capabilities, but a lot of compute queues are compute only

AmesingFlank: If the compute backend and graphics backend are both Vulkan, will we be using the same `Device` for both? It might be the case that we will have cleaner code if we always created two devices, one for compute and one for graphics. This way,  we can re-use some of the logic for compute-graphics data sharing for different compute backends.

taichi/backends/device.h
```diff
@@ -31,6 +31,103 @@ enum class DeviceCapability : uint32_t {
   vk_has_spv_variable_ptr,
 };
 
+class Device;
+struct DeviceAllocation;
+struct DevicePtr;
+
+// TODO: Figure out how to support images. Temporary solutions is to have all
+// opque types such as images work as an allocation
+struct DeviceAllocation {
```
k-ye, line 40: I'm a bit worried about this design: In AOT mode, users likely have a set of buffers. This would prevent them from using their own.

taichi/backends/vulkan/vulkan_device.cpp
bobcao3: Using two would be creating an assumption that the memory between them are not unified, which could be a potential performance hurdle.

bobcao3: Ideally in the future taichi will create a device per detected physical device. So a user with a Nvidia card will boot up taichi with both a VulkanDevice and a CudaDevice. If the user has another integrated intel, we might also have another VulkanDevice maps to that intel GPU

taichi/backends/device.h
```diff
@@ -31,6 +31,103 @@ enum class DeviceCapability : uint32_t {
...
+  virtual void bind_resources(ResourceBinder &binder) = 0;
+  virtual void buffer_barrier(DevicePtr ptr, size_t size) = 0;
+  virtual void buffer_barrier(DeviceAllocation alloc) = 0;
+  virtual void memory_barrier() = 0;
+  virtual void buffer_copy(DevicePtr dst, DevicePtr src, size_t size) = 0;
+  virtual void buffer_fill(DevicePtr ptr, size_t size, uint32_t data) = 0;
+  virtual void dispatch(uint32_t x, uint32_t y = 1, uint32_t z = 1) = 0;
+
+  // These are not implemented in compute only device
+  virtual void draw(uint32_t num_verticies, uint32_t start_vertex = 0) {
```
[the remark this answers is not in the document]
bobcao3: Maybe the other way around? Or something to represent a combined list

taichi/backends/device.h
bobcao3: We can have the device API support importing external buffers. Assuming in AOT mode the Vulkan runtime is still there

k-ye: Sounds good thx!

taichi/backends/device.h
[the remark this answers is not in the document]
k-ye: I was originally wondering if other backends could follow this pattern. E.g. in Metal there are different kinds of encoders. IIUC, this interface means that the Metal impl would implicitly begin and end encoders (e.g. from a `dispatch()` call to a `draw()` call)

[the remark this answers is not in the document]
bobcao3: We can certainly expose a `finalize()` or `end()` call. As for Metal, I'm not sure how separated are compute / graphics encoder so I don't really have an idea. On DX and Vulkan they are pretty much merged i think.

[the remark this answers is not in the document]
k-ye: Sounds good. FYI Metal's blit/compute/rendering are completely separated: https://developer.apple.com/documentation/metal/mtlcommandbuffer.

taichi/backends/vulkan/runtime.cpp
```diff
@@ -61,16 +53,16 @@ class StopWatch {
 };
 
 using BufferEnum = TaskAttributes::Buffers;
-using InputBuffersMap = std::unordered_map<BufferEnum, VkBufferWithMemory *>;
+using InputBuffersMap = std::unordered_map<BufferEnum, DeviceAllocation *>;
```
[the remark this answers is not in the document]
bobcao3: Left as a TODO since it's causing quite some trouble rn as the context_buffers are in unique ptrs which can be null if the kernel doesn't use args, and the code gets a bit messy since de-refercing a null is definitely not good but otherwise there need to be if statements to pass in a null allocation depending on the unique pointer's "null-ness"

taichi/backends/vulkan/vulkan_device.cpp
bobcao3: Left as a TODO

AmesingFlank (approved):

g1n0st (approved):

k-ye (approved):
Thanks! Looking forward to this new API!

PR #2703, opened 2021-08-16
[Lang] Redesign Ndarray class and add ti.any_arr() annotation

strongoier:
Related issue = #

Details:
- Like `Field`, there should also be `Ndarray`/`ScalarNdarray`/`MatrixNdarray` classes (`MatrixNdarray` not implemented yet) instead of a standalone `Ndarray` class.
- There's no need to let users change underlying implementation of `Ndarray`, so other ways of constructing `Ndarray` are removed.
- `Ndarray` is not an external array so we shouldn't use `ti.ext_arr()` to pass it into Taichi kernels. `ti.any_arr()` annotation is added to accept both `Ndarray` and external arrays.


python/taichi/lang/kernel_arguments.py
```diff
@@ -36,6 +36,29 @@ def extract(self, x):
...
+"""Alias for :class:`~taichi.lang.kernel_arguments.ArgAnyArray`.
+
+Example::
+
+    >>> @ti.kernel
+    >>> def to_numpy(x: ti.any_arr(), y: ti.any_arr()):
+    >>>     for i in range(n):
+    >>>         x[i] = y[i]
+    >>>
+    >>> y = ti.ndarray(ti.f64, shape=n)
```
ailzhang, line 55: Nit: if `ti.ndarray` will be a first class citizen in the future, shall we also name the annotation accordingly? 

I was thinking about `ti.annotation.ndarray()` as an example but that might require a bit refactor. 

`ti.ext_arr()` doesn't look like a type annotation from its name :P

strongoier: Nice catch - I agree that the current annotations (`ti.template()` for fields, `ti.ext_arr()` for external arrays) are not straightforward to users. Instead, they are more like directives for compilers regarding which information is available at compile time/runtime. As for `ti.any_arr()` here, the final name is still under discussion but we are aiming for a general annotation which accepts external arrays, Taichi ndarrays and dense Taichi fields. Maybe we need a better naming system in the future.

ailzhang: Yup agreed. I feel like this doesn't have to be resolved in this PR, but we probably need to keep this in mind to make it more straightforward to users before we officially release ndarray. :D

strongoier: Sure :-)

ailzhang (approved):
LGTM! Will leave the final review for @k-ye since I wasn't in the ndarray API design :D

python/taichi/lang/kernel_arguments.py
```diff
@@ -36,6 +36,29 @@ def extract(self, x):
 """
 
 
+class ArgAnyArray:
+    """Type annotation for arbitrary arrays, including external arrays and Taichi ndarrays."""
+    def extract(self, x):
+        return to_taichi_type(x.dtype), len(x.shape)
```
k-ye, line 42: TODO: We should include information like AOS/SOA, matrix shape

k-ye (approved):
LGTM!

python/taichi/lang/kernel_arguments.py
strongoier: BTW should I support these in `ext_arr()` first or just implement them in `any_arr()` and deprecate `ext_arr()` in the future?

k-ye: >  just implement them in any_arr()

This sounds good

Issue #2690, opened 2021-08-15
Test `test_listgen_opt_with_offsets` might be flaky


ailzhang:
I've seen twice on two different PRs that `test_listgen_opt_with_offsets` failed in CI but rerunning passed. 
It looks like a flaky test to me. Some logs just in case it's useful:
```
[gw0] PASSED ../../../hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/taichi/tests/test_custom_float.py::test_cache_read_only 

=================================== FAILURES ===================================
_ hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/taichi/tests/test_async.py _
[gw1] darwin -- Python 3.7.11 /Users/runner/hostedtoolcache/Python/3.7.11/x64/bin/python
worker 'gw1' crashed while running 'hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/taichi/tests/test_async.py::test_listgen_opt_with_offsets'
=============================== warnings summary ===============================
../../../hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/autograd/numpy/numpy_wrapper.py:20: 12 warnings
  /Users/runner/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/autograd/numpy/numpy_wrapper.py:20: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    int_types = {_np.int, _np.int8, _np.int16, _np.int32, _np.int64, _np.integer}

-- Docs: https://docs.pytest.org/en/stable/warnings.html
=========================== short test summary info ============================
FAILED ../../../hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/taichi/tests/test_async.py::test_listgen_opt_with_offsets
====== 1 failed, 863 passed, 1 skipped, 12 warnings in 587.61s (0:09:47) =======
```

We should keep an eye on this and maybe disable it first if it also happens to other people.

ailzhang:
Closing as it's no longer reporducible

PR #2708, opened 2021-08-17
[ci] Fixed pypi version

Leonz5288:
The first nightly release has failed because the version does not meet the requirement of PyPi.
```
<title>400 '0.7.29.2021-08-17.899707f' is an invalid value for Version. Error: Start and end with a letter or numeral containing only ASCII numeric and '.', '_' and '-'. See https://packaging.python.org/specifications/core-metadata for more information.</title>
```
As PyPi requires versioning of projects meets the PEP440 standards, which does not allow commit hash in project version, I discard adding commit hash in nightly versions and made version meet PEP440 standards.
```
DVCS based version labels

Many build tools integrate with distributed version control systems like Git and Mercurial in order to add an identifying hash to the version identifier. As hashes cannot be ordered reliably such versions are not permitted in the public version field.

As with semantic versioning, the public .devN suffix may be used to uniquely identify such releases for publication, while the original DVCS based label can be stored in the project metadata.

Identifying hash information may also be included in local version labels.
```
For hash information of each nightly release, find it in github action workflows pages.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,9 +39,9 @@ TI_VERSION_PATCH = 29
 version = f'{TI_VERSION_MAJOR}.{TI_VERSION_MINOR}.{TI_VERSION_PATCH}'
 if project_name == 'taichi-nightly':
     from datetime import date
-    today = date.today().strftime('%Y_%m_%d')
+    today = date.today().strftime('%Y%m%d')
     commit_sha = os.getenv('COMMIT_SHA')
-    version += f'.{today}.{commit_sha}'
+    version += f'.dev{today}'
 
 data_files = glob.glob('python/lib/*')
 print(data_files)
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -39,9 +39,9 @@
 version = f'{TI_VERSION_MAJOR}.{TI_VERSION_MINOR}.{TI_VERSION_PATCH}'
 if project_name == 'taichi-nightly':
     from datetime import date
-    today = date.today().strftime('%Y_%m_%d')
+    today = date.today().strftime('%Y%m%d')
     commit_sha = os.getenv('COMMIT_SHA')
-    version += f'.{today}.{commit_sha}'
+    version += f'.dev{today}'
```
ailzhang, line 44: Do we still want `commit_sha`? ;)

ailzhang: sorry just saw the PR summary, does it work if we keep the first 6 digits of commit sha?

Leonz5288: I believe not, because as PEP440 mentions, the version must be in a meaningful order. Apparently, commit hash does not guarantee this property...

Leonz5288: And I am already only keeping the first 6 digits XD.

ailzhang:
Btw the CI failure is not related to this PR. I've reported a similar issue in #2690 . We probably need to disable this flaky test.

ailzhang (approved):
LGTM!

PR #2710, opened 2021-08-17
[ci] Pin yapf version to 0.31.0.

ailzhang:
Recently noticed that `yapf` in the code format server no longer pins to 0.29. Changing yapf version is pretty annoying to dev workflow so let's pin it to the current version in the code format server (which is 0.31.0). 
Nit: I'm not sure if yapf should be a dependency for taichi user, maybe we can remove it from requirements.txt.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -38,7 +38,7 @@ jobs:
           git remote add upstream https://github.com/taichi-dev/taichi.git
           git fetch upstream master
           sudo apt-get install clang-format
-          python3 -m pip install --user yapf gitpython colorama isort
+          python3 -m pip install --user yapf==0.31.0 gitpython colorama isort
           python3 python/taichi/code_format.py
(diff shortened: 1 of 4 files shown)


.github/workflows/presubmit.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           git remote add upstream https://github.com/taichi-dev/taichi.git
           git fetch upstream master
           sudo apt-get install clang-format
-          python3 -m pip install --user yapf gitpython colorama isort
+          python3 -m pip install --user -r requirements_dev.txt
```
strongoier, line 41: Will this make this step too slow? IMO checking code format needs to be as fast as possible.

ailzhang: Hmmm good point, IIRC this isn't too slow but yea, let's just explicitly install it then. Just another call site to change manually :P

Leonz5288 (approved):
LGTM!

PR #2721, opened 2021-08-17
[vulkan] Fix dumb memory error

bobcao3:
std::vector will resize when elements are pushed into it, and in this process the pointers to the elements will change. This fixes crashing on Linux (and probably windows as well)


taichi/backends/vulkan/vulkan_device.cpp
```diff
@@ -190,22 +190,37 @@ void VulkanResourceBinder::framebuffer_depth_stencil(DeviceAllocation image) {
 void VulkanResourceBinder::write_to_set(uint32_t index,
                                         VulkanDevice &device,
                                         VkDescriptorSet set) {
-  std::vector<VkDescriptorBufferInfo> buffer_info;
+  std::vector<VkDescriptorBufferInfo> buffer_infos;
   std::vector<VkWriteDescriptorSet> desc_writes;
```
k-ye, line 194: Just to be safe, we can `reserve()` in the beginning.

k-ye (approved):
LGTM! Thanks for the fix

taichi/backends/vulkan/vulkan_device.cpp
bobcao3: Since I won't know the size until I went through all the descriptors, I don't think we can use reserve here? Using a second pass should work.

PR #2726, opened 2021-08-18
[ci] Reduce the number of python wheels built nightly

Leonz5288:
Related issue = $


ailzhang (approved):
Sounds good! nit: it's okay to remove these lines instead of commenting them out ;) mainly because they're pretty easy to add back.

Leonz5288:
> Sounds good! nit: it's okay to remove these lines instead of commenting them out ;) mainly because they're pretty easy to add back.

True!

PR #2727, opened 2021-08-18
[Lang] Enable treating external arrays as Taichi vector/matrix fields

strongoier:
Related issue = #

Previously, external arrays can only be treated as Taichi scalar fields. Therefore, people have to copy them to Taichi vector/matrix fields first in order to treat them as Taichi vector/matrix fields and operate them with another ti.Vector/ti.Matrix. After this PR, people can directly pass external arrays into Taichi kernels without any data copy and let them act as if they are Taichi vector/matrix fields:
```python
@ti.kernel
def func1(a: ti.any_arr(element_shape=(2, 2))):
    for i in range(5):
        for j, k in ti.ndrange(2, 2):
            a[i][j, k] = j * j + k * k

m = np.empty((5, 2, 2), dtype=np.int32)
func1(m)

@ti.kernel
def func2(b: ti.any_arr(element_shape=(10,), is_soa=True)):
    for i in range(5):
        for j in range(4):
            b[i][j * j] = j * j

v = np.empty((10, 5), dtype=np.int32)
func2(v)
```


python/taichi/lang/ext_array.py
```diff
@@ -38,3 +38,59 @@ def loop_range(self):
...
+
+class AnyArray:
+    """Class for arbitrary arrays in Python AST.
+
+    Args:
+        ptr (taichi_core.Expr): A taichi_core.Expr wrapping a taichi_core.ExternalTensorExpression.
+        element_shape (Tuple[Int]): () if scalar elements (default), (n) if vector elements, and (n, m) if matrix elements.
+        is_soa (bool): Whether to change the layout from AOS (default) to SOA.
+    """
+    def __init__(self, ptr, element_shape, is_soa):
```
k-ye, line 51: I'd suggest to use `layout`, so that we can have  a wider range of options (e.g. AOS, SOA, AoSoA)

python/taichi/lang/impl.py
```diff
@@ -158,6 +158,25 @@ def subscript(value, *indices):
             ])
         else:
             return Expr(_ti_core.subscript(var, indices_expr_group))
+    elif isinstance(value, AnyArray):
+        field_dim = int(value.ptr.get_attribute("dim"))
```
k-ye, line 162: nit: Leave a TODO so that we don't do this..?

python/taichi/lang/impl.py
```diff
@@ -158,6 +158,25 @@ def subscript(value, *indices):
...
+        element_dim = len(value.element_shape)
+        if field_dim != index_dim + element_dim:
+            raise IndexError(
+                f'Field with dim {field_dim - element_dim} accessed with indices of dim {index_dim}'
+            )
+        if element_dim == 0:
+            return Expr(_ti_core.subscript(value.ptr, indices_expr_group))
+        else:
```
k-ye, line 170: nit: no need to `else` the rest of the code

python/taichi/lang/kernel_arguments.py
```diff
@@ -37,9 +37,35 @@ def extract(self, x):
...
+    """Type annotation for arbitrary arrays, including external arrays and Taichi ndarrays.
+
+    For external arrays, we can treat it as a Taichi field with Vector or Matrix elements by specifying element shape and layout.
+
+    Args:
+        element_shape (Tuple[Int], optional): () if scalar elements (default), (n) if vector elements, and (n, m) if matrix elements.
+        is_soa (bool, optional): Whether to change the layout from AOS (default) to SOA.
+    """
+    def __init__(self, element_shape=(), is_soa=False):
+        assert len(
```
k-ye, line 49: Raise errors instead of using assert. The idea is that the former handles user errors, while the later catches Taichi implementation bugs

k-ye (commented):
A few nits, great work!

python/taichi/lang/ext_array.py
strongoier: Well the reason why I use bool instead of ti.AOS/SOA is circular import...

k-ye: Can we move around `ti.AOS/SOA` to break the cycle?

strongoier: I cannot use `ti.AOS/SOA` as default values of method parameters. It seems that the Matrix class hacks it using `None` as default value and changes it into `ti.AOS` in the method...

strongoier: If we want use `ti.xxx` then no. If we want to move it to another file and directly use `taichi.lang.xxx.xx` then possible.

python/taichi/lang/impl.py
strongoier: Do you mean that we are going to put `subscript()` in their individual classes?

python/taichi/lang/ext_array.py
k-ye: > directly use taichi.lang.xxx.xx then possible.

SG!

python/taichi/lang/impl.py
k-ye: sorry, i mean the "dim" part

python/taichi/lang/ext_array.py
strongoier: Shall we merge this first and add a separate PR to do this? This is a slight refactoring which will affect a couple of places. I plan to make an Enum for Layout instead of a class with a true/false var inside (current implementation). This will allow us to support more potential layouts in the future. Furthermore, should we expose this to users as `ti.Layout.AOS` or keep the current `ti.AOS`?

python/taichi/lang/kernel_arguments.py
strongoier: done

python/taichi/lang/impl.py
strongoier: done

python/taichi/lang/impl.py
strongoier: done

k-ye (approved):
LGTM!

PR #2735, opened 2021-08-20
[test] Fix potential memory error when DecoratorRecorder hasn't been reset correctly

squarefk:


strongoier (approved):
Awesome!!! Thanks!

PR #2739, opened 2021-08-20
[ci] Fix clang-format version to 10

ailzhang:
We were using system default which is clang-format-10 before. Let's still fix it to 6.0.


Leonz5288 (approved):
LGTM

Issue #2760, opened 2021-08-20
Provide dev installation guide with conda


k-ye:
**Concisely describe the proposed feature**

I can share some tips on this topic, so that devs do not have to pollute their default python environments.

PR #2764, opened 2021-08-21
[Doc] Add instructions for how to use conda

k-ye:
Related issue = #2760


conda/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+We need an equivalent bat script for Windows.
+:::
+
+Download and install the `conda` matching your OS: https://docs.conda.io/en/latest/miniconda.html.
+
+Run the following commands:
+
+```bash
+$ conda env create -f conda_env.yaml
```
ailzhang, line 14: nit: would be nice to add a `cd conda` before this ;)

conda/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+$ conda activate taichi-dev
+
+# This will only install the dependencies to the 'taichi-dev' environment.
+(taichi-dev) $ python3 -m pip install --user -r ../requirements_dev.txt
```
ailzhang, line 19: In this case I think it's safe and cleaner to do `pip install -r ../requirements_dev.txt`?

conda/scripts/activate_env_vars.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -x
+
+### Path to clang++ ###
```
ailzhang, line 5: IIUC this file will be run when we conda activate taichi-dev environment? (great idea!)
Nit: maybe it's good to do some additional checks here? For example we know on linux if CXX is not set taichi will likely pick up g++ instead of clang++, which is not what we want? 
Similar for llvm, IIUC we either need llvm to be in PATH or set LLVM_DIR? we can probably check those here and give nicer errors than at cmake :D 
Same for LLVM_DIR, maybe we can check `llvm-config --version` here to make sure it's in path?

docs/lang/articles/contribution/dev_install.md
```diff
@@ -126,6 +125,12 @@ installer.
 3. Execute `python3 -m taichi test` to run all the tests. It may take
   up to 5 minutes to run all tests.
 
+## Conda
+
+To avoid directly installing Taichi's dependencies into your existing
+Python environment, we have provided a pre-defined `conda` environment.
+You can find the instructions [here](https://github.com/taichi-dev/taichi/conda/README.md).
+
```
ailzhang, line 133: It also worth mentioning that once conda is set up following the README, they still need to run `python setup.py develop`?

ailzhang (approved):
Looks great! Only a few minor nits.

conda/README.md
k-ye: right :-)

docs/lang/articles/contribution/dev_install.md
k-ye: Done

conda/scripts/activate_env_vars.sh
k-ye: Good point... Not exactly sure what additional checks we need here. I feel like CMake is still better at catching such errors, given that it handles these env settings in a more structured way. (E.g. we don't have to write some fancy script just to figure out where LLVM headers/libs are) WDYT?

From my point of view, the bigger issue is that users who change this file would need to remember to reset it, or their working tree becomes dirty...

ailzhang: Yea originally I was thinking about some simple checks like "check CXX must be set on linux/llvm-config must be runnable or LLVM_DIR is set".(cmake errors are clear enough but it doesn't remind setting CXX/LLVM_DIR as workarounds :P )  But yea I agree it's kinda hard to remember updating this file compared to normal bashrc/zshrc I guess. I'm totally fine that we start with this version first :D

PR #2769, opened 2021-08-23
[Refactor] Move cc backend to use runtime_lib_dir() and remove runtime_tmp_dir.

ailzhang:
* #2770
* __->__ #2769


ailzhang:
note to self: we do need to separate the generated .c/.o files in a multi threaded environment. Will come back to this later.

PR #2774, opened 2021-08-23
[Refactor] Make Layout an Enum class and move it away from impl.py

strongoier:
Related issue = #2223
Discussed in #2727 Review

Reasons:
- Make `Layout` an `Enum` class: Using `Enum` is more clear and extensible than using a class with a bool inside.
- Move `Layout` away from `impl.py`: Avoid circular import when we want to set default parameter values to some layout (previously it was hacked with a `None` value); Avoid exposing everything directly under `ti`.

Effects:
- Now users need to use `ti.Layout.AOS/SOA`. `ti.AOS/SOA` is still available but marked as deprecated.

Side Effects:
- A new way to handle deprecated names is implemented. We only need to directly add the mapping to `deprecated_names` in `taichi/__init__.py`.

### DIFF
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -4,7 +4,7 @@ from copy import deepcopy as _deepcopy
 
 from taichi.core.util import ti_core as _ti_core
 from taichi.lang import impl
-from taichi.lang.enums import *
+from taichi.lang.enums import Layout
 from taichi.lang.exception import InvalidOperationError
 from taichi.lang.impl import *
 from taichi.lang.kernel_arguments import any_arr, ext_arr, template
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -3,6 +3,7 @@ import os
 from copy import deepcopy as _deepcopy
 
 from taichi.core.util import ti_core as _ti_core
+from taichi.lang.enums import *
 from taichi.lang import impl
 from taichi.lang.exception import InvalidOperationError
 from taichi.lang.impl import *
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -3,8 +3,8 @@ import os
 from copy import deepcopy as _deepcopy
 
 from taichi.core.util import ti_core as _ti_core
-from taichi.lang.enums import *
 from taichi.lang import impl
+from taichi.lang.enums import *
 from taichi.lang.exception import InvalidOperationError
 from taichi.lang.impl import *
 from taichi.lang.kernel_arguments import any_arr, ext_arr, template
--- a/python/taichi/lang/ext_array.py
+++ b/python/taichi/lang/ext_array.py
@@ -72,7 +72,9 @@ class AnyArray:
         if element_dim == 0:
             return ret
         else:
-            return ret[element_dim:] if self.layout == Layout.SOA else ret[:-element_dim]
+            return ret[
+                element_dim:] if self.layout == Layout.SOA else ret[:
+                                                                    -element_dim]
 
 
 class AnyArrayAccess:
--- a/python/taichi/lang/ext_array.py
+++ b/python/taichi/lang/ext_array.py
@@ -1,4 +1,5 @@
 from taichi.core.util import ti_core as _ti_core
+from taichi.lang.enums import Layout
 from taichi.lang.expr import Expr, make_expr_group
 from taichi.lang.util import taichi_scope
 
@@ -46,13 +47,13 @@ class AnyArray:
     Args:
         ptr (taichi_core.Expr): A taichi_core.Expr wrapping a taichi_core.ExternalTensorExpression.
         element_shape (Tuple[Int]): () if scalar elements (default), (n) if vector elements, and (n, m) if matrix elements.
-        is_soa (bool): Whether to change the layout from AOS (default) to SOA.
+        layout (Layout): Memory layout.
     """
-    def __init__(self, ptr, element_shape, is_soa):
+    def __init__(self, ptr, element_shape, layout):
         assert ptr.is_external_var()
         self.ptr = ptr
         self.element_shape = element_shape
-        self.is_soa = is_soa
+        self.layout = layout
 
     @property
     @taichi_scope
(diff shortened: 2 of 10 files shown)


python/taichi/lang/impl.py
```diff
@@ -607,20 +607,6 @@ def ndarray(dtype, shape):
     return ScalarNdarray(dtype, shape)
 
 
-class Layout:
-    def __init__(self, soa=False):
-        self.soa = soa
-
-
-SOA = Layout(soa=True)
```
k-ye, line 615: Let's not deprecate this yet, or it would break users. Could you annotate this with a deprecated warning?

python/taichi/lang/__init__.py
```diff
@@ -4,6 +4,7 @@
 
 from taichi.core.util import ti_core as _ti_core
 from taichi.lang import impl
+from taichi.lang.enums import *
```
k-ye, line 7:
```suggestion
from taichi.lang.enums import Layout
```

k-ye (approved):
Great! Do we need this on the C++ side?

strongoier:
> Great! Do we need this on the C++ side?

No. `Layout` is for `Matrix`, and `Matrix` is a pure Python frontend thing...

k-ye:
I see. In the future we likely need to move Matrix/Vector into C++ as well :-)

python/taichi/lang/impl.py
strongoier: Do you mean we still keep this class for use? Or should I directly refer `ti.SOA` to `ti.Layout.SOA`? (But in this case I don't know how to give a deprecated warning)

k-ye: This can be a fancy way to achieve that: https://stackoverflow.com/questions/922550/how-to-mark-a-global-as-deprecated-in-python

python/taichi/lang/ext_array.py
```diff
@@ -71,7 +72,9 @@ def shape(self):
         if element_dim == 0:
             return ret
         else:
-            return ret[element_dim:] if self.is_soa else ret[:-element_dim]
+            return ret[
+                element_dim:] if self.layout == Layout.SOA else ret[:
```
ailzhang, line 76: lol yapf has done some hard work here...

python/taichi/lang/matrix.py
```diff
@@ -27,7 +27,7 @@ class Matrix(TaichiOperations):
         shape ( Union[int, tuple of int], optional): the shape of a matrix field.
         offset (Union[int, tuple of int], optional): The coordinate offset of all elements in a field.
         empty (Bool, deprecated): True if the matrix is empty, False otherwise.
-        layout (TypeVar, optional): The filed layout(AOS or SOA).
+        layout (TypeVar, optional): The filed layout (AOS or SOA).
```
ailzhang, line 30: nit: Layout.AOS or Layout.SOA

python/taichi/lang/matrix.py
```diff
@@ -914,7 +914,7 @@ def field(cls,
               name="",
               offset=None,
               needs_grad=False,
-              layout=None):  # TODO(archibate): deprecate layout
+              layout=Layout.AOS):  # TODO(archibate): deprecate layout
```
ailzhang, line 917: is this TODO still valid after we change it to Layout enum?

python/taichi/lang/matrix.py
```diff
@@ -925,7 +925,7 @@ def field(cls,
             name (string, optional): The custom name of the field.
             offset (Union[int, tuple of int], optional): The coordinate offset of all elements in a field.
             needs_grad (bool, optional): Whether the Matrix need gradients.
-            layout (:class:`~taichi.lang.impl.Layout`, optional): The field layout, i.e., Array Of Structure(AOS) or Structure Of Array(SOA).
+            layout (Layout, optional): The field layout, i.e., Array Of Structure (AOS) or Structure Of Array (SOA).
```
ailzhang, line 928: If `:class:` still works, it might be good to keep it since it'll link to Layout docstring on the website?

ailzhang (approved):
LGTM!

python/taichi/lang/matrix.py
strongoier: I don't know what was planned in the long term from this TODO..

python/taichi/lang/matrix.py
strongoier: well type annotations in the parentheses can be automatically linked..

python/taichi/lang/impl.py
strongoier: Wow interesting. However

> The downside is that you are now performing two lookups when you access the module, so there is a slight performance hit.

Can we afford this?

python/taichi/lang/matrix.py
k-ye: Let's remove that. We do want to support AOS/SOA layout

python/taichi/lang/impl.py
k-ye: Should be good :-)

strongoier: A problem arises - If we wrap `impl` as such class, then we can no longer use `from taichi.lang.impl import *` which is currently everywhere now. And we cannot expose things inside as `ti.xxx` either.

k-ye: OK, let's just make `AOS = Layout.AOS` in `__init__.py` for now

strongoier: Well I solved it just now - there's a new feature in Python 3.7 which allows us to directly add `__getattr__` to modules. Now I added a general way to handle deprecated names - directly add the mapping to `deprecated_names` in `taichi/__init__.py`.

python/taichi/lang/__init__.py
strongoier: done

python/taichi/lang/matrix.py
strongoier: done

python/taichi/lang/matrix.py
strongoier: done

python/taichi/lang/impl.py
k-ye: We support 3.6....

strongoier: Oh... Maybe directly do `AOS = Layout.AOS` for 3.6?

k-ye: Sounds good

PR #2786, opened 2021-08-24
[ci] Changed nightly tag from pre-release to post-release

Leonz5288:
Related issue = #

### DIFF
--- a/python/build.py
+++ b/python/build.py
@@ -48,7 +48,9 @@ def build(project_name):
             f'cd ..; {get_python_executable()} setup.py {project_tag} bdist_wheel -p manylinux1_x86_64'
         )
     else:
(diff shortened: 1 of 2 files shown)


python/build.py
```diff
@@ -39,13 +39,18 @@ def build(project_name):
         f'{get_python_executable()} ../misc/make_changelog.py origin/master ../ True'
     )
 
+    os.environ['PROJECT_NAME'] = project_name
```
ailzhang, line 42: Nit: it'd be nice if we can add a comment saying this env is used by `setup.py` below. It's easy to forget updating this env when we change setup.py below.

ailzhang (approved):
LGTM! Thanks for fixing it!

PR #2789, opened 2021-08-24
[vulkan] Graphics Device API

AmesingFlank:
This PR adds API definitions for Graphics device and the vulkan implemention.

credits @bobcao3 who proposed this api and provided the vk impl

### DIFF
--- a/taichi/backends/device.cpp
+++ b/taichi/backends/device.cpp
@@ -21,5 +21,30 @@ void Device::memcpy(DevicePtr dst, DevicePtr src, uint64_t size) {
 
 #endif
 }
+
+void GraphicsDevice::image_transition(DeviceAllocation img,
+                                      ImageLayout old_layout,
+                                      ImageLayout new_layout) {
+  auto cmd_list = new_command_list({CommandListType::Graphics});
+  cmd_list->image_transition(img, old_layout, new_layout);
+  submit_synced(cmd_list.get());
+}
+void GraphicsDevice::buffer_to_image(DeviceAllocation dst_img,
+                                     DevicePtr src_buf,
+                                     ImageLayout img_layout,
+                                     const BufferImageCopyParams &params) {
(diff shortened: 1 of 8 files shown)


taichi/backends/device.h
```diff
@@ -253,16 +398,74 @@ class Surface {
...
+  BufferFormat format;
+  uint32_t offset{0};
+};
+
+struct SurfaceConfig {
+  uint32_t width;
+  uint32_t height;
+  bool vsync;
+  void *window_handle;
+};
```
g1n0st, line 424: Should these have initial values?

taichi/backends/device.cpp
```diff
@@ -21,5 +21,30 @@ void Device::memcpy(DevicePtr dst, DevicePtr src, uint64_t size) {
...
+}
+void GraphicsDevice::image_to_buffer(DevicePtr dst_buf,
+                                     DeviceAllocation src_img,
+                                     ImageLayout img_layout,
+                                     const BufferImageCopyParams &params) {
+  auto cmd_list = new_command_list({CommandListType::Graphics});
+  cmd_list->image_to_buffer(dst_buf, src_img, img_layout, params);
+  submit_synced(cmd_list.get());
+}
+
```
KuribohG, line 48: These APIs will get new command list and submit it immediately. Will this cause too many command lists submitted in one frame?

bobcao3: it will, but this is the "immediate mode" API. If possible we should avoid this and use the image_to_buffer from the command list directly. A potential issue with the current design is that each Taichi kernel dispatch is a unique command list. In TVM's design the commands are accumulated into one buffer and dispatched together upon a sync. Since we don't have a good sync boundary, the current design will start dispatching work to the GPU immediately but not necessarily wait for them to complete. That said, this is a choice by the taichi backend not the RHI, the RHI will support all cases.

g1n0st (approved):

PR #2791, opened 2021-08-24
[ci] Added gpu test timeout

Leonz5288:
Related issue = #


ailzhang:
Nit: one way to test whether if this will work before merging is - set the timeout to a small number (say 10mins) and see whether it'll be killed as expected ;)

ailzhang (approved):
Looks like it worked! Thanks so much!!

k-ye:
Probably need to restrict M1 as well... https://github.com/taichi-dev/taichi/runs/3408750566?check_suite_focus=true :joy:

PR #2793, opened 2021-08-24
[gui] GGUI 5/n: remove some stuff

AmesingFlank:
Related issue = #2646 

This is the 5th of a series of PRs that adds a GPU-based GUI to taichi. 

Since GGUI 4/n, the system had went through some major changes. This PR mainly removes the code from previous PRs that are no longer needed (because most of them have been moved to device api), and make some of the pre-existing code compatible with the new system.

The PR also slightly changes the point lights API so that there're no limits to the amount of lights, and make the cuda-vk interop functions take in an additional offset parameter


k-ye (approved):
LGTM!

PR #2798, opened 2021-08-25
[gui] GGUI 8/n: Renderable class

AmesingFlank:
Related issue = #2646 

This is the 8th of a series of PRs that adds a GPU-based GUI to taichi.

This PR adds a `Renderable` class, which manages a graphics pipeline, a VBO, an IBO, and optionally an UBO and a SSBO. It also implements methods of updating the VBO and IBO using data from a taichi field (represented by a `FieldInfo`)

### DIFF
--- a/taichi/ui/backends/vulkan/renderable.cpp
+++ b/taichi/ui/backends/vulkan/renderable.cpp
@@ -1,0 +1,301 @@
+#include "taichi/ui/backends/vulkan/renderable.h"
+#include "taichi/ui/utils/utils.h"
+
+#include "taichi/ui/backends/vulkan/vulkan_cuda_interop.h"
+#include "taichi/ui/backends/vulkan/renderables/kernels.h"
+
+TI_UI_NAMESPACE_BEGIN
+
+namespace vulkan {
+
+using namespace taichi::lang;
+using namespace taichi::lang::vulkan;
+
+void Renderable::init(const RenderableConfig &config, AppContext *app_context) {
+  config_ = config;
+  app_context_ = app_context;
+}
+
+void Renderable::init_render_resources() {
+  create_graphics_pipeline();
+  init_buffers();
+}
+
+void Renderable::free_buffers() {
+  app_context_->device().dealloc_memory(vertex_buffer_);
+  app_context_->device().dealloc_memory(staging_vertex_buffer_);
+  app_context_->device().dealloc_memory(index_buffer_);
+  app_context_->device().dealloc_memory(staging_index_buffer_);
+
+  destroy_uniform_buffers();
+  destroy_storage_buffers();
+}
+
+void Renderable::init_buffers() {
+  create_vertex_buffer();
+  create_index_buffer();
+  create_uniform_buffers();
+  create_storage_buffers();
+
+  create_bindings();
+
+  if (app_context_->config.ti_arch == Arch::cuda) {
+    auto [vb_mem, vb_offset, vb_size] =
+        app_context_->device().get_vkmemory_offset_size(vertex_buffer_);
+
+    auto [ib_mem, ib_offset, ib_size] =
+        app_context_->device().get_vkmemory_offset_size(index_buffer_);
+
+    auto block_size = VulkanDevice::kMemoryBlockSize;
+
+    vertex_buffer_device_ptr_ =
+        (Vertex *)get_memory_pointer(vb_mem, block_size, vb_offset, vb_size,
+                                     app_context_->device().vk_device());
+    index_buffer_device_ptr_ =
+        (int *)get_memory_pointer(ib_mem, block_size, ib_offset, ib_size,
+                                  app_context_->device().vk_device());
+  }
+}
+
+void Renderable::update_data(const RenderableInfo &info) {
+  int num_vertices = info.vertices.shape[0];
+  int num_indices;
+  if (info.indices.valid) {
+    num_indices = info.indices.shape[0];
+    if (info.indices.dtype != PrimitiveType::i32 &&
+        info.indices.dtype != PrimitiveType::u32) {
+      throw std::runtime_error("dtype needs to be 32-bit ints for indices");
+    }
+  } else {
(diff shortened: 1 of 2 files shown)


taichi/ui/backends/vulkan/renderable.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+  RasterParams raster_params;
+  raster_params.prim_topology = config_.topology_type;
+  raster_params.depth_test = true;
+  raster_params.depth_write = true;
+
+  std::vector<VertexInputBinding> vertex_inputs = {{0, sizeof(Vertex), false}};
+  std::vector<VertexInputAttribute> vertex_attribs = {
+      {0, 0, BufferFormat::rgb32f, offsetof(Vertex, pos)},
+      {1, 0, BufferFormat::rgb32f, offsetof(Vertex, normal)},
```
bobcao3, line 204: Can we potentially do RGB8 or RGBA8 for normals and colors? Should be doable since these buffers are coming from a CUDA kernel right

taichi/ui/backends/vulkan/renderable.cpp
```diff
@@ -0,0 +1,301 @@
...
+        indexed_ = false;
+      }
+      app_context_->device().unmap(staging_index_buffer_);
+    }
+    app_context_->device().memcpy(vertex_buffer_.get_ptr(0),
+                                  staging_vertex_buffer_.get_ptr(0),
+                                  config_.vertices_count * sizeof(Vertex));
+    app_context_->device().memcpy(index_buffer_.get_ptr(0),
+                                  staging_index_buffer_.get_ptr(0),
+                                  config_.indices_count * sizeof(int));
```
bobcao3, line 163: btw when needing more than 1 copy, we should get a new command list and record two copy commands. The device memcpy call is synced and we should try to reduce the number of command lists submitted

taichi/ui/backends/vulkan/renderable.cpp
```diff
@@ -0,0 +1,301 @@
...
+    if (info.indices.valid) {
+      indexed_ = true;
+      update_renderables_indices_cuda(index_buffer_device_ptr_,
+                                      (int *)info.indices.data, num_indices);
+    } else {
+      indexed_ = false;
+    }
+
+  } else if (info.vertices.field_source == FieldSource::TaichiX64) {
+    {
```
bobcao3, line 120: looks like a good place to let in numpy arrays & not supported backends

taichi/ui/backends/vulkan/renderable.cpp
AmesingFlank: Yep we can. Although do you think it's necessary to do that right now? Currently I'm using a single kernel to copy different vertex attributes: 

https://github.com/taichi-dev/taichi/blob/5d4796e0c852cd021502c11fc82d57a44c8a8e76/taichi/ui/backends/vulkan/renderables/kernels.h#L5-L14

taichi/ui/backends/vulkan/renderable.cpp
AmesingFlank: good point, will fix.

taichi/ui/backends/vulkan/renderable.cpp
AmesingFlank: Yep.

Although I intended `FieldSource::TaichiX64` to represent fields directly from a `ti.field` from a x64 backend. I will add a TODO for a `FieldSource::Numpy` here.

I suspected that the data from a x64 field could be different from those dumped to an np array. There're a few things at play: power-of-2-fields, packed mode, and multiple level dense layouts... I haven't given much thought into how to organise support for all of this.

taichi/ui/backends/vulkan/renderable.cpp
bobcao3: It would be less data to move around, you should be able to do that from within the same kernel.

AmesingFlank: OK. And do we really wanna do this for normals? Does 8 bit normals have visible artifacts?

bobcao3: should be enough at least for GGUI since our lighting don't rely on normals that much.

AmesingFlank: we didn't define rgb8 in `device.h`...

bobcao3: Use RGBA8, RGB8 is a 3 byte awkward format

AmesingFlank: Let's leave this for now and add a TODO. I don't want to have to change device api and vulkan device in this PR..

bobcao3: We should not change device API. RGB8 don't exist in Vulkan images as well since it's not aligned. Usual practice is just use RGBA8 and ignore the alpha channel

AmesingFlank: Screenshot from 2021-08-25 04-23-06
Screenshot from 2021-08-25 04-26-15

AmesingFlank: oops sorry for the dual screen.

AmesingFlank: tried using 8 bit normals, looks really weird around the edges.

AmesingFlank: gonna leave this for now :)

bobcao3: Right by default it rgba8 means rgba8unorm, you will need the signed normalized version or do the range conversion (-128..127) to (0..255) manually

AmesingFlank: K. will open a separate pr to add a `BufferFormat::rgba8signed` and fix this :)

Issue #2771, opened 2021-08-23
Add `ti.ndarray` as an alternative to Taichi dense fields


strongoier:
**Concisely describe the proposed feature**
I would like to add `ti.ndarray`/`ti.Vector.ndarray`/`ti.Matrix.ndarray` as an alternative to dense `ti.field`/`ti.Vector.field`/`ti.Matrix.field`. From a user's perspective, `ti.ndarray` will have at least the same behavior as dense `ti.field`. However, `ti.ndarray` will be implemented using contiguous memory, which will be better aligned with external arrays like numpy arrays or torch tensors, and provide more flexibility to support operations like slicing (#2669).

**Describe the solution you'd like (if any)**
First use `torch.Tensor` as internal implementation to avoid memory allocation/deallocation, and finish the prototype.

- [x] Scalar Ndarray. (#2676, #2703)
- [ ] Vector/Matrix Ndarray.

Then write our own memory allocator, and support operations like slicing.

- [ ] Memory allocator.
- [ ] Slicing.

PR #2808, opened 2021-08-26
[Lang] Add ti.Vector.ndarray and ti.Matrix.ndarray

strongoier:
Related issue = #2771

This PR adds `ti.Vector.ndarray` and `ti.Matrix.ndarray`, which support Python-scope and Taichi-scope access. When passing them into kernels, we need to use `ti.any_arr()` annotation with `element_shape` and optional `layout` (default value: `ti.Layout.AOS`) to enforce type consistency.
```python
@ti.kernel
def func(a: ti.any_arr(element_shape=(2, 2), layout=ti.Layout.SOA)):
    for i in range(5):
        for j, k in ti.ndrange(2, 2):
            a[i][j, k] = j * j + k * k

m = ti.Matrix.ndarray(2, 2, ti.i32, 5, layout=ti.Layout.SOA)
func(m)
```


python/taichi/lang/ndarray.py
```diff
@@ -84,11 +81,32 @@ def shape(self):
...
 
     def __repr__(self):
         return '<ti.ndarray>'
+
+
+class NdarrayHostAccess:
+    """Class for accessing VectorNdarray/MatrixNdarray in Python scope.
+    Args:
+        arr (Union[VectorNdarray, MatrixNdarray]): See above.
+        indices_first (Tuple[Int]): Indices of first-level access.
```
k-ye, line 98: What are first level and second level?

python/taichi/lang/kernel_arguments.py
```diff
@@ -57,8 +58,25 @@ def __init__(self, element_shape=(), layout=Layout.AOS):
         self.layout = layout
 
     def extract(self, x):
-        shape = tuple(x.shape)
+        from taichi.lang.matrix import MatrixNdarray, VectorNdarray
```
k-ye, line 61: move to top? (or will it cause circular import?)

strongoier: It will cause circular import :-(

python/taichi/lang/ndarray.py
strongoier: First level refers to coordinates in the field, while second level refers to indices inside the vector/matrix. Maybe I should make it clearer?

strongoier: Using first/second is to tell users this is what they write in the first/second pair of brackets.

tests/python/test_ndarray.py
```diff
@@ -29,3 +111,95 @@ def run(x: ti.any_arr(), y: ti.any_arr()):
...
+    v = ti.Vector.ndarray(10, ti.i32, 5, layout=layout)
+    func(v)
+    assert v[0][6] == 0
+    assert v[1][0] == 0
+    assert v[2][1] == 1
+    assert v[3][4] == 4
+    assert v[4][9] == 9
+
+
+# number of compiled functions
```
ailzhang, line 184: is this comment not longer needed?

python/taichi/lang/matrix.py
```diff
@@ -1435,3 +1482,96 @@ def empty(self):
...
+    def __getitem__(self, key):
+        key = () if key is None else (
+            key, ) if isinstance(key, numbers.Number) else tuple(key)
+        return Matrix.with_entries(self.n, self.m, [
+            NdarrayHostAccess(self, key, (i, j)) for i in range(self.n)
+            for j in range(self.m)
+        ])
+
+    def __repr__(self):
+        return f'<{self.n}x{self.m} ti.Matrix.ndarray>'
```
ailzhang, line 1535: IMHO printing out layout information will also help?

python/taichi/lang/matrix.py
```diff
@@ -1435,3 +1482,96 @@ def empty(self):
...
+
+    @python_scope
+    def __getitem__(self, key):
+        key = () if key is None else (
+            key, ) if isinstance(key, numbers.Number) else tuple(key)
+        return Matrix.with_entries(
+            self.n, 1,
+            [NdarrayHostAccess(self, key, (i, )) for i in range(self.n)])
+
+    def __repr__(self):
```
ailzhang, line 1576: ditto

ailzhang (commented):
LGTM!

tests/python/test_ndarray.py
strongoier: It's used to separate three different parts of tests of this file.

python/taichi/lang/matrix.py
strongoier: Nice suggestion. Done.

python/taichi/lang/matrix.py
strongoier: done

python/taichi/lang/ndarray.py
strongoier: Explanations are added.

ailzhang (approved):
LGTM!

PR #2826, opened 2021-08-28
[Lang] Let rescale_index support SNode as input parameter

Jack12xl:
Related issue = 

this pr basically let `ti.rescale_index` support `SNode` as input parameter. The reason is that sometimes an `SNode` might not have '**placed**' a `Field` but we want to rescale the cell index based on the `SNode` anyway.

See the [example](https://github.com/Jack12xl/taichi_elements/blob/8c4a16cc0bfb45a667ec0d62e248c445bcea53e2/engine/mpm_solver.py#L321) from the **Taichi_elements**. We need to rescale the index from `pid(ti.root.pointer.pointer.dynamic)` to `block(ti.root.pointer.pointer)`. But `block` did not have a cell placed beneath (but we still need to rescale according to `block`).


strongoier (approved):
LGTM! Thanks!

PR #2831, opened 2021-08-28
[util] fix ModuleNotFoundError in Windows

quadpixels:
Add back the code that copies taichi_core.pyd from VS build folder
to the binary directory (TAICHI_REPO_DIR/runtimes) and add the
binary directory to sys.path in order to fix the ModuleNotFoundError.

Tested: Built with cmake and VS in Windows 10, then ran Python as
follows:
```
C:\Users\USER_NAME\taichi_my>python
Python 3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:59:51) [MSC v.1914 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import taichi
[Taichi] version 0.0.0, llvm 10.0.0, commit 10c2cbb2, win, python 3.7.0
>>>
```

Related issue = #

### DIFF
--- a/python/taichi/core/util.py
+++ b/python/taichi/core/util.py
@@ -130,6 +130,27 @@ if settings.get_os_name() != 'win':
     # For llvm jit to find the runtime symbols
     if not os.path.exists(link_dst):
         os.symlink(link_src, link_dst)
+else:
+    bin_dir = os.path.join(settings.get_repo_directory(), 'runtimes')
+    possible_folders = ['Debug', 'RelWithDebInfo', 'Release']
+    detected_dlls = []
+    for folder in possible_folders:
+        dll_path = os.path.join(bin_dir, folder, 'taichi_core.dll')
+        if os.path.exists(dll_path):
+            detected_dlls.append(dll_path)
+    if len(detected_dlls) == 0:
+        raise FileNotFoundError(
+            f'Cannot find Taichi core dll under {bin_dir}/{possible_folders}'
+        )
+    elif len(detected_dlls) != 1:
+        print('Warning: multiple Taichi core dlls found: %s' %
+            ','.join(detected_dlls))
+        print(f'Using {detected_dlls[0]}')
+    dll_path = detected_dlls[0]
+    old_wd = os.getcwd()
+    os.chdir(bin_dir)
+    shutil.copy(dll_path, os.path.join(bin_dir, 'taichi_core.pyd'))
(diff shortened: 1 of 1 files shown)


ailzhang:
Hey @quadpixels, thanks for contributing! May I ask how did you build taichi from source? We'd recommend using `python setup.py develop` instead of the old `TAICHI_REPO_DIR` way. Would you mind giving it a try?   https://docs.taichi.graphics/docs/lang/articles/contribution/dev_install

quadpixels:
> 
> 
> Hey @quadpixels, thanks for contributing! May I ask how did you build taichi from source? We'd recommend using `python setup.py develop` instead of the old `TAICHI_REPO_DIR` way. Would you mind giving it a try? https://docs.taichi.graphics/docs/lang/articles/contribution/dev_install

Hello!
Yes I built by `cmake ..` and then open taichi.sln in VS, switched configuration to RelWithDbgInfo, then built the solution.
I tried the `python setup.py develop --user` method as suggested and am able to build the module just like I did with cmake and VS before; a small difference appears to be this method uses the Release configuration by default. When I run `python -m taichi`, it seems it is still calling the `python/taichi/core/util.py` during module import, so the error this PR is trying to resolve still applies to the `python setup.py develop --user` method.
Is there anything else I need to test?
Thanks!

ailzhang:
Hey @quadpixels, thanks a ton for the detailed response! 
>  a small difference appears to be this method uses the Release configuration by default.  
FYI if you do `DEBUG=1 python setup.py develop --user` it should build with the DEBUG mode. ;) 

One more thing that I'm curious, do this issue still repro if you use `python setup.py install`? We'd definitely want to fix the issue for develop mode, but I'm curious why it didn't show up on our windows CI. :D

k-ye:
@bobcao3 Hmm, have you encountered similar issue when building on Win?

quadpixels:
I can check that the issue I'm trying to resolve was due to my own error. I guess it's because `setup.py` did not run till the end when I ran it.

Confirming that the ModuleNotFoundError issue does not exist when `setup.py` runs correctly. Module works correctly after `setup.py` copies `runtimes/Release/taichi_core.dll` to `python\taichi/lib/taichi_core.pyd`, which is the pyddir in line 30 of util.py (`pyddir = os.path.join(package_root(), 'lib')`).

Closing and abandoning this change.

PR #2835, opened 2021-08-29
[refactor] Decouple opengl codegen compile with Program.

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #2839
* #2837
* #2836
* __->__ #2835
* #2834
* #2833

PR #2836, opened 2021-08-29
[refactor] Decouple MemoryPool with Program.

ailzhang:
* #2839
* #2837
* __->__ #2836
* #2835
* #2834
* #2833

### DIFF
--- a/taichi/program/program.cpp
+++ b/taichi/program/program.cpp
@@ -39,13 +39,6 @@
 
 namespace taichi {
 namespace lang {
-namespace {
-inline uint64 *allocate_result_buffer_default(Program *prog) {
-  return (uint64 *)prog->memory_pool->allocate(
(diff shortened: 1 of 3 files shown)


taichi/program/program.cpp
```diff
@@ -39,13 +39,6 @@
 
 namespace taichi {
 namespace lang {
-namespace {
-inline uint64 *allocate_result_buffer_default(Program *prog) {
```
k-ye, line 43: nit: Maybe keep this but pass in the `memory_pool` itself? Your call..

k-ye (approved):
LGTM!

taichi/program/program.cpp
ailzhang: Was planning to do this memory allocation in each arch's impl so separated them out. ;) Let's keep it for now, we might be able to unify them to a proper API (possibly in memory_pool) in the future.

PR #2843, opened 2021-08-30
[ci] Fixed bug of wrong os parameter

Leonz5288:
Related issue = #


ailzhang (approved):
Thanks for fixing this! 
Btw we probably need to think about how to shake out all the bugs before a release or merging changes to release.yaml. https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release Maybe it'll make it easier if we can figure out a way to trigger this workflow before "a release is published"?

PR #2844, opened 2021-08-30
[ci] Enable verbose on M1 CI to collect more info on hanging jobs.

ailzhang:
I've occasionally seen M1 job hanging in some tests and finally timed out. Enabling verbose option here so that we can have more information about the hang.


k-ye (approved):
I think `-v` itself is fine. retry usually exacerbates the problem, IIRC..

ailzhang:
> I think `-v` itself is fine. retry usually exacerbates the problem, IIRC..

Agh agreed, was thinking that we can increase the failure rate to get more info for now :P

Leonz5288 (approved):
lgtm

PR #2848, opened 2021-08-30
[CUDA] [benchmark] CUPTI 1/3 : an experimental kernel profiler that provides low and deterministic profiling overhead with more accurate results

yolo2themoon:
This is the 1st of a series of PRs that adds CUPTI kernel profiler to taichi for CUDA backend

1st : python API and cpp code ( without the actual call to CUPTI ) 
2nd : add CUPTI headers and APIs
3rd : doc and misc   

now we can use : 
` ti.init(kernel_profiler = True) `  [ CUDA::cuEvent ]
` ti.init(kernel_profiler = ti.profiler.cuda_event) `  [ CUDA::cuEvent ]
` ti.init(kernel_profiler = ti.profiler.cupti_onepass) ` [ CUPTI::onepass ]
` ti.init(kernel_profiler = ti.profiler.cupti_detailed) ` [ CUPTI::detailed ]
to select different kernel profiler


python/taichi/core/__init__.py
```diff
@@ -1,5 +1,6 @@
 from taichi.core.logging import *
 from taichi.core.primitive_types import *
+from taichi.core.profiler_modes import *
```
k-ye, line 3: Don't do this please

taichi/backends/cuda/cuda_context.h
```diff
@@ -26,13 +27,21 @@ class CUDAContext {
   std::mutex lock;
   KernelProfilerBase *profiler;
   CUDADriver &driver;
+  CUDAProfiler &cuda_kernel_profiler;
```
k-ye, line 30: `_`

taichi/backends/cuda/cuda_profiler.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+CUDAProfiler &CUDAProfiler::get_instance_without_context() {
+  static CUDAProfiler *instance = new CUDAProfiler();
+  return *instance;
+}
+
+CUDAProfiler &CUDAProfiler::get_instance() {
+  CUDAContext::get_instance();
+  return get_instance_without_context();
```
k-ye, line 112: Any specific reason we need to do this? If possible, I'd like us to avoid this

k-ye (commented):
Just a quick pass, will read more carefully later :-)

python/taichi/core/profiler_modes.py
```diff
@@ -0,0 +1,20 @@
...
+cuda_detailed = _ti_core.KernelProfilerMode.cuda_detailed
+
+kernel_profiler_enums = [
+    kernel_profiler_disable, kernel_profiler_enable, cuda_accurate,
+    cuda_detailed
+]
+kernel_profiler_enums_ids = [id(t) for t in kernel_profiler_enums]
+
+__all__ = [
+    'kernel_profiler_disable',
```
ailzhang, line 16: Hmmm personally I'd prefer these enum names to be more explicit, e.g. `DISABLED, CUEVENT, CUPTI_ONEPASS, CUPTI_DETAILED` so that I know what I'm calling (and potentially easier to add new profiling modes in the future ;) Wdyt?

taichi/program/program.cpp
```diff
@@ -68,9 +70,19 @@ Program::Program(Arch desired_arch) : snode_rw_accessors_bank_(this) {
   if (config.debug)
     config.check_out_of_bound = true;
 
-  profiler = make_profiler(config.arch);
+  profiler = make_profiler(config.arch, config.kernel_profiler);
   llvm_program_ = std::make_unique<LlvmProgramImpl>(config, profiler.get());
 
+#if defined(TI_WITH_CUDA)
```
ailzhang, line 76: Can you move this back to llvm_program.cpp? :P We plan to decouple backend-specific logic from Program class.

taichi/backends/cuda/cuda_context.h
ailzhang: Hmm would you mind explaining a bit why we have two profilers in CUDAcontext? It'd be nice if we can tell the difference from their name or adding a note :D

python/taichi/core/profiler_modes.py
yolo2themoon: Of course. Enum names have not yet been finalized, perhaps a short discussion is needed.

yolo2themoon: `disable` and `enable` are used for compatibility with the current `True` and `False`. In general, we can follow the current method: `kernel_profiler = True`

yolo2themoon: For CUDA backend, `enable` and `CUEVENT` (default kernel profiler) are equivalent, we can also add this enum name to explicitly specify CUEVENT profiler.

taichi/program/program.cpp
yolo2themoon: Previously, I encountered a segfault with a null pointer ( `profiler` ). I will move the code back and try to solve the problem

yolo2themoon:
> Hmm would you mind explaining a bit why we have two profilers in CUDAcontext? It'd be nice if we can tell the difference from their name or adding a note :D
_Originally posted by @ailzhang in https://github.com/taichi-dev/taichi/pull/2848#discussion_r698946279_

`CUDAProfiler` has its own back-end methods that different from `KernelProfilerBase`. 
This commit tries not to change `KernelProfilerBase`, so this new profiler class is added.
I will think about how to better express and distinguish these two classes.

taichi/backends/cuda/cuda_context.h
yolo2themoon: `CUDAProfiler` has its own back-end methods that different from `KernelProfilerBase`.
This commit tries not to change `KernelProfilerBase`, so this new profiler class is added.
I'll think about how to better express and distinguish these two classes.

taichi/backends/cuda/cuda_profiler.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+CUDAProfiler::~CUDAProfiler() {
+  if (profiler_config_.profiler_type == CUDA_KERNEL_PROFILER_CUPTI) {
+    end_profiling();
+    deinit_cupti();
+  }
+}
+
+bool CUDAProfiler::set_profiling_mode(KernelProfilerMode profiling_mode) {
+  bool enable = profiling_mode == KernelProfilerMode::enable |
```
ailzhang, line 19: I'd encapsulate this logic into a method on KernelProfilerMode::is_enabled() so that we only need to change one place when adding a new mode.

taichi/backends/cuda/cuda_profiler.cpp
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  CUDAKernalProfiler profiler_type =
+      (profiling_mode == KernelProfilerMode::cuda_accurate) ||
+              (profiling_mode == KernelProfilerMode::cuda_detailed)
+          ? CUDA_KERNEL_PROFILER_CUPTI
+          : CUDA_KERNEL_PROFILER_EVENT;
+  profiler_config_.profiling_mode = profiling_mode;
+
+  if (enable && profiler_type == CUDA_KERNEL_PROFILER_EVENT) {
```
ailzhang, line 33: Since we return instantly when enable==false, it's probably safe to assume enable=true after line 22?

taichi/backends/cuda/cuda_profiler.cpp
```diff
@@ -0,0 +1,150 @@
...
+    profiler_config_.enable = true;
+    profiler_config_.profiler_type = CUDA_KERNEL_PROFILER_EVENT;
+    profiler_config_.profiling_mode = KernelProfilerMode::enable;
+    TI_TRACE("CUDA_KERNEL_PROFILER_EVENT : enable");
+    TI_TRACE("CUDAProfiler enable = {}", profiler_config_.enable);
+    TI_TRACE("profiler_type : {}", profiler_config_.profiler_type);
+    return true;
+  }
+
+#if !defined(TI_WITH_TOOLKIT_CUDA)
```
ailzhang, line 43: Hmmm it's not clear yet when we should define TI_WITH_TOOLKIT_CUDA, IIUC this should happen in cmake in your next PR when adding CUPTI?

taichi/backends/cuda/cuda_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+  CUDA_KERNEL_PROFILER_UNDEF = 0,
+  CUDA_KERNEL_PROFILER_EVENT = 1,
+  CUDA_KERNEL_PROFILER_CUPTI = 2
+} CUDAKernalProfiler;
+
+struct profilerConfig {
+  bool enable = false;
+  CUDAKernalProfiler profiler_type = CUDA_KERNEL_PROFILER_UNDEF;
+  KernelProfilerMode profiling_mode = KernelProfilerMode::disable;
+  uint32_t num_ranges = 1024 * 16;
```
k-ye, line 23: ?

taichi/backends/cuda/cuda_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+  std::vector<std::string> metric_names;
+#if defined(TI_WITH_TOOLKIT_CUDA)
+  CUpti_ProfilerRange profiler_range = CUPTI_AutoRange;
+  CUpti_ProfilerReplayMode profiler_replay_mode = CUPTI_KernelReplay;
+#endif
+};
+
+struct CUDAKernelTracedRecord {
+  std::string kernel_name;
+  float kernel_elapsed_time_in_ms;
```
k-ye, line 33: initialize these member variables

taichi/backends/cuda/cuda_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+  double kernel_gloabl_load_byets;
+  double kernel_gloabl_store_byets;
+  float utilization_ratio_sm;
+  float utilization_ratio_mem;
+};
+
+struct profilerRawData {
```
k-ye, line 40: naming convention (class, member)

taichi/backends/cuda/cuda_profiler.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+bool CUDAProfiler::printTracedRecords() {
+  size_t result_nums = traced_records_.size();
+  if (!result_nums) {
+    TI_WARN("traced_records_ is empty!");
+    return false;
+  }
+
+  for (size_t resultIndex = 0; resultIndex < result_nums; ++resultIndex) {
+    TI_INFO("Kernel: name[{}] time[{:6.6f}]ms ldg[{}]bytes stg[{}]bytes",
```
ailzhang, line 123: Maybe add a print() method for CUDAKernelTracedRecord?

taichi/backends/cuda/cuda_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+  std::vector<uint8_t> counterDataScratchBuffer;
+  std::vector<uint8_t> counterAvailabilityImage;
+};
+
+class CUDAProfiler {
+ public:
+  CUDAProfiler();
+  ~CUDAProfiler();
+
+  static CUDAProfiler &get_instance();
```
k-ye, line 54: Do we have to use singleton?

taichi/backends/cuda/cuda_profiler.cpp
```diff
@@ -0,0 +1,150 @@
...
+  static CUDAProfiler *instance = new CUDAProfiler();
+  return *instance;
+}
+
+CUDAProfiler &CUDAProfiler::get_instance() {
+  CUDAContext::get_instance();
+  return get_instance_without_context();
+}
+
+bool CUDAProfiler::printTracedRecords() {
```
k-ye, line 115: naming convention

taichi/backends/cuda/cuda_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+typedef enum CUDA_KERNEL_PROFILER {
+  CUDA_KERNEL_PROFILER_UNDEF = 0,
+  CUDA_KERNEL_PROFILER_EVENT = 1,
+  CUDA_KERNEL_PROFILER_CUPTI = 2
+} CUDAKernalProfiler;
+
+struct profilerConfig {
+  bool enable = false;
+  CUDAKernalProfiler profiler_type = CUDA_KERNEL_PROFILER_UNDEF;
+  KernelProfilerMode profiling_mode = KernelProfilerMode::disable;
```
ailzhang, line 22: IMHO we might want to clarify what's the relationship/difference between `enable` and `profiling_mode` here, e.g. if there's any invariants that we should be aware of .

taichi/backends/cuda/cuda_profiler.cpp
k-ye, line 19: why not just `!= KernelProfilerMode::disable`?

taichi/program/kernel_profiler.h
```diff
@@ -12,18 +12,43 @@
...
+
+KernelProfilerMode kernel_profiler_from_name(const std::string &mode);
+
 struct KernelProfileRecord {
   std::string name;
   int counter;
   double min;
   double max;
   double total;
+  double ldg;
```
ailzhang, line 32: I'd appreciate more descriptive names for ldg and stg :P

ailzhang (commented):
Overall looks good!

taichi/backends/cuda/cuda_context.h
k-ye: Can we have  different implementation for `KernelProfilerBase` ? E.g. `CudaEventBasedProfiler`, `CuptiBasedProfiler`, etc?

taichi/backends/cuda/cuda_profiler.cpp
yolo2themoon: I encapsulated this logic into `is_cuda_profiler()` and simplified init logic. thanks a lot :smile:

taichi/backends/cuda/cuda_profiler.cpp
yolo2themoon: thanks, the `enable` here is somewhat redundant. i have deleted it in commit e05ee990e753783bab8d57c1638e5d78893675f6

taichi/backends/cuda/cuda_profiler.cpp
yolo2themoon: This part of code is used for prototype testing, I have deleted it in commit e05ee990e753783bab8d57c1638e5d78893675f6. There will be a trace mode to print the profile result for each kernel.

taichi/backends/cuda/cuda_profiler.cpp
yolo2themoon: thanks, I encapsulated this logic into is_cuda_profiler() and simplified init logic in commit e05ee990e753783bab8d57c1638e5d78893675f6

taichi/backends/cuda/cuda_profiler.h
yolo2themoon: my bad :sob:  commit 291dd51f199f5f8ea704a8aae9e381f13fa327fd

taichi/program/kernel_profiler.h
yolo2themoon: my bad :sob:  commit 291dd51

taichi/backends/cuda/cuda_context.h
yolo2themoon: Removed from `CUDAContext`in commit 2eb03ed, CUPTI is now one of KernelProfilerCUDA's member variables, named `cupti_toolkit_`

taichi/backends/cuda/cuda_profiler.cpp
yolo2themoon: Solved in commit f4db404, thanks !

python/taichi/core/__init__.py
yolo2themoon: fixed in commit 2231dad

taichi/backends/cuda/cuda_profiler.h
yolo2themoon: I'll change variable name `num_ranges` in the next PR that uses CUPTI toolkit

PR #2849, opened 2021-08-30
[Vulkan] Use reference counting based wrapper layer

bobcao3:
Wraps the core vulkan object creation API in shared ptrs and provide references to keep track of object lifetime. This is a much cleaner approach than trying to track them in the RHI.

And renamed `vulkan_api.cpp` to `embedded_device.cpp`, so that the light resource wrapper can be called `vulkan_api.cpp`


bobcao3:
+ It might be useful to put this thin wrapper into another repo so we can use it in any other vulkan projects.

imgui.ini
```diff
@@ -0,0 +1,24 @@
+[Window][Debug##Default]
```
AmesingFlank, line 1: remove this lol. I apologize that GGUI keeps generating this... Maybe we could put it in a gitignore.

AmesingFlank:
Otherwise lgtm.

AmesingFlank (approved):

Issue #2772, opened 2021-08-23
Add `ti.any_arr` annotation to accept all kinds of dense arrays into kernels


strongoier:
**Concisely describe the proposed feature**
Previously, we have `ti.ext_arr` annotation to accept external arrays like numpy arrays and torch tensors into kernels. However, the support is limited - no struct fors are allowed, and they can only act as a Taichi scalar field. For user convenience as well as ease of AOT deployment, we would like to provide a more general and powerful annotation `ti.any_arr`, which accepts all kinds of dense arrays into kernels, including external arrays, user (device) buffers, Taichi ndarrays, and potentially Taichi dense fields. `ti.any_arr` will also support struct fors, and be able to act as Taichi vector/matrix fields.

**Describe the solution you'd like (if any)**

- [x] Add `ti.any_arr` annotation for external arrays and Taichi ndarrays. (#2703)
- [x] Add ability to act as Taichi vector/matrix fields. (#2727)
- [ ] Support struct fors.
- [ ] See whether Taichi dense fields can be handled together.

PR #2850, opened 2021-08-31
[refactor] Make ti.ext_arr a special case of ti.any_arr

strongoier:
Related issue = #2772

Now `ti.any_arr` annotation is added, and `ti.ext_arr` is for backward compatibility and indeed a special case of `ti.any_arr` with no support for element shape. In this PR, I would like to remove all the duplicated code and directly turn a `ti.ext_arr` into a `ti.any_arr` with only default arguments. This aims for better code maintainability.


k-ye (approved):
LGTM!

PR #2852, opened 2021-08-31
[GGUI] Visual & shader tweaks

bobcao3:
1. Fixed attenuation of point lights
2. Tone mapping & gamma correction for 3d scene
3. Replaced lambertian diffuse with oren nayer with a hard coded roughness


maajor (approved):
LGTM!

AmesingFlank:
I'm unsure about the point light attenuation:

If the user uses a big scene, is it possible that most of the light is attenuated away?

AmesingFlank (changes requested):
Let's remove point light attenuation for now... The api specified `point_light(pos,color)`, and I'd like for the users to not have to think about radiance.

python/taichi/shaders/Postprocessing_vk.vert
```diff
@@ -0,0 +1,14 @@
...
+vec2 positions[] = {
+  vec2(-1.0, 4.0),
+  vec2(-1.0, -1.0),
+  vec2(4.0, -1.0)
+};
+
+layout(location = 0) out vec2 uv;
+
+void main() {
+  uv = positions[gl_VertexIndex] * 0.5 + 0.5;
```
AmesingFlank, line 12: If you use `gl_VertexIndex`, it will be harder for us to move GGUI to OpenGL once we have OpenGLDevice. AFAIK opengl doesn't support `gl_VertexIndex`

bobcao3: In OpenGL there is gl_VertexID and it is fully equivalent, so we don't need to worry about that.

bobcao3:
> 
> 
> Let's remove point light attenuation for now :(( The api specified `point_light(pos,color)`, and I'd like for the users to not have to think about radiance.

Disagree. This is counter intuitive to the mental model of light where the further away the dimmer it is, and it degrades the sense of depth in the image greatly as again it's against the observation of lights in real world

bobcao3:
For a large scene the user either need to make a very strong point light or just use a directional light, or even bumping up the ambient lighting. Since this is the practice since before 2000s I don't think people will have a hard time dealing with attenuation

AmesingFlank:
> > Let's remove point light attenuation for now :(( The api specified `point_light(pos,color)`, and I'd like for the users to not have to think about radiance.
> 
> Disagree. This is counter intuitive to the mental model of light where the further away the dimmer it is, and it degrades the sense of depth in the image greatly as again it's against the observation of lights in real world

I understand that. But the visual improvements that this brings does not overweigh the additional 心智负担。The purpose of taichi's built-in UI is to be a preview tool, and should be as easy to use as possible.

bobcao3:
I stand by my view that this does not bring any burden to the user. there's
no point in using a point light if the user only wants something to show
up. We should just tell them to set the ambient light or use a "sun light".
A point light having a position implies there's a distance.

bobcao3:
Let's deal with compatibility first

PR #2855, opened 2021-09-01
[ci] M1 release

Leonz5288:
Related issue = #2254

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
+      matrix_osx: ${{ steps.set-matrix.outputs.matrix_osx }}
     steps:
       - id: set-matrix
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -81,6 +85,53 @@ jobs:
...
+      run:
+        shell: /bin/bash -l {0}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+
+      - name: Build
+        run: |
+          conda activate $PYTHON
```
ailzhang, line 104: if conda activate worked, the python executable should have been callable without adding to PATH?

.github/workflows/release.yml
```diff
@@ -81,6 +85,53 @@ jobs:
...
+          export PATH=/Users/github/miniforge3/envs/$PYTHON/bin:$PATH
+          arch -arm64 python3 -m pip uninstall taichi -y
```
ailzhang, line 106: No need to set arch -arm64 anymore, see #2845

ailzhang (commented):
Nice job! Thanks! Only a few nits in the comments.

.github/workflows/release.yml
Leonz5288: I tried the default trick, but it does not work for the bash I use...

.github/workflows/release.yml
```diff
@@ -81,6 +85,51 @@ jobs:
...
+      run:
+        shell: "/usr/bin/arch -arch arm64e /bin/bash --noprofile --norc -eo pipefail {0}"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+
+      - name: Build
+        run: |
+          export PATH=/Users/github/miniforge3/envs/$PYTHON/bin:$PATH
```
k-ye, line 104: Is this how conda is enabled 😅  ?

Leonz5288: Well... This is complicated... We can use conda activate $PYTHON. However, this won't change the PATH to python for some reason. Even if we activate conda env, we still need to manually set python path... So we can just get rid of the activate.

Leonz5288: BTW, the conda related github actions all requires setting PATH

ailzhang (approved):
Looks great! Thanks!

PR #2856, opened 2021-09-01
[Doc] Update the API reference section.

rexwangcc:
Related issue = https://github.com/taichi-dev/docs.taichi.graphics/issues/83

## Changes

- Remove redundant api references for clarity, also update the main API page with the correct link.

## Questions

- Shall we remove the existing md files?
- Does the wording of the notice make sense?

PR #2866, opened 2021-09-02
[doc] Remove links from documentation articles to API reference

rexwangcc:
Related issue = This is a follow up of #2856 

- We need to remove links from documentation articles to API reference so the documentation site can build without issues.
- This is more of an **interim** step, we will build smarter MDX components to help:
    - link from articles to API reference
    - link between API reference sections


ailzhang (approved):
Thanks!

PR #2869, opened 2021-09-02
[Lang] Support calling external function from bitcode

squarefk:
The original solution for calling external C++ function is based on .so and python ctypes. The function type support for ctypes is bad.


squarefk:
Wait. Some polishment is needed before merging.. Forget to

PR #2871, opened 2021-09-03
[Lang] Fix ti func with template and add corresponding tests

squarefk:
Related issue = #

According to https://docs.taichi.graphics/docs/lang/articles/basic/syntax#advanced-arguments-1, we can use `ti.template()` as type-hint to force arguments to be passed by reference to `ti.func`. However, current implementation doesn't match this description. This PR aims at fixing this mismatch.

Example:
```python
@ti.func
def my_func(x: ti.template()):
    x = x + 1  # will change the original value of x


@ti.kernel
def my_kernel():
    x = 233
    my_func(x)
    print(x)  # 234
```

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -164,7 +164,6 @@ class Func:
             _ti_core.make_func_call_expr(
                 self.taichi_functions[key.instance_id], non_template_args))
 
-    # TODO: Rename this to show only compiling func not kernel
     def do_compile(self, key, args):
         src = _remove_indent(oinspect.getsource(self.func))
         tree = ast.parse(src)
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -164,6 +164,7 @@ class Func:
             _ti_core.make_func_call_expr(
                 self.taichi_functions[key.instance_id], non_template_args))
 
+    # TODO: Rename this to show only compiling func not kernel
     def do_compile(self, key, args):
         src = _remove_indent(oinspect.getsource(self.func))
         tree = ast.parse(src)
--- a/tests/python/test_function.py
+++ b/tests/python/test_function.py
@@ -166,6 +166,24 @@ def test_default_templates():
     def func4(x):
         x += 1
 
+    @ti.func
+    def func1_field(x: ti.template()):
+        x[None] = 1
+
+    @ti.func
+    def func2_field(x: ti.template()):
+        x[None] += 1
+
+    @ti.func
+    def func3_field(x):
+        x[None] = 1
+
+    @ti.func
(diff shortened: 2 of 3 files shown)


strongoier:
Could you also add tests for passing a field into `ti.template()` under this modification?

tests/python/test_function.py
```diff
@@ -148,8 +148,73 @@ def run(self) -> ti.i32:
...
+    def func4_field(x):
+        x[None] += 1
+
+    v = ti.Vector.field(3, dtype=ti.i32, shape=())
+
+    @ti.kernel
+    def run_func():
+        a = 0
+        func1(a)
+        assert a == 1
```
strongoier, line 191: `assert` in `ti.kernel` only works in debug mode :-)

tests/python/test_function.py
```diff
@@ -148,8 +148,73 @@ def run(self) -> ti.i32:
...
+        x[None] = 1
+
+    @ti.func
+    def func4_field(x):
+        x[None] += 1
+
+    v = ti.Vector.field(3, dtype=ti.i32, shape=())
+
+    @ti.kernel
+    def run_func():
```
strongoier, line 188: You need to actually call `run_func()` :-)

tests/python/test_function.py
```diff
@@ -148,8 +148,73 @@ def run(self) -> ti.i32:
...
+        a = 0
+        func1(a)
+        assert a == 1
+        b = 0
+        func1(b)
```
strongoier, line 193:
```suggestion
        func2(b)
```

tests/python/test_function.py
```diff
@@ -148,8 +148,73 @@ def run(self) -> ti.i32:
...
+        assert b == 1
+        c = 0
+        func1(c)
```
strongoier, line 196:
```suggestion
        func3(c)
```

tests/python/test_function.py
```diff
@@ -148,8 +148,73 @@ def run(self) -> ti.i32:
...
+        assert c == 0
+        d = 0
+        func1(d)
```
strongoier, line 199:
```suggestion
        func4(d)
```

python/taichi/lang/kernel_impl.py
```diff
@@ -164,6 +164,7 @@ def func_call_rvalue(self, key, args):
             _ti_core.make_func_call_expr(
                 self.taichi_functions[key.instance_id], non_template_args))
 
+    # TODO: Rename this to show only compiling func not kernel
```
strongoier, line 167: But this method is inside `Func` class?

strongoier:
```suggestion
```

strongoier (approved):
LGTM!

Issue #2847, opened 2021-08-30
Re-enable test_remove_clear_list_from_fused_serial on CUDA backend


ailzhang:
This is a flaky test on GPU CI so we disabled it: https://pipelines.actions.githubusercontent.com/FbByQYkP8Upm7Ytd8H6oTOQALwY9F2xlXc4iGS6gaDqx1Yahr3/_apis/pipelines/1/runs/13359/signedlogcontent/23?urlExpires=2021-08-30T07%3A25%3A09.8417842Z&urlSigningMethod=HMACV1&urlSignature=oNiUWK9jPo83M1H1HahyJwHm8fbo7KlQhJmt8PffpNU%3D

qiao-bo:
If this is not urgent, I would like to have a look. Would you indicate which example is this? I cannot open the link.

ailzhang:
Thanks @qiao-bo ! Yea please feel free to take it!

qiao-bo:
Some findings:

- single thread: local test ok, CI test ok
- two threads: local test ok, CI test ok
- four threads: local test ok, CI test fails as follows
```
=========================== short test summary info ============================
FAILED ../../../../.local/lib/python3.8/site-packages/taichi/tests/test_torch_ad.py::test_torch_ad
FAILED ../../../../.local/lib/python3.8/site-packages/taichi/tests/test_torch_io.py::test_io_3d
FAILED ../../../../.local/lib/python3.8/site-packages/taichi/tests/test_while.py::test_while
FAILED ../../../../.local/lib/python3.8/site-packages/taichi/tests/test_torch_ad.py::test_torch_ad_gpu
FAILED ../../../../.local/lib/python3.8/site-packages/taichi/tests/test_torch_io.py::test_io_zeros
====== 5 failed, 879 passed, 2 skipped, 36 warnings in 809.56s (0:13:29) =======
``` 
@ailzhang Interestingly, the sfg test hasn't failed yet. comments are appreciated.

ailzhang:
This test is enabled back since it's not easy to reproduce. We'll keep an eye on it and reopen this issue if it happens again in the future.

PR #2874, opened 2021-09-06
[ci] Re-enable sfg test on CUDA

qiao-bo:
Related issue = #2847


qiao-bo:
This sfg test cannot be reproduced for now. Suggest to re-enable it until someone observes the failure again.

ailzhang (approved):
Thanks!

AmesingFlank:
I am observing failure in #3067 ... It is the only test that failed. I'm disabling it.

PR #2875, opened 2021-09-06
[Lang] Refine semantics of ti.any_arr

strongoier:
Related issue = #2772

Changes:
- For external arrays, we can treat it as a Taichi field with Vector or Matrix elements by specifying element **dim** (instead of **shape**) and layout.
- For Taichi vector/matrix ndarrays, we will **automatically** identify element dim and layout. If they are explicitly specified, we will check compatibility between the actual arguments and the annotation.

### DIFF
--- a/python/taichi/lang/kernel_arguments.py
+++ b/python/taichi/lang/kernel_arguments.py
@@ -64,17 +64,26 @@ class ArgAnyArray:
         layout = Layout.AOS if self.layout is None else self.layout
         shape = tuple(x.shape)
         if len(shape) < element_dim:
-            raise ValueError(f"Invalid argument into ti.any_arr() - required element_dim={element_dim}, but the argument has only {len(shape)} dimensions")
-        element_shape = () if element_dim == 0 else shape[:element_dim] if layout == Layout.SOA else shape[-element_dim:]
+            raise ValueError(
+                f"Invalid argument into ti.any_arr() - required element_dim={element_dim}, but the argument has only {len(shape)} dimensions"
+            )
+        element_shape = (
+        ) if element_dim == 0 else shape[:
+                                         element_dim] if layout == Layout.SOA else shape[
+                                             -element_dim:]
         return to_taichi_type(x.dtype), len(shape), element_shape, layout
 
     def check_element_dim(self, arg, arg_dim):
         if self.element_dim is not None and self.element_dim != arg_dim:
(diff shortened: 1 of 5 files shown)


python/taichi/lang/kernel_arguments.py
```diff
@@ -30,52 +30,51 @@ def ext_arr():
...
+        from taichi.lang.ndarray import ScalarNdarray
+        if isinstance(x, ScalarNdarray):
+            if self.element_dim is not None and self.element_dim != 0:
                 raise ValueError("Invalid argument passed to ti.any_arr()")
-            if isinstance(x, VectorNdarray) and (element_dim != 1 or
-                                                 self.element_shape[0] != x.n
-                                                 or self.layout != x.layout):
+            return x.dtype, len(x.shape), (), Layout.AOS
+        if isinstance(x, VectorNdarray):
+            if self.element_dim is not None and self.element_dim != 1:
```
ailzhang, line 56: nit: as our semantic here varies case by case, maybe it's helpful to put more hints in the error message? E.g `Invalid argument passed to ti.any_arr(). This is a VectorNdarray but got element_dim = {}`. Similar improvements can be applied to layout check as well. 
Given it's repeated three times for scalar/vector/matrix it might also worth adding a helper function to do element_dim check ;)

tests/python/test_ndarray.py
```diff
@@ -231,6 +231,9 @@ def func(a: ti.any_arr(element_shape=(10, ))):
     func(v)
     assert ti.get_runtime().get_num_compiled_functions() == 1
     import torch
-    v = torch.zeros((7, 10), dtype=torch.int32)
+    v = torch.zeros((6, 11), dtype=torch.int32)
```
ailzhang, line 234: Please feel free to ignore if we already have one ;) : it might also worth adding a test that triggers element_dim/layout `ValueError`s above in `ArgAnyArray` so that we make sure it's properly handled.

ailzhang (approved):
Looks great! Only a few minor suggestions.

python/taichi/lang/kernel_arguments.py
strongoier: Helpful suggestion! Done.

tests/python/test_ndarray.py
strongoier: Helpful suggestion! Done.

PR #2886, opened 2021-09-07
[Llvm] Increase the number of arguments allowed in a kernel

strongoier:
This PR aims at increasing the number of arguments allowed in a kernel in the LLVM backend (for user requirements). However, this is a temporary solution which may make the context a bit larger and not fully utilized. A better way for argument passing needs to be utilized in the future.


ailzhang (approved):
LTGM!

Issue #2546, opened 2021-07-19
[IR] Unify global pointers


xumingkuan:
**Concisely describe the proposed feature**
Related issue = #2193 
Currently, Taichi has 5 types of global pointers: `GlobalPtrStmt`, `ExternalPtrStmt`, `BlockLocalPtrStmt`, `ThreadLocalPtrStmt`, `GlobalTemporaryStmt`. Ad-hoc checking these classes can cause some confusions when more types is going to be added (see #2543 as an example).

**Describe the solution you'd like (if any)**
There are multiple solutions to unify them.
1. Unify all global pointers to a class `GlobalPtrStmt`, and use an enum class to represent the *domain* of the global pointer.
  - Pros: simple inheritance hierarchy; easily knowing each statement belongs to which domain
  - Cons: too many fields in the class, since different domains need different fields and we must store all of them (or extensively use `std::variant`)
2. Let all global pointer classes inherit from a class `GlobalPtrStmt` (rename the current `GlobalPtrStmt`), and put everything common to them into the base class
  - Pros: very convenient if we only want to treat a part of global pointers specifically while treating other ones together
  - Cons: complex inheritance hierarchy
3. Keep current classes and add a helper function `is_global_ptr(Stmt *)`
  - Pros: minimal code change
  - Cons: easy to forget to change the function when adding a new global pointer; hard to use if we want to treat all global pointers together

**Additional comments**

PR #2888, opened 2021-09-08
[bug] [opt] Disable putting pointers into global tmp buffer

### LINKED ISSUE #2884: Pointers between offloaded tasks are not handled properly
**Describe the bug**
Existence of one variable appearing in multiple offloaded tasks will introduce a global tmp var, which is a pointer to the actual value. However, if such variable itself is a pointer, the global tmp var needs to be a pointer to a pointer, which seems not supported now. This leads to failure to handle external pointers between offloaded tasks.

**To Reproduce**
```py
import taichi as ti
import numpy as np

ti.init(print_preprocessed=True, print_ir=True)

@ti.kernel
def test(a: ti.any_arr()):
    a[0] = 1
    for i in range(1, 10):
        a[i] = 1

x = np.zeros(10)
test(x)
```

IR before offloading:
```
kernel {
  <i32> $0 = const [0]
  <*f64> $1 = arg[0]
  <*f64> $2 = external_ptr <$1>, [$0]
  <f64> $3 = const [1.0]
  $4 : global store [$2 <- $3]
  <i32> $5 = const [1]
  <i32> $6 = const [10]
  $7 : for in range($5, $6) (vectorize 1) (bit_vectorize -1) block_dim=adaptive {
    <i32> $8 = loop $7 index 0
    <*f64> $9 = external_ptr <$1>, [$8]
    $10 : global store [$9 <- $3]
  }
}
```

IR after offloading:
```
kernel {
  $0 = offloaded  
  body {
    <i32> $1 = const [0]
    <*f64> $2 = arg[0]
    <*f64> $3 = global tmp var (offset = 0 B) // SHOULD BE <**f64>
    <f64> $4 = cast_value<f64> $2 // FAILURE point due to casting, but the root cause is stated above
    $5 : global store [$3 <- $4]
    <*f64> $6 = external_ptr <$2>, [$1]
    <f64> $7 = const [1.0]
    $8 : global store [$6 <- $7]
    <i32> $9 = const [1]
    <i32> $10 = const [10]
  }
  $11 = offloaded range_for(1, 10) grid_dim=0 block_dim=32 
  body {
    <i32> $12 = loop $11 index 0
    <*f64> $13 = global tmp var (offset = 0 B)
    <*f64> $14 = external_ptr <$13>, [$12]
    <f64> $15 = const [1.0]
    $16 : global store [$14 <- $15]
  }
}
```

strongoier:
Related issue = close #2884

Previously, `GlobalPtrStmt` is not put into global tmp buffer due to lack of support for pointers to pointers. Same thing should also happen to other pointers including `ExternalPtrStmt` and `ArgLoad` (if the arg is a pointer) so that bug #2884 can be resolved.

### DIFF
--- a/taichi/transforms/offload.cpp
+++ b/taichi/transforms/offload.cpp
@@ -557,7 +557,7 @@ class FixCrossOffloadReferences : public BasicStmtVisitor {
     if (local_to_global_offset.find(op) == local_to_global_offset.end()) {
       TI_ASSERT_INFO(op->is<ConstStmt>() || op->is<PtrOffsetStmt>() ||
                          op->is<GlobalTemporaryStmt>() ||
-                         op->is<ArgLoadStmt>() || op->is<ExternalPtrStmt>(),
+                         op->is<ExternalPtrStmt>() || (op->is<ArgLoadStmt>() && op->as<ArgLoadStmt>()->is_ptr),
                      "{} is not allowed here.", op->type());
       // For cases like ConstStmt
       auto copy = op->clone();
--- a/taichi/transforms/offload.cpp
+++ b/taichi/transforms/offload.cpp
@@ -333,7 +333,7 @@ class IdentifyValuesUsedInOtherOffloads : public BasicStmtVisitor {
       return;
     auto top_level_ptr = SquashPtrOffset::run(stmt);
     // We don't support storing a pointer for now.
-    if (top_level_ptr->is<GlobalPtrStmt>())
+    if (top_level_ptr->is<GlobalPtrStmt>() || stmt->is<ExternalPtrStmt>() || (stmt->is<ArgLoadStmt>() && stmt->as<ArgLoadStmt>()->is_ptr))
       return;
     // Not yet allocated
     if (local_to_global.find(top_level_ptr) == local_to_global.end()) {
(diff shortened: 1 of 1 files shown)


taichi/transforms/offload.cpp
```diff
@@ -555,7 +556,8 @@ class FixCrossOffloadReferences : public BasicStmtVisitor {
 
     if (local_to_global_offset.find(op) == local_to_global_offset.end()) {
       TI_ASSERT_INFO(op->is<ConstStmt>() || op->is<PtrOffsetStmt>() ||
-                         op->is<GlobalTemporaryStmt>(),
+                         op->is<GlobalTemporaryStmt>() ||
+                         op->is<ArgLoadStmt>() || op->is<ExternalPtrStmt>(),
```
k-ye, line 560: nit: Should this also check for `ArgLoadStmt::is_ptr`?

k-ye (approved):
LGTM!

taichi/transforms/offload.cpp
strongoier: Yes that would be more strict and accurate. Fixed.

taichi/transforms/offload.cpp
```diff
@@ -333,7 +333,8 @@ class IdentifyValuesUsedInOtherOffloads : public BasicStmtVisitor {
       return;
     auto top_level_ptr = SquashPtrOffset::run(stmt);
     // We don't support storing a pointer for now.
-    if (top_level_ptr->is<GlobalPtrStmt>())
+    if (top_level_ptr->is<GlobalPtrStmt>() || stmt->is<ExternalPtrStmt>() ||
```
ailzhang, line 336: nit: can we wrap these checks into a helper function so that we can have a single place to change the logic in the future? ;)

strongoier: Completely solving the problem of identifying pointers will be done in #2546. As @xumingkuan said, only adding a helper function might not be good enough.
Meanwhile, we will be able to support pointers here (no need of these ad-hoc checks) once we improve our type system.

PR #2890, opened 2021-09-08
[ci] Install vulkan sdk in Dockerfile

AmesingFlank:


ci/Dockerfile
```diff
@@ -34,11 +34,8 @@ RUN apt-get update && \
 
 
 # Install Vulkan
-WORKDIR /
-RUN wget -qO - https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo apt-key add -
-RUN sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.2.182-focal.list https://packages.lunarg.com/vulkan/1.2.182/lunarg-vulkan-1.2.182-focal.list
-RUN sudo apt update
-RUN sudo apt install vulkan-sdk
+COPY install_vulkan.py /tmp/
+RUN python3 /tmp/install_vulkan.py
```
ailzhang, line 38: Hmmm ideally the install logic will stay in this Dockerfile so that it's portable (e.g. you can just send a single Dockerfile without `install_vulkan.py` to others so that they can build from scrach :D Also the logic in the python file can be controlled by the `UBUNTU` argument passed in. But I'm definitely okay with this being an intermediate state. :D

AmesingFlank: Yes It should be possible to translate the logic in the python file into shell scripts, and include that script in a `RUN` block of the Dockerfile. I chose to do it this way because I find shell scripts kind of dirty and less readable (even more so when we write it in a Dockerfile, because we have to put a \ at the end of every line). I would be happy to do change this in the future when we encounter the need of only sending a single Dockerfile.

Issue #2893, opened 2021-09-08
Error occurs when trying to assign a value to a scalar member in a struct from python scope


JeffreyXiang:
**Describe the bug**
Error occurs when trying to assign a value to a scalar member in a struct. Namely members with type ti.i32, ti.f32 and so on.

**To Reproduce**
```py
import taichi as ti

Mytype = ti.types.struct(val=ti.i32)

a = Mytype(0)
a.val = 1
```

**Log/Screenshots**
```
[Taichi] version 0.7.31, llvm 10.0.0, commit fa68168a, linux, python 3.8.8
Traceback (most recent call last):
  File "/******/test.py", line 6, in <module>
    a.val = 1
  File "/******/anaconda3/lib/python3.8/site-packages/taichi/lang/util.py", line 207, in wrapped
    return func(*args, **kwargs)
  File "/******/anaconda3/lib/python3.8/site-packages/taichi/lang/struct.py", line 103, in setter
    self[key] = value
  File "/******/anaconda3/lib/python3.8/site-packages/taichi/lang/struct.py", line 79, in __setitem__
    self.entries[key].set_entries(value)
AttributeError: 'int' object has no attribute 'set_entries'
```

**Additional comments**
A potential fix is as follows
Original code in /python/taichi/lang/struct.py class Struct:
```py
def __setitem__(self, key, value):
    _taichi_skip_traceback = 1
    if isinstance(self.entries[key], SNodeHostAccess):
        self.entries[key].accessor.setter(value, *self.entries[key].key)
    else:
        if in_python_scope():
            self.entries[key].set_entries(value)
        else:
            self.entries[key] = value
```
Fixed code:
```py
def __setitem__(self, key, value):
    _taichi_skip_traceback = 1
    if isinstance(self.entries[key], SNodeHostAccess):
        self.entries[key].accessor.setter(value, *self.entries[key].key)
    else:
        if in_python_scope() and (isinstance(self.entries[key], Struct) or isinstance(self.entries[key], Matrix)):
            self.entries[key].set_entries(value)
        else:
            self.entries[key] = value
```

PR #2894, opened 2021-09-08
[bug] Fix the Bug that cannot assign a value to a scalar member in a struct from python scope

JeffreyXiang:
Related issue = #2893

### DIFF
--- a/python/taichi/lang/struct.py
+++ b/python/taichi/lang/struct.py
@@ -75,8 +75,16 @@ class Struct(TaichiOperations):
         if isinstance(self.entries[key], SNodeHostAccess):
             self.entries[key].accessor.setter(value, *self.entries[key].key)
         else:
-            if in_python_scope() and (isinstance(self.entries[key], Struct) or isinstance(self.entries[key], Matrix)):
-                self.entries[key].set_entries(value)
+            if in_python_scope():
+                if isinstance(self.entries[key], Struct) or isinstance(self.entries[key], Matrix):
+                    self.entries[key].set_entries(value)
+                else:
+                    if isinstance(value, (numbers.Number, expr.Expr)):
+                        self.entries[key] = value
+                    else:
+                        raise TypeError(
+                            "Number or Expr expected"
+                        )
             else:
                 self.entries[key] = value
 
--- a/python/taichi/lang/struct.py
+++ b/python/taichi/lang/struct.py
@@ -75,7 +75,7 @@ class Struct(TaichiOperations):
         if isinstance(self.entries[key], SNodeHostAccess):
             self.entries[key].accessor.setter(value, *self.entries[key].key)
         else:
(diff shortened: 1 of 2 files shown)


victoriacity (approved):

victoriacity:
Looks good! It would be great to also add a test case for this.

JeffreyXiang:
I found some other problems:
1. Vectors and Matrices can be assigned values of both types (int and float) whatever it is initialized to be in Python scope.
This can cause errors when users try to assign them to a Vector/Matrix field.

To Reproduce
```py
import taichi as ti

ti.init(arch=ti.cpu)

Vec3i = ti.types.vector(3, ti.i32)

a = Vec3i(0)
b = Vec3i.field(shape=())

print(a)
a[0] = 0.0
print(a)
b[None] = a
```

I think there should have strict type cast in python scope to avoid potential problems

2. In the new code I commit, Scalars can be assigned any Objects, because I directly use `self.entries[key] = value`. This is not as expected. I will fix this in the next commit.

python/taichi/lang/struct.py
```diff
@@ -76,7 +76,15 @@ def __setitem__(self, key, value):
             self.entries[key].accessor.setter(value, *self.entries[key].key)
         else:
             if in_python_scope():
-                self.entries[key].set_entries(value)
+                if isinstance(self.entries[key], Struct) or isinstance(self.entries[key], Matrix):
+                    self.entries[key].set_entries(value)
+                else:
+                    if isinstance(value, (numbers.Number, expr.Expr)):
```
victoriacity, line 82:
```suggestion
                    if isinstance(value, numbers.Number):
```
`expr.Expr` is used in Taichi scope. Since we already checked the method is called in Python scope, we can just verify whether it is a `Number` object.

python/taichi/lang/struct.py
```diff
@@ -76,7 +76,15 @@ def __setitem__(self, key, value):
...
+                        self.entries[key] = value
+                    else:
+                        raise TypeError(
+                            "Number or Expr expected"
```
victoriacity, line 86:
```suggestion
                           "A number is expected when assigning struct members"
```

python/taichi/lang/struct.py
JeffreyXiang: OK, I see. Thanks a lot!

victoriacity:
> I found some other problems:
> ...
> expected. I will fix this in the next commit.

Great catch! The compound types in `ti.types` module is actually more similar to a factory instance which produces matrices or structs with a predefined shape than strict custom typing like `typedef` in C++. This is due to the nature of Python being dynamically typed, and thus typing has not been enforced in python-scope local matrices and structs.

victoriacity (approved):

victoriacity:
Merged! Thank you so much for your contribution!

PR #2899, opened 2021-09-09
[CUDA] Add a compilation option for CUDA toolkit

yolo2themoon:
Related issue = #

Taichi loads the CUDA Driver functions from `libcuda.so` at runtime, instead of linking libraries during compilation.
Although this simplifies the configuration steps for using Taichi, the paths to CUDA headers and libraries are not introduced.

`backends/cuda/cuda_types.h` copys some of the definitions from `<cuda.h>` to avoid this problem.
Some profiling modules may depend on libraries in the CUDA toolkits directory.

**Describe the solution you'd like**
Add the `TI_WITH_CUDA_TOOKIT` compile option, set to OFF by default.
Get the CUDA toolkits directory from environment variable `CUDA_TOOLKIT_ROOT_DIR`.
Add header guards for codes in `cuda_types.h` to be compatible with both configurations ( with/without CUDA header ).


yolo2themoon:
I have tested all GGUI examples with : 
```
TAICHI_CMAKE_ARGS="-DTI_WITH_GGUI:BOOL=ON -DTI_WITH_CUDA_TOOLKIT:BOOL=OFF" python3 setup.py develop --user
TAICHI_CMAKE_ARGS="-DTI_WITH_GGUI:BOOL=ON -DTI_WITH_CUDA_TOOLKIT:BOOL=ON" python3 setup.py develop --user
```
looks good.

AmesingFlank:
> I have tested all GGUI examples with :
> 
> ```
> TAICHI_CMAKE_ARGS="-DTI_WITH_GGUI:BOOL=ON -DTI_WITH_CUDA_TOOLKIT:BOOL=OFF" python3 setup.py develop --user
> TAICHI_CMAKE_ARGS="-DTI_WITH_GGUI:BOOL=ON -DTI_WITH_CUDA_TOOLKIT:BOOL=ON" python3 setup.py develop --user
> ```
> 
> looks good.

Many Thanks!

AmesingFlank (approved):

cmake/TaichiCore.cmake
```diff
@@ -228,6 +229,17 @@ if (TI_WITH_CUDA)
     target_link_libraries(${LIBRARY_NAME} ${llvm_ptx_libs})
 endif()
 
+if (TI_WITH_CUDA_TOOLKIT)
+    message("TI_WITH_CUDA_TOOLKIT ON")
+    message("CUDA_TOOLKIT_ROOT_DIR=$ENV{CUDA_TOOLKIT_ROOT_DIR}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DTI_WITH_CUDA_TOOLKIT")
```
ailzhang, line 235: Maybe error out if `TI_WITH_CUDA_TOOLKIT` is on && `CUDA_TOOLKIT_ROOT_DIR` is not set?
Also there might be a better way to do it (although I haven't tested it :P ) https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#module:FindCUDAToolkit

yolo2themoon: I tried before this PR: put `FindCUDA.cmake` in the `cmake/utils` directory and `include(cmake/utils/FindCUDA.cmake)`  in `CMakeLists.txt` like TVM did, but it didn’t work then. ( the name `CUDA_TOOLKIT_ROOT_DIR` comes from cmake :P )

yolo2themoon: It's a great idea and i'll eventually work out a solution to this problem.

ailzhang (approved):
Thanks!

PR #2903, opened 2021-09-09
[vulkan] Support for multiple SNode trees in Vulkan

AmesingFlank:


taichi/backends/vulkan/runtime.cpp
```diff
@@ -442,26 +467,41 @@ class VkRuntime ::Impl {
...
+
     cmdlist->buffer_fill(global_tmps_buffer_->get_ptr(0), gtmp_buffer_size,
                          /*data=*/0);
     stream->submit_synced(cmdlist.get());
   }
 
-  const SNodeDescriptorsMap *const snode_descriptors_;
+  void add_root_buffer() {
+#pragma message("Vulkan buffers size hardcoded")
+    size_t root_buffer_size = 64 * 1024 * 1024;
```
bobcao3, line 478: After materializing the SNode tree, we should be able to get a size for that tree?

bobcao3: Since we are not handling sparse yet

bobcao3 (approved):
Super great work!

taichi/backends/vulkan/codegen_vulkan.cpp
```diff
@@ -70,9 +72,18 @@ class TaskCodegen : public IRVisitor {
     allow_undefined_visitor = true;
     invoke_default_visitor = true;
 
+    fill_node_to_root();
     ir_ = std::make_shared<spirv::IRBuilder>(params.device);
   }
 
+  void fill_node_to_root() {
```
k-ye, line 79: Nit: snode

taichi/backends/vulkan/codegen_vulkan.cpp
```diff
@@ -239,20 +254,22 @@ class TaskCodegen : public IRVisitor {
   void visit(SNodeLookupStmt *stmt) override {
     // TODO: SNodeLookupStmt -> GetSNodeCellStmt ?
     bool is_root{false};  // Eliminate first root snode access
+    int root_id = node_to_root_.at(stmt->snode->id);
```
k-ye, line 257: nit: const

taichi/backends/vulkan/codegen_vulkan.cpp
```diff
@@ -1224,16 +1245,19 @@ class TaskCodegen : public IRVisitor {
...
 
-    auto bind_buffer = [&](TaskAttributes::Buffers buffer) {
+    auto bind_buffer = [&](BufferInfo buffer) {
       result.push_back({buffer, binding});
       buffer_binding_map_[buffer] = binding++;
     };
 
-    bind_buffer(BuffersEnum::Root);
-    bind_buffer(BuffersEnum::GlobalTmps);
+    for (int root = 0; root < compiled_structs_.size(); ++root) {
```
k-ye, line 1253: BTW, if a kernel only uses the second root, will this still work? (I.e. do we still need to bind the first root?)

k-ye (approved):
LGTM!

taichi/backends/vulkan/codegen_vulkan.cpp
bobcao3: I think it's fine, as the descriptor layout is locked.

taichi/backends/vulkan/runtime.cpp
AmesingFlank: hmm yeah. lemme try.

taichi/backends/vulkan/codegen_vulkan.cpp
AmesingFlank: This will work, but there will indeed be an extra unused binding in that case.

PR #2904, opened 2021-09-10
[wasm] Fix WASM AOT module builder order

k-ye:
Related issue = #

### DIFF
--- a/taichi/program/program.cpp
+++ b/taichi/program/program.cpp
@@ -495,8 +495,8 @@ Program::~Program() {
 
 std::unique_ptr<AotModuleBuilder> Program::make_aot_module_builder(Arch arch) {
   // FIXME: This couples the runtime backend with the target AOT backend. E.g.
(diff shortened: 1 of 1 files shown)


taichi/program/program.cpp
```diff
@@ -494,11 +494,16 @@ Program::~Program() {
 }
 
 std::unique_ptr<AotModuleBuilder> Program::make_aot_module_builder(Arch arch) {
+  // FIXME: This couples the runtime backend with the target AOT backend. E.g.
+  // If we want to build a Metal AOT module, we have to be on the macOS
+  // platform. Consider decoupling this part
+  if (arch == Arch::wasm) {
+    // Have to check WASM first, or it dispatches to the LlvmProgramImpl.
+    return std::make_unique<wasm::AotModuleBuilderImpl>();
```
ailzhang, line 502: ohhh nice catch! yea I think the `arch_uses_llvm` below isn't accurate, we can probably also fix it later.

ailzhang (approved):
Thanks!

Issue #2804, opened 2021-08-26
When using a bit masked field in Metal, number of activated masks can be wrong


bsavery:
**Describe the bug**
On metal, 2D bit masks don't seem to activate all coordinates

**To Reproduce**
So I noticed this when trying to bit mask a 2D image.  Hopefully this example will be clear.  I start by activating the whole 2D field, then just check how many are active.  

```py
import taichi as ti

ti.init(arch=ti.gpu)

some_val = ti.field(dtype=float)
width, height = 10, 10
ptr = ti.root.bitmasked(ti.ij, (width, height))
ptr.place(some_val)
num_active = ti.field(dtype=int, shape=())

@ti.kernel
def init():
    num_active[None] = 0
    for x, y in ti.ndrange(width, height):
        ti.activate(ptr, [x, y])
        some_val[x, y] = 5
        num_active[None] += 1

@ti.kernel
def run():
    num_active[None] = 0
    for x, y in some_val:
        num_active[None] += 1
        

init()
print('num activated', num_active[None])
run()
print("active = ", num_active[None])
```

**Log/Screenshots**
5x5:
num activated 25
active =  16

6x6:
num activated 36
active =  28

10x10:
num activated 100
active =  64

**Additional comments**
CPU seems to do the right thing FYI.

k-ye:
Not sure if this is relevant: https://github.com/taichi-dev/taichi/issues/2260, but I'll just link here for future reference

bsavery:
I'm not sure.  I removed the "sibling field" in this case `num_active` but still got the same result.  Unless it's the same bug as #2260 but deeper.

k-ye:
Oops, short term solution is to use `ti.init(..., packed=True)`.

Long answer: By default Taichi pads a field to a power-of-two shape. So (10, 10) becomes (16, 16), so only the upper-left 10x10 region of the 16x16 is valid, meaning that the linearized indices are no longer consecutive. However, when iterating over a sparse field, we still used the linearized indices in [0, 100), so it checked for a few invalid places..

k-ye:
This should be fixed by the next release :-)

PR #2915, opened 2021-09-11
[metal] Fix upperbound for list-gen and struct-for

k-ye:
Related issue = #2804 

## Repro

I will use a smaller example compared to #2804 , so that the entire runtime information can be analyzed easily.

```py
import taichi as ti

ti.init(arch=ti.metal, print_ir=True)

some_val = ti.field(dtype=int)
width, height = 2, 3
ptr = ti.root.bitmasked(ti.ij, (width, height))
ptr.place(some_val)
num_active = ti.field(dtype=int, shape=())

@ti.kernel
def init():
    num_active[None] = 0
    for x, y in ti.ndrange(width, height):
        # ti.activate(ptr, [x, y])
        some_val[x, y] = x * 3 + y
        num_active[None] += 1

@ti.kernel
def run():
    num_active[None] = 0
    for x, y in some_val:
        print('x=', x, 'y=', y)
        num_active[None] += 1
        

init()
print('num activated', num_active[None])
run()
print("active = ", num_active[None])
```

1. `init() `
`init()` iterates over a 2D range of `(width, height)`. Internally, this gets converted into a linearized 1D range of `width x height = 6`.

Here's the corresponding Metal source code Taichi has generated (with simplication):

```cpp
void mtl_k0001_init_c4_0_1_func(const int linear_loop_idx_) {
  std::cout << "linear_loop_idx_=" << linear_loop_idx_ << '\n';
  const int tmp5 = linear_loop_idx_;
  constexpr int32_t tmp6 = 3;
  const int32_t tmp111 = (tmp5 / tmp6);
  constexpr int32_t tmp112 = 0;
  const int32_t tmp113 = -(tmp5 < tmp112);
  const int32_t tmp115 = (tmp111 * tmp6);
  const int32_t tmp116 = -(tmp113 != tmp112);
  const int32_t tmp117 = -(tmp5 != tmp112);
  const int32_t tmp118 = -(tmp115 != tmp5);
  const int32_t tmp119 = (tmp116 & tmp117);
  const int32_t tmp120 = (tmp119 & tmp118);
  const int32_t tmp121 = (tmp111 + tmp120);
  const int32_t tmp8 = (tmp121 * tmp6);
  const int32_t tmp9 = (tmp5 - tmp8);
  const int32_t tmp10 = (tmp8 + tmp9);
  // S0 tmp93(root_addr);
  // S0_ch tmp95 = tmp93.children(tmp112);
  // S1 tmp96 = tmp95.get0(runtime_, mem_alloc_);
  // std::cout << "  S0.children tmp112=" << tmp112 << std::endl;
  constexpr int32_t tmp124 = 1;
  const int32_t tmp125 = (tmp121 & tmp124);
  const int32_t tmp129 = (tmp9 & tmp6);
  constexpr int32_t tmp149 = 2;
  const int32_t tmp150 = (tmp125 << tmp149);
  const int32_t tmp146 = (tmp129 + tmp150);
  // tmp96.activate(tmp146);
  std::cout << "  activate tmp146=" << tmp146 << std::endl;
  // S1_ch tmp100 = tmp96.children(tmp146);
  // device int32_t *tmp101 = tmp100.get0(runtime_, mem_alloc_).val;
  // *tmp101 = tmp10;
  // const auto tmp80 = *tmp75;
  // const int32_t tmp81 = (tmp80 + tmp124);
  // *tmp75 = tmp81;
}
```

Ignoring the quirk math details, `linear_loop_idx_` iterates over `[0, 6)`, whereas `tmp146` is the linearized index of `[x, y]`.

Running the above code in any language, there is a discrepancy between these two indices.

| `linear_loop_idx_` | 0 | 1 | 2 | 3 | 4 | 5 |
|--------------------|---|---|---|---|---|---|
| `tmp146`           | 0 | 1 | 2 | 4 | 5 | 6 |

Note that there are "holes" in `tmp146`.

The reason behind this is that `some_val`'s internal shape is not `(2, 3)`. Rather, it is padded to the power-of-two, `(2, 4)`. This ensures that each axis takes a known number of bits. When accessing an element via a 2-D coordinate, each component is bit-shifted and bit-or'ed together to form the linearized index. In this example:

- `some_val` : externally it is a 2-D field of shape `(2, 3)`. Internally it is `(2, 4)`, so the first axis takes 1 bit, and the second 2 bits.
- 2-D coordinate: Taking `(1, 2)` as an example. The x-component 1 is left shifted by 2 bits, which is then bit-or'ed with the y-component 2. Therefore the linearized coordinate is `(1 << 2) | 2 = 6`. If not padded, this should have been `1 * 3 + 2 = 5`.

| 0: [0, 0] | 1: [0, 1] | 2: [0, 2] | 3: INVALID |
|-----------|-----------|-----------|------------|
| 4: [1, 0] | 5: [1, 1] | 6: [1, 2] | 7: INVALID |

2. `run()`
`run()` iterates over the field `some_val` itself. This is called a *struct-for*. Because `some_val` is a sparse field, Taichi will need to iterate over all the activated coordinates of the 2x3 region.

Here's a snippet of the generated Metal source code:

```cpp
  for (int ii = utid_; ii < 6; ii += ugrid_size_) {  // Note the 6
    const int parent_idx_ = (ii / child_num_slots);
    if (parent_idx_ >= parent_list.num_active()) break;
    const int child_idx_ = (ii % child_num_slots);
    const auto parent_elem_ = parent_list.get<ListgenElement>(parent_idx_);
    device auto *parent_addr_ =
        mtl_lgen_snode_addr(parent_elem_, root_addr, runtime_, mem_alloc_);
    if (!is_active(parent_addr_, parent_meta, child_idx_)) continue;
    ElementCoords elem_coords_;
    refine_coordinates(parent_elem_.coords, runtime_->snode_extractors[1],
                       child_idx_, &elem_coords_);
    mtl_k0003_run_c6_0_2_func(root_addr, global_tmps_addr, runtime_addr,
                              print_assert_addr, elem_coords_, tls_buffer_, ii);
  }
```

Unfortunately, this upperbound `6` is not correct. According to the above table, this only iterates over [0, 0], [0, 1], [0, 2], INVALID, [1, 0] and [1, 1]. The correct upperbound is `8` in this case, and `is_active()` knows to skip the INVALID holes as well as the inactivated coordinates.

We can make it even simpler, by not putting an upperbound limit at all. It works because `parent_idx_ >= parent_list.num_active()` serves as a natural upperbound check.

This upperbound was once necessary (see the discussion in https://github.com/taichi-dev/taichi/pull/986#discussion_r427712872). How it worked back then was beyond my memory.

### DIFF
--- a/tests/python/test_bitmasked.py
+++ b/tests/python/test_bitmasked.py
@@ -271,7 +271,6 @@ def test_bitmasked_2d_power_of_two():
     def init():
         num_active[None] = 0
         for x, y in ti.ndrange(width, height):
-            ti.activate(ptr, [x, y])
             some_val[x, y] = 5
             num_active[None] += 1
 
--- a/tests/python/test_bitmasked.py
+++ b/tests/python/test_bitmasked.py
@@ -260,28 +260,28 @@ def test_bitmasked_offset_child():
 
 @ti.test(require=ti.extension.sparse)
 def test_bitmasked_2d_power_of_two():
-  some_val = ti.field(dtype=float)
-  width, height = 10, 10
-  total = width * height
-  ptr = ti.root.bitmasked(ti.ij, (width, height))
-  ptr.place(some_val)
-  num_active = ti.field(dtype=int, shape=())
+    some_val = ti.field(dtype=float)
+    width, height = 10, 10
+    total = width * height
+    ptr = ti.root.bitmasked(ti.ij, (width, height))
+    ptr.place(some_val)
+    num_active = ti.field(dtype=int, shape=())
 
-  @ti.kernel
-  def init():
-      num_active[None] = 0
-      for x, y in ti.ndrange(width, height):
-          ti.activate(ptr, [x, y])
-          some_val[x, y] = 5
-          num_active[None] += 1
+    @ti.kernel
+    def init():
+        num_active[None] = 0
+        for x, y in ti.ndrange(width, height):
+            ti.activate(ptr, [x, y])
+            some_val[x, y] = 5
+            num_active[None] += 1
 
-  @ti.kernel
-  def run():
-      num_active[None] = 0
-      for x, y in some_val:
-          num_active[None] += 1
+    @ti.kernel
+    def run():
+        num_active[None] = 0
+        for x, y in some_val:
+            num_active[None] += 1
 
-  init()
-  assert num_active[None] == total
-  run()
-  assert num_active[None] == total
+    init()
+    assert num_active[None] == total
+    run()
+    assert num_active[None] == total
--- a/tests/python/test_bitmasked.py
+++ b/tests/python/test_bitmasked.py
@@ -256,3 +256,32 @@ def test_bitmasked_offset_child():
 
     func()
     assert s[None] == 7
+
+
+@ti.test(require=ti.extension.sparse)
+def test_bitmasked_2d_power_of_two():
+  some_val = ti.field(dtype=float)
+  width, height = 10, 10
+  total = width * height
+  ptr = ti.root.bitmasked(ti.ij, (width, height))
+  ptr.place(some_val)
+  num_active = ti.field(dtype=int, shape=())
+
+  @ti.kernel
+  def init():
+      num_active[None] = 0
+      for x, y in ti.ndrange(width, height):
+          ti.activate(ptr, [x, y])
+          some_val[x, y] = 5
+          num_active[None] += 1
+
+  @ti.kernel
+  def run():
+      num_active[None] = 0
+      for x, y in some_val:
+          num_active[None] += 1
+
+  init()
+  assert num_active[None] == total
+  run()
+  assert num_active[None] == total
--- a/taichi/backends/metal/codegen_metal.cpp
+++ b/taichi/backends/metal/codegen_metal.cpp
@@ -1181,7 +1181,8 @@ class KernelCodegenImpl : public IRVisitor {
     emit("const int child_num_slots = parent_meta.num_slots;");
     // Grid-stride loops:
     // Each thread begins at thread_index, and incremets by grid_size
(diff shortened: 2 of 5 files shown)


ailzhang (commented):
Thanks for the detailed issue summary and explanation! This is really helpful!

strongoier (commented):
Thanks for your clear explanation! I have a small question: if all of your description is correct, then is the function name `num_active()` in your snippet of the generated Metal source code a little misleading?

tests/python/test_bitmasked.py
```diff
@@ -256,3 +256,32 @@ def func():
...
+    total = width * height
+    ptr = ti.root.bitmasked(ti.ij, (width, height))
+    ptr.place(some_val)
+    num_active = ti.field(dtype=int, shape=())
+
+    @ti.kernel
+    def init():
+        num_active[None] = 0
+        for x, y in ti.ndrange(width, height):
+            ti.activate(ptr, [x, y])
```
strongoier, line 274: Is it necessary to manually call `activate()`?

ailzhang (approved):
sorry forget to approve in my previous review ;P

k-ye:
> Thanks for your clear explanation! I have a small question: if all of your description is correct, then is the function name `num_active()` in your snippet of the generated Metal source code a little misleading?

My description is the one that's misleading here... In this case we have one `bitmasked` container, which has 6 cells. So `num_active()` for the bitmasked container itself is always 1. If `bitmasked` has another non-place child SNode, then the list-gen process will figure out, out of the 6 cells, which of them are active

tests/python/test_bitmasked.py
k-ye:
```suggestion
```

strongoier (approved):
LGTM!

PR #2918, opened 2021-09-13
[Example] Add mgpcg and reflection ehanced stable fluid

erizmr:
Related issue = #
Update the current naive Jacobi based stable fluid with MGPCG and reflection, GGUI is also an option for visualization.

### DIFF
--- a/examples/simulation/mgpcg_stable_fluid.py
+++ b/examples/simulation/mgpcg_stable_fluid.py
@@ -38,7 +38,8 @@ class Dye:
 
     @ti.kernel
     def apply_impulse(self, vf: ti.template(), dyef: ti.template(),
-                      imp_data: ti.ext_arr(), dt: ti.template(), mean_res: float):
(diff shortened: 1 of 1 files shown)


k-ye:
Awesome! but this is probably too huge to serve as a basic example. It would be better to place this in https://github.com/taichi-dev/advanced_examples :-)

erizmr:
I think maybe I can try to reuse the `mgpcg_advanced.py` in the `examples/algorithm` to make a minimal version for this file (limit the number of lines comparable to the original `stable_fluid.py`). How does this sound to you? :)

k-ye:
> I think maybe I can try to reuse the `mgpcg_advanced.py` in the `examples/algorithm` to make a minimal version for this file (limit the number of lines comparable to the original `stable_fluid.py`). How does this sound to you? :)

We likely will move `mgpcg_advanced.py` to advanced_examples as well :rofl:

erizmr:
Good to know that.🤣 Then let's put them in the advanced_examples.

PR #2919, opened 2021-09-13
[ci] Enable tmate session if release test fails.

ailzhang:
Looks like our nightly release started to fail randomly, this PR automatically creates an ssh session **when it fails** so that we can debug from there.


k-ye (approved):
Thanks! Do we need a timeout :-) ?

Leonz5288 (approved):
lgtm

Issue #2736, opened 2021-08-20
Unified backend device API


k-ye:
Let's use this bug to track the unified device API works.

* [ ] LLVM
* [ ] Vulkan
* [ ] Metal
* [ ] OpenGL

bobcao3:
I think we should also add in a part where we refactors the program.cpp and kernel.cpp

PR #2925, opened 2021-09-13
[opengl] Initial Device API work

bobcao3:
Related Issue = https://github.com/taichi-dev/taichi/issues/2736

Added the initial GLDevice and related compute only implementation


taichi/backends/opengl/opengl_device.h
```diff
@@ -0,0 +1,246 @@
...
+    return binding_map_;
+  }
+
+ private:
+  std::unordered_map<uint32_t, GLuint> binding_map_;
+};
+
+class GLPipeline : public Pipeline {
+ public:
+  GLPipeline(const PipelineSourceDesc &desc, std::string name);
```
k-ye, line 64: nits: `const std::string &`

taichi/backends/opengl/opengl_device.h
```diff
@@ -0,0 +1,246 @@
...
+  struct Bindings : public ResourceBinder::Bindings {
+    // OpenGL has no sets, default set = 0
+    uint32_t binding{0};
+    GLuint buffer{0};
+    GLuint image{0};
+  };
+
+  std::unique_ptr<ResourceBinder::Bindings> materialize() override;
+
+  // In Vulkan this is called Storage Buffer (shader can store)
```
k-ye, line 25: These comments can be removed

taichi/backends/opengl/opengl_device.h
```diff
@@ -0,0 +1,246 @@
...
+  void run_commands();
+
+ private:
+  struct Cmd {
+    virtual void execute() {
+    }
+  };
+
+  struct CmdBindPipeline : public Cmd {
+    GLuint program;
```
k-ye, line 133: nit: Init these member variables?

taichi/backends/opengl/opengl_device.h
```diff
@@ -0,0 +1,246 @@
...
+class GLDevice : public GraphicsDevice {
+ public:
+  ~GLDevice() override;
+
+  DeviceAllocation allocate_memory(const AllocParams &params) override;
+  void dealloc_memory(DeviceAllocation handle) override;
+
+  std::unique_ptr<Pipeline> create_pipeline(
+      PipelineSourceDesc &src,
+      std::string name = "Pipeline") override;
```
k-ye, line 189: const ref

k-ye (commented):
Awesome! But would you break this into smaller PRs in the future?

taichi/backends/opengl/opengl_device.h
bobcao3: I thought since these are private member are only ever created through functions within this class, the overhead of initialization could be saved

ailzhang:
btw I self-requested a review to learn a bit more about device api, but you don't need to block on my review ;)

taichi/backends/opengl/opengl_device.h
k-ye: The initialization here is mostly to make the new code looks consistent :-) (And I don't think there will be any discernible difference here..)

taichi/backends/opengl/opengl_device.cpp
```diff
@@ -0,0 +1,425 @@
...
+  glBindBuffer(GL_SHADER_STORAGE_BUFFER, ptr.alloc_id);
+  glUnmapBuffer(GL_SHADER_STORAGE_BUFFER);
+}
+
+void GLDevice::unmap(DeviceAllocation alloc) {
+  glBindBuffer(GL_SHADER_STORAGE_BUFFER, alloc.alloc_id);
+  glUnmapBuffer(GL_SHADER_STORAGE_BUFFER);
+}
+
+void GLDevice::memcpy_internal(DevicePtr dst, DevicePtr src, uint64_t size) {
```
k-ye, line 304: nit: better assert `dist` and `src` have the same device ID.

k-ye (approved):
Great work!

PR #2926, opened 2021-09-14
[ci] Disable a few vulkan flaky tests.

ailzhang:
Our GPU CI is pretty flaky these days, I happen to see a few failed tests on my linux workstation which I think are pretty relevant. I'm going to disable these first to make sure CI is back to normal state. 
Screen Shot 2021-09-14 at 9 40 46 AM

### DIFF
--- a/tests/python/test_ad_basics.py
+++ b/tests/python/test_ad_basics.py
@@ -76,6 +76,7 @@ def test_poly():
 
 
 @if_has_autograd
+@ti.test(exclude=[ti.vulkan])
 def test_trigonometric():
     grad_test(lambda x: ti.tanh(x), lambda x: np.tanh(x))
     grad_test(lambda x: ti.sin(x), lambda x: np.sin(x))
--- a/tests/python/test_clear_all_gradients.py
+++ b/tests/python/test_clear_all_gradients.py
@@ -1,7 +1,7 @@
 import taichi as ti
 
 
-@ti.test()
+@ti.test(exclude=[ti.vulkan])
 def test_clear_all_gradients():
     x = ti.field(ti.f32)
     y = ti.field(ti.f32)
--- a/tests/python/test_element_wise.py
+++ b/tests/python/test_element_wise.py
@@ -11,7 +11,7 @@ def _c_mod(a, b):
 
 @pytest.mark.parametrize('lhs_is_mat,rhs_is_mat', [(True, True), (True, False),
                                                    (False, True)])
-@ti.test(fast_math=False)
+@ti.test(fast_math=False, exclude=[ti.vulkan])
 def test_binary_f(lhs_is_mat, rhs_is_mat):
     x = ti.Matrix.field(3, 2, ti.f32, 16)
     if lhs_is_mat:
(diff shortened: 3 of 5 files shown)


tests/python/test_ad_basics.py
```diff
@@ -76,6 +76,7 @@ def test_poly():
 
 
 @if_has_autograd
+@ti.test(exclude=[ti.vulkan])
 def test_trigonometric():
     grad_test(lambda x: ti.tanh(x), lambda x: np.tanh(x))
     grad_test(lambda x: ti.sin(x), lambda x: np.sin(x))
```
g1n0st, line 82: `ti.vulkan` seems not support autograd yet...

tests/python/test_clear_all_gradients.py
```diff
@@ -2,6 +2,7 @@
 
 
 @ti.test()
+@ti.test(exclude=[ti.vulkan])
```
g1n0st, line 5: ditto

tests/python/test_element_wise.py
```diff
@@ -243,6 +243,7 @@ def func():
 
 
 @ti.test()
+@ti.test(exclude=[ti.vulkan])
```
g1n0st, line 246: This might be the accuracy problem, rel=1e-5 probably not enough.

g1n0st (approved):
LGTM!

tests/python/test_ad_basics.py
ailzhang: This seems to be running fine on vulkan, except for an accuracy problem too :P 
`FAILED tests/python/test_ad_basics.py::test_trigonometric - assert 1.3344652652740479 == 1.3346064424467945 ± 1...`

g1n0st: But it's also need to note that the accuracy in Vulkan backend seems lower than other backends...

tests/python/test_clear_all_gradients.py
```diff
@@ -2,6 +2,7 @@
 
 
 @ti.test()
```
strongoier, line 4:
```suggestion
```
You don't have to repeat `ti.test()` :-)

tests/python/test_element_wise.py
```diff
@@ -243,6 +243,7 @@ def func():
 
 
 @ti.test()
```
strongoier, line 245:
```suggestion
```
ditto

tests/python/test_mpm88.py
```diff
@@ -127,6 +127,7 @@ def test_mpm88_async():
 
 
 @ti.test()
```
strongoier, line 129:
```suggestion
```
ditto

tests/python/test_struct.py
```diff
@@ -63,6 +63,7 @@ def test_linear_nested_aos():
 
 
 @ti.test()
```
strongoier, line 65:
```suggestion
```
ditto

tests/python/test_clear_all_gradients.py
ailzhang: oops nice catch! thanks!

PR #2929, opened 2021-09-14
[llvm] Remove duplicated set dim attribute for GlobalVariableExpression in LLVM

ailzhang:
Looks like this is already done in
https://github.com/taichi-dev/taichi/blob/master/taichi/program/snode_expr_utils.cpp#L69
for all backends. So `materialize_snode_expr_attributes` seems
redundant.

### DIFF
--- a/taichi/llvm/llvm_program.cpp
+++ b/taichi/llvm/llvm_program.cpp
@@ -193,7 +193,6 @@ void LlvmProgramImpl::materialize_snode_tree(
     SNodeTree *tree,
     std::vector<std::unique_ptr<SNodeTree>> &snode_trees_,
     std::unordered_map<int, SNode *> &snodes,
-    SNodeGlobalVarExprMap &snode_to_glb_var_exprs_,
     uint64 *result_buffer) {
   auto *const root = tree->root();
   auto host_module = clone_struct_compiler_initial_context(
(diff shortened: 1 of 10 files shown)


taichi/llvm/llvm_program.cpp
```diff
@@ -221,12 +219,6 @@ void LlvmProgramImpl::materialize_snode_tree(
   }
 }
 
-void LlvmProgramImpl::materialize_snode_expr_attributes(
-    SNodeGlobalVarExprMap &snode_to_glb_var_exprs_) {
-  for (auto &[snode, glb_var] : snode_to_glb_var_exprs_) {
```
k-ye, line 226: Hmm, we don't need to do this at all now..?

strongoier: As @ailzhang says, this is already done in
https://github.com/taichi-dev/taichi/blob/c0a5ebe070262cd691c6b71dd99127ac0bcea409/taichi/program/snode_expr_utils.cpp#L69
which is
https://github.com/taichi-dev/taichi/blob/c0a5ebe070262cd691c6b71dd99127ac0bcea409/taichi/ir/frontend_ir.h#L437-L440
Therefore it is redundant code.

strongoier (approved):
LGTM!

PR #2931, opened 2021-09-14
[example] 3D fratal example

AmesingFlank:
This is the code for this example:

 https://www.bilibili.com/video/BV1pf4y1J7rM?spm_id_from=333.999.0.0


k-ye (approved):
Very cool!

PR #2939, opened 2021-09-15
[refactor] Remove Program::KernelProxy

### LINKED ISSUE #2723: Remove Program::KernelProxy
`KernelProxy` is defined around:

https://github.com/taichi-dev/taichi/blob/63140f6fa56322c5af606662a60f48cb50acea2b/taichi/program/program.h#L194-L202

It complicates the readability without adding much value

https://github.com/taichi-dev/taichi/blob/63140f6fa56322c5af606662a60f48cb50acea2b/python/taichi/lang/kernel_impl.py#L434-L452

* L434: Creates a `KernelProxy` using https://github.com/taichi-dev/taichi/blob/63140f6fa56322c5af606662a60f48cb50acea2b/taichi/python/export_lang.cpp#L762-L765
* L452: Creates an actual `Kernel` using https://github.com/taichi-dev/taichi/blob/63140f6fa56322c5af606662a60f48cb50acea2b/taichi/program/program.h#L199-L201

Just pass `taichi_ast_generator` to `Kernel`'s constructor directly. We don't have to use KernelProxy.

lin-hitonami:
Related issue = close #2723

### DIFF
--- a/taichi/python/export_lang.cpp
+++ b/taichi/python/export_lang.cpp
@@ -424,15 +424,6 @@ void export_lang(py::module &m) {
       .def("serialize", &ExprGroup::serialize);
 
   py::class_<Stmt>(m, "Stmt");
-  py::class_<Program::KernelProxy>(m, "KernelProxy")
-      .def(
-          "define",
-          [](Program::KernelProxy *ker,
-             const std::function<void()> &func) -> Kernel * {
-            py::gil_scoped_release release;
-            return ker->def(func);
-          },
-          py::return_value_policy::reference);
 
   m.def("insert_deactivate", [](SNode *snode, const ExprGroup &indices) {
     return Deactivate(snode, indices);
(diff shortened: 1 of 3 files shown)


taichi/python/export_lang.cpp
```diff
@@ -772,10 +763,13 @@ void export_lang(py::module &m) {
   m.def("get_external_tensor_shape_along_axis",
         Expr::make<ExternalTensorShapeAlongAxisExpression, const Expr &, int>);
 
-  m.def("create_kernel",
-        [&](std::string name, bool grad) -> Program::KernelProxy {
-          return get_current_program().kernel(name, grad);
-        });
+  m.def(
+      "create_kernel",
+      [&](const std::function<void()> &body, const std::string &name = "",
```
ailzhang, line 768: nit: kernel_name was a required argument before, is there any reason that we want to make it optional here? ;)

ailzhang (approved):
Overall looks great! Thanks!

taichi/python/export_lang.cpp
lin-hitonami: fixed

taichi/python/export_lang.cpp
```diff
@@ -424,15 +424,6 @@ void export_lang(py::module &m) {
       .def("serialize", &ExprGroup::serialize);
 
   py::class_<Stmt>(m, "Stmt");
-  py::class_<Program::KernelProxy>(m, "KernelProxy")
-      .def(
-          "define",
-          [](Program::KernelProxy *ker,
-             const std::function<void()> &func) -> Kernel * {
-            py::gil_scoped_release release;
```
ailzhang, line 432: @lin-hitonami This `gil_scoped_release` was accidentally removed in the refactor? Would you mind adding it back? 
Thanks @strongoier for catching this!

PR #2943, opened 2021-09-15
[misc] Throw proper error if real function is not annotated.

### LINKED ISSUE #2938: Confusing error if I forget to add type annotation in real function support
Repro: 
```
import taichi as ti

ti.init(ti.cpu, print_ir=False, print_preprocessed=True, experimental_real_function=True)

@ti.func
def add(a, b: ti.i32) -> ti.i32:
    return a + b

@ti.kernel
def run():
    print(add(40, 2))


run()
```
In this case I forget to add a type annotation for `a` and the error message is 
```
  File "repro.py", line 16, in <module>
    run()
  File "/Users/ailzhang/github/taichi/python/taichi/lang/kernel_impl.py", line 681, in wrapped
  File "/Users/ailzhang/github/taichi/python/taichi/lang/kernel_impl.py", line 608, in __call__
    key = self.ensure_compiled(*args)
  File "/Users/ailzhang/github/taichi/python/taichi/lang/kernel_impl.py", line 599, in ensure_compiled
  File "/Users/ailzhang/github/taichi/python/taichi/lang/kernel_impl.py", line 465, in materialize
  File "/Users/ailzhang/github/taichi/python/taichi/lang/kernel_impl.py", line 460, in taichi_ast_generator
    compiled()
  File "repro.py", line 13, in run
    print(add(40, 2))
  File "/Users/ailzhang/github/taichi/python/taichi/lang/kernel_impl.py", line 72, in decorated
    return fun.__call__(*args)
  File "/Users/ailzhang/github/taichi/python/taichi/lang/kernel_impl.py", line 147, in __call__
    self.do_compile(key=key, args=args)
  File "/Users/ailzhang/github/taichi/python/taichi/lang/kernel_impl.py", line 193, in do_compile
    compile(tree, filename=oinspect.getsourcefile(self.func), mode='exec')
ValueError: None disallowed in expression list
```

ailzhang:
fixes #2938

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -226,6 +226,13 @@ class Func:
             if annotation is inspect.Parameter.empty:
                 if i == 0 and self.classfunc:
                     annotation = template()
+                # TODO: pyfunc also need type annotation check when real function is enabled,
+                #       but that has to happen at runtime when we know which scope it's called from.
+                elif not self.pyfunc and impl.get_runtime(
+                ).experimental_real_function:
+                    raise KernelDefError(
+                        f'Taichi function `{self.func.__name__}` parameter `{arg_name}` must be type annotated'
+                    )
             else:
                 if not id(annotation
                           ) in primitive_types.type_ids and not isinstance(
--- a/python/taichi/tools/patterns.py
+++ b/python/taichi/tools/patterns.py
@@ -2,12 +2,7 @@ import taichi as ti
 
 
 @ti.func
-def _inside(p, c, r):
-    return (p - c).norm_sqr() <= r * r
-
-
-@ti.func
-def taichi_logo(pos, scale=1 / 1.11):
+def taichi_logo(pos: ti.template(), scale: float = 1 / 1.11):
     p = (pos - 0.5) / scale + 0.5
     ret = -1
     if not (p - 0.50).norm_sqr() <= 0.52**2:
--- a/tests/python/test_function.py
+++ b/tests/python/test_function.py
@@ -1,3 +1,5 @@
+import pytest
+
 import taichi as ti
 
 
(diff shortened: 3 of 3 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -226,6 +226,14 @@ def extract_arguments(self):
             if annotation is inspect.Parameter.empty:
                 if i == 0 and self.classfunc:
                     annotation = template()
+                # TODO: pyfunc also need type annotation check when real function is enabled,
+                #       but that has to happen at runtime when we know which scope it's called from.
+                elif not self.pyfunc and impl.get_runtime(
+                ).experimental_real_function:
+                    _taichi_skip_traceback = 1
```
k-ye, line 233: Context: we wish to remove this

```suggestion
```

python/taichi/lang/kernel_impl.py
```diff
@@ -226,6 +226,14 @@ def extract_arguments(self):
...
+                    raise KernelDefError(
```
k-ye, line 234: Ah, how about we use `TaichiSyntaxError` for consistency?

ailzhang: Are we going to keep both `TaichiSyntaxError` and `KernelDefError` in the future? Asking since all the surrounding code uses `KernelDefError` :P

k-ye: OK.. Then maybe we should change the missing return type to `KernelDefError` as well :-)

ailzhang: haha let me show you this, you'll be amazed. ;) 
the missing return type is checked at runtime (where we use a lot  of `TaichiSyntaxError`): https://github.com/taichi-dev/taichi/blob/master/python/taichi/lang/stmt_builder.py#L631
while the missing annotation is checked at compile time (where we use a lot of `KernelDefError`) ;) https://github.com/taichi-dev/taichi/blob/master/python/taichi/lang/kernel_impl.py#L118 
For example, if you look at the tests I added below, you can trigger the error for missing input arg annotation by defining a ti.func. But triggering the error for missing return annotation have to be done by calling the ti.func :D

ailzhang: We might need a bit discussion to decide which one (or two) errors to unify, maybe I just keep it locally consistent for now? ;)

k-ye (approved):
Thx!

PR #2951, opened 2021-09-16
[perf] [CUDA] Add missed AtomicOpType::min&max of warp reduction

yolo2themoon:
Related issue = #2487

This commit solves the problem of AtomicOpType::min & max not being added to the optimization list when warp reduction is performed on `ti.field` data.

But the one on `ti.ndarray` still unfixed.

### DIFF
--- a/taichi/transforms/make_thread_local.cpp
+++ b/taichi/transforms/make_thread_local.cpp
@@ -14,6 +14,7 @@ TLANG_NAMESPACE_BEGIN
 
 namespace {
 
+// TODO : bit_and, bit_or, bit_xor
 bool is_atomic_op_linear(AtomicOpType op_type) {
   return op_type == AtomicOpType::add || op_type == AtomicOpType::sub ||
          op_type == AtomicOpType::min || op_type == AtomicOpType::max;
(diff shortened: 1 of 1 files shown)


yolo2themoon:
FAILED ../../../../Library/Python/3.8/lib/python/site-packages/taichi/tests/test_atomic.py::test_atomic_max_expr_evaled

taichi/transforms/make_thread_local.cpp
```diff
@@ -14,8 +14,10 @@ TLANG_NAMESPACE_BEGIN
 
 namespace {
 
+// TODO : bit_and, bit_or, bit_xor
 bool is_atomic_op_linear(AtomicOpType op_type) {
-  return op_type == AtomicOpType::add || op_type == AtomicOpType::sub;
+  return op_type == AtomicOpType::add || op_type == AtomicOpType::sub ||
```
strongoier, line 19:
```suggestion
  return op_type == AtomicOpType::add || 
```

strongoier (commented):
https://github.com/taichi-dev/taichi/blob/1a11f762a1b888eb774c099a972a4ddc25c6a425/taichi/transforms/make_thread_local.cpp#L170-L173

As we can see, `AtomicOpType::add` is hard-coded here so only add can work. We need to check whether only one op is applied to the destination and use that op here.

PR #2957, opened 2021-09-18
[gui] Move all ggui kernels to python by using taichi fields as staging buffers

AmesingFlank:
This PR moves all ggui kernels to python by using taichi fields as staging buffers. This removes the need for `ui_kernels.cpp/.ll/.bc` and allows the logic for future ggui backends to be less graphics-api-specific. 

On the C++ side, no more kernels needs to launched, only `memcpy` and `cuda_memcpy` are needed. In the future, these should be replaced by `device.memcpy()`.

The PR also fixed the bug where `set_image` behaves incorrectly when packed mode is enabled.


bobcao3:
In theory, going from here, for backends without known interop or without an implemented interop, we can direct the taichi kernel outputs to an extarr and use host device copy like the old GUI, right?

bobcao3 (approved):
Clean and clear, looks great!

AmesingFlank:
> In theory, going from here, for backends without known interop or without an implemented interop, we can direct the taichi kernel outputs to an extarr and use host device copy like the old GUI, right?

Yep. It should be quite straightforward to implement  [user input] -> [staging buffer] -> [np array] -> [actual gpu buffers].

k-ye (commented):
Thank you! Random nit: could you help add the docstrings for the public APIs like `Canvas`, `Scene` and `Window`? :-)

PR #2963, opened 2021-09-22
[cuda] Cuda Device API 1/n: memory allocation

AmesingFlank:
This PR begins the work on CUDA deivce API. 

The PR only implements memory allocation and deallocation, and makes use of those methods in `LlvmProgram` and `MemoryPool`. 

This would be very useful for unifying the `memcpy` logic used in GGUI, which would facilitate implementing GGUI on other backends.

### DIFF
--- a/taichi/backends/cuda/cuda_device.cpp
+++ b/taichi/backends/cuda/cuda_device.cpp
@@ -2,11 +2,9 @@
 
 TLANG_NAMESPACE_BEGIN
 
+namespace cuda {
 
-namespace cuda{
-
-
-CudaDevice::AllocInfo  CudaDevice::get_alloc_info(DeviceAllocation handle){
+CudaDevice::AllocInfo CudaDevice::get_alloc_info(DeviceAllocation handle) {
   validate_device_alloc(handle);
   return allocations_[handle.alloc_id];
 }
@@ -14,16 +12,15 @@ CudaDevice::AllocInfo  CudaDevice::get_alloc_info(DeviceAllocation handle){
 DeviceAllocation CudaDevice::allocate_memory(const AllocParams &params) {
   AllocInfo info;
 
-  if(params.host_read || params.host_write){
+  if (params.host_read || params.host_write) {
     CUDADriver::get_instance().malloc_managed(&info.ptr, params.size,
                                               CU_MEM_ATTACH_GLOBAL);
-  }
-  else{
+  } else {
     CUDADriver::get_instance().malloc(&info.ptr, params.size);
   }
-  
+
   info.size = params.size;
-  
+
   DeviceAllocation alloc;
   alloc.alloc_id = allocations_.size();
   alloc.device = this;
@@ -34,13 +31,13 @@ DeviceAllocation CudaDevice::allocate_memory(const AllocParams &params) {
 
 void CudaDevice::dealloc_memory(DeviceAllocation handle) {
   validate_device_alloc(handle);
-  AllocInfo& info = allocations_[handle.alloc_id];
-  if(info.ptr == nullptr){
+  AllocInfo &info = allocations_[handle.alloc_id];
+  if (info.ptr == nullptr) {
     TI_ERROR("the DeviceAllocation is already deallocated");
   }
   CUDADriver::get_instance().mem_free(info.ptr);
   info.ptr = nullptr;
 }
 
-}
+}  // namespace cuda
 TLANG_NAMESPACE_END
--- a/taichi/backends/cuda/cuda_device.cpp
+++ b/taichi/backends/cuda/cuda_device.cpp
@@ -13,8 +13,15 @@ CudaDevice::AllocInfo  CudaDevice::get_alloc_info(DeviceAllocation handle){
 
 DeviceAllocation CudaDevice::allocate_memory(const AllocParams &params) {
   AllocInfo info;
-  //TODO: unified memory for host read/write? Need to query CUDA capabilities.
-  CUDADriver::get_instance().malloc(&info.ptr, params.size);
+
+  if(params.host_read || params.host_write){
+    CUDADriver::get_instance().malloc_managed(&info.ptr, params.size,
+                                              CU_MEM_ATTACH_GLOBAL);
+  }
+  else{
+    CUDADriver::get_instance().malloc(&info.ptr, params.size);
+  }
+  
   info.size = params.size;
   
   DeviceAllocation alloc;
--- a/taichi/backends/cuda/cuda_device.cpp
+++ b/taichi/backends/cuda/cuda_device.cpp
@@ -1,0 +1,39 @@
+#include "taichi/backends/cuda/cuda_device.h"
+
+TLANG_NAMESPACE_BEGIN
+
+
+namespace cuda{
+
+
+CudaDevice::AllocInfo  CudaDevice::get_alloc_info(DeviceAllocation handle){
+  validate_device_alloc(handle);
+  return allocations_[handle.alloc_id];
+}
+
+DeviceAllocation CudaDevice::allocate_memory(const AllocParams &params) {
+  AllocInfo info;
+  //TODO: unified memory for host read/write? Need to query CUDA capabilities.
+  CUDADriver::get_instance().malloc(&info.ptr, params.size);
+  info.size = params.size;
+  
+  DeviceAllocation alloc;
+  alloc.alloc_id = allocations_.size();
+  alloc.device = this;
+
+  allocations_.push_back(info);
+  return alloc;
+}
+
+void CudaDevice::dealloc_memory(DeviceAllocation handle) {
+  validate_device_alloc(handle);
+  AllocInfo& info = allocations_[handle.alloc_id];
+  if(info.ptr == nullptr){
+    TI_ERROR("the DeviceAllocation is already deallocated");
+  }
+  CUDADriver::get_instance().mem_free(info.ptr);
+  info.ptr = nullptr;
+}
+
+}
+TLANG_NAMESPACE_END
--- a/taichi/backends/cuda/cuda_device.cpp
+++ b/taichi/backends/cuda/cuda_device.cpp
@@ -1,6 +1,7 @@
 #include "taichi/backends/cuda/cuda_device.h"
 
-TLANG_NAMESPACE_BEGIN
+namespace taichi {
+namespace lang {
 
 namespace cuda {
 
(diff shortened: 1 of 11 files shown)


taichi/backends/cuda/cuda_device.h
```diff
@@ -0,0 +1,124 @@
...
+#include "taichi/backends/cuda/cuda_context.h"
+#include "taichi/backends/device.h"
+
+TLANG_NAMESPACE_BEGIN
+
+namespace cuda {
+
+class CudaResourceBinder : public ResourceBinder {
+ public:
+  virtual ~CudaResourceBinder() override {
```
bobcao3, line 16: nit: mark implementation of virtual functions with override and remove that virtual in the beggining

taichi/backends/cuda/cuda_device.cpp
```diff
@@ -0,0 +1,43 @@
...
+TLANG_NAMESPACE_BEGIN
+
+namespace cuda {
+
+CudaDevice::AllocInfo CudaDevice::get_alloc_info(DeviceAllocation handle) {
+  validate_device_alloc(handle);
+  return allocations_[handle.alloc_id];
+}
+
+DeviceAllocation CudaDevice::allocate_memory(const AllocParams &params) {
```
k-ye, line 12: Note that there is another memory allocation mechanism in Taichi, where you pre-allocate a huge chunk of memory and use it as the pool. We should think about how to better unify these two approaches.

taichi/backends/cuda/cuda_device.h
```diff
@@ -0,0 +1,117 @@
+#pragma once
+#include <vector>
+#include <set>
+
+#include "taichi/common/core.h"
+#include "taichi/backends/cuda/cuda_driver.h"
+#include "taichi/backends/cuda/cuda_context.h"
+#include "taichi/backends/device.h"
+
+TLANG_NAMESPACE_BEGIN
```
k-ye, line 10: nit: let's stop using macros

taichi/backends/cuda/cuda_device.cpp
AmesingFlank: I agree. 

Currently, the `CudaDevice::allocate_memory` is used by the memory pool when it requires memory from the cuda driver. This is the exact opposite as our vulkan device, whose `VulkanDeivce::allocate_memory` uses an underlying memory pool.

AmesingFlank: How about we move the cuda memory pool inside the device api, so that Device APIs `allocate_memory` can always be used as-is?

bobcao3: https://github.com/NVIDIA/cnmem Looks interesting. There isn't a VulkanMemoryAllocator equivalent in CUDA i think. (There is an DX12MemoryAllocator lol)

taichi/backends/cuda/cuda_device.h
AmesingFlank: Let's have @taichi-gardener submit a PR that removes all of them?

taichi/program/program_impl.h
```diff
@@ -55,6 +56,14 @@ class ProgramImpl {
    */
   virtual std::unique_ptr<AotModuleBuilder> make_aot_module_builder() = 0;
 
+  virtual Device *get_compute_device() {
+    return nullptr;
+  }
+
```
qiao-bo, line 62: Just for my own naive understanding ;), these two functions are not pure virtual because they don't have to be implemented or?

AmesingFlank: I didn't make them pure virtual because for now returning a `nullptr` is a reasonable default behaviour. Currently, not every backend has a `Device` in their `ProgramImpl`.

taichi/backends/cuda/cuda_device.h
k-ye: Sounds good!

taichi/backends/cuda/cuda_device.cpp
k-ye: > How about we move the cuda memory pool inside the device api in the future, so that Device APIs allocate_memory can always be used as-is?

SG, right now https://github.com/taichi-dev/taichi/blob/e5914a458776bb921dd95cb9a00126dafe675b7a/taichi/system/snode_tree_buffer_manager.cpp#L43-L48 kind of achieves what we want. (It invokes the allocation function that is JIT compiled from LLVM, which makes things more complicated than necessary.) FYI @qiao-bo

taichi/llvm/llvm_program.cpp
```diff
@@ -92,6 +92,7 @@ LlvmProgramImpl::LlvmProgramImpl(CompileConfig &config_,
       CUDAContext::get_instance().set_profiler(nullptr);
     }
     CUDAContext::get_instance().set_debug(config->debug);
+    device_ = std::unique_ptr<Device>(new cuda::CudaDevice());
```
ailzhang, line 95: use make_unique instead?

taichi/system/memory_pool.cpp
```diff
@@ -1,10 +1,12 @@
 #include "memory_pool.h"
 #include "taichi/system/timer.h"
 #include "taichi/backends/cuda/cuda_driver.h"
+#include "taichi/backends/cuda/cuda_device.h"
 
 TLANG_NAMESPACE_BEGIN
 
-MemoryPool::MemoryPool(Arch arch) : arch_(arch) {
+MemoryPool::MemoryPool(Arch arch, Device *device)
+    : arch_(arch), device_(device) {
```
ailzhang, line 9: It's a bit unclear to me if how arch_ and device_ should align in MemoryPool class. I'm definitely okay with this as an intermediate state but would love to see a comment here that describes desired end state. :D

taichi/system/unified_allocator.cpp
```diff
@@ -30,8 +32,16 @@ UnifiedAllocator::UnifiedAllocator(std::size_t size, Arch arch)
...
     auto _ = CUDAContext::get_instance().get_guard();
-    CUDADriver::get_instance().malloc_managed(&_cuda_data, size,
-                                              CU_MEM_ATTACH_GLOBAL);
+
+    Device::AllocParams alloc_params;
+    alloc_params.size = size;
+    alloc_params.host_read = true;
+    alloc_params.host_write = true;
+
+    cuda::CudaDevice *cuda_device = static_cast<cuda::CudaDevice *>(device);
```
ailzhang, line 41: DeviceAllocation carries a ptr to Device as well. Random thought: can we delay initializing device_(device_ = std::unique_ptr<Device>(new cuda::CudaDevice());) until here instead of passing it all the way down from LlvmProgramImpl?

taichi/backends/cuda/cuda_device.cpp
k-ye: > It invokes the allocation function that is JIT compiled from LLVM, which makes things more complicated than necessary

I should have clarified here: The LLVM module needs to have a function for memory allocation, because the sparse SNode system needs that. This part is completely fine. However, I feel like it's simpler to implement the entire memory allocation functionality on the host side, then passing in a function pointer into the JIT module, rather than the other way around. Right now the situation is a bit complicated:

* If CUDA unified memory is enabled, then we are passing in the function pointer to the JIT module: https://github.com/taichi-dev/taichi/blob/e5914a458776bb921dd95cb9a00126dafe675b7a/taichi/llvm/llvm_program.cpp#L345-L350
* Otherwise, the memory allocation is done completely inside the JIT module: https://github.com/taichi-dev/taichi/blob/e5914a458776bb921dd95cb9a00126dafe675b7a/taichi/runtime/llvm/runtime.cpp#L783-L814

taichi/system/memory_pool.cpp
AmesingFlank: I understand your concern and I find it a bit unclear as well. The desired end state is to move the MemoryPool inside each Device, so that the memory allocated from each Device can be used as-is. I will add a comment as well.

taichi/backends/cuda/cuda_device.cpp
AmesingFlank: That makes sense, let's aim to only implement memory allocation on the host, and always pass a func ptr to LLVMRuntime.

taichi/system/unified_allocator.cpp
AmesingFlank: Hmm I'm not quite following, are you suggesting that we let `unified_allocator` own the `device`? Also I imagine in the future the `device_` will be used in many other places, so very soon this might not be its first usage anymore. I find that it makes more sense to initialize it within `ProgramImpl`.

taichi/system/unified_allocator.h
```diff
@@ -56,6 +58,9 @@ class UnifiedAllocator {
   }
 
   UnifiedAllocator operator=(const UnifiedAllocator &) = delete;
+
+ private:
+  Device *device_;
```
k-ye, line 63:
```suggestion
  Device *device_{nullptr};
```

taichi/system/unified_allocator.cpp
ailzhang: Yea I was indeed thinking about simplifying the ownership and trying to avoid passing around a device ptr. ;)  But if our desired end state is to let Device own the memoryPool this makes sense to me!
(nit: IMHO an easier process is creating a `DesiredDevice` class and make it work e2e, then rename & remove the old Device class in the end to show people the difference before and after the refactor ;)

ailzhang (approved):
LGTM!

taichi/system/unified_allocator.cpp
AmesingFlank: What would be the difference between `DesiredDevice` and `Device`?

AmesingFlank:
I think there's something wrong with `liscense/cla` :((

taichi/system/unified_allocator.cpp
ailzhang: I'd imagine we first create a `DesiredDevice` to include the MemoryPool, keep adding it to the codebase until it works e2e. And finally remove `Device` class and rename `DesiredDevice` to `Device`. ;) The main advantage is we always have the old and new classes for reviewers to compare even in the intermediate PRs. But ofc this is my personal choice to make smaller PRs haha.  It's totally up to you how to split the refactor into PRs ;)

AmesingFlank: I see! That makes sense 🤔

PR #2968, opened 2021-09-22
[refactor] Avoid passing device strings into torch

strongoier:
For now (prototyping stage) we are using `torch.Tensor` as storage for Taichi ndarrays. When constructing ndarrays, we will pass device strings `cpu` or `cuda:0` into `torch.zeros`. However, there are some known issues like https://github.com/pytorch/pytorch/issues/50779, https://discuss.pytorch.org/t/std-regex-replace-not-work-with-libtorch-1-7-0/102253, which may prevent users from using Taichi as a part of their software. Therefore I'm getting rid of these device strings in this PR.

### DIFF
--- a/python/taichi/lang/ndarray.py
+++ b/python/taichi/lang/ndarray.py
@@ -16,13 +16,10 @@ class Ndarray:
         assert has_pytorch(
         ), "PyTorch must be available if you want to create a Taichi ndarray."
         import torch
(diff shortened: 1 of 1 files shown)


k-ye (approved):
LGTM!

ailzhang (approved):
Thanks!

PR #2971, opened 2021-09-22
[refactor] Minimize Python context

strongoier:
Related issue = #282

### Background
The Taichi compiler first translates the AST of a kernel/function into another Python function, and runs that function to generate C++ AST. In order to run that function, the Python context (global vars and non-local vars) needs to be passed in.

### Problem
Previous way is to use `func.__globals__` for all global vars (not limited to what are used in that `func`) and ad-hoc manipulation for non-local vars. The former may make the Python context too large and the latter might be error-prone.

### Solution
Use `inspect.getclosurevars(func)` instead. This provides a minimum Python context, which only captures actually used global/non-local variables. This may also serve as a first step towards cleaning up the compilation process.

### Other details
After AST translation of force-inlined functions, the annotations of arguments should be removed (effects have been translated into function body) to avoid using `ti.template()` before importing `taichi`. Comments of force-inlined functions are also fixed by the way (which should be done in #2871).

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -290,17 +290,8 @@ class KernelArgError(Exception):
 
 
 def _get_global_vars(func):
-    # Discussions: https://github.com/taichi-dev/taichi/issues/282
-    global_vars = copy.copy(func.__globals__)
-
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Awesome thanks!

PR #2976, opened 2021-09-23
[misc] Remove unnecessary symlink step.

ailzhang:
When I debugged the nightly release failure I realized that `pip uninstall taichi -y` won't fully delete the `taichi` folder in the python `site-packages`. This happens since we generates a symlinked
`libtaichi_core.so` at runtime that `pip` isn't aware of. So `pip`
leaves it in `site-packages/taichi/lib` during uninstall. This is **BAD** since it means even after
you uninstall taichi, you'll still be able to `import taichi`(altho empty!) due to the
`site-packages/taichi` folder.

Note this only happens when you install taichi from a wheel, `python
setup.py install/develop` works fine.

Related issue = #

### DIFF
--- a/python/taichi/core/util.py
+++ b/python/taichi/core/util.py
@@ -124,12 +124,6 @@ def get_unique_task_id():
 
 
 sys.path.append(os.path.join(package_root(), 'lib'))
-if settings.get_os_name() != 'win':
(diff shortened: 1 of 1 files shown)


Leonz5288 (approved):
Great!

PR #2981, opened 2021-09-23
[cuda] Cuda Device API 1/n: memory allocation

AmesingFlank:
This PR begins the work on CUDA deivce API.

The PR only implements memory allocation and deallocation, and makes use of those methods in LlvmProgram and MemoryPool.

This would be very useful for unifying the memcpy logic used in GGUI, which would facilitate implementing GGUI on other backends.


AmesingFlank:
Re-submitted the PR because the previous one is stuck on license/cla

ailzhang (approved):
LGTM！

PR #2982, opened 2021-09-23
[ci] Disable fail-fast matrix on release jobs.

ailzhang:
This fits release jobs better than presubmit jobs. (in the sense that we
want release builds to finish as many as possible but in presubmit jobs
we want to save minutes.

Related issue = #


k-ye (approved):
Great, thank you!

PR #2987, opened 2021-09-24
[Doc] Remove 'Versioning and releases'; Fix 'Documentation writing guide'

strongoier:
This PR does the following:
- Removing `Versioning and releases` from documentation because it is out-dated and we are not going to expose this to public for now
- Fixing typos, grammars, non-working syntax and other nits in `Documentation writing guide`


ailzhang (approved):
Looks great , thanks!

PR #2995, opened 2021-09-25
[misc] Add Vulkan as a target for ti diagnose

### LINKED ISSUE #2959: Add Vulkan as a target for TI diagnose
At the moment, Taichi system diagnose does not print system support for Vulkan. 
```
$ti diagnose
...
cc: True
cpu: True
metal: False
opengl: True
cuda: True
```
Would be nice to have `Vulkan: true/false` as well

KuribohG:
Related issue = fixes #2959

### DIFF
--- a/python/taichi/diagnose.py
+++ b/python/taichi/diagnose.py
@@ -46,7 +46,7 @@ def main():
     print('')
     try_print('import', 'ti')
     print('')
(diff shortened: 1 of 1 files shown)


k-ye (approved):
Thanks!

Issue #2719, opened 2021-08-17
[Doc] Improve documentation for developer installation 


qiao-bo:
The [Developer Installation](https://docs.taichi.graphics/docs/lang/articles/contribution/dev_install#linux) page has a few points could be improved. Following the page step-by-step might still encounter some installation errors. Take Linux/Ubuntu as an example:

- Additional dependencies: The following packages are needed when building Taichi (v0.7.26) on a freshly installed Ubuntu 20.04.2 LTS. `libxi-dev` `libxcursor-dev` `libxinerama-dev` `libxrandr-dev` `libx11-dev` `libgl-dev`. Based on errors from CMakeLists.txt.
- Setting up CUDA: On machines with existing installations and/or older GPUs, pay attention to Application/Driver/Runtime/CC compatibility. Suggest to at least point to the [Nvidia CUDA Installation Guide](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html) and [The Compatibility Document](https://docs.nvidia.com/deploy/cuda-compatibility/).

KuribohG:
Also, `libtinfo5` is missing. `libtinfo-dev` in the documentation is not working for me.

PR #2962, opened 2021-09-22
[misc] Support clean command in setup.py.

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* **#2962**
* #2961

```
$ python setup.py clean
running clean
removing '/Users/ailzhang/github/taichi/build' (and everything under it)
removing 'bin' (and everything under it)
removing 'python/taichi/assets' (and everything under it)
removing 'python/taichi/lib' (and everything under it)
removing 'python/taichi/examples' (and everything under it)
removing 'python/taichi/tests' (and everything under it)
removing 'python/taichi.egg-info' (and everything under it)
removing generated file taichi/common/commit_hash.h
removing generated file taichi/common/version.h
removing generated file taichi/runtime/llvm/runtime_arm64.bc
removing generated file taichi/runtime/llvm/runtime_arm64.ll

$ python setup.py clean --dry-run
```

Note this command only cleans up local information of your previous builds.
It mainly allows a fresh build without any cache from previous builds.
This is a different use case than `pip uninstall taichi` which uninstalls taichi package from your python environment.

PR #2996, opened 2021-09-25
[Doc] Developer installation update

qiao-bo:
Related issue = #2719


ailzhang (commented):
btw would you mind adding a small section about `python setup.py clean` since it's now supported? ;) https://github.com/taichi-dev/taichi/pull/2962 Thanks a lot!

qiao-bo:
> btw would you mind adding a small section about `python setup.py clean` since it's now supported? ;) #2962 Thanks a lot!

good point, will do

ljcc0930 (approved):
LGTM!

PR #3012, opened 2021-09-26
[Doc] Update documentation for fields (advanced)

strongoier:
Related issue = #

### DIFF
--- a/docs/lang/articles/advanced/layout.md
+++ b/docs/lang/articles/advanced/layout.md
@@ -42,9 +42,9 @@ Normally, you don't have to worry about the performance nuances between
 different layouts, and you can just use the default one (simply by specifying
 `shape` when creating fields) as a start.
 
-However, when the field gets large, a proper data layout can be critical to
+However, when a field gets large, a proper data layout may be critical to
 performance, especially for memory-bound applications. A carefully designed
-data layout can have much better spacial locality, which will significantly
+data layout has much better spatial locality, which will significantly
 improve cache/TLB-hit rates and cache line utilization.
 
 Taichi decouples computation from data structures, and the Taichi compiler
--- a/docs/lang/articles/advanced/layout.md
+++ b/docs/lang/articles/advanced/layout.md
@@ -193,7 +193,7 @@ ti.root.dense(ti.i, N).place(pos, vel)
 ```
 
 Then `vel[i]` is placed right next to `pos[i]`, which can increase spatial
-locality and therefore increase performance.
+locality and therefore improve performance.
 
 ### Flat layouts versus hierarchical layouts
 
--- a/docs/lang/articles/advanced/layout.md
+++ b/docs/lang/articles/advanced/layout.md
@@ -22,10 +22,12 @@ have a larger internal shape when some of its dimensions are not power-of-two.
 The downside is that the runtime performance will regress slightly.
 
 A switch named `packed` for `ti.init()` decides whether to use packed mode:
+
 ```python
 ti.init()  # default: packed=False
 a = ti.field(ti.i32, shape=(18, 65))  # padded to (32, 128)
 ```
+
 ```python
 ti.init(packed=True)
 a = ti.field(ti.i32, shape=(18, 65))  # no padding
@@ -33,24 +35,26 @@ a = ti.field(ti.i32, shape=(18, 65))  # no padding
 
 ## Advanced data layouts
 
-Besides shape, you can actually specify the data layout of a field. Defining a
-proper layout can be critical to performance, especially for memory-bound
-applications. A carefully designed data layout can significantly improve
-cache/TLB-hit rates and cache line utilization.
+Apart from shape and data type, you can actually specify the data layout of a
+field in a recursive manner. This may allow you to achieve better performance.
 
-After you are comfortable with the default layout specification (simply by specifying
-`shape` when creating fields), and you realize performance is your top
-priority, you can try to migrate to more advanced recursively-defined data
-layouts.
+Normally, you don't have to worry about the performance nuances between
+different layouts, and should just use the default one (simply by specifying
+`shape` when creating fields) as a start.
+
+However, when the field gets large, a proper data layout can be critical to
+performance, especially for memory-bound applications. A carefully designed
+data layout can have much better spacial locality, which will significantly
+improve cache/TLB-hit rates and cache line utilization.
 
 Taichi decouples computation from data structures, and the Taichi compiler
 automatically optimizes data accesses on a specific data layout. This allows
-programmers to quickly experiment with different data layouts and figure out
-the most efficient one on a specific task and computer architecture.
+you to quickly experiment with different data layouts and figure out the most
+efficient one on a specific task and computer architecture.
 
-### From `shape` to `ti.root.X`
+### Layout 101: from `shape` to `ti.root.X`
 
-For example, this declares a 0-D field:
+The following declares a 0-D field:
 
 ```python {1-2}
 x = ti.field(ti.f32)
@@ -59,7 +63,7 @@ ti.root.place(x)
 x = ti.field(ti.f32, shape=())
 ```
 
-This declares a 1-D field of size `3`:
+The following declares a 1D field of shape `3`:
 
 ```python {1-2}
 x = ti.field(ti.f32)
@@ -68,7 +72,7 @@ ti.root.dense(ti.i, 3).place(x)
 x = ti.field(ti.f32, shape=3)
 ```
 
-This declares a 2-D field of shape `(3, 4)`:
+The following declares a 2D field of shape `(3, 4)`:
 
 ```python {1-2}
 x = ti.field(ti.f32)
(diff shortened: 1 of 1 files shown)


docs/lang/articles/advanced/layout.md
```diff
@@ -2,30 +2,59 @@
...
+## Packed mode
 
-In Taichi, the layout is defined in a recursive manner.
-We suggest starting with the default layout specification (simply
-by specifying `shape` when creating fields using
-`ti.field/ti.Vector.field/ti.Matrix.field`), and then migrate to more
-advanced layouts using the `ti.root.X` syntax if necessary.
+By default, all non-power-of-two dimensions of a field are automatically
+padded to a power of two. For instance, a field of shape `(18, 65)` will
+have internal shape `(32, 128)`. Although the padding has many benefits
```
victoriacity, line 14:
```suggestion
have an internal shape of `(32, 128)`. Although the padding has many benefits
```

docs/lang/articles/advanced/layout.md
```diff
@@ -2,30 +2,59 @@
...
+such as allowing fast and convenient bitwise operations for coordinate
+handling, it will consume potentially much more memory than you originally
+think.
```
victoriacity, line 17:
```suggestion
such as allowing fast and convenient bitwise operations for accessing coordinates
of hierarchical data layouts, it will consume potentially much more memory than
the user expects.
```
I wonder if we shall describe "coordinate handing" more explicitly, i.e., use bitwise operations to traverse different levels in a hierarchical data structure.

docs/lang/articles/advanced/layout.md
```diff
@@ -2,30 +2,59 @@
...
 
-## From `shape` to `ti.root.X`
+If you indeed want smaller memory usage, you could use the optional packed
+mode. In packed mode, no more padding will be applied so a field will not
+have a larger internal shape when some of its dimensions are not power-of-two.
+The downside is that the runtime performance will regress slightly.
```
victoriacity, line 22:
```suggestion
If you would like to reduce memory usage, you can use the optional packed
mode. In packed mode, no padding is applied such that a field does not
have a larger internal shape than when some of its dimensions are not powers oftwo.
The downside is that the runtime performance will regress slightly.
```

docs/lang/articles/advanced/layout.md
```diff
@@ -2,30 +2,59 @@
...
+```
+
+```python
+ti.init(packed=True)
+a = ti.field(ti.i32, shape=(18, 65))  # no padding
+```
+
+## Advanced data layouts
+
+Apart from shape and data type, you can actually specify the data layout of a
```
victoriacity, line 38:
```suggestion
Apart from shape and data type, you can also specify the data layout of a
```

docs/lang/articles/advanced/layout.md
```diff
@@ -2,30 +2,59 @@
...
+field in a recursive manner. This may allow you to achieve better performance.
+
+Normally, you don't have to worry about the performance nuances between
+different layouts, and should just use the default one (simply by specifying
```
victoriacity, line 42:
```suggestion
different layouts, and you can just use the default one (simply by specifying
```

docs/lang/articles/advanced/layout.md
```diff
@@ -2,30 +2,59 @@
...
+`shape` when creating fields) as a start.
+
+However, when the field gets large, a proper data layout can be critical to
+performance, especially for memory-bound applications. A carefully designed
+data layout can have much better spacial locality, which will significantly
+improve cache/TLB-hit rates and cache line utilization.
```
victoriacity, line 48:
```suggestion
However, when a field gets large, a proper data layout may be critical to
performance, especially for memory-bound applications. A carefully designed
data layout has much better spatial locality, which will significantly
improve cache/TLB-hit rates and cache line utilization.
```

docs/lang/articles/advanced/layout.md
```diff
@@ -52,44 +81,41 @@ ti.root.dense(ti.ij, (3, 4)).place(x)
...
-## Row-major versus column-major
+After being comfortable with these equivalent definitions, you can move forward
+and see how to change the data layout.
 
-Let's start with the simplest layout.
+### Row-major versus column-major
 
-Since address spaces are linear in modern computers, for 1D Taichi
-fields, the address of the `i`-th element is simply `i`.
+As you may already know, address spaces are linear in modern computers. To
```
victoriacity, line 89:
```suggestion
As you might have learned in a computer architecture course,
address spaces are linear in modern computers. To
```

docs/lang/articles/advanced/layout.md
```diff
@@ -52,44 +81,41 @@ ti.root.dense(ti.ij, (3, 4)).place(x)
...
+simplify the discussions, data type size will not be considered and will always
+be treated as 1. Assume the starting address of a field is `base`. Then for 1D
+Taichi fields, the address of the `i`-th element is simply `base + i`.
 
-To store a multi-dimensional field, however, it has to be flattened, in
-order to fit into the 1D address space. For example, to store a 2D field
-of size `(3, 2)`, there are two ways to do this:
+However, a multi-dimensional field has to be flattened in order to fit into the
+1D address space. For example, to store a 2D field of size `(3, 2)`, there are
+two ways:
```
victoriacity, line 96:
```suggestion
1D address space. For example, there are two ways to store a 2D field of size `(3, 2)`:
```

docs/lang/articles/advanced/layout.md
```diff
@@ -52,44 +81,41 @@ ti.root.dense(ti.ij, (3, 4)).place(x)
...
+#       x:  x[0, 0]  x[0, 1]  x[1, 0]  x[1, 1]  x[2, 0]  x[2, 1]
+#       y:  y[0, 0]  y[1, 0]  y[2, 0]  y[0, 1]  y[1, 1]  y[2, 1]
 ```
 
-What do we find here? `x` first increases the first index (i.e. row-major), while `y`
-first increases the second index (i.e. column-major).
-
 :::note
 
 For those people from C/C++, here's what they look like:
```
victoriacity, line 121:
```suggestion
For those who are familiar with C/C++, here is what they look like in C code:
```

docs/lang/articles/advanced/layout.md
```diff
@@ -100,54 +126,50 @@ int y[2][3];  // column-major
...
-start. However, locality sometimes have a significant impact on the
-performance, especially when the field is huge.
-
-**To improve spatial locality of memory accesses (i.e. cache hit rate /
-cacheline utilization), it's sometimes helpful to place the data
-elements within relatively close storage locations if they are often
-accessed together.** Take a simple 1D wave equation solver for example:
+**To improve spatial locality of memory accesses, it's sometimes helpful to
+place the data elements within relatively close addresses if they are often
+accessed together.** Take a simple 1D wave equation solver as an example:
```
victoriacity, line 172:
```suggestion
**To improve spatial locality of memory accesses, it may be helpful to
place data elements that are often accessed together within
relatively close addresses.** Take a simple 1D wave equation solver as an example:
```

docs/lang/articles/advanced/layout.md
```diff
@@ -162,40 +184,37 @@ def step():
...
+locality and poor performance. A better way is to place them together:
 
 ```python
 ti.root.dense(ti.i, N).place(pos, vel)
 ```
 
-Then `vel[i]` is placed right next to `pos[i]`, this can increase the
-cache-hit rate and therefore increase the performance.
+Then `vel[i]` is placed right next to `pos[i]`, which can increase spatial
+locality and therefore increase performance.
```
victoriacity, line 196:
```suggestion
locality and therefore improve performance.
```

docs/lang/articles/advanced/layout.md
```diff
@@ -162,40 +184,37 @@ def step():
...
 
-## Flat layouts versus hierarchical layouts
+### Flat layouts versus hierarchical layouts
 
-By default, when allocating a `ti.field`, it follows the simplest data
-layout.
+As you may already know from the above discussions,
```
victoriacity, line 200:
```suggestion
From the above discussions,
```

docs/lang/articles/advanced/layout.md
```diff
@@ -162,40 +184,37 @@ def step():
...
-leads to poor access locality under certain computation tasks.
-Specifically, in tasks such as texture trilinear interpolation, the two
-elements are not even within the same `4KB` pages, creating a huge
-cache/TLB pressure.
+However, at times this data layout can be suboptimal for certain types of
+computation tasks. For example, in texture trilinear interpolation,
+`val[i, j, k]` and `val[i + 1, j, k]` are often accessed together. With the
+above layout, they are very far away (`32 KB`) from each other, and not even
+within the same `4KB` pages. This creates a huge cache/TLB pressure and leads
+to pretty poor performance.
```
victoriacity, line 217:
```suggestion
However, at times this data layout may be suboptimal for certain types of
computation tasks. For example, in trilinear texture interpolation,
`val[i, j, k]` and `val[i + 1, j, k]` are often accessed together. With the
above layout, they are very far away (32 KB) from each other, and not even
within the same 4 KB pages. This creates a huge cache/TLB pressure and leads
to poor performance.
```

docs/lang/articles/advanced/layout.md
```diff
@@ -206,69 +225,146 @@ ti.root.dense(ti.ijk, (8, 16, 32)).dense(ti.ijk, (4, 4, 4)).place(val)
...
+order.
 
-If `A` is hierarchical, it will be iterated level by level. This
-maximizes the memory bandwidth utilization in most cases.
+If `A` is hierarchical, it will be iterated level by level. This maximizes the
+memory bandwidth utilization in most cases.
 
-Struct-for loops on sparse fields follow the same philosophy, and will
-be discussed further in [Sparse computation](./sparse.md).
+As you may notice, we only cover dense data layouts in this section. For sparse
```
victoriacity, line 247:
```suggestion
As you may notice, only dense data layouts are covered in this section. For sparse
```

docs/lang/articles/advanced/layout.md
```diff
@@ -206,69 +225,146 @@ ti.root.dense(ti.ijk, (8, 16, 32)).dense(ti.ijk, (4, 4, 4)).place(val)
...
-    ti.root.dense(ti.i, 1024).place(pos(i))
+    ti.root.dense(ti.i, 1024).place(pos.get_scalar_field(i))
 for i in range(3):
-    ti.root.dense(ti.i, 1024).place(vel(i))
+    ti.root.dense(ti.i, 1024).place(vel.get_scalar_field(i))
+```
+
+## Dynamic field allocation and destruction
+
+Use the `FieldsBuilder` class for dynamic field allocation and destruction.
```
victoriacity, line 300:
```suggestion
You can use the `FieldsBuilder` class for dynamic field allocation and destruction.
```

docs/lang/articles/advanced/layout.md
strongoier: Here I meant `a field does not have a larger internal shape than the defined shape when some of its dimensions are not powers of two`. Is there a better way to express that?

docs/lang/articles/advanced/layout.md
strongoier: I have a little concern about using `hierarchical data layouts` before actually introducing this term to users.

docs/lang/articles/advanced/layout.md
victoriacity: I think this is already clear enough 😄

docs/lang/articles/advanced/layout.md
victoriacity: Yes this totally makes sense. Let's just use "coordinate handling" and leave hierarchical layouts to the following sections.

victoriacity (approved):
LGTM!

PR #3017, opened 2021-09-26
[ci] Reduce Docker image size to 1/2 by removing build cache (Ubuntu 18.04 -> Ubuntu 20.04)

### LINKED ISSUE #3016: The latest Docker image is broken and the Dockerfile built with error: No module named taichi_core
**Describe the bug**

The latest docker image on Docker Hub was pushed 2 months ago, and it is broken.

```
$ docker run -it taichidev/taichi:latest /bin/bash

# Inside the container
root@0a46ea7585f6:/taichi-dev/taichi# taichi
bash: taichi: command not found

# Empy taichi folder inside the container
root@0a46ea7585f6:/taichi-dev/taichi# ls
# Empty folder
```

Thus I built the image using the Dockerfile from latest source code, but failed to execute `taichi` and `ti` in the docker container.

**To Reproduce**

- Ubuntu 20.04
- Docker 19.03.13

```bash
# After cloning the latest version from github master branch
$ docker build -t taichi:latest .
$ docker run -it taichi:latest /bin/bash
root@e7a6a7986674 # taichi
```

**Log/Screenshots**
```
root@e7a6a7986674:/taichi-dev/taichi# taichi
Share object taichi_core import failed, check this page for possible solutions:
https://docs.taichi.graphics/docs/lang/articles/misc/install
  File "/taichi-dev/taichi/bin/taichi", line 3, in <module>
    import taichi
  File "/taichi-dev/taichi/python/taichi/__init__.py", line 3, in <module>
    from taichi.core import *
  File "/taichi-dev/taichi/python/taichi/core/__init__.py", line 1, in <module>
  File "/taichi-dev/taichi/python/taichi/core/logging.py", line 4, in <module>
    from taichi.core import util
  File "/taichi-dev/taichi/python/taichi/core/util.py", line 127, in <module>
    import_ti_core()
  File "/taichi-dev/taichi/python/taichi/core/util.py", line 45, in import_ti_core
    raise e from None
  File "/taichi-dev/taichi/python/taichi/core/util.py", line 33, in import_ti_core
    import taichi_core as core
ModuleNotFoundError: No module named 'taichi_core'
Consider installing libtinfo5: sudo apt-get install libtinfo5
```

wuhanstudio:
Related issue =  close #3016

Hi, I find the Dockerfile is a little outdated, and the following improvements drastically reduce the docker image size:

* Use Docker multi-stage builds to exclude temporary build files from the final image

* Upgrade base image from Ubuntu 18.04 to Ubuntu 20.04

~~* Install LLVM-10 via apt rather than build from source~~

* Install CMake via apt rather than build from source

The original docker image is around 6GB which is a little intimidating for a new user, now the final released docker image is here (~200MB extra layer and ~600MB in total):

https://hub.docker.com/r/wuhanstudio/taichi/tags?page=1&ordering=last_updated

Hope you find it helpful,
Han

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Taichi Dockerfile for development
-FROM nvidia/cuda:10.0-devel-ubuntu18.04
+FROM nvidia/cuda:11.0-devel-ubuntu20.04 as builder
 
 LABEL maintainer="https://github.com/taichi-dev"
 
@@ -7,10 +7,11 @@ LABEL maintainer="https://github.com/taichi-dev"
 # docker image is upgraded to Ubuntu 20.04 this will
 # install Python 3.8 by default
 RUN apt-get update && \
+    DEBIAN_FRONTEND="noninteractive" \
     apt-get install -y software-properties-common \
                        python3-pip \
                        libtinfo-dev \
-                       clang-8 \
+                       clang-10 \
                        wget \
                        git \
                        libx11-dev \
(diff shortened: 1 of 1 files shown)


qiao-bo:
Thanks again for this PR, the image size reduction is really useful.

qiao-bo:
we might need to check if the Ubuntu-provided LLVM satisfies our requirement. Generally, we use a customized built version.

wuhanstudio:
> we might need to check if the Ubuntu-provided LLVM satisfies our requirement. Generally, we use a customized built version.

Hi, I just updated this PR.

Indeed, perhaps it's better to preserve the customized build version of LLVM, so that it does not introduce too many chages.

The docker image is still as small as 0.6GB using customized LLVM.

qiao-bo:
seems like libz3-4 still needs to be installed.

wuhanstudio:
> seems like libz3-4 still needs to be installed.

Indeed, somehow libz3-4 is still required for the latest version of Taichi.

These are the minimum requirements I find for the runtime:

```
RUN apt-get update && \
    apt-get install -y python3-pip \
                       libx11-dev \
                       libtinfo5 \
                       libz3-4
```

qiao-bo:
A note here, this Dockerfile is meant to provide an environment for Taichi development. Namely, both building Taichi and running Taichi tests on available backends (such as x86 and CUDA). Only passing the a whl between the docker stages is not sufficient. We do need to keep the building environment...  ;-)

wuhanstudio:
Eventually, I find a balance between a minimal runtime and  a full development environment.

Now, the Dockerfile still uses the customized LLVM with full CUDA support and taichi source code.

```
COPY --from=builder /tmp/llvm-install/ /usr/local/
COPY --from=builder /taichi-dev/taichi /taichi-dev/taichi
```

The docker image size is reduced:

```
From:
10.0-devel-ubuntu18.04 (2.24GB) + extra Taichi layer (4.17GB) = 6.41GB
To:
11.0-devel-ubuntu20.04 (3.66GB) + extra Taichi layer (2.04)= 5.70GB
```

So that the final released docker image contains no temporary build cache, but can be used as a development environment as well.

feisuzhu:
Closing ancient PR

PR #3020, opened 2021-09-27
[lang] [opt] Memory allocator for ti.ndarray

qiao-bo:
Related issue = #2771 

**Proposed feature**
Using Taichi's own memory allocator to move away from the dependency on Pytorch.

**Status**
supported targets
- [ ] x86
- [ ] CUDA

### DIFF
--- a/taichi/program/ndarray.cpp
+++ b/taichi/program/ndarray.cpp
@@ -1,0 +1,63 @@
+#include <numeric> 
+#include "ndarray.h"
+
+namespace taichi {
+namespace lang {
+
+Ndarray::Ndarray(Program *prog, const DataType type, 
+                 const std::vector<int> &shape)
+    : program(prog), dtype(type), shape(shape), 
+      data_ptr(nullptr), nelement(1), element_size(1) {
+  LlvmProgramImpl *prog_ = program->get_llvm_program_impl();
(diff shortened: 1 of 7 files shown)


taichi/program/ndarray.cpp
```diff
@@ -0,0 +1,68 @@
...
+      element_size(1) {
+  LlvmProgramImpl *prog_ = program->get_llvm_program_impl();
+  TaichiLLVMContext *tlctx = prog_->get_llvm_context(program->config.arch);
+
+  nelement = std::accumulate(std::begin(shape), std::end(shape), 1,
+                             std::multiplies<>());
+  element_size = data_type_size(dtype);
+  auto *const runtime_jit = tlctx->runtime_jit_module;
+  runtime_jit->call<void *, std::size_t, std::size_t>(
+      "runtime_snode_tree_allocate_aligned", prog_->get_llvm_runtime(),
```
k-ye, line 24: tiny nit: create a separate PR just to rename this function first?

k-ye:
> ndarray_buffer_manager is very similar to snode_tree_buffer_manager. Maybe there can be a parent class and let these two inherit from it.

If possible, let's just get rid of this manager... What we really need is just an allocator

PR #3021, opened 2021-09-27
[misc] Get changelogs via tags instead of commit messages

strongoier:
Related issue = #

Previously, `make_changelog.py` identifies the last commit with `[release]` message as the last version. After this PR, `make_changelog.py` identifies the last commit with a tag as the last version. This allows a more flexible releasing process.

### DIFF
--- a/misc/make_changelog.py
+++ b/misc/make_changelog.py
@@ -19,6 +19,7 @@ def load_pr_tags():
 
 def main(ver='master', repo_dir='.'):
     g = Repo(repo_dir)
+    commits_with_tags = set([tag.commit for tag in g.tags])
     commits = list(g.iter_commits(ver, max_count=200))
     begin, end = -1, 0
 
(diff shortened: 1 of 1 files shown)


misc/make_changelog.py
```diff
@@ -32,7 +33,7 @@ def format(c):
 
     for i, c in enumerate(commits):
         s = format(c)
-        if s.startswith('[release]'):
+        if c in commits_with_tags:
```
ailzhang, line 36: tag can also has names: so maybe we can name our release tag v0.x.x[rc] and check here?

strongoier: That sounds good. However, we support commands like `python3 make_changelog.py v0.7.32` to show changelog of `v0.7.32`. This requires us to maintain backward-compatibility. Unfortunately, early versions of Taichi doesn't follow this naming rule (e.g., no `v`). IMO if we only use tags for releases, this implementation without format checking will be fine.

ailzhang: okay hmmm let's do this for now, would like to make it more restricted in the future ;). Since github doesn't have protected tags, our make_changelog.py witout any enforcement of "only use tags for releases".

ailzhang (approved):
LGTM!

Issue #2715, opened 2021-08-17
Put repeating code blocks in Github CI workflows into scripts.


Leonz5288:
Currently we have several ci workflows running every PR/release, and considerate amount of code in these workflows are repeating(having the same functionality).

For example, basically every workflows requires we build taichi from source. Therefore, if these kinds of code can be put into bash/shell scripts, the readability of these workflows can be much higher and they can also be much more easier to fix/debug.

To begin this work, files under `.github/workflows/` can be a place to start.

lucywsq:
Hi, @frostming  please join this slack channel, then we would respond to your problems on time. We can communicate faster as well.

ailzhang:
Closing as this is mostly done, thanks!

PR #3034, opened 2021-09-28
[ci] Run shell scripts in CI

Leonz5288:
Related issue = #2715
Work In Progress


.github/workflows/scripts/unix_test.sh
```diff
@@ -0,0 +1,13 @@
+TAICHI_REPO_DIR=`pwd`
+TI_LIB_DIR=`python -c "import taichi;print(taichi.__path__[0])" | tail -1`
+TI_LIB_DIR="$TI_LIB_DIR/lib" ./build/taichi_cpp_tests
+export PATH=$TAICHI_REPO_DIR/taichi-llvm/bin/:PATH
+export PATH=$PATH:$HOME/.local/bin
```
ailzhang, line 5: This line doesn't look right, if you installed without --user, this shouldn't be added to PATH.

.github/workflows/scripts/unix_build.sh
```diff
@@ -0,0 +1,10 @@
+export PATH=`pwd`/taichi-llvm/bin/:$LLVM_PATH:$PATH
+python3 -m pip uninstall taichi -y
```
ailzhang, line 2: taichi and taichi-nightly

.github/workflows/scripts/unix_test.sh
Leonz5288: But this was in every Test section in presubmit.yml.

ailzhang (approved):
Thanks!

ailzhang:
Btw this PR need an update before merging (#3064 #3068 ) otherwise it might fail on master. (no rush!)

PR #3036, opened 2021-09-28
[LLVM] Rename runtime memory allocation function

qiao-bo:
Give the runtime memory allocation function a more generic name.


strongoier (approved):
LGTM!

PR #3047, opened 2021-09-29
[Lang] [bug] Fix subscripting user classes in Taichi kernels

### LINKED ISSUE #3046: Subscription of user-defined classes worked in v0.7.26 but not v0.8.0
The following code snippet:
```python
import taichi as ti

ti.init()

class MyList:
    def __init__(self, elements):
        self.elements = elements

    def __getitem__(self, index):
        return self.elements[index]

@ti.kernel
def func():
    for i in ti.static(range(3)):
        print(a[i])

a = MyList([1, 2, 3])
func()
```
worked well in v0.7.26.

However in v0.8.0, `TypeError: Subscription (e.g., "a[i, j]") only works on fields or external arrays.` was reported.

strongoier:
Related issue = fixes #3046

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -196,10 +196,8 @@ def subscript(value, *indices):
                 f'Field with dim {field_dim} accessed with indices of dim {index_dim}'
             )
         return Expr(_ti_core.subscript(value.ptr, indices_expr_group))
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Awesome, thanks for fixing it!

PR #3055, opened 2021-09-30
[python] Add context manager support for ui.Gui

### LINKED ISSUE #3015: Use python's `with` construct for IMGUI in GGUI
Currently, each IMGUI window in GGUI is declared by a `window.GUI.begin()` and a `window.GUI.end()`. For example, to create a subwindow with a single button in it, one would write
```python
window.GUI.begin("window", 0.1, 0.1, 0.5, 0.5)
window.GUI.button("my button")
window.GUI.end()
```
This API can be made more elegant using python's `with` construct:
```python
with window.GUI.subwindow() as w:
    w.button("my button")
```

Refer to [context management in python](https://stackoverflow.com/questions/1984325/explaining-pythons-enter-and-exit)

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

Related issue = close #3015

### DIFF
--- a/python/taichi/ui/gui.py
+++ b/python/taichi/ui/gui.py
@@ -13,6 +13,16 @@ class Gui:
     def __init__(self, gui) -> None:
         self.gui = gui  #reference to a PyGui
 
+    def __enter__(self, name, x, y, width, height):
+        """Creating a context manager for subwindows, all args are the same with `begin`.
+        """
+        return self.begin(name, x, y, width, height)
+
+    def __exit__(self):
+        """Exiting a context manager for subwindows.
+        """
+        self.end()
+
     def begin(self, name, x, y, width, height):
         """Creates a subwindow that holds imgui widgets.
 


Zheaoli:
LGTM , would you mind adding some extra unit tests for the change?

Xuanwo:
> LGTM , would you mind adding some extra unit tests for the change?

Of course. Can you guide me to the existing unit tests about the GUI class?

python/taichi/ui/gui.py
```diff
@@ -13,6 +13,16 @@ class Gui:
     def __init__(self, gui) -> None:
         self.gui = gui  #reference to a PyGui
 
+    def __enter__(self, name, x, y, width, height):
+        """Creating a context manager for subwindows, all args are the same with `begin`.
+        """
+        return self.begin(name, x, y, width, height)
+
+    def __exit__(self):
```
yihong0618, line 21: exit may has args.

yihong0618 (commented):
```python
def __exit__(self, exc_type, exc_val, exc_tb):

```

use may better 

```python
def __exit__(self, *args)
````

File: python/taichi/ui/gui.py (hunk 1 above)
Comment (yihong0618) [reply python/taichi/ui/gui.py:21]: doc:
https://docs.python.org/3/library/stdtypes.html?highlight=__exit__#contextmanager.__exit__

Comment (Zheaoli) [reply python/taichi/ui/gui.py]: make sense

Comment (Xuanwo) [reply python/taichi/ui/gui.py]: fixed.

Comment (yihong0618) [reply python/taichi/ui/gui.py]: cool

**Comment by Zheaoli** [comment]
LGTM

**Review by Zheaoli — APPROVED** [review]
LGTM

**Comment by Zheaoli** [comment]
PTAL @ailzhang

File: python/taichi/ui/gui.py [hunk 2]
```diff
@@ -13,6 +13,16 @@ class Gui:
     def __init__(self, gui) -> None:
         self.gui = gui  #reference to a PyGui
 
+    def __enter__(self, name, x, y, width, height):
```
frostming, line 16: `__enter__` does not accept other arguments than `self`
https://docs.python.org/zh-cn/3/reference/datamodel.html#object.__enter__

Xuanwo: Oops, I will revisit the docs of `__enter__`.

Xuanwo: @frostming Fixed, PTAL

Xuanwo:
This PR will introduce a new API called `subwindow`. The function name is chosen without much deep thought. Please feel free to propose other options.

python/taichi/ui/gui.py
```diff
@@ -13,6 +14,21 @@ class Gui:
     def __init__(self, gui) -> None:
         self.gui = gui  #reference to a PyGui
 
+    @contextmanager
+    def subwindow(self, name, x, y, width, height):
```
Zheaoli, line 18: maybe `sub_window` is better. FYI https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

python/taichi/ui/gui.py
```diff
@@ -13,6 +14,21 @@ class Gui:
...
+        """Creating a context manager for subwindow, all args are the same with `begin`.
+
+        Usage:
+
+        with gui.subwindow(name, x, y, width, height) as g:
```
k-ye, line 23: nit: One more indentation level, or sphinx can't recognize this as codeblock (IIUC): https://github.com/taichi-dev/taichi/blob/cc2dd3420f239c8811aa58f9dcfb89b9fbdfd8d3/python/taichi/lang/kernel_arguments.py#L94-L102

Xuanwo: Looks nice now！

python/taichi/ui/gui.py
```diff
@@ -13,6 +14,21 @@ class Gui:
     def __init__(self, gui) -> None:
         self.gui = gui  #reference to a PyGui
 
+    @contextmanager
+    def sub_window(self, name, x, y, width, height):
+        """Creating a context manager for subwindow, all args are the same with `begin`.
```
ailzhang, line 19: `all args are the same with begin` : since this is user facing we should copy paste the args over from `begin` and add an additional comment to developers saying "all args of this method should align with begin`. 
Alternatively how about we only expose this context manager to users? This is cleaner and easier for users? cc: @AmesingFlank

AmesingFlank: Sounds good, let's remove `begin` and `end` and only keep `sub_window`. Although @Xuanwo , no need to worry about doing that in this PR, and many thanks for contributing!

AmesingFlank: Additionally, removing `begin` and `end` and only keeping `sub_window` can help users prevent mismatched `begin`s and `end`s. It's a really good idea.

Xuanwo: I created an issue to track this idea, and I will fix it later.

python/taichi/ui/gui.py
```diff
@@ -13,6 +14,30 @@ class Gui:
     def __init__(self, gui) -> None:
         self.gui = gui  #reference to a PyGui
 
+    @contextmanager
+    def sub_window(self, name, x, y, width, height):
+        """Creating a context manager for subwindow
+
+        Note::
+            All args of this method should align with `begin`.
```
ailzhang, line 22: nit: this line should be removed along with begin/end in the next PR ;)

ailzhang (approved):
Nice work, thanks for contributing!

python/taichi/ui/gui.py
```diff
@@ -13,6 +14,30 @@ class Gui:
...
+
+        Args::
+            x (float): The x-coordinate (between 0 and 1) of the top-left corner of the subwindow, relative to the full window.
+            y (float): The y-coordinate (between 0 and 1) of the top-left corner of the subwindow, relative to the full window.
+            width (float): The width of the subwindow relative to the full window.
+            height (float): The height of the subwindow relative to the full window.
+
+        Usage::
```
k-ye, line 30: About the single `:` versus double `::` situation, we only need to use the double `::` before a code block. So ti would still be `Args:`, but `Usage::` :-)

Xuanwo: Got.

Xuanwo:
Merge with master to fix the code format check.

Issue #2846, opened 2021-08-30
Exception raising in Field APIs


victoriacity:
The current Field APIs, including ScalarField, MatrixField, and Structfield (after #2627) use assertions statements to check input validity, for example,

https://github.com/taichi-dev/taichi/blob/03e2923d103704048c101051d337fa31f9dbb1de/python/taichi/lang/matrix.py#L207

In software engineering practice, assertion is used to verify whether the program itself is implemented correctly regardless of user inputs, while exceptions/errors are used to check validity of the inputs. Therefore quite a few occurrences of `assert` statements in [`field.py`](https://github.com/taichi-dev/taichi/blob/master/python/taichi/lang/field.py) and [`matrix.py`](https://github.com/taichi-dev/taichi/blob/master/python/taichi/lang/matrix.py) are supposed to raise errors.

Contributions are greatly welcomed!

ljjsalt:
Could you please provide an example of changing an `assert` statement to an exception?
Also, any resources that help write exceptions in Taichi are appreciated.

gaocegege:
> Could you please give an example of changing an `assert` statement to an exception? Also, any resources that help write exceptions in Taichi are appreciated.

I refined the exception when converting from/to numpy array in the PR https://github.com/taichi-dev/taichi/pull/3058. And there are many other asserts in the codebase. Welcome other contributions! You can have a look at the PR, as an example.

ljjsalt:
> > Could you please give an example of changing an `assert` statement to an exception? Also, any resources that help write exceptions in Taichi are appreciated.
> 
> I refined the exception when converting from/to numpy array in the PR #3058. And there are many other asserts in the codebase. Welcome other contributions! You can have a look at the PR, as an example.
> 
> I am unassigning myself since it is an umbrella issue.

Thanks, gocegege! I have created a pull request.

PR #3058, opened 2021-09-30
[Lang] Use more user-friendly exception when converting from numpy array

gaocegege:
Related issue = #2846

Test

```python
x = ti.field(ti.f32, 2)
x_np = np.array([1, 7, 3, 5])
x.from_numpy(x_np)

```

Result

```
  File "./demo/test.py", line 30, in <module>
    x.from_numpy(x_np)
  File "/Users/gaocedidi/code/taichi/python/taichi/lang/util.py", line 207, in wrapped
    return func(*args, **kwargs)
  File "/Users/gaocedidi/code/taichi/python/taichi/lang/field.py", line 258, in from_numpy
    "ti.field shape {} does not equal to the numpy array shape {}".format(self.shape, arr.shape))
ValueError: ti.field shape (2,) does not equal to the numpy array shape (4,)
```

BTW, should we catch the exceptions in the field and hide the stacktrace to be more user-friendly?


python/taichi/lang/field.py
```diff
@@ -251,6 +253,9 @@ def to_torch(self, device=None):
     @python_scope
     def from_numpy(self, arr):
         assert len(self.shape) == len(arr.shape)
+        if self.shape != arr.shape:
+            raise ValueError(
+                "ti.field shape {} does not equal to the numpy array shape {}".format(self.shape, arr.shape))
```
yihong0618, line 258: why not f-string?

gaocegege: SGTM. Will update it.

python/taichi/lang/field.py
```diff
@@ -250,9 +250,13 @@ def to_torch(self, device=None):
 
     @python_scope
     def from_numpy(self, arr):
-        assert len(self.shape) == len(arr.shape)
-        for i in range(len(self.shape)):
-            assert self.shape[i] == arr.shape[i]
+        if self.shape != arr.shape:
+            raise ValueError(f"ti.field shape {self.shape} does not equal"
```
ailzhang, line 254: nit: does not match the numpy array shape

python/taichi/lang/field.py
```diff
@@ -250,9 +250,13 @@ def to_torch(self, device=None):
 
     @python_scope
     def from_numpy(self, arr):
-        assert len(self.shape) == len(arr.shape)
-        for i in range(len(self.shape)):
-            assert self.shape[i] == arr.shape[i]
+        if self.shape != arr.shape:
```
ailzhang, line 253: is it `len(self.shape) == len(arr.shape)`?

python/taichi/lang/field.py
```diff
@@ -250,9 +250,13 @@ def to_torch(self, device=None):
...
+            raise ValueError(f"ti.field shape {self.shape} does not equal"
+                             f" to the numpy array shape {arr.shape}")
+        for i, _ in enumerate(self.shape):
+            if self.shape[i] != arr.shape[i]:
+                raise ValueError(f"ti.field shape {self.shape} does not equal"
+                                 f" to the numpy array shape {arr.shape}")
```
ailzhang, line 259: ditto

ailzhang (approved):
Thanks for contributing! Made a few small inline suggestions :D

gaocegege:
Thanks for your review, I will fix them ASAP

python/taichi/lang/field.py
gaocegege: Yeah, it should be len(self.shape) != len(arr.shape) here. I will update it and add one test case.

gaocegege (commented):
I fixed the potential bug and added some test cases. PTAL again.
/cc @ailzhang

ailzhang (approved):
Looks great, thanks!

gaocegege:
I rebased the upstream master to pass all CI tests

Please take a look, thanks for your time :smile:

Issue #2659, opened 2021-08-11
Supporting chained assignment e.g. ``a = b = 0`` in ti.kernel 


Ricahrd-Li:
**Concisely describe the proposed feature**
I'd like to support chained assignment like ``a = b = 0`` in ti.kernel 

**Describe the solution you'd like (if any)**
Thinking now. 

**Additional comments**
current error:
```python
>>> import taichi as ti 
[Taichi] mode=release
[Taichi] preparing sandbox at /var/folders/n_/mhc67_bn7vx8gyv83h7w29_c0000gn/T/taichi-61l517c2
[Taichi] version 0.7.24, llvm 10.0.0, commit 25778773, osx, python 3.8.3
>>> @ti.kernel 
... def func():
...     a = b = 0
... 
>>> func()
[Taichi] materializing...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/lizhehao/opt/anaconda3/lib/python3.8/site-packages/taichi/lang/kernel_impl.py", line 635, in wrapped
    return primal(*args, **kwargs)
  File "/Users/lizhehao/opt/anaconda3/lib/python3.8/site-packages/taichi/lang/kernel_impl.py", line 562, in __call__
    key = self.ensure_compiled(*args)
  File "/Users/lizhehao/opt/anaconda3/lib/python3.8/site-packages/taichi/lang/kernel_impl.py", line 553, in ensure_compiled
    self.materialize(key=key, args=args, arg_features=arg_features)
  File "/Users/lizhehao/opt/anaconda3/lib/python3.8/site-packages/taichi/lang/kernel_impl.py", line 391, in materialize
    visitor.visit(tree)
  File "/Users/lizhehao/opt/anaconda3/lib/python3.8/site-packages/taichi/lang/transformer.py", line 53, in visit
    self.pass_Preprocess.visit(tree)
  File "/Users/lizhehao/opt/anaconda3/lib/python3.8/ast.py", line 363, in visit
    return visitor(node)
  File "/Users/lizhehao/opt/anaconda3/lib/python3.8/site-packages/taichi/lang/transformer.py", line 650, in visit_Module
    self.generic_visit(node)
  File "/Users/lizhehao/opt/anaconda3/lib/python3.8/site-packages/taichi/lang/transformer.py", line 168, in generic_visit
    value = self.visit(value)
  File "/Users/lizhehao/opt/anaconda3/lib/python3.8/ast.py", line 363, in visit
[...]

PR #3062, opened 2021-09-30
[Lang] Support chained assignments

gaocegege:
Fix https://github.com/taichi-dev/taichi/issues/2659

This PR is to support chained assignments

Test case:

```python
@ti.kernel
def func():
    (a, b) = (c, d) = (7, 8)
    aa = bb = -1
    print(a, b, c, d)
    print(aa, bb)

func()
```

Python generated code

```python
Initial AST:
def func():
    a, b = c, d = 7, 8
    aa = bb = -1
    print(a, b, c, d)
    print(aa, bb)

Preprocessed:
def func():
    import taichi as ti
    if 1:
        if 1:
            __tmp_tuple = ti.expr_init_list((7, 8), 2)
            a = ti.expr_init(__tmp_tuple[0])
            b = ti.expr_init(__tmp_tuple[1])
            del __tmp_tuple
        if 1:
            __tmp_tuple = ti.expr_init_list((7, 8), 2)
            c = ti.expr_init(__tmp_tuple[0])
            d = ti.expr_init(__tmp_tuple[1])
            del __tmp_tuple
    if 1:
        aa = bb = ti.expr_init(-1)
        aa = bb = ti.expr_init(-1)
    ti.ti_print(a, b, c, d)
    ti.ti_print(aa, bb)
    del bb
    del aa
    del d
    del c
    del b
    del a
```

Result:

```
7 8 7 8
-1 -1
```


gaocegege:
/cc @k-ye 

Do we have the unit tests or integration tests for the Python code generation logic?

k-ye:
Yep, they're inside https://github.com/taichi-dev/taichi/tree/master/tests/python. See also https://docs.taichi.graphics/docs/lang/articles/contribution/write_test :-)

gaocegege:
> Yep, they're inside https://github.com/taichi-dev/taichi/tree/master/tests/python. See also https://docs.taichi.graphics/docs/lang/articles/contribution/write_test :-)

Thanks. I will add more test cases for the feature.

gaocegege:
Just mark it WIP now.

gaocegege:
I added some test cases (unpack, chained, chained + unpack, assign) for the feature, and the PR is ready to review, I think.

PTAL.

tests/python/test_assign.py
```diff
@@ -0,0 +1,47 @@
+import pytest
+
+import taichi as ti
+
+
+@ti.test()
```
k-ye, line 6: Thanks! Could you write the tests in this way:

1. Make each kernel a separate test case.
2. Use 0-D Taichi fields instead of local variables `a`, `b`; then do the assertion outside the `@ti.kernel`. The reason is that assertions inside a Taichi kernel is enabled only when `init(debug=True)` 😓

python/taichi/lang/stmt_builder.py
```diff
@@ -107,85 +106,95 @@ def build_Assign(ctx, node):
...
-                if is_local and ctx.is_creation(target.id):
-                    var_name = target.id
-                    target.ctx = ast.Store()
+        # Keep all generated assign statements and compose single one at last.
+        # The variable is introduced to support chained assignments.
+        # Ref https://github.com/taichi-dev/taichi/issues/2659.
+        assign_stmts = []
+        for node_target in node.targets:
+            if isinstance(node_target, ast.Tuple):
+                targets = node_target.elts
```
k-ye, line 115: nit: wrap  this part (L115-L166) in a dedicated function?

tests/python/test_assign.py
gaocegege: SGTM.

python/taichi/lang/stmt_builder.py
gaocegege: SGTM, it should work.

tests/python/test_assign.py
gaocegege: BTW The code will run without any warning if `debug=True` is not set, should we drop a warning message in such condition?

python/taichi/lang/stmt_builder.py
gaocegege: https://github.com/taichi-dev/taichi/pull/3062/files#diff-25d296ad7a362b20f650e4ccb9b456798aecaf36f7162a406f8e51db09f4255cR149 is added, I am not sure if the func name works for you, PTAL

gaocegege:
Please take a look, thanks for your time :smile:

lin-hitonami:
https://github.com/taichi-dev/taichi/blob/fe5b54d0e5a36afb5ad4d74008e6914330a4a47c/python/taichi/lang/stmt_builder.py#L118-L145
Can we make this part a static method and use it in both tuple and non-tuple cases?

PR #3082, opened 2021-10-02
[refactor] Style police

bobcao3:
We can hire this style police called `clang-tidy` for basically free. Not quite sure whether this guy plays well with the previous formatter, well who knows at this point.


bobcao3:
funny...

/format

taichi/backends/cuda/cuda_context.h
```diff
@@ -46,46 +46,46 @@ class CUDAContext {
               std::size_t shared_mem_bytes);
 
   void set_profiler(KernelProfilerBase *profiler) {
-    this->profiler = profiler;
+    this->profiler_ = profiler;
   }
 
   void set_debug(bool debug_) {
```
bobcao3, line 52: Very funny

ailzhang (commented):
I'm super happy to see clang-tidy being enforced on the codebase :D  nit: we'll need to check it in the format server as well. (maybe with a more accurate name than `format`)https://github.com/taichi-dev/taichi/blob/master/misc/format_server.py

yuanming-hu:
Awesome! I'll leave the code-style decision to @k-ye  and @ailzhang. My two cents: it would be nice to use @taichi-gardener to open up the initial code style unifying PR (basically adding `_`s to private variables), so that you don't get blamed for creating a large changeset!

bobcao3:
From the looks of the compilation results, there are quite some errors
caused by formatting mostly due to name conflicts. I think I will fix those
first and then somehow transfer those commits to gardener

bobcao3:
I think this will do for now, then I will use taichi gardener to make a PR on the entire fix

run-clang-tidy.py
```diff
@@ -0,0 +1,360 @@
+#!/usr/bin/env python
```
k-ye, line 1: nit: Can we put these two files under `scripts/` ?

k-ye:
> I'm super happy to see clang-tidy being enforced on the codebase :D nit: we'll need to check it in the format server as well. (maybe with a more accurate name than `format`)https://github.com/taichi-dev/taichi/blob/master/misc/format_server.py

`format_server.py` is probably no longer in use. But yeah we should integrate this into our CI format workflow https://github.com/taichi-dev/ci_workflows/blob/main/.github/workflows/auto_format.yml

k-ye (approved):
Awesome to have this!

taichi/llvm/llvm_program.cpp
```diff
@@ -9,6 +9,9 @@
 #include "taichi/codegen/codegen.h"
 #include "taichi/ir/statements.h"
 #include "taichi/backends/cpu/cpu_device.h"
+#include "taichi/backends/cuda/cuda_device.h"
+
+#include "taichi/backends/cuda/cuda_device.h"
```
yuanming-hu, line 14:
```suggestion
```

yuanming-hu (approved):
Awesome! LGTM now! Thanks.

Issue #3018, opened 2021-09-27
Multiple SNode root support


bobcao3:
Related #3004
Related #2736
Related #3024

Remaining backends to support this feature

* [ ] OpenGL
* [x] Metal

k-ye:
Disengaging this from https://github.com/taichi-dev/taichi/projects/7. We can just release this whenever the feature is ready

PR #3083, opened 2021-10-02
[lang] Add a method to get all SNodes under a root

k-ye:
Related issue = #3018

Adding this helper to make it easier for us to implement multi-root support

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,7 @@ include(cmake/TaichiCXXFlags.cmake)
 include(cmake/TaichiCore.cmake)
 include(cmake/TaichiMain.cmake)
 
-option(TI_BUILD_TESTS "Build the CPP tests" OFF)
+option(TI_BUILD_TESTS "Build the CPP tests" ON)
 
 if (TI_BUILD_TESTS)
   add_subdirectory(external/googletest EXCLUDE_FROM_ALL)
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ include(cmake/TaichiCXXFlags.cmake)
 include(cmake/TaichiCore.cmake)
 include(cmake/TaichiMain.cmake)
 
-option(TI_BUILD_TESTS "Build the CPP tests" OFF)
+option(TI_BUILD_TESTS "Build the CPP tests" ON)
```
k-ye, line 56:
```suggestion
option(TI_BUILD_TESTS "Build the CPP tests" OFF)
```

taichi/struct/snode_tree.cpp
```diff
@@ -2,6 +2,18 @@
 
 namespace taichi {
 namespace lang {
+namespace {
+
+void get_snodes_to_root_id_impl(const SNode &node,
+                                const int root_id,
+                                std::unordered_map<int, int> *map) {
```
strongoier, line 9: Is there a reason we use the `*`-way instead of `&`-way to pass the map?

strongoier (approved):
LGTM!

ailzhang (approved):
LGTM!

taichi/struct/snode_tree.cpp
k-ye: It's mostly Google C++ Coding Style -- it suggests that mutating params being specified in the end and passed in as pointers

PR #3086, opened 2021-10-04
[metal] Pull out the Runtime MSL code into its own module

k-ye:
Related issue = #3018

Decoupling the runtime part from the root/SNodes, so that it is easier for us to support multiple SNode roots.

### DIFF
--- a/taichi/backends/metal/struct_metal.cpp
+++ b/taichi/backends/metal/struct_metal.cpp
@@ -13,7 +13,8 @@
 #include "taichi/math/arithmetic.h"
 #include "taichi/util/line_appender.h"
 
-TLANG_NAMESPACE_BEGIN
+namespace taichi {
+namespace lang {
 namespace metal {
 namespace {
 namespace shaders {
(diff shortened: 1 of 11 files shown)


taichi/backends/metal/struct_metal.cpp
```diff
@@ -340,15 +336,39 @@ class StructCompiler {
...
   void emit_runtime_structs() {
     line_appender_.append_raw(shaders::kMetalRuntimeStructsSourceCode);
     emit("");
     emit("struct Runtime {{");
-    emit("  SNodeMeta snode_metas[{}];", max_snodes_);
-    emit("  SNodeExtractors snode_extractors[{}];", max_snodes_);
-    emit("  ListManagerData snode_lists[{}];", max_snodes_);
-    emit("  NodeManagerData snode_allocators[{}];", max_snodes_);
-    emit("  NodeManagerData::ElemIndex ambient_indices[{}];", max_snodes_);
+    emit("  SNodeMeta snode_metas[{}];", kMaxNumSNodes);
```
bobcao3, line 367: How big is kMaxNumSNodes? Will it be a concern in the future

bobcao3 (approved):
I'm not very familiar with this backend but looks good to me

taichi/backends/metal/struct_metal.cpp
k-ye: Good point! This is 1024 now https://github.com/taichi-dev/taichi/blob/fce9108c7c3cfda1b3b828212606f38ded7144ba/taichi/inc/constants.h#L12

Unfortunately the sparse runtime part cannot be easily separated into a per-root buffer yet, so all the sparse info (for multiple roots) still share a single global buffer.

PR #3089, opened 2021-10-05
[Doc] Fix an example in the documentation for Coordinate offsets

0xrabbyte:
Related issue = #
I think the original example given in the code is not right according to Ln 18.


strongoier (approved):
Welcome to the Taichi community! Thanks for catching this! Would you please sign the CLA so that the PR can be merged after passing CI?

yuanming-hu (approved):
Welcome, @0xrabbyte! Thanks for the contribution. Fixing this error will make readers better understand the coordinate offset system. It's interesting that the error stays there for this long :-)

Issue #3069, opened 2021-10-01
ti.init(arch=ti.opengl) failed with Taichi 0.8.1


Y7K4:
**Describe the bug**
With Taichi 0.8.1, `ti.init(arch=ti.opengl)` doesn't work.

Btw, `ti.opengl` with v0.7.32 and `ti.cpu` with v0.8.1 both work.

**To Reproduce**
A minimal Python file to reproduce:

```py
import taichi as ti

ti.init(arch=ti.opengl)
```

**Log/Screenshots**
Full error message:

```
(venv) kyc@kyc-T470p:~$ python3 example.py 
[Taichi] version 0.8.1, llvm 10.0.0, commit cc2dd342, linux, python 3.8.10
[Taichi] Starting on arch=opengl
[E 09/30/21 23:36:16.302 52849] [opengl_device.cpp:check_opengl_error@30] glClearBufferSubData: GL_INVALID_OPERATION


***********************************
* Taichi Compiler Stack Traceback *
***********************************
/home/kyc/venv/lib/python3.8/site-packages/taichi/core/../lib/taichi_core.so: taichi::Logger::error(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool)
/home/kyc/venv/lib/python3.8/site-packages/taichi/core/../lib/taichi_core.so: taichi::lang::opengl::check_opengl_error(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
/home/kyc/venv/lib/python3.8/site-packages/taichi/core/../lib/taichi_core.so: taichi::lang::opengl::GLCommandList::CmdBufferFill::execute()
/home/kyc/venv/lib/python3.8/site-packages/taichi/core/../lib/taichi_core.so: taichi::lang::opengl::GLStream::submit(taichi::lang::CommandList*)
/home/kyc/venv/lib/python3.8/site-packages/taichi/core/../lib/taichi_core.so: taichi::lang::opengl::GLStream::submit_synced(taichi::lang::CommandList*)
/home/kyc/venv/lib/python3.8/site-packages/taichi/core/../lib/taichi_core.so: taichi::lang::opengl::OpenGlRuntime::OpenGlRuntime()
[...]

PR #3091, opened 2021-10-06
[OpenGL] Fix minor bugs in GL Device

bobcao3:
(Maybe) Related issue = #3069

- `glCreateBuffers` is the DSA variants of creating objects, since we are targeting GL4.3 without DSA, we should use genBuffers
- This might have caused the issue #3069, however I'm not sure and I don't have intel device to test this with. Detailed theory will be posted in the issue
- Caught another small bug (not impacting anything yet) as we forgot to record the offset field of the clear buffer command


k-ye (approved):
LGTM, thanks!

PR #3094, opened 2021-10-06
[perf] Batched constant folding JIT evaluator

bobcao3:
And this new evaluator blows up Vulkan codegen.. lol

### DIFF
--- a/taichi/backends/vulkan/embedded_device.cpp
+++ b/taichi/backends/vulkan/embedded_device.cpp
@@ -18,7 +18,7 @@ namespace vulkan {
 namespace {
 
 // FIXME: NDEBUG is broken, so just manually enable this if necessary.
-constexpr bool kEnableValidationLayers = false;
+constexpr bool kEnableValidationLayers = true;
 
 const std::vector<const char *> kValidationLayers = {
     "VK_LAYER_KHRONOS_validation",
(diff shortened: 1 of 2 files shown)


bobcao3:
I think I should revert the batching part but still put in the part that cleans up the compile config of the evaluators. Those are pretty big.

ailzhang:
> According to @bobcao3 's benchmark, the major overhead comes from LLVM/SPIRV's optimizers. 

Sorry for missing this, would you mind adding a pointer to the benchmark result? ;) Thanks!

k-ye:
Could you also revert the part regarding `SPIRV-Tools` :-) ?

bobcao3:
> 
> 
> Could you also revert the part regarding `SPIRV-Tools` :-) ?

I think the submodule part can stay as it does give us a consistency benefit so we are always using the same version of this tool instead of depending on the user / build bot's Vulkan SDK version

bobcao3:
> 
> 
> > According to @bobcao3 's benchmark, the major overhead comes from LLVM/SPIRV's optimizers.
> 
> Sorry for missing this, would you mind adding a pointer to the benchmark result? ;) Thanks!

When using batched version, since all the evaluators are batched we can't cache compiled evaluators anymore. Which means we have a few more compilation to do. And in addition, the profiler says that 96.27% of the compile time is spent on the backend compiler for LLVM, and around 77% for Vulkan SPIR-V optimizer. And in both CPU and Vulkan backends, total compile time are over 90% of the constant folding pass' time. The old approach caches compiled evaluators, and since the evaluator is essentially a tuple of OpType, RHS, LHS, RET types, there are quite a lot of hits.

PR #3108, opened 2021-10-08
[perf] Constant folding optimization

bobcao3:
Moved from https://github.com/taichi-dev/taichi/pull/3094

---

This PR turns off a bunch of the default optimization passes for the JIT evaluator kernels which should speed up constant folding time as the kernel compiling time is non-trivial.

In addition, integer to FP casts and bit casts are done on the CPU as there will be no difference as all backends follow IEEE standards. SPIRV-Tools is also moved to a submodule so that we do not depend on the version provided by the users' Vulkan SDK

### DIFF
--- a/taichi/transforms/constant_fold.cpp
+++ b/taichi/transforms/constant_fold.cpp
@@ -56,21 +56,12 @@ class ConstantFold : public BasicStmtVisitor {
       current_ast_builder().insert(std::move(ret));
     };
 
-    auto program_compile_config_org = program->config;
-    program->config.advanced_optimization = false;
-    program->config.cfg_optimization = false;
-    program->config.simplify_before_lower_access = false;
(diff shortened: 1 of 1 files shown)


bobcao3:
CPU backend & mpm88 test:

Before: `68.426 ms 23.43%  taichi::lang::irpass::constant_fold [40 x   1.711 ms]`
After: `53.021 ms 18.78%  taichi::lang::irpass::constant_fold [40 x   1.326 ms]`

Vulkan:

Before: `27.120 ms 10.55%  taichi::lang::irpass::constant_fold [40 x 678.000 us]`
After: `15.804 ms  6.46%  taichi::lang::irpass::constant_fold [40 x 395.100 us]`

bobcao3:
Oops...

PR #3129, opened 2021-10-09
[test] Parametrize test_ad_basics.py

k-ye:
Related issue = #3124


strongoier:
> We should get rid of the internal ti.test instead of the outer one :P
> We usually add ti.test decorators to test_xxx() which are pytest-aware. It's a bit confusing that it doesn't work in this file :P

IIUC it indeed works, and the problem is double decorators which work independently.

PR #3130, opened 2021-10-09
[misc] Delete accidentally included spirv dump

bobcao3:
Related issue = #


k-ye:
Why would this require a /format ...

Issue #3126, opened 2021-10-09
Speed up CI build by caching pip dependencies


ailzhang:
@yihong0618 reminded that we can cache installed pip dependencies to speed up build time. https://github.com/taichi-dev/taichi/commit/bc0182ccffb25c68f8493fc9037e47ea615bde59#r57732885 Thanks!
Some more references: https://medium.com/ai2-blog/python-caching-in-github-actions-e9452698e98d

PR #3139, opened 2021-10-10
[CI] Add pip cache

zstone12:
Related issue = #3126 

1.I’m not good at give the step a name,like `Check code format for git init` can be improved.
2.I'm not sure these places whether need to add pip cache,like line 234 `python -m pip install -r requirements_dev.txt`。if need to do it, I am willing to do it.:D
3.refer to @yihong0618,these is a step Set Variable.I think it's used for adapting diferent enviroment .In this case,we just choose the python3.8 and ubuntu,So I don't add this step.
My English isn’t that good,forgive me.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -32,7 +32,7 @@ jobs:
         with:
           python-version: 3.8
           
-      - name: Check code format for git init 
+      - name: Setup git & clang-format
         run: |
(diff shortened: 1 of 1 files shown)


zstone12:
> Yea caching for code_format check is good enough for now since we don't update it often. If we want to add caching for other jobs it's a bit tricky to make sure we invalidate cache correctly as they're updated more frequently ;)

Got it.

PR #3141, opened 2021-10-11
[Bug] Throw proper error message when creating more than 32 snode trees in LLVM backend.

ailzhang:
* #3143
* #3142
* __->__ #3141

Noticed this `mem double free/segfault` when trying out @qiao-bo @strongoier's memory allocator
script, thought that we should throw a proper error message here
instead.


qiao-bo:
Can someone explain briefly why it's 32?

strongoier:
> Can someone explain briefly why it's 32? not like 64?

Maybe we need to @ljcc0930.

k-ye:
> Can someone explain briefly why it's 32? not like 64?

It's just some random number we picked...

PR #3175, opened 2021-10-12
[vulkan] Fix ti.sync() for vulkan

AmesingFlank:
Our vk backend doesn't submit a cmd list until an operation that requires syncing is performed.
Thus, during `synchronize()`, we should also check if there's a pending cmd list, and submit it if so.


k-ye:
Oops, PR title

PR #3179, opened 2021-10-13
[vulkan] Fixing some type mismatches

bobcao3:
Related issue = #3164

Not sure whether this causes #3164 but this comes up in validation, and could be a reason.


g1n0st:
@bobcao3 Should we merge this?

Issue #3110, opened 2021-10-08
Drawing lines with `canvas.lines()` does not work as expected


bcolloran:
**Describe the bug**

I've been having trouble drawing lines with `canvas.lines()`.

The short script below defines 10 points in a semicircle around one point at the center of the semicircle, and then defines edges between the central point and all others. Only five of the expected 10 lines are drawn.

**To Reproduce**
```python
import numpy as np
import taichi as ti

ti.init(arch=ti.gpu)

x_np = np.pi * np.arange(0, 1, 0.1)
x_np = np.vstack([np.cos(x_np), np.sin(x_np)])
x_np = np.hstack([[[0], [0]], x_np]).T
x_np = 0.3 * x_np + 0.5

edges_np = np.array([[0, i] for i in range(1, x_np.shape[0])])

n_vertices = x_np.shape[0]
n_edges = edges_np.shape[0]

x = ti.Vector.field(2, shape=n_vertices, dtype=float)
x.from_numpy(x_np)

edges = ti.Vector.field(2, shape=n_edges, dtype=int)
edges.from_numpy(edges_np)

print(edges) # output included below

window = ti.ui.Window("draw lines", res=(512, 512))
canvas = window.get_canvas()

while window.running:
    canvas.lines(vertices=x, indices=edges, width=0.002)
    canvas.circles(centers=x, radius=0.005)
    window.show()
```

**Log/Screenshots**
here is the full terminal output when i run this script:

```
[Taichi] version 0.8.0, llvm 10.0.0, commit 181c9039, linux, python 3.9.7
[Taichi] Starting on arch=cuda
[[ 0  1]
 [ 0  2]
 [ 0  3]
 [ 0  4]
 [ 0  5]
 [ 0  6]
 [ 0  7]
 [ 0  8]
 [ 0  9]
 [ 0 10]]
[W 10/07/21 22:54:13.619 48481] [vulkan_device.cpp:buffer@526] Overriding last binding
[W 10/07/21 22:54:13.626 48481] [vulkan_device.cpp:buffer@526] Overriding last binding
```

Note that the list of `edges` that is printed includes an edge from point zero to all 10 other points, but in the screenshot below only 5 lines are drawn:
![image](https://user-images.githubusercontent.com/2380975/136505737-94999dae-aa9f-4cd0-b2a0-531c2eb2aa90.png)

**Additional comments**
I have just recently discovered Taichi and I *love it*. Thank you, and keep up the amazing work!

[...]

PR #3181, opened 2021-10-13
[gui] Accept array of vectors as indices for lines and triangles

Leonz5288:
Related issue = #3110
This method is somehow hacky and needs to be discussed.
My perspective:
In renderable.cpp update_data(), indices are processed without differences about drawings. It makes more sense to modify the code in their separate scope. However, in these situations, all we can do is to modify the shape of the indices.
I might miss something about the process of rendering, so please give any suggestions.

### DIFF
--- a/taichi/ui/backends/vulkan/renderable.cpp
+++ b/taichi/ui/backends/vulkan/renderable.cpp
@@ -41,10 +41,9 @@ void Renderable::update_data(const RenderableInfo &info) {
   int num_vertices = info.vbo.shape[0];
   int num_indices;
   if (info.indices.valid) {
-    num_indices = info.indices.shape[0];
-    if (info.indices.matrix_cols == 1) {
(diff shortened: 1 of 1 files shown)


Leonz5288:
I believe this is what you meat. @AmesingFlank

AmesingFlank:
> I believe this is what you meant. @AmesingFlank However, I am kind of worry if this would affect other drawing APIs in a bad way?

I don't think this will break anything. Just to be sure, could you try running each of the files in `examples/ggui_examples`? Again, I am terribly sorry that I haven't implemented any automated testing mechanisms yet.

Leonz5288:
> > I believe this is what you meant. @AmesingFlank However, I am kind of worry if this would affect other drawing APIs in a bad way?
> 
> I don't think this will break anything. Just to be sure, could you try running each of the files in `examples/ggui_examples`? Again, I am terribly sorry that I haven't implemented any automated testing mechanisms yet.

Already tested, nothing is broken :)

Issue #3019, opened 2021-09-27
`property` decorator is not supported in `data_oriented` class


ljcc0930:
Found while updating ODOP docs. #3006

Reproduce code:
```py
import taichi as ti

ti.init(arch=ti.cuda)

@ti.data_oriented
class Test:
    @property
    @ti.kernel
    def test(self) -> ti.i32:
        return 0

a = Test()
print(a.test)
```

Traceback:
```bash
Traceback (most recent call last):
  File "/home/ljcc/repo/taichi/work.py", line 13, in <module>
    print(a.test)
  File "/home/ljcc/repo/taichi/python/taichi/lang/kernel_impl.py", line 767, in getattr
    x = super(cls, self).__getattribute__(item)
  File "/home/ljcc/repo/taichi/python/taichi/lang/kernel_impl.py", line 660, in wrapped
    assert not hasattr(clsobj, '_data_oriented')
AssertionError
```

ljcc0930:
Another error example:
```py
import taichi as ti

ti.init(arch=ti.cuda)

@ti.data_oriented
class Counter():
    num_ = ti.field(dtype=ti.i32, shape=(32, ))
    def __init__(self, data_range):
        self.range = data_range
        self.add(data_range[0], data_range[1], 1)

    @classmethod
    @ti.kernel
    def add(cls, l: ti.i32, r: ti.i32, d: ti.i32):
        for i in range(l, r):
            cls.num_[i] += d

    @property
    @ti.kernel
    def num(self) -> ti.i32:
        ret = 0
        for i in range(self.range[0], self.range[1]):
            ret += self.num_[i]
        return ret

a = Counter((0, 5))
b = Counter((4, 10))
print(b.num)
```

yihong0618:
I've looked at the code and it doesn't currently support this kind of  ` decortator stacking`

you need to write another method

yihong0618:
@ltt1598 

About `@classmethod` stacking I think just raise maybe a better way,  what do you think ?

ltt1598:
> @ltt1598
> 
> About `@classmethod` stacking I think just raise maybe a better way, what do you think ?

Uh sorry, what does it mean by "just raise"?

yihong0618:
```
import taichi as ti

ti.init(arch=ti.cuda)

@ti.data_oriented
class Test:
    @classmethod
    @ti.kernel
    def test(self) -> ti.i32:
        return 0

a = Test()

--->

[...]

PR #3186, opened 2021-10-14
[Lang] [bug] Fix support staticmethod decorator for data_oriented class

yihong0618:
Related issue = #3019

Also support the `@staticmethod` decorator


yihong0618:
@ailzhang @Zheaoli @yuanming-hu can you take a look, and make some change for `@property` support.

Issue #2711, opened 2021-08-17
[vulkan] Clear unnecessary GLSL shader files on Vulkan backend


g1n0st:
Currently, the GLSL shader files in folder `taichi/backends/vulkan/shaders` are outdated. Welcome contributors to remove them and make sure the dependency is clean. : )

PR #3211, opened 2021-10-17
[vulkan] Clear unnecessary GLSL shader files on Vulkan backend

0xzhang:
Related issue = #2711


0xzhang:
Time to do more thing! :smile:

PR #3215, opened 2021-10-18
[Example] Fix the ti example problems of sparse matrix demos

Hanke98:
Related issue = #3216 #3116 #3081

When you type `ti example stable_fluid`, it will not run the example successfully:
```
[Taichi] version 0.8.3, llvm 10.0.0, commit 306cdbf3, linux, python 3.9.5

**      Taichi Programming Language      **

Docs:   https://taichi.rtfd.io/en/stable
GitHub: https://github.com/taichi-dev/taichi
Forum:  https://forum.taichi.graphics

Running example stable_fluid ...
usage: stable_fluid.py [-h] [-s]
stable_fluid.py: error: unrecognized arguments: example stable_fluid
```
So does the implicit mass-spring demo.

The reason is that the `runpy.run_path` method will pass the `sys.argv` which contains the arguments of the command: `ti example ...` to the example scripts and some of these arguments will not be recognized by the demos.

Here, I use a quick solution: use `ArgumentParser.parse_known_args()`  instead of `ArgumentParser.parse_args()`. 

In the long run, I think we should support passing the arguments correctly from the `ti` CLI to examples since some arguments are important to the examples.

### DIFF
--- a/python/taichi/main.py
+++ b/python/taichi/main.py
@@ -181,7 +181,7 @@ class TaichiMain:
             action='store_true',
             help="Save source code to current directory instead of running it")
 
-        #TODO: Pass the arguments to donwstreams correctly(#3216).
+        # TODO: Pass the arguments to downstream correctly(#3216).
         args = parser.parse_args(arguments)
 
         examples_dir = TaichiMain._get_examples_dir()


ailzhang:
Alternatively if we could only pass ` sys.argv[2:]` to downstream( ofc add a comment saying that we skipped ti, example) that seems better?

PR #3229, opened 2021-10-19
[vulkan] Fix fence timeout

AmesingFlank:


bobcao3:
lmao, should we just make it UINT64_MAX so that extremely long compute will survive?

PR #2637, opened 2021-08-04
[IR] Support local tensor

squarefk:
## IR
Python:
```
a = ti.Vector([1., 2., 3.])

print(a[?])
```

Initial IR:
```
<[TensorType> $1 = alloca @tmp     // FrontendAllocaStmt(IdExpression, shape, element_type)
@tmp[0] = elements[0]              // FrontendAssignStmt(LocalTensorElementExpression(IdExpression, indices), elements)
@tmp[1] = elements[1]
@tmp[2] = elements[2]
$? = eval @tmp[?]                  // FrontendEvalStmt
print $?
```

Lower AST (eliminate IdExpression):
```
<TensorType> $1 = alloca @tmp      // AllocaStmt
<Type*> $2 = shift ptr [$1 + $?]   // ShiftGlobalPtrStmt
<Type> $3 = local store [$2 <- $?] // LocalStore

<Type> $4 = local load [$2]        // LocalLoad
```

Offloaded:
AllocaStmt may be converted into GlobalTemporaryStmt or just keep AllocaStmt

## Issues
Currently, there are three issues:
- <del>Offload pass only deal with ConstStmt type for offset in PtrOffsetStmt. This will introduce some code restrictions but should be easy to resolve. </del>. Done with another PR.
- During make_local_block pass, Alloca TensorType need LocalStore to initialize which introduces extra difficulty. Hence, it is disabled.
- Loop index keep using individual variable instead of TensorType for now. We may want to change this with another PR. At least the following aspects should be considered.
  2. for I in ti.grouped(ti.ndrange()): need to make sure Alloca TensorType inside loop body.
  3. Loop indices are looked up by Alloca id in some places. Extra work is needed to extract different dim indices from the same Alloca TensorType

Issue #2590, opened 2021-07-27
Support dynamic indexing in ti.Vector/ti.Matrix


k-ye:
The original issue was #1004. This issue tracks the progress of supporting the new feature. Detailed design stories are commented below.

strongoier:
Dynamic indexing refers to the following two cases:
```py
@ti.kernel
def dynamic_index_local_vector(i: ti.i32) -> ti.f32:
    a = ti.Vector([0.0, 1.0, 2.0])
    return a[i]

b = ti.Vector.field(3, ti.f32, shape=())

@ti.kernel
def dynamic_index_vector_field(i: ti.i32) -> ti.f32:
    return b[None][i]
```

We would like to gradually support this feature with `ti.init(dynamic_index=True)`.

- [ ] If we can finally confirm that this feature will not harm performance of non-dynamic-indexing code, we will make this behavior as default.

A few months ago, @squarefk made a prototype for this in #2543, #2637, #2673, #2773. He also gave a talk (https://github.com/taichi-dev/taichicon/releases/download/taichicon02/Dynamic.Indexing-Freely.Access.Vector.and.Matrix.Elements.at.Runtime-Yu.Fang.pdf) at TaichiCon02 to present his ideas - common use cases can be translated to the form of `pointer + offset` by the compiler. Here are some important technical details of his implementation:
- `TensorType` is added in CHI IR for vectors and matrices;
- `PtrOffsetStmt` is added in CHI IR for accessing a memory address of form `pointer + offset`;
- `TensorElementExpression` is added in frontend IR to represent a vector/matrix with dynamic indices.

However, there are still lots of engineering efforts to be done in order to make this idea fully fit into the language and the compiler. Recently, I've been driving the development of this feature. Related work includes #3218, #3228, #3237, #3244, #3468, #3517, #3524, #3546, #3599. Also, this feature pushes us to support frontend type check (#3301). Now we can completely use dynamic indexing of local vectors/matrices now.

My next focus is to fully support dynamic indexing of elements of vector/matrix fields. A n-D vector field is currently implemented as a tuple of n scalar fields, and we will refer to them as sub-fields below. Here is the preliminary roadmap:
- [ ] Enforce each sub-field has the same shape.
- [ ] Develop an algorithm to identify the cases where the dynamic indexing can be turned into the form of `pointer + offset` at compile time, and use that form for these cases.
- [ ] Handle other rare use cases with a fallback plan:
```py
# assume v is an element of a vector field
[...]

PR #3237, opened 2021-10-20
[opt] Enable CFG optimization for local tensors

strongoier:
Related issue = #2590, #2637, #3218, #3228

Although #2637 introduced local tensors, related CFG optimization was not implemented, resulting in redundant local memory allocation/load/store in many cases. This PR enables CFG optimization for local tensors, and eliminates the overhead of #3218 and #3228. Let's look at a tiny example:
```python
import taichi as ti

ti.init(dynamic_index=True, print_preprocessed=True, print_ir=True)

@ti.kernel
def my_func():
    a = ti.Vector([1, 2])
    print(a.sum())

my_func()
```

Before this PR, the final IR for the kernel is:
```
kernel {
  $0 = offloaded  
  body {
    <[Tensor (2) i32]> $1 = alloca
    <i32> $2 = const [1]
    <i32> $3 = const [0]
    <*i32> $4 = shift ptr [$1 + $3]
    <i32> $5 : local store [$4 <- $2]
    <i32> $6 = const [2]
    <i32> $7 = const [4]
    <*i32> $8 = shift ptr [$1 + $7]
    <i32> $9 : local store [$8 <- $6]
    <i32> $10 = local load [ [$4[0]]]
    <i32> $11 = local load [ [$8[0]]]
    <i32> $12 = add $10 $11
    print $12, "\n"
  }
}
```

After this PR, the final IR for this kernel is:
```
kernel {
  $0 = offloaded  
  body {
    <i32> $1 = const [3]
    print $1, "\n"
  }
}
```

Details:
1. Add a not-too-conservative alias analysis for `PtrOffsetStmt`, which is able to produce `definitely same/different` results.
2. Stop treating a `TensorType alloca` as a store to enable store-to-load forwarding. This is because currently local tensors must be initialized with values (`ti.Vector([1, 2, 3])`). Then we don't treat `TensorType alloca` itself as a valid forwarding source.
3. Stop exposing a `PtrOffsetStmt` with an `alloca` origin to other offloaded tasks (it shouldn't appear in final node `live_in`) to enable dead store elimination.

### DIFF
--- a/taichi/analysis/alias_analysis.cpp
+++ b/taichi/analysis/alias_analysis.cpp
@@ -31,9 +31,11 @@ AliasResult alias_analysis(Stmt *var1, Stmt *var2) {
   if (origin1 != nullptr && origin2 != nullptr) {
     if (origin1 == origin2) {
       if (var1->is<PtrOffsetStmt>() && var2->is<PtrOffsetStmt>()) {
-        auto diff = value_diff_ptr_index(var1->cast<PtrOffsetStmt>()->offset, var2->cast<PtrOffsetStmt>()->offset);
+        auto diff = value_diff_ptr_index(var1->cast<PtrOffsetStmt>()->offset,
+                                         var2->cast<PtrOffsetStmt>()->offset);
         if (diff.is_diff_certain) {
-          return diff.diff_range == 0 ? AliasResult::same : AliasResult::different;
+          return diff.diff_range == 0 ? AliasResult::same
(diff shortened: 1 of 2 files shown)


strongoier:
> LGTM! (Just curious, is it possible that there is a local tensor without an initial value?)

Current it is impossible. But maybe we will support that in the future, and the CFG part will be updated accordingly.

PR #3245, opened 2021-10-21
[metal] Add mem_offset_in_parent to AOT module

k-ye:
Related issue = #

This PR stores the memory offset of each field within the root buffer.

### DIFF
--- a/taichi/backends/wasm/aot_module_builder_impl.h
+++ b/taichi/backends/wasm/aot_module_builder_impl.h
@@ -24,7 +24,8 @@ class AotModuleBuilderImpl : public AotModuleBuilder {
   void add_per_backend_tmpl(const std::string &identifier,
                             const std::string &key,
                             Kernel *kernel) override;
-  void add_per_backend_field(const std::string &Identifier,
+  void add_field_per_backend(const std::string &Identifier,
+                             const SNode *rep_snode,
                              bool is_scalar,
                              DataType dt,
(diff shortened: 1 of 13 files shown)


taichi/backends/wasm/aot_module_builder_impl.h
```diff
@@ -24,7 +24,8 @@ class AotModuleBuilderImpl : public AotModuleBuilder {
   void add_per_backend_tmpl(const std::string &identifier,
                             const std::string &key,
                             Kernel *kernel) override;
-  void add_per_backend_field(const std::string &Identifier,
+  void add_field_per_backend(const std::string &Identifier,
```
bobcao3, line 27: When we have the snode, is the identifier still important?

bobcao3 (approved):
other than that LGTM!

ailzhang (approved):
Looks great! 
(now I know what you meant about missing mem_offset in the opengl backend :P will add one to the opengl backend as well! 
Btw shall we rebase this on top of master and enable aot test on metal backend as well? https://github.com/taichi-dev/taichi/blob/master/tests/python/test_aot.py#L31

taichi/backends/wasm/aot_module_builder_impl.h
k-ye: It kind of is. Users will be able to retrieve the info of the field (represented by this SNode) via the identifier.

PR #3314, opened 2021-10-29
[cuda] Perf: Use the min between saturation grid dim and const range-for dim

bobcao3:
Related issue = #


qiao-bo:
@bobcao3 shall we get this PR into the master? why the change is gone? there are some opengl stuff instead?

bobcao3:
I think I messed up my branches 😂

Issue #3151, opened 2021-10-11
Join us as a new Taichi contributor!


lucywsq:
Below are features that are relatively friendly to new comers. Everyone is welcome to participate :-)
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-403e542b-7fff-8acf-11d2-f6591ba3c26f"><div dir="ltr" style="margin-left:0pt;" align="left">

**Welcome contribution**

Issue | Link
-- | --
Support more atomic operations | https://github.com/taichi-dev/taichi/issues/2675
Support efficient operations for large matrix | https://github.com/taichi-dev/taichi/issues/2696
[refactor] Make Callable::Args/Ret serializable | https://github.com/taichi-dev/taichi/issues/2625
Deprecate Expr::operator= | https://github.com/taichi-dev/taichi/issues/2684
[vulkan] Clear unnecessary GLSL shader files on Vulkan backend | https://github.com/taichi-dev/taichi/issues/2711
Put repeating code blocks in Github CI workflows into scripts. | https://github.com/taichi-dev/taichi/issues/2715
Invoke materialize() only before calling a Taichi kernel | https://github.com/taichi-dev/taichi/issues/2730

</div></b>

lucywsq:
These issues haven't been assigned yet. You can still choose one to do some excellent work : )

Put repeating code blocks in Github CI workflows into scripts. | #2715
-- | --
[lang] Expose thread index in @ti.kernel | #2725
Support more atomic operations | #2675
Support efficient operations for large matrix | #2696
Deprecate Expr::operator= | #2684

hantsy:
Why not add these issues to the yearly hacker event: https://hacktoberfest.digitalocean.com/, and hackerfest hashtag on Twitter.

lucywsq:
@hantsy Thanks for your suggestion :)

lucywsq:
Here are some new issues with the tag "good first issue". Welcome to take one and implement it. 
We are looking forward to your contributions! : )


Issue | Link
-- | --
[refactor] Clean up keep_raw | https://github.com/taichi-dev/taichi/issues/2718
Support setting the random seed during initialization | https://github.com/taichi-dev/taichi/issues/2734
Exception raising in Field APIs | https://github.com/taichi-dev/taichi/issues/2846
[docs] Add doc about building python part without compiling taichi-core | https://github.com/taichi-dev/taichi/issues/3056

lucywsq:
[...]

Issue #2725, opened 2021-08-18
[lang] Expose thread index in @ti.kernel


yolo2themoon:
**Concisely describe the proposed feature**
Getting the thread index is a common operation in parallel computing, and exposing the thread index in `@ti.kernel` with something like `ti.block_idx()` and `ti.thread_idx()` might be useful for some applications.

**Describe the solution you'd like (if any)**
For the CUDA backend, all I can think of is that the thread index and block index are stored in special registers and can be accessed directly. As for the frontend, not sure if we can refer to `ti.block_dim(block_dim)` to implement `ti.thread_idx()`

lucywsq:
Hi @Luke-Skycrawler  please join this slack channel, then we would respond to your problems on time. We can communicate faster as well.

yolo2themoon:
#3319  has been merged into the master. Many thanks to @Luke-Skycrawler . cc @lucywsq

PR #3319, opened 2021-10-29
[Lang] Implement ti.global_thread_idx

Luke-Skycrawler:
Related issue = #2725 #3151

### DIFF
--- a/taichi/transforms/lower_ast.cpp
+++ b/taichi/transforms/lower_ast.cpp
@@ -190,10 +190,6 @@ class LowerAST : public IRVisitor {
     // do nothing
   }
 
-  void visit(GlobalThreadIndexStmt *stmt) override {
-    // do nothing
-  }
-
   void visit(BinaryOpStmt *stmt) override {
     // do nothing
   }


k-ye (commented):
Great start && welcome to the community!

`LoopLinearIndexStmt` is specifically used for the BLS case, and represents the threadIdx within a CUDA block. What we need here is a global thread index, i.e. `blockDim.x * blockIdx.x + threadIdx.x`. So I'd suggest that we have a specific IR, e.g. `GlobalThreadIndexStmt`. This would also work on CPU backends, as we can just use the thread ID (within the thread pool) :-)

(I do think we can merge these two thread ID statements in the future.)

taichi/python/export_lang.cpp
```diff
@@ -828,6 +828,18 @@ void export_lang(py::module &m) {
   m.def("bit_vectorize", BitVectorize);
   m.def("block_dim", BlockDim);
 
+  m.def("insert_thread_idx_expr", [&](){
+    TI_ERROR_IF(get_current_program().config.arch != Arch::cuda,
+            "ti.thread_idx() is only available in cuda context.");
+    auto loop = scope_stack.size() ? scope_stack.back()->list->parent_stmt : nullptr;
+    if(!(loop && loop->is<FrontendForStmt>()))
```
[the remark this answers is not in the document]
Luke-Skycrawler: Appologies for the delay & thanks for the input here. I've changed the runtime function call and class name accordingly.

k-ye (approved):
Great job, LGTM!

Luke-Skycrawler:
Commit `ab29999` has passed all the tests. Reviewers please ignore commit `b04bb86`

python/taichi/lang/__init__.py
```diff
@@ -603,6 +603,7 @@ def loop_unique(val, covers=None):
 vectorize = _ti_core.vectorize
 bit_vectorize = _ti_core.bit_vectorize
 block_dim = _ti_core.block_dim
+thread_idx = _ti_core.insert_thread_idx_expr
```
yolo2themoon, line 606: Thanks a lot! 
Nits after a brief discussion with @k-ye  : 
We prefer to name `ti.thread_idx()` as `ti.global_thread_idx()` to make it self-explaining and avoid divergence from thread_idx() in the codebase.
```suggestion
global_thread_idx = _ti_core.insert_thread_idx_expr
```

Luke-Skycrawler:
advice taken & PR title changed

yolo2themoon (approved):

tests/python/test_thread_idx.py
```diff
@@ -0,0 +1,16 @@
+import taichi as ti
+
+
+@ti.test(arch=[ti.cuda, ti.cpu])
+def test_thread_idx():
```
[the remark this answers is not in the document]
yolo2themoon: Sorry, this comment was sent yesterday but I pended it by mistake 😿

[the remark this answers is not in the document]
yolo2themoon: We have temporarily disabled clang-tidy check #3458, please rebase and merge branch 'master'. 🙈

yolo2themoon:
I’m trying to fix an error introduced by using the undeclared identifier ‘EvalExpression’ 
[Auto Format · taichi-dev/taichi@f8f64e1 (github.com)](https://github.com/taichi-dev/taichi/runs/4176288367?check_suite_focus=true) 

Sorry for messing up your branch.

strongoier:
We are not using `EvalExpression` anymore and it has been removed in #3448. Would you add a new `Expression` in the frontend named `GlobalThreadIndexExpression`?

yolo2themoon:
Just found that, thanks a lot!

taichi/analysis/gather_uniquely_accessed_pointers.cpp
```diff
@@ -34,6 +34,11 @@ class LoopUniqueStmtSearcher : public BasicStmtVisitor {
       loop_unique_[stmt] = stmt->index;
   }
 
+  void visit(GlobalThreadIndexStmt *stmt) override {
```
strongoier, line 37: A thread index is not unique among the top-level loop. Consider the case that one sets the number of cpu threads to 1. So IMHO we shouldn't do this here, and we don't need to record the loop in this statement.

taichi/ir/frontend_ir.h
```diff
@@ -557,6 +557,19 @@ class TensorElementExpression : public Expression {
   }
 };
 
+class GlobalThreadIndexExpression : public Expression {
+ public:
+  Stmt *loop_stmt;
+  GlobalThreadIndexExpression(Stmt *stmt) : loop_stmt(stmt) {
+  }
+
```
strongoier, line 565: Could you also add a `type_check` member function (similar to that in `RandExpression`) which simply returns `Primitive::i32`?

strongoier (commented):
Thanks for the work! There are a few suggestions though :-)

taichi/analysis/gather_uniquely_accessed_pointers.cpp
Luke-Skycrawler: I pressumed marking it as unique in the top level might mean that the statement is optimized in the lower-level loops but preserved in the top-level?

strongoier: `loop_unique` was introduced in #1961 and the intention was to mark/analyze those stmts with definitely different values across loop iterations so that no atomic operations need to involve when those stmts serve as field indices.

taichi/transforms/lower_ast.cpp
```diff
@@ -190,6 +190,10 @@ class LowerAST : public IRVisitor {
     // do nothing
   }
 
+  void visit(GlobalThreadIndexStmt *stmt) override {
```
strongoier, line 193: This can be removed :-)

taichi/transforms/ir_printer.cpp
```diff
@@ -599,6 +599,11 @@ class IRPrinter : public IRVisitor {
           stmt->loop->name());
   }
 
+  void visit(GlobalThreadIndexStmt *stmt) override {
+    print("{}{} = loop {} global thread index", stmt->type_hint(),
+          stmt->name());
```
strongoier, line 604:
```suggestion
    print("{}{} = global thread index", stmt->type_hint(), stmt->name());
```

strongoier (approved):
LGTM! Thanks!

PR #3328, opened 2021-10-30
[OpenGl] Set result_buffer for opengl aot.

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #3329
* __->__ #3328

PR #3329, opened 2021-10-30
[OpenGl] Support preprocessing glsl code in aot.

ailzhang:
* __->__ #3329
* #3328

I've tested this manually but haven't figured out a good way to put this
in the tests yet, e.g. not sure how to force generate code containing a
macro. Suggestions are welcome!

### DIFF
--- a/taichi/backends/opengl/aot_module_builder_impl.cpp
+++ b/taichi/backends/opengl/aot_module_builder_impl.cpp
@@ -1,4 +1,5 @@
 #include "taichi/backends/opengl/aot_module_builder_impl.h"
+#include <stdio.h>
 #include "glad/glad.h"
 
 namespace taichi {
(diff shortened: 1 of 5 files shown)


k-ye:
WDYT if we have a default empty `preprocess()` impl, and override that in the OpenGL AOT builder 🤔 ?

taichi/backends/opengl/aot_module_builder_impl.cpp
```diff
@@ -11,6 +15,23 @@ AotModuleBuilderImpl::AotModuleBuilderImpl(
   aot_data_.root_buffer_size = compiled_structs_.root_size;
 }
 
+void AotModuleBuilderImpl::preprocess_kernels() {
+  for (auto &aot_compiled_kernel : aot_data_.kernels) {
```
k-ye, line 19: nit: leave a TODO saying that we should try to move this to the base class?

k-ye (approved):
LGTM!

Issue #3298, opened 2021-10-27
Make GL_NV_shader_* extensions configurable in the GLSL codegen


k-ye:
Use case: Sometimes it is necessary for us to generate the code for floating atomic adds using the atomicCAS approach, even if the hardware supports native float atomic add (i.e., the `GL_NV_shader_atomic_float` extension is available). We should make these extensions configurable or ignorable.

PR #3331, opened 2021-10-30
[opengl] Provide an option to disable NV extensions during codegen

k-ye:
Related issue = #3298

RIght now the users still don't have a way to disable the usage of extensions during JIT. After https://github.com/taichi-dev/taichi/pull/3329, hopefully we only need to disable this in the AOT module.


ailzhang (approved):
LGTM! 
btw IIUC we'll expose this option to python as a followup?

k-ye:
> LGTM! btw IIUC we'll expose this option to python as a followup?

Yep, we can add a flag to `CompileConfig` or something like that

PR #3330, opened 2021-10-30
[vulkan] Fix GGUI

AmesingFlank:
fixes #3327 

GGUI started crashing since #3253, but that PR wasn't the issue. The real problem is that the `vk_has_external_memory` flag wasn't set properly.

PR #3333, opened 2021-10-31
[gui] GGUI Image IO  (well, it's actually just O...)

AmesingFlank:
This PR adds image IO for GGUI. The implementation is somewhat based on [this](https://github.com/SaschaWillems/Vulkan/blob/master/examples/screenshot/screenshot.cpp).

Notice that this PR alone doesn't enable headlessly running GGUI: it only allows you to dump an image file while the window is also running.

In the next PR, I'll add in headless support. And then we can start adding tests for GGUI.

Not working on MoltenVK on macOS... will try to fix.

And finally, this PR should be merged after #3330, otherwise GGUI won't work at all.

### DIFF
--- a/taichi/backends/device.h
+++ b/taichi/backends/device.h
@@ -242,9 +242,9 @@ struct BufferImageCopyParams {
 };
 
 struct ImageCopyParams {
-  uint32_t width;
-  uint32_t height;
-  uint32_t depth;
+  uint32_t width{1};
+  uint32_t height{1};
+  uint32_t depth{1};
 };
 
 class CommandList {
--- a/taichi/backends/device.h
+++ b/taichi/backends/device.h
@@ -219,7 +219,8 @@ enum class ImageLayout {
   depth_attachment,
   depth_attachment_read,
   transfer_dst,
-  transfer_src
+  transfer_src,
+  present_src
 };
 
 struct BufferImageCopyParams {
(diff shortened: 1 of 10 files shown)


taichi/backends/device.h
```diff
@@ -240,6 +241,12 @@ struct BufferImageCopyParams {
   uint32_t image_layer_count{1};
 };
 
+struct ImageCopyParams {
+  uint32_t width;
```
k-ye, line 245: default init

taichi/backends/device.h
```diff
@@ -303,6 +310,20 @@ class CommandList {
                                const BufferImageCopyParams &params) {
     TI_NOT_IMPLEMENTED
   }
+  virtual void copy_image(DeviceAllocation dst_img,
+                          DeviceAllocation src_img,
+                          ImageLayout dst_img_layout,
+                          ImageLayout src_img_layout,
+                          const ImageCopyParams &params) {
+    TI_NOT_IMPLEMENTED
```
k-ye, line 318: nit: If this is a pure virtual class, make all these methods `=0`?

taichi/backends/device.h
```diff
@@ -421,6 +442,9 @@ class Surface {
   virtual std::pair<uint32_t, uint32_t> get_size() = 0;
   virtual BufferFormat image_format() = 0;
   virtual void resize(uint32_t width, uint32_t height) = 0;
+  virtual DeviceAllocation get_image_data() {
```
k-ye, line 445: ditto

k-ye (commented):
Great work!

taichi/backends/device.h
bobcao3: That might make implementation of each device a little bit more complicated if they don't support something

k-ye (approved):
LGTM!

AmesingFlank:
> LGTM!

help me merge it? :D

k-ye:
3333!

PR #3336, opened 2021-11-01
[ci] Update windows postsubmit job timeout to 90mins.

ailzhang:
This matches the number we have in presubmit. We'll probably use a template for presubmit/postsubmit to avoid divergence in the future.

Related issue = #


k-ye (approved):
LGTM!

PR #3339, opened 2021-11-01
[CI] Linux CD containerization

Leonz5288:
Related issue = #3274

### DIFF
--- a/.github/workflows/scripts/unix_docker_build.sh
+++ b/.github/workflows/scripts/unix_docker_build.sh
@@ -23,7 +23,7 @@ fi
 cd python
 # This is for changelog
 git fetch origin master
-TAICHI_CMAKE_ARGS=$CI_SETUP_CMAKE_ARGS python3 build.py build --project_name $PEOJECT_NAME
+TAICHI_CMAKE_ARGS=$CI_SETUP_CMAKE_ARGS python3 build.py build --project_name $PROJECT_NAME
 # Run basic cpp tests
 cd ..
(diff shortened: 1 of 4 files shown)


Leonz5288:
Tested here: https://github.com/taichi-dev/test_actions/runs/4064868178?check_suite_focus=true
The four linux workflows at the beginning.

Leonz5288:
They did not pass because the PYPI password is missing.

qiao-bo (approved):
Thanks! LGTM!

.github/workflows/scripts/unix_docker_build.sh
```diff
@@ -5,7 +5,8 @@ set -ex
 # Parse ARGs
 PY=$1
 GPU_BUILD=$2
-CI_SETUP_CMAKE_ARGS=$3
+PROJECT_NAME=$3
```
ailzhang, line 8: can we make project_name optional :P (which saves us from specifying it in each pre/post submit jobs)?

ailzhang (approved):
Looks great! Thanks!

.github/workflows/scripts/unix_docker_build.sh
Leonz5288: If we use one script for all, --project-name has to follow a name and cannot be empty... So I don't think this can be optional.

Leonz5288: I'll merge this first so tomorrow's auto release can test this. If we have better idea I'll open another PR :).

PR #3341, opened 2021-11-01
[opengl] Only run preprocess_kernels when glslc is available.

ailzhang:
thanks to @squarefk for reporting the issue!


k-ye (approved):
LGTM!

Issue #3301, opened 2021-10-27
Support frontend type check


strongoier:
**Concisely describe the proposed feature**

Currently, Taichi doesn't have frontend type check. All type related stuff are not handled until Taichi kernels are turned into CHI IRs. This causes the following issues:
1. Type error messages are not readable. It is very hard for users to distinguish between a user error and a compiler internal error, as they always see a bunch of purple stuff. Even for an experienced user, in some cases it is almost impossible to figure out the connections between the reported CHI IR-level errors and the original Taichi kernels.
2. It is not possible to support local variable definition requiring type inference. For example, we may want a local array with only initial values (which may or may not be constants) provided.

To solve these issues, I hope to make Taichi support frontend type check.

**Describe the solution you'd like (if any)**

Add `ret_type` (return type) member variable and `type_check` member function for the `Expression` class (which represents frontend expressions) in Taichi core, and call `type_check` when a Python-side `Expr` is constructed. Note that the type inference/check rules need to be consistent with the `type_check` pass of CHI IR. Our goal is to infer all frontend types and report all type errors at this stage, and the `type_check` pass of CHI IR should only serve the compiler itself.

Progress:

- [x] ArgLoadExpression
- [x] RandExpression
- [x] UnaryOpExpression
- [x] BinaryOpExpression
- [x] TernaryOpExpression
- [x] InternalFuncCallExpression
- [x] ExternalFuncCallExpression
- [x] ExternalTensorExpression
- [x] GlobalVariableExpression
- [x] GlobalPtrExpression
- [x] TensorElementExpression
- [x] <del>EvalExpression</del>
- [x] RangeAssumptionExpression
- [x] LoopUniqueExpression
- [x] IdExpression
- [x] AtomicOpExpression
- [x] SNodeOpExpression
- [x] LocalLoadExpression
- [x] GlobalLoadExpression
- [x] ConstExpression
- [x] ExternalTensorShapeAlongAxisExpression
- [x] FuncCallExpression

PR #3345, opened 2021-11-01
[Lang] Better type error messages

strongoier:
Related issue = #3301

This PR aims at setting up a mechanism to give better type error messages to users. To enable this, this PR also separates `type_check` from `Expression` constructors.

Let's look at a simple example:
```python
import taichi as ti

ti.init()

@ti.kernel
def func():
    a = 1
    b = 3.1
    c = a & b

func()
```

Before this PR, the error message is: (in fact, this is the simplest one I can get in Taichi - others usually involve very long purple stuff)
```
[Taichi] version 0.8.3, llvm 10.0.0, commit 5c1cf297, linux, python 3.8.10
[Taichi] Starting on arch=x64
[W 11/01/21 16:41:23.253 3711558] [type_check.cpp:operator()@308] [$15] Error: bitwise operations can only apply to integral types. at
[W 11/01/21 16:41:23.253 3711558] [type_check.cpp:operator()@310] 
  File "type_test.py", line 9, in func
    c = a & b
  File "/home/xuyi15/taichi/python/taichi/lang/common_ops.py", line 96, in __and__
    return ti.bit_and(self, other)
  File "/home/xuyi15/taichi/python/taichi/lang/ops.py", line 78, in wrapped
    return imp_foo(a, b)
  File "/home/xuyi15/taichi/python/taichi/lang/ops.py", line 63, in imp_foo
    return foo(x, y)

[W 11/01/21 16:41:23.253 3711558] [type_check.cpp:operator()@311] Compilation stopped due to type mismatch.
  File "type_test.py", line 11, in <module>
    func()
  File "/home/xuyi15/taichi/python/taichi/lang/kernel_impl.py", line 796, in wrapped
  File "/home/xuyi15/taichi/python/taichi/lang/kernel_impl.py", line 724, in __call__
  File "/home/xuyi15/taichi/python/taichi/lang/kernel_impl.py", line 682, in func__
    t_kernel(launch_ctx)
RuntimeError: Binary operator type mismatch
```

After this PR, the error message is:
```
[Taichi] version 0.8.4, llvm 10.0.0, commit 8d947831, linux, python 3.8.10
[Taichi] Starting on arch=x64
[Taichi] Compilation failed
  File "type_test.py", line 9, in func
    c = a & b
TaichiTypeError: unsupported operand type(s) for &: 'i32' and 'f32'
```

### DIFF
--- a/.github/workflows/postsubmit.yml
+++ b/.github/workflows/postsubmit.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Build
         run: |
           mkdir -m777 wheel
-          docker create -v `pwd`/wheel:/wheel --user dev --name taichi_build ghcr.io/taichi-dev/taichidev-cpu-ubuntu18.04:v0.1.0 /home/dev/taichi/.github/workflows/scripts/unix_docker_build.sh $PY $GPU_BUILD "$CI_SETUP_CMAKE_ARGS"
+          docker create -v `pwd`/wheel:/wheel --user dev --name taichi_build ghcr.io/taichi-dev/taichidev-cpu-ubuntu18.04:v0.1.0 /home/dev/taichi/.github/workflows/scripts/unix_docker_build.sh $PY $GPU_BUILD $PROJECT_NAME "$CI_SETUP_CMAKE_ARGS"
           # Docker cp preserves the ownership in the host machine. However, the user in the host machine won't be recognized
           # by the docker image we set ahead of time. Therefore, we use tar to set the user of the folder that is copied to the container.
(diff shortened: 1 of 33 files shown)


k-ye (commented):
Very cool! One suggestion: do not use assertions for user errors, let's change to TI_ERROR macros

strongoier:
> Very cool! One suggestion: do not use assertions for user errors, let's change to TI_ERROR macros

We are on our way to give better user error messages, and the new way is to throw `std::runtime_error` in C++ and catch it in Python for re-formatting. However, we should only do these for definite user-errors. For other cases, we stick to assertions to maintain invariants.

k-ye (approved):
LGTM!

PR #3348, opened 2021-11-02
[gui] Headless GGUI

AmesingFlank:
This PR enables headlessly running GGUI. I tested it by sshing into my remote linux machine and running a ggui example within the ssh session.

There is a small bug: IMGUI widgets are not shown in the output images when running in headless mode. I will investigate this.

We can now begin to add tests, and I will submit another PR soon to update the docs to include this feature.


g1n0st (approved):
Awesome! It's really helpful!

bobcao3 (approved):
LGTM

PR #3349, opened 2021-11-02
[misc] Use configured headers to track version and commit hash

sjwsl:
Instead of generating headers manually, use [a more standard way](https://cmake.org/cmake/help/latest/guide/tutorial/A%20Basic%20Starting%20Point.html#adding-a-version-number-and-configured-header-file) to track the version and commit hash. It is easier to maintain and understand now.

Test on my local machine by building different versions.

This is the prerequisite for #3354.


CMakeLists.txt
```diff
@@ -103,14 +91,8 @@ foreach(arch IN LISTS HOST_ARCH CUDA_ARCH)
...
 
-
-FILE(WRITE ${CMAKE_CURRENT_LIST_DIR}/taichi/common/version.h
-        "#pragma once\n"
-        "#define TI_VERSION_MAJOR \"${TI_VERSION_MAJOR}\"\n"
-        "#define TI_VERSION_MINOR \"${TI_VERSION_MINOR}\"\n"
-        "#define TI_VERSION_PATCH \"${TI_VERSION_PATCH}\"\n"
-        "#define TI_CUDAVERSION \"${CUDA_VERSION}\"\n"
-        )
+configure_file(taichi/common/version.h.in version.h)
```
ailzhang, line 94: Can we somehow still keep these headers in taichi/common folder? ;)

ailzhang (commented):
Yesss! Nice work!

CMakeLists.txt
sjwsl: Sure. Which one do you think is better?
1. add `.in` suffix to distinguish the files before and after processing
2. or just replace them in place

ailzhang: @sjwsl let's keep the `.in` suffix so that we know what happened there :D Thanks a lot!

sjwsl: Got it. But I accidentally deleted my local branch so can't update this PR now... Could you help merge this first and then I will give another PR for this. Thanks!

ailzhang: sure!

ailzhang (approved):
LGTM!

Issue #3335, opened 2021-10-31
Improve quality of the C++ code base


sjwsl:
Recently I noticed some compiler warnings while refactoring the C++ codebase. Warnings like unused variables may be irrelevant. But there are some critical ones, for example, non-virtual base class destructor, use of uninitialized variables, and narrowing conversion (#3322). These warnings are not about style or readability but affect the behavior of the code. They may cause performance issues, memory leaks, and in the worst case, undefined behavior. For example #3323, the different behaviors in debug and release mode are most likely due to unknown undefined behaviors.

The quality of the C++ codebase is in urgent need of improvement. The existing CI process is not enough, because `ti test` cannot detect all these errors. We need to firstly use static analysis tools to improve the situation. I have some initial ideas:

1. Add static analysis as a mandatory check in CI. We should discuss which warnings must not exist in the code base and make them fail CI.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
2. Internally maintain a panel of all warnings (except for some really insignificant ones), and deal with all the current warnings as one of follows:
    - Mark as false positive
    - Mark as intentional 
    - Mark as confirmed and fix
3. Organize the result of discussions into a set of code standards, which will serve as a reference for both the R&D team and the community to avoid similar mistakes in the future.

These are just my random thinkings and I'm currently working on the second one based on [Clang Static Analyzer](https://clang-analyzer.llvm.org) and [Clang-tidy](https://clang.llvm.org/extra/clang-tidy). The final goal is to avoid potential problems caused by unsafe code as much as possible. Any comments and ideas are welcome :)

PR #3354, opened 2021-11-02
[ci] Add clang-tidy in CI

sjwsl:
Related issue = #3335 

Add `clang-tidy` as a mandatory stage in CI. Currently only one check (exactly detected one warning in the codebase) is enabled and this PR includes the fix.

This PR must be merged after #3349. The reason is that compilation before depends on the headers (`version.h` and `commit_hash.h`) generated by scripts. Only after that PR `cmake` is able to generate the complete compilation database for `clang-tidy` without `setup.py`.


.github/workflows/presubmit.yml
```diff
@@ -21,6 +21,32 @@ jobs:
         env:
           PR_TITLE: ${{ github.event.pull_request.title }}
 
+  check_clang_tidy:
```
k-ye, line 24: 1. Let's move this below `check_code_format`
2. Use Docker to set up the environment? https://github.com/taichi-dev/taichi/blob/da2fa48f4ff8e9ead8ef7246a506bff597fe3a36/.github/workflows/presubmit.yml#L94-L98

sjwsl: 2. Yes we can. But we can't reuse the existing Dockerfiles because `ti.test` is included. Do we need a more general Dockerfile just to setup the environment?

sjwsl:
#3082 introduced `clang-tidy` to our repo. But I think the current naming checks are only triggered manually. What do you think to unify them into CI so that there is only one `.clang-tidy`? @bobcao3 @k-ye

bobcao3:
Let's leave out the naming check for now. It breaks the code because we have fipped names and things like that, causing a bunch of conflicts. We should focus on reliability/safety related ones first, like virtual/override, return values, value initialization etc.

sjwsl:
> Let's leave out the naming check for now. It breaks the code because we have fipped names and things like that, causing a bunch of conflicts. We should focus on reliability/safety related ones first, like virtual/override, return values, value initialization etc.

Sounds good. But IMO `.clang-tidy` under the root directory should represent "the rules we follow and check in CI", as a guideline for developers. Will it be better to move the manual naming checks to another folder like `scripts/`?

.github/workflows/presubmit.yml
qiao-bo: > 1. Yes we can. But we can't reuse the existing Dockerfiles because `ti.test` is included. Do we need a more general Dockerfile just to setup the environment?

You can try the dockerfile in current ci/ folder. No test or build are included there.

sjwsl: OK, thanks! BTW what is the usage of `ci/scripts/ubuntu_build_test`? I found current CI only invokes scripts under `.github/workflows/scripts.`.

qiao-bo: > OK, thanks! BTW what is the usage of `ci/scripts/ubuntu_build_test`? I found current CI only invokes scripts under `.github/workflows/scripts.`.

that is used to build and test the image generated by the dockerfile in ci folder. For CPU test, we directly use the image generated by that dockerfile (v0.1.0). For GPU test, we added some packages on top of the image built from the Dockerfile  (v0.1.1).

k-ye:
> > Let's leave out the naming check for now. It breaks the code because we have fipped names and things like that, causing a bunch of conflicts. We should focus on reliability/safety related ones first, like virtual/override, return values, value initialization etc.
> 
> Sounds good. But IMO `.clang-tidy` under the root directory should represent "the rules we follow and check in CI", as a guideline for developers. Will it be better to move the manual naming checks to another folder like `scripts/`?

Let's focus on the security issues first, SG if we move the naming checks away later  :-)

.github/workflows/presubmit.yml
```diff
@@ -72,6 +72,28 @@ jobs:
...
+      - name: Get docker images
+        run: |
+          # https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#upgrading-a-workflow-that-accesses-ghcrio
+          echo $CR_PAT | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+          docker pull ghcr.io/taichi-dev/taichidev-cpu-ubuntu18.04:v0.1.0
+        env:
+          CR_PAT: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run clang-tidy
+        run: |
+          docker create --user root --name check_clang_tidy ghcr.io/taichi-dev/taichidev-cpu-ubuntu18.04:v0.1.0 /home/dev/taichi/.github/workflows/scripts/check_clang_tidy.sh "$CI_SETUP_CMAKE_ARGS"
```
qiao-bo, line 91: we need root user here because we need to apt install later? Then, it might be better to install the package in the image.

.github/workflows/scripts/check_clang_tidy.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+CI_SETUP_CMAKE_ARGS=$1
+
+apt install -y clang-tidy
+
+cd taichi
+python3 -m pip install -r requirements_dev.txt
```
qiao-bo, line 8: do we need these packages to run clang-tidy?

qiao-bo (commented):
Thanks!

.github/workflows/scripts/check_clang_tidy.sh
sjwsl: Yep, I found cmake need some of these packages like GitPython.

.github/workflows/presubmit.yml
sjwsl: Huh, actually only clang-tidy binary needs to be installed later. I didn't add to it the base image because all the other tests don't use it. WDYT?

qiao-bo: i guess one package is fine. The risk of allowing root user is higher. I think running clang-tidy should just like build/test on non-root mode.

.github/workflows/scripts/check_clang_tidy.sh
qiao-bo: ok, installing these packages can take quite some time...

.github/workflows/presubmit.yml
sjwsl: IMO all commands take effects only in the container. So I can't get what the risk of allowing root user is.

qiao-bo: a quick google search gives me these ;/

[Rootless mode allows running the Docker daemon and containers as a non-root user to mitigate potential vulnerabilities in the daemon and the container runtime.](https://docs.docker.com/engine/security/rootless/)

[Because anyone who accesses your container running as root can start undesirable processes in it, such as injecting malicious code. And running a process in your container as root makes it possible to change the user id (UID) or group id (GID) when starting the container, which makes your application vulnerable](https://engineering.bitnami.com/articles/why-non-root-containers-are-important-for-security.html#:~:text=Because%20anyone%20who%20accesses%20your,which%20makes%20your%20application%20vulnerable.)

sjwsl:
I have containerized clang-tidy and it works now. Could you @bobcao3 @Leonz5288 help review this again? Thanks a lot!

scripts/run-clang-tidy.py
```diff
@@ -82,6 +82,7 @@ def get_tidy_invocation(f, clang_tidy_binary, checks, tmpdir, build_path,
                         config):
     """Gets a command line for clang-tidy."""
     start = [clang_tidy_binary]
+    start.append('-warnings-as-errors=*')
```
k-ye, line 85: nit: later let's change this to `run_clang_tidy.py`

taichi/transforms/ir_printer.cpp
```diff
@@ -62,7 +62,7 @@ class IRPrinter : public IRVisitor {
 
   void print_raw(std::string f) {
     for (int i = 0; i < current_indent; i++)
-      f = "  " + f;
+      f.insert(0, "  ");
```
k-ye, line 65: is this change relevant?

k-ye (approved):
LGTM!

scripts/run-clang-tidy.py
sjwsl: Agreed!

taichi/transforms/ir_printer.cpp
sjwsl: This is the only warning in our codebase detected by current rules. It is used to test whether CI result changes as expected.

qiao-bo (approved):
Thanks!

PR #3355, opened 2021-11-02
[IR] Add type_check for UnaryOpExpression

strongoier:
Related issue = #3301

Now for
```python
import taichi as ti

ti.init()

@ti.kernel
def func():
    a = 1
    b = ti.ceil(a)

func()
```

The error message is:
```
[Taichi] version 0.8.4, llvm 10.0.0, commit d577b876, linux, python 3.8.10
[Taichi] Starting on arch=x64
[Taichi] Compilation failed
  File "type_test.py", line 8, in func
    b = ti.ceil(a)
TaichiTypeError: 'ceil' takes real inputs only, however 'i32' is provided
```

### DIFF
--- a/tests/cpp/ir/frontend_type_inference_test.cpp
+++ b/tests/cpp/ir/frontend_type_inference_test.cpp
@@ -47,5 +47,17 @@ TEST(FrontendTypeInference, BinaryOp) {
   EXPECT_EQ(truediv_f64->ret_type, PrimitiveType::f64);
 }
 
+TEST(FrontendTypeInference, UnaryOp) {
+  auto const_i16 = Expr::make<ConstExpression, int16>(-(1 << 10));
+  const_i16->type_check();
+  EXPECT_EQ(const_i16->ret_type, PrimitiveType::i16);
+  auto cast_i8 = cast(const_i16, PrimitiveType::i8);
+  cast_i8->type_check();
+  EXPECT_EQ(cast_i8->ret_type, PrimitiveType::i8);
+  auto bit_not_i16 = ~const_i16;
+  bit_not_i16->type_check();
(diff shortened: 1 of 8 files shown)


tests/cpp/ir/frontend_type_inference_test.cpp
```diff
@@ -47,5 +47,17 @@ TEST(FrontendTypeInference, BinaryOp) {
   EXPECT_EQ(truediv_f64->ret_type, PrimitiveType::f64);
 }
 
+TEST(FrontendTypeInference, UnaryOp) {
```
k-ye, line 50: FYI there's an `EXPECT_THROW` checker :-) https://google.github.io/googletest/reference/assertions.html#EXPECT_THROW

k-ye (approved):
LGTM!

tests/cpp/ir/frontend_type_inference_test.cpp
strongoier: Wow great! I'm just thinking about whether it is better to put the error checks in Python tests or C++ tests. Putting them in Python tests can directly give some illegal examples, while putting them in C++ tests can make `type_check` tests more uniform. WDYT?

ailzhang: Haha this is a hard one, I'm more used to putting fine grained tests in C++ and keep python tests sparse but representable. (sorry I know this is vague...

ailzhang (approved):
Looks great!

tests/cpp/ir/frontend_type_inference_test.cpp
strongoier: > putting fine grained tests in C++ and keep python tests sparse but representable

Sounds good. Will follow this in future PRs.

PR #3358, opened 2021-11-02
[opengl] Off-screen context using EGL & Support GLES

bobcao3:
Related issue = https://github.com/taichi-dev/taichi/issues/3364

GLES can be enabled by changing the flag in the `opengl_api.cpp`, and due to the need of GLSL preprocessing, it's only available on linux / BSD. (well other platforms don't really support GLES 3.1 anyways)

GLAD2 is introduced so that we can use its EGL loader to create our off-screen context. GLFW will be attempted first, if GLFW can not initialize a window, the code will try EGL. Therefore this does not introduce extra dependencies.

### DIFF
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -58,6 +58,47 @@ std::string opengl_atomic_op_type_cap_name(AtomicOpType type) {
   return type_names[type];
 }
 
+#if !defined(TI_PLATFORM_WINDOWS)
+#include<sys/wait.h>
+#endif
+
+std::string preprocess_kernel(std::string src) {
+#if !defined(TI_PLATFORM_WINDOWS)
+  int child_status;
+  auto child_pid = fork();
+  if(child_pid == 0) {
+    char* const argv[] = {"glslc", "--version", nullptr};
+    execvp("glslc", argv);
+  } else {
+    pid_t tpid = waitpid(child_pid, &child_status, 0);
+  }
+  if (!child_status) {
+    // For debugging only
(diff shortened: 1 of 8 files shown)


taichi/backends/opengl/opengl_api.cpp
```diff
@@ -29,6 +30,8 @@ namespace opengl {
 int opengl_max_block_dim = 1024;
 int opengl_max_grid_dim = 1024;
 
+constexpr bool use_gles = false;
```
k-ye, line 33: Note that we need to make this user configurable later

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -58,6 +58,49 @@ std::string opengl_atomic_op_type_cap_name(AtomicOpType type) {
...
 }
 
+#if !defined(TI_PLATFORM_WINDOWS)
+#include<sys/wait.h>
+#endif
+
+std::string preprocess_kernel(std::string src) {
+  if (!is_gles()) return src;
+
+#if !defined(TI_PLATFORM_WINDOWS)
```
k-ye, line 68: Huh, let's probably undo https://github.com/taichi-dev/taichi/pull/2486 to avoid using macros or `glslc` @ailzhang  ?

ailzhang (approved):
Btw I've tried the EGL commit and it works nicely on my workstation! Thanks a lot! 
Opened an issue for reverting the macros here https://github.com/taichi-dev/taichi/issues/3360 
I'll get to it later today/tmr if @bobcao3 haven't ;)

k-ye (approved):
Great job LGTM!

PR #3365, opened 2021-11-03
[Test] Fix uninitialized tests for ndarray

qiao-bo:
Did not fail before because Torch memory is initialized to zero.

### DIFF
--- a/tests/python/test_ndarray.py
+++ b/tests/python/test_ndarray.py
@@ -303,7 +303,24 @@ def test_matrix_ndarray_taichi_scope_struct_for(layout):
     _test_matrix_ndarray_taichi_scope_struct_for(layout)
 
 
-def _test_vector_ndarray_python_scope(layout):
+@pytest.mark.parametrize('layout', layouts)
+@pytest.mark.skipif(not ti.has_pytorch(), reason='Pytorch not installed.')
+@ti.test(exclude=ti.opengl)
+def test_vector_ndarray_python_scope_torch(layout):
+    a = ti.Vector.ndarray(10, ti.i32, 5, layout=layout)
+    for i in range(5):
+        for j in range(4):
+            a[i][j * j] = j * j
+    assert a[0][6] == 0 # torch memory initialized to zero
+    assert a[1][0] == 0
+    assert a[2][1] == 1
+    assert a[3][4] == 4
+    assert a[4][9] == 9
+
+
+@pytest.mark.parametrize('layout', layouts)
+@ti.test(arch=[ti.cpu, ti.cuda], ndarray_use_torch=False)
+def test_vector_ndarray_python_scope(layout):
     a = ti.Vector.ndarray(10, ti.i32, 5, layout=layout)
     for i in range(5):
         for j in range(4):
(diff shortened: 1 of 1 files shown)


strongoier (commented):
PyTorch-based implementation will initialize ndarray to 0. Shall we keep the same behavior?

qiao-bo:
> PyTorch-based implementation will initialize ndarray to 0. Shall we keep the same behavior?

I don't think these two tests intend to test that (at least is not obvious by name). These will fail after we switched to DeviceAllocation since there is no initialization here. I don't think we should assume ndarray is initialized unless users explicitly do so.

qiao-bo:
maybe we can add a test for uninitialized memory?

strongoier:
> @strongoier maybe we can add a test for uninitialized memory?

In that case we may not get deterministic behavior. How can we ensure it is *uninitialized*?

qiao-bo:
> > @strongoier maybe we can add a test for uninitialized memory?
> 
> In that case we may not get deterministic behavior. How can we ensure it is _uninitialized_?

Actually, my point is that these two tests create confusion at this moment. We have other tests in test_ndarray where first the ndarray is filled/initialized, then the corresponding positions are checked. In these two tests, we have one assert that checks a location where we did not fill/initialize explicitly. This might be OK for torch, but for our own allocation we are asserting an uninitialized value. is this what we want in these tests?

k-ye:
> > > @strongoier maybe we can add a test for uninitialized memory?
> > 
> > 
> > In that case we may not get deterministic behavior. How can we ensure it is _uninitialized_?
> 
> Actually, my point is that these two tests create confusion at this moment. We have other tests in test_ndarray where first the ndarray is filled/initialized, then the corresponding positions are checked. In these two tests, we have one assert that checks a location where we did not fill/initialize explicitly. This might be OK for torch, but for our own allocation we are asserting an uninitialized value. is this what we want in these tests?

I think both are valid points. Note that Taichi field uses zero as the default value, so it's consistent to keep it that way. That said, this test itself should only check for what it has written. We can have a separate test just for uninitialized Ndarrays. WDYT?

qiao-bo:
> > > > @strongoier maybe we can add a test for uninitialized memory?
> > > 
> > > 
> > > In that case we may not get deterministic behavior. How can we ensure it is _uninitialized_?
> > 
> > 
> > Actually, my point is that these two tests create confusion at this moment. We have other tests in test_ndarray where first the ndarray is filled/initialized, then the corresponding positions are checked. In these two tests, we have one assert that checks a location where we did not fill/initialize explicitly. This might be OK for torch, but for our own allocation we are asserting an uninitialized value. is this what we want in these tests?
> 
> I think both are valid points. Note that Taichi field uses zero as the default value, so it's consistent to keep it that way. That said, this test itself should only check for what it has written. We can have a separate test just for uninitialized Ndarrays. WDYT?

I agree. How about we do the following: First, we keep the test, but rename it to something like `_test_vector_ndarray_python_scope_torch`, and after this assert we add a comment says `// this assert checks the default initialized zero value`. Then, we put another test with the [0][6] fixed. The no-torch-version should use this test.

strongoier:
> > > > > @strongoier maybe we can add a test for uninitialized memory?
> > > > 
> > > > 
> > > > In that case we may not get deterministic behavior. How can we ensure it is _uninitialized_?
> > > 
> > > 
> > > Actually, my point is that these two tests create confusion at this moment. We have other tests in test_ndarray where first the ndarray is filled/initialized, then the corresponding positions are checked. In these two tests, we have one assert that checks a location where we did not fill/initialize explicitly. This might be OK for torch, but for our own allocation we are asserting an uninitialized value. is this what we want in these tests?
> > 
> > 
> > I think both are valid points. Note that Taichi field uses zero as the default value, so it's consistent to keep it that way. That said, this test itself should only check for what it has written. We can have a separate test just for uninitialized Ndarrays. WDYT?
> 
> I agree. How about we do the following: First, we keep the test, but rename it to something like `_test_vector_ndarray_python_scope_torch`, and after this assert we add a comment says `// this assert checks the default initialized zero value` (we can also create a separate test for this). Then, we put another test with the [0][6] fixed. The no-torch-version should use this test.

Sounds good.

strongoier (approved):
LGTM!

PR #3366, opened 2021-11-03
[lang] [refactor] Use DeviceAllocation for Ndarray

qiao-bo:
In this PR we first replace the runtime-based memory allocator with device allocation. Only the basic allocate/deallocate functionalities for now.


taichi/llvm/llvm_program.cpp
```diff
@@ -473,6 +473,13 @@ void LlvmProgramImpl::finalize() {
     cuda_device()->dealloc_memory(preallocated_device_buffer_alloc);
   }
 #endif
+  for (auto &alloc : ndarray_allocs_) {
+    if (config->arch == Arch::cuda) {
+      cuda_device()->dealloc_memory(alloc);
```
k-ye, line 478: Here and below: it seems that Device is the only difference in many of the function definitions. Consider de-duplicating the code w/ sth like this:

```cpp
auto * device = (arch == cuda) ? cuda_device() : cpu_device();
// use `device`
```

taichi/program/ndarray.cpp
```diff
@@ -11,22 +11,15 @@ Ndarray::Ndarray(Program *prog,
...
-      prog->get_llvm_program_impl()->get_llvm_context(prog->config.arch);
-  auto *const runtime_jit = tlctx->runtime_jit_module;
-  TI_TRACE("allocating memory for Ndarray");
-  runtime_jit->call<void *, std::size_t, std::size_t>(
-      "runtime_memory_allocate_aligned",
-      prog->get_llvm_program_impl()->get_llvm_runtime(),
-      nelement_ * element_size_, taichi_page_size);
-  data_ptr_ = prog->get_llvm_program_impl()->fetch_result<uint64_t *>(
-      taichi_result_buffer_runtime_query_id, prog->result_buffer);
+  LlvmProgramImpl *prog_impl_ = prog->get_llvm_program_impl();
```
k-ye, line 19:
```suggestion
  LlvmProgramImpl *prog_impl = prog->get_llvm_program_impl();
```

k-ye (commented):
Cool! Let's try to dedupe the code here

taichi/llvm/llvm_program.cpp
```diff
@@ -564,5 +571,39 @@ DevicePtr LlvmProgramImpl::get_snode_tree_device_ptr(int tree_id) {
...
+  if (config->arch == Arch::cuda) {
+#if defined(TI_WITH_CUDA)
+    Device::AllocParams device_buffer_alloc_params;
+    device_buffer_alloc_params.size = alloc_size;
+    DeviceAllocation alloc =
+        cuda_device()->allocate_memory(device_buffer_alloc_params);
+    ndarray_allocs_.push_back(alloc);
+    return alloc;
+#else
+    TI_NOT_IMPLEMENTED
```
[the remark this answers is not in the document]
qiao-bo: @k-ye maybe we can have a separate issue to address this as a whole? The macro itself is defined with a `;` [here](https://github.com/taichi-dev/taichi/blob/22fd5abd6c513860a0e3f46adb12ef2aff0119c1/taichi/common/logging.h#L108). A grep in our code base shows lots of occurrences go without `;`. '/
IIUC, macros should not have `;` in definition, unless they are sure to be used alone as a statement.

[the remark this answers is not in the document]
k-ye: SG

taichi/llvm/llvm_program.cpp
```diff
@@ -564,5 +567,26 @@ DevicePtr LlvmProgramImpl::get_snode_tree_device_ptr(int tree_id) {
...
+  ndarray_allocs_.push_back(alloc);
+  return alloc;
+}
+
+uint64_t *LlvmProgramImpl::get_ndarray_alloc_info_ptr(DeviceAllocation &alloc) {
+  if (config->arch == Arch::cuda) {
+#if defined(TI_WITH_CUDA)
+    return (uint64_t *)cuda_device()->get_alloc_info(alloc).ptr;
+#else
+    TI_NOT_IMPLEMENTED
```
ailzhang, line 585: I'm always curious if this if-else is necessary :P When do we have arch== cuda but TI_WITH_CUDA not enabled? (In this case I thought we already fallback to cpu?

ailzhang (approved):
Looks great!

taichi/llvm/llvm_program.cpp
qiao-bo: when someone wrote ti.init(arch=cuda) but build Taichi with TAICHI_CMAKE_ARGS="-DTI_WITH_CUDA:BOOL=OFF"? ;)))

strongoier (approved):
LGTM!

taichi/llvm/llvm_program.cpp
ailzhang: In practice this never happens since we build our wheels using TI_WITH_CUDA=ON? hmmm I'm just a bit confused :P

ailzhang: (fyi we figured out this macro is indeed turned off on mac so nvm :D

Issue #3360, opened 2021-11-03
revert https://github.com/taichi-dev/taichi/pull/2486


ailzhang:
https://github.com/taichi-dev/taichi/pull/2486

Macros are not supported on GLES and we had to hack around it in https://github.com/taichi-dev/taichi/pull/3358/files. We might prefer to revert back to the version not using macros. 

I gave this a quick try using git revert but it (unsurprisingly) has conflicts. So we'll have to revert this manually.

ailzhang:
Closing via #3369 and #3376, thanks @k-ye !

PR #3369, opened 2021-11-03
[opengl] Do not use macros in GLSL codegen

k-ye:
Related issue = #3360


taichi/backends/opengl/shaders/atomics_macro_f64.glsl.h
```diff
@@ -1,54 +0,0 @@
-// vim: ft=glsl
```
ailzhang, line 1: lol is this file not used at all?

ailzhang (approved):
Awesome thanks!

taichi/backends/opengl/shaders/atomics_macro_f64.glsl.h
k-ye: nope 😂

PR #3370, opened 2021-11-03
[Lang] Define __str__ method for DataType

### LINKED ISSUE #3367: Properly support printing taichi data type string
I'd expect this to print `ti.u16` instead of the object info
```
>>> str(ti.u16)
'<taichi.lib.taichi_core.DataType object at 0x7f8af4a9b4f0>'
```
This will make composing error message string much easier. 
(note this information can be retrieved by `ti.u16.to_string()` but this is not very pythonic. :P

knight42:
Signed-off-by: knight42 <anonymousknight96@gmail.com>
fixes #3367

### DIFF
--- a/taichi/python/export_lang.cpp
+++ b/taichi/python/export_lang.cpp
@@ -103,6 +103,7 @@ void export_lang(py::module &m) {
       .def(py::self == py::self)
       .def("__hash__", &DataType::hash)
       .def("to_string", &DataType::to_string)
+      .def("__str__", &DataType::to_string)
       .def(
           "get_ptr", [](DataType *dtype) -> Type * { return *dtype; },
           py::return_value_policy::reference)
--- a/tests/python/test_basics.py
+++ b/tests/python/test_basics.py
@@ -116,3 +116,12 @@ def test_while_global_load():
     run()
 
     assert y[None] == 3
+
+
+@ti.test()
+def test_datatype_string():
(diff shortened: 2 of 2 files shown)


taichi/python/export_lang.cpp
```diff
@@ -103,6 +103,7 @@ void export_lang(py::module &m) {
       .def(py::self == py::self)
       .def("__hash__", &DataType::hash)
       .def("to_string", &DataType::to_string)
+      .def("__str__", &DataType::to_string)
```
yihong0618, line 106: do we need add `__repr__` too?

yihong0618 (approved):

taichi/python/export_lang.cpp
knight42: SGTM!

ailzhang: `__repr__` is more like for debugging purpose which object information. 
I usually think in this way (maybe incorrect :P):
information from `__str__` should be human readable (but can lose info)
But you can use information from `__repr__` to reconstruct this object easily. 
So IMHO only handling `__str__` is good enough for now and we might consider adding `__repr__` in the future :D

ailzhang (approved):
Nice thanks!

ailzhang (approved):
Thanks!!

yihong0618:
cool

PR #3371, opened 2021-11-03
[Lang] User-friendly exception when copying between ti.field

ljjsalt:
Related issue = #2846

Test
```
x = ti.field(ti.f32, (2, 3))
y = ti.field(ti.f32, (2, 3, 4))
x.copy_from(y)
```

Result
```
  File "/Users/ohm/OSS/taichi/copy_from_test.py", line 7, in <module>
    x.copy_from(y)
  File "/Users/ohm/OSS/taichi/python/taichi/lang/util.py", line 207, in wrapped
    return func(*args, **kwargs)
  File "/Users/ohm/OSS/taichi/python/taichi/lang/field.py", line 162, in copy_from
    raise ValueError(f"ti.field dimension {len(self.shape)} does not match"
ValueError: ti.field dimension 2 does not match the source field dimension 3
```

### DIFF
--- a/python/taichi/lang/ir_builder.py
+++ b/python/taichi/lang/ir_builder.py
@@ -30,12 +30,8 @@ class IRBuilder(Builder):
         # The variable is introduced to support chained assignments.
         # Ref https://github.com/taichi-dev/taichi/issues/2659.
         for node_target in node.targets:
-            if isinstance(node_target, ast.Tuple):
-                IRBuilder.build_assign_unpack(ctx, node_target, node.value.ptr,
-                                              is_static_assign)
-            else:
-                IRBuilder.build_assign_basic(ctx, node_target, node.value.ptr,
-                                             is_static_assign)
+            IRBuilder.build_assign_unpack(ctx, node_target, node.value.ptr,
+                                          is_static_assign)
         return node
 
     @staticmethod
@@ -50,7 +46,9 @@ class IRBuilder(Builder):
             values: A node/list representing the values.
             is_static_assign: A boolean value indicating whether this is a static assignment
         """
-
+        if not isinstance(node_target, ast.Tuple):
+            return IRBuilder.build_assign_basic(ctx, node_target, values,
+                                                is_static_assign)
         targets = node_target.elts
         tmp_tuple = values if is_static_assign else ti.expr_init_list(
             values, len(targets))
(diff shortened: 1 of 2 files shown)


yihong0618 (approved):
LGTM

tests/python/test_field.py
```diff
@@ -99,6 +99,37 @@ def test_scalr_field_from_numpy_with_mismatch_shape(dtype, shape):
...
+def test_field_copy_from(dtype, shape):
+    x = ti.field(dtype, shape)
+    other = ti.field(dtype, shape)
+    x.copy_from(other)
+
+
+@pytest.mark.parametrize('dtype', data_types)
+@pytest.mark.parametrize('shape', field_shapes)
+@ti.test(arch=ti.get_host_arch_list())
+def test_field_copy_from_with_non_field_object(dtype, shape):
```
ailzhang, line 114: for these two error checking tests you don't have to parametrize the tests ;) Simply test one case is good enough.

tests/python/test_field.py
```diff
@@ -99,6 +99,37 @@ def test_scalr_field_from_numpy_with_mismatch_shape(dtype, shape):
         x.from_numpy(arr)
 
 
+@pytest.mark.parametrize('dtype', data_types)
+@pytest.mark.parametrize('shape', field_shapes)
+@ti.test(arch=ti.get_host_arch_list())
+def test_field_copy_from(dtype, shape):
+    x = ti.field(dtype, shape)
+    other = ti.field(dtype, shape)
```
ailzhang, line 107: can we also initialize `other` differently and assert on the copied `x`'s value?

ailzhang (commented):
LGTM!

tests/python/test_field.py
ljjsalt: Thanks for reviewing, It will be removed.

tests/python/test_field.py
ljjsalt: Since copying between different shapes is allowed, how do I assert the copied value? Comparing the overlap? 

And `field.copy_from` calls `taichi.lang.meta.tensor_to_tensor` to perform the copy, maybe it's better to do that test in `meta`'s test file.

ailzhang: oh good point! Sorry I just noticed that we allow copying from a different shape hmmm...yea in that case it's even nicer to have tests covering this. You can simply assert what's the expected `x` value directly. Thanks a lot!

ljjsalt: Hi Ail,

Can you confirm if copying from a different shape is allowed in ```copy_from```? 
The original code only assert the length of the shape is equal, but the comment says 
> The shape of the other field needs to be the same as `self`.

And if it's allowed, I am confused about the intended usage of ```copy_from```.

```
>> x = ti.field(ti.f32, (5,5))
>> y = ti.field(ti.f64, (4,3))
>> y.fill(1)
>> x.copy_from(y); x.to_numpy()
array([[1., 1., 1., 0., 1.],
       [1., 1., 1., 0., 1.],
       [1., 1., 1., 0., 1.],
       [1., 1., 1., 0., 1.],
       [1., 1., 1., 0., 1.]], dtype=float32)
```
It's not simply padded with 0.

ailzhang: Hi @ljjsalt I'd prefer to ban this behavior as long as it passes tests ;) (in other words, nobody depends on this behavior).
I do agree that it's confusing to allow copying from a different shape.

ailzhang (approved):
Nice work, thanks! Welcome to taichi community!

ailzhang:
Oops @ljjsalt I'm really sorry, I tried to rebase your branch on top of master to pass CI but seemed to have messed it up, would you mind reopening a new PR for it?

ljjsalt:
@ailzhang 
No worries, I would like to reopen a new PR and see what's going on.

PR #3373, opened 2021-11-04
[llvm] Basic f16 support

ailzhang:
* #3384
* #3374
* __->__ #3373

This PR supports f16 on unary & binary ops, RandStmt and PrintStmt on
CPU & CUDA.
Milestone here is we now support running fractal simulation on f16.

I'll write some documentation for fp16 support in a bit.

Related issue = https://github.com/taichi-dev/taichi/issues/2992

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -19,7 +19,7 @@ from taichi.lang.util import (cook_dtype, has_pytorch, is_taichi_class,
                               python_scope, taichi_scope, to_pytorch_type)
 from taichi.misc.util import deprecated, get_traceback, warning
 from taichi.snode.fields_builder import FieldsBuilder
-from taichi.type.primitive_types import f32, f64, i32, i64, u32, u64
+from taichi.type.primitive_types import f16, f32, f64, i32, i64, u32, u64
 
 import taichi as ti
 
@@ -296,7 +296,7 @@ class PyTaichi:
         return len(self.compiled_functions) + len(self.compiled_grad_functions)
 
     def set_default_fp(self, fp):
-        assert fp in [f32, f64]
+        assert fp in [f16, f32, f64]
         self.default_fp = fp
         default_cfg().default_fp = self.default_fp
 
@@ -427,6 +427,9 @@ def make_constant_expr(val):
             return Expr(_ti_core.make_const_expr_f32(val))
         elif pytaichi.default_fp == f64:
             return Expr(_ti_core.make_const_expr_f64(val))
+        elif pytaichi.default_fp == f16:
+            # Use f32 to interact with python
+            return Expr(_ti_core.make_const_expr_f32(val))
         else:
             assert False
     else:
--- a/python/taichi/type/primitive_types.py
+++ b/python/taichi/type/primitive_types.py
@@ -18,7 +18,7 @@ f64 = float64
 float16 = ti_core.DataType_f16
 f16 = float16
 
-real_types = [f32, f64, float]
+real_types = [f16, f32, f64, float]
 real_type_ids = [id(t) for t in real_types]
 
 # Integer types
@@ -68,6 +68,7 @@ __all__ = [
     'f32',
     'float64',
     'f64',
+    'float16',
     'f16',
     'int8',
     'i8',
--- a/python/taichi/type/primitive_types.py
+++ b/python/taichi/type/primitive_types.py
@@ -15,6 +15,9 @@ f64 = float64
 """Alias for :const:`~taichi.type.primitive_types.float64`
 """
 
+float16 = ti_core.DataType_f16
+f16 = float16
+
 real_types = [f32, f64, float]
 real_type_ids = [id(t) for t in real_types]
 
@@ -65,6 +68,7 @@ __all__ = [
     'f32',
     'float64',
     'f64',
+    'f16',
     'int8',
     'i8',
     'int16',
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -191,7 +191,7 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
     // functions from libdevice
     auto input = llvm_val[stmt->operand];
     auto input_taichi_type = stmt->operand->ret_type;
-    if (stmt->operand->ret_type->is_primitive(PrimitiveTypeID::f16)) {
+    if (input_taichi_type->is_primitive(PrimitiveTypeID::f16)) {
       // Promote to f32 since we don't have f16 support for extra unary ops in
       // libdevice.
       input =
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -162,7 +162,8 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
 
         auto value_type = tlctx->get_data_type(arg_stmt->ret_type);
         auto value = llvm_val[arg_stmt];
-        if (arg_stmt->ret_type->is_primitive(PrimitiveTypeID::f32)) {
+        if (arg_stmt->ret_type->is_primitive(PrimitiveTypeID::f32) ||
+            arg_stmt->ret_type->is_primitive(PrimitiveTypeID::f16)) {
           value_type = tlctx->get_data_type(PrimitiveType::f64);
           value = builder->CreateFPExt(value, value_type);
         }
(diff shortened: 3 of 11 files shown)


taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -190,6 +191,14 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
     // functions from libdevice
     auto input = llvm_val[stmt->operand];
     auto input_taichi_type = stmt->operand->ret_type;
+    if (stmt->operand->ret_type->is_primitive(PrimitiveTypeID::f16)) {
```
k-ye, line 194:
```suggestion
    if (input_taichi_type->is_primitive(PrimitiveTypeID::f16)) {
```

taichi/codegen/codegen_llvm.cpp
```diff
@@ -162,13 +162,28 @@ void CodeGenLLVM::visit(AllocaStmt *stmt) {
 }
 
 void CodeGenLLVM::visit(RandStmt *stmt) {
-  llvm_val[stmt] = create_call(
-      fmt::format("rand_{}", data_type_name(stmt->ret_type)), {get_context()});
+  if (stmt->ret_type->is_primitive(PrimitiveTypeID::f16)) {
+    // Promoting to f32 since there's no rand_f16 support in runtime.cpp.
+    llvm_val[stmt] = create_call("rand_f32", {get_context()});
```
k-ye, line 167: nit: use a local var to hold this intermediary result?

taichi/codegen/codegen_llvm.cpp
```diff
@@ -162,13 +162,28 @@ void CodeGenLLVM::visit(AllocaStmt *stmt) {
...
+    llvm_val[stmt] =
+        create_call(fmt::format("rand_{}", data_type_name(stmt->ret_type)),
+                    {get_context()});
+  }
 }
 
 void CodeGenLLVM::emit_extra_unary(UnaryOpStmt *stmt) {
   auto input = llvm_val[stmt->operand];
   auto input_taichi_type = stmt->operand->ret_type;
+  if (stmt->operand->ret_type->is_primitive(PrimitiveTypeID::f16)) {
```
k-ye, line 180:
```suggestion
  if (input_taichi_type->is_primitive(PrimitiveTypeID::f16)) {
```

taichi/codegen/codegen_llvm.cpp
```diff
@@ -653,8 +625,83 @@ void CodeGenLLVM::visit(BinaryOpStmt *stmt) {
...
+    }
+    if (stmt->lhs->ret_type->is_primitive(PrimitiveTypeID::f16)) {
+      rhs = builder->CreateFPExt(rhs, llvm::Type::getFloatTy(*llvm_context));
+    }
+    if (ret_type->is_primitive(PrimitiveTypeID::f16)) {
+      ret_type = PrimitiveType::f32;
+    }
+
+    if (op == BinaryOpType::atan2) {
+      if (arch_is_cpu(current_arch())) {
```
k-ye, line 644: We should separate these into codegen_cpu and codegen_cuda in the future...

k-ye (approved):
Great job, thanks!

taichi/codegen/codegen_llvm.cpp
ailzhang: Agreed! Opened an issue at https://github.com/taichi-dev/taichi/issues/3378

python/taichi/lang/impl.py
```diff
@@ -427,6 +427,9 @@ def make_constant_expr(val):
             return Expr(_ti_core.make_const_expr_f32(val))
         elif pytaichi.default_fp == f64:
             return Expr(_ti_core.make_const_expr_f64(val))
+        elif pytaichi.default_fp == f16:
+            # Use f32 to interact with python
```
strongoier, line 431: This will affect the type system because the constant is marked as `f32` in this implementation. Are we only using `float16` as storage types and using at least `float32` for all calculations? If so, this is fine.

python/taichi/type/primitive_types.py
```diff
@@ -15,6 +15,9 @@
 """Alias for :const:`~taichi.type.primitive_types.float64`
 """
 
+float16 = ti_core.DataType_f16
+f16 = float16
+
 real_types = [f32, f64, float]
```
strongoier, line 21:
```suggestion
real_types = [f16, f32, f64, float]
```

python/taichi/type/primitive_types.py
```diff
@@ -65,6 +68,7 @@
     'f32',
     'float64',
     'f64',
+    'f16',
```
strongoier, line 71:
```suggestion
    'float16',
    'f16',
```

sjwsl (approved):
LGTM!

taichi/program/kernel.cpp
```diff
@@ -195,6 +195,9 @@ void Kernel::LaunchContextBuilder::set_arg_float(int arg_id, float64 d) {
     ctx_->set_arg(arg_id, (uint32)d);
   } else if (dt->is_primitive(PrimitiveTypeID::u64)) {
     ctx_->set_arg(arg_id, (uint64)d);
+  } else if (dt->is_primitive(PrimitiveTypeID::f16)) {
```
strongoier, line 198: Should the same modification happen in `Kernel::LaunchContextBuilder::set_arg_int`?

taichi/codegen/codegen_llvm.cpp
```diff
@@ -653,8 +625,83 @@ void CodeGenLLVM::visit(BinaryOpStmt *stmt) {
...
-    TI_NOT_IMPLEMENTED
+    // This branch contains atan2 and pow which use runtime.cpp function for
+    // **real** type. We don't have f16 support there so promoting to f32 is
+    // necessary.
+    llvm::Value *lhs = llvm_val[stmt->lhs];
+    llvm::Value *rhs = llvm_val[stmt->rhs];
+    if (stmt->lhs->ret_type->is_primitive(PrimitiveTypeID::f16)) {
+      lhs = builder->CreateFPExt(lhs, llvm::Type::getFloatTy(*llvm_context));
+    }
+    if (stmt->lhs->ret_type->is_primitive(PrimitiveTypeID::f16)) {
```
strongoier, line 636:
```suggestion
    if (stmt->rhs->ret_type->is_primitive(PrimitiveTypeID::f16)) {
```

strongoier (commented):
Have you tested if the memory usage can be reduced to half with f16?

taichi/codegen/codegen_llvm.cpp
```diff
@@ -407,13 +440,16 @@ void CodeGenLLVM::visit(UnaryOpStmt *stmt) {
...
   UNARY_INTRINSIC(floor)
   UNARY_INTRINSIC(ceil)
-  else emit_extra_unary(stmt);
+  else {
+    emit_extra_unary(stmt);
+  }
 #undef UNARY_INTRINSIC
 }
 
 void CodeGenLLVM::visit(BinaryOpStmt *stmt) {
```
sjwsl, line 449: Should we add a helper function like `CreateBinaryOp(BinaryOpType op, bool is_real)` to get rid of these tons of if/else? And seems this whole file is full of if/else...

python/taichi/lang/impl.py
ailzhang: float16 is a storage type, but also a computation type for some ops e.g. add/mul on cuda. It's indeed a bit tricky since we don't have a native f16 type in python/c++. We can discuss this later, fp16 might need a bit tweak here. :P

taichi/codegen/codegen_llvm.cpp
ailzhang: nice catch! thanks!

taichi/codegen/codegen_llvm.cpp
ailzhang: Yup agreed! We should definitely refactor this, it indeed created some trouble with all the branches when working on this

sjwsl: I have opened issue #3382 for this. Thanks!

taichi/program/kernel.cpp
ailzhang: @strongoier that was a bit confusing that we need to handle f32/f64 in set_arg_int... I did a quick experiment by removing them and tests passed. So I created #3384 to see if we can just remove them :P

strongoier: My understanding is that, `set_arg_int` is passing an int into the kernel, while the kernel might have a f32 argument in its signature, so we need to cast the type here.

ailzhang: @strongoier in that case `set_arg_float` will be called right? https://github.com/taichi-dev/taichi/blob/master/python/taichi/lang/kernel_impl.py#L591-L599

ailzhang:
> Have you tested if the memory usage of fields or ndarrays can be reduced to half with f16? 
@strongoier Good question! Not yet ;) , the way I verified the correctness of this PR is by looking at the generated llvm ir and ptx assembly, making sure the corresponding computation is done on half type :D We'll do some more memory usage  & speed check later.

taichi/program/kernel.cpp
strongoier: Ah casting also done here... Yep you're right. They can be removed.

PR #3374, opened 2021-11-04
[Lang] Fp16 interacts with pytorch.

ailzhang:
* #3384
* __->__ #3374
* #3373

### DIFF
--- a/tests/python/test_f16.py
+++ b/tests/python/test_f16.py
@@ -1,6 +1,7 @@
 import math
 
 import numpy as np
+import pytest
 
 import taichi as ti
 from taichi import approx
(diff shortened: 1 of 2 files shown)


tests/python/test_f16.py
```diff
@@ -51,6 +52,45 @@ def init():
         assert (z[i] == i * 3)
 
 
+@pytest.mark.skipif(not ti.has_pytorch(), reason='Pytorch not installed.')
+@ti.test(arch=archs_support_f16)
+def test_to_torch():
+    n = 16
+    x = ti.field(ti.f16, shape=(n))
```
strongoier, line 59: `ti.field` can accept a tuple or an int as its shape. However, `(n)` is an int which looks like a tuple. It might be clearer to make it more specific, i.e., `n` or `(n, )`.

strongoier (approved):
LGTM!

k-ye (commented):
LGTM!

Do we know how torch implements float16 🤔 ?

PR #3377, opened 2021-11-04
[IR] Add type_check for expressions related to fields and matrices

strongoier:
Related issue = #3301

This PR handles type checking for expressions related to fields and matrices:
-  `type_check` for `TensorElementExpression` and `GlobalPtrExpression` are added.
- Type inference for `LocalLoadExpression` and `GlobalLoadExpression` are handled in constructors because there's no chance to call `type_check` for them in Python side.
- `GlobalVariableExpression` and `ExternalTensorExpression` doesn't need `type_check` for now because the types are encoded in the expression names and there aren't types for them yet.

After this PR, the error for the following code snippet:
```python
import taichi as ti

ti.init()
a = ti.ndarray(ti.i32, shape=(10, 10))

@ti.kernel
def my_func(x: ti.any_arr()):
    print(x[3, 1.1])

my_func(a)
```
can be reported as:
```
[Taichi] version 0.8.4, llvm 10.0.0, commit 0237dd47, linux, python 3.8.10
[Taichi] Starting on arch=x64
[Taichi] Compilation failed
  File "easy-ir-test.py", line 8, in my_func
    print(x[3, 1.1])
TaichiTypeError: indices must be integers, however 'f32' is provided as index 1
```


tests/python/test_type_check.py
```diff
@@ -24,3 +24,16 @@ def bitwise_float():
...
+@pytest.mark.skipif(not ti.has_pytorch(), reason='Pytorch not installed.')
+@ti.test(arch=ti.cpu)
+def test_subscript():
+    a = ti.ndarray(ti.i32, shape=(10, 10))
+
+    @ti.kernel
+    def any_array(x: ti.any_arr()):
+        b = x[3, 1.1]
+
+    with pytest.raises(SystemExit):
```
ailzhang, line 38: potential followup: maybe it's also good to check error message thrown here? 
similar to C++, as our type system mature, we might also need to assert on accurate error messages. https://github.com/google/googletest/issues/952#issuecomment-521361666

ailzhang (approved):
Nice work!

tests/python/test_type_check.py
strongoier: I agree. It seems that Python 3.6 doesn't allow arbitrarily modify the traceback stack of an `Exception` so I'm manually printing the modified stack and doing `sys.exit(1)` for error reports. If we can make it into the `Exception` class, this will be easier to check IMO.

ailzhang: @strongoier Would you mind opening an issue elaborating a bit more on the issue (maybe with an example? ;) I'm a bit curious if we can find a workaround for it

strongoier: #3422

Issue #3346, opened 2021-11-01
Fail to setup dev env under Apple M1


WenheLI:
**Describe the bug**
I was trying to set up the dev env for my M1 Mac. However, I encountered the following error while running `python setup.py develop --user`

```bash
Building commit 470e47dcc8199d9bc14a6233a946093000260319
[  0%] Built target generate_commit_hash
/usr/local/bin/llvm-as: runtime_arm64.ll:2540:74: error: expected ')' at end of argument list
define void @Element_get_pcoord(%struct.PhysicalCoordinates* noalias sret(%struct.PhysicalCoordinates) align 4 %0, %struct.Element* %1) #0 {
```
My llvm is 10.0.0
And both C compiler and C++ compiler are 11.1.0
And the python version is 3.9.7.
Any one know how to fix it?

k-ye:
Hmm, what's your `/usr/local/bin/llvm-as --version`? Note that it might not pick up the correct binary.

Relevant issues:

* https://github.com/taichi-dev/taichi/issues/3320
* https://github.com/taichi-dev/taichi/issues/3183

WenheLI:
@k-ye - llvm-as aligns with llvm and both of them are 10.0.0. 
```bash
/usr/local/bin/llvm-as --version
LLVM (http://llvm.org/):
  LLVM version 10.0.0
  Optimized build with assertions.
  Default target: arm-apple-darwin21.1.0
  Host CPU: cyclone
```
Also the cmake prepare shows I am using the correct binaries:
```
---------- Running CMake prepare ----------------------------------------
-- The C compiler identification is Clang 11.1.0
-- The CXX compiler identification is Clang 11.1.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/homebrew/opt/llvm@11/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /opt/homebrew/opt/llvm@11/bin/clang++ - skipped
```

Quite unsure where the problem is.

k-ye:
Huh, could you replace these with the actual command?

https://github.com/taichi-dev/taichi/blob/dfd1e4cb131be2380bd9f8df49b661bdc6dfc490/CMakeLists.txt#L99-L100

[...]

PR #3379, opened 2021-11-04
[misc] Improve the mechanism to find clang

sjwsl:
Related issue = #3346

1. Use CMAKE_CXX_COMPILER as clang executable if applicable.
2. Raise fatal if clang version >= 13 on macOS, >=12 otherwise. 
3. Raise warning if versions of clang and llvm-as don't match.

Note: `sysroot` must be explicitly specify on macOS when running `clang` as a regular binary.


sjwsl:
I can't find a better way to get the version of clang so use regex now. Any better ideas?

sjwsl:
This breaks CI on MI since it uses clang-12 now. Should we replace it with clang-10? @rexwangcc @k-ye

CMakeLists.txt
```diff
@@ -70,14 +70,43 @@ if (TI_WITH_CUDA)
...
+  endif()
+endif()
+
+find_program(CLANG_EXECUTABLE NAMES ${CLANG_PATH} clang clang-7 clang-8 clang-9 clang-10 clang-11)
 if (NOT CLANG_EXECUTABLE)
     message(FATAL_ERROR "Cannot find any clang executable.")
 endif()
 
+execute_process(COMMAND ${CLANG_EXECUTABLE} --version OUTPUT_VARIABLE CLANG_VERSION_OUTPUT)
+message(${CLANG_VERSION_OUTPUT})
```
k-ye, line 87:
```suggestion
message("clang --version: ${CLANG_VERSION_OUTPUT}")
```

CMakeLists.txt
```diff
@@ -70,14 +70,43 @@ if (TI_WITH_CUDA)
     set(CUDA_ARCH "cuda")
 endif()
 
-find_program(CLANG_EXECUTABLE NAMES clang clang-7 clang-8 clang-9 clang-10)
+if (CLANG_PATH)
```
k-ye, line 73: I'm a bit confused of what's the best approach for setting clang paths now.. AFAICT, we have env var `CC`, `CXX`, and cmake variables `CLANG_PATH` and/or `CMAKE_C_COMPILER`?

k-ye (commented):
Thanks!

CMakeLists.txt
sjwsl: Done.

CMakeLists.txt
sjwsl: Yeah, it's indeed confusing... The point is that we need two `clang` to compile taichi itself and the LLVM runtime respectively. 

Specifically, `CMAKE_C_COMPILER` determines the compiler for taichi, and if not set it is obtained from ENV `CC`. Same for `CMAKE_CXX_COMPILER` and `CXX`. They have no effect on the compiler of the LLVM runtime because that is detected from `find_program` as a normal binary. So the new `CLANG_EXCUTABLE` is used to affect the behavior of `find_program`.

Actually I prefer to use the same `clang` for both if possible... But it will break our CI due to the strange search order of `clang` on macOS.

ailzhang: @sjwsl  is it possible that we change the search order so that we can use the same clang? ;) 
We indeed want to use only one clang IMHO.

sjwsl: Agreed! So can we completely prohibit the GNU compilers? Otherwise we still need to keep these confusing parameters.

yihong0618:
> I can't find a better way to get the version of clang so use regex now. Any ideas?

grep + sed? But I think its the same as `regrex`, maybe no other better way/

CMakeLists.txt
ailzhang: @sjwsl Sure I think we made it clear in the doc that we only support clang so it should be safe to prohibit gcc as long as we throw proper error messages?

sjwsl: Yep. Anyway `clang` is needed for llvm runtime, so it will not bring more troubles. :)

k-ye (approved):
LGTM!

PR #3389, opened 2021-11-05
[gui] Allowing recreating GGUI windows after ti.reset().

AmesingFlank:
I have been trying to write tests for GGUI, and one problem I encountered was that I couldn't run GGUI multiple times separated by ti.reset() calls. There were two reasons for this:

1. ti.reset() doesn't release the resources for the GGUI window. This PR adds an explicit destory() method that works around this. (i.e. after each test, I will call destroy() manually)

2. The vulkan_cuda interop function did not consider the fact that the VulkanDevice and CudaDevice could change. This is fixed in this PR.


bobcao3 (approved):
LGTM! The title seemed a bit confusing (I thought it mean multiple GGUI windows at the same time), so I changed it

Issue #3318, opened 2021-10-29
Put some old docs back


k-ye:
Thanks @strongoier for pointing this out! We used to have API references in https://taichi.readthedocs.io/en/stable/scalar_tensor.html, but they are removed (#2856) from the new doc site: https://docs.taichi.graphics/ site. Some of the content needs to be placed back.

Progress (an item contains an old path before #2856):
- [x] `docs/lang/api/arithmetics.md`
- [x] `docs/lang/api/atomic.md `
- [ ] `docs/lang/api/matrix.md`
- [x] <del>`docs/lang/api/reference/field.md`</del> (nothing more than current docs)
- [ ] `docs/lang/api/reference/ti.md`
- [x] `docs/lang/api/scalar_field.md`
- [ ] `docs/lang/api/snode.md`
- [ ] `docs/lang/api/struct.md`
- [ ] `docs/lang/api/vector.md`

PR #3394, opened 2021-11-05
[Doc] Put back content from old docs/lang/api/arithmetics.md

strongoier:
Related issue = #3318

This PR puts back content from old `docs/lang/api/arithmetics.md` into our current type system section. The organization structure of the type system section is also refined.


ailzhang (approved):
Thanks!!

sjwsl (approved):
LGTM!

PR #3395, opened 2021-11-05
[lang] [refactor] Use preallocated memory via device allocation for Ndarray

qiao-bo:
Use the preallocated memory for Ndarray (CPU+CUDA)

### DIFF
--- a/taichi/backends/cpu/cpu_device.cpp
+++ b/taichi/backends/cpu/cpu_device.cpp
@@ -31,7 +31,8 @@ DeviceAllocation CpuDevice::allocate_memory_runtime(const AllocParams &params,
                                                     LLVMRuntime *runtime,
                                                     uint64 *result_buffer) {
   AllocInfo info;
-  info.ptr = allocate_llvm_runtime_memory_jit(runtime_jit, runtime, params.size, result_buffer);
+  info.ptr = allocate_llvm_runtime_memory_jit(runtime_jit, runtime, params.size,
+                                              result_buffer);
   info.size = params.size;
 
   DeviceAllocation alloc;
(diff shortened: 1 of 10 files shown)


taichi/backends/cpu/cpu_device.cpp
```diff
@@ -26,6 +26,28 @@ DeviceAllocation CpuDevice::allocate_memory(const AllocParams &params) {
   return alloc;
 }
 
+DeviceAllocation CpuDevice::allocate_memory_runtime(const AllocParams &params,
```
k-ye, line 29: nit: let's unify the impl between cpu and cuda

taichi/backends/cuda/cuda_device.cpp
```diff
@@ -31,6 +31,28 @@ DeviceAllocation CudaDevice::allocate_memory(const AllocParams &params) {
   return alloc;
 }
 
+DeviceAllocation CudaDevice::allocate_memory_runtime(const AllocParams &params,
+                                                     JITModule *runtime_jit,
+                                                     LLVMRuntime *runtime,
+                                                     uint64 *result_buffer) {
+  AllocInfo info;
+  if (params.host_read || params.host_write) {
```
strongoier, line 39: Can we make more code into the shared impl? IMO except L39 here, `allocate_memory_runtime` in `CpuDevice` and `CudaDevice` are still the same

qiao-bo: I have thought about it. First, line 39 as well as line 46 are GPU specific. Second, if we move the allocation code, the other allocate_memory function should be changed as well. Maybe there can be another PR later to just add something like a LLVMDevice layer. WDYT?

strongoier: Sounds good.

strongoier (approved):
Thanks!

Issue #3038, opened 2021-09-28
(Prototype) Generate the frontend AST when traversing the Python one


k-ye:
**Concisely describe the proposed feature**

Right now Taichi generates its C++ AST in the following steps:

1. Traverse the Python AST, substitutes many parts into function calls to build the C++ AST
2. Compile this transformed Python AST back to a Python function
3. Invoke the Python function, so that the C++ AST is built

**Describe the solution you'd like (if any)**

Build the C++ AST directly while traversing the Python AST.

## AST Node Implementation Progress

Based on https://docs.python.org/3.9/library/ast.html#abstract-grammar

### mod
- [x] Module
- [x] ~~Interactive~~
- [x] ~~Expression~~
- [x] ~~FunctionType~~

### stmt 
- [x] FunctionDef
  - [x] return value 
  - [x] kernel arguments
    - [x] ti.template
    - [x] ti.sparse_matrix_builder
    - [x] ti.any_arr
    - [x] scalar
  - [x] function arguments

- [x] ~~AsyncFunctionDef~~
- [x] ~~ClassDef~~
- [x] Return

- [x] ~~Delete~~
- [x] Assign
  - [x] static assign
  - [x] normal assign
- [x] AugAssign
- [x] ~~AnnAssign~~

- [x] For
  - [x] static range/ndrange for
  - [x] static grouped for
  - [x] range for
  - [x] ndrange for
  - [x] grouped ndrange for
  - [x] struct for
  - [x] grouped struct for
- [x] ~~AsyncFor~~
- [x] While
- [x] If
  - [x] static if
  - [x] normal if 
- [x] ~~With~~
- [x] ~~AsyncWith~~

- [x] ~~Raise~~
- [x] ~~Try~~
- [x] Assert

- [x] ~~Import~~
- [x] ~~ImportFrom~~

- [x] ~~Global~~
- [x] ~~Nonlocal~~
- [x] Expr
- [x] Pass
- [x] Break
- [x] Continue

### expr
- [x] BoolOp
- [x] NamedExpr
- [x] BinOp
- [x] UnaryOp
- [x] ~~Lambda~~
- [x] IfExp
- [x] Dict
- [x] ~~Set~~
- [x] ListComp
- [x] ~~SetComp~~
- [x] DictComp
- [x] ~~GeneratorExp~~
- [x] ~~Await~~
- [x] ~~Yield~~
- [x] ~~YieldFrom~~
- [x] Compare
[...]

PR #3398, opened 2021-11-05
[refactor] Add Assert, BoolOp, NamedExpr and dict to the new AST builder

lin-hitonami:
Related issue = #3038


ailzhang (approved):
LGTM!

tests/python38/test_ast_refactor.py
```diff
@@ -0,0 +1,16 @@
+from sys import version_info
+
+import pytest
+
+import taichi as ti
+
+
+@ti.test(experimental_ast_refactor=True)
```
ailzhang, line 8: can we add a note here to explain why this is a separate file/link to the github issue? :P

ailzhang (commented):
inline comment can be a followup PR ;)

strongoier (approved):
LGTM!

tests/python38/test_ast_refactor.py
lin-hitonami: I will add it in the next pr.

PR #3399, opened 2021-11-05
[ci] Skip the full test when the PR just get Docs change

Zheaoli:
Related issue = #


ailzhang (approved):
Niceeee! Thanks a lot!! 
(yea we do need a followup to avoid sprinkling these lines in every step :P

frostming:
That was a duplicate of code. We can simply make use of [`on.pull_requests.paths_ignore`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths) in workflow.

frostming:
I've made the changes in PR #3409

PR #3404, opened 2021-11-05
[LLVM] Fix logging formats

sjwsl:
1. Avoid splitting one log into two lines.
2. Use correct atomic operation names in logs.


ailzhang (approved):
LGTM!

PR #3407, opened 2021-11-06
[refactor] First clang-tidy pass

taichi-gardener:
Related = #3354

### DIFF
--- a/scripts/run_clang_tidy.sh
+++ b/scripts/run_clang_tidy.sh
@@ -6,4 +6,6 @@ mkdir -p build_clang_tidy/
 cd build_clang_tidy
 cmake .. -DCMAKE_CXX_COMPILER=clang -DCMAKE_C_COMPILER=clang -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
 cd ..
-python3 scripts/run_clang_tidy.py $PWD/taichi -header-filter="$PWD/taichi/" -p build_clang_tidy -j16 -fix
+TAICHI_SRC=$PWD/taichi
+VAR=${1:-${TAICHI_SRC}}
+python3 scripts/run-clang-tidy.py $PWD/taichi -header-filter="$PWD/taichi/" -p build_clang_tidy -j16 -fix
--- a/scripts/run_clang_tidy.sh
+++ b/scripts/run_clang_tidy.sh
@@ -8,4 +8,4 @@ cmake .. -DCMAKE_CXX_COMPILER=clang -DCMAKE_C_COMPILER=clang -DCMAKE_EXPORT_COMP
 cd ..
 TAICHI_SRC=$PWD/taichi
 VAR=${1:-${TAICHI_SRC}}
-python3 scripts/run-clang-tidy.py $PWD/taichi -header-filter="$PWD/taichi/" -p build_clang_tidy -j16 -fix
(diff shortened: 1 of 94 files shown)


scripts/run_clang_tidy.sh
```diff
@@ -6,4 +6,6 @@ mkdir -p build_clang_tidy/
 cd build_clang_tidy
 cmake .. -DCMAKE_CXX_COMPILER=clang -DCMAKE_C_COMPILER=clang -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
 cd ..
-python3 scripts/run_clang_tidy.py $PWD/taichi -header-filter="$PWD/taichi/" -p build_clang_tidy -j16 -fix
+TAICHI_SRC=$PWD/taichi
+VAR=${1:-${TAICHI_SRC}}
+python3 scripts/run-clang-tidy.py $PWD/taichi -header-filter="$PWD/taichi/" -p build_clang_tidy -j16 -fix
```
sjwsl, line 11: It has been renamed to run_clang_tidy.py in #3354.

taichi/analysis/clone.cpp
```diff
@@ -12,7 +12,7 @@ TLANG_NAMESPACE_BEGIN
 class IRCloner : public IRVisitor {
  private:
   IRNode *other_node;
```
k-ye, line 14: Lovely, wonder what `clang-tidy` criteria is for picking which set of private members to fix...

sjwsl:
Awesome! I guess I can add the naming checks into CI now? This way we can maintain only one configuration under the root directory.

taichi/analysis/clone.cpp
sjwsl: Huh, both False negatives and false positives seem to be inevitable for static analysis tools.

sjwsl:
May I ask about the source of our current naming conventions? Should we follow [Google C++ style] instead(https://google.github.io/styleguide/cppguide.html#Naming)?

k-ye:
> May I ask about the source of our current naming conventions? Should we follow [Google C++ Style](https://google.github.io/styleguide/cppguide.html#Naming) instead?

Good question. We follow Google's C++ style with one exception: the function name still uses the `snake_case`.

taichi/analysis/clone.cpp
qiao-bo: quite interesting that the one in taichi/analysis/same_statements.cpp has been picked up :))

qiao-bo (commented):
cool, thanks a lot! this PR also includes some manual fix and some added glsl header change, right? how is it supposed to be merged?

ailzhang (approved):
LGTM! Thanks!

taichi-gardener:
> cool, thanks a lot! this PR also includes some manual fix and some added glsl header change, right? how is it supposed to be merged?

Not quite sure, it is probably impossible to make separate PRs on them. Should we just merge it in rn or should we wait?

taichi/analysis/clone.cpp
bobcao3: It could also be that sources outside of the src tree referenced this and since tidy only changes the specified source tree, it disabled this. Only a theory, not sure whether it's the actual case here.

k-ye (changes requested):
Let's wait @g1n0st

k-ye (approved):
Ready to proceed 🚀  (unblocked by https://github.com/taichi-dev/taichi/issues/3608)

Issue #2880, opened 2021-09-06
Simplify `ti.Matrix`


strongoier:
Currently `ti.Matrix` has become a single class with more than 1000 lines, with all deprecated/obsolete/unused/rarely used features mixed together. For better code maintainability, we need to carefully decide what should be exposed to users and clean up this class.

victoriacity:
I'd recommend moving the linear algebra methods (`inverse()`, `diag()`, `determinant()`) to `linalg.py` so that we have a clear distinction between API and math implementations. Then `Matrix.inverse()` and similar methods will become an alias of `ti.inverse()`, similar to NumPy ndarrays and PyTorch tensors.

PR #3408, opened 2021-11-06
[refactor] Remove Matrix.new

strongoier:
Related issue = #2880

`Matrix.new` is an internal API which is used only once. This adds unnecessary complexity to the `Matrix` class.


k-ye (approved):
LGTM!

Issue #3413, opened 2021-11-07
Taichi LLVM installation


NiclasSchwalbe:
I do not seem to get LLVM to work with CMake. I already massacred google because of it.

I downloaded llvm via git and put it into the Programm(x86) folder
In the llvm-repo I made a build folder and wrote

cmake ..\llvm\

Then I went into VS2019 and "Build all targets" 
Which was a huge task and I froze once...
In Taichi we have this requirement in the CMake:

find_package(LLVM REQUIRED CONFIG)

This fails, LLVM is not found.
So I try to add LLVM_INCLUDE_DIRS manually. But this does not seem to work...

What would my  LLVM_INCLUDE_DIRS be? Considering I built in  Programm(x86) folder

http://llvm.org/docs/CMake.html#embedding-llvm-in-your-project

_________________________________________________________________________________

With the help of the LLVM Guys, I have brought it to work.
I made a PULL - Request. But I am not 100% this will do it...

sjwsl:
Hi! Could you try adding `-DLLVM_DIR=C:/Program Files(x86)/LLVM/build/installed/lib/cmake/llvm` when running cmake for taichi?

sjwsl:
Or you can download our [prebuilt LLVM](https://github.com/taichi-dev/taichi_assets/releases/download/llvm10/taichi-llvm-10.0.0-msvc2019.zip). Then use `LLVM_DIR={prebuilt_llvm_directory}/lib/cmake/llvm python3 setup.py develop --user` to build taichi.

PR #3414, opened 2021-11-07
[build] Please help in correcting CMake...

NiclasSchwalbe:
Is LLVM_AS_EXECUTABLE redudant?

Related issue = #3413 

I have got no idea. But it is an absolute pain in the ass to install llvm and use it with this build.
I also encountered that the variable LLVM_AS_EXECUTABLE is actually false, while LLVM_AS was installed (seemingly)
I think "NOT TARGET LLVM_AS" would be better.
Can CMAKE search in "ProgramFiles (x86)" for LLVM?

### DIFF
--- a/cmake/TaichiCore.cmake
+++ b/cmake/TaichiCore.cmake
@@ -200,6 +200,10 @@ if(DEFINED ENV{LLVM_DIR})
     message("Getting LLVM_DIR=${LLVM_DIR} from the environment variable")
 endif()
 
+if(WIN)
+list(APPEND CMAKE_PREFIX_PATH "C:/Program Files(x86)/LLVM")
+endif()
+
 # http://llvm.org/docs/CMake.html#embedding-llvm-in-your-project
 find_package(LLVM REQUIRED CONFIG)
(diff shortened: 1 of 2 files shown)


sjwsl:
Thanks for contribution! Could you check if `llvm-as` is indeed in `PATH` by running it in command line? `find_program` should find it as long as you can run it.

cmake/TaichiCore.cmake
```diff
@@ -200,6 +200,10 @@ if(DEFINED ENV{LLVM_DIR})
     message("Getting LLVM_DIR=${LLVM_DIR} from the environment variable")
 endif()
 
+if(WIN)
+list(APPEND CMAKE_PREFIX_PATH "C:/Program Files(x86)/LLVM")
```
k-ye, line 204: I don't think we should make such assumption about where the LLVM lib is installed

ailzhang:
FYI to add `llvm-as` in the path, https://github.com/taichi-dev/taichi/pull/3410/files#diff-5dc3724742e2d0d09dc98e551427757c551cf1f614cd220df59d4120b26ce680R39-R40 might give some hints :D

PR #3417, opened 2021-11-08
[opengl] Save opengl aot data in json format.

ailzhang:
This PR does two things:
1. It converts `TextSerializer` to output json format instead of a customized format. I tried to not introduce any extra dependency like jsoncpp but just make our output json compliant. (two regex tricks are used :P) 
2. Since Json doesn't support multiple line string, it saves each glsl kernel source code in a `.glsl` file and only saves the filename in json. 

Example generated aot json file: https://gist.github.com/ailzhang/48b6126cabae2e9f2d35da02b210a719 

Currently I only changed opengl backend to json, will change metal later together with enabling its tests.

### DIFF
--- a/taichi/common/serialization.h
+++ b/taichi/common/serialization.h
@@ -18,6 +18,7 @@
 #include <type_traits>
 #include <unordered_map>
 #include <vector>
+#include <regex>
 
 #ifdef TI_INCLUDED
 TI_NAMESPACE_BEGIN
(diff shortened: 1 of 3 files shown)


taichi/common/serialization.h
```diff
@@ -636,9 +638,25 @@ class TextSerializer : public Serializer {
     this->process(key, t);
   }
 
+  // Entry to make an AOT json file
+  template <typename T>
+  void serialize_to_json(const char *key, const T &t) {
```
k-ye, line 643: Hmm, is this all we need for generating conforming JSON... ?

k-ye (approved):
Great, LGTM!

taichi/common/serialization.h
ailzhang: Since our output structure is kinda stable so yea it's pretty simple to generate a conforming JSON. To simplify the problem, I didn't use `[]` at all, but keep using `{}` for now which is sufficient for simple cases. :P

PR #3418, opened 2021-11-08
[OpenGL] Add mem_offset_in_parent to serialized file in AOT.

ailzhang:
Related issue = #

### DIFF
--- a/taichi/backends/opengl/aot_data.h
+++ b/taichi/backends/opengl/aot_data.h
@@ -27,6 +27,7 @@ struct CompiledFieldData {
   std::string field_name;
   uint32_t dtype;
   std::string dtype_name;
+  size_t mem_offset_in_parent{0};
   std::vector<int> shape;
   bool is_scalar{false};
   int row_num{0};
@@ -35,6 +36,7 @@ struct CompiledFieldData {
   TI_IO_DEF(field_name,
             dtype,
             dtype_name,
+            mem_offset_in_parent,
             shape,
             is_scalar,
             row_num,
--- a/taichi/backends/opengl/aot_module_builder_impl.cpp
+++ b/taichi/backends/opengl/aot_module_builder_impl.cpp
@@ -65,6 +65,16 @@ void AotModuleBuilderImpl::add_per_backend(const std::string &identifier,
   aot_data_.kernels.push_back({compiled, identifier});
 }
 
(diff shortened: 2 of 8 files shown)


taichi/backends/opengl/aot_data.h
```diff
@@ -27,6 +27,7 @@ struct CompiledFieldData {
   std::string field_name;
   uint32_t dtype;
   std::string dtype_name;
+  size_t mem_offset_in_parent;
```
k-ye, line 30:
```suggestion
  size_t mem_offset_in_parent{0};
```

taichi/backends/opengl/aot_module_builder_impl.cpp
```diff
@@ -9,6 +9,24 @@ namespace taichi {
 namespace lang {
 namespace opengl {
 
+namespace {
+bool all_fields_are_dense_in_container(const SNode *container) {
```
k-ye, line 13: Can we move this to a common place..

taichi/program/aot_module_builder.cpp
```diff
@@ -31,5 +31,31 @@ void AotModuleBuilder::add_kernel_template(const std::string &identifier,
   add_per_backend_tmpl(identifier, key, kernel);
 }
 
+bool AotModuleBuilder::all_fields_are_dense_in_container(
```
k-ye, line 34: I feel like this can have an even wider applicability, but the base AOT builder seems good enough :-)

taichi/program/aot_module_builder.h
```diff
@@ -48,6 +48,10 @@ class AotModuleBuilder {
   virtual void add_per_backend_tmpl(const std::string &identifier,
                                     const std::string &key,
                                     Kernel *kernel) = 0;
+
+  bool all_fields_are_dense_in_container(const SNode *container);
```
k-ye, line 52: nit: static

k-ye (approved):
LGTM!

taichi/program/aot_module_builder.h
ailzhang: nice catch!

taichi/program/aot_module_builder.cpp
ailzhang: yup agreed! I'd just keep this here for now since we know it'll be useful for all aot backends for sure. Later we can move it to other places if needed :D

PR #3419, opened 2021-11-08
[Doc] Update dev install about m1 prebuilt llvm.

ailzhang:
Related issue = #


ailzhang:
cc: @Vissidarte-Herman

qiao-bo (approved):
Thanks! Looks great!

Vissidarte-Herman:
If you *encounter* "" on macOS, there are two ways to work around *this*: 
- (One-off) `System configurations` > `xxx` > `yyy`.
- (Permanent) Run `sudo spctl --master-disable` in your terminal to allow your MacBook to download apps from anywhere by default.

Vissidarte-Herman (changes requested):
If you encounter `llvm-as can’t be opened because Apple cannot check it for malicious software` on macOS, you have two options to work around this:

(One-off) `System configurations` > `Security & Privacy` > `General` > `Allow anyway`.
(Permanent) Run sudo spctl --master-disable` in your terminal to allow your MacBook to download apps from anywhere by default.

PR #3427, opened 2021-11-08
[ci] Move required cpu check to AWS machine

Leonz5288:
Related issue = #
Let's try aws machine with the required cpu test first!


ailzhang (approved):
Nice!

PR #3428, opened 2021-11-08
[llvm] Support atomic operations of f16

sjwsl:
Support atomic add/max/min of `f16` by CAS (`CreateAtomicCmpXchg`) on CPU and CUDA backends. 
CPU: Cast to `i16` pointer.
CUDA: Extend to `i32` pointer in a hacky way because CAS for 16-bit values is not supported. The correctness depends on the memory arrangement.

### DIFF
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -331,6 +331,35 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
         llvm::AtomicOrdering::SequentiallyConsistent);
   }
 
+  // A huge hack for supporting f16 atomic add/max/min!
+  // The reason is that LLVM10 does not support generating atomicCAS for f16 on NVPTX backend.
+  //
+  // int32 is used for the atomicCAS operation. So atomicCAS reads and writes 32 bit values from
+  // the memory, which is larger than the memory size required by the original
+  // atomic binary operation. We mask off the last two bits of the output_address
+  // and use the result as an address to read the 32 bit values from the memory.
+  //
+  // This can avoid out of bound memory accesses, based on the assumption:
+  // All buffers are 4 byte aligned and have a size of 4N.
+  //
+  // The pseudo code is shown below.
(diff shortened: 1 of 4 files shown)


bobcao3:
oops, vulkan f16 atomics failed.

tests/python/test_f16.py
```diff
@@ -218,3 +218,60 @@ def paint(t: float):
...
+        for i in range(1000):
+            f[0] += 1.12
+
+        # Serial sum
+        for _ in range(1):
+            for i in range(1000):
+                f[1] = f[1] + 1.12
+
+    foo()
+    assert (f[0] == f[1])
```
bobcao3, line 239: This can fail, we need an epsilon when comparing floats as its not commutative

sjwsl: Thanks!

sjwsl:
> oops, vulkan f16 atomics failed.

Ah guess we have not supported atomic ops on Vulkan? Now I temporarily exclude Vulkan in related tests to pass CI.

ailzhang (approved):
Awesome!! Thanks!

taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -330,15 +331,174 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
         llvm::AtomicOrdering::SequentiallyConsistent);
   }
 
+  // A huge hack for supporting f16 atomic add/max/min! Borrowed from
+  // https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/gpu/ir_emitter.cc#L378-L490
```
k-ye, line 335: tiny nit: https://github.com/tensorflow/tensorflow/blob/470d58a83470f8ede3beaa584e6992bc71b7baa6/tensorflow/compiler/xla/service/gpu/ir_emitter.cc#L378-L490

sjwsl: Thanks!

PR #3430, opened 2021-11-09
[gui] GGUI Tests

AmesingFlank:
I stored a few ground truth images in the `tests/python/images` folder. I'm not sure if this is the best thing to do. Lmk if you have any suggestions.


ailzhang (approved):
Thanks!! I'm happy with this as a temporary solution as long as we don't change it often. Later we might consider either only asserts on a few values in the images? 

Nit: maybe rename the folder to `expected`?

ailzhang:
ohhhh @AmesingFlank the error here comes from we don't ship these image files in our wheel. 
Tbh we should use pytest instead of `ti test`  in CI (`ti test` isn't supposed to be user facing)

AmesingFlank:
> ohhhh @AmesingFlank the error here comes from we don't ship these image files in our wheel. Tbh we should use pytest instead of `ti test` in CI (`ti test` isn't supposed to be user facing)

Hmm, how do I include these images into the wheel?

https://github.com/taichi-dev/taichi/blob/bc8eaa6be60798a8f09523ac02e2111eedbda323/setup.py#L85

Feels like the entire `tests` directly is being copied.

AmesingFlank:
Ah nevermind. Figured it out. Needed to update `MANIFEST.IN`.

bobcao3 (approved):
Great work!

Issue #3431, opened 2021-11-09
Use triangles to simulate lines in GGUI


AmesingFlank:
Currently, during `canvas.lines()`, GGUI directly asks the GPU to draw lines. However, drawing lines with arbitrary width is not a well-supported operation. It would be better to generate triangles that simulate these lines instead.

Issue #3416, opened 2021-11-07
Vulkan backend on m1 Mac ( device does not support wide lines)


Linyou:
The problem was occurred when running ggui_example/mass_spring_game_ggui.py

```
❯ python mass_spring_game_ggui.py                                          
[Taichi] version 0.8.4, llvm 10.0.0, commit f665c28c, osx, python 3.9.7
[Taichi] Starting on arch=vulkan
[W 11/07/21 22:14:21.787 427294] [embedded_device.cpp:create_logical_device@407] Potential non-conformant Vulkan implementation, enabling VK_KHR_portability_subset
[W 11/07/21 22:14:21.787 427294] [embedded_device.cpp:create_logical_device@464] Taichi GPU GUI requires wide lines support
[W 11/07/21 22:14:21.986 427294] [vulkan_device.cpp:buffer@527] Overriding last binding
[W 11/07/21 22:14:21.991 427294] [vulkan_device.cpp:buffer@527] Overriding last binding
[mvk-error] VK_ERROR_FEATURE_NOT_PRESENT: vkCmdSetLineWidth(): The current device does not support wide lines.
[E 11/07/21 22:14:22.000 427294] [vulkan_device.cpp:submit_synced@1387] Vulkan Error : -8 : failed to submit command buffer
```

```
                            * Taichi Core - Stack Traceback *                             
==========================================================================================
|                       Module |  Offset | Function                                      |
|----------------------------------------------------------------------------------------|
*               taichi_core.so |     120 | taichi::Logger::error(std::__1::basic_string< |
                                         | char, std::__1::char_traits<char>, std::__1:: |
                                         | allocator<char> > const&, bool)               |
*               taichi_core.so |     540 | taichi::lang::vulkan::VulkanStream::submit_sy |
                                         | nced(taichi::lang::CommandList*)              |
*               taichi_core.so |     508 | taichi::ui::vulkan::Renderer::draw_frame(taic |
                                         | hi::ui::vulkan::Gui*)                         |
*               taichi_core.so |      32 | taichi::ui::vulkan::Window::show()            |
*               taichi_core.so |     108 | void pybind11::cpp_function::initialize<pybin |
                                         | d11::cpp_function::cpp_function<void, taichi: |
                                         | :ui::PyWindow, pybind11::name, pybind11::is_m |
                                         | ethod, pybind11::sibling>(void (taichi::ui::P |
                                         | yWindow::*)(), pybind11::name const&, pybind1 |
[...]

PR #3432, opened 2021-11-09
[gui] Fix canvas.lines on macOS

AmesingFlank:
Related issue = #3416 

This is a temporary work around. The right thing to do would be to generate triangles that simulate lines. This is tracked by issue #3431.


g1n0st (approved):
LGTM!

bobcao3 (approved):

Issue #2658, opened 2021-08-11
[Opt] Refactor advanced_optimization into optimization_level


xumingkuan:
**Concisely describe the proposed feature**
We need more flexibility on which optimizations to be performed. A simple switch `advanced_optimization` is probably not sufficient.

See also: https://github.com/taichi-dev/taichi/issues/1416#issuecomment-654379663 https://github.com/taichi-dev/taichi/pull/1403#pullrequestreview-442632149 https://github.com/taichi-dev/taichi/issues/2375#issuecomment-852722639 https://github.com/taichi-dev/taichi/issues/1469#issuecomment-657168352

**Describe the solution you'd like (if any)**
1. Implement a new config variable `optimization_level` with values 0, 1, 2, 3, and deprecate `advanced_optimization` into setting the `optimization_level` variable.
2. Remove `advanced_optimization` after several releases.

PR #3434, opened 2021-11-09
[Lang] Implement opt_level for accelerating compiling

squarefk:
Related issue = https://github.com/taichi-dev/taichi/issues/2658

Currently, `opt_level` has two options 0 and 1.
- 0 will disable CFG Optimization passes.
- 1 will do nothing.

To run tests with opt_level, simply changed [`opt_level=1;`](https://github.com/squarefk/taichi/blob/d101d9a4a3deff33fc9c5cf50a7764b99a5ca19e/taichi/program/compile_config.cpp#L10) into `opt_level = 0;`

What I have done to fix broken tests after disabling CFG Optimization.
1. AssertStmt in Opengl backend missing (Avoid crash and mark TODO)
2. constant_fold from llvm doesn't behave as expected (Adjust test to avoid calling llvm's constant_fold)
3. Internal function called from CC backend (Disable such tests)
4. LocalStore following Atomic Op has illed type for CustomInt/CustomFloat ([Skip LocalStore is val->is<AtomicOpStmt>() with CustomInt/CustomFloat](https://github.com/squarefk/taichi/blob/d101d9a4a3deff33fc9c5cf50a7764b99a5ca19e/taichi/codegen/codegen_llvm.cpp#L1028-L1034))
5. BLS needs ([enforce opt_level = 1 in when .blocked_local() get called](https://github.com/squarefk/taichi/blob/d101d9a4a3deff33fc9c5cf50a7764b99a5ca19e/python/taichi/lang/__init__.py#L568-L572))
6. Autodiff needs CFG to determine ad variables stack size ([enforce opt_level = 1 when .grad() get called](https://github.com/squarefk/taichi/blob/d101d9a4a3deff33fc9c5cf50a7764b99a5ca19e/python/taichi/lang/kernel_impl.py#L720-L724))

We might want discuss whether there is a better way to resolve 4. to 6.


strongoier (commented):
Looks great!
1. Could you run some experiments and show the acceleration number?
2. In the future I think we need to provide a more systematic mechanism for optimization levels (e.g. a level-to-enable property for each optimization pass).

python/taichi/lang/__init__.py
```diff
@@ -565,6 +565,9 @@ def block_local(*args):
     Args:
         *args (List[Field]): A list of sparse Taichi fields.
     """
+    if impl.current_cfg().opt_level == 0:
+        warning("""opt_level = 1 is enforced to enable bls analysis.""")
```
ailzhang, line 569: for consistency with other code, shall we use `ti.warn`? ;)

tests/python/test_internal_func.py
```diff
@@ -9,7 +9,7 @@ def all_archs_for_this(test):
     return ti.archs_excluding(ti.metal, ti.opengl, ti.cuda, ti.vulkan)(test)
 
 
-@ti.test(exclude=[ti.metal, ti.opengl, ti.cuda, ti.vulkan])
+@ti.test(exclude=[ti.metal, ti.opengl, ti.cuda, ti.vulkan, ti.cc])
```
ailzhang, line 12: would you mind elaborating on why we are disabling these tests on cc backend? Is it only for a certain opt_level or for both? :D

ailzhang (commented):
LGTM! Thanks!

tests/python/test_internal_func.py
k-ye: cc backend is not in a maintained mode (but still serves as a nice codegen reference), we don't have to let it get in our way.

ailzhang: @k-ye yea I totally agree, I guess I'm just curious why it failed and we should make sure it's understood and added in the comment ;)

squarefk: cc backend doesn't have functions from taichi/runtime/llvm/internal_functions.h

python/taichi/lang/__init__.py
squarefk: https://github.com/taichi-dev/taichi/pull/3434/commits/9ccdd16813af75332c2fe738a4c5763ecab86476

ailzhang (approved):
LGTM! 
Btw I totally agree with @strongoier that we need some result numbers to show how different opt_levels impact the compilation speed. 
I'll merge this PR now but Looking forward to followups with more numners!  (maybe write a documentation for opt_level? ;)

PR #3440, opened 2021-11-10
[Doc] Put back content from old docs/lang/api/atomic.md

strongoier:
Related issue = #3318

This PR puts back content from old `docs/lang/api/atomic.md` into our current type system section. The format and content organization are refined to fit current doc better. `ti.atomic_max/min` is also added into the doc.

### DIFF
--- a/docs/lang/articles/basic/type.md
+++ b/docs/lang/articles/basic/type.md
@@ -185,6 +185,74 @@ pow(x, y)  # Same as `x ** y`.
 ti.random(dtype=float)
 ```
 
+### Supported atomic operations
+
+In Taichi, augmented assignments (e.g., `x[i] += 1`) are automatically
+[atomic](https://en.wikipedia.org/wiki/Fetch-and-add).
+
+:::caution
+
+When modifying global variables in parallel, make sure you use atomic
+operations. For example, to sum up all the elements in `x`, :
+
+    @ti.kernel
+    def sum():
+        for i in x:
+            # Approach 1: OK
+            total[None] += x[i]
+
+            # Approach 2: OK
+            ti.atomic_add(total[None], x[i])
+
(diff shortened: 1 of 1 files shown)


ailzhang (approved):
LGTM！
cc: @Vissidarte-Herman but we might want to refine this file as a whole later?

docs/lang/articles/basic/type.md
```diff
@@ -185,6 +185,75 @@ pow(x, y)  # Same as `x ** y`.
...
+| `ti.atomic_or(x, y)`  | atomically compute <code>x &#124; y</code>, store the result in `x`, and return the old value of `x` |
+| `ti.atomic_xor(x, y)` | atomically compute `x ^ y`, store the result in `x`, and return the old value of `x`                 |
+| `ti.atomic_max(x, y)` | atomically compute `max(x, y)`, store the result in `x`, and return the old value of `x`             |
+| `ti.atomic_min(x, y)` | atomically compute `min(x, y)`, store the result in `x`, and return the old value of `x`             |
+
+:::note
+
+Supported atomic operations on each backend:
+
+| type | CPU/CUDA | OpenGL | Metal | C source |
```
sjwsl, line 247: Add Vulkan, i16, and f16?

sjwsl (commented):
> ````

strongoier:
> LGTM！ cc: @Vissidarte-Herman but we might want to refine this file as a whole later?

Ah yes. I'll stop at this point. We just need to remember that we already have something, so there is no need to start from scratch if we decide to write related materials.

docs/lang/articles/basic/type.md
strongoier: Yep they should be added maybe in a future PR.

PR #3441, opened 2021-11-10
[refactor] [bug] Eliminate failing tests on the new AST builder

lin-hitonami:
Related issue = #3038

### DIFF
--- a/misc/code_format.py
+++ b/misc/code_format.py
@@ -133,7 +133,8 @@ def main(all=False, diff=None):
         if re.match(r'.*examples\/[a-z_]+\d\d+\.py$', fn):
             print(f'Skipping example file "{fn}"...')
             continue
-        if fn.find(os.path.join('tests', 'python38')) != -1 and sys.version_info < (3, 8):
+        if fn.find(os.path.join(
+                'tests', 'python38')) != -1 and sys.version_info < (3, 8):
             print(f'Skipping "{fn}"...')
             continue
         if not format_file(fn):
--- a/misc/code_format.py
+++ b/misc/code_format.py
@@ -133,6 +133,9 @@ def main(all=False, diff=None):
         if re.match(r'.*examples\/[a-z_]+\d\d+\.py$', fn):
             print(f'Skipping example file "{fn}"...')
             continue
+        if fn.find('tests/python38/') != -1 and sys.version_info < (3, 8):
+            print(f'Skipping "{fn}"...')
+            continue
         if not format_file(fn):
             print(f'Skipping "{fn}"...')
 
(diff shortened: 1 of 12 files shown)


misc/code_format.py
```diff
@@ -133,6 +133,9 @@ def find_diff_or_empty(s):
         if re.match(r'.*examples\/[a-z_]+\d\d+\.py$', fn):
             print(f'Skipping example file "{fn}"...')
             continue
+        if fn.find('tests/python38/') != -1 and sys.version_info < (3, 8):
```
k-ye, line 136:
```suggestion
        if fn.find(os.path.join('tests', 'python38')) != -1 and sys.version_info < (3, 8):
```

tests/python/test_type_check.py
```diff
@@ -26,17 +26,17 @@ def bitwise_float():
...
-    @ti.kernel
-    def select():
-        a = 1.1
-        b = 3
-        c = 3.6
-        d = b if a else c
-
-    with pytest.raises(SystemExit):
-        select()
+# @ti.test(arch=ti.cpu)
```
k-ye, line 29: Is this disabling intentional?

python/taichi/lang/__init__.py
```diff
@@ -396,7 +396,7 @@ def __init__(self):
         self.gdb_trigger = False
         self.excepthook = False
         self.experimental_real_function = False
-        self.experimental_ast_refactor = False
+        self.experimental_ast_refactor = True
```
g1n0st, line 399: IIUC, after this PR, `experimental_ast_refactor` will turn on by default?

tests/python/test_type_check.py
strongoier: This is still in discussion - whether we should expand `IfExp` in Python side or C++ side. In current implementation,  if `IfExp` is expanded to if statements, they will not be type-checked and no exception will be raised here.

python/taichi/lang/__init__.py
lin-hitonami: IIUC it can be turned on by default if no more bugs are found? @k-ye @ailzhang

g1n0st: MeshTaichi is now rely on the old AST generator, better if turn it on when MeshTaichi merges back and migrates to the new AST generator?

lin-hitonami, line 399:
```suggestion
        self.experimental_ast_refactor = False
```

ailzhang: @lin-hitonami Let's keep experimental_ast_refactor=False in this PR so that @g1n0st can give it a try to rebase MeshTaichi. 
@g1n0st Would you mind giving it a try after PR merges and work with @lin-hitonami to migrate MeshTaichi to this new ast generator asap? We need to get rid of the old ast generator soon otherwise we'll have to maintain both paths ;)

g1n0st: Sure~ I will try it next week and work with @lin-hitonami . : )

strongoier (approved):
LGTM!

PR #3442, opened 2021-11-10
[Lang] User-friendly exception when copying between ti.field

ljjsalt:
Related issue = #2846

Test 0
```
x = ti.field(ti.f32, (2, 3))
y = ti.field(ti.f32, (2, 3, 4))
x.copy_from(y)
```

Result 0
```
ValueError: ti.field shape (2, 3) does not match the source field shape (2, 3, 4)
```

Test 1
```
x = ti.field(ti.f32, (2, 3))
y = np.zeros((2, 3))
x.copy_from(y)
```

Result 1
```
TypeError: Cannot copy from a non-field object
```


ljjsalt:
@ailzhang

python/taichi/lang/field.py
```diff
@@ -1,3 +1,5 @@
+from typing import Type
```
ailzhang, line 1: this import is not used?

ailzhang (commented):
cc: @ljjsalt the test failures on GPU seem related?

python/taichi/lang/field.py
ljjsalt: I am sorry, it's removed.

tests/python/test_field.py
```diff
@@ -146,3 +146,36 @@ def test_field_name():
     for i in range(10):
         d.append(ti.field(dtype=ti.f32, shape=(2, 3), name=f'd{i}'))
         assert d[i].name == f'd{i}'
+
+
+@ti.test()
+@pytest.mark.parametrize('shape', field_shapes)
+def test_field_copy_from(shape):
+    shapes = [ti.i32, ti.f32]  # Metal kernel only supports <= 32-bit data
```
ailzhang, line 154: shape doesn't match what's inside the array :P 
Using parameterize for dtype should work and pass CI  :D 
```
+@ti.test()
+@pytest.mark.parametrize('shape', field_shapes)
+@pytest.mark.parametrize('dtype', [ti.i32, ti.f32])
+def test_field_copy_from(shape, dtype):
+    x = ti.field(dtype=ti.f32, shape=shape)
+    other = ti.field(dtype=dtype, shape=shape)
+    other.fill(1)
+    x.copy_from(other)
+    convert = lambda arr: arr[0] if len(arr) == 1 else arr
+    assert (convert(x.shape) == shape)
+    assert (x.dtype == ti.f32)
+    assert ((x.to_numpy() == 1).all())
```

ailzhang (changes requested):

tests/python/test_field.py
ljjsalt: @ailzhang 
Hi, I finally have time to keep working on this PR, how can I test locally? I have run the pytest script.

ailzhang (approved):
LGTM! Thanks a lot for contributing!

PR #3450, opened 2021-11-10
[misc] Fix python wheel versioning.

ailzhang:
Related issue = #


setup.py
```diff
@@ -35,6 +35,10 @@
 ]
 
 project_name = os.getenv('PROJECT_NAME', 'taichi')
+TI_VERSION_MAJOR = 0
+TI_VERSION_MINOR = 8
+TI_VERSION_PATCH = 5
+version = f'{TI_VERSION_MAJOR}.{TI_VERSION_MINOR}.{TI_VERSION_PATCH}'
```
rexwangcc, line 41: FTR: this is the source of truth of the version info.

rexwangcc (approved):
👍 LGTM! Thanks!!

strongoier:
It seems that this change reverts what was done in #3354. Which way should we take? We need to avoid modify this back-and-forth IMO :-)

frostming:
How are these values to be updated? Maybe we can get them from git tags automatically. I am happy to provide an implementation.

strongoier:
> How are these values to be updated? Maybe we can get them from git tags automatically. I am happy to provide an implementation.

@ailzhang @rexwangcc WDYT? In this case we no longer need to manually set versions.

ailzhang:
Hey @frostming @strongoier, sorry for the late reply! @frostming would you mind describing your "fetching from git tags automatically" idea in detail in a new issue? 
My main concern is if someone download our zipped source code (it is a folder without `.git/` in this case), then will they be able to get the version information? As long as this work I'll be super happy to see this process being automatic! :D

PR #3454, opened 2021-11-10
[misc] Remove regex in TextSerializer.

ailzhang:
Also added mpm88 in aot tests.

Related issue = #


k-ye (approved):
LGTM!

@sjwsl feel free to explore a simpler way if you wish. I think the fact whether we need to append a comma can be decided at compile time :-)

PR #3459, opened 2021-11-11
[misc] Add default values of TI_VERSION

sjwsl:
This makes `cmake` stage self-contained.


CMakeLists.txt
```diff
@@ -5,9 +5,12 @@
 cmake_minimum_required(VERSION 3.12)
 
 project(taichi)
-SET(TI_VERSION_MAJOR ${TI_VERSION_MAJOR})
-SET(TI_VERSION_MINOR ${TI_VERSION_MINOR})
-SET(TI_VERSION_PATCH ${TI_VERSION_PATCH})
+
+if (NOT DEFINED TI_VERSION_MAJOR)
```
ailzhang, line 9: Can we also send a warning message when we do this? :D

ailzhang (approved):
LGTM!

CMakeLists.txt
sjwsl: Yep. Actually users and even developers should never run `cmake` directly. So it's reasonable to give a warning if they really get here.

PR #3461, opened 2021-11-11
[IR] Enforce type check for all expressions

strongoier:
Related issue = #3301

This PR adds `type_check` for remaining expressions and removes all `TODO: assert no unknowns after type_check for all expressions are implemented`.


taichi/ir/frontend_ir.cpp
```diff
@@ -128,10 +131,12 @@ void UnaryOpExpression::serialize(std::ostream &ss) {
 }
 
 void UnaryOpExpression::type_check() {
-  // TODO: assert no unknowns after type_check for all expressions are
-  // implemented
-  if (operand->ret_type == PrimitiveType::unknown)
-    return;
+  TI_ASSERT_INFO(operand->ret_type != PrimitiveType::unknown,
```
ailzhang, line 134: nit: maybe this assert worth its own macro :P

ailzhang (approved):
LGTM!

taichi/ir/frontend_ir.cpp
strongoier: Nice suggestion! Done.

PR #3468, opened 2021-11-11
[Lang] Enable local tensors as arithmetic operation results

strongoier:

This PR rewrites arithmetic operations of matrices so as to simplify the implementation as well as to enable local tensors as arithmetic operation results (`empty_copy()` in the old implementation has `disable_local_tensor=True`).


k-ye (approved):
LGTM!

PR #3472, opened 2021-11-12
[Lang] Fix pylint rule R1703

LJHG:
Related issue = #3151

### DIFF
--- a/python/taichi/lang/ast/checkers.py
+++ b/python/taichi/lang/ast/checkers.py
@@ -86,14 +86,11 @@ class KernelSimplicityASTChecker(ast.NodeVisitor):
     def visit_For(self, node):
         # TODO: since autodiff is enhanced, AST checker rules should be relaxed. This part should be updated.
         return
-        if (isinstance(node.iter, ast.Call)
-                and isinstance(node.iter.func, ast.Attribute)
-                and isinstance(node.iter.func.value, ast.Name)
-                and node.iter.func.value.id == 'ti'
-                and node.iter.func.attr == 'static'):
-            is_static = True
-        else:
-            is_static = False
+        is_static = bool(isinstance(node.iter, ast.Call)
+                            and isinstance(node.iter.func, ast.Attribute)
+                            and isinstance(node.iter.func.value, ast.Name)
+                            and node.iter.func.value.id == 'ti'
+                            and node.iter.func.attr == 'static')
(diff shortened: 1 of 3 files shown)


python/taichi/lang/ast/checkers.py
```diff
@@ -86,14 +86,11 @@ def generic_visit(self, node):
...
         return
-        if (isinstance(node.iter, ast.Call)
-                and isinstance(node.iter.func, ast.Attribute)
-                and isinstance(node.iter.func.value, ast.Name)
-                and node.iter.func.value.id == 'ti'
-                and node.iter.func.attr == 'static'):
-            is_static = True
-        else:
-            is_static = False
+        is_static = bool(isinstance(node.iter, ast.Call)
```
yihong0618, line 89: why add bool() here?
maybe use all?
```python
all([a, b, c])
```

python/taichi/lang/stmt_builder.py
```diff
@@ -236,10 +236,7 @@ def build_If(ctx, node):
             node.test.func, ast.Attribute)
         if is_static_if:
             attr = node.test.func
-            if attr.attr == 'static':
-                is_static_if = True
-            else:
-                is_static_if = False
+            is_static_if = bool(attr.attr == 'static')
```
yihong0618, line 239: and we do need `bool`?

python/taichi/lang/ast/checkers.py
LJHG: I use bool according to the log of pylint.
`R1703: The if statement can be replaced with 'var = bool(test)'`
Do we need change the code into the `all([a, b, c])` format?

LJHG: @ailzhang  Thanks. I will fix that.

ailzhang (approved):
Thanks! Welcome to taichi community!

sjwsl:
Could you also add this pylint rule in CI to verify your work? See here for more details. Thanks!

ailzhang (changes requested):
oops yea we need to enable this in CI as well. Requesting changes for to make sure we don't merge it accidentally. Thanks @sjwsl !

LJHG:
@ailzhang I have enabled this in CI.

ailzhang (approved):
LGTM! Thanks!

PR #3456, opened 2021-11-10
[vulkan] Try to enable Vulkan test in macOS presubmit

bobcao3:
Related issue = #

PR #3474, opened 2021-11-12
[ci] Disable arch fallback on CI

### LINKED ISSUE #3469: Provide an option to disable arch fallback
This feature will be particularly useful in CI. 
With the current smart fallbacks, we sometimes miss testing a backend entirely when fallback happens, which is pretty inconvenient. 

A simple implementation I have in mind is just `ti.init(enable_fallback=False)`, most of the change can happen in python side, so this shouldn't be too hard to implement IMHO.

lin-hitonami:
Related issue = close #3469


.github/workflows/presubmit.yml
```diff
@@ -311,14 +317,15 @@ jobs:
...
-          docker create --user dev --name taichi_test --gpus all -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix registry.taichigraphics.com/taichidev-ubuntu18.04:v0.1.1 /home/dev/unix_docker_test.sh $PY $GPU_TEST
+          docker create --user dev --name taichi_test --gpus all -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix registry.taichigraphics.com/taichidev-ubuntu18.04:v0.1.1 /home/dev/unix_docker_test.sh $PY $GPU_TEST $TI_WANTED_ARCHS
           docker cp .github/workflows/scripts/unix_docker_test.sh taichi_test:/home/dev/unix_docker_test.sh
           docker cp wheel/*.whl taichi_test:/home/dev/
           docker start -a taichi_test
         env:
           PY: py38
           GPU_TEST: ON
           DISPLAY: :1
+          TI_WANTED_ARCHS: "cuda,vulkan,opengl"
```
ailzhang, line 328: let's also include cpu and cc here to make sure it works (since this is our release environment :P

python/taichi/lang/__init__.py
```diff
@@ -1007,11 +1008,11 @@ def is_arch_supported(arch):
         return False
 
 
-def supported_archs():
-    """Gets all supported archs on the machine.
+def expected_archs():
```
ailzhang, line 1011: Can we move this function from init to testing.py? Do we still want to expose it to users?

ailzhang (commented):
Nice! Would you mind also adding a note to `test` in testing.py to describe it behavior? ;)

.github/workflows/presubmit.yml
lin-hitonami: it seems that the GPU test's cc support is disabled in the cmake argument now, should we turn it on?

.github/workflows/presubmit.yml
```diff
@@ -440,4 +449,7 @@ jobs:
           python3 examples/algorithm/laplace.py
           TI_LIB_DIR=`python3 -c "import taichi;print(taichi.__path__[0])" | tail -1`
           TI_LIB_DIR="$TI_LIB_DIR/lib" ./build/taichi_cpp_tests
+          echo wanted archs: $TI_WANTED_ARCHS
           ti test -vr2 -t4 -x
+        env:
+          TI_WANTED_ARCHS: "metal"
```
ailzhang, line 455: metal and cpu?

ailzhang (approved):
Nice work!!

.github/workflows/presubmit.yml
ailzhang: let's give it a try? IIRC it should work

ailzhang:
sorry for the prolonged review process, IIRC we have enable vulkan on macs as well, shall we add that on all mac jobs too?

ailzhang:
Oh actually that PR is not in yet https://github.com/taichi-dev/taichi/pull/3456, cc: @bobcao3 this PR might make your life a bit easier in making sure vulkan is enabled in CI. ;)

PR #3475, opened 2021-11-12
[misc] Enable clang-tidy check in CI

sjwsl:
Related issue = #

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -99,6 +99,7 @@ jobs:
           fi
 
   check_clang_tidy:
+    # TODO: Move installation of clang-tidy to docker image
     name: Check clang-tidy
     runs-on: ubuntu-latest
     needs: check_files
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -98,6 +98,36 @@ jobs:
             exit 1
           fi
 
+  check_clang_tidy:
+    name: Check clang-tidy
+    runs-on: ubuntu-latest
+    needs: check_files
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/presubmit.yml
```diff
@@ -98,6 +98,36 @@ jobs:
...
+        env:
+          CR_PAT: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run clang-tidy
+        run: |
+          if [[ ${{needs.check_files.outputs.run_job}} == false ]]; then
+            exit 0
+          fi
+          docker run -id --user dev --name check_clang_tidy ghcr.io/taichi-dev/taichidev-cpu-ubuntu18.04:v0.1.0 /bin/bash
+          tar -cf - ../${{ github.event.repository.name }} --mode u=+rwx,g=+rwx,o=+rwx --owner dev --group dev | docker cp - check_clang_tidy:/home/dev/
+          docker exec --user root check_clang_tidy apt install -y clang-tidy-10
```
ailzhang, line 126: Can we add a todo here saying that we should install clang-tidy-10 inside the docker image when we rebuild ;) ?

ailzhang (approved):
Nice work! Btw I took a look at our log and it's a bit too verbose :P . Can we have a followup that: 
- If clang-tidy passes, let's just print one sentence and celebrate
- If it fails, clearly list the offending file names and a command to reproduce.

.github/workflows/presubmit.yml
sjwsl: Finished! Thanks :)

ailzhang (approved):
Awesome thanks!

PR #3486, opened 2021-11-12
[Lang] Fix pylint rule W0235

tssujt:
Related issue = #3151

### DIFF
--- a/python/taichi/ui/scene.py
+++ b/python/taichi/ui/scene.py
@@ -76,9 +76,6 @@ def gen_normals(vertices, indices):
 class Scene(_ti_core.PyScene):
     """A 3D scene, which can contain meshes and particles, and can be rendered on a canvas
     """
-    def __init__(self):
-        super().__init__()
-
     def set_camera(self, camera):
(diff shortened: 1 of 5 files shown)


python/taichi/ui/scene.py
```diff
@@ -134,8 +131,5 @@ def particles(self,
         vbo_info = get_field_info(vbo)
         super().particles(vbo_info, has_per_vertex_color, color, radius)
 
-    def point_light(self, pos, color):
+    def point_light(self, pos, color):  # pylint: disable=W0235
```
k-ye, line 134: Why handling this differently :-) ?

k-ye (approved):
Thanks, welcome to the community!

python/taichi/ui/scene.py
tssujt: https://github.com/taichi-dev/taichi/runs/4191773093

@k-ye CI fails, but I'm not familiar with pybind11.

tssujt:
Thanks @sjwsl

PR #3493, opened 2021-11-13
[Lang] Fix pylint rule W0101

darkSheep404:
I noticed that one of the two codes that needed to be fixed has a  `TODO`,
so I just commented on the code after `return` instead of delete

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -88,7 +88,7 @@ jobs:
             exit 0
           fi
           # Make sure pylint doesn't regress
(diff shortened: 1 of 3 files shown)


sjwsl:
Would you mind signing the CLA so that we can accept your contribution? Thanks!

darkSheep404:
> Would you mind signing the CLA so that we can accept your contribution? Thanks!
Sorry, i signed with a different email with my github email before
I has been re-signed

sjwsl:
Huh... It seems that #3494 introduces some new errors to your changes. Could you refer to the error message in CI and #3494 to fix it?

darkSheep404:
ok,i  have fixed it and pushed now

sjwsl (approved):
LGTM, thanks!

PR #3500, opened 2021-11-14
[Lang] Fix pylint rule E1101

DeepDuke:
Signed-off-by: DeepDuke czm.hit@gmail.com
Related issue = #3151


sjwsl:
1. Please use `disable` since `disable-msg`  is deprecated as of pylint 0.21.0.
2. Enabling this check but skipping it everywhere does nothing but increase the number of lines of code... Do you have a better way to resolve them? If not, I think probably this check should not be enabled at all.

DeepDuke:
> 1. Please use `disable` since `disable-msg`  is deprecated as of pylint 0.21.0.
> 2. Enabling this check but skipping it everywhere does nothing but increase the number of lines of code... Do you have a better way to resolve them? Otherwise, I think probably this check should not be enabled at all. WDYT?

It seems pylint cannot parse pytorch ,opencv and some other special cases.  But pylint does not fail in all situations. If  don’t enable it, we will miss some checks ? I tried some other methods, but not work...

strongoier (commented):
A few suggestions:
1. Mark `E1101` as `True` in `scripts/generate_pylint_tags.py`
2. Use `disable` instead of `disable-msg`

DeepDuke:
> A few suggestions:
> 1. Mark `E1101` as `True` in `scripts/generate_pylint_tags.py`
> 2. Use `disable` instead of `disable-msg`

I will fix it soon.

sjwsl (approved):
LGTM, thank you!

DeepDuke:
> LGTM, thank you!
@sjwsl @strongoier 
Thanks for your kind suggestions  and guidance. It's my honor to participate in taichi project！

PR #3502, opened 2021-11-14
[CUDA] Fix a misuse of std::move: address of stack memory associated with temporary object of type std::lock_guard<std::mutex> returned to caller

PragmaTwice:
Related issue = #

In the original code, the returned rvalue reference is obviously a dangling reference, and the scoped mutex has no effect (destructed immediately after construction).

### DIFF
--- a/taichi/backends/cuda/cuda_context.h
+++ b/taichi/backends/cuda/cuda_context.h
@@ -95,8 +95,8 @@ class CUDAContext {
     return ContextGuard(this);
   }
 
-  std::lock_guard<std::mutex> get_lock_guard() {
-    return std::lock_guard<std::mutex>(lock_);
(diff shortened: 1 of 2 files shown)


k-ye (approved):
LGTM && welcome to the community!

PragmaTwice:
Some fact to help understand it:
- std::move is implicit and mandatory while a local variable is returned (in the standard, overload resolution will select a move ctor first, as treat the lvalue as an rvalue, since 14882:2011, addressed in 12.8/32)
- RVO is a mandatory optimization since c++17, so even move ctor will not be called. Using std::move here will prevent RVO. (of course, the return type can never be a reference type while you want to return a local variable (with auto storage duration))

ailzhang (approved):
Nice catch! Thank you!

PR #2922, opened 2021-09-13
[refactor] [benchmark] Add ti.profiler in python scope

yolo2themoon:
Since #2909 added a new structure, `KernelProfileTracedRecord` to store traced records,
now we can keep the profiling data in its original form (records) on the backend, 
and do the processing in python scope.

**Describe the solution you'd like** :
Add `class Profiler` in `python/taichi/profiler/profiler.py` 
`Profiler.update_records()` copys  records from backend to python scope
`Profiler.count_results()` does some statistics on the raw data ( the back-end implementations will be deprecated ) 
`Profiler.print_info()` has two mode : count(default) and trace(print records of launched kernel)

**For compatibility of interfaces**  : 
`ti.print_kernel_profile_info()` now calls `profiler.print_info()`
`ti.query_kernel_profile_info(name)` now calls `profiler.query_info(name)` 
`ti.clear_kernel_profile_info()` now calls `profiler.clear_info()`
`ti.kernel_profiler_total_time()` now calls `profiler.get_total_time()`

```
=========================================================================
CUDA Profiler(count)
=========================================================================
[      %     total   count |      min       avg       max   ] Kernel name
[100.00%   0.000 s     10x |    0.025     0.025     0.026 ms] memcpy_c8_0_kernel_4_range_for
------------------------------------------------------------------------
[100.00%] Total kernel execution time:   0.253 s   number of records:  1
=========================================================================
```

```
====================================
CUDA Profiler(trace)
====================================
[      % |     time    ] Kernel name
[ 10.10% |    0.026  ms] memcpy_c8_0_kernel_4_range_for
[  9.92% |    0.025  ms] memcpy_c8_0_kernel_4_range_for
[  9.82% |    0.025  ms] memcpy_c8_0_kernel_4_range_for
[  9.82% |    0.025  ms] memcpy_c8_0_kernel_4_range_for
[ 10.25% |    0.026  ms] memcpy_c8_0_kernel_4_range_for
[  9.97% |    0.025  ms] memcpy_c8_0_kernel_4_range_for
[  9.91% |    0.025  ms] memcpy_c8_0_kernel_4_range_for
[  9.95% |    0.025  ms] memcpy_c8_0_kernel_4_range_for
[ 10.31% |    0.026  ms] memcpy_c8_0_kernel_4_range_for
[  9.95% |    0.025  ms] memcpy_c8_0_kernel_4_range_for
====================================
```

PR #3504, opened 2021-11-14
[refactor] Decouple KernelProfilerBase::sync() from Program::synchronize()

yolo2themoon:
Related issue = #2922

### DIFF
--- a/taichi/backends/cuda/cuda_context.cpp
+++ b/taichi/backends/cuda/cuda_context.cpp
@@ -84,7 +84,7 @@ void CUDAContext::launch(void *func,
 
   KernelProfilerBase::TaskHandle task_handle;
   // Kernel launch
-  if (profiler_ && profiler_->is_enabled()) {
+  if (profiler_) {
     KernelProfilerCUDA *profiler_cuda =
         dynamic_cast<KernelProfilerCUDA *>(profiler_);
     profiler_cuda->trace(task_handle, task_name, func, grid_dim, block_dim, 0);
@@ -111,7 +111,7 @@ void CUDAContext::launch(void *func,
                           dynamic_shared_mem_bytes, nullptr,
                           arg_pointers.data(), nullptr);
   }
-  if (profiler_ && profiler_->is_enabled())
+  if (profiler_)
     profiler_->stop(task_handle);
 
   if (debug_) {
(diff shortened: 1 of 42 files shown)


taichi/backends/cuda/cuda_context.cpp
```diff
@@ -84,7 +84,7 @@ void CUDAContext::launch(void *func,
 
   KernelProfilerBase::TaskHandle task_handle;
   // Kernel launch
-  if (profiler_) {
+  if (profiler_ && profiler_->is_enabled()) {
```
qiao-bo, line 87: Here might need a bit clarification: when profiler_ is not null, i assume it's should be enabled? and vice versa?

qiao-bo (commented):
nice, thanks!

taichi/backends/cuda/cuda_context.cpp
yolo2themoon: It seems that from the beginning make_profiler has been designed to return valid pointers
[taichi/kernel_profiler.cpp at v0.7.10 ](https://github.com/taichi-dev/taichi/blob/v0.7.10/taichi/program/kernel_profiler.cpp#L174) vs [taichi/kernel_profiler.cpp at master](https://github.com/taichi-dev/taichi/blob/master/taichi/program/kernel_profiler.cpp#L116)
[taichi/program.cpp at v0.7.10](https://github.com/taichi-dev/taichi/blob/v0.7.10/taichi/program/program.cpp#L123)  vs  [taichi/program.cpp at master](https://github.com/taichi-dev/taichi/blob/master/taichi/program/program.cpp#L71)

For CUDA backends, the difference is whether or not the profiler_ in the CUDAContext is set.
If kernel_profiler is not set, it will not be executed when kernels are launched.
[taichi/program.cpp at v0.7.10](https://github.com/taichi-dev/taichi/blob/v0.7.10/taichi/program/program.cpp#L130)  vs  [taichi/llvm_program.cpp at master](https://github.com/taichi-dev/taichi/blob/master/taichi/llvm/llvm_program.cpp#L94)

I tried to return `nullptr` but test on m1 did not pass [taichi-dev/taichi@f278285](https://github.com/taichi-dev/taichi/actions/runs/1456235456)

So I added an additional conditional statement to prevent external modifications from causing the profiler to be called accidentally

ailzhang: Hmm but this check complicates the logic here, @yolo2themoon do we know why m1 failed with returning nullptr?

yolo2themoon: removed the enable-related code, thanks to @ailzhang 's help in debugging

qiao-bo (approved):
LGTM!

ailzhang (approved):
Nice! Thanks!

PR #3507, opened 2021-11-14
[llvm] Add a TI_WITH_LLVM option

AmesingFlank:
This PR adds a TI_WITH_LLVM option, so that one can optionally disable compiling with LLVM. (For example, I'm trying to emscripten the entire taichi project, and LLVM doesn't like that).

TI_WITH_LLVM is defaulted to ON, so this PR won't affect any of our existing CI/CD/build tools.

### DIFF
--- a/cmake/TaichiCore.cmake
+++ b/cmake/TaichiCore.cmake
@@ -1,4 +1,5 @@
 option(USE_STDCPP "Use -stdlib=libc++" OFF)
+option(TI_WITH_LLVM "Build with LLVM backends" OFF)
 option(TI_WITH_CUDA "Build with the CUDA backend" ON)
 option(TI_WITH_CUDA_TOOLKIT "Build with the CUDA toolkit" OFF)
 option(TI_WITH_OPENGL "Build with the OpenGL backend" ON)
@@ -42,6 +43,10 @@ endif()
 if (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/external/glad/src/gl.c")
     set(TI_WITH_OPENGL OFF)
     message(WARNING "external/glad submodule not detected. Settings TI_WITH_OPENGL to OFF.")
+endif()
+
+if(NOT TI_WITH_LLVM)
+    set(TI_WITH_CUDA OFF)
 endif()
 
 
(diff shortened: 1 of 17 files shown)


cmake/TaichiCore.cmake
```diff
@@ -44,6 +45,10 @@ if (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/external/glad/src/gl.c")
     message(WARNING "external/glad submodule not detected. Settings TI_WITH_OPENGL to OFF.")
 endif()
 
+if(NOT TI_WITH_LLVM)
+    set(TI_WITH_CUDA OFF)
```
qiao-bo, line 49:
```suggestion
    set(TI_WITH_CUDA OFF)
    set(TI_WITH_CUDA_TOOLKIT OFF)
```

qiao-bo (commented):
Thanks. A quick grep into our cpp and h files shows 263 occurrences of `endif`, many of which are small blocks. I feel like this may hinder readability in the long term.

AmesingFlank:
> Thanks. A quick grep into our cpp and h files shows 263 occurrences of `endif`, many of which are small blocks. I feel like this may hinder readability in the long term.

I agree :(( The right thing to do here would be to:
* Move `taichi/llvm` `taichi/codegen` (and some other stuff such as SNodeBufferManager) into a folder in `taichi/backend`, and use cmake to omit compiling that folder
* Keep refactoring `program.cpp` until it is fully backend-independent (i.e. remove all the llvm stuff).

I'm happy to create an issue to track this, but it would be a rather big effort.

qiao-bo:
> > Thanks. A quick grep into our cpp and h files shows 263 occurrences of `endif`, many of which are small blocks. I feel like this may hinder readability in the long term.
> 
> I agree :(( The right thing to do here would be to:
> 
> * Move `taichi/llvm` `taichi/codegen` (and some other stuff such as SNodeBufferManager) into a folder in `taichi/backend`, and use cmake to omit compiling that folder
> * Keep refactoring `program.cpp` until it is fully backend-independent (i.e. remove all the llvm stuff).
> 
> I'm happy to create an issue to track this, but it would be a rather big effort.

Indeed a big effort. Having a llvm folder in taichi/backend seems reasonable to me. This can also move some llvm-specific code from device.h/cpp. For now, i think this is fine.

qiao-bo (approved):
LGTM!

PR #3509, opened 2021-11-15
[ci] Add a helper script for Dockerfile generation.

rexwangcc:
Related issue = #

- Implement a helper script for easier Dockerfile generation.


ci/Dockerfile_generator.py
```diff
@@ -0,0 +1,286 @@
...
+        def f(os: str, version: str) -> str:
+            info(f"OS: {os}, version: {version}")
+            base_block = CPU_BASE_BLOCK.format(os=os, version=version)
+            scripts_block = SCRIPTS_BLOCK.format(
+                script=f"{os}_build_test_cpu.sh")
+            install_block = CPU_APT_INSTALL_BLOCK
+
+            # ubuntu 18.04 needs special treatments
+            if os == "ubuntu" and version == "18.04":
+                install_block += """ \\
```
qiao-bo, line 245:
```suggestion
                install_block = install_block.rstrip() + """ \\
```

ci/Dockerfile_generator.py
```diff
@@ -0,0 +1,286 @@
...
+
+        def f(os: str, version: str) -> str:
+            info(f"OS: {os}, version: {version}")
+            base_block = GPU_BASE_BLOCK.format(version=version)
+            scripts_block = SCRIPTS_BLOCK.format(script=f"{os}_build_test.sh")
+            install_block = GPU_APT_INSTALL_BLOCK
+
+            # ubuntu 20.04 needs special treatments
+            if os == "ubuntu" and version == "20.04":
+                install_block += """ \\
```
qiao-bo, line 267:
```suggestion
                install_block = install_block.rstrip() + """ \\
```

qiao-bo (commented):
Gerat, thanks! just some nits newline thingy

qiao-bo:
@rexwangcc do we need to commit the changes on the Dockerfiles as well?

rexwangcc:
> @rexwangcc do we need to commit the changes on the Dockerfiles as well?

Great question! I prefer to do it since the updated Dockerfiles are the results of the script, and we need to do it anyways from when we start to use the script, what do you think?

qiao-bo:
> > @rexwangcc do we need to commit the changes on the Dockerfiles as well?
> 
> Great question! I prefer to do it since the updated Dockerfiles are the results of the script, and we need to do it anyways from when we start to use the script, what do you think?

sounds good, are they also the latest? 'cause we had a change in lines.

rexwangcc:
> > > @rexwangcc do we need to commit the changes on the Dockerfiles as well?
> > 
> > 
> > Great question! I prefer to do it since the updated Dockerfiles are the results of the script, and we need to do it anyways from when we start to use the script, what do you think?
> 
> sounds good, are they also the latest? 'cause we had a change in lines.

will give it a try, one sec.

qiao-bo (approved):
Great! Thanks a lot!

PR #3510, opened 2021-11-15
[metal] Add a TI_WITH_METAL option

AmesingFlank:
Add a TI_WITH_METAL option, defaulted to true, so that one can optionally disable building for metal.

### DIFF
--- a/cmake/TaichiCore.cmake
+++ b/cmake/TaichiCore.cmake
@@ -130,7 +130,7 @@ endif()
 
 # By default, TI_WITH_METAL is ON for all platforms.
 # As of right now, on non-macOS platforms, the metal backend won't work at all.
-# We have future plans to allow metal AOT to run on non-macOS machines.
+# We have future plans to allow metal AOT to run on non-macOS devices.
 if (TI_WITH_METAL)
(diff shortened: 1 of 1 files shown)


sjwsl:
Do we also need to disable this option and raise a warning on non-macOS? Like
https://github.com/taichi-dev/taichi/blob/00b66a8c633cb2e71e7c76ed0473c9cfb5dc90a5/cmake/TaichiCore.cmake#L16-L19

AmesingFlank:
> Do we also need to disable this option and raise a warning on non-macOS? Like
> 
> https://github.com/taichi-dev/taichi/blob/00b66a8c633cb2e71e7c76ed0473c9cfb5dc90a5/cmake/TaichiCore.cmake#L16-L19

@k-ye thoughts? I understand that metal has an AOT module so it might be useful to have the metal backend even when we're not on macOS.

ailzhang (approved):
LGTM!

ailzhang:
@AmesingFlank Metal AOT isn't fully working  without a metal runtime (due to jit evaluators) yet but we plan to fix it later :P 
So yea I'm fine with either way as long as we have a note explaining its behavior :D

PR #3516, opened 2021-11-15
[refactor] Enable the new ast builder by default

lin-hitonami:
issue = #3038

When fixing pylint R1705 no-else-return, some of the raise statements are moved outside the `else` scope of the static-if, resulting raising errors when traversing to the Raise node. For example:
```python
    if ti.static(A.n == 2): 
        ret = svd2d(A, dt)
        return ret
    elif ti.static(A.n == 3):
        return svd3d(A, dt)
    else:
        raise Exception("SVD only supports 2D and 3D matrices.")
```
was changed to 
```python
    if ti.static(A.n == 2):
        ret = svd2d(A, dt)
        return ret
    if ti.static(A.n == 3):
        return svd3d(A, dt)
    raise Exception("SVD only supports 2D and 3D matrices.")
```
and the AST builder will continue to visit the Raise node because it is not in the else branch of static-if.
Immediate termination of traversing when meeting return statement can be supported in the future.

### DIFF
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -396,7 +396,7 @@ class _SpecialConfig:
         self.gdb_trigger = False
         self.excepthook = False
         self.experimental_real_function = False
-        self.experimental_ast_refactor = False
+        self.experimental_ast_refactor = True
 
 
 def prepare_sandbox():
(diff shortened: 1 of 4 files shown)


g1n0st (approved):
Thanks!

ailzhang (approved):
Thanks!

strongoier (approved):
Thanks!

PR #3517, opened 2021-11-15
[Lang] Enable local tensors as writeback binary operation results

strongoier:

This PR rewrites `element_wise_writeback_binary` to enable local tensor results. To make it work, `Matrix.__init__` is refactored a little bit to clean the logic and cover missing cases.


k-ye (approved):
LGTM!

PR #3521, opened 2021-11-15
[misc] Fix redefined reset function (#3227)

### LINKED ISSUE #3227: [misc] Function redifined in python/taichi/lang/__init__.py
**Describe the bug**

`python/taichi/lang/__init__.py`:

```python
from taichi.lang.impl import *

def reset():
    """Resets Taichi to its initial state.

    This would destroy all the fields and kernels.
    """
    _ti_core.reset_snode_access_flag()
    impl.reset()
    global runtime
    runtime = impl.get_runtime()
```

There is a func `reset` in python/taichi/lang/__init__.py, which is already defined at taichi.lang.impl

`python/taichi/lang/impl.py`:

```python
def reset():
    global pytaichi
    old_kernels = pytaichi.kernels
    pytaichi.clear()
    pytaichi = PyTaichi(old_kernels)
    for k in old_kernels:
        k.reset()
    _ti_core.reset_default_compile_config()

```

The IDE always raises an error here.

u2386:
Related issue = close #3227

### DIFF
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -12,29 +12,48 @@ import taichi.lang.meta
 from taichi.core.util import locale_encode
 from taichi.core.util import ti_core as _ti_core
 from taichi.lang import _random, impl, types
+from taichi.lang._ndarray import ScalarNdarray
+from taichi.lang.any_array import AnyArray, AnyArrayAccess
 from taichi.lang.ast.transformer import TaichiSyntaxError
 from taichi.lang.enums import Layout
 from taichi.lang.exception import InvalidOperationError
-from taichi.lang.impl import *
+from taichi.lang.expr import Expr, make_expr_group
+from taichi.lang.field import Field, ScalarField
+from taichi.lang.impl import (axes, begin_frontend_if,
+                              begin_frontend_struct_for, call_internal,
+                              chain_compare, current_cfg, expr_init,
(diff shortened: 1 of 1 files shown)


ailzhang (commented):
nice catch! I'm happy to see this `impl` gone, tho I'm more concerned with `from taichi.lang.impl import *` than this line :P 
I might prefer to make the line below more explicit (instead of `import *`) as a followup, what do you think?

u2386:
> make the line below more explicit

Thanks for your suggestion. I believe that this is the right way to fix the issue and we should find out imports that are defined in `taichi.lang.impl` but imported directly from `taichi.lang`. Working on it 😆

ailzhang:
@u2386 Ah I guess removing `from taichi.lang.impl import *` indeed is not an easy task since it affects what we expose from `taichi` package. 
Maybe defining a `__all__` in taichi.lang.impl is cleaner than importing all of them explicitly. wdyt?

u2386:
To add an `__all__` in taichi.lang.impl,  we also need to find all the exposed imports, and this way may not work either as `import *` also exposes what taichi.lang.impl package imports directly.

I grepped all references which prefix with `ti.` and check if they are defined in taichi.lang.impl one by one.

I think now it's equivalent to the way it was before. May you approve the CI pipeline.

ailzhang (approved):
Nice thanks a lot for fixing it!

PR #3523, opened 2021-11-15
[vulkan] Fix a few errors encountered when building for x86

AmesingFlank:
Lessons learned:
* Always prefer `VK_NULL_HANDLE` to `nullptr`
* Always prefer `VkDeviceSize` to `size_t`


k-ye (approved):
Thanks!

PR #3524, opened 2021-11-16
[Lang] Remove disable_local_tensor in most cases

strongoier:

After this PR, `disable_local_tensor` only appears in `ti.Matrix.empty`. Removing that requires significant refactoring, which will be done in a future PR.


ailzhang (approved):
Nice!

PR #3526, opened 2021-11-16
[Lang] Add check version function to taichi main

Leonz5288:
Initial version of checking version for users when using Taichi. The method of getting users' platforms might looks hacky. Welcome any suggestions on how to map users' platform.

### DIFF
--- a/python/taichi/main.py
+++ b/python/taichi/main.py
@@ -127,8 +127,10 @@ class TaichiMain:
 
         # We do not want request exceptions break users' usage of Taichi.
         try:
-            response = requests.post('http://ec2-54-90-48-192.compute-1.amazonaws.com/check_version',
-                                     json=payload, timeout=5)
+            response = requests.post(
+                'http://ec2-54-90-48-192.compute-1.amazonaws.com/check_version',
+                json=payload,
+                timeout=5)
             response.raise_for_status()
         except requests.exceptions.ConnectionError as err:
             print('No internet:', err)
@@ -146,9 +148,9 @@ class TaichiMain:
         response = response.json()
         if response['status'] == 1:
             print(
-                f'Your Taichi version {version} is outdated. The latest version is {response["latest_version"]}, you can use\n'+
-                f'pip install taichi=={response["latest_version"]}\n'+'to upgrade to the latest Taichi!'
-            )
+                f'Your Taichi version {version} is outdated. The latest version is {response["latest_version"]}, you can use\n'
+                + f'pip install taichi=={response["latest_version"]}\n' +
+                'to upgrade to the latest Taichi!')
         elif response['status'] == 0:
             # Status 0 means that user already have the latest Taichi. The message here prompts this infomation to users.
             print(response['message'])
--- a/python/taichi/main.py
+++ b/python/taichi/main.py
@@ -1,4 +1,5 @@
 import argparse
+import json
 import math
 import os
 import platform
@@ -7,13 +8,12 @@ import shutil
 import subprocess
 import sys
 import timeit
-import requests
-import json
 from collections import defaultdict
 from functools import wraps
 from pathlib import Path
 
 import numpy as np
+import requests
 import taichi.cc_compose
 import taichi.diagnose
 from colorama import Fore
--- a/python/taichi/main.py
+++ b/python/taichi/main.py
@@ -1,5 +1,4 @@
 import argparse
-import json
 import math
 import os
 import platform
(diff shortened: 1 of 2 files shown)


python/taichi/main.py
```diff
@@ -88,6 +91,48 @@ def __call__(self):
...
+            payload['python'] = 'cp37'
+        elif python_version.startswith('3.8'):
+            payload['python'] = 'cp38'
+        elif python_version.startswith('3.9'):
+            payload['python'] = 'cp39'
+
+        import requests
+        response = requests.post('http://54.90.48.192/check_version', json=payload)
+        import json
+        response_json = json.loads(response.text)
```
yihong0618, line 131: we do not need import json
requests can use like:
```
j = response.json()
```

python/taichi/main.py
```diff
@@ -88,6 +91,48 @@ def __call__(self):
...
         return getattr(self, args.command)(sys.argv[2:])
 
+    @staticmethod
+    def _check_version():
+        # Check Taichi version for the user.
+        print('Checking your Taichi version...')
+        major = _ti_core.get_version_major()
+        minor = _ti_core.get_version_minor()
+        patch = _ti_core.get_version_patch()
+        version = '{}.{}.{}'.format(major, minor, patch)
```
yihong0618: f-string for pylint?

python/taichi/main.py
```diff
@@ -88,6 +91,48 @@ def __call__(self):
...
+            payload['python'] = 'cp38'
+        elif python_version.startswith('3.9'):
+            payload['python'] = 'cp39'
+
+        import requests
+        response = requests.post('http://54.90.48.192/check_version', json=payload)
+        import json
+        response_json = json.loads(response.text)
+        if response_json['status'] == 1:
+            print('Your Taichi version {} is outdated. The latest version is {}, you can use pip to upgrade to the latest Taichi!'.format(version, response_json['latest_version']))
```
yihong0618: f-string

python/taichi/main.py
Leonz5288: Great, thanks for the info!

python/taichi/main.py
```diff
@@ -88,6 +92,48 @@ def __call__(self):
...
+
+        response = requests.post('http://54.90.48.192/check_version',
+                                 json=payload).json()
+        if response['status'] == 1:
+            print(
+                f'Your Taichi version {version} is outdated. The latest version is {response["latest_version"]}, you can use pip to upgrade to the latest Taichi!'
+            )
+        elif response['status'] == 0:
+            print(response['message'])
+
```
rexwangcc, line 136: Several unexpected things could happen during an HTTP call, it can timeout, the response might not be JSON-serializable, etc... I'd probably:

- add a timeout to the POST method (which might raise an exception)
- store the response as is and use something like `response.raise_for_status()` before trying to parse the response into JSON. (which might raise a different type of exception)
- Wrap the caller of this static method in a try catch finally block to catch the above exceptions so even if the HTTP call failed, user won't get exposed to some bizarre issues due to Network failure.

WDYT?

rexwangcc: optional nit: we may want to add the link to our docs or print the explicit pip install command directly in the helper message for users.

requirements_test.txt
```diff
@@ -5,3 +5,4 @@ pytest-rerunfailures
 pytest-cov
 numpy
 autograd
+requests
```
rexwangcc, line 8: nit: HTTP clients constantly break, I'd suggest adding a constraint on the versions 😄

python/taichi/main.py
```diff
@@ -88,6 +92,48 @@ def __call__(self):
...
+        if python_version.startswith('3.6'):
+            payload['python'] = 'cp36'
+        elif python_version.startswith('3.7'):
+            payload['python'] = 'cp37'
+        elif python_version.startswith('3.8'):
+            payload['python'] = 'cp38'
+        elif python_version.startswith('3.9'):
+            payload['python'] = 'cp39'
+
+        response = requests.post('http://54.90.48.192/check_version',
```
rexwangcc, line 128: Maybe hold on merging this PR until the IP gets a TLS+DNS, random ip address scares ppl :P

rexwangcc (commented):
Looks great, thanks! I left my 2 cents :octocat:

python/taichi/main.py
```diff
@@ -88,6 +92,48 @@ def __call__(self):
...
+                                 json=payload).json()
+        if response['status'] == 1:
+            print(
+                f'Your Taichi version {version} is outdated. The latest version is {response["latest_version"]}, you can use pip to upgrade to the latest Taichi!'
+            )
+        elif response['status'] == 0:
+            print(response['message'])
```
rexwangcc, line 135: What's expected in the `message`? Could you add a brief explanation to the docstring? Thanks!

requirements_test.txt
Leonz5288: OK

python/taichi/main.py
Leonz5288: What about the address that contains 'aws' and 'amazon'?

python/taichi/main.py
Leonz5288: Of course

python/taichi/main.py
rexwangcc: That may be better, but still a bit awkward, but we can update it later.

rexwangcc (approved):

python/taichi/main.py
```diff
@@ -11,6 +12,7 @@
 from pathlib import Path
 
 import numpy as np
+import requests
```
AmesingFlank, line 15: need to add this to `requirements.txt`

PR #3530, opened 2021-11-16
[Refactor] Fix typo and type in docstring, and format too long string

gaoxinge:
Related issue = #


python/taichi/lang/matrix.py
```diff
@@ -26,9 +26,9 @@ class Matrix(TaichiOperations):
     """The matrix class.
 
     Args:
-        n (int): the first dimension of a matrix.
+        n (int, list, tuple, np.ndarray): the first dimension of a matrix.
```
ailzhang, line 29: nit: `n (Union[int, list, tuple], np.ndarray)`

ailzhang (approved):
Looks great! Thanks! 
(with one minor suggestion ;)

python/taichi/lang/matrix.py
gaoxinge: @ailzhang I have improve it.

PR #3531, opened 2021-11-16
[refactor] Remove variable() of Matrix/Struct and empty() of Matrix/StructType

strongoier:
Related issue = #2880

There are lots of dependencies of the `ti.Matrix.empty` internal method (#3524). This PR serves as the first step towards cleaning them up.

### DIFF
--- a/python/taichi/lang/struct.py
+++ b/python/taichi/lang/struct.py
@@ -305,7 +305,7 @@ class IntermediateStruct(Struct):
     """The Struct type class for compiler internal use only.
 
     Args:
-        entries (Dict[str, Union[Dict, Expr, Matrix, Struct]]): keys and values for struct members.
+        entries (Dict[str, Union[Expr, Matrix, Struct]]): keys and values for struct members.
     """
     def __init__(self, entries):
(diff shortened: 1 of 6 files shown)


python/taichi/lang/struct.py
```diff
@@ -542,41 +513,37 @@ def __call__(self, *args, **kwargs):
...
+                    entries[k] = cast(struct.entries[k], dtype)
+        return Struct(entries)
 
-    def empty(self):
-        """
-        Create an empty instance of the given compound type.
-        Nested structs and matrices need to be recursively handled.
-        """
-        struct = Struct.empty(self.members.keys())
+    def scalar_filled(self, value):
```
lin-hitonami, line 539: How about renaming to filled_with_scalar or something like that? IMHO scalar_filled maybe has a tendency to be confused with scalar field

strongoier: Sounds good.

strongoier: done

lin-hitonami (approved):
LGTM!

PR #3538, opened 2021-11-17
[bug] Remove fallback in C++ code

lin-hitonami:
Related issue = #
The fallback-related code in C++ is broken and deprecated, so we should remove it.

### DIFF
--- a/taichi/program/program.cpp
+++ b/taichi/program/program.cpp
@@ -76,27 +76,18 @@ Program::Program(Arch desired_arch)
     TI_ERROR("This taichi is not compiled with LLVM");
 #endif
   } else if (config.arch == Arch::metal) {
-    if (!metal::is_metal_api_available()) {
-      TI_ERROR("No Metal API detected.");
-    } else {
-      program_impl_ = std::make_unique<MetalProgramImpl>(config);
-    }
+    TI_ASSERT(metal::is_metal_api_available());
+    program_impl_ = std::make_unique<MetalProgramImpl>(config);
   } else if (config.arch == Arch::vulkan) {
 #ifdef TI_WITH_VULKAN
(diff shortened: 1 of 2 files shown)


taichi/program/program.cpp
```diff
@@ -73,53 +73,44 @@ Program::Program(Arch desired_arch)
...
     program_impl_ = std::make_unique<LlvmProgramImpl>(config, profiler.get());
 #else
-    TI_NOT_IMPLEMENTED
+    TI_ERROR("This taichi is not compiled with LLVM");
 #endif
   } else if (config.arch == Arch::metal) {
     if (!metal::is_metal_api_available()) {
-      TI_WARN("No Metal API detected.");
-      config.arch = host_arch();
+      TI_ERROR("No Metal API detected.");
```
ailzhang, line 80: Shall we just do TI_ASSERT(metal::is_metal_api_available()) and get rid of the if-else? ;)

taichi/program/program.cpp
```diff
@@ -73,53 +73,44 @@ Program::Program(Arch desired_arch)
...
+    TI_ERROR("No C backend detected.");
 #endif
+  } else {
+    TI_NOT_IMPLEMENTED
   }
 
-  if (config.arch != desired_arch) {
-    TI_WARN("Falling back to {}", arch_name(config.arch));
-  }
+  TI_ERROR_IF(!program_impl_, "Backend not found.");
```
ailzhang, line 110: ditto :D

ailzhang: Would you mind also adding a comment about invariant program_impl_ is not empty?

ailzhang (approved):
a few nits, otherwise looks great! Thanks!

PR #3540, opened 2021-11-17
[CI] Add version database update in CD

Leonz5288:
In release.yml, we will automatically update version database in every production release. We will need to add two new secrets to the repository though.

### DIFF
--- a/misc/save_new_version.py
+++ b/misc/save_new_version.py
@@ -16,5 +16,23 @@ username = os.getenv('METADATA_USERNAME')
 password = os.getenv('METADATA_PASSWORD')
 url = os.getenv('METADATA_URL')
 
-response = requests.post('http://'+url+'/add_version/main', json=payload, auth=(username, password))
-print(response.text)
+try:
+    response = requests.post('http://'+url+'/add_version/main', json=payload, auth=(username, password), timeout=5)
+    response.raise_for_status()
+except requests.exceptions.ConnectionError as err:
+    print('Updating latest version failed: No internet,', err)
+    exit(1)
+except requests.exceptions.HTTPError as err:
+    print('Updating latest version failed: Server error,', err)
+    exit(1)
+except requests.exceptions.Timeout as err:
+    print(
+        'Updating latest version failed: Time out when connecting server,',
+        err)
+    exit(1)
+except requests.exceptions.RequestException as err:
+    print('Updating latest version failed:', err)
+    exit(1)
+
+response = response.json()
+print(response['message'])
--- a/misc/save_new_version.py
+++ b/misc/save_new_version.py
@@ -24,7 +24,10 @@ password = os.getenv('METADATA_PASSWORD')
 url = os.getenv('METADATA_URL')
 
 try:
-    response = requests.post('http://'+url+'/add_version/main', json=payload, auth=(username, password), timeout=5)
+    response = requests.post('http://' + url + '/add_version/main',
+                             json=payload,
+                             auth=(username, password),
+                             timeout=5)
     response.raise_for_status()
 except requests.exceptions.ConnectionError as err:
     print('Updating latest version failed: No internet,', err)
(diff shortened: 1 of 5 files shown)


misc/save_new_version.py
```diff
@@ -0,0 +1,28 @@
...
+    'major': major,
+    'minor': minor,
+    'patch': patch,
+    'date': release_date
+}
+
+username = os.getenv('METADATA_USERNAME')
+password = os.getenv('METADATA_PASSWORD')
+
+response = requests.post('http://54.90.48.192/add_version/main',
```
rexwangcc, line 25: How about making the IP address (i.e. `http://54.90.48.192`) an ENV VAR (e.g. `METADATA_SERVICE_URL`) as well so later if we update it to something like HTTPS://DOMAIN we don't need to change the code again? Same for the other place.

rexwangcc: Side note: could you add all newly created credentials (username, password, url, etc.) to Vault? (and we may let Action authenticate with Vault in the future) Thanks!

misc/save_new_version.py
```diff
@@ -0,0 +1,28 @@
...
+                         json=payload,
+                         auth=requests.auth.HTTPBasicAuth(username, password))
+print(response.text)
```
rexwangcc, line 28: Curious: what would happen if the response code is not 200/201? Do we need a `response.raise_for_status()` and/or try catch here? Same question applies to the other place.

rexwangcc (commented):
Thanks! LGTM in general, I added a few comments.

misc/save_new_version.py
Leonz5288: Added to vault

Leonz5288: > How about making the IP address (i.e. `http://54.90.48.192`) an ENV VAR (e.g. `METADATA_SERVICE_URL`) as well so later if we update it to something like [HTTPS://DOMAIN](HTTPS://DOMAIN) we don't need to change the code again? Same for the other place.

Good point.

python/build.py
```diff
@@ -81,6 +81,15 @@ def parse_args():
...
     return parser.parse_args()
 
+def upload_taichi_version():
+    username = os.getenv('METADATA_USERNAME')
+    password = os.getenv('METADATA_PASSWORD')
+    filename = os.listdir('../dist')[0]
+    filename = filename[:len(filename)-4]
+    parts = filename.split('-')
+    payload = {'version': parts[1], 'platform': parts[4], 'python': parts[2]}
+    response = requests.post('http://54.90.48.192/add_version/detail', json=payload, auth=requests.auth.HTTPBasicAuth(username, password))
```
yihong0618: for HTTPBasicAuth 
auth=(username, password)  is fine

Leonz5288: Thanks!

python/build.py
```diff
@@ -82,6 +84,38 @@ def parse_args():
...
+def upload_taichi_version():
+    username = os.getenv('METADATA_USERNAME')
+    password = os.getenv('METADATA_PASSWORD')
+    url = os.getenv('METADATA_URL')
+    filename = os.listdir('../dist')[0]
+    filename = filename[:len(filename) - 4]
+    parts = filename.split('-')
+    payload = {'version': parts[1], 'platform': parts[4], 'python': parts[2]}
+    try:
+        response = requests.post('http://' + url + '/add_version/detail',
```
rexwangcc, line 96:
```suggestion
        response = requests.post(f'http://{url}/add_version/detail',
```

misc/save_new_version.py
```diff
@@ -0,0 +1,47 @@
...
+    'patch': patch,
+    'date': release_date
+}
+
+username = os.getenv('METADATA_USERNAME')
+password = os.getenv('METADATA_PASSWORD')
+url = os.getenv('METADATA_URL')
+
+try:
+    response = requests.post('http://' + url + '/add_version/main',
```
rexwangcc, line 27:
```suggestion
    response = requests.post(f'http://{url}/add_version/main',
```

rexwangcc (approved):
LGTM!

PR #3541, opened 2021-11-17
[Lang] Add ti.round op

gaoxinge:
Related issue = #


python/taichi/lang/ops.py
```diff
@@ -271,6 +271,19 @@ def _rsqrt(a):
...
 
+@unary
+def round(a):
+    """The round function.
+
+    Args:
+        a (Union[:class:`~taichi.lang.expr.Expr`, :class:`~taichi.lang.matrix.Matrix`]): A number or a matrix.
+
+    Returns:
+        Round of `a`.
```
strongoier, line 282:
```suggestion
        The nearest integer of `a`.
```

strongoier (commented):
Looks great! Thanks for your work! A little suggestion: could you also add the op in `backends/vulkan/codegen_vulkan.cpp` and `backends/metal/data_types.cpp`?

python/taichi/lang/ops.py
gaoxinge: @strongoier Follow your advice, I modify the docstring for more precise meaning.

gaoxinge:
@strongoier I have add round op code in `backends/vulkan/codegen_vulkan.cpp` (follow by [link](https://www.khronos.org/registry/SPIR-V/specs/1.0/GLSL.std.450.html)) and `backends/metal/data_types.cpp`, but I'm not sure whether it will work properly.

strongoier (commented):
Your modifications look pretty good to me. To verify your modifications, you can remove https://github.com/taichi-dev/taichi/blob/17a0a119bad15e5caeaf242548a72476f643fd7e/tests/python/test_element_wise.py#L245 and see if CI can pass. However, our CI seems buggy on your PR (#3568), and you may need to wait for a fix for that. Sorry about that.

strongoier:
Now our CI is fixed. You could try my comment above now.

strongoier (approved):
LGTM! Thanks!

PR #3542, opened 2021-11-17
[ci] Dockerfile for CPU manylinux2014 compliant

qiao-bo:
This dockerfile enables a CentOS 7 based environment to build Taichi (CPU only). The resulting wheel is fully [manylinux2014](https://www.python.org/dev/peps/pep-0599/) compliant.


ci/scripts/manylinux_build_wheel.sh
```diff
@@ -0,0 +1,34 @@
...
+TAICHI_CMAKE_ARGS="-DTI_WITH_VULKAN:BOOL=OFF -DTI_WITH_CUDA:BOOL=OFF -DTI_WITH_OPENGL:BOOL=OFF -DTI_WITH_CC:BOOL=OFF" python3 setup.py install
+
+# Add Docker specific ENV
+export TI_IN_DOCKER=true
+
+# Run tests
+ti diagnose
+
+# Build wheel
+cd python && python build.py build
```
frostming, line 34: This performs build twice(the first at Line 25). And `build.py` is to be removed by #3537 .

Maybe we can modify Line 25 to build the wheel and install it in two steps.

qiao-bo: sure, we can stay aligned here.

rexwangcc (approved):
I was able to build and test the Dockerfile locally, LGTM!

ailzhang (approved):
LGTM!

ci/Dockerfile.manylinux.cpu
```diff
@@ -0,0 +1,57 @@
...
+USER dev
+
+# Install miniconda
+RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
+    bash Miniconda3-latest-Linux-x86_64.sh -p /home/dev/miniconda -b
+ENV PATH="/home/dev/miniconda/bin:$PATH"
+
+# Set up multi-python environment
+RUN conda init bash
+RUN conda create -n py36 python=3.6 -y
```
messense, line 47: FYI, Python 3.6-3.10 is already installed in /`opt/python` so conda may not be necessary.

qiao-bo: @messense, thx, good to know that!

PR #3546, opened 2021-11-18
[Lang] Remove disable_local_tensor and empty() from Matrix

strongoier:

After this PR, dynamic index of local tensors are completely supported with `dynamic_index=True`.


ailzhang (approved):
Nicee! Thanks!

PR #3547, opened 2021-11-18
[llvm] Make taichi's Ndarray carry a ptr to its DeviceAllocation.

ailzhang:
Sorry I left a few todos in this PR since it enables some unification
between cpu/cuda backend, as well as some cleanups.
But they're not relevant to the goal of this PR so I'll send them
separately.

This PR mainly enforces an invariant that:
- Taichi ndarrays' ptr points to a DeviceAllocation instead of the raw
data like torch-based ndarray.

The invariant above makes it easier to unify our Ndarray implementation
between multiple backends. I'll add an opengl implementation of ndarray
on top of this.

Related issue = #

### DIFF
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -62,31 +62,55 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
       Kernel::LaunchContextBuilder ctx_builder(kernel, &context);
       bool transferred = false;
       for (int i = 0; i < (int)args.size(); i++) {
-        if (args[i].is_external_array && args[i].size > 0) {
-          // Note: both numpy and PyTorch support arrays/tensors with zeros
-          // in shapes, e.g., shape=(0) or shape=(100, 0, 200). This makes
-          // args[i].size = 0.
+        if (args[i].is_external_array) {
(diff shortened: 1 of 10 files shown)


taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -62,31 +62,55 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
...
+                                    ->get_ndarray_alloc_info_ptr(*ptr);
+            // We compare arg_buffers[i] and device_buffers[i] later to check
+            // if transfer happened.
+            // TODO: this logic can be improved but I'll leave it to a followup
+            // PR.
+            arg_buffers[i] = device_buffers[i];
+
+            ctx_builder.set_arg_external_array(i, (uint64)device_buffers[i],
+                                               args[i].size,
+                                               /*is_device_allocation=*/false);
```
qiao-bo, line 112: `false` here is correct?

ailzhang: Yup since we unwraps it and replaces it with a raw ptr, so it's no longer a device allocation.

qiao-bo (commented):
Thanks for the effort!

ailzhang:
I'll close this and open a ghstack PR instead ;)

PR #3549, opened 2021-11-18
[Lang] [bug] Fix numpy from and to ndarray matrix

qiao-bo:
Related issue = #

### DIFF
--- a/python/taichi/lang/_ndarray.py
+++ b/python/taichi/lang/_ndarray.py
@@ -90,33 +90,6 @@ class Ndarray:
         raise NotImplementedError()
 
     @python_scope
-    def fill(self, val):
-        """Fills ndarray with a specific scalar value.
-
-        Args:
-            val (Union[int, float]): Value to fill.
-        """
-        raise NotImplementedError()
-
-    @python_scope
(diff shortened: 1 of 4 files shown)


strongoier (approved):
LGTM!

python/taichi/lang/_ndarray.py
```diff
@@ -89,34 +89,41 @@ def __getitem__(self, key):
...
 
         Returns:
             numpy.ndarray: The result numpy array.
         """
         if impl.current_cfg().ndarray_use_torch:
             return self.arr.cpu().numpy()
+
         arr = np.zeros(shape=self.arr.shape, dtype=to_numpy_type(self.dtype))
-        taichi.lang.meta.ndarray_to_ext_arr(self, arr)
+        if len(self.element_shape) == 0:
```
strongoier, line 115: nit: avoid manually checking this in the parent class

ailzhang (approved):
Thanks for the quick fix!!

python/taichi/lang/_ndarray.py
qiao-bo: @strongoier fixed!

PR #3554, opened 2021-11-19
[llvm] Make taichi's Ndarray carry a ptr to its DeviceAllocation.

ailzhang:
* #3556
* #3555
* __->__ #3554

Sorry I left a few todos in this PR since it enables some unification
between cpu/cuda backend, as well as some cleanups.
But they're not relevant to the goal of this PR so I'll send them
separately.

This PR mainly enforces an invariant that:
- Taichi ndarrays' ptr points to a DeviceAllocation instead of the raw
data like torch-based ndarray.

The invariant above makes it easier to unify our Ndarray implementation
among multiple backends. I'll add an opengl implementation of ndarray
on top of this.

note: we keep the data_ptr_ pointing to a raw ptr (so added a new ptr
for DeviceAllocation* instead) to avoid breaking TNG.


taichi/backends/cuda/codegen_cuda.cpp
```diff
@@ -62,31 +62,57 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
...
+                                    ->get_ndarray_alloc_info_ptr(*ptr);
+            // We compare arg_buffers[i] and device_buffers[i] later to check
+            // if transfer happened.
+            // TODO: this logic can be improved but I'll leave it to a followup
+            // PR.
+            arg_buffers[i] = device_buffers[i];
+
+            ctx_builder.set_arg_external_array(i, (uint64)device_buffers[i],
+                                               args[i].size,
+                                               /*is_device_allocation=*/false);
```
k-ye, line 114: nit: comment on why `false` here?

k-ye (approved):
LGTM!

qiao-bo (approved):
LGTM!

taichi/program/ndarray.cpp
```diff
@@ -31,6 +36,10 @@ intptr_t Ndarray::get_data_ptr_as_int() const {
   return reinterpret_cast<intptr_t>(data_ptr_);
 }
 
+intptr_t Ndarray::get_device_allocation_ptr_as_int() const {
+  return reinterpret_cast<intptr_t>(device_allocation_ptr_);
```
k-ye, line 40: super nit here:

1. If we don't store this as an additional field, we can just `reinterpret_cast<intptr_t>(&device_allocation_ptr_)`
2. If we do, why not just store it as an `intptr_t`...

k-ye (approved):
LGTM!

taichi/program/ndarray.cpp
ailzhang: Right... I was dumb....thanks for catching this!

k-ye (approved):

PR #3555, opened 2021-11-19
[opengl] Support taichi ndarray on opengl backend and enable tests.

ailzhang:
* #3556
* __->__ #3555
* #3554

Note running the same kernel with different shape of taichi ndarrays is
fixed in this PR.

Update: when I move the DeviceAllocation dealloc logic to Ndarray
destructor, I had to share the `Device` object in OpenglRuntime with
Ndarrays so that the destruction order is guaranteed. But I'm open to
other suggestions if there's a better design :D


python/taichi/lang/__init__.py
```diff
@@ -552,6 +552,14 @@ def init(arch=None,
         _ti_core.set_tmp_dir(locale_encode(prepare_sandbox()))
     print(f'[Taichi] Starting on arch={_ti_core.arch_name(ti.cfg.arch)}')
 
+    # Torch based ndarray on opengl backend allocates memory on host instead of opengl backend.
+    # So it won't work.
+    if ti.cfg.arch == opengl and ti.cfg.ndarray_use_torch:
+        ti.warn(
```
k-ye, line 558: Just raise an error?

taichi/program/ndarray.cpp
```diff
@@ -17,11 +17,14 @@ Ndarray::Ndarray(Program *prog,
...
-  LlvmProgramImpl *prog_impl = prog->get_llvm_program_impl();
-  ndarray_alloc_ = prog_impl->allocate_memory_ndarray(nelement_ * element_size_,
-                                                      prog->result_buffer);
-
-  data_ptr_ = prog_impl->get_ndarray_alloc_info_ptr(ndarray_alloc_);
+  ndarray_alloc_ = prog->allocate_memory_ndarray(nelement_ * element_size_,
+                                                 prog->result_buffer);
+
+  if (arch_is_cpu(prog->config.arch) || prog->config.arch == Arch::cuda) {
+    // Keep this information for TNG.
```
k-ye, line 24:
```suggestion
    // For the LLVM backends, device allocation is a physical pointer
```

taichi/backends/opengl/opengl_api.h
```diff
@@ -100,6 +103,9 @@ class DeviceCompiledProgram {
 
   DeviceAllocation args_buf_{kDeviceNullAllocation};
   DeviceAllocation ret_buf_{kDeviceNullAllocation};
+  // Only saves numpy/torch cpu based external array since they don't have
+  // DeviceAllocation.
+  // Taichi |Ndarray| manages their own DeviceAllocation so it's not saved here.
   DeviceAllocation arr_bufs_[taichi_max_num_args]{kDeviceNullAllocation};
```
k-ye, line 109:
```suggestion
  mutable DeviceAllocation arr_bufs_[taichi_max_num_args]{kDeviceNullAllocation};
```

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -340,25 +340,35 @@ void CompiledProgram::set_used(const UsedFeature &used) {
...
       continue;
-    void *baseptr = device_->map(arr_bufs_[i]);
-    if (program_.check_ext_arr_read(i)) {
-      std::memcpy((char *)baseptr, (void *)ctx.args[i], size);
+    if (args[i].size == 0)
+      continue;
+
+    if (!ctx.is_device_allocation[i]) {
+      void *host_ptr = (void *)ctx.args[i];
+      const_cast<DeviceAllocation &>(arr_bufs_[i]) = device_->allocate_memory(
```
k-ye, line 362: Hmm, this leaks SSBO memory. I believe the correct logic is:

1. If `arr_bufs_[i]` is empty, allocate
2. If `args[i].size` doesn't match the current `arr_bufs_[i]`, deallocate the existing `arr_bufs_[i]`, then allocate a new one
3. Deallocate in the destructor.

python/taichi/lang/__init__.py
ailzhang: @k-ye since `ndarray_use_torch=true` is default mode, making it an error might is a bit inconvenient for users. Esp when they don't know about `ndarray_use_torch`. So I might prefer keep this as a warning and turn it into an error when ndarray_use_torch=false becomes default?

taichi/backends/opengl/opengl_api.cpp
ailzhang: ohhhh nice catch! Thanks a lot!

python/taichi/lang/__init__.py
k-ye: SG!

taichi/llvm/llvm_program.h
```diff
@@ -98,7 +98,7 @@ class LlvmProgramImpl : public ProgramImpl {
   void finalize();
 
   DeviceAllocation allocate_memory_ndarray(std::size_t alloc_size,
-                                           uint64 *result_buffer);
+                                           uint64 *result_buffer) override;
```
qiao-bo, line 101: This does not override anything, does it?

tests/python/test_ndarray.py
```diff
@@ -603,7 +607,7 @@ def test_arg_not_match():
     _test_arg_not_match()
 
 
-@ti.test(arch=ti.opengl, ndarray_use_torch=True)
+@ti.test(arch=ti.vulkan, ndarray_use_torch=True)
```
qiao-bo, line 610: just to confirm, this is vulkan instead of opengl?

taichi/program/ndarray.cpp
```diff
@@ -15,21 +15,31 @@ Ndarray::Ndarray(Program *prog,
...
   // we'll get rid of it soon.
   device_allocation_ptr_ = (uint64_t *)&ndarray_alloc_;
-#else
-  TI_ERROR("Llvm disabled");
-#endif
+}
+
+Ndarray::~Ndarray() {
+  if (device_) {
+    device_->dealloc_memory(ndarray_alloc_);
```
qiao-bo, line 41: just for my understanding, how this works with preallocated memory? where we don't deallocate per instance.

qiao-bo (commented):
Thanks!

taichi/backends/opengl/opengl_api.cpp
```diff
@@ -406,8 +419,15 @@ void DeviceCompiledProgram::launch(RuntimeContext &ctx,
     // TODO: properly assert and throw if we bind more than allowed SSBOs.
     //       On most devices this number is 8. But I need to look up how
     //       to query this information so currently this is thrown from OpenGl.
-    for (const auto &[arg_id, bind_id] : program_.used.arr_arg_to_bind_idx) {
-      binder->buffer(0, bind_id, arr_bufs_[arg_id]);
+    for (const auto [arg_id, bind_id] : program_.used.arr_arg_to_bind_idx) {
+      if (!ctx.is_device_allocation[arg_id]) {
```
k-ye, line 423: nit: 
```suggestion
      if (ctx.is_device_allocation[arg_id]) {
```

then flip the two branches

taichi/program/ndarray.h
```diff
@@ -31,13 +31,21 @@ class Ndarray {
...
   DeviceAllocation ndarray_alloc_{kDeviceNullAllocation};
+  // Invariant:
+  //   data_ptr_ is not nullptr iff arch is a llvm backend
   uint64_t *data_ptr_{nullptr};
   uint64_t *device_allocation_ptr_{nullptr};
   std::size_t nelement_{1};
   std::size_t element_size_{1};
+  // Ndarrays manage their own |DeviceAllocation| so this must be shared with
+  // |OpenGlRuntime|. Without the ownership, when the program exits |device_|
+  // mighted be destructed earlier than Ndarray object, leaving a segfault when
```
k-ye, line 46:
```suggestion
  // might be destructed earlier than Ndarray object, leaving a segfault when
```

k-ye: Hmm, I see why this is necessary, but I wonder if there's a better way to handle this. The other way around is to track all the Ndarrays in the Program, which is more complicated to implement. Does torch track its allocated tensors, though?

taichi/backends/opengl/opengl_program.cpp
```diff
@@ -32,6 +32,17 @@ void OpenglProgramImpl::materialize_runtime(MemoryPool *memory_pool,
   TI_NOT_IMPLEMENTED;
 #endif
 }
+DeviceAllocation OpenglProgramImpl::allocate_memory_ndarray(
+    std::size_t alloc_size,
+    uint64 *result_buffer) {
+  return opengl_runtime_->device->allocate_memory(
+      {alloc_size, /*host_write=*/true, /*host_read=*/true,
```
bobcao3, line 39: These two flags should be used carefully since it not only declares a capability, it also hints the device backend to place this memory into different memory heaps (with potentially degraded on-gpu performance or host-device sync overhead)

tests/python/test_ndarray.py
ailzhang: ah yes, since we supported opengl I changed it to vulkan which is still not supported yet. :P Thanks will change the test name!

taichi/program/ndarray.cpp
ailzhang: device_ is nullptr for llvm based backends so this is a no-op for them :D

taichi/program/ndarray.h
ailzhang: As we discussed offline, I'll keep the dealloc logic in ndarray destructor for now, until we start to implement DTR :D

taichi/backends/opengl/opengl_program.cpp
ailzhang: @bobcao3 These were mostly from https://github.com/taichi-dev/taichi/blob/v0.8.4/taichi/backends/opengl/opengl_api.cpp#L374. But I did try setting `host_read` to false and it also passed tests. (setting `host_write` to false will break tests) But I'm not 100% if I understood `host_read` correctly. Do you have any insights here?

qiao-bo: i'm also curious here since in our [CUDA device allocate_memory](https://github.com/taichi-dev/taichi/blob/46b27c7c2f0970ec822c6a197f8c50f9bab6318b/taichi/backends/cuda/cuda_device.cpp#L16), these two flags lead to the usage of `malloc_managed` which hints unified memory, which has been dropped out?

k-ye: >  setting host_read to false and it also passed tests.

https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glMapBufferRange.xhtml My feeling is that WRITE implies the memory is readable.

Anyway, we can do fancier analysis later. For now it's safe to turn on both read and write.

> these two flags lead to the usage of malloc_managed which hints unified memory, which has been dropped out?

I don't think we have unified the memory allocation mechanism yet. For now the one goes through the preallocated memory is in https://github.com/taichi-dev/taichi/blob/7027cc300cf459f57dcc882dd5ddcae14ae7281e/taichi/backends/cuda/cuda_device.cpp#L38-L44

k-ye (approved):

PR #3556, opened 2021-11-19
[opengl] Serialize ndarrays and ndarray-based kernels in AOT.

ailzhang:
* __->__ #3556
* #3555
* #3554

Milestone: serialize a ndarray version of mpm88 with AOT module.

Serialized data structure is subject to change, will need to carefully
rethink about what's the minimal information should be included with an
ndarray kernel.


taichi/backends/opengl/aot_data.h
```diff
@@ -43,14 +43,35 @@ struct CompiledFieldData {
             column_num);
 };
 
+struct CompiledNdarrayData {
+  std::string field_name;
+  uint32_t dtype;
+  std::string dtype_name;
+  size_t dim{0};
+  std::vector<int> shape;
```
k-ye, line 51: Actually we don't want to restrict the shape here?

ailzhang: Yup it's more like a FYI to show an example shape, `dim` info above is more important.

k-ye: Ack, let's just remove it :-)

qiao-bo (approved):
LGTM!

taichi/backends/opengl/aot_data.h
k-ye: Can we remove `shape`?

k-ye (approved):
LGTM thanks!

Issue #3528, opened 2021-11-16
Reduce end to end time for presubmit jobs 


ailzhang:
As discussed offline with @lin-hitonami , here are a few planned steps to reduce end to end CI time, target is to get CI signal in 30mins. 

1. Get rid of "required CPU" job, directly spawn python3.6 & python3.9 CPU jobs
   - Should reduce CPU e2e time by half: 1hr -> 30mins.      
2. Setup sccache for linux jobs: 
   - build time on CPU: 15mins -> 5mins (expected) 
3. Reduce windows build time: maybe also use sccache? (not sure about this one yet)
   - Windows job takes 1hr e2e which will be a main bottleneck after step1&2. 
   - If you have any ideas how can this be improved, please feel free to let us know!

ailzhang:
Closing as it's mostly done

PR #3559, opened 2021-11-19
[ci] Add sccache to ci building process on Linux and Mac jobs

lin-hitonami:
Related issue = #3528


.github/workflows/scripts/unix_build.sh
```diff
@@ -4,7 +4,26 @@ python3 -m pip uninstall taichi taichi-nightly -y
 python3 -m pip install -r requirements_dev.txt
 python3 -m pip install -r requirements_test.txt
 git fetch origin master
-
+export SCCACHE_DIR=`pwd`/sccache_cache
+export SCCACHE_CACHE_SIZE="128M"
+echo sccache dir: $SCCACHE_DIR
+if [[ $OSTYPE == "linux-"* ]]; then
+    wget https://github.com/mozilla/sccache/releases/download/v0.2.15/sccache-v0.2.15-x86_64-unknown-linux-musl.tar.gz
+    tar -xzf sccache-v0.2.15-x86_64-unknown-linux-musl.tar.gz
```
frostming, line 12: Is musl build usable on GNU linux?

bobcao3: Seems like it can, circa https://www.musl-libc.org/faq.html

lin-hitonami: It seems in this case it's working fine.

lin-hitonami:
There seems to be some permission problem when I use `docker create -v sccache_cache:/home/dev/sccache_cache` to mount the cache to the docker. Even if I have already set the permissions all to 777 by `chmod -R 777 sccache_cache` outside the docker, the sccache still reports `Operation not permitted` error when reading the cache files. Copying the files to another directory, deleting the original files, and then moving the copied files to the original directory inside the docker can solve this problem. Generating a tarball outside the docker, copying the tarball into the docker by `docker cp`, and then extracting it inside the docker also solves the problem. I don't know if there is a better way to do this.

frostming:
Why was windows change reverted, it is the biggest bottleneck 😢

lin-hitonami:
Yes, but I have limited knowledge on windows and it seems that my configuration isn't working. It would be great if you could help us set up sccache on windows (maybe on another pr). @frostming

.github/workflows/scripts/unix_docker_build.sh
```diff
@@ -7,12 +7,20 @@ PY=$1
 GPU_BUILD=$2
 PROJECT_NAME=$3
 CI_SETUP_CMAKE_ARGS=$4
-
+export SCCACHE_DIR=/home/dev/sccache_cache
```
ailzhang, line 10: Let's put sccache related setup code together? ;) Now it's split by conda & pip :P

.github/workflows/presubmit.yml
```diff
@@ -184,14 +192,19 @@ jobs:
             exit 0
           fi
           mkdir -m777 wheel
+          mkdir -p sccache_cache
```
ailzhang, line 195: As we discussed, let's add a note about why a tarball is needed here.

ailzhang (commented):
LGTM with minor suggestions!
cc: @frostming

frostming:
> There seems to be some permission problem when I use `docker create -v sccache_cache:/home/dev/sccache_cache` to mount the cache to the docker on the github-hosted runners

Can adding `:rw` to the mount solve the problem?

`docker create -v sccache_cache:/home/dev/sccache_cache:rw`

lin-hitonami:
> > There seems to be some permission problem when I use `docker create -v sccache_cache:/home/dev/sccache_cache` to mount the cache to the docker on the github-hosted runners
> 
> Can adding `:rw` to the mount solve the problem?
> 
> `docker create -v sccache_cache:/home/dev/sccache_cache:rw`

It doesn't work. Even if I did not add rw, I can read, write, copy and delete the files. It seems that sccache performs some operations that only the file owner can do like maybe changing the mode of the files or something.

frostming:
> > > There seems to be some permission problem when I use `docker create -v sccache_cache:/home/dev/sccache_cache` to mount the cache to the docker on the github-hosted runners
> > 
> > 
> > Can adding `:rw` to the mount solve the problem?
> > `docker create -v sccache_cache:/home/dev/sccache_cache:rw`
> 
> It doesn't work. Even if I do not add rw, I can read, write, copy and delete the files. It seems that sccache performs some operations that only the file owner can do like maybe changing the mode of the files or something.

Ok, let's use the `docker cp` solution

ailzhang (approved):
Thanks! This will help a looot! 
Btw I've been wondering why we have both unix_build.sh and unix_docker_build.sh and it make code a lot messier.... Gonna open an issue and see if anyone has context on that.

Issue #3515, opened 2021-11-15
[Bug maybe][from_numpy] from_numpy() seemd take uint8 to int8


SoTosorrow:
**Describe the bug**
image input to from_numpy() get a wrong result
image put to a np.ones then input to from_numpy() get a right result
the difference seemd because from_numpy() take uint8 to int8

```
$ python report.py
image = Image.open("C:/Users/User/Desktop/1.jpg")
image = np.array(image)
w = len(image[0])
h = len(image[:])

# image turn to taichi1
pix = ti.field(ti.f32,(h,w,3))
n = np.zeros([h,w,3])
for i in range(h):
    for j in range(w):
        for k in range(3):
            n[i][j][k] = image[i][j][k]
pix.from_numpy(n)

# image turn to taichi2
pix2 = ti.field(ti.f32,(h,w,3))
pix2.from_numpy(image) # wrong result
# pix2.from_numpy(image.astype(np.float32))  #right result

print(n == image)  # same
print(pix == pix2) # not same


```

SoTosorrow:
When i alter the "from_numpy(image)" to "from_numpy(image.astype(np.float32)"  (prompt by @archibate)
then get right result
from_numpy() seemd take uint8 to int8
is it a bug or feature?

strongoier:
This should be fixed in #3560. Could you wait for Taichi v0.8.6 and help verify this?

SoTosorrow:
> This should be fixed in #3560. Could you wait for Taichi v0.8.6 and help verify this?

My pleasure, thx

SoTosorrow:
> This should be fixed in #3560. Could you wait for Taichi v0.8.6 and help verify this?

I have tested my code in the v0.8.6, it is correct now
I will do more verification tomorrow maybe, and the issue can be closed for now
thanks

PR #3560, opened 2021-11-19
[Bug] [llvm] Fix FP<->UInt castings

strongoier:
Related issue = #3515

Castings between floating point numbers and unsigned integers are not handled correctly in LLVM backends. This PR aims at fixing it.


sjwsl (approved):
LGTM!

Issue #3557, opened 2021-11-19
Cannot implicit cast from ti.f64 from ti.f16


victoriacity:
**Describe the bug**
The implicit type conversion from `ti.f64` to `ti.f16` is not available on the CPU backend.

**Platform**
Ubuntu 20.04, Python 3.8

**To Reproduce**

```py
import taichi as ti
ti.init()

@ti.kernel
def copy(a: ti.any_arr(), b: ti.any_arr()):
    for i in a:
        b[i] = a[i]

x = ti.ndarray(shape=(8,), dtype=ti.f64)
y = ti.ndarray(shape=(8,), dtype=ti.f16)
copy(x, y)
```

**Log/Screenshots**

```
JIT session error: Symbols not found: [ __truncdfhf2 ]
[E 11/18/21 22:14:15.167 3836053] [jit_cpu.cpp:lookup@165] Function "copy_c48_0_kernel_0_serial" not found
```

ailzhang:
fixed via #3561, closing

PR #3561, opened 2021-11-19
[LLVM] Fix casting f64 to f16

sjwsl:
Related issue = #3557

### DIFF
--- a/tests/python/test_f16.py
+++ b/tests/python/test_f16.py
@@ -279,9 +279,19 @@ def test_atomic_min_f16():
     foo()
     assert (f[0] == approx(f[1], rel=1e-3))
 
-@pytest.mark.parametrize('dtype', [ti.f32, ti.f64])
 @ti.test(arch=archs_support_f16)
-def test_cast_to_f16(dtype):
+def test_cast_f32_to_f16():
+    @ti.kernel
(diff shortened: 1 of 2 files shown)


tests/python/test_f16.py
```diff
@@ -278,3 +278,15 @@ def foo():
...
     assert (f[0] == approx(f[1], rel=1e-3))
+
+
+@pytest.mark.parametrize('dtype', [ti.f32, ti.f64])
+@ti.test(arch=archs_support_f16)
+def test_cast_to_f16(dtype):
+    @ti.kernel
+    def func() -> ti.f16:
+        a = ti.cast(23.0, ti.f64)
+        b = ti.cast(4.0, ti.f64)
```
victoriacity, line 289: Is it intended to be
```suggestion
        a = ti.cast(23.0, dtype)
        b = ti.cast(4.0, dtype)
```
so that the `dtype` argument can be used?

victoriacity (approved):
LGTM with a suggestion in the test case, thanks!

strongoier (approved):
LGTM!

tests/python/test_f16.py
sjwsl: Thanks!

PR #3564, opened 2021-11-20
[Bug] [metal] Fix data type alignment for arguments and return values

### LINKED ISSUE #3563: Invalid data type alignment for arguments and return values in the Metal backend
The following code snippet:
```python
import taichi as ti

ti.init(ti.metal)

@ti.kernel
def func(a: ti.i16) -> ti.f32:
    return 3.0

print(func(255))
```
prints `2.304855714121459e-41` instead of `3.0`. However, if we add another argument:
```python
@ti.kernel
def func(a: ti.i16, b: ti.i16) -> ti.f32:
    return 3.0
```
it will print `3.0` as expected.

This behavior suggests that there might be a data alignment problem for writing return values. Same thing happens for reading arguments:
```python
import taichi as ti

ti.init(ti.metal)

@ti.kernel
def func(a: ti.i8, b: ti.i32):
    print(b)

func(255, 2)
```
prints `767` instead of `2`. The binary representation of `767` is `0x2ff`, where `ff` comes from the argument `a=255`.

If we look at the generated Metal source, we will find that this is indeed the case:
```metal
class mtl_k0000_func_c46_0_args {
 public:
  explicit mtl_k0000_func_c46_0_args(device byte* addr) : addr_(addr) {}
  device int8_t* arg0() {
    // scalar, size=1 B
    return (device int8_t*)(addr_ + 0);
  }
  device int32_t* arg1() {
    // scalar, size=4 B
    return (device int32_t*)(addr_ + 1);
  }
  
  int32_t extra_arg(int i, int j) {
    device int32_t* base = (device int32_t*)(addr_ + 5);
    return *(base + (i * 8) + j);
  }
 private:
  device byte* addr_;
};
```
Accessing an `int32_t` at `addr_ + 1` is invalid behavior. The language specification shows that it follows normal data type alignment rules (https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf). We need to pad the address to `addr_ + 4` here.

strongoier:
This PR fixes #3563.

### DIFF
--- a/taichi/backends/metal/kernel_utils.cpp
+++ b/taichi/backends/metal/kernel_utils.cpp
@@ -121,6 +121,7 @@ KernelContextAttributes::KernelContextAttributes(const Kernel &kernel)
     for (int i : scalar_indices) {
       auto &attribs = (*vec)[i];
       const size_t dt_bytes = metal_data_type_bytes(attribs.dt);
+      // Align bytes to the nearest multiple of dt_bytes
       bytes = (bytes + dt_bytes - 1) / dt_bytes * dt_bytes;
       attribs.offset_in_mem = bytes;
       bytes += attribs.stride;
--- a/taichi/backends/metal/kernel_utils.cpp
+++ b/taichi/backends/metal/kernel_utils.cpp
@@ -120,12 +120,16 @@ KernelContextAttributes::KernelContextAttributes(const Kernel &kernel)
     // Put scalar args in the memory first
     for (int i : scalar_indices) {
       auto &attribs = (*vec)[i];
+      const size_t dt_bytes = metal_data_type_bytes(attribs.dt);
+      bytes = (bytes + dt_bytes - 1) / dt_bytes * dt_bytes;
       attribs.offset_in_mem = bytes;
       bytes += attribs.stride;
     }
     // Then the array args
     for (int i : array_indices) {
       auto &attribs = (*vec)[i];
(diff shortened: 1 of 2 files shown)


taichi/backends/metal/kernel_utils.cpp
```diff
@@ -120,12 +120,16 @@ KernelContextAttributes::KernelContextAttributes(const Kernel &kernel)
     // Put scalar args in the memory first
     for (int i : scalar_indices) {
       auto &attribs = (*vec)[i];
+      const size_t dt_bytes = metal_data_type_bytes(attribs.dt);
+      bytes = (bytes + dt_bytes - 1) / dt_bytes * dt_bytes;
```
sjwsl, line 124: Could you add a comment here explaining how this magic transformation works?

strongoier: I'm just thinking about what comment is most proper here. What about `Align bytes to the nearest multiple of dt_bytes`?

sjwsl: This makes sense to me :)

strongoier: Done

sjwsl (approved):
LGTM!

ailzhang (approved):
Nice catch! Thanks!

k-ye:
Thanks!

PR #3565, opened 2021-11-20
[misc] MoltenVK dynamic library & allow overriding CLANG_EXECUTABLE

bobcao3:
Related issue = #

MoltenVK moved to using dynamic library (as a static one is not available on 10.14, vulkan sdk only provides dynamic), and now we also search for it in both homebrew and the SDK. The user (in runtime) does not need to do anything different.

Also users can use `-DCLANG_EXECUTABLE=clang` to override the clang used. (useful when cmake did not select the correct binary)


sjwsl:
As discussed in #3379, it's kind of confusing to have both `CXX` and `CLANG_EXECUTABLE`for specifying a `clang`. This new option only makes sense when the detected compiler is GCC or MSVC. In other cases, we determine to use the same `clang` for taichi and the LLVM runtime. The flexibility improvement will make the dev install guide more difficult to understand unless we come up with a good way to explain it. WDYT? @k-ye @ailzhang @qiao-bo

bobcao3:
> As discussed in #3379, it's kind of confusing to have both `CXX` and `CLANG_EXECUTABLE`for specifying a `clang`. This new option only makes sense when the detected compiler is GCC or MSVC. In other cases, we determine to use the same `clang` for taichi and the LLVM runtime and `CXX` can be used to override `clang` for both. The flexibility improvement will make the dev install guide more difficult to understand unless we come up with a good way to explain it. WDYT? @k-ye @ailzhang @qiao-bo

In my specific case when I set CMAKE_CXX_COMPILER I felt into a fail loop and Cmake regenerates an empty profile.

ailzhang:
Note that @bobcao3 opened https://github.com/taichi-dev/taichi/issues/3578 to track the find clang issue, so let's merge this to unblock release. cc: @sjwsl @Leonz5288

ailzhang (approved):
LGTM!

Issue #3568, opened 2021-11-20
CI is still skipped when there are non-doc modifications


strongoier:
https://github.com/taichi-dev/taichi/runs/4252103913?check_suite_focus=true is an example run. Although `docs/lang/articles/basic/type.md` is modified, there are other non-doc modifications. However, CI is skipped in this case.

The logic here:
https://github.com/taichi-dev/taichi/blob/17a0a119bad15e5caeaf242548a72476f643fd7e/.github/workflows/presubmit.yml#L29-L32
is incorrect. We can skip CI only if all modifications are within docs.

strongoier:
Fixed by #3569.

PR #3569, opened 2021-11-20
[ci] Fix skipping presubmit checks

sjwsl:
Related issue = #3568

Fix and simplify the logic for skipping CI when there are no non-doc modifications.


ailzhang (approved):
Nice cleanup! Thanks!

PR #3571, opened 2021-11-21
[Bug] [vulkan] Fix data type alignment for arguments and return values

strongoier:
This is a fix similar to #3564, but for our Vulkan backend.


g1n0st (approved):
LGTM!

PR #3576, opened 2021-11-22
[Lang] Add get_element_size for ndarray

qiao-bo:
Add `get_element_size()` and `get_nelement()` for more convenient access of size in bytes.


YuCrazing (approved):
Thanks!

strongoier (approved):
LGTM!

PR #3588, opened 2021-11-22
[Lang] Add time check before performing version check

Leonz5288:
Related issue = #

### DIFF
--- a/python/taichi/main.py
+++ b/python/taichi/main.py
@@ -1,4 +1,5 @@
 import argparse
+import datetime
 import math
 import os
 import platform
@@ -7,7 +8,6 @@ import shutil
 import subprocess
 import sys
 import timeit
-import datetime
 from collections import defaultdict
 from functools import wraps
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


python/taichi/main.py
```diff
@@ -69,7 +70,25 @@ def __init__(self, test_mode: bool = False):
...
+                last_time = f.readlines()[0].rstrip()
+            target_date = (cur_date -
+                           datetime.timedelta(days=14)).strftime('%Y-%m-%d')
+            if target_date > last_time:
+                with open(timestamp_path, 'w') as f:
+                    f.write(cur_date.strftime('%Y-%m-%d'))
+                    f.truncate()
+                self._check_version()
+        else:
+            with open(timestamp_path, 'w') as f:
```
ailzhang, line 89: nit: instead of saving the cur_date, shall we just save (cur_date+14 days) in the file so that your just need to perform a simple comparison every time? ;)

ailzhang (approved):
LGTM!

python/taichi/main.py
Leonz5288: Make sense!

PR #3591, opened 2021-11-22
[ci] Turn off vulkan build on macos

Leonz5288:
Related issue = #


ailzhang (approved):
FYI this PR matches presubmit behavior that vulkan is turned off on non-M1 macos.

PR #3592, opened 2021-11-22
[gui] GGUI initial alpha transparency support

bobcao3:
Related issue = #

Just some initial work, I think we should modify device api to have the ability to specify blending ops instead of hard code it


g1n0st (approved):
Thanks!

AmesingFlank:
Thanks!

Please also change the `ti.scene(color = )` API to support rgba as well. (need to update `python/ui/scene.py` and `taichi/python/export_ggui.cpp`)

AmesingFlank:
> Thanks!
> 
> Please also change the `ti.scene(color = )` API to support rgba as well. (need to update `python/ui/scene.py` and `taichi/python/export_ggui.cpp`)

But we can merge this one first to unblock YuChang.

AmesingFlank (approved):

Issue #3463, opened 2021-11-11
SNodeOpStmt is not allowed here


FantasyVR:
**Describe the bug**
Related forum post: [here](https://forum.taichi.graphics/t/cpu-gpu/2003/8)

**To Reproduce**
```py
import taichi as ti
ti.init(arch=ti.gpu)

@ti.kernel
def run(a: ti.i32):
    b = a + 1
    for x in range(1):
        for i in range(b):
            print('OK')

run(2)
```

**Log/Screenshots**
```
[Taichi] version 0.8.5, llvm 10.0.0, commit 45c6ad48, linux, python 3.8.11
[Taichi] Starting on arch=cuda
[E 11/11/21 17:42:28.226 640199] [offload.cpp:visit_operand@558] SNodeOpStmt is not allowed here.
```

PR #3597, opened 2021-11-23
[Bug] [ir] Fix the IdentifyValuesUsedInOtherOffloads pass

### LINKED ISSUE #3467: Finish the implementation of IdentifyValuesUsedInOtherOffloads
The implementation of `IdentifyValuesUsedInOtherOffloads` is not finished yet: https://github.com/taichi-dev/taichi/blob/153f338df549174a4c000d85ff5936e7a36923e1/taichi/transforms/offload.cpp#L269-L272

More specifically, most statements currently go through: https://github.com/taichi-dev/taichi/blob/153f338df549174a4c000d85ff5936e7a36923e1/taichi/transforms/offload.cpp#L345-L351

However, the operands involved in many of them are not present at the `operands` member variable, therefore they will not be handled. #3463 is an example of a result bug and #3466 is an example of the corresponding fix.

To avoid future appearance of such bugs, we need to finish the implementation of this class.

strongoier:
Related issue = #3463, #3467

#3467 didn't point out the accurate root cause of #3463. `RangeForStmt` was indeed visited with https://github.com/taichi-dev/taichi/blob/1ca4c408a69ba7475fd792aadd2e45cebf218615/taichi/ir/basic_stmt_visitor.cpp#L33-L36 instead of https://github.com/taichi-dev/taichi/blob/1ca4c408a69ba7475fd792aadd2e45cebf218615/taichi/transforms/offload.cpp#L374-L380

Having this, the most effective way to fix #3467 is to fill in the `preprocess_container_stmt()` member function common to all container statements, which is done in this PR.

### DIFF
--- a/taichi/transforms/offload.cpp
+++ b/taichi/transforms/offload.cpp
@@ -345,13 +345,6 @@ class IdentifyValuesUsedInOtherOffloads : public BasicStmtVisitor {
     TI_ASSERT(current_offloaded);
   }
 
-  void visit(RangeForStmt *stmt) override {
-    test_and_allocate(stmt->begin);
-    test_and_allocate(stmt->end);
-    if (stmt->body)
-      stmt->body->accept(this);
-  }
-
   void test_and_allocate(Stmt *stmt) {
     if (stmt == nullptr)
       return;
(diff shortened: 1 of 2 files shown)


lin-hitonami (approved):
LGTM!

PR #3598, opened 2021-11-23
[Lang] Check version when importing taichi instead of when using ti

Leonz5288:
Related issue = #


python/taichi/lang/__init__.py
```diff
@@ -456,6 +523,26 @@ def init(arch=None,
...
+            last_time = f.readlines()[0].rstrip()
+        if cur_date.strftime('%Y-%m-%d') > last_time:
+            check_version()
+            with open(timestamp_path, 'w') as f:
+                f.write((cur_date +
+                         datetime.timedelta(days=14)).strftime('%Y-%m-%d'))
+                f.truncate()
+    else:
+        check_version()
+        with open(timestamp_path, 'w') as f:
```
rexwangcc, line 542:
```suggestion
        with open(timestamp_path, 'wb') as f:
```
maybe use binary mode? (if so, same for all R/W...)

rexwangcc (commented):
Looks great and thanks! I left one comment, and guess @ailzhang could provide insights about the new deps (requests).

python/taichi/lang/__init__.py
Leonz5288: Is binary mode better?

lin-hitonami (approved):
LGTM！

PR #3599, opened 2021-11-23
[mesh] Make ti.Mesh compatible with dynamic index

strongoier:
Related issue = #3567

Just remember to use `load_if_ptr` for any `Expr` when constructing Taichi frontend IR.


g1n0st (approved):
LGTM! Thanks!

PR #3607, opened 2021-11-24
[bug] Enable reassignment of scalar arguments

### LINKED ISSUE #3605: [AST] Cannot reassign numerical arguments of kernels
### Describe the bug
After AST refactor has been fully deployed (#3516 and #3527), numerical arguments passed to Taichi kernels cannot be reassigned. This is because in the new AST, the argument variable is regarded as a rvalue and thus cannot be modified.

### To Reproduce
```py
import taichi as ti
ti.init()
@ti.kernel
def add(a: ti.f32, b: ti.f32) -> ti.f32:
    a = a + b
    return a
add(1.0, 2.0)
```

### Log/Screenshots
```
# when compiling `a = a + b`
File "path\to\taichi\python\taichi\lang\ops.py", line 838, in assign
  _ti_core.expr_assign(a.ptr, b.ptr, stack_info())
RuntimeError: [export_lang.cpp:taichi::lang::expr_assign@55] Assertion failure: lhs->is_lvalue()
```

### Additional comments
**Before** the AST refactor, the following IR is produced for this kernel which indicates it correctly creates a new variable when reassigning `a`: 
```
kernel {
  $0 = offloaded
  body {
    <f32> $1 = arg[0]
    <f32> $2 = arg[1]
    <f32> $3 = add $1 $2
    <f32> $4 : return $3
  }
}
```

lin-hitonami:
Related issue = close #3605

### DIFF
--- a/python/taichi/lang/ir_builder.py
+++ b/python/taichi/lang/ir_builder.py
@@ -340,8 +340,9 @@ class IRBuilder(Builder):
                 else:
                     ctx.create_variable(
                         arg.arg,
-                        ti.lang.kernel_arguments.decl_scalar_arg(
-                            ctx.func.argument_annotations[i]))
+                        ti.expr_init(
+                            ti.lang.kernel_arguments.decl_scalar_arg(
+                                ctx.func.argument_annotations[i])))
             # remove original args
             node.args.args = []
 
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Thanks for the quick fix!

PR #3609, opened 2021-11-24
[misc] Remove legacy torch_io.py

strongoier:
Currently `from_torch` and `to_torch` are implemented in our `Field` classes, e.g.,
https://github.com/taichi-dev/taichi/blob/e26d9c24b788539946f8f858aae487e5367e179c/python/taichi/lang/field.py#L139-L140
https://github.com/taichi-dev/taichi/blob/e26d9c24b788539946f8f858aae487e5367e179c/python/taichi/lang/field.py#L116-L117

The legacy implementaion in `torch_io.py` is no longer used and needed. This PR aims at cleaning them up.


ailzhang (approved):
Yay!! Thanks! 
(btw not sure if we need to deprecate it first lol, was it widely used before? ;)

strongoier:
> Yay!! Thanks! (btw not sure if we need to deprecate it first lol, was it widely used before? ;)

It seems that `torch_io.py` was added in https://github.com/taichi-dev/taichi/commit/754df758626b032a7c562b3d2cc71ef26e9ebcdd (Oct 2019) and soon replaced by https://github.com/taichi-dev/taichi/commit/d8d6cab9f5a19860e485bdc78d0a593f247bb75d (Dec 2019). It should be removed at that time :-)

PR #3197, opened 2021-10-14
[infra] Disable default signal handler if envar defined

Renkai:
Related issue = https://github.com/taichi-dev/taichi/issues/3165

Issue #3165, opened 2021-10-12
Add an environment variable as switch to close the default signal handler


Zheaoli:
**Concisely describe the proposed feature**
Add an environment variable as a switch to close the default signal handler.

The code here https://github.com/taichi-dev/taichi/blob/7e54f56d305f370f773264864f7fcf117470ddf1/taichi/system/hacked_signal_handler.cpp#L66 add some signal handler to handle the signal such as SIGABRT etc to print some formatted message by tichi.

But in some circumstances, people may want to use some tools such as breakpads/crashpad to collect crash messages on their own. All of those tools will have some custom signal handlers that are override by taichi default.

So we need a switch to close the default TAICHi signal handler.

PR #3613, opened 2021-11-24
[lang] Disable signal handlers when TI_DISABLE_SIGNAL_HANDLERS=1

k-ye:
Related issue = #3165

FYI @Renkai, sorry for taking over your work in #3197. This feature is kind of urgent for us :)


strongoier (approved):
LGTM!

PR #3614, opened 2021-11-24
[gui] Respect drawing order specified by user

AmesingFlank:
Respect drawing order specified by user.

### DIFF
--- a/taichi/ui/backends/vulkan/renderer.cpp
+++ b/taichi/ui/backends/vulkan/renderer.cpp
@@ -80,12 +80,23 @@ void Renderer::scene(Scene *scene) {
   }
   float aspect_ratio = swap_chain_.width() / (float)swap_chain_.height();
   scene->update_ubo(aspect_ratio);
-  for (int i = 0; i < scene->mesh_infos_.size(); ++i) {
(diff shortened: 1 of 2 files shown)


g1n0st (approved):
LGTM! Thanks!
https://github.com/taichi-dev/taichi/blob/4c182a052d253a28ddcc37c538305e97735043c1/taichi/python/export_ggui.cpp#L127
I think there should be `vec4`?

https://github.com/taichi-dev/taichi/blob/f5c46fc7112f28faa931ddc5910d9c7955653172/taichi/python/export_ggui.cpp#L143

ditto

AmesingFlank:
> LGTM! Thanks!
> 
> https://github.com/taichi-dev/taichi/blob/4c182a052d253a28ddcc37c538305e97735043c1/taichi/python/export_ggui.cpp#L127
> 
> 
> I think there should be `vec4`?
> https://github.com/taichi-dev/taichi/blob/f5c46fc7112f28faa931ddc5910d9c7955653172/taichi/python/export_ggui.cpp#L143
> 
> ditto

Hmm no. We haven't implemented alpha channel for `color` yet. Only `per_vertex_color` can use RGBA for now.

PR #3615, opened 2021-11-25
[ci] Widen the GPU test timeout

k-ye:
Related issue = #

Sample failures due to the early cancellation:

* https://github.com/taichi-dev/taichi/runs/4318329900?check_suite_focus=true
* https://github.com/taichi-dev/taichi/runs/4318962108?check_suite_focus=true


k-ye:
Fixed by #3619

PR #3632, opened 2021-11-27
[Lang] Add option short_circuit_operators for short-circuiting boolean ops

### LINKED ISSUE #3572: Support short-circuiting logical expressions
The following code snippet,
```python
import taichi as ti

ti.init(debug=True)

a = ti.field(ti.i32, shape=10)

@ti.kernel
def func(i: ti.i32):
    if i >= 0 and a[i] != 0:
        print(a[i])

func(-1)
```
will give a runtime error due to negative indices. This is because in Taichi, logical expressions are completely evaluated anyway. However, most programming languages are using short-circuiting logical expressions, so this behavior difference may cause confusion to newcomers.

I suggest that Taichi can support short-circuiting logical expressions. To avoid affecting performance of existing code, we can have an option in compiler configurations to decide whether to enable the new behavior.

re-xyr:
Related issue = Fixes #3572

cc @strongoier 

```python
ti.init(short_circuit_operators=True)
```

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -477,16 +477,67 @@ class ASTTransformer(Builder):
         return node
 
     @staticmethod
+    def build_short_circuit_and(operands):
+        if len(operands) == 1:
+            return operands[0].ptr
+
+        val = ti.expr_init(None)
+        lhs = operands[0].ptr
+        ti.begin_frontend_if(lhs)
+
+        ti.core.begin_frontend_if_true()
+        rhs = ASTTransformer.build_short_circuit_and(operands[1:])
+        val.assign(rhs)
+        ti.core.pop_scope()
+
+        ti.core.begin_frontend_if_false()
+        val.assign(0)
+        ti.core.pop_scope()
+
+        return val
+
+    @staticmethod
+    def build_short_circuit_or(operands):
+        if len(operands) == 1:
+            return operands[0].ptr
+
+        val = ti.expr_init(None)
+        lhs = operands[0].ptr
+        ti.begin_frontend_if(lhs)
+
+        ti.core.begin_frontend_if_true()
+        val.assign(1)
+        ti.core.pop_scope()
+
+        ti.core.begin_frontend_if_false()
+        rhs = ASTTransformer.build_short_circuit_or(operands[1:])
+        val.assign(rhs)
+        ti.core.pop_scope()
+
+        return val
+
+    @staticmethod
+    def build_normal_bool_op(op):
+        def inner(operands):
+            result = op(operands[0].ptr, operands[1].ptr)
+            for i in range(2, len(operands)):
+                result = op(result, operands[i].ptr)
+            return result
+
+        return inner
+
+    @staticmethod
     def build_BoolOp(ctx, node):
         node.values = build_stmts(ctx, node.values)
-        op = {
-            ast.And: ti.logical_and,
-            ast.Or: ti.logical_or,
-        }.get(type(node.op))
-        result = op(node.values[0].ptr, node.values[1].ptr)
-        for i in range(2, len(node.values)):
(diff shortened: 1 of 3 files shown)


tests/python/test_short_circuit.py
```diff
@@ -0,0 +1,49 @@
+import taichi as ti
+
+
+@ti.test(short_circuit_operators=True)
+def test_and_shorted():
+    a = ti.field(ti.i32, shape=10)
+
+    @ti.func
+    def explode() -> ti.i32:
+        return a[-1]
```
strongoier, line 10: To actually make this explode, you need to add `debug=True`. When `debug=False`, indexing out of bounds is an undefined behavior in Taichi, which may or may not trigger a runtime error.

strongoier (commented):
Great work! Only a tiny thing to notice below.

lin-hitonami (changes requested):
IIUC, this method always needs `len(node.values) - 1` times of comparison when executing. When the program meets a `false` in `and` operation, the comparison after it is not necessary. This can be achieved by generating nested ifs.

strongoier:
> IIUC, this method always needs `len(node.values) - 1` times of comparison when executing. When the program meets a `false` in `and` operation, the comparison after it is not necessary. This can be achieved by generating nested ifs.

I agree. Your comment reminds me that Python AST treats `BoolOp` differently from `BinaryOp`, which naturally fits your suggestion.

re-xyr:
> this method always needs `len(node.values) - 1` times of comparison when executing

nice catch - will fix this. Not sure why Python doesn't just make boolean operators right-associative though.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -476,17 +476,68 @@ def build_UnaryOp(ctx, node):
         node.ptr = op(node.operand.ptr)
         return node
 
+    @staticmethod
+    def build_short_circuit_and(operands):
+        if len(operands) == 1:
+            return operands[0].ptr
```
lin-hitonami, line 482: Should we convert it to a bool value? Maybe by ti.logical_not(ti.logical_not(operands[0].ptr)) or something like that?

re-xyr: Chi IR will already give a compile error at wrong types, so I think this is unnecessary?

lin-hitonami: In rare circumstances there might be some people who writes codes like `a = 123 + (1 and 2)`, the value of `1 and 2` should be `1`(`True`) instead of `2`. IIUC in the current implementation the value of `1 and 2` is the value of the last operands which is `2`?

re-xyr: Ah yes, this is indeed the case. It is also Python semantics (`1 and 2 == 2`), but normal (non short-circuiting) taichi semantics is casting that into boolean, so I'll change this.

strongoier: Your comment definitely makes sense. However, I have to say that the result of bool expressions in Taichi is undefined for now. For example, having the following code snippet,
```python
@ti.kernel
def func(a: ti.i32, b: ti.i32) -> ti.i32:
    return a and b
```
, you'll get `func(2, 3) == 2`. This is because Taichi uses `bitwise and` to implement `logical and`.

Therefore, I think this is not an issue for this PR. Instead, I think we need to open another issue to discuss what is the expected behavior of bool expressions in Taichi.

lin-hitonami (approved):
Although the expected behavior of bool expressions is to be determined, other parts look good to me. Thanks!

python/taichi/lang/ast/ast_transformer.py
strongoier: We can discuss this in #3635.

strongoier (approved):
LGTM! Thanks!

PR #3637, opened 2021-11-29
[error] Add line number and source code to exception

### LINKED ISSUE #3612: Print the related source code and its position in the file when an error occurs
I would like to add the feature that prints the related source code and its position in the file when an error occurs when building the AST.

Keep a copy of the source code of the taichi kernel/function inside the context and utilize the position information in the AST node to find out the source code related to the node. When an exception is thrown, we can catch it in the transformer and add the extra information in the error message and then throw it.

lin-hitonami:
Related issue = close #3612

### DIFF
--- a/python/taichi/lang/exception.py
+++ b/python/taichi/lang/exception.py
@@ -6,5 +6,5 @@ class InvalidOperationError(Exception):
     pass
 
 
-class TaichiError(Exception):
+class TaichiCompilationError(Exception):
     pass
--- a/python/taichi/lang/exception.py
+++ b/python/taichi/lang/exception.py
@@ -4,3 +4,7 @@ class TaichiSyntaxError(Exception):
 
 class InvalidOperationError(Exception):
     pass
+
+
+class TaichiError(Exception):
+    pass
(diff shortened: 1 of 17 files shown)


python/taichi/lang/exception.py
```diff
@@ -4,3 +4,7 @@ class TaichiSyntaxError(Exception):
 
 class InvalidOperationError(Exception):
     pass
+
+
+class TaichiError(Exception):
```
strongoier, line 9: What about renaming this to `TaichiCompilationError`? `TaichiError` sounds a bit weird.

strongoier (commented):
Looks great! I feel like the modifications in #3439 is no longer needed now. Do I understand correctly?

lin-hitonami:
> Looks great! I feel like the modifications in #3439 is no longer needed now. Do I understand correctly?

We can define some exceptions in pybind11 ffi part so that the customized exceptions in C++ can be converted to python exceptions naturally.

strongoier (approved):
LGTM! Thanks!

Issue #3625, opened 2021-11-26
some error on version 0.8.6


bolin12:
I encountered some ploblems:
version 0.8.6 will display a totally chaos of this script https://github.com/taichiCourse01/taichi_ray_tracing/blob/master/4_0_path_tracing.py

and also report  " 'float' object has no attribute 'ptr' " error of this
https://github.com/taichiCourse01/taichi_ray_tracing/blob/master/3_2_whitted_style_ray_tracing.py

as I switched back to earlier version 0.8.4, problems above are gone.

Are there problems related to my machine? thanks!

strongoier:
Could you provide the complete error message? Thanks!

bolin12:
[Taichi] version 0.8.6, llvm 10.0.0, commit d5f18ffd, linux, python 3.8.10
[Taichi] Starting on arch=cuda
Traceback (most recent call last):
  File "3_2_whitted_style_ray_tracing.py", line 242, in <module>
    render()
  File "/home/lbl/miniconda3/lib/python3.8/site-packages/taichi/lang/kernel_impl.py", line 702, in wrapped
    return primal(*args, **kwargs)
  File "/home/lbl/miniconda3/lib/python3.8/site-packages/taichi/lang/kernel_impl.py", line 628, in __call__
    key = self.ensure_compiled(*args)
  File "/home/lbl/miniconda3/lib/python3.8/site-packages/taichi/lang/kernel_impl.py", line 619, in ensure_compiled
    self.materialize(key=key, args=args, arg_features=arg_features)
  File "/home/lbl/miniconda3/lib/python3.8/site-packages/taichi/lang/kernel_impl.py", line 466, in materialize
    taichi_kernel = _ti_core.create_kernel(taichi_ast_generator,
  File "/home/lbl/miniconda3/lib/python3.8/site-packages/taichi/lang/kernel_impl.py", line 461, in taichi_ast_generator
    visitor.visit(tree)
  File "/home/lbl/miniconda3/lib/python3.8/site-packages/taichi/lang/ast/transformer.py", line 47, in visit
    tree = IRBuilder()(ctx, tree)
  File "/home/lbl/miniconda3/lib/python3.8/site-packages/taichi/lang/ast_builder_utils.py", line 16, in __call__
    return method(ctx, node)
  File "/home/lbl/miniconda3/lib/python3.8/site-packages/taichi/lang/ir_builder.py", line 425, in build_Module
[...]

PR #3638, opened 2021-11-29
[Bug] [lang] Fix copying a Matrix/Struct from Python scope to Taichi scope

strongoier:
Related issue = #3625

A `Matrix/Struct` defined in Python scope is read-only in Taichi scope. To assign it to a local variable in Taichi scope, we have to do explicit copy. After this PR, https://github.com/taichiCourse01/taichi_ray_tracing/blob/master/3_2_whitted_style_ray_tracing.py should work fine.


lin-hitonami (approved):
LGTM!

PR #3640, opened 2021-11-29
[refactor] Get rid of ti.task.

ailzhang:
* #3641
* __->__ #3640
* #3639

Seems like `ti.task` is not used and maintained so let's remove it.


k-ye (approved):
Yay

Issue #3642, opened 2021-11-29
Feature Proposal: Taichi's Ahead of Time (AOT) Module


k-ye:
We'd like to share our ideas on how to implement the AOT feature in Taichi. AOT refers to the process of using Taichi as a GPU compute shader/kernel compiler: The AOT users can compile their Taichi kernels into an *AOT module*, package this module alongside their apps (likely without a Python environment, e.g. an Android app), and load/execute the compiled shaders from this module in the app.

Note that AOT is already an ongoing work, hence some of the tasks have already been implemented. For a quick peak of the Taichi AOT workflow, please check out [this test](https://github.com/taichi-dev/taichi/blob/master/tests/python/test_aot.py).

# Goals

* Provide a workflow to package the Taichi kernels into a module that can be loaded and executed from a non-Python environment (e.g. an Android app using Vulkan).
* Priotitize on the mobile platforms deployment, including OpenGL ES, Vulkan and Apple Metal.
* Priotitize on the *dense* Taichi fields. However, we have plan to further decouple the AOT from Taichi fields (see the *Ndarray* section).

# API Proposal

Taichi provides a utility, [`taichi.aot.Module`](https://github.com/taichi-dev/taichi/blob/fdf665e4dfe6c1cee41e67b12b288749b2c8cb8f/python/taichi/aot/module.py#L58), for compiling the Taichi kernels and fields info into a module file. It provides these APIs:

* `add_kernel(kernel_fn)`: Add a Taichi kernel to the AOT module.
* `add_kernel_template(kernel_templ_fn)`: Add a *Taichi kernel template* to the AOT module.
* (*) `add_field(name, field)` : Add a Taichi field to the AOT module. However, we hope that Ndarray can serve as a more convenient dense data container in the AOT use cases.
* `save(filepath, filename)`: Save this AOT module to `filepath/filename`.

We will walk through the `Module` usage with the following example.

```py
x = ti.Vector.field(2, float, n_particles)
v = ti.Vector.field(2, float, n_particles)
...

@ti.kernel
def init():
  ...

@ti.kernel
def substep():
  ...

def run_jit():
    gui = ti.GUI('mpm88')
    # Driver logic
    init()
    while True:
        for _ in range(50):
            substep()
    gui.circles(x.to_numpy())
    gui.show()

def run_aot():
    # 1
    mod = ti.aot.Module(arch=ti.opengl)
    # 2
    mod.add_kernel(init)
    mod.add_kernel(substep)
[...]

PR #3647, opened 2021-11-29
[Vulkan] Add initial support of AOT

ghuau-innopeak:

### DIFF
--- a/taichi/backends/vulkan/aot_module_builder_impl.cpp
+++ b/taichi/backends/vulkan/aot_module_builder_impl.cpp
@@ -1,0 +1,73 @@
+#include "taichi/backends/vulkan/aot_module_builder_impl.h"
+
+#include <fstream>
+#include <type_traits>
+
+#include "taichi/backends/vulkan/codegen_vulkan.h"
+
+namespace taichi {
+namespace lang {
+namespace vulkan {
+
+AotModuleBuilderImpl::AotModuleBuilderImpl(
+    VkRuntime *runtime,
+    const std::vector<CompiledSNodeStructs> &compiled_structs)
+    : compiled_structs_(compiled_structs), runtime_(runtime) {
+}
+
+void AotModuleBuilderImpl::write_spv_file(
+    const std::string &output_dir,
+    const TaskAttributes &k,
+    const std::vector<uint32_t> &source_code) const {
+  const std::string spv_path = fmt::format("{}/{}.spv", output_dir, k.name);
+  std::ofstream fs(spv_path, std::ios_base::binary | std::ios::trunc);
+  fs.write((char *)source_code.data(), source_code.size() * sizeof(uint32_t));
+  fs.close();
+}
+
+void AotModuleBuilderImpl::dump(const std::string &output_dir,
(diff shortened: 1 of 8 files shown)


ghuau-innopeak:
There is one issue with the current PR I'm not entirely sure: why it's needed to run `lower(kernel);` in `AotModuleBuilderImpl::add_per_backend(...)`, otherwise, we don't get an OffloadStatement but a FrontendStatement

k-ye:
> There is one issue with the current PR I'm not entirely sure: why it's needed to run `lower(kernel);` in `AotModuleBuilderImpl::add_per_backend(...)`, otherwise, we don't get an OffloadStatement but a FrontendStatement

Yup, `lower()` is the mechanism to translate the frontend AST to the lower-level, SSA compliant IR (namely CHI IR)

taichi/backends/vulkan/aot_module_builder_impl.h
```diff
@@ -0,0 +1,52 @@
...
+
+class AotModuleBuilderImpl : public AotModuleBuilder {
+ public:
+  explicit AotModuleBuilderImpl(
+      VkRuntime *runtime,
+      const std::vector<CompiledSNodeStructs> &compiled_structs);
+  void dump(const std::string &output_dir,
+            const std::string &filename) const override;
+
+ protected:
```
k-ye, line 25: nit: I belive this class is not gonna be further inherited.
```suggestion
 private:
```

k-ye (commented):
Thanks! The PR looks good, I'll leave this to @ailzhang for additional comments, if any :-)

taichi/backends/vulkan/aot_module_builder_impl.cpp
```diff
@@ -0,0 +1,75 @@
...
+    : compiled_structs_(compiled_structs), runtime_(runtime) {
+}
+
+void AotModuleBuilderImpl::write_spv_file(
+    const std::string &output_dir,
+    const std::string &filename,
+    const TaskAttributes &k,
+    const std::vector<uint32_t> &source_code) const {
+  const std::string spv_path =
+      fmt::format("{}/{}_{}.spv", output_dir, filename, k.name);
```
ailzhang, line 24: btw we recently found out that ignoring the filename prefix is probably more convenient (https://github.com/taichi-dev/taichi/pull/3648), I'm curious if you want the same for vulkan backend ;)

ailzhang (approved):
Looks great! Thanks!!

taichi/backends/vulkan/aot_module_builder_impl.h
ghuau-innopeak: I agree, I moved it to private and if it needs to be inherited and used later... we can still change it back to protected :).

taichi/backends/vulkan/aot_module_builder_impl.cpp
ghuau-innopeak: Definitely! I pushed the same changes to this PR, that would avoid to do multiple one :). I tested it and seems to work as well.

k-ye (approved):

PR #3648, opened 2021-11-30
[refactor] Ignore filename prefix in opengl aot files.

ailzhang:
It seems good enough to use folders here so let's keep the filenames
simple and match kernel names.

Related issue = #


Leonz5288 (approved):
LGTM!

PR #3649, opened 2021-11-30
[Bug] [lang] Fix copying Matrix/StructField elements in Taichi scope

strongoier:
Related issue = #3625

Accidentally, assignments from Matrix/StructField elements to local variables are by reference:
```python
import taichi as ti

ti.init(debug=True)

a = ti.Vector.field(3, ti.i32, shape=())
a[None] = ti.Vector([1, 2, 3])

@ti.kernel
def test():
    b = a[None]
    b[0] = 5
    b[1] = 9
    b[2] = 7
    print(a[None][0])  # 5
    print(a[None][1])  # 9
    print(a[None][2])  # 7

test()
```

However, they are indented to be by value. The PR aims at fixing it. After this PR, https://github.com/taichiCourse01/taichi_ray_tracing/blob/master/4_0_path_tracing.py should work fine.

### DIFF
--- a/tests/python/test_custom_struct.py
+++ b/tests/python/test_custom_struct.py
@@ -315,3 +315,24 @@ def test_copy_python_scope_struct_to_taichi_scope():
         assert b.b == 4
 
     test()
+
+
+@ti.test(debug=True)
+def test_copy_struct_field_element_to_taichi_scope():
+    a = ti.Struct.field({'a': ti.i32, 'b': ti.i32}, shape=())
+    a[None].a = 2
+    a[None].b = 3
+
+    @ti.kernel
+    def test():
+        b = a[None]
+        assert b.a == 2
+        assert b.b == 3
+        b.a = 5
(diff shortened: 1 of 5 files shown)


lin-hitonami (approved):
LGTM!

tests/python/test_custom_struct.py
```diff
@@ -315,3 +315,24 @@ def test():
         assert b.b == 4
 
     test()
+
+
+@ti.test(debug=True)
+def test_copy_struct_field_element_to_taichi_scope():
```
ailzhang, line 321: OOC: why do we need to set `debug=True` here? ;)

ailzhang (approved):
LGTM!

tests/python/test_custom_struct.py
strongoier: Because assertions in kernels are enabled only when `debug=True` :-)

Issue #3630, opened 2021-11-27
[IR] Building CHI IR examples 


AmesingFlank:
I'm trying to lean to use CHI IR builder, but I'm having some troubles building `chi_examples`. From what I can tell

* The include paths aren't set up correctly. This is likely because we have added quite a few dependencies.
* The build args aren't set-up correctly.  The cmake file seem to be linux-specific, and on windows, I'm getting things like "ERROR: C++ 14 required". I'm trying to fix it but I could use some help. @BillXu2000

BillXu2000:
Dependency: #3636

AmesingFlank:
Instead of fixing this CMakeList, I suggest we move it under `cmake` and add a `TI_BUILD_CHI_EXAMPLE` to optionally enable it (similar to how taichi cpp tests are handled). This way, we can share all the build configurations of the main taichi library.

BillXu2000:
`chi_examples` was supposed to be an example to show how to independently compile and run taichi kernels in cpp with an exported taichi core shared library, decoupled from cmake settings in Taichi repository. Though it seems difficult now to maintain its availability while Taichi is developing so fast. I am considering whether deleting this `chi_examples` or not after fix bugs for current Taichi version.

By the way, [this test](https://github.com/taichi-dev/taichi/blob/master/tests/cpp/ir/ir_builder_test.cpp#L87-L119) might be useful  as an example to construct and run kernels in cpp.

AmesingFlank:
It would indeed be more ideal if the example can be built separately. But if this is difficult, then having it coupled with the main taichi library is still better then deleting it completely... Even leaving it as its current state would be better than having it deleted i would say, because people can still use it as a reference.

k-ye:
Let's just add this to CI, which is the best way to prevent regression :-)

PR #3652, opened 2021-11-30
[test] Fix chi examples and add them to CI

BillXu2000:
Related issue = #3630

Fix CHI example for current Taichi version, and add CHI example as a test to CI.
Not sure whether this works for platforms other than Ubuntu.


BillXu2000:
CI is only set up for Ubuntu now.

BillXu2000:
closing #3630

chi_examples/README.md
```diff
@@ -6,6 +6,8 @@ Follow the steps in `https://docs.taichi.graphics/lang/articles/contribution/dev
 
 Add option `-DTI_EXPORT_CORE=ON` to your `cmake` command (i.e. use `cmake .. DTI_EXPORT_CORE=ON`).
 
+Set environment variable `$TAICHI_REPO_DIR` to your Taichi repository path.
```
k-ye, line 9: Hmm, do we still need this?

tests/run_tests.py
```diff
@@ -20,6 +20,29 @@ def _test_cpp():
         subprocess.check_call(f'./{cpp_test_filename}',
                               env={'TI_LIB_DIR': ti_lib_dir},
                               cwd=build_dir)
+        export_core_filename = 'libtaichi_export_core.so'
+        if os.path.exists(os.path.join(build_dir, export_core_filename)):
+            chi_dir = os.path.join(curr_dir, '../chi_examples')
+            subprocess.run('cmake . -B build',
```
k-ye, line 26: Could you use `[cmake, '.', '-B', 'build']` and make `shell=False`?

chi_examples/README.md
ailzhang: no we should get rid of it :D

tests/run_tests.py
BillXu2000: Sure, though I wonder the reason to switch it to `shell=False`.

BillXu2000:
Problem has been solved in #3630, so I closed this PR.

PR #3653, opened 2021-11-30
[perf] Accelerate _inside_class()

strongoier:
When profiling my own application, I found that the `_inside_class()` function of taichi was the bottleneck: https://github.com/taichi-dev/taichi/blob/2166c8fe14e5444dc237d8b343a378a317962877/python/taichi/lang/kernel_impl.py#L649-L660

As we only needed the frame at `level_of_class_stackframe`, getting the whole stack was a waste of time. I accelerated this function by mimicking the implementation of `inspect.stack()`, but only fetching the target frame out.

This PR decreases the running time of my own application from 22.2s to 0.8s. It also decreases the running time of `import taichi as ti` from 1.7s to 0.5s on my machine.

References:
https://github.com/python/cpython/blob/9a5dec4e978f68ab956e979858ce3aa178436fa4/Lib/inspect.py#L1559-L1569
https://github.com/python/cpython/blob/9a5dec4e978f68ab956e979858ce3aa178436fa4/Lib/inspect.py#L1583-L1589

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -643,9 +643,8 @@ _KERNEL_CLASS_STACKFRAME_STMT_RES = [
 def _inside_class(level_of_class_stackframe):
     try:
         maybe_class_frame = inspect.currentframe()
-        while level_of_class_stackframe > 0:
+        for _ in range(level_of_class_stackframe):
             maybe_class_frame = maybe_class_frame.f_back
-            level_of_class_stackframe -= 1
         statement_list = inspect.getframeinfo(maybe_class_frame)[3]
         first_statment = statement_list[0].strip()
         for pat in _KERNEL_CLASS_STACKFRAME_STMT_RES:
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -642,9 +642,7 @@ _KERNEL_CLASS_STACKFRAME_STMT_RES = [
 
 def _inside_class(level_of_class_stackframe):
     try:
-        maybe_class_frame = inspect.currentframe()
-        for _ in range(level_of_class_stackframe):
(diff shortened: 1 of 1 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -641,10 +641,12 @@ def __call__(self, *args, **kwargs):
 
 
 def _inside_class(level_of_class_stackframe):
-    frames = oinspect.stack()
     try:
-        maybe_class_frame = frames[level_of_class_stackframe]
-        statement_list = maybe_class_frame[4]
+        maybe_class_frame = inspect.currentframe()
+        while level_of_class_stackframe > 0:
```
frostming, line 646: I'd prefer a for-loop to be more pythonic:

```python
for _ in range(level_of_class_stackframe):
```

strongoier: Updated :-)

sjwsl (changes requested):

python/taichi/lang/kernel_impl.py
```diff
@@ -641,10 +641,11 @@ def __call__(self, *args, **kwargs):
...
 
 def _inside_class(level_of_class_stackframe):
-    frames = oinspect.stack()
     try:
-        maybe_class_frame = frames[level_of_class_stackframe]
-        statement_list = maybe_class_frame[4]
+        maybe_class_frame = inspect.currentframe()
+        for _ in range(level_of_class_stackframe):
+            maybe_class_frame = maybe_class_frame.f_back
+        statement_list = inspect.getframeinfo(maybe_class_frame)[3]
```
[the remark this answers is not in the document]
strongoier: Nice catch! This can make the code even cleaner. Updated.

sjwsl (approved):
`import taichi as ti` has been dramatically faster!! LGTM!

yihong0618:
Cool

PR #3655, opened 2021-11-30
[Refactor] Reduce reference count introduced by object closure

qiao-bo:
Before the change, any `Ndarray` obj will have an extra ref count due to the introduced `obj.__closure___`. This can be checked by `import gc 
  gc.get_referrers`. The problem is that the destructor will not be called until the program finish, due to this extra count. This prevents optimizations such as memory recycling.

After the change, the extra ref count is eliminated and object can be properly recycled when out of scope.


strongoier (approved):
LGTM!

ailzhang (approved):
LGTM!

PR #3657, opened 2021-12-01
[cuda] Query the max block count from the hardware

bobcao3:
Related issue = #


bobcao3:
Is there a way to trigger an regression benchmark / CI on a PR?

k-ye (approved):
LGTM!

taichi/backends/cuda/cuda_driver.h
```diff
@@ -30,6 +30,7 @@ constexpr uint32 CU_STREAM_NON_BLOCKING = 0x1;
 constexpr uint32 CU_MEM_ATTACH_GLOBAL = 0x1;
 constexpr uint32 CU_MEM_ADVISE_SET_PREFERRED_LOCATION = 3;
 constexpr uint32 CU_DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_X = 2;
+constexpr uint32 CU_DEVICE_ATTRIBUTE_MAX_BLOCKS_PER_MULTIPROCESSOR = 106;
```
qiao-bo, line 33: why 106 here? do we hardcode these values to a specific card or?

qiao-bo (commented):
Thanks!

taichi/backends/cuda/cuda_driver.h
bobcao3: This is a enum from cuda headers

Issue #3634, opened 2021-11-28
[RFC] Caching memory allocator for ndarray


qiao-bo:
**Concisely describe the proposed feature**
We want to have a caching memory allocator for `Ndarray`, which can improve memory utilization on e.g., GPU device memory.

**Describe the solution you'd like (if any)**
Currently, Ndarray's own memory allocator does not perform recycling: 
https://github.com/taichi-dev/taichi/blob/71dd17426b62327864a21c65ca570812997c33ba/taichi/program/ndarray.cpp#L20
We need to add a release/deallocate mechanism in the destructor to recycle the memory into to a memory pool:
https://github.com/taichi-dev/taichi/blob/71dd17426b62327864a21c65ca570812997c33ba/taichi/program/ndarray.cpp#L34
Right now memory actions of different backends in Taichi are unified into [`DeviceAllocation`](https://github.com/taichi-dev/taichi/blob/71dd17426b62327864a21c65ca570812997c33ba/taichi/backends/device.h#L368). Here we take CUDA device as an example to present a rough picture of the proposal: CUDA device can have a `cuda_caching_allocator` or `ccalloc`. which keeps track of a block pool for all allocated/released memory blocks. `ccalloc` exposes two methods:
- `Ptr allocate(sz)` returns the requested memory block of size `sz`. ccalloc rounds up the size based on alignment, searches the block pool, and performs one of the followings:
    - return if sz is available
    - take the smallest possible block that is larger than sz, and split the block into `returned` and `remaining`. return the `returned` and recycle the `remaining` back info the pool.
    - otherwise go ahead and allocate a new block
- `release(ptr, sz)` simply put the block back into the pool.   

The mentioned solution performs the so-called block splitting, which facilitates the reuse of larger blocks (size > sz). In addition, it is also possible to perform block merging, which targets small blocks recycling. To realize block merging, each block could maintain a list with `prev` and `next` that pointing to its neighbors (the only way to reuse smaller blocks is to make sure they are continuous in memory). When a block is released, we can merge with its neighboring blocks if possible. The idea is to put a large block instead of many small pieces back to the memory pool, such that the `allocate` can be hit with cached as much as possible. 

**Additional comments**
Ref: [Pytorch CUDACachingAllocator](https://github.com/pytorch/pytorch/blob/a31aea8eaa99a5ff72b5d002c206cd68d5467a5e/c10/cuda/CUDACachingAllocator.h)

Comments/suggestions are welcome!

PR #3659, opened 2021-12-01
[Refactor] Simplify allocation logic in runtime

qiao-bo:
Related issue = #3634


taichi/runtime/llvm/runtime.cpp
```diff
@@ -809,13 +809,10 @@ Ptr LLVMRuntime::allocate_from_buffer(std::size_t size, std::size_t alignment) {
...
-    auto alignment_bytes =
-        alignment - 1 -
-        ((std::size_t)preallocated_head + alignment - 1) % alignment;
-    size += alignment_bytes;
-    if (preallocated_head + size <= preallocated_tail) {
-      ret = preallocated_head + alignment_bytes;
-      preallocated_head += size;
+    auto size_aligned = taichi::iroundup(size, alignment);
+    if (preallocated_head + size_aligned <= preallocated_tail) {
+      ret = preallocated_head;
```
ailzhang, line 814: Looks like we've changed `ret` here, does it matter?

qiao-bo: we do, that's how it's simplified. The invariant does not change, which is `ret` should be always aligned. Now we can just update `preallocated_head` instead of `preallocated_head`+`alignment_bytes`.

qiao-bo:
Thanks to the discussion with @strongoier @ailzhang, The change in this PR only applies when alignment is fixed (const). The previous implementation can cope with changed alignment. We close this PR.

PR #3662, opened 2021-12-01
[opengl] [refactor] Expose use_gles in CompileConfig.

ailzhang:
This PR converts use_gles from a constexpr to a static variable set in
`initialize_opengl` so that users can optionally turn it on/off using
`ti.init(use_gles=True)`.

Related issue = #


k-ye (approved):
Cool!

PR #3669, opened 2021-12-01
[error] Fix tab and long line in error messages

### LINKED ISSUE #3666: Error message not correctly printed when tab appears in the source code
**Describe the bug**
We now convert all tabs to 4 spaces in the source code. However, a tab can be 1-4 spaces long depending on its position. So, the tabs are not correctly handled in the error messages. Using textwrap.fill can solve this problem.
Many people like using tabs to align numbers in arrays.
Here is an example:
```python
foo = [1,	2,	3,	4,	5,	6,	]
bar = [12,	34,	56,	7,	89,	10,	]
```
Now, it looks like
```python
foo = [1,    2,    3,    4,    5,    6,    ]
bar = [12,    34,    56,    7,    89,    10,    ]
```
If it is correctly indented with tab size=4, it will look like
```python
foo = [1,   2,  3,  4,  5,  6,  ]
bar = [12,  34, 56, 7,  89, 10, ]
```

lin-hitonami:
Related issue = close #3666
I have to disable code format check on test_exception.py, otherwise it will always turn the tabs to spaces and I cannot disable it even if I disabled yapf.

### DIFF
--- a/tests/python/test_exception.py
+++ b/tests/python/test_exception.py
@@ -84,3 +84,61 @@ On line {lineno + 5} of file "{file}":
 TypeError: 'NoneType' object is not callable"""
     print(e.value.args[0])
     assert e.value.args[0] == msg
+
+
+@ti.test()
+def test_tab():
+    frameinfo = getframeinfo(currentframe())
+    with pytest.raises(ti.TaichiCompilationError) as e:
+        # yapf: disable
+        @ti.kernel
+        def foo():
+            a(11,	22,	3)
+        foo()
+        # yapf: enable
+    lineno = frameinfo.lineno
+    file = frameinfo.filename
+    if version_info < (3, 8):
+        msg = f"""\
+On line {lineno + 5} of file "{file}":
+            a(11,   22, 3)
+TypeError: 'NoneType' object is not callable"""
+    else:
+        msg = f"""\
+On line {lineno + 5} of file "{file}":
+            a(11,   22, 3)
+            ^^^^^^^^^^^^^^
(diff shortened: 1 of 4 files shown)


tests/python/test_exception.py
```diff
@@ -84,3 +84,61 @@ def foo():
...
+        def foo():
+            a(11,	22,	3)
+        foo()
+        # yapf: enable
+    lineno = frameinfo.lineno
+    file = frameinfo.filename
+    if version_info < (3, 8):
+        msg = f"""\
+On line {lineno + 5} of file "{file}":
+            a(11,   22, 3)
```
strongoier, line 104: I'm a little curious how do we determine the number of spaces here. IMHO the number of spaces shown depends on editor configurations. For example, in my browser this line doesn't align with the function definition. What is the goal we aim to achieve?

lin-hitonami: We have to convert tabs to spaces in the source code when printing because otherwise we cannot generate the hint line accurately. As python allows to indent using tab instead of 4 spaces, I just assumed that the tabsize of python program is 4. Or should we make this configurable?

strongoier: What is the advantage of using `textwrap.fill` over replacing `\t` with 4 spaces?

lin-hitonami: Updated the example in the issue. Also, this is the right way to treat tabs, and replacing `\t` with 4 spaces is simply wrong.

strongoier (approved):
LGTM!

Issue #2489, opened 2021-07-05
Clean up _taichi_skip_traceback


k-ye:
`_taichi_skip_traceback` was added to skip a few non-important Taichi internal stack frames. This is way too ad-hoc. We should modularize Taichi itself, and remove this.

ailzhang:
cc: @turbo0628 This issue is fairly simple in the sense that it just need to grep and delete all the _taichi_skip_traceback lines. But it might give you some sense how the frontend works :D

turbo0628:
Thanks! It's pretty interesting to trace the code execution path :)

strongoier:
Closed via #3902.

PR #3671, opened 2021-12-01
[refactor] Remove unused functions in misc.

ailzhang:
Related issue = #


python/taichi/misc/util.py
```diff
@@ -230,26 +197,6 @@ def get_kernel_stats():
     return _ti_core.get_kernel_stats()
 
 
-def print_async_stats(include_kernel_profiler=False):
```
strongoier, line 233: This function seems related to our async project. @k-ye wdyt?

ailzhang: @strongoier I deleted mainly because it wasn't even in the `__all__` in this file, so it's very likely that this one is not used...

python/taichi/misc/error.py
```diff
@@ -1,47 +0,0 @@
-import sys
-import traceback
-
-from colorama import Fore
-
-
-def enable_excepthook():
```
strongoier, line 7: Not sure if we definitely want to remove this. If we finally decide to remove this, we can remove all `_taichi_skip_traceback` in the codebase and all docs or configs related to this feature (https://github.com/taichi-dev/taichi/search?q=excepthook).

As we have @lin-hitonami 's new error report mechanism now, do we still want to keep the hacky pretty printing of the traceback stack? @k-ye

ailzhang: I'm fully down to remove _taichi_skip_traceback :D

k-ye: +1, https://github.com/taichi-dev/taichi/issues/2489

python/taichi/misc/util.py
k-ye: This is some random util function...

strongoier (approved):
LGTM! BTW do you plan to remove `_taichi_skip_traceback` related stuff in another PR?

ailzhang:
@strongoier Yea that one should be separate PR for sure, I'll open an issue first to see if anyone is interested ;)

PR #3672, opened 2021-12-01
[refactor] Move image.py from misc/ to tools/ and remove it from top level package

strongoier:
Related issue = #


ailzhang (approved):
LGTM!

PR #3680, opened 2021-12-02
[ci] Run cpp tests in unix_docker_test.sh.

ailzhang:
The current situation is a bit confusing that we run c++ tests in
unix_docker_build.sh. It didn't run in unix_docker_test.sh since we
didn't copy the build/ into docker image. This PR fixes the issue that
we now run c++ tests in unix_docker_test.sh.

Related issue = #


ailzhang:
Okay I give up... docker is so mysterious :( I'll open an issue for this instead since it's not critical.

PR #3681, opened 2021-12-02
[refactor] Turn off empty root buffer warning for finalize_for_aot.

ailzhang:
This is a duplicate of #3674.

Now that we support ndarrays for AOT, it's possible/valid to have all
ndarray inputs and no field input as shown in test_mpm88_ndarray. So
let's turn off the warning here to avoid confusion.

Related issue = #


k-ye (approved):
Thanks!

PR #3688, opened 2021-12-03
[Lang] Make example_any_arrays optional.

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #3692
* #3691
* #3690
* #3689
* __->__ #3688
* #3687

Now that users have options to either specify `element_shapes &
field_dim` in the kernel argument annotation, or provided a dict of
example any_arr.

PR #3687, opened 2021-12-03
[Lang] Add optional element_shapes & field_dim annotation to ndarray.

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #3692
* #3691
* #3690
* #3689
* #3688
* __->__ #3687

PR #3691, opened 2021-12-03
[refactor] [opengl] For each taichi kernel, save a {kernel_name, CompiledProgram} pair instead of a vector.

ailzhang:
* #3692
* __->__ #3691
* #3690
* #3689
* #3688
* #3687

Nested data structures are confusing... Let's just use an unordered_map with taichi kernel_name as
keys.

### DIFF
--- a/taichi/backends/opengl/aot_data.h
+++ b/taichi/backends/opengl/aot_data.h
@@ -9,20 +9,6 @@ namespace taichi {
 namespace lang {
 namespace opengl {
 
-struct AotCompiledKernel {
-  CompiledProgram program;
-  std::string identifier;
-
-  TI_IO_DEF(program, identifier);
-};
-
-struct AotCompiledKernelTmpl {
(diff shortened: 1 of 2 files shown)


taichi/backends/opengl/aot_data.h
```diff
@@ -44,8 +30,8 @@ struct CompiledFieldData {
 };
 
 struct AotData {
-  std::vector<AotCompiledKernel> kernels;
-  std::vector<AotCompiledKernelTmpl> kernel_tmpls;
+  std::unordered_map<std::string, CompiledProgram> kernels;
```
k-ye, line 33: Shall we rename this to `CompiledTaichiKernel`?

k-ye (approved):

taichi/backends/opengl/aot_data.h
ailzhang: Agreed! Will do a followup PR for that!

PR #3692, opened 2021-12-03
[refactor] [opengl] Properly save both scalar args and array args.

ailzhang:
* __->__ #3692
* #3691
* #3690
* #3689
* #3688
* #3687

### DIFF
--- a/taichi/backends/opengl/opengl_api.h
+++ b/taichi/backends/opengl/opengl_api.h
@@ -49,7 +49,13 @@ struct CompiledOffloadedTask {
   TI_IO_DEF(name, src, workgroup_size, num_groups);
 };
 
-struct CompiledArrayData {
+struct ScalarArg {
+  size_t offset_in_bytes_in_args_buf{0};
+
+  TI_IO_DEF(offset_in_bytes_in_args_buf);
+};
+
+struct CompiledArrayArg {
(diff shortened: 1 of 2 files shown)


taichi/backends/opengl/opengl_api.h
```diff
@@ -88,14 +94,16 @@ struct CompiledProgram {
   std::unordered_map<int, irpass::ExternalPtrAccess> ext_arr_access;
   std::vector<std::string> str_table;
   UsedFeature used;
-  mutable std::unordered_map<int, CompiledArrayData> arr_args;
+  std::unordered_map<int, ScalarArg> scalar_args;
+  mutable std::unordered_map<int, CompiledArrayArg> arr_args;
```
k-ye, line 98: nit: I feel like we can remove `mutable` now?

k-ye (approved):
LGTM!

taichi/backends/opengl/opengl_api.h
ailzhang: hmmm not yet for now, currently this is coupled with the JIT runtime logic "if runtime size doesn't change for **external** (note this is not ndarray) array, don't allocate_memory again, just reuse". 
We can consider decoupling external array and ndarray in the future. :D

PR #3695, opened 2021-12-03
[refactor] Let build_xxx in ASTTransformer return node.ptr instead of node

### LINKED ISSUE #3694: Random value is returned when kernel return type is specified but no return statement is executed
**Describe the bug**
If a taichi kernel has a return type but no return statement is executed, the kernel returns a random value. 

**To Reproduce**

```py
import taichi as ti
ti.init()
@ti.kernel
def foo()->ti.i32:
    pass

print(foo())
```

**Log/Screenshots**

```
[Taichi] version 0.8.7, llvm 10.0.0, commit 4590adca, linux, python 3.6.13
[Taichi] Starting on arch=x64
-1377067008
```

lin-hitonami:
Related issue = #3694
This is a necessary refactoring to resolve #3694.
Since we do not modify the AST anymore after the refactoring, we do not need to assign the new node back to the old node in the AST. So, we can simplify the programming by returning node.ptr in build_xxx of expression nodes and None for build_xxx of statement nodes. In this way, many codes like `a = build_stmt(ctx, b).ptr` can be simplified to `a = build_stmt(ctx, b)`.

ASTTransformerCheck will be removed in the near future, the modification to it is just a workaround to pass the tests of real functions with multiple returns.


strongoier (changes requested):
Would you merge with master branch? I think there must be something to resolve as #3709 is in now.

strongoier (approved):
LGTM!

PR #3696, opened 2021-12-03
[refactor] Add doc string for ndarray element_shapes & field_dim.

ailzhang:
* #3697
* __->__ #3696


python/taichi/type/annotations.py
```diff
@@ -6,6 +6,8 @@ class ArgAnyArray:
 
     Args:
         element_dim (Union[Int, NoneType], optional): None if not specified (will be treated as 0 for external arrays), 0 if scalar elements, 1 if vector elements, and 2 if matrix elements.
+        element_shapes (Union[Tuple[Int], NoneType]): None if not specified, shapes of each element. For example, element_shapes must be 1d for vector and 2d tuple for matrix.
```
ailzhang, line 9: re variable name here: do we prefer `element_shapes` or `element_shape` here? @k-ye

k-ye: Let's go with `element_shape`

python/taichi/type/annotations.py
```diff
@@ -6,25 +6,27 @@ class ArgAnyArray:
 
     Args:
         element_dim (Union[Int, NoneType], optional): None if not specified (will be treated as 0 for external arrays), 0 if scalar elements, 1 if vector elements, and 2 if matrix elements.
+        element_shape (Union[Tuple[Int], NoneType]): None if not specified, shapes of each element. For example, element_shape must be 1d for vector and 2d tuple for matrix.
```
strongoier, line 9:
```suggestion
        element_shape (Union[Tuple[Int], NoneType]): None if not specified, shape of each element. For example, element_shape must be 1d for vector and 2d tuple for matrix. This argument is ignored for external arrays for now.
```

python/taichi/type/annotations.py
```diff
@@ -6,25 +6,27 @@ class ArgAnyArray:
...
+        field_dim (Union[Int, NoneType]): None if not specified, number of field dimensions.
```
strongoier, line 10:
```suggestion
        field_dim (Union[Int, NoneType]): None if not specified, number of field dimensions. This argument is ignored for external arrays for now.
```

strongoier (changes requested):

strongoier (approved):
LGTM!

PR #3699, opened 2021-12-03
[ci] Auto generate manylinux dockerfile

qiao-bo:
Add an option in the script to generate our CentOS based dockerfile: `python Dockerfile_generator.py -o manylinux2014 -t cpu`. Locally tested.


rexwangcc (approved):
LGTM, thanks! Btw, the generated Dockerfile using the cli is slightly different from the existing manylinux one, shall we overwrite the existing one?

ailzhang (commented):
Can we also add "#generated by python bla.. config" on top of the generated Dockerfile? ;)

Issue #3679, opened 2021-12-02
Cross compile taichi_core.so to other targets


ailzhang:
In [Taichi AOT Module RFC](https://github.com/taichi-dev/taichi/issues/3642) we mainly covered how to serialize taichi kernels and fields/ndarrays. One missing piece is how to load & execute on a different target. 

For example, in the android demo @Leonz5288 is working on (will be released soon), we first serialized the kernels/ndarrays for GLES, but still had to deserialize the JSON output to proper data structures in Java and wrote a runtime to bind SSBOs and execute those shaders. 

An alternative that worth exploring is cross compiling taichi_core.so to different targets, say Android so that we can reuse what the deserialization part and a lot of data structures/logic used in Taichi JIT runtime. This is something that @ghuau-innopeak is looking into (thanks a ton!!). 

cc: @k-ye @ghuau-innopeak please feel free correct me if anything is wrong!

leirbag:
One idea I have in mind (seems to work so far...) is to introduce a new option `TI_WITH_PYTHON` to the CMake build system. By default, the option could be ON but in the case of cross-compiling to a platform that doesn't support Python (Android for example), we could set this option to OFF.
With this option, we could compile the `taichi/python/` source code to a static library `libtaichi_python.a` that would be linked to libtaichi_core.so. So for example, if `TI_WITH_PYTHON` is OFF, we libtaichi_core.so would not contain any Python code but if this is set, it will be linked to libtaichi_core.so as it was before.

I think in the long term, we might even want to move libtaichi_python to a shared library, like the backends, but this would involve too many changes in the Python code, so this is why I'm thinking of this path for now, and this is pretty flexible too... :)

ghuau-innopeak:
One idea I have in mind (seems to work so far...) is to introduce a new option `TI_WITH_PYTHON` to the CMake build system. By default, the option could be ON but in the case of cross-compiling to a platform that doesn't support Python (Android for example), we could set this option to OFF.
With this option, we could compile the `taichi/python/` source code to a static library `libtaichi_python.a` that would be linked to libtaichi_core.so. So for example, if `TI_WITH_PYTHON` is OFF, we libtaichi_core.so would not contain any Python code but if this is set, it will be linked to libtaichi_core.so as it was before.

I think in the long term, we might even want to move libtaichi_python to a shared library, like the backends, but this would involve too many changes in the Python code, so this is why I'm thinking of this path for now, and this is pretty flexible too... :)

[...]

PR #3703, opened 2021-12-03
[Vulkan] Add support of loading AOT modules and fields

ghuau-innopeak:
Related issue = #3679

### DIFF
--- a/taichi/backends/vulkan/aot_module_builder_impl.cpp
+++ b/taichi/backends/vulkan/aot_module_builder_impl.cpp
@@ -15,6 +15,32 @@ AotModuleBuilderImpl::AotModuleBuilderImpl(
     : compiled_structs_(compiled_structs), runtime_(runtime) {
 }
 
+uint32_t AotModuleBuilderImpl::to_vk_dtype_enum(DataType dt) {
+  if (dt == PrimitiveType::u64) {
+    return 0;
+  } else if (dt == PrimitiveType::i64) {
+    return 1;
+  } else if (dt == PrimitiveType::u32) {
+    return 2;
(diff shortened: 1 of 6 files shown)


ghuau-innopeak:
For `to_vk_dtype_enum`, right now, I'm using constant values with a static function, but I'm wondering if we even need to serialize dtype as we already serialize the name?

taichi/backends/vulkan/aot_module_builder_impl.cpp
```diff
@@ -46,6 +86,26 @@ void AotModuleBuilderImpl::dump(const std::string &output_dir,
   ts.write_to_file(txt_path);
 }
 
+void AotModuleBuilderImpl::load(const std::string &output_dir,
+                                const std::string &filename) {
+  const std::string bin_path =
+      fmt::format("{}/{}_metadata.tcb", output_dir, filename);
```
ailzhang, line 92: nit: filename is no longer used? ;)

ailzhang (approved):
Looks great! Thanks! 
Shall we add a C++ test saving & loading as followup? ;)

taichi/backends/vulkan/aot_module_builder_impl.cpp
ghuau-innopeak: Oops, my bad! should be fixed :)

ghuau-innopeak:
> Looks great! Thanks! Shall we add a C++ test saving & loading as followup? ;)

@ailzhang Indeed, and some example code in C++ for `aot_save` was merged recently that could be re-used for testing too (#3707)

PR #3702, opened 2021-12-03
[LLVM] Add missing pre-processor macro when LLVM is disabled

ghuau-innopeak:
Some components were still referencing LLVM with TI_WITH_LLVM set to
false triggering a build failure.

Related issue = #3679

PR #3704, opened 2021-12-03
[CPU][Vulkan] Add missing ifdef macros if CPU is disabled

ghuau-innopeak:
Related issue = #3679

### DIFF
--- a/taichi/backends/device.cpp
+++ b/taichi/backends/device.cpp
@@ -1,9 +1,11 @@
 #include <taichi/backends/device.h>
-#include <taichi/backends/cpu/cpu_device.h>
 
 #if TI_WITH_VULKAN
 #include <taichi/backends/vulkan/vulkan_device.h>
 #include <taichi/backends/interop/vulkan_cpu_interop.h>
+#if TI_WITH_LLVM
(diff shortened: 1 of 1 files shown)


AmesingFlank:
hmm, can't we just use `TI_WITH_LLVM` as guards?

ghuau-innopeak:
> hmm, can't we just use `TI_WITH_LLVM` as guards?

Correct... :) I completely missed that the CPU source code was only added when LLVM was enabled. I will update the changes, the question now is... should I close this PR and push the changes in the other PR related to missing TI_WITH_LLVM? (#3702) Or should I still keep this one with the updated changes?

AmesingFlank:
> > hmm, can't we just use `TI_WITH_LLVM` as guards?
> 
> @AmesingFlank Correct... :) I completely missed that the CPU source code was only added when LLVM was enabled. I will update the changes, the question now is... should I close this PR and push the changes in the other PR related to missing TI_WITH_LLVM? (#3702) Or should I still keep this one with the updated changes?

I think either is fine. Up to u really. I just approved #3702 btw.

ghuau-innopeak:
Closing, PR merged with #3702 as all the changes are now related to LLVM.

PR #3705, opened 2021-12-04
[example] Build cpp examples

### LINKED ISSUE #3630: [IR] Building CHI IR examples

AmesingFlank:
fixes #3630

This PR fixes the cmake for what was originally called `chi_examples`. The cmake file works for both windows and linux, and It will be built in CI as well. 

I renamed it to `cpp_examples`, because I intend to write a few AOT examples as well.


k-ye (approved):
LGTM!

PR #3707, opened 2021-12-04
[example] SPIR-V AOT example in C++

AmesingFlank:
This PR adds an example of using the C++ API to AOT compile a kernel to SPIRV.


cpp_examples/main.cpp
```diff
@@ -327,8 +327,84 @@ void autograd() {
   std::cout << std::endl;
 }
 
+void aot() {
```
ailzhang, line 330: maybe `aot_save` is more accurate? ;)

ailzhang (approved):
Nice! Thanks! 
Btw can we split these into different files in the followup PR? :P

Issue #2610, opened 2021-07-29
[Lang] Support annotated assignment


xumingkuan:
**Concisely describe the proposed feature**
We would like to support annotated assignment in the future, for example,
```python
a: ti.i32 = 1
```

Since Taichi is *statically typed*, code snippet like this (which compiles in Python) would not compile in Taichi:
```python
a: int = 1
a: float = 2.0
```

**Describe the solution you'd like (if any)**
Add a method `build_AnnAssign` in `stmt_builder.py`. The content of the method will be similar to `build_Assign`, and the statement `a: ti.i32 = 1` should be translated into the following in the CHI IR:
```
<i32> $0 = alloca
<i32> $1 = const [1]
<i32> $2 : local store [$0 <- $1]
```

**Additional comments**
The statement corresponds to `ast.AnnAssign` in the Python AST. See https://docs.python.org/3/library/ast.html#ast.AnnAssign for more information.

woooodyye:
I would like to work on this issue. A simple approach would be cast the value based on the type annotation we provide. In case of already-created variable, we compare the type annotation with the type established in the `ctx`. If the types are different, we would raise a syntaxError

strongoier:
Welcome to the Taichi community! Your proposal looks great. This issue is assigned to you :-)

YouJiacheng:
Can we make `taichi` ignore annotation in annotated assignment when annotation is not a primitive type, or allow `ti.template()`/`typing.Any` to serve as a dummy annotation(without any compile-time or runtime effect, only edit-time effect for developer experience)?
With this we can improve developer experience leveraging the existing intelliSense(auto complete, jump to source, semantic highlight) infrastructure for Python.
Use case:
```python
from typing import Generic, TypeVar

import taichi as ti
from taichi.lang.struct import StructType

T = TypeVar('T')

vec3 = ti.types.vector(3, ti.float32)
vec2 = ti.types.vector(2, ti.float32)

# make IntelliSense happy
class TypedStructType(Generic[T], StructType): ...

def ti_dataclass(cls: type[T]) -> TypedStructType[T]:
    return ti.dataclass(cls)

def typing(_: TypedStructType[T], obj) -> type[T]:
    return obj

@ti_dataclass
class Triangle:
    V0: vec3
    V1: vec3
    V2: vec3
TriangleT = typing(Triangle, ti.template())

@ti_dataclass
class TextureTriangle:
    V0: vec2
    V1: vec2
    V2: vec2
[...]

PR #3709, opened 2021-12-04
[Lang] Support annotated assignment

woooodyye:
Related issue = #2610


strongoier:
Hi @woooodyye. Which PR is the version you would like to get reviewed, this or #3701?

woooodyye:
hi, can I get this PR draft reviewed ? I think #3701 has formatting issues. Thanks. @strongoier

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -18,6 +18,53 @@ def build_Name(ctx, node):
...
+                    "Static assign cannot be used on elements in arrays")
+            ctx.create_variable(target.id, value)
+            var = value
+        elif is_local and not ctx.is_var_declared(target.id):
+            var = cast_type(value, annotation)
+            var = ti.expr_init(var)
+            ctx.create_variable(target.id, var)
+        else:
+            var = target.ptr
+            if str(var.ptr.get_ret_type()) != annotation:
```
strongoier, line 62: 1. Why do we use `str` here?
2. Could you also add a test for this invalid case?

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -18,6 +18,53 @@ def build_Name(ctx, node):
         node.ptr = ctx.get_var_by_name(node.id)
         return node
 
+    @staticmethod
+    def build_AnnAssign(ctx, node):
+        annotation = getattr(node.annotation, 'attr')
```
strongoier, line 23: Why not `node.annotation.attr`?

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -18,6 +18,53 @@ def build_Name(ctx, node):
...
+         Args:
+            ctx (ast_builder_utils.BuilderContext): The builder context.
+            target (ast.Name): A variable name. `target.id` holds the name as
+            a string.
+            annotation: A type we hope to assign to the target
+            value: A node representing the value.
+            is_static_assign: A boolean value indicating whether this is a static assignment
+        """
+        is_local = isinstance(target, ast.Name)
+        if is_static_assign:
```
strongoier, line 50: We should not allow annotated static assign.

strongoier (changes requested):
Great!

woooodyye:
please ignore the review request

woooodyye:
Right Now. Since annotations are not considered as variables, we need to explicitly define a primitive type as a value for the primitive types to be considered as variables.

For Instance:
`a: ti.i32 = 1`
would not work.

Instead, we need to do something like
```
my_int = ti.i32
a : my_int = 1
```

### Note: the primitive type you declare doesn't necessarily need to be the type annotation you want to use, It's totally fine to do
```
my_int = ti.i32
a : ti.f32 = 1.
b : ti.f64 = 2
.......
```

The whole point of this `my_int = ti.i32` declaration statement is to introduce ti.primitivetypes into the global variables.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -18,6 +18,53 @@ def build_Name(ctx, node):
         node.ptr = ctx.get_var_by_name(node.id)
         return node
 
+    @staticmethod
+    def build_AnnAssign(ctx, node):
+
```
strongoier, line 23:
```suggestion
```

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -18,6 +18,53 @@ def build_Name(ctx, node):
...
+        node.ptr = ASTTransformer.build_ann_assign_basic(
+            ctx, node.target, node.value.ptr, is_static_assign,
+            node.annotation.ptr)
+        return node
+
+    @staticmethod
+    def build_ann_assign_basic(ctx, target, value, is_static_assign,
+                               annotation):
+        """Build basic assginment like this: target : annotation = value.
```
strongoier, line 40:
```suggestion
        """Build an annotated assignment like this: target: annotation = value.
```

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -18,6 +18,53 @@ def build_Name(ctx, node):
...
+            node.value, ast.Call) and ASTResolver.resolve_to(
+                node.value.func, ti.static, globals())
+
+        node.ptr = ASTTransformer.build_ann_assign_basic(
+            ctx, node.target, node.value.ptr, is_static_assign,
+            node.annotation.ptr)
+        return node
+
+    @staticmethod
+    def build_ann_assign_basic(ctx, target, value, is_static_assign,
```
strongoier, line 38: I suggest renaming it as `build_assign_annotated` to match other names like `build_assign_basic`, `build_assign_unpack`.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -18,6 +18,53 @@ def build_Name(ctx, node):
...
+        """
+        is_local = isinstance(target, ast.Name)
+        anno = ti.expr_init(annotation)
+        if is_static_assign:
+            raise TaichiSyntaxError(
+                "Static assign cannot be used on annotated assignment")
+        if is_local and not ctx.is_var_declared(target.id):
+            var = ti.cast(value, anno)
+            var = ti.expr_init(var)
+
```
strongoier, line 58:
```suggestion
```

tests/python/test_assign.py
```diff
@@ -57,3 +57,33 @@ def func_assign():
         assert a == 1
 
     func_assign()
+
+
+@ti.test(debug=True)
+def test_assign_ann():
+    @ti.kernel
+    def func_ann():
+        #need to introduce it
```
strongoier, line 66:
```suggestion
        # need to introduce ti as a global var
```

tests/python/test_assign.py
```diff
@@ -57,3 +57,33 @@ def func_assign():
...
+
+@ti.test()
+def test_assign_ann_over():
+    @ti.kernel
+    def func_ann_over():
+        my_int = ti.i32
+        d: my_int = 2
+        d: ti.f32 = 2.0
+
+    with pytest.raises(ti.lang.exception.TaichiCompilationError) as e:
```
strongoier, line 86:
```suggestion
    with pytest.raises(ti.TaichiCompilationError):
```

tests/python/test_assign.py
```diff
@@ -57,3 +57,33 @@ def func_assign():
...
+        func_ann_over()
+
+    assert e.type is ti.lang.exception.TaichiCompilationError
```
strongoier, line 89:
```suggestion
```
This is redundant.

strongoier (changes requested):
Great work! Only some nits.

strongoier (approved):
LGTM!

PR #3711, opened 2021-12-05
[vulkan] Further decouple SPIRV codegen from Vulkan runtime

### LINKED ISSUE #3708: Vulkan AOT requires materializing runtime
See the AOT example in PR #3707:
```cpp
void aot() {
  using namespace taichi;
  using namespace lang;
  auto program = Program(Arch::vulkan);

  program.config.advanced_optimization = false;

  int n = 10;
  program.materialize_runtime();
  auto *root = new SNode(0, SNodeType::root);
  auto *pointer = &root->dense(Axis(0), n, false);
  auto *place = &pointer->insert_children(SNodeType::place);
  place->dt = PrimitiveType::i32;
  program.add_snode_tree(std::unique_ptr<SNode>(root), /*compile_only=*/false);

  auto aot_builder = program.make_aot_module_builder(Arch::vulkan);

  std::unique_ptr<Kernel> kernel_init, kernel_ret;

  {
    /*
    @ti.kernel
    def init():
      for index in range(n):
        place[index] = index
    */
    IRBuilder builder;
    auto *zero = builder.get_int32(0);
    auto *n_stmt = builder.get_int32(n);
    auto *loop = builder.create_range_for(zero, n_stmt, 1, 0, 4);
    {
      auto _ = builder.get_loop_guard(loop);
      auto *index = builder.get_loop_index(loop);
      auto *ptr = builder.create_global_ptr(place, {index});
      builder.create_global_store(ptr, index);
    }

    kernel_init =
        std::make_unique<Kernel>(program, builder.extract_ir(), "init");
  }

  {
    /*
    @ti.kernel
    def ret():
      sum = 0
      for index in place:
        sum = sum + place[index];
      return sum
    */
    IRBuilder builder;
    auto *sum = builder.create_local_var(PrimitiveType::i32);
    auto *loop = builder.create_struct_for(pointer, 1, 0, 4);
    {
      auto _ = builder.get_loop_guard(loop);
      auto *index = builder.get_loop_index(loop);
      auto *sum_old = builder.create_local_load(sum);
      auto *place_index =
          builder.create_global_load(builder.create_global_ptr(place, {index}));
      builder.create_local_store(sum, builder.create_add(sum_old, place_index));
    }
    builder.create_return(builder.create_local_load(sum));

    kernel_ret = std::make_unique<Kernel>(program, builder.extract_ir(), "ret");
    kernel_ret->insert_ret(PrimitiveType::i32);
  }

  aot_builder->add_field("place", place, true, place->dt, {n}, 1, 1);
  aot_builder->add("init", kernel_init.get());
  aot_builder->add("ret", kernel_ret.get());
  aot_builder->dump(".", "aot.tcb");
  std::cout << "done" << std::endl;
}
```
If the line `program.materialize_runtime();` is removed, then this example crashes.

AmesingFlank:
fixes #3708 

Previously, the SPIRV codegen requires an instance of `VkRuntime`. The runtime is in fact only used to provide a `Device*` and the compiled SNode structs. Moreover, the `Device*` is only used to query device capabilities.

This PR removes the `VkRuntime*` from the codegen. When there isn't a runtime, the `VulkanProgram` will own the compiled structs, and feed them to the codegen. The PR also adds a `AotTargetDevice`, which is used to report device capabilities to the codegen.

The result of this PR is that #3708 is fixed. In other words, we can run Vulkan AOT to generate SPIRV without ever creating a `VulkanDevice` or a `VkRuntime`.

### DIFF
--- a/taichi/backends/vulkan/aot_module_builder_impl.cpp
+++ b/taichi/backends/vulkan/aot_module_builder_impl.cpp
@@ -25,29 +25,33 @@ class AotTargetDevice : public Device {
   }
 
   DeviceAllocation allocate_memory(const AllocParams &params) override {
-    TI_NOT_IMPLEMENTED
+    TI_NOT_IMPLEMENTED;
   }
-  void dealloc_memory(DeviceAllocation handle)
-      override{TI_NOT_IMPLEMENTED} std::unique_ptr<Pipeline> create_pipeline(
-          const PipelineSourceDesc &src,
-          std::string name = "Pipeline") override {
-    TI_NOT_IMPLEMENTED
+  void dealloc_memory(DeviceAllocation handle) override {
+    TI_NOT_IMPLEMENTED;
+  }
+  std::unique_ptr<Pipeline> create_pipeline(
+      const PipelineSourceDesc &src,
+      std::string name = "Pipeline") override {
+    TI_NOT_IMPLEMENTED;
   }
   void *map_range(DevicePtr ptr, uint64_t size) override {
-    TI_NOT_IMPLEMENTED
+    TI_NOT_IMPLEMENTED;
   }
   void *map(DeviceAllocation alloc) override {
-    TI_NOT_IMPLEMENTED
+    TI_NOT_IMPLEMENTED;
   }
   void unmap(DevicePtr ptr) override {
-    TI_NOT_IMPLEMENTED
+    TI_NOT_IMPLEMENTED;
   }
   void unmap(DeviceAllocation alloc) override {
-    TI_NOT_IMPLEMENTED
+    TI_NOT_IMPLEMENTED;
   }
-  void memcpy_internal(DevicePtr dst, DevicePtr src, uint64_t size) override{
-      TI_NOT_IMPLEMENTED} Stream *get_compute_stream() override {
-    TI_NOT_IMPLEMENTED
+  void memcpy_internal(DevicePtr dst, DevicePtr src, uint64_t size) override {
+    TI_NOT_IMPLEMENTED;
+  }
+  Stream *get_compute_stream() override {
+    TI_NOT_IMPLEMENTED;
   }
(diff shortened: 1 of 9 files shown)


ailzhang:
@AmesingFlank Hmmm I wonder what'll happen if we want to use jit evaluator in constant folding without a runtime?

AmesingFlank:
> @AmesingFlank Hmmm I wonder what'll happen if we want to use jit evaluator in constant folding without a runtime?

the program will crash :((

ailzhang:
@AmesingFlank Does it mean aot compilation might potentially crash if jit evaluation happens? hmmm that's not perfect

AmesingFlank:
> @AmesingFlank Does it mean aot compilation might potentially crash if jit evaluation happens? hmmm that's not perfect

I agree, but that that had always been the case.
* Before this PR:
     With runtime: Everything works
     Without runtime: Will crash, with or without constant folding.
* After this PR:
     With runtime: Everything works
     Without runtime: Will work if constant folding disabled. Won't work if it is enabled.

So this PR doesn't break anything new. It should be a separate task (cpu-side jit evaluator) to make constant folding work without a runtime.

taichi/backends/vulkan/aot_module_builder_impl.cpp
```diff
@@ -9,10 +9,53 @@ namespace taichi {
 namespace lang {
 namespace vulkan {
 
+// Only responsible for reporting device capabilities
+class AotTargetDevice : public Device {
```
bobcao3, line 13: Very nice design! Maybe put this in a higher level so that GL aot can benefit from this

bobcao3 (approved):
LGTM

ailzhang (approved):
Oh I see, so you've only removed materialize_runtime in the example. This is all good. Thanks!

PR #3712, opened 2021-12-06
[misc] Use ccache & prefer Ninja over Make as cmake generator.

ailzhang:
This PR makes it more automatic using ccache + ninja. 
Ninja is very fast (i.e., instant) for incremental builds so it fits development need where we need to rebuild for small changes. Here we use it only when it's available to avoid breaking people's workflow.


frostming:
Nice, it seems sccache has better support for Ninja on Windows.

lin-hitonami (approved):
LGTM!

PR #3716, opened 2021-12-06
[bug] Fix template arguments of ti.func with default values

lin-hitonami:
Related issue = #
Injecting template arguments into globals is only necessary in kernels, and it should not happen on functions.


ailzhang (approved):
LGTM!

strongoier (commented):
Should we add a test for this?

qiao-bo (approved):
Thanks a lot for the quick fix!

strongoier (approved):
LGTM!

PR #3718, opened 2021-12-06
[misc] Error handle and TLS

Leonz5288:
Related issue = #

### DIFF
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -533,8 +533,8 @@ def init(arch=None,
         else:
             check_version()
             with open(timestamp_path, 'w') as f:
-                f.write(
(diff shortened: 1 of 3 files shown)


python/taichi/lang/__init__.py
```diff
@@ -473,7 +472,7 @@ def check_version():
     try:
         payload = json.dumps(payload)
         payload = payload.encode()
-        req = request.Request('http://metadata.taichi.graphics/check_version',
+        req = request.Request('https://metadata.taichi.graphics/check_version',
```
ailzhang, line 475: In the future we might want to change this without changing source code ;)

ailzhang (approved):
LGTM!

python/taichi/lang/__init__.py
Leonz5288: Good point. Fetching from vault might be a solution in the future.

Issue #3580, opened 2021-11-22
Test examples in CI


ailzhang:
Currently, we only run `ti example laplace` in CI, which cannot prevent us from breaking other examples in PRs. This leads to occasionally unstable releases of Taichi.

The main blocker is that most of the examples require GUI, and we need to add a CI mode for them to only verify correctness.

strongoier:
I've found a way to test examples in CI without modifying the original examples too much. See #3721 as an example. 

Basically, we can:
- Wrap the original main procedure of an example (which often involves GUI) into `main()`;
- Write a test which imports the example and runs correctness checks.

However, it is not easy to verify correctness of most examples. For them, as a first step, I think we can run the calculation of a couple of frames by providing some random info (e.g. positions of the mouse) without involving GUI. This can at least ensure that the examples can successfully run, i.e., no compilation or runtime errors will occur.

If we want to fully make sure the results of the examples are also correct, for now I think it is inevitable to manually run them to check the results before a release. That said, adding the tests described above can still help reduce the chances of breaking the examples, which will certainly ease the burden before a release.

After skimming through all the examples, I think we can start with the following:
- *examples/algorithm/*
    - [x] laplace
    - [ ] marching_squares
    - [ ] mgpcg
    - [ ] mgpcg_advanced
    - [ ] print_offset
- *examples/autodiff/*
    - [ ] minimization
- *examples/features/sparse/*
    - [ ] explicit_activation
    - [ ] taichi_bitmasked
    - [ ] taichi_dynamic
    - [ ] taichi_sparse
    - [ ] tutorial
- *examples/rendering/*
    - [ ] cornell_box
    - [ ] rasterizer
    - [ ] sdf_renderer
    - [ ] simple_uv
    - [ ] taichi_logo
- *examples/simulation/*
    - [ ] ad_gravity
    - [ ] comet
    - [ ] fem99
    - [ ] fractal
    - [ ] game_of_life
    - [ ] initial_value_problem
    - [ ] mass_spring_game
    - [ ] mpm99
    - [ ] mpm_lagrangian_forces
    - [ ] nbody
    - [ ] odop_solar
    - [ ] pbf2d
    - [ ] physarum
    - [ ] vortex_rings
    - [ ] waterwave

Contributions are highly welcome!

ReFantasy:
I would like to help "fractal"

Zydiii:
I would like to help "minimization"

nasnoisaac:
[...]

PR #3721, opened 2021-12-06
[test] Refine the way to test the laplace example

strongoier:
Related issue = #3580

Just as #3580 states, we currently only run `ti example laplace` in CI because many examples require GUI. In this PR, instead of adding a CI mode to the examples, I would like to write tests for them, so that users will see examples (mostly) unchanged and developers can have examples tested at the same time. Specifically, the modification includes:
- Wrap the original main procedure of the example into `main()`;
- Write a test which imports the example and runs non-GUI operations/correctness checks.

This PR serves as a kickoff and only covers the `laplace` example.


tests/python/examples/algorithm/test_laplace.py
```diff
@@ -0,0 +1,11 @@
+from taichi.examples.algorithm.laplace import *
```
ailzhang, line 1: shall we just import laplace here?

ailzhang (approved):
LGTM! Thanks!

tests/python/examples/algorithm/test_laplace.py
strongoier: We can `from taichi.examples.algorithm import laplace`, but then we have to add `laplace.` prefix to all variables and functions. Since the test is exclusively designed for the corresponding example, is it a bit tedious to add that?

ailzhang: @strongoier  oh I meant we can do `from taichi.examples.algorithm.laplace import laplace` ;)

strongoier: Ah yes. Updated.

Issue #2734, opened 2021-08-20
Support setting the random seed during initialization


k-ye:
**Concisely describe the proposed feature**

Something like `ti.init(random_seed=42)`. This would be very useful during the development. We can start with the LLVM backends.

siddharth1704:
can you explain more about it cause i am new to this i feel i will be able to do it

knight42:
@k-ye Hi! It seems that we could already specify the random seed now https://github.com/taichi-dev/taichi/blob/17adcdb92db10b1573c692168ed5a2879f1632ca/taichi/python/export_lang.cpp#L168

Is this issue still valid? Or are there any remaining work to do?

k-ye:
> @k-ye Hi! It seems that we could already specify the random seed now
> 
> https://github.com/taichi-dev/taichi/blob/17adcdb92db10b1573c692168ed5a2879f1632ca/taichi/python/export_lang.cpp#L168
> 
> Is this issue still valid? Or are there any remaining work to do?

Hey! Sorry for the delayed reply. Could you check if this is passed to the backend? (IIRC, Metal at least doesn't use this yet. Likely the case for Vulkan)

knight42:
@k-ye Hi! As for the Vulkan and OpenGL backend, looks like they initialize the random numbers using a similar ad-hoc method:
* opengl: https://github.com/taichi-dev/taichi/blob/master/taichi/backends/opengl/shaders/random.glsl.h#L19
* vulkan: https://github.com/taichi-dev/taichi/blob/master/taichi/codegen/spirv/spirv_ir_builder.cpp#L936

I am not sure how should we adjust the intialisation to accept a random seed, could you provide me some guidance? @k-ye 

In addition, do we need to pass random seed to other backends(like `cc` and `wasm`)? If so, maybe I could change them first.

PR #3724, opened 2021-12-06
[Metal] Pass random seed to metal backend

knight42:
Signed-off-by: knight42 <anonymousknight96@gmail.com>

Related issue = #2734

/cc @k-ye 

I am not sure if the current implementation is desired since I don't know much about cpp 😢 , any advice is welcome!

### DIFF
--- a/tests/python/test_random.py
+++ b/tests/python/test_random.py
@@ -124,7 +124,7 @@ def test_random_seed_per_program():
     assert not np.allclose(result[0], result[1])
 
 
-@ti.test(arch=[ti.cpu, ti.cuda, ti.metal])
+@ti.test(arch=[ti.cpu, ti.cuda])
 def test_random_f64():
     '''
     Tests the granularity of float64 random numbers.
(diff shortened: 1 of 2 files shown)


tests/python/test_random.py
```diff
@@ -2,11 +2,7 @@
 from taichi import approx
 
 
-def archs_support_random(func):
```
knight42, line 5: I have grep-ed this function in the project and it seems this function is not used anywhere, so I delete it in this PR. Please feel free to correct me if I am missing something.

tests/python/test_random.py
```diff
@@ -128,7 +124,7 @@ def gen():
     assert not np.allclose(result[0], result[1])
 
 
-@ti.test(arch=[ti.cpu, ti.cuda])
+@ti.test(arch=[ti.cpu, ti.cuda, ti.metal])
```
k-ye, line 127: Our CI's Metal doesn't support 64-bit data yet

```suggestion
@ti.test(arch=[ti.cpu, ti.cuda])
```

knight42: Got it!

knight42:
@k-ye Hi! All tests passed now! Would you like to take a second look?

k-ye (approved):
LGTM Thanks!

PR #3728, opened 2021-12-07
[refactor] Separate Cpp examples into different files

AmesingFlank:
This is a followup to #3705. The three examples are separated into different files.


k-ye (approved):
LGTM!

PR #3731, opened 2021-12-07
[refactor] Move ti.lib to ti._lib and move ti.core to ti._lib.core

lin-hitonami:
Related issue = #


lin-hitonami:
Since the vulkan support for M1 without installing molten-vk or vulkan sdk seems broken even in the master branch, I cannot test whether my change broke it or not.

strongoier (changes requested):
There are some conflicts to be handled as #3774 is in :-)

ailzhang (approved):
LGTM! Thanks!

PR #3733, opened 2021-12-07
[Lang] Support sparse solver datatype configuration

FantasyVR:
Related issue = #2906

### DIFF
--- a/taichi/program/sparse_solver.cpp
+++ b/taichi/program/sparse_solver.cpp
@@ -30,7 +30,7 @@ namespace lang {
 
 template <class EigenSolver>
 bool EigenSparseSolver<EigenSolver>::compute(const SparseMatrix &sm) {
-  solver_.compute(sm.get_matrix().cast<float>());
+  solver_.compute(sm.get_matrix());
   if (solver_.info() != Eigen::Success) {
     return false;
   } else
--- a/taichi/program/sparse_solver.cpp
+++ b/taichi/program/sparse_solver.cpp
@@ -15,7 +15,7 @@
 
 using Triplets = std::tuple<std::string, std::string, std::string>;
 namespace {
-struct key_hash : public std::unary_function<Triplets, std::size_t> {
+struct key_hash {
   std::size_t operator()(const Triplets &k) const {
     auto h1 = std::hash<std::string>{}(std::get<0>(k));
     auto h2 = std::hash<std::string>{}(std::get<1>(k));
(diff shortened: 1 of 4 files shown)


taichi/program/sparse_solver.cpp
```diff
@@ -55,20 +57,23 @@ bool EigenSparseSolver<EigenSolver>::info() {
...
-          MAKE_SOLVER(LLT, COLAMD),
-          MAKE_SOLVER(LDLT, AMD),
-          MAKE_SOLVER(LDLT, COLAMD),
-      };
-  std::pair<std::string, std::string> solver_key =
-      std::make_pair(solver_type, ordering);
+          MAKE_SOLVER(float32, LLT, AMD), MAKE_SOLVER(float32, LLT, COLAMD),
+          MAKE_SOLVER(float32, LDLT, AMD), MAKE_SOLVER(float32, LDLT, COLAMD)};
+  static const std::unordered_map<std::string, std::string> dt_map = {
+      {"f32", "float32"}, {"f64", "float64"}};
```
Hanke98, line 70: Is float64 type supported now? Should we add float64 in `solver_factory`? (or maybe later after float64 is fully supported.

Hanke98 (commented):
Sorry for the delayed response.
LGTM mostly, I only have one point that I am not so sure

BTW, should we add some test cases against data type?

strongoier (changes requested):
Would you fix the compilation error on the Windows platform?

taichi/program/sparse_solver.cpp
```diff
@@ -55,20 +57,23 @@ bool EigenSparseSolver<EigenSolver>::info() {
...
-          MAKE_SOLVER(LLT, AMD),
-          MAKE_SOLVER(LLT, COLAMD),
-          MAKE_SOLVER(LDLT, AMD),
-          MAKE_SOLVER(LDLT, COLAMD),
-      };
-  std::pair<std::string, std::string> solver_key =
-      std::make_pair(solver_type, ordering);
+          MAKE_SOLVER(float32, LLT, AMD), MAKE_SOLVER(float32, LLT, COLAMD),
+          MAKE_SOLVER(float32, LDLT, AMD), MAKE_SOLVER(float32, LDLT, COLAMD)};
+  static const std::unordered_map<std::string, std::string> dt_map = {
```
strongoier, line 69: Using strings to refer to types (e.g. "f32") is not good practice, so maybe in the future some refactoring needs to be done here :-)

strongoier (approved):
LGTM! Only one suggestion below.

PR #3735, opened 2021-12-07
[ci] Use clang+Ninja to build Taichi on windows

bobcao3:
Related issue = #


bobcao3:
Cache is all miss, not sure why. The build works

frostming:
> Cache is all miss, not sure why. The build works

Maybe need to add `-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache`

bobcao3:
> > Cache is all miss, not sure why. The build works
> 
> Maybe need to add `-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache`

No need. ccache is called, it's just that ccache thinks it's all miss. (in the cmakelist it detects the ccache program and automatically set the launchers)

setup.py
```diff
@@ -187,7 +187,7 @@ def prepare_package(self):
                     shutil.copy(moltenvk_path,
                                 os.path.join(target, 'libMoltenVK.dylib'))
             else:
-                shutil.copy('runtimes/Release/taichi_core.dll',
+                shutil.copy('runtimes/taichi_core.dll',
```
ailzhang, line 190: OOC do we need `Release` here (seems like we explicitly opened a Release folder above for ninja as well) ?

lin-hitonami:
Do all the compilations happen in `C:/Users/runneradmin/AppData/Local/Temp/pip-req-build-n_d7za0z/` ? This seems to be a randomly generated directory and it needs some configuration to enable sharing caches when compiling in different directories. See https://ccache.dev/manual/4.5.1.html#_compiling_in_different_directories . (btw sccache does not support that at all)

frostming:
> Do all the compilations happen in `C:/Users/runneradmin/AppData/Local/Temp/pip-req-build-n_d7za0z/` ? This seems to be a randomly generated directory and it needs some configuration to enable sharing caches when compiling in different directories. See https://ccache.dev/manual/4.5.1.html#_compiling_in_different_directories . (btw sccache does not support that at all)

Yes, because it is built by `pip`, which sets up a temp dir for build. Change it to `python setup.py install` to check if cache hits

setup.py
bobcao3: The release folder is generated by msbuild. I can't think of a good way to detect whether msbuild or ninja is used. Maybe just enforce ninja?

ailzhang: IIUC this line already implicitly enforced ninja? ;) If so we'll need to be extra careful and make sure the dev install is updated correctly. ;)

bobcao3: Yes... kind of sad, but I can't think of a way to cleanly deal with this.

ailzhang:
@frostming 
>  but I have to mention invoking setup.py directly is not a good practice. 

Is it better to use `pip install .` instead ? If so feel free to send PRs to update our CI & doc sites :D!

.github/workflows/scripts/win_build.ps1
```diff
@@ -63,12 +76,13 @@ python -m pip install wheel
 python -m pip install -r requirements_dev.txt
 python -m pip install -r requirements_test.txt
 WriteInfo("Building Taichi")
-$env:CXX = "$libsDir\taichi_clang\bin\clang++.exe"
+$env:TAICHI_CMAKE_ARGS += " -DCLANG_EXECUTABLE=$libsDir\\taichi_clang\\bin\\clang++.exe"
+$env:TAICHI_CMAKE_ARGS += " -DLLVM_AS_EXECUTABLE=$libsDir\\taichi_llvm\\bin\\llvm-as.exe"
 if ($install) {
     if ($develop) {
-        python -m pip install -v -e .
+        python setup.py develop
```
ailzhang, line 83: These commands are kinda equivalent (and IIUC pip install version is preferred :P so let's keep this unchanged?

setup.py
ailzhang: @bobcao3 I don't mind switching to ninja for now ;) but this PR better stays consistent ;) . For example, `if (CMAKE_GENERATOR EQUAL "Ninja")` above kinda hints that we still support msbuild :D 
If we are sure ninja is easy to install on windows and we know how to update dev install doc, we could just error out on msbuild in cmake?

ailzhang (commented):
Nice work, thanks!!

.github/workflows/scripts/win_build.ps1
bobcao3: In the pip version caching won't work because it uses a different path every time and the generators use absolute path everywhere

bobcao3:
I tried CMAKE_GENERATOR but it doesn't really work, it did not detect Ninja at all so... not sure

frostming:
The changes made here LGTM and I hope this will be landed in the master soon as it dramatically accelerates the windows build 💯 

We can figure out how to use the preferred build approach `pip install` later in another PR.

ailzhang:
@bobcao3 Would you mind resolving the merge conflict so that we can get it in? Thanks a lot!

docs/lang/articles/contribution/dev_install.md
```diff
@@ -406,18 +435,37 @@ The `develop` command serves the developers' needs better because edits to the P
...
-   ```shell
-   .\install_deps.ps1
-   ```
-3. Reboot your computer.
-4. Run powershell as administrator and run [this script](https://github.com/taichi-dev/taichi_assets/releases/download/windows_scripts/build_taichi.ps1) to build Taichi from souce:
 
-   ```shell
-   .\build_taichi.ps1
-   ```
+```shell
```
k-ye, line 443: Let's add a three-space indentation to this code block

docs/lang/articles/contribution/dev_install.md
```diff
@@ -406,18 +435,37 @@ The `develop` command serves the developers' needs better because edits to the P
...
+git clone --recursive https://github.com/taichi-dev/taichi
+
+cd taichi
+
+python -m pip install --user -r requirements_dev.txt
+```
+
+4. Build taichi by using `python setup.py develop`
+
+:::note
```
ailzhang, line 453: Did we just add this part (or move from other places)? I somehow couldn't find where it was moved from hmmm

bobcao3: Previously we asked the user to run `build_taichi.ps1`. Now we are asking them to build it through command line, the build script is for CI, not really for end user

ailzhang (approved):
LGTM！

PR #3737, opened 2021-12-07
[refactor] Add namespaces for cpp_examples.

ailzhang:
C++ build fails in my linux machine complaining about multiple
definition of `main`. Not sure why this only happened with `ninja`
but not `make`. But it's good to add namespaces anyway. :D

Related issue = #

### DIFF
--- a/cpp_examples/main.cpp
+++ b/cpp_examples/main.cpp
@@ -2,6 +2,8 @@
 #include "taichi/ir/statements.h"
 #include "taichi/program/program.h"
 
+namespace taichi {
+namespace example {
 void run_snode();
 void autograd();
 void aot_save();
@@ -12,3 +14,5 @@ int main() {
   aot_save();
   return 0;
 }
+}  // namespace example
+}  // namespace taichi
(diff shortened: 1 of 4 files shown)


AmesingFlank:
I've seen this error before, it was because in I was using `examples/*.cpp` in `TaichiExamples.cmake`, and somehow it included the dummy `.cpp` file that cmake implicit generates for checking c++ compiler. Although I've already explicitly listed out the files in cmake, so I don't know why it's happening now.

cpp_examples/main.cpp
```diff
@@ -2,6 +2,8 @@
 #include "taichi/ir/statements.h"
 #include "taichi/program/program.h"
 
+namespace taichi {
```
AmesingFlank, line 5: if we surround `main()` with `nampespace xx{}`, then who's the program entry function now?

ailzhang:
nvm I'll close this PR for now.... something weird happened in my linux machine :P

ailzhang:
@AmesingFlank I don't know what happened exactly but `git clean -xfd` fixed it.. Although usually `python setup.py clean` was good enough... :P

PR #3738, opened 2021-12-08
[Doc] Fix a heading level in syntax.md

ran-huang:
Related issue = #

`Functions` should be a level 2 heading just like `Kernels`.


ailzhang (approved):
Thanks!!

PR #3742, opened 2021-12-08
[refactor] Move ti.randn out of taichi.lang.

ailzhang:
`ti.randn` is more like a standard lib on top of taichi.lang. Leaving it inside `taichi.lang` can easily cause circular import and import order related issues. Let's move such `ti.func`s out. Potentially we can have `ti.std` in the future but I'll put them in `taichi/_funcs.py` and expose them to `ti` namespace manually for now.


strongoier (approved):
LGTM!

PR #3753, opened 2021-12-08
[Perf] Unnecessary assignment as it is redefined

### LINKED ISSUE #3752: Unnecessary assignment as it is redefined
Hi, guys

We found that some assignments are unnecessary as they are redefined.

For example, in /python/taichi/lang/impl.py line 693.

`elif isinstance(_var, (list, tuple)):
                res = _var# the first assignment
                # If the first element is '__ti_format__', this list is the result of ti_format.
                if len(_var) > 0 and isinstance(
                        _var[0], str) and _var[0] == '__ti_format__':
                    res = _var[1:] # the 2nd assignment
                else:
                    res = list_ti_repr(_var) # the 3nd assignment ` 

here the 1st assignment is not necessary.

Even though this will not cause CVEs, we think that it will still influence the performance.

We create a PR for this one.

Thanks

tszdanger:
Related issue = #

Related issue = close #3752


ailzhang (approved):
Nice catch! Thanks!

PR #3755, opened 2021-12-08
[Android] Add initial support of the platform

ghuau-innopeak:
Related issue = #3679

This is an initial support so a lot of features are not supported yet such as the logging with logcat or traceback.

### DIFF
--- a/taichi/backends/vulkan/vulkan_program.cpp
+++ b/taichi/backends/vulkan/vulkan_program.cpp
@@ -1,7 +1,10 @@
 #include "taichi/backends/vulkan/vulkan_program.h"
 #include "taichi/backends/vulkan/aot_module_builder_impl.h"
 
+#ifdef ANDROID
+#else
 #include "GLFW/glfw3.h"
+#endif
 
 using namespace taichi::lang::vulkan;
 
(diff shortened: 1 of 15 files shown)


ghuau-innopeak:
I added a new tag [Android] as this is a new platform, but I can rename it to [Linux] if needed. Also, for testing, this is what I'm currently using:
```
TAICHI_CMAKE_ARGS="-DCMAKE_TOOLCHAIN_FILE=$ANDROID_HOME/ndk-bundle/build/cmake/android.toolchain.cmake -DANDROID_NATIVE_API_LEVEL=29 -DANDROID_ABI=arm64-v8a -DTI_WITH_VULKAN:BOOL=ON -DTI_WITH_CUDA:BOOL=OFF -DTI_WITH_OPENGL:BOOL=OFF -DTI_WITH_LLVM:BOOL=OFF -DTI_EXPORT_CORE:BOOL=ON" python3 setup.py build_ext
```

taichi/backends/vulkan/vulkan_program.cpp
```diff
@@ -68,6 +85,9 @@ void VulkanProgramImpl::materialize_runtime(MemoryPool *memory_pool,
   *result_buffer_ptr = (uint64 *)memory_pool->allocate(
       sizeof(uint64) * taichi_result_buffer_entries, 8);
 
+// TODO: Support for Android if possible... Should we move this code to be the
+// responsability of the caller to this library?
```
bobcao3, line 89: When embedded in other applications the design is to leave the creation of VkDevice and related states to the caller/host. VulkanDeviceCreator is only used when Taichi is running on its own.

bobcao3 (approved):
Looks great!

taichi/backends/vulkan/vulkan_program.cpp
bobcao3: I think it's good for now!

PR #3756, opened 2021-12-09
[misc] Speed-up builds by removing LLVM includes from llvm_program.h

bobcao3:
This also reduces the warnings coming out from the LLVM headers, especially on windows, thus improving caching hit rates.


ailzhang (approved):
Nice!!

PR #3757, opened 2021-12-09
[Lang] Fix pylint rule W1309 (f-string-without-interpolation)

kxxt:
Related issue = #3151


ailzhang (approved):
LGTM! Thanks for fixing it!

PR #3759, opened 2021-12-09
[test] Enable ndarray tests

qiao-bo:
Let's try to enable these tests, I can no longer reproduce the flaky behavior locally.


ailzhang (approved):
LGTM!

PR #3761, opened 2021-12-09
[Lang] Enforce members of a matrix field to have same shape

strongoier:
Related issue = #2590

After this PR, the following code snippet:
```py
import taichi as ti

ti.init()
a = ti.Matrix.field(2, 2, ti.i32)
ti.root.dense(ti.i, 10).place(a.get_scalar_field(0, 0))
ti.root.dense(ti.i, 11).place(a.get_scalar_field(0, 1))
ti.root.dense(ti.i, 10).place(a.get_scalar_field(1, 0))
ti.root.dense(ti.i, 11).place(a.get_scalar_field(1, 1))

@ti.kernel
def func():
    pass

func()
```
will give the following error:
```
RuntimeError: Members of the following field have different shapes [(10,), (11,), (10,), (11,)]:
  File "test_matrix_field_member_shape.py", line 4, in <module>
    a = ti.Matrix.field(2, 2, ti.i32)

```

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -354,14 +354,14 @@ class PyTaichi:
         _root_fb._finalize_for_aot()
 
     @staticmethod
-    def _get_tb_of_global_var(_var):
+    def _get_tb(_var):
         return getattr(_var, 'declaration_tb', str(_var.ptr))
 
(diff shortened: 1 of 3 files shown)


lin-hitonami (approved):
LGTM!

PR #3766, opened 2021-12-09
[Vulkan] Update AOT Loader support to new API

ghuau-innopeak:
Related issue = #3642

### DIFF
--- a/taichi/backends/vulkan/aot_module_builder_impl.cpp
+++ b/taichi/backends/vulkan/aot_module_builder_impl.cpp
@@ -13,6 +13,9 @@ AotModuleBuilderImpl::AotModuleBuilderImpl(
     const std::vector<CompiledSNodeStructs> &compiled_structs)
     : compiled_structs_(compiled_structs) {
   aot_target_device_ = std::make_unique<AotTargetDevice>(Arch::vulkan);
+  if (!compiled_structs.empty()) {
+    ti_aot_data_.root_buffer_size = compiled_structs[0].root_size;
+  }
 }
 
 uint32_t AotModuleBuilderImpl::to_vk_dtype_enum(DataType dt) {
@@ -51,20 +54,6 @@ void AotModuleBuilderImpl::write_spv_file(
   fs.close();
 }
 
-std::vector<uint32_t> AotModuleBuilderImpl::read_spv_file(
-    const std::string &output_dir,
-    const TaskAttributes &k) {
-  const std::string spv_path = fmt::format("{}/{}.spv", output_dir, k.name);
-  std::vector<uint32_t> source_code;
-  std::ifstream fs(spv_path, std::ios_base::binary | std::ios::ate);
-  size_t size = fs.tellg();
-  fs.seekg(0, std::ios::beg);
-  source_code.resize(size / sizeof(uint32_t));
-  fs.read((char *)source_code.data(), size);
-  fs.close();
-  return source_code;
-}
-
 void AotModuleBuilderImpl::dump(const std::string &output_dir,
                                 const std::string &filename) const {
   TI_WARN_IF(!filename.empty(),
@@ -84,23 +73,6 @@ void AotModuleBuilderImpl::dump(const std::string &output_dir,
   TextSerializer ts;
   ts.serialize_to_json("aot_data", ti_aot_data_);
   ts.write_to_file(txt_path);
-}
-
-void AotModuleBuilderImpl::load(const std::string &output_dir) {
-  const std::string bin_path = fmt::format("{}/metadata.tcb", output_dir);
-  read_from_binary_file(ti_aot_data_, bin_path);
-  for (int i = 0; i < ti_aot_data_.kernels.size(); ++i) {
-    auto k = ti_aot_data_.kernels[i];
-    std::vector<std::vector<uint32_t>> spirv_sources_codes;
-    for (int j = 0; j < k.tasks_attribs.size(); ++j) {
-      std::vector<uint32_t> res = read_spv_file(output_dir, k.tasks_attribs[j]);
-      spirv_sources_codes.push_back(res);
-    }
-
-    VkRuntime::RegisterParams params;
-    params.kernel_attribs = ti_aot_data_.kernels[i];
-    params.task_spirv_source_codes = spirv_sources_codes;
-  }
 }
 
 void AotModuleBuilderImpl::add_per_backend(const std::string &identifier,
--- a/taichi/backends/metal/aot_module_builder_impl.h
+++ b/taichi/backends/metal/aot_module_builder_impl.h
@@ -6,7 +6,7 @@
 
 #include "taichi/backends/metal/aot_utils.h"
 #include "taichi/backends/metal/struct_metal.h"
-#include "taichi/program/aot_module_builder.h"
+#include "taichi/program/aot_module.h"
 
 namespace taichi {
 namespace lang {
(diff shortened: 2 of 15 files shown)


ghuau-innopeak:
@ailzhang Before writing the AOT Test cases, I think we will need to re-add support for loading with this MR, so for example, we can save a module, re-load it and make sure all the kernel have been re-loaded properly with proper name/content/root_size/etc.

There is a change I'm not necessary a huge fan but I'm not sure if there is a better way to do it:
- Creating accessors to get the kernels/fields in Vulkan Program: Had to do it like this now because the AotModuleBuilderImpl is not exposed to host/caller of the library

ailzhang (commented):
@ghuau-innopeak  Nice work! 
Btw when I think about this I wonder if `VulkanProgramImpl::load(std::unique_ptr<AotModuleBuidler> aot_module)` fits better? In that case you don't need to carry a AotModuleBuilder in program but can optionally load from it. Also in side this function you'll have access to a vulkan_runtime to register the kernels. Wdyt? 
cc: @k-ye

ghuau-innopeak:
> @ghuau-innopeak Nice work! Btw when I think about this I wonder if `VulkanProgramImpl::load(std::unique_ptr<AotModuleBuidler> aot_module)` fits better? In that case you don't need to carry a AotModuleBuilder in program but can optionally load from it. Also in side this function you'll have access to a vulkan_runtime to register the kernels. Wdyt? cc: @k-ye

@ailzhang The problem is I need the same loader for the accessors `get_kernel()`, `get_field`, etc.
And in the case of using Taichi in external CPP app, the `vulkan_runtime_` inside Program is not used, we actually create an instance from the app. I should probably try to push some tests code on a repo and share it to have a better idea but it's still a WIP :) and I'm trying to stay as close as possible to  #3642, right now, it looks like:

```
    VulkanProgramImpl program(config, "./mpm88");

    ...

    // Create Runtime
    VkRuntime::Params params;
    params.host_result_buffer = result_buffer;
    params.device = embedded_device->device();
    auto vulkan_runtime = std::make_unique<VkRuntime>(std::move(params));

    // Retrieve the Kernels
    VkRuntime::RegisterParams init_kernel, substep_kernel;
    program.get_kernel("init", init_kernel);
    program.get_kernel("substep", substep_kernel);

    // Register the kernels to the VkRuntime
    vulkan_runtime->add_root_buffer(root_size);
    auto init_kernel_handle = vulkan_runtime->register_taichi_kernel(init_kernel);
    auto substep_kernel_handle = vulkan_runtime->register_taichi_kernel(substep_kernel);

    // Execute an AOT kernel
    vulkan_runtime->launch_kernel(init_kernel_handle, &host_ctx);
```

taichi/backends/vulkan/aot_module_builder_impl.cpp
```diff
@@ -96,11 +99,33 @@ void AotModuleBuilderImpl::load(const std::string &output_dir) {
...
+  }
+}
 
-    VkRuntime::RegisterParams params;
-    params.kernel_attribs = ti_aot_data_.kernels[i];
-    params.task_spirv_source_codes = spirv_sources_codes;
+bool AotModuleBuilderImpl::get_kernel(const std::string &name,
+                                      VkRuntime::RegisterParams &kernel) {
+  for (int i = 0; i < ti_aot_data_.kernels.size(); ++i) {
+    // Kernel names encode more than the name of the function, but for AOT, only
```
ailzhang, line 109: IIUC this is  more like an offloaded task than a kernel (a kernel (which is ti.kernel in python) consists of many offloaded tasks ;)

ailzhang:
> I should probably try to push some tests code on a repo and share it to have a better idea but it's still a WIP :) and I'm trying to stay as close as possible to **#3642**

@ghuau-innopeak Yea it's a great idea to have a test that works e2e so that we can work refining on the APIs together :D I guess the first prototype doesn't have to be perfect as long as it can serialize and deserialize from AotModuleBuilderImpl?  

I'm thinking that `GLProgram` in the RFC doesn't have to be exactly `VulkanProgramImpl` so we can adjust it to fit our needs :D

ghuau-innopeak:
@ailzhang https://github.com/ghuau-innopeak/taichi-aot-tests I pushed our test code for now for Android/Desktop and some instructions on how to build :), desktop/main.cpp should give you an idea on how we register the kernels/fields/etc. If we go the way with `VulkanProgramImpl::load(std::unique_ptr<AotModuleBuidler> aot_module)`, how would you register the kernel to the Vulkan Runtime? As this is not the one used inside the Program.

I'm completely open to any idea for initial integration :)

ailzhang:
Hey @ghuau-innopeak! Thanks a ton for sharing the e2e test code! It's super helpful in understanding the design! 

After seeing the e2d test I feel like my point about `VulkanProgramImpl::load` wasn't accurate so let me clarify a few things, @ghuau-innopeak @k-ye  please correct me if I'm wrong!

- If we look at [AotModuleBuilderImpl::load()](https://github.com/taichi-dev/taichi/blob/master/taichi/backends/vulkan/aot_module_builder_impl.cpp#L90-L103) it actually fits better in a new class ` AotModuleLoaderImpl` which only carries a `TaichiAotData ti_aot_data_` and provides a `get_kernel()` interface? IMHO separating aot module builder and loader will make it easier to manage. Wdyt?
- With that we also don't need instantiate `VulkanProgramImpl` in [desktop/main.cpp](https://github.com/ghuau-innopeak/taichi-aot-tests/blob/main/desktop/main.cpp#L41) anymore.  Looks like loading with a `AotModuleLoader` instance  then register the kernels to vulkan_runtime is good enough? This is roughly what I have in mind: 
```
// pseudo code
class AotModuleLoader {
  bool get_kernel(const std::string &name, VkRuntime::RegisterParams &kernel);

  bool get_field(const std::string &name, aot::CompiledFieldData &field);

  size_t get_root_size() const;
private:
TaichiAotData ti_aot_data_;
};

// desktop/main.cpp
taichi::lang::AotModuleLoader loader("../mpm88");
taichi::lang::vulkan::VkRuntime::RegisterParams init_kernel, substep_kernel;
bool ret = loader.get_kernel("init", init_kernel);
if (!ret) {
    printf("Cannot find 'init' kernel\n");
    return -1;
}
ret = loader.get_kernel("substep", substep_kernel);
if (!ret) {
    printf("Cannot find 'substep' kernel\n");
    return -1;
}
auto root_size = loader.get_root_size();
printf("root buffer size=%d\n", root_size);

```
- I believe the APIs described in https://github.com/taichi-dev/taichi/issues/3642 is more like when we write a brand new runtime & data structures for AOT. IMHO `GLProgram` there is supposed to be a brand new class and is not related to `GLProgramImpl`, IIUC we don't have to put `get_kernel` into `VulkanProgramImpl` ;) . 


Please feel free to correct me if I'm wrong! Thanks!!

ghuau-innopeak:
@ailzhang it makes completely sense, I actually do prefer your approach :), and it completely remove the usage of VulkanProgram as you said. It also create a separation between the Builder and Loader (not sure why I tried to keep everything together here at first...).

I pushed some updates to reflect this new approach and updated the test case too. Right now, there is only 2 main things that was changed:
- Renamed `aot_module_builder.cpp/h` to `aot_module.cpp/h` as it contains definition used by both the Builder and Loader (for example the `CompiledFieldData`.
-  Created a new AotModuleLoaderImpl for Vulkan that can be used directly in our external C++ code

ailzhang (approved):
Awesome!! Thanks a ton!

PR #3772, opened 2021-12-11
[vulkan] Fix SPIRV types

AmesingFlank:
We always use integral types for buffer type. This leads to SPIRV with invalid type errors.
For example, looking at the AOTed SPIRV of mpm 99, one can find things like:
```
        %575 = OpAccessChain %_ptr_StorageBuffer_int %root_buffer_0 %int_0 %574
     %tmp402 = OpAtomicFAddEXT %float %575 %int_1 %int_0 %tmp400
```
This is not valid, because we're treating a int pointer as a float pointer. Tools like `spirv-val` complains about this.

btw, I don't understand why `get_primitive_buffer_type` was introduced in the first place hmm.


bobcao3 (approved):
LGTM! I thought I have fixed this qutie a while ago when switching to linear indexing, apparently didn't

PR #3773, opened 2021-12-11
[Doc] Fix a misspelling

0xrabbyte:
PS: The examples here are not enough to make people understand.
Related issue = #


yuanming-hu (approved):
Thanks! 

> PS: The examples here are not enough to make people understand.

Agreed. It seems that the GUI section here needs some systematic improvements.

PR #2870, opened 2021-09-02
[dx] [WIP] DirectX backend

quadpixels:
This is a work-in-progress DirectX backend.

What works now:
  * A simplified mpm88 (points are aligned to a regular grid instead of randomly generated)

Todo items:
  * Add random number generators
  * Look into performance problems
  * Make other examples work
  * Use SPIR-V

Related issue = #992

PR #3776, opened 2021-12-12
[DX11] DX11 backend based on spirv_cross

quadpixels:
Related issue = #992

This is a rework of Pull Request #2870; while PR #2870 used the source-to-source method to generate an HLSL shader, this one uses spirv_cross similarly to generate an HLSL shader much similarly to the Vulkan backend from @bobcao3.
Because of the usage of spirv_cross, the functionalities should be on parity with the Vulkan backend (it supports what the Vulkan backend supports and does not support what the Vulkan backend does not support).


cmake/TaichiCore.cmake
```diff
@@ -287,7 +289,9 @@ endif()
 if (TI_WITH_OPENGL)
     add_subdirectory(external/SPIRV-Cross)
     target_include_directories(${CORE_LIBRARY_NAME} PRIVATE external/SPIRV-Cross)
-    target_link_libraries(${CORE_LIBRARY_NAME} spirv-cross-glsl spirv-cross-core)
+    target_link_libraries(${CORE_LIBRARY_NAME} spirv-cross-glsl)
+    target_link_libraries(${CORE_LIBRARY_NAME} spirv-cross-hlsl)
```
bobcao3, line 293: Should add an separate TI_WITH_DX11 flag somewhere and use that to link hlsl cross compiler. Then we can add the cross compiler core if either SPIR-V based OpenGL or DX is used.

python/taichi/examples/chi_examples/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 3.12)
```
bobcao3, line 1: Should remove this file

python/taichi/examples/chi_examples/README.md
```diff
@@ -0,0 +1,23 @@
+# How to use CHI IR Builder
```
bobcao3, line 1: Should remove this file

python/taichi/examples/chi_examples/main.cpp
```diff
@@ -0,0 +1,332 @@
+#include <bits/stdc++.h>
```
bobcao3, line 1: Should remove this file

taichi/backends/dx/dx_api.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+ID3D11Device *g_device;
+ID3D11DeviceContext *g_context;
+IDXGISwapChain *g_swapchain;
+ID3D11Buffer *g_args_i32_buf, *g_args_f32_buf, *g_data_i32_buf, *g_data_f32_buf,
+    *g_extr_i32_buf, *g_extr_f32_buf, *g_locks_buf, *tmp_arg_buf;
+ID3D11UnorderedAccessView *g_args_i32_uav, *g_args_f32_uav, *g_data_i32_uav,
+    *g_data_f32_uav, *g_extr_i32_uav, *g_extr_f32_uav, *g_locks_uav;
+
+ID3D11Device *GetD3D11Device() {
```
bobcao3, line 19: We should rename these to `snake_case`

taichi/backends/dx/dx_api.cpp
```diff
@@ -0,0 +1,320 @@
...
+    TI_TRACE("D3D11 device has already been created.");
+  }
+  return true;
+}
+
+bool is_dx_api_available() {
+  return initialize_dx();
+}
+
+std::unique_ptr<Device> get_dx_device() {
```
bobcao3, line 312: This should really be called `make_dx_device()`

taichi/backends/dx/dx_device.cpp
```diff
@@ -0,0 +1,464 @@
...
+namespace taichi {
+namespace lang {
+namespace directx11 {
+
+extern IDXGISwapChain *g_swapchain;
+
+static uint32_t g_alloc_id = 0;
+
+// Mapping from binding ID to Buffer
+static std::unordered_map<uint32_t, ID3D11Buffer *> g_binding2buf;
```
bobcao3, line 16: These should be put inside DX11Device (similar to the Vulkan backend), we should refrain from using global variables

taichi/backends/dx/dx_device.cpp
```diff
@@ -0,0 +1,464 @@
...
+                              uint32_t binding,
+                              DevicePtr ptr,
+                              size_t size) {
+  TI_NOT_IMPLEMENTED;
+}
+
+void DxResourceBinder::buffer(uint32_t set,
+                              uint32_t binding,
+                              DeviceAllocation alloc) {
+  rw_buffer(set, binding, alloc);
```
bobcao3, line 57: `Buffer` means constant buffer in DX / uniform buffer in Vulkan. Buffer should not be bind as an RWByteAccessBuffer

taichi/backends/dx/dx_device.cpp
```diff
@@ -0,0 +1,464 @@
...
+  TI_NOT_IMPLEMENTED;
+}
+
+DxPipeline::DxPipeline(const PipelineSourceDesc &desc, const std::string &name)
+    : compute_shader_(nullptr) {
+  TI_ASSERT(desc.type == PipelineSourceType::hlsl_src ||
+            desc.type == PipelineSourceType::spirv_binary);
+
+  if (desc.type == PipelineSourceType::hlsl_src) {
+    TI_NOT_IMPLEMENTED;
```
bobcao3, line 85: Can we still somehow preserve this path?

taichi/backends/dx/dx_device.cpp
```diff
@@ -0,0 +1,464 @@
...
+}
+
+DxPipeline::~DxPipeline() {
+}
+
+ResourceBinder *DxPipeline::resource_binder() {
+  return &binder_;
+}
+
+DxCommandList::~DxCommandList() {
```
bobcao3, line 121: https://docs.microsoft.com/en-us/windows/win32/direct3d11/overviews-direct3d-11-render-multi-thread-command-list In the future we should use DX11's built-in command list instead of emulating it on the CPU. This is a possible improvement, don't need to do it rn. Let's get the back end merged first

taichi/backends/opengl/opengl_device.cpp
```diff
@@ -150,6 +150,7 @@ void GLCommandList::bind_resources(ResourceBinder *_binder) {
     auto cmd = std::make_unique<CmdBindBufferToIndex>();
     cmd->buffer = buffer;
     cmd->index = binding;
+    printf("[gl bind_resources] %u, %u\n", buffer, binding);
```
bobcao3, line 153: Should remove this

taichi/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -978,17 +978,12 @@ void IRBuilder::init_random_function(Value global_tmp_) {
...
-  // how RAND_STATE changes, `gl_GlobalInvocationID.x` can still help
-  // us to set different seeds for different threads.
-  // Discussion:
-  // https://github.com/taichi-dev/taichi/pull/912#discussion_r419021918
-  Value tmp9 = load_var(rand_gtmp_, t_int32_);
-  Value tmp10 = new_value(t_int32_, ValueKind::kNormal);
-  ib_.begin(spv::OpIAdd)
-      .add_seq(t_int32_, tmp10, tmp9, _1)
+
+  // FIXME: Only do this for the first thread or something
```
bobcao3, line 982: FYI here for the other reviewers: this is needed so that the SPIR-V codegen satisfies the race condition detection in HLSL compiler

taichi/program/program.cpp
```diff
@@ -88,6 +90,13 @@ Program::Program(Arch desired_arch)
   } else if (config.arch == Arch::opengl) {
     TI_ASSERT(opengl::initialize_opengl(config.use_gles));
     program_impl_ = std::make_unique<OpenglProgramImpl>(config);
+  } else if (config.arch == Arch::dx11) {
+    if (!directx11::is_dx_api_available()) {
+      TI_WARN("No DX API detected.");
```
bobcao3, line 95: nit: Should change it to "DirectX 11 Unavailable, can not detect a supported device" or something like that. DX API sounds a bit vague.

bobcao3 (changes requested):
Great work and many thanks! Just need to fix the global variables and the cmake files a bit and this looks good to me! Maybe we can get DX backend in by the next release

AmesingFlank (commented):
All the changes in chi_examples shouldn't be there.

taichi/backends/dx/dx_api.cpp
```diff
@@ -0,0 +1,320 @@
+#include "taichi/backends/dx/dx_api.h"
+
+#include <memory>
+
+#include "taichi/backends/dx/dx_device.h"
+
+TLANG_NAMESPACE_BEGIN
```
ailzhang, line 7: let also just use explicit namespaces instead of this macro ;)

taichi/backends/dx/dx_api.h
```diff
@@ -0,0 +1,43 @@
...
+#include "taichi/common/core.h"
+#include "taichi/backends/device.h"
+#include <d3d11.h>
+#include <d3dcompiler.h>
+
+TLANG_NAMESPACE_BEGIN
+
+namespace directx11 {
+
+bool initialize_dx(bool error_tolerance = false);
```
ailzhang, line 14: is `error_tolerance` used here?

ailzhang (commented):
LGTM!

taichi/inc/archs.inc.h
```diff
@@ -11,7 +11,7 @@ PER_ARCH(wasm)   // WebAssembly
 PER_ARCH(cuda)    // NVIDIA CUDA
 PER_ARCH(metal)   // Apple Metal
 PER_ARCH(opengl)  // OpenGL Compute Shaders
-PER_ARCH(dx)      // Microsoft DirectX, N/A
+PER_ARCH(dx11)      // Microsoft DirectX, N/A
```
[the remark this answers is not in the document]
quadpixels: Done.

taichi/backends/opengl/opengl_device.cpp
quadpixels: Done.

taichi/backends/dx/dx_api.cpp
quadpixels: Done.

python/taichi/examples/chi_examples/main.cpp
quadpixels: Done. Removed this file (main.cpp)

python/taichi/examples/chi_examples/README.md
quadpixels: Done. Removed this file (README.md)

python/taichi/examples/chi_examples/CMakeLists.txt
quadpixels: Done. Removed this file.

cmake/TaichiCore.cmake
quadpixels: Added a check for TI_WITH_DX11 and TI_WITH_OPENGL and enabling spirv_hlsl and/or spirv_cross as neded.

taichi/backends/dx/dx_api.cpp
quadpixels: Done.

taichi/backends/dx/dx_api.cpp
quadpixels: Removed this function altogether, this was only used once in materialize_runtime, and now materialize_runtime just calls the ctor of DxDevice.

taichi/backends/dx/dx_api.h
quadpixels: Ah ..! It is not used and is now removed. (actually initialize_dx is removed too)

taichi/program/program.cpp
quadpixels: Done.

taichi/backends/dx/dx_device.cpp
quadpixels: Moved the binding to alloc ID map to DxResourceBinder.
Moved the alloc_id to Buffer and to UAV maps to DxDevice.

taichi/backends/dx/dx_device.cpp
quadpixels: It looks this function has never been called through the test cases I had run, so I'm setting this to "not implemented".

taichi/backends/dx/dx_device.cpp
quadpixels: It seems PipelineSourceType::hlsl_src still generates SPIRV binary for now, I have removed the check here and will modify this code path after figuring out how to generate HLSL source directly.

taichi/backends/dx/dx_device.cpp
quadpixels: Acknowledged.

taichi/codegen/spirv/spirv_ir_builder.cpp
quadpixels: Acknowledged.

k-ye:
One general suggestion: please consider breaking this into smaller PRs so that we can kick off the merge-back process :-)

quadpixels:
> One general suggestion: please consider breaking this down into smaller PRs so that we can kick off the merge-back process :-)

Acknowledged, I'll revisit the changes and apply the "small fixes" (like moving global variables into the device class and using snake case) from the beginning and update this PR (and multiple PRs when this one gets broken down).

PR #3779, opened 2021-12-12
[llvm] [bug] Support atomic min/max for unsigned int type

### LINKED ISSUE #3751: LLVM backends not handle atomic max/min for unsigned int
**Describe the bug**
Currently LLVM codegen not support atomic_max/atomic_min for unsigned int type (ti.u32/ & ti.u64). But there are not error throwing in frontend and codegen. So the uint atomic_min/atomic_max behavior in cuda and cpu backends is not correctly.

**To Reproduce**

```py
import taichi as ti
ti.init(arch=ti.cpu)

x = ti.field(ti.u64, shape=100)

@ti.kernel
def test():
    for I in x:
        x[I] = ti.cast(1, ti.u64) << 63 # line (1), output: 9223372036854775808
        # x[I] = 10000 # line (2), output: 100
    x[1] = 100
    for I in x:
        ti.atomic_min(x[0], x[1])
    print(x[0])

test()
```

g1n0st:
Related issue = close #3751


tests/python/test_atomic.py
```diff
@@ -47,6 +47,33 @@ def test_atomic_add_global_f32():
...
+def test_atomic_min_max_uint():
+    x = ti.field(ti.u64, shape=100)
+
+    @ti.kernel
+    def test0():
+        for I in x:
+            x[I] = 0
+        x[1] = ti.cast(1, ti.u64) << 63
+        for I in x:
+            ti.atomic_max(x[0], x[1])
```
strongoier, line 60:
```suggestion
            ti.atomic_max(x[0], x[I])
```
Is this the original intention?

tests/python/test_atomic.py
```diff
@@ -47,6 +47,33 @@ def test_atomic_add_global_f32():
...
+    test0()
+    assert x[0] == 9223372036854775808
+
+    @ti.kernel
+    def test1():
+        for I in x:
+            x[I] = ti.cast(1, ti.u64) << 63
+        x[1] = 100
+        for I in x:
+            ti.atomic_min(x[0], x[1])
```
strongoier, line 71: ditto

strongoier (changes requested):
Looks great!

tests/python/test_atomic.py
g1n0st: Fixed. I think both are OK.

tests/python/test_atomic.py
g1n0st: Fixed.

strongoier (approved):
LGTM!

ailzhang (approved):
Thanks for fixing it!

Issue #3782, opened 2021-12-13
Python frontend API cleanup 


ailzhang:
## Motivation:
We are in the process of revamping our [API doc site](https://api-docs.taichi.graphics/) and found that we currently expose a lot of internal data structures / functions to users. While this problem can be partially resolved by not showing them on the website through a config file, we'd like to take this opportunity to refactor & cleanup our python frontend as well. We believe it'll make it easier to further develop in our python frontend. 

There are two main blockers:
- Currently Taichi uses a fair amount of `from ... import *` in the python codebase which results in many undocumented modules/functions/classes
- There're some circular imports inside `taichi.lang` which resulted in many hacks/workarounds in the python implementation. 

## Goals & Milestones: 
- [ ] Be mindful of which module to expose under `taichi` namespace, an desired end state is:
```python
taichi/
  ti.aot
  ti.ad
  ti.types
  ti.ui
  ti.linalg
  ti.profiler
  ti.tools
```
	
  - This is currently WIP by @strongoier , @lin-hitonami  and @ailzhang.
  - [ ] `taichi.lang` refactor involves a lot of circular imports and it's currently WIP.

- [ ] Be mindful of which functions / classes to expose under `taichi` namespace
- [ ] Make sure top level APIs have **doc strings**.
- [ ] Be mindful the sub-package structure exposed under `taichi` namespace, for example `taichi.aot.*, taichi.ad.*`. 
- [ ]  `import taichi` should just do imports (thus it should be blazing fast). In other words we need to reduce the code execution happening here.

## Design discussions:

-  Top level or structured API?
    - Historically `taichi` exposes most of its language related public APIs in the `taichi` namespace (instead of `taichi.lang.*`). This is flexible because python file movement won't affect end users as long as we keep passing and exposing them in `taichi`.
    - **[proposal]** We'll keep this tradition to maximize flexibility, but for sub-components `aot/ad/ui/profiler` etc we'll develop them in sub-packages to make it more structured.
- Circular imports!
    - Taichi has suffered from circular imports for a long time, and we plan to clean it up during project. One of the main issues here is using `ti.kernel`s and `ti.func`s inside `taichi/lang/`, which is supposed to be the python wrapper of taichi programming language. Those kernels and funcs are more like standard lib functions on top of the taichi language, so we should make this clear during the refactor.
[...]

PR #3788, opened 2021-12-13
[refactor] Move diagnose.py and cc_compose.py to tools/

lin-hitonami:
Related issue = #3782
`yaml.load` without specifying `Loader` is deprecated, so I fixed it by the way.


ailzhang (commented):
Shall we move this into tools/ as well?

ailzhang (approved):
LGTM!

Issue #3781, opened 2021-12-12
[RFC] Taichi on Javascript


AmesingFlank:
I'm writing this RFC to share my thoughts and plans to build a version of Taichi completely embedded in Javascript, with the hope of getting production-level applications to use Taichi, and thereby increasing the impact of Taichi as a programming language.

Currently, Taichi uses Python as the host language. Thanks to Python's simplicity, Taichi enables programmers with little GPU experiences to have fun and experiment with GPU computing. Javascript shares the simplicity of Python (to some extent), and additionally, it has the advantage of being used extensively in production-level, user-facing applications. For this reason, if we offer Taichi as a (deployable) Javascript package, it would provide professional web developers with a brand new way of building and serving compute-heavy or graphics-heavy web programs. This would certainly boost the popularity of Taichi.

To be more specifically, I'm planning to build a Javascript library that enables this:

```javascript
const ti = require("taichi")

x = ti.Vector.field(3, ti.f32, 512)

let k = ti.kernel(() => {
    for (i of ti.range(512)) {
        x[i] = x[i] + 1
    }
})

k()
```

Here, `taichi` would be a normal Javascript module, one that could be installed simply by running  `npm install taichi`. Most importantly, it must be a pure Javascript module, meaning that it must have **no binary code dependencies** attached to it. This is of course vital if we want web developers to be able to use it in their own products. We will of course implement interop routines with normal Javascript arrays, HTML canvases, and maybe even provide a GGUI. 

# Use cases and Impacts

From the top of my head I can think of a lot of use cool cases for a Javascript version of Taichi

* Graphics applications on the web. 
No doubt there are lots and lots of graphics demands on web pages, and these are done most often via WebGL, which is flexible but cumbersome. As a language directly embedded in JS, this could be very appealing for beginner-level front-end engineers who don’t want to learn WebGL.

* Compute heavy applications on the web
These are considerably rarer, but part of the reason for that is exactly because it’s difficult to develop high performance programs on the web -- and taichi just might change that.

* Other fancy apps:
    * A (much better) taichi zoo, where actually everything happens on the frontend, and the backend server does nothing but serving (instead of compiling).
    * A compute resource sharing platform:
This is something I thought about a few years ago. We could build a site where people can submit compute tasks written in taichi, and other people could simple open up the website and contribute their (GPU) computing resources.
    * A transpiler between python-taichi and JS-taichi, which helps improve the impact of the python language.
 

# Technological Considerations and Challenges

[...]

Issue #3764, opened 2021-12-09
Any parallel sorting on the GPU with Taichi or thirdparty ?


GrapixLeGrand:
Hello,

I could not find a thread (nor in the docs) that would cover this subject, therefore I created one. Is there any parallel sorting algorithm in taichi? I'm looking for a radix/counting sort. Is there a known taichi implementation of a radix/counting sort out there? I was considering writing my own following this [paper](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=5395307). However, according to the doc (under the [atomics](https://taichi.readthedocs.io/en/stable/atomic.html?highlight=atomic) section), the atomic assign operation is missing and it would be needed for step 4 of the algorithm  `b[c[r[a[i]]]++] = a[i];`. 

Best,
Quentin

qiao-bo:
[](url)Hi Quentin, 
I am not aware of an existing GPU-accelerated sorting algo in Taichi, perhaps other people can comment on this.

Regarding to the atomic assign operation, would this work [(doc)](https://docs.taichi.graphics/lang/articles/basic/type#supported-atomic-operations)?
```
idx = ti.atomic_add(c[b[a[i]]], 1)
b[idx] = a[i]
```

GrapixLeGrand:
Hello back,

It worked for the atomics, thanks! Now, do you know how to get the max number of global threads?
Best,
Quentin

turbo0628:
> We do think a sort intrinsic can be useful for many Taichi users. We are considering add some parallel intrinsics to the language. Let us know what else you find useful ;)

Is that possible to add some hooks to for external libraries? Many such math functions / permutation utilities are already implemented in vendor libraries, CUB for example.

qiao-bo:
> > We do think a sort intrinsic can be useful for many Taichi users. We are considering add some parallel intrinsics to the language. Let us know what else you find useful ;)
> 
> Is that possible to add some hooks to for external libraries? Many such math functions / permutation utilities are already implemented in vendor libraries, CUB for example.

Very good point! CUB exposes very efficient permutation utilities, and it's hard to be better ;). Although in Taichi we want to avoid Vendor lock-in, having some external libs could still be a good start...

turbo0628:
> Very good point! CUB exposes very efficient permutation utilities, and it's hard to be better ;). Although in Taichi we want to avoid Vendor lock-in, having some external libs could still be a good start... 

[...]

PR #3790, opened 2021-12-14
[lang] Add parallel sort utility

AmesingFlank:
This is an implementation of parallel odd-even merge sort.

related issue = #3764 

References:
 
https://developer.nvidia.com/gpugems/gpugems2/part-vi-simulation-and-numerical-algorithms/chapter-46-improved-gpu-sorting
https://en.wikipedia.org/wiki/Batcher_odd%E2%80%93even_mergesort

This will be very useful for #3781 as well. We likely won't be having atomics in WebGPU (sadge), but one can always use a sort to simulate atomic operations. To demonstrate this, I will open another PR to add versions of `mpm99.py` and `pbf2d.py` that are atomics-free but uses sorting.

### DIFF
--- a/python/taichi/tools/sort.py
+++ b/python/taichi/tools/sort.py
@@ -1,6 +1,10 @@
 import taichi as ti
 
 
+# Odd-even merge sort
+# References:
+# https://developer.nvidia.com/gpugems/gpugems2/part-vi-simulation-and-numerical-algorithms/chapter-46-improved-gpu-sorting
+# https://en.wikipedia.org/wiki/Batcher_odd%E2%80%93even_mergesort
 def parallel_sort(x):
     N = x.shape[0]
 
--- a/python/taichi/tools/sort.py
+++ b/python/taichi/tools/sort.py
@@ -1,3 +1,5 @@
+from taichi.core.util import ti_core as _ti_core
+
 import taichi as ti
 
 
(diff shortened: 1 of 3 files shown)


python/taichi/tools/sort.py
```diff
@@ -0,0 +1,29 @@
...
+                        x[a] = val_b
+                        x[b] = val_a
+
+    p = 1
+    while p < N:
+        k = p
+        while k >= 1:
+            invocations = int((N - k - k % p) / (2 * k)) + 1
+            sort_stage(x, N, p, k, invocations)
+            ti.sync()
```
AmesingFlank, line 27: This simply shouldn't be here, It shouldn't be necessary and will make the program slower. However, on Vulkan, if this is removed, the results are wrong. We probably need to add a `buffer_barrier` or something between consecutive compute dispatches.

AmesingFlank: created #3791 to fix the vulkan problem. After this is fixed, I will submit another PR to remove the `ti.sync()`

bobcao3:
Maybe add an option where we can sort based on a sort key? Which takes two arrays `(N, k)`, which will make this a lot more useful I'd imagine.

qiao-bo:
Hi, thanks a lot for the quick implementation. I wonder do we have another place to put these parallel intrinsics? Since more will be added later, IMHO looking at the tools folder, they seems not fit into the same category ;)

AmesingFlank:
> Hi, thanks a lot for the quick implementation. I wonder do we have another place to put these parallel intrinsics? Since more will be added later, IMHO looking at the `tools` folder, they seem not fitting into the same category ;)

Yeah I'm not sure if this is right place either. I chose the `tools` folder because this folder contains things that are not core to the taichi language, but rather utilities and helpers, and that's what this `parallel_sort` function is. I'm happy to move it to another place though.

qiao-bo:
> > Hi, thanks a lot for the quick implementation. I wonder do we have another place to put these parallel intrinsics? Since more will be added later, IMHO looking at the `tools` folder, they seem not fitting into the same category ;)
> 
> Yeah I'm not sure if this is right place either. I chose the `tools` folder because this folder contains things that are not core to the taichi language, but rather utilities and helpers, and that's what this `parallel_sort` function is. I'm happy to move it to another place though.
> 
> @ailzhang @k-ye @qiao-bo suggestions on where to put this?

We have some linalg implementations at `python/lang/linalg_impl.py`, such as `ti.svd()`. Maybe also in lang? But i am not sure if we will move these impl to C++ IR in the future?

AmesingFlank:
I don't really think it belongs in `lang`, as it is not part of the language, but rather just a utility function. And I don't think we will be moving it to C++ IR, it's a lot less readable.

qiao-bo:
ehmm, in this case it is a utility helper correct. I thought we could offer functions such as parallel sort, reduce, scan... as the parallel intrinsics offered by the Taichi language. Nevertheless, i am also fine with keeping it here at `tools` for now, unless someone has a better suggestion ;/

AmesingFlank:
> I thought we could offer functions such as parallel sort, reduce, scan... as the parallel intrinsics offered by the Taichi language.

That's a pretty good idea. We should group these into the same directory once we also have an API for reduce and scan. But for now I think let's just leave it in tools.

AmesingFlank:
Can i please get a stamp :) @qiao-bo @ailzhang @k-ye @bobcao3

qiao-bo (approved):
Thanks!

PR #3792, opened 2021-12-14
[Lang] Add "In" support in static scope

### LINKED ISSUE #3783: Add support for operator "in"
Taichi itself does not support operator `in`. However, before the refactoring of the AST transformer, code like `if ti.static(dtype in [ti.f32, ti.f64]):` can run perfectly because they were handled by Python instead of taichi AST transformer. Sadly, these code can no longer run after the refactoring because all the expressions are evaluated by the AST transformer now. Similar usages are widely used by the users, so operator `in` should be supported in taichi. 

The current implementation of `list` as a taichi variable does not support `in` because it converts all the variables it contains to taichi objects when it is assigned. so we should only allow operator `in` on normal `list`s and `dict`s.

We can wrap the `list`s in a new wrapper class in `expr_init`, an disallow operator `in` on that class. (Maybe throw an exception in \_\_contains\_\_)

lin-hitonami:
Related issue = close #3783


strongoier (approved):
LGTM!

PR #3795, opened 2021-12-14
[Lang] Enforce single deterministic return in taichi kernels and functions

### LINKED ISSUE #3694: Random value is returned when kernel return type is specified but no return statement is executed

lin-hitonami:
Related issue = close #3694


python/taichi/lang/ast/ast_transformer_utils.py
```diff
@@ -150,7 +150,7 @@ def loop_status(self):
     def set_loop_status(self, status):
         self.loop_scopes[-1].status = status
 
-    def is_in_static(self):
+    def is_in_static_for(self):
         if len(self.loop_scopes):
```
yihong0618, line 154: maybe more pythonic?
```python 
if self.loop_scopes:
```

tests/python/test_syntax_errors.py
```diff
@@ -275,8 +275,9 @@ def safe_sqrt(a):
     def kern(a: float):
         print(safe_sqrt(a))
 
-    with pytest.raises(ti.TaichiCompilationError,
```
ailzhang, line 278: can we also add a test for static case just to make sure it passes ;) ?

ailzhang (approved):
LGTM! Thanks!

PR #3799, opened 2021-12-14
[bug] Create core folder if it doesn't exist.

ailzhang:
Building taichi from source on master failed in my workstation, this PR should fix it


k-ye (approved):

ailzhang:
@lin-hitonami  I can repro this bug if I do `git clean -xfd; python setup.py clean` before `python setup.py develop`. IIUC this is related to our hack here https://github.com/taichi-dev/taichi/blob/master/setup.py#L171-L173

PR #3803, opened 2021-12-14
[Bug] [opengl] Modify the implementation of norm_sqr()

### LINKED ISSUE #3796: Cornell box example is broken on opengl backend
It's shows below on current master. I've confirmed it was correct on 0.8.5 release so we need to do a bisect. 
image

strongoier:
Related issue = fixes #3796

The phenomenon in #3796 appears after #3517. However, #3517 is not the root cause - after #3517, since the initial IR of that example changes, the optimization of `pow` in CHI IR goes into a different path and lets the OpenGL backend to calculate the result, which exposes a bug which already exists.

Let's look at a code snippet:
```py
import taichi as ti

ti.init(ti.opengl)

@ti.kernel
def test(a: ti.f32, b: ti.f32) -> ti.f32:
    return a ** b

print(test(-1.0, 2.0))
```
It will give `nan` as the result because in OpenGL `pow(x, y)` has undefined behavior when `x < 0` (https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/pow.xhtml). Therefore, we should avoid doing such calculation.

In that example, `pow()` comes from `norm_sqr()` of the `Matrix` class. To fix the problem, I replace `pow(x, 2)` with `x * x` in this PR.

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -628,7 +628,7 @@ class Matrix(TaichiOperations):
     @kern_mod.pyfunc
     def norm_sqr(self):
         """Return the sum of the absolute squares of its elements."""
-        return (self**2).sum()
+        return (self * self).sum()
 
     @kern_mod.pyfunc
     def max(self):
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Nice catch! Thanks!

PR #3812, opened 2021-12-15
[misc] [bug] Fix legacy benchmarks/run.py

yolo2themoon:
Related issue = #

### DIFF
--- a/benchmarks/async_cases.py
+++ b/benchmarks/async_cases.py
@@ -4,7 +4,7 @@ import sys
 
 import taichi as ti
 
-sys.path.append(os.path.join(ti.core.get_repo_dir(), 'tests', 'python'))
+sys.path.append(os.path.join(sys.path[0], '../', 'tests', 'python/'))
 
 from fuse_test_template import (template_fuse_dense_x2y2z,
                                 template_fuse_reduction)
--- a/benchmarks/async_cases.py
+++ b/benchmarks/async_cases.py
@@ -2,9 +2,11 @@ import math
 import os
 import sys
 
+from utils import get_benchmark_dir
+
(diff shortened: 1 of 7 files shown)


benchmarks/async_cases.py
```diff
@@ -4,7 +4,7 @@
 
 import taichi as ti
 
-sys.path.append(os.path.join(ti.core.get_repo_dir(), 'tests', 'python'))
+sys.path.append(os.path.join(sys.path[0], '../', 'tests', 'python/'))
```
qiao-bo, line 7: Can we briefly comment here on the path ;)? Before it was clear where it points to, now we depend on sys.path[0]?

qiao-bo (commented):
Thanks!

benchmarks/async_cases.py
yolo2themoon: `ti.core.get_repo_dir()` now returns the directory`/home/username/.taichi/`,
and `sys.path[0]` points to the directory where the current file is located.

qiao-bo: maybe `__file__` is better? `sys.path` is a list of search path i believe.

yolo2themoon: Thanks for the tip, when trying to use `__file__`, I found that the absolute path was missing
sys.path[0] = `/home/username/.../taichi_dir/benchmarks`
__file__ = `run.py`

yolo2themoon: How about we use `os.path.dirname(os.path.realpath(__file__))`, which also serves the purpose

qiao-bo (approved):
Thanks!

PR #3816, opened 2021-12-16
[misc] Slient & non-blocking version check

Leonz5288:
Now version check runs silently per 7 days. Nothing will be printed in terminal.
If user's version is outdated, a latest_version file will be written to ~/.taichi and the latest version will be recorded.
After that, no more version check in 7 days. When user use Taichi again, if the current version is lower than the version recorded in the file, the latest version will be printed with the Taichi header. For example:
`[Taichi] version 0.8.7, latest version 0.9.0, llvm 10.0.0, commit abcdef, linux, python 3.8.10`

### DIFF
--- a/python/taichi/_lib/utils.py
+++ b/python/taichi/_lib/utils.py
@@ -141,6 +141,7 @@ def require_version(major, minor=None, patch=None):
 
 at_startup()
 
+
 def compare_version(latest, current):
     latest_num = map(int, latest.split('.'))
     current_num = map(int, current.split('.'))
@@ -152,6 +153,7 @@ def compare_version(latest, current):
         else:
             continue
     return False
+
 
 def _print_taichi_header():
     header = '[Taichi] '
--- a/python/taichi/_lib/utils.py
+++ b/python/taichi/_lib/utils.py
@@ -153,11 +153,11 @@ def _print_taichi_header():
     header += f'version {ti_core.get_version_string()}, '
 
     try:
-        version_path = os.path.join(ti_core.get_repo_dir(), 'latest_version')
-        if os.path.exists(version_path):
+        timestamp_path = os.path.join(ti_core.get_repo_dir(), 'timestamp')
+        if os.path.exists(timestamp_path):
             latest_version = ''
-            with open(version_path, 'r') as f:
-                latest_version = f.readlines()[0].rstrip()
+            with open(timestamp_path, 'r') as f:
+                latest_version = f.readlines()[1].rstrip()
             if compare_version(latest_version, ti_core.get_version_string()):
                 header += f'latest version {latest_version}, '
     except:
--- a/python/taichi/_lib/utils.py
+++ b/python/taichi/_lib/utils.py
@@ -148,10 +148,9 @@ def compare_version(latest, current):
     for x, y in zip(latest_num, current_num):
         if x > y:
             return True
-        elif x < y:
+        if x < y:
             return False
-        else:
-            continue
+        continue
     return False
 
 
--- a/python/taichi/_lib/utils.py
+++ b/python/taichi/_lib/utils.py
@@ -154,7 +154,7 @@ def _print_taichi_header():
                 latest_version = f.readlines()[0].rstrip()
             if latest_version > ti_core.get_version_string():
                 header += f'latest version {latest_version}, '
-    except Exception as error:
+    except Exception:
         pass
 
     llvm_version = ti_core.get_llvm_version_string()
--- a/python/taichi/_lib/utils.py
+++ b/python/taichi/_lib/utils.py
@@ -141,6 +141,17 @@ def require_version(major, minor=None, patch=None):
 
 at_startup()
 
+def compare_version(latest, current):
+    latest_num = map(int, latest.split('.'))
+    current_num = map(int, current.split('.'))
+    for x, y in zip(latest_num, current_num):
+        if x > y:
+            return True
+        elif x < y:
+            return False
+        else:
+            continue
+    return False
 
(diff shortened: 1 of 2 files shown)


Leonz5288:
Now this service runs in a new thread and main process will not wait for the new thread to finish (prevent blocking user).
I also tested a few cases related to network issues, for example, timeout, domain mismatch, network down, etc. Nothing will be printed on terminal and Taichi main process will run as normal.
If there are other cases that comes to your mind, please tell me so I can test more.

python/taichi/lang/__init__.py
```diff
@@ -453,18 +453,15 @@ def check_version():
...
-            elif response['status'] == 0:
-                # Status 0 means that user already have the latest Taichi. The message here prompts this infomation to users.
-                print(response['message'])
-    except Exception as error:
-        print('Checking lastest version failed:', error)
+                version_path = os.path.join(_ti_core.get_repo_dir(),
+                                            'latest_version')
+                with open(version_path, 'w') as f:
+                    f.write(response['latest_version'])
+    except Exception:
```
ailzhang, line 463: Wild but might be a good fit here as we discussed @rexwangcc @strongoier :P  : 
```
except:
    pass
```

python/taichi/lang/__init__.py
```diff
@@ -481,15 +478,14 @@ def try_check_version():
...
-                    f.truncate()
         else:
             check_version()
             with open(timestamp_path, 'w') as f:
                 f.write((cur_date +
                          datetime.timedelta(days=7)).strftime('%Y-%m-%d'))
     # Wildcard exception to catch potential file writing errors.
-    except Exception as error:
-        print('Checking lastest version failed:', error)
+    except Exception:
```
ailzhang, line 487: ditto

python/taichi/lang/__init__.py
```diff
@@ -521,7 +517,10 @@ def init(arch=None,
     # Check version for users every 7 days if not disabled by users.
     skip = os.environ.get("TI_SKIP_VERSION_CHECK")
     if skip != 'ON':
-        try_check_version()
+        # We don't join this thread because we do not wish to block the user.
+        check_version_thread = threading.Thread(target=try_check_version,
```
ailzhang, line 521: Just to double check: have you tried locally that just raise in this thread and see if it crash the main process?

ailzhang (commented):
LGTM!

python/taichi/lang/__init__.py
Leonz5288: OK, didn't know it can work like this

python/taichi/lang/__init__.py
Leonz5288: I tried crash the thread and not catch error. The sub thread will crash and a bunch of output will be printed, but the main thread can run as normal.

python/taichi/lang/__init__.py
```diff
@@ -481,15 +478,14 @@ def try_check_version():
                 with open(timestamp_path, 'w') as f:
```
frostming, line 478: We may not need a timestamp file. Reading the mtime should work.

Leonz5288: What do you mean by mtime?

python/taichi/lang/__init__.py
frostming: Yeah, the exceptions raised in sub thread can't fail the main thread. But I am not sure if `sys.exit` would work

python/taichi/lang/__init__.py
frostming: last modified time returned by `os.stat(filename)`

python/taichi/_lib/utils.py
```diff
@@ -142,10 +142,33 @@ def require_version(major, minor=None, patch=None):
 at_startup()
 
 
+def compare_version(latest, current):
+    latest_num = map(int, latest.split('.'))
+    current_num = map(int, current.split('.'))
+    for x, y in zip(latest_num, current_num):
```
frostming, line 148: The `tuple[int]` can be compared directly with `<`/`>`/`==`

Leonz5288: Thanks for the info!

python/taichi/lang/__init__.py
Leonz5288: I see. However, I don't think we can guarantee that users won't modify the file. Maybe we should stay timestamp for now?

frostming: Reasonable, but can we store timestamp with the version in the same file, maybe a json file for convenience. In this way we don't scatter misc files around.

python/taichi/lang/__init__.py
rexwangcc: didn't know it either, tested with:

```python
try:
     raise BaseException(99)
except:
     print(42)
```

rexwangcc (approved):
LGTM, thanks! (let's resolve all comments before merging :smile: )

python/taichi/lang/__init__.py
Leonz5288: Understood, let me adjust this.

python/taichi/lang/__init__.py
Leonz5288: Can you elaborate about sys.exit please?

frostming: I may have understood your need, if you don't want it to fail the main program, then subthread is OK.

Leonz5288: Yes, the main program is the last thing we wish to fail.

PR #3818, opened 2021-12-16
[vulkan] Fix command serial ordering & uses less queue submits

### LINKED ISSUE #3791: [Vulkan] `ti.sync()` changes program behavior on Vulkan
Example of when `ti.sync()` changes program behavior on Vulkan: #3790. In this example, the program only behaves correctly when the `ti.sync()` is added. 

On vulkan, we aggressively batch consecutive kernel launches into a single command buffer. We probably need to add a buffer_barrier or something between consecutive compute dispatches. I will investigate this.

bobcao3:
Fixes #3791

### DIFF
--- a/taichi/backends/vulkan/runtime.cpp
+++ b/taichi/backends/vulkan/runtime.cpp
@@ -118,9 +118,9 @@ class HostDeviceContextBlitter {
 #undef TO_DEVICE
   }
 
-  void device_to_host() {
+  bool device_to_host(CommandList *cmdlist) {
     if (ctx_attribs_->empty()) {
-      return;
+      return false;
     }
 
     bool require_sync = ctx_attribs_->rets().size() > 0;
@@ -134,9 +134,9 @@ class HostDeviceContextBlitter {
     }
 
     if (require_sync) {
-      device_->get_compute_stream()->command_sync();
+      device_->get_compute_stream()->submit_synced(cmdlist);
     } else {
-      return;
+      return false;
     }
 
     char *const device_base =
@@ -205,6 +205,8 @@ class HostDeviceContextBlitter {
 #undef TO_HOST
 
     device_->unmap(*host_shadow_buffer_);
+
+    return true;
   }
 
   static std::unique_ptr<HostDeviceContextBlitter> maybe_make(
@@ -259,17 +261,8 @@ CompiledTaichiKernel::CompiledTaichiKernel(const Params &ti_params)
     }
   }
   const auto ctx_sz = ti_kernel_attribs_.ctx_attribs.total_bytes();
-  if (!ti_kernel_attribs_.ctx_attribs.empty()) {
-    ctx_buffer_ = ti_params.device->allocate_memory_unique(
-        {size_t(ctx_sz),
-         /*host_write=*/true, /*host_read=*/false,
-         /*export_sharing=*/false, AllocUsage::Storage});
-    ctx_buffer_host_ = ti_params.device->allocate_memory_unique(
-        {size_t(ctx_sz),
-         /*host_write=*/false, /*host_read=*/true,
-         /*export_sharing=*/false, AllocUsage::Storage});
-    input_buffers_[BufferType::Context] = ctx_buffer_.get();
-  }
+
+  ctx_buffer_size_ = ctx_sz;
 
   const auto &task_attribs = ti_kernel_attribs_.tasks_attribs;
   const auto &spirv_bins = ti_params.spirv_bins;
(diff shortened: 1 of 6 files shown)


taichi/backends/vulkan/runtime.cpp
```diff
@@ -411,26 +407,46 @@ VkRuntime::KernelHandle VkRuntime::register_taichi_kernel(
...
+  if (ti_kernel->get_ctx_buffer_size()) {
+    ctx_buffer = device_->allocate_memory_unique(
+        {ti_kernel->get_ctx_buffer_size(),
+           /*host_write=*/true, /*host_read=*/false,
+           /*export_sharing=*/false, AllocUsage::Storage});
+    ctx_buffer_host = device_->allocate_memory_unique(
+        {ti_kernel->get_ctx_buffer_size(),
+           /*host_write=*/false, /*host_read=*/true,
+           /*export_sharing=*/false, AllocUsage::Storage});
+  }
```
AmesingFlank, line 422: r we creating new buffers every time we launch a kernel?

bobcao3: yes

AmesingFlank: .. that can't be good

bobcao3: It is fine, it is actually faster than before

AmesingFlank: be that as it may, we should still at least try to reuse some of these buffers. It is fast now because VMA is doing a superb job, but what if we want to unify the runtime code for different backends?

bobcao3: We can't reuse them until we know for sure they are done. A better solution is a ring buffer, but in terms of memory usage it's the same.

bobcao3: Also for DX11 and OpenGL this is going to be faster than a ring buffer because the drivers know for sure there isn't overlap and is free to use them async.

bobcao3: So in the end only maybe Vulkan / DX12 / and Metal will benefit from a ring buffer, and all of these requires a robust memory allocator anyways.

AmesingFlank: What about the overhead of allocating buffers on DX and GL? Are we confident that the performance gain from having things async outweighs these overheads?

bobcao3: If it's buffer reuse it should be implemented in the allocator instead of the runtime

bobcao3: Let's do the naive way for now, we can figure out a better way later

taichi/backends/vulkan/runtime.cpp
```diff
@@ -293,15 +286,14 @@ size_t CompiledTaichiKernel::num_pipelines() const {
...
+size_t CompiledTaichiKernel::get_ctx_buffer_size() const {
+  return ctx_buffer_size_;
 }
 
-DeviceAllocation *CompiledTaichiKernel::ctx_buffer_host() const {
-  return ctx_buffer_host_.get();
-}
-
-void CompiledTaichiKernel::command_list(CommandList *cmdlist) const {
+void CompiledTaichiKernel::command_list(
```
k-ye, line 293: Out of curiosity, what does the function name `command_list` mean here, `run_command_list`?

taichi/backends/vulkan/runtime.cpp
```diff
@@ -411,26 +408,46 @@ VkRuntime::KernelHandle VkRuntime::register_taichi_kernel(
 
 void VkRuntime::launch_kernel(KernelHandle handle, RuntimeContext *host_ctx) {
   auto *ti_kernel = ti_kernels_[handle.id_].get();
+
+  std::unique_ptr<DeviceAllocationGuard> ctx_buffer_host, ctx_buffer;
```
k-ye, line 412:
```suggestion
  std::unique_ptr<DeviceAllocationGuard> ctx_buffer_host{nullptr}, ctx_buffer{nullptr};
```

taichi/backends/vulkan/runtime.cpp
```diff
@@ -411,26 +408,46 @@ VkRuntime::KernelHandle VkRuntime::register_taichi_kernel(
...
+
+  if (ti_kernel->get_ctx_buffer_size()) {
+    ctx_buffer = device_->allocate_memory_unique(
```
k-ye, line 415: Would this result in memory fragmentation? Or the VMA should take care of it?

taichi/backends/vulkan/runtime.cpp
bobcao3: This means record command list

taichi/backends/vulkan/runtime.cpp
bobcao3: It would, let's do the naive one for now until we kick ext arr out of context buffer and add a ring buffer

taichi/backends/vulkan/runtime.cpp
k-ye: nit: Can we just rename it to `record_command_list` then :-) ?

taichi/backends/vulkan/runtime.cpp
k-ye: SG

k-ye (approved):
LGTM!

ailzhang (approved):
Nice work! Thanks!

PR #3822, opened 2021-12-17
[bug] Fix static grouped for

### LINKED ISSUE #3819: A potential bug for `ti.static` for `Taichi >= 0.8.6`
```python
import taichi as ti

ti.init(ti.cpu)


@ti.kernel
def test_range():
    for I in ti.static(ti.grouped(ti.ndrange((4, 5), (3, 5), 5))):
        tmp = I.cast(float)  # not work
        print(tmp)


@ti.kernel
def test_range2():
    for I in ti.grouped(ti.ndrange((4, 5), (3, 5), 5)):
        tmp = I.cast(float)  # it works
        print(tmp)


if __name__ == "__main__":
    test_range2()
    test_range()

```

Starting from `taichi == 0.8.6`, the  `test_range` function would raise an **error**:

```bash
taichi.lang.exception.TaichiCompilationError: On line 9 of file 
        tmp = I.cast(float)
              ^^^^^^
AttributeError: 'tuple' object has no attribute 'cast'
```

while the `test_range2` would not.

Here the only difference between these two `test_range` functions is the `ti.static`. 

**So I wonder what happened to the `ti.static` during the recent update?**

What also confuses me more is a fix to this is kind of weird.  In `taichi_elements` repo, if we update the `Taichi>=0.8.6 ` , we will need a fix like [this](https://github.com/taichi-dev/taichi_elements/pull/86). 

```python
for offset in ti.static(ti.grouped(self.stencil_range())):
	dpos = ti.Vector(offset).cast(float) - fx
	g_v = grid_v_in[base + offset]
```

It seems that the `offset` here is actually not a normal `python tuple` right? So I wonder if this is a bug during recent update?

lin-hitonami:
Related issue = fixes #3819

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -673,7 +673,7 @@ class ASTTransformer(Builder):
                     f"Group for should have 1 loop target, found {len(targets)}"
                 )
             target = targets[0]
-            for value in ndrange_arg:
+            for value in ti.grouped(ndrange_arg):
                 with ctx.variable_scope_guard():
                     ctx.create_variable(target, value)
                     build_stmts(ctx, node.body)
--- a/tests/python/test_ast_refactor.py
+++ b/tests/python/test_ast_refactor.py
@@ -998,3 +998,16 @@ def test_default_template_args_on_func():
         return bar()
 
     assert foo() == 123
+
+
+@ti.test()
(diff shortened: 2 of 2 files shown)


strongoier (changes requested):
Add a test?

strongoier (approved):
LGTM!

yuanming-hu (approved):

PR #3827, opened 2021-12-17
[refactor] Remove import taichi in ast_transformer

lin-hitonami:
Related issue = #3782
In `taichi/lang/__init.py`, `from taichi.lang.ndrange import ndrange` is a bit confusing both to me and the IDE because it imports the class `ndrange` and shadows the module `ndrange`. But at the time of loading `ast_transformer.py`, this line is not yet executed. So, when I `from taichi.lang import ndrange` in `ast_transformer.py`, it imports the module. In order to solve this confusing situation, I moved the import of `ndrange` to `taichi/__init__.py`


k-ye (approved):
LGTM!

PR #3829, opened 2021-12-19
[lang] Fix ndarray cuda dealloc when using preallocated memory

qiao-bo:
When using preallocated memory, deallocation is handled by LLVMProgramImpl [finalize](https://github.com/taichi-dev/taichi/blob/3818ba3142288752ebdd8c57bd47e92501aafcac/taichi/llvm/llvm_program.cpp#L479) instead of CUDA device.

### DIFF
--- a/taichi/backends/cuda/cuda_device.cpp
+++ b/taichi/backends/cuda/cuda_device.cpp
@@ -23,6 +23,7 @@ DeviceAllocation CudaDevice::allocate_memory(const AllocParams &params) {
   info.size = params.size;
   info.is_imported = false;
   info.use_cached = false;
+  info.use_preallocated = false;
 
   DeviceAllocation alloc;
   alloc.alloc_id = allocations_.size();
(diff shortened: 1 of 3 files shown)


taichi/backends/cuda/cuda_device.cpp
```diff
@@ -69,7 +71,7 @@ void CudaDevice::dealloc_memory(DeviceAllocation handle) {
       TI_ERROR("the CudaCachingAllocator is not initialized");
     }
     caching_allocator_->release(info.size, (uint64_t *)info.ptr);
-  } else {
+  } else if (!info.use_preallocated) {
```
ailzhang, line 74: would you mind also explain (in code) what's the difference between `use_cached` and `use_preallocated`? ;) I feel like they might be merged into one flag.

qiao-bo: added in the decl part

ailzhang (approved):
LGTM！Thanks!

PR #3832, opened 2021-12-20
[opengl] Only bind buffers per ti.kernel.

ailzhang:
* #3835
* #3834
* #3833
* __->__ #3832

Binding won't change among different offloaded tasks in the same
ti.kernel so we can save some time here.

For my ndarray based mpm88 benchmark, this improves from 12.50 fps to
13.01 fps.


k-ye (approved):
LGTM!

PR #3833, opened 2021-12-20
[refactor] Rename is_external_array -> is_array and arr_bufs_ -> ext_arr_bufs_

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #3835
* #3834
* __->__ #3833
* #3832

This PR is a simple massive rename PR which serves as the first step to
separate internal ndarray and external array implementation.

PR #3834, opened 2021-12-20
[opengl] Only sync per ti.kernel when there's external array arg.

ailzhang:
* #3835
* __->__ #3834
* #3833
* #3832

For ndarray based mpm88 we now only sync when we plot on GUI, which is
after 500 substeps in my benchmark. This PR improves from 13.01fps to
16.67fps.


k-ye (approved):
LGTM!

PR #3835, opened 2021-12-20
[opengl] Create new args buf instead of reusing one args buf.

ailzhang:
* (to be filled)

When reusing one args_buf for different gl kernels we spend a lot of
time on glMapBufferRange() in syncing. But these syncs are indeed not
necessary. Now we just create one args_buf per kernel launch.
This improve mpm88 benchmark from 16.67fps to 30fps.

Thanks to @k-ye and @bobcao3 for helping!


ailzhang:
Closing this since I've opened a new one at #3848

PR #3837, opened 2021-12-20
[ci] Enforce using ninja on Windows in both release and normal testing

strongoier:
This is a follow-up PR of #3735 to make the release work properly.


qiao-bo (approved):
LGTM!

PR #3778, opened 2021-12-12
[gui] Fix vector to fast image

bobcao3:
Fixes #3675
And thus fixes cornell_box.py

PR #3839, opened 2021-12-20
[refactor] Recover 'in' operator in vector_to_fast_image()

strongoier:
In #3778, the use of `in` is temporarily removed due to lack of support. After #3792, `in` is supported again so we can add it back.


ailzhang (approved):
LGTM!

PR #3841, opened 2021-12-20
[bug] [cuda] Fix memset before reusing cuda cached memory

qiao-bo:
This prevents any side effect when read/write to the cached memory returned by ndarray memory allocator.


strongoier (approved):
LGTM!

ailzhang (approved):
lgtm!

PR #3843, opened 2021-12-20
[Lang] Use ndarray own memory allocator by default

qiao-bo:
This PR disables the usage of torch for ndarray by default, by switching to Taichi's own memory allocator.


strongoier (approved):
LGTM!

PR #3845, opened 2021-12-21
[Android] [gui] Add initial support of Android in GGUI

ghuau-innopeak:
Related issue = #3679

### DIFF
--- a/taichi/ui/backends/vulkan/app_context.h
+++ b/taichi/ui/backends/vulkan/app_context.h
@@ -5,17 +5,26 @@
 #include "taichi/backends/vulkan/vulkan_loader.h"
 #include "taichi/backends/vulkan/vulkan_device.h"
 #include "taichi/ui/backends/vulkan/swap_chain.h"
+#ifdef ANDROID
+#include <android/native_window.h>
+#endif
 
 TI_UI_NAMESPACE_BEGIN
+
+#ifdef ANDROID
+using TaichiWindow = ANativeWindow;
+#else
+using TaichiWindow = GLFWwindow;
+#endif
 
 namespace vulkan {
 
 class AppContext {
  public:
-  void init(GLFWwindow *glfw_window, const AppConfig &config);
+  void init(TaichiWindow *window, const AppConfig &config);
   void cleanup();
 
-  GLFWwindow *glfw_window() const;
+  TaichiWindow *taichi_window() const;
 
   taichi::lang::vulkan::VulkanDevice &device();
   const taichi::lang::vulkan::VulkanDevice &device() const;
@@ -30,7 +39,7 @@ class AppContext {
   // not owned
   taichi::lang::vulkan::VulkanDevice *vulkan_device_{nullptr};
 
-  GLFWwindow *glfw_window_{nullptr};
+  TaichiWindow *taichi_window_{nullptr};
 };
 
 }  // namespace vulkan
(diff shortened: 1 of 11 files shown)


ghuau-innopeak:
This is an initial draft to allow using GGUI on Android, I'm not an entire fan of the changes in GGUI to run with Android but this is the simple transition I could do without impacting too much the code right now, the main issue is that GLFW code is a bit everywhere :).
Also, I'm still not 100% sure of the best solution for handling the `DeviceAllocation`, right now, I just added a new field in the struct as we cannot use the `snode` as it will require a program.

Here is a quick example on how to use it for Android as well: https://github.com/ghuau-innopeak/taichi-aot-tests/blob/main/android/app/src/main/cpp/main.cpp (it's test code... :))

AmesingFlank:
This is amazing! 
Regarding the device allocation, I'm happy with the current solution, but it's best if we add a comment to clarify when where the `DeviceAllocation` be used and when will the `snode` be used. (also, nit, I'd rename `dalloc` to `dev_alloc`)

taichi/ui/backends/vulkan/renderable.cpp
```diff
@@ -144,7 +154,7 @@ void Renderable::create_graphics_pipeline() {
 void Renderable::create_vertex_buffer() {
   size_t buffer_size = sizeof(Vertex) * config_.max_vertices_count;
 
-  Device::AllocParams vb_params{buffer_size, false, false,
+  Device::AllocParams vb_params{buffer_size, true, true,
```
AmesingFlank, line 157: May I ask why this is needed for Android? 
If this is really necessary, then we should guard it with `#if ANDROID` so that it doesn't hurt the performance of GGUI on other platforms.

taichi/ui/backends/vulkan/app_context.h
```diff
@@ -6,16 +6,23 @@
 #include "taichi/backends/vulkan/vulkan_device.h"
 #include "taichi/ui/backends/vulkan/swap_chain.h"
 
+#ifdef ANDROID
+#include <android/native_window.h>
+typedef ANativeWindow TaichiWindow;
+#else
+typedef GLFWwindow TaichiWindow;
+#endif
```
AmesingFlank, line 14: tiny nit, let's use `using A = B` instead of `typedef B A`, and maybe put `TaichiWindow` inside `taichi::ui`?

taichi/ui/backends/vulkan/renderable.cpp
ghuau-innopeak: This is not needed, this is something I forget to revert during my testing... good catch! :D, it should be fixed.

taichi/ui/backends/vulkan/app_context.h
ghuau-innopeak: Done! My habits of C developer :), I left the `#include` outside as well to avoid having all the functions from this header inside the Taichi namespace, I think I could have used `namespace { #include <android/native_window.h> }` instead to avoid the extra `ifdef ANDROID` but not sure what is preferred here.

ghuau-innopeak:
@AmesingFlank Perfect :), I added a comment in the struct to clarify the usage of both fields.

taichi/ui/backends/vulkan/app_context.h
AmesingFlank: I think this is fine.

AmesingFlank (approved):
LGTM! Looking forward to seeing a demo on Android :D

bobcao3 (approved):
LGTM

PR #3848, opened 2021-12-21
[opengl] Allocate new arg bufs per kernel launch

ailzhang:
When reusing one args_buf for different gl kernels we spend a lot of
time on glMapBufferRange() in syncing. But these syncs are indeed not
necessary. Now we just create one args_buf per kernel launch.
This improve mpm88 benchmark from 16.67fps to 33fps.

Thanks to @k-ye and @bobcao3 for helping!

Will try to clean this up (some logic can be shared with vulkan runtime iiuc) in followup PRs. 

Related issue = #

### DIFF
--- a/taichi/backends/opengl/opengl_api.cpp
+++ b/taichi/backends/opengl/opengl_api.cpp
@@ -367,6 +367,15 @@ void DeviceCompiledTaichiKernel::launch(RuntimeContext &ctx,
   // If we have external array args we'll have to do host-device memcpy.
   // Whether we get external array arg is runtime information.
   bool has_ext_arr = false;
+  bool synced = false;
+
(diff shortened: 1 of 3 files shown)


ailzhang:
Note to self: I started with a version without saving allocated buffer in OpenGLRuntime and caused some random `nan`s in tests. This is due to deallocating arg_buf way too aggressively (at the end of kernel launch) so sometimes when we try to copy data back from device to host, the buffer on device was already freed. (I was dumb...)

ailzhang:
FYI the reason that I didn't enforce a max/fixed length for saved_arg_bufs here as it might trigger another nan issue that I'd like to fix in a separate PR.

k-ye (approved):

PR #3858, opened 2021-12-23
[lang] Cleanup parallel sort utility

AmesingFlank:
This PR moves the `parallel_sort` function to a different place, so that we don't use `import taichi as ti`. It also moves the kernel definition outside the python function, so that the kernel isn't recompiled everytime the function is called. 

On a tangential note, this sort function is kinda slow. On my macbook, sorting the particles in mpm99 is so slow that the simulation goes down to around 1.2 FPS. This is quite bad, I might try to implement a fast radix sort.


qiao-bo (approved):
LGTM!

ailzhang:
btw I feel it's okay to do `import taichi as ti` in tools :P Is it the only motivation for this change? ;) 
`import taichi as ti` in taichi/lang might be problematic :P

AmesingFlank:
> btw I feel it's okay to do `import taichi as ti` in tools :P Is it the only motivation for this change? ;) `import taichi as ti` in taichi/lang might be problematic :P

In my original code in `tools`, the `@ti.kernel` was inside a python function, so the `kernel` decorator is not used until the function is called. As soon as I move the kernel definition to global scope, I started getting error messages that `kernel` doesn't exists. It is necessary to move the kernel definition out, because otherwise the kernel recompiles every time `parallel_sort` is called. So basically, putting a taichi kernel inside `tools` won't work.

PR #3862, opened 2021-12-23
[refactor] Add _MatrixFieldElement class

strongoier:
Related issue = #2590, #2880

This PR makes one use case of `_IntermediateMatrix` class, an element of a matrix field, more explicit. It can unblock the implementation of enabling dynamic indexing of matrix field elements.


g1n0st (approved):
LGTM, thanks!

Issue #3863, opened 2021-12-23
Provide stmt name for auto-diff related assert info


erizmr:
Provide stmt name for auto-diff related assert info for ```basic_verify```

```
TI_ASSERT_INFO(
          found,
          "IR broken: stmt {} cannot have operand {}."
          " If you are using autodiff, please check"
          " https://docs.taichi.graphics/lang/articles/advanced/"
          "differentiable_programming#kernel-simplicity-rule"
          " If it doesn't help, please report this bug by opening an issue at"
          " https://github.com/taichi-dev/taichi to help us improve."
          " Thanks in advance!",
          stmt->id, op->id);
    }
```

```
TI_ASSERT_INFO(stmt->parent == current_block_,
               "stmt({})->parent({}) != current_block({})", stmt->id,
               fmt::ptr(stmt->parent), fmt::ptr(current_block_));
```

PR #3864, opened 2021-12-23
[autodiff] Provide stmt name for auto-diff related assert info

erizmr:
Related issue = #3863

### DIFF
--- a/taichi/analysis/verify.cpp
+++ b/taichi/analysis/verify.cpp
@@ -47,14 +47,14 @@ class IRVerifier : public BasicStmtVisitor {
       }
       TI_ASSERT_INFO(
           found,
-          "IR broken: stmt {} cannot have operand {}."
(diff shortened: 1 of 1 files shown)


taichi/analysis/verify.cpp
```diff
@@ -54,7 +54,7 @@ class IRVerifier : public BasicStmtVisitor {
           " If it doesn't help, please report this bug by opening an issue at"
           " https://github.com/taichi-dev/taichi to help us improve."
           " Thanks in advance!",
-          stmt->id, op->id);
```
ailzhang, line 57: ah I feel like stmt->id might still be useful in some cases. Shall we do `{} {} cannot..., stmt->type(), stmt->id()` instead ?

erizmr: Sure. Let me do that.

ailzhang (approved):
Thanks!

PR #3865, opened 2021-12-23
[Lang] Enable dynamic indexing of matrix field elements when possible

strongoier:
Related issue = #2590

After this PR, dynamic indexing of matrix field elements will be enabled when `dynamic_index=True` is set during `ti.init()` and `dynamic_index_stride` is successfully calculated (#3854).

To achieve that and keep code simple, this PR does a lot of refactoring:
- Unify `local_subscript_with_offset` and `global_subscript_with_offset` into `make_tensor_element_expr`;
- Pass `dynamic_index_stride` into `TensorElementExpression` instead of calculating stride inside `TensorElementExpression`, which is not feasible now.


k-ye (approved):
LGTM!

PR #3880, opened 2021-12-25
[dx11] Add DX11 device interface definition

quadpixels:
Related issue = #992

This is the first part in an attempt to break down and refactor PR #3776.

During the refactor, a few things have changed (example: I would want to add a class to read information form the DX11 info queue for some functional testing), but it seems no matter what the refactor looks like, this interface boilerplate would be the same. So creating a pull request here.


bobcao3 (approved):
LGTM!

python/taichi/lang/__init__.py
```diff
@@ -16,6 +16,7 @@
 import taichi.lang.linalg_impl
 import taichi.lang.meta
 from taichi._lib import core as _ti_core
+print(_ti_core.__file__)
```
ailzhang, line 19: not necessary?

ailzhang (changes requested):
Great work!
nit: please remove the additional print before merging :D

python/taichi/lang/__init__.py
quadpixels: This was added for resolving an installation issue, it is now removed.

quadpixels:
> Great work! nit: please remove the additional print before merging :D

Thanks for the comments!
Removed the print statement in the new version of the pull request.

ailzhang (approved):
Thanks!

PR #3883, opened 2021-12-27
[opengl] Reduce repeated read to args buffer.

ailzhang:
* #3884
* __->__ #3883

Note this change doesn't show any improvemance on desktop (likely due to
caching), but it improves perf on android (a snapdragon device which
might not have L1 cache) from 100ms per
frame to 75ms per frame.  Thanks to @Leonz5288 for profiling!


k-ye (approved):
LGTM!

PR #3884, opened 2021-12-27
[opengl] Optimize range_for for ndarrays

ailzhang:
* #3901
* #3900
* __->__ #3884

Note this PR reduces number of generated shaders for mpm88 from 10 to 6.
There's still one gtmp related shader remaining for temporary values
used across multiple shaders. Whether to further remove that shader need
more benchmark so let's just get rid of addtional shaders introduced by
ndarray range_for for now.

### DIFF
--- a/taichi/ir/statements.h
+++ b/taichi/ir/statements.h
@@ -378,6 +378,10 @@ class ExternalTensorShapeAlongAxisStmt : public Stmt {
 
   ExternalTensorShapeAlongAxisStmt(int axis, int arg_id);
 
+  bool has_global_side_effect() const override {
+    return false;
+  }
+
   TI_STMT_DEF_FIELDS(ret_type, axis, arg_id);
   TI_DEFINE_ACCEPT_AND_CLONE
 };
@@ -729,6 +733,7 @@ class RangeForStmt : public Stmt {
   int num_cpu_threads;
   int block_dim;
   bool strictly_serialized;
+  bool range_of_array{false};
 
   RangeForStmt(Stmt *begin,
                Stmt *end,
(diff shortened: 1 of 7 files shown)


taichi/ir/statements.h
```diff
@@ -756,7 +762,8 @@ class RangeForStmt : public Stmt {
                      bit_vectorize,
                      num_cpu_threads,
                      block_dim,
-                     strictly_serialized);
+                     strictly_serialized,
+                     range_of_array);
```
bobcao3, line 766: This parameter name seems slightly confusing?

taichi/transforms/offload.cpp
```diff
@@ -85,19 +85,35 @@ class Offloader {
         } else {
           offloaded->block_dim = s->block_dim;
         }
-        if (auto val = s->begin->cast<ConstStmt>()) {
+
+        // TODO: We need to update codegen for each backend gradually so let's
```
bobcao3, line 89: When we merge this in let's also notify people working on the vulkan ndarray this change

tests/python/test_aot.py
```diff
@@ -202,7 +202,7 @@ def init(d: ti.i32, density1: ti.any_arr(), density2: ti.any_arr(),
 
 @ti.test(arch=ti.opengl)
 def test_opengl_exceed_max_ssbo():
-    # 7 ndarrays + gtmp + args > 8 (maximum allowed)
+    # 8 ndarrays + args > 8 (maximum allowed)
```
bobcao3, line 205: Do we really want this test? This seems a bit arch specific & a lot of devices support more

bobcao3 (approved):
LGTM other than a few minor nitpicks

tests/python/test_aot.py
ailzhang: Yea this test was mainly a self reminder to double check how many ssbos we create in normal cases - we can remove it later :D

PR #3897, opened 2021-12-28
[autodiff] Restrict Independent Block scope for cases with atomic operations on global variables

erizmr:
Related issue = #3912


ailzhang (commented):
Awesome!

ailzhang (approved):
Thanks!!

PR #3899, opened 2021-12-28
[Test] [aot] Add initial AOT CPP test (#3850)

ghuau-innopeak:
Only test the basic feature as the API is still under development. Just
make sure we can serialize and de-serialize the data properly for the
kernels.

Related issue = #3642


ailzhang (approved):
LGTM!

PR #3901, opened 2021-12-29
[opengl] Cache loaded arr elements.

ailzhang:
* __->__ #3901
* #3900
* #3884

WIP: implementation is a bit hacky, for testing only...

Noticed that we're reloading the same elements in ndarrays again and
again in the opengl generated code. This doesn't affect JIT perf on NV
GPUs but I'm suspicious that it's one source of bottleneck on snapdragon
android.

Unlike read-only shape information saved in args buffer, cached ndarray
elements should be invalidated once it's overwritten.


ailzhang:
superseded by #3992

PR #3911, opened 2021-12-30
[Refactor] Remove all occurrences of print_preprocessed and print_ast

### LINKED ISSUE #3748: Remove `print_preprocessed` which is no longer used.
After the ast generator refactor `print_preprocessed` is no longer used(replaced by `print_preprocessed_ir`). We should remove it as well as `print_ast` here. https://github.com/taichi-dev/taichi/blob/master/python/taichi/lang/ast/ast_transformer_utils.py#L199-L204 
This task should be as simple as removing all of `print_preprocessed` and `print_ast`s. 
cc: @lin-hitonami

mzmzm:
Related issue = close #3748


strongoier (approved):
LGTM!

PR #3919, opened 2021-12-30
Add example of color_edit_3

Vineyo:
It seems this is what supposed to be.


k-ye (approved):
LGTM! Thanks for the fix

AmesingFlank (approved):
nice catch. Thanks!

PR #3921, opened 2021-12-31
[refactor] Optimize vector and matrix ndarray fill

qiao-bo:
Speedup ndarray `fill` using `cuMemset32D` and `std::fill` instead of the kernel-based approach. Measured speedups (using the test script from [PR 3907](https://github.com/taichi-dev/taichi/pull/3907):

`ti.Matrix.ndarray(4, 4, ti.f32, N)`

1. `ti.init(arch=ti.cuda) `

| `iterations` | Master | Head |
| ------------- | ------------- | ------------- |
| 10  | 0.457 s| 0.050 ms  |
| 100  | 4.705 s | 0.475 ms |
| 1000  |  65.971 s | 4.374 ms |

2. `ti.init(arch=ti.cpu) `

| `iterations` | Master | Head |
| ------------- | ------------- | ------------- |
| 10  | 0.407 s| 0.150 s  |
| 100  | 4.144 s | 1.488 s |
| 1000  | 42.127 s  | 14.693 s |

### DIFF
--- a/python/taichi/lang/_ndarray.py
+++ b/python/taichi/lang/_ndarray.py
@@ -83,7 +83,9 @@ class Ndarray:
         """
         if impl.current_cfg().ndarray_use_torch:
             self.arr.fill_(val)
-        elif impl.current_cfg().arch != _ti_core.Arch.cuda and impl.current_cfg().arch != _ti_core.Arch.x64:
+        elif impl.current_cfg(
+        ).arch != _ti_core.Arch.cuda and impl.current_cfg(
+        ).arch != _ti_core.Arch.x64:
             self.fill_by_kernel(val)
         elif self.dtype == primitive_types.f32:
             self.arr.fill_float(val)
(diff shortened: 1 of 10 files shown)


python/taichi/lang/_ndarray.py
```diff
@@ -280,6 +269,9 @@ def __deepcopy__(self, memo=None):
         ret_arr.copy_from(self)
         return ret_arr
 
+    def buffer_fill(self, val):
```
strongoier, line 272:
```suggestion
    def fill_by_kernel(self, val):
```

python/taichi/lang/matrix.py
```diff
@@ -1350,7 +1350,7 @@ def __getitem__(self, key):
 
     @python_scope
     def fill(self, val):
-        self.ndarray_matrix_fill(val)
+        self.ndarray_fill(val)
```
strongoier, line 1353: What about removing this and renaming `ndarray_fill()` of the base `Ndarray` class to `fill()`?

strongoier (changes requested):

strongoier (approved):
LGTM!

Issue #3915, opened 2021-12-30
Taichi Python Runtime Overhead Optimization (Ndarray)


qiao-bo:
Currently, Taichi experiences an noticeable amount of python overhead when executing light kernels using `Field` or `Ndarray`. Using cProfile and/or pyinstrument, we can identify a few hot spots that incur this python runtime overhead. Here is an example using pyinstrument:

```
import taichi as ti
from pyinstrument import Profiler
from pyinstrument.renderers import ConsoleRenderer

ti.init(arch=ti.cuda)

N = 4096
a = ti.ndarray(ti.i32, N)

@ti.kernel
def fill_ones(a: ti.any_arr()):
    for i in a:
        a[i] = 1

def repeat_execution(n):
    for _ in range(n):
        fill_ones(a)

# initial run
repeat_execution(100)

# start profiling
profiler = Profiler()
profiler.start()
repeat_execution(10000)
session = profiler.stop()
profile_renderer = ConsoleRenderer(unicode=True, color=True, show_all=True)
print(profile_renderer.render(session))
```
On my PC with rtx 3080+i9 11900K, the output is
```
[Taichi] version 0.8.9, llvm 10.0.0, commit 345f0f04, linux, python 3.9.7
[Taichi] Starting on arch=cuda
[I 12/23/21 14:57:27.289 296821] [llvm_program.cpp:LlvmProgramImpl@84] CUDA max blocks per SM = 16

  _     ._   __/__   _ _  _  _ _/_   Recorded: 14:57:27  Samples:  205
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.205     CPU time: 0.206
/   _/                      v4.1.1

Program: test.py

0.205 <module>  test1.py:1
└─ 0.205 repeat_execution  test1.py:16
   └─ 0.203 wrapped  taichi/lang/kernel_impl.py:712
      ├─ 0.198 __call__  taichi/lang/kernel_impl.py:634
      │  ├─ 0.134 func__  taichi/lang/kernel_impl.py:485
      │  │  ├─ 0.123 [self]  
      │  │  ├─ 0.005 current_cfg  taichi/lang/impl.py:906
      │  │  │  └─ 0.005 PyCapsule.current_compile_config  <built-in>:0
      │  │  └─ 0.003 match_ext_arr  taichi/lang/kernel_impl.py:619
      │  ├─ 0.056 ensure_compiled  taichi/lang/kernel_impl.py:626
[...]

PR #3922, opened 2021-12-31
[refactor] [perf] Reduce python overhead for frequent lightweight kernel launches by caching contexts

turbo0628:
Related issue = #3915 

This pull request is an ad-hoc patch to the performance issue in #3915. It caches the kernel launch contexts and keys in order to reduce python overhead for very frequent small kernel launches.

The runtime of the micro-benchmark program on my CPU is as follows:
0.8.8 (7bae9c77): 1.516s
HEAD: 0.994s

Briefings of modifications:
1) Extract the body of `func__` into a standalone function.
2) Use the args.__hash__() as its signature, add caches for launch_ctx and keys.
 
Potential issues and concerns:
1) Hope the modifications won't break other applications.
2) A toggle is absent -- It's just that I didn't figure out where to put it.
3) (Possibly) more memory usage.

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -350,6 +350,8 @@ class Kernel:
         self.grad = None
         self.argument_annotations = []
         self.argument_names = []
+        self.argument_launch_context_cache = {}
+        self.argument_key_cache = {}
         self.return_type = None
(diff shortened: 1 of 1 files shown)


turbo0628:
This solution is deprecated as the signature method is wrong.

Issue #3608, opened 2021-11-24
Internal research project: ti.Mesh


g1n0st:
We are developing an internal research project ti.Mesh (#3567). It offers first-class support for triangular/tetrahedral meshes and allows efficient computation on these irregular data structures. Now it's an early-stage prototype in progress, so we keep it undiscussed. We will soon release more details and welcome contributions from the community. : )

yang-si-hang:
I want to generate a Tetrahedral Mesh from vertices.
I don't know how to use the Function "taichi.TetMesh()". Have you accomplished this Function?
Thanks!

g1n0st:
@yang-si-hang  Thank you for being so interested in our work! The work about `ti.Mesh` has recently been accepted by SIGGRAPH Asia 2022, and we are refactoring APIs and example cases. We will release complete examples at https://github.com/taichi-dev/meshtaichi and provide documentation in the following days.

yang-si-hang:
> @yang-si-hang Thank you for being so interested in our work! The work about `ti.Mesh` has recently been accepted by SIGGRAPH Asia 2022, and we are refactoring APIs and example cases. We will release complete examples at https://github.com/taichi-dev/meshtaichi and provide documentation in the following days.

Thanks for your reply! I look forward to reading your paper.

g1n0st:
Since the internal research project ti.Mesh will make it public soon, so I will close this issue. New problems related to MeshTaichi could open with a new thread or move to https://github.com/taichi-dev/meshtaichi. Any discussions and contributions from the community are welcomed!

PR #3923, opened 2021-12-31
[mesh] Demote from-end element attribute atomic op

g1n0st:
Related issue = #3608 

It is worth noting that in a mesh-for loop, the attribute of the from-end element should only be accessed in one loop, like the following case:
``` Python
for c in model.cells:
    for i in c.cells.size:
        c.a += c.cells[i].b
```
The variable `c.a` is a unique address in the whole loop scope so the atomic operation `+=` should be demoted.
Note that we ignore the following case since it's not reasonable for a parallel program.
 ``` Python
for c in model.cells:
    for i in c.cells.size:
        c.a += c.cells[i].a
```

### DIFF
--- a/taichi/analysis/gather_uniquely_accessed_pointers.cpp
+++ b/taichi/analysis/gather_uniquely_accessed_pointers.cpp
@@ -130,26 +130,29 @@ class UniquelyAccessedSNodeSearcher : public BasicStmtVisitor {
 
   void visit(GlobalPtrStmt *stmt) override {
     // mesh-for loop unique
-    if (stmt->indices.size() == 1 && stmt->indices[0]->is<MeshIndexConversionStmt>()) {
+    if (stmt->indices.size() == 1 &&
+        stmt->indices[0]->is<MeshIndexConversionStmt>()) {
       auto idx = stmt->indices[0]->as<MeshIndexConversionStmt>()->idx;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by BillXu2000** [comment]
``` python
@ti.kernel
def fun():
    for u in bunny.verts:
        u.s += 1
        for i in range(u.verts.size):
            v = u.verts[i]
            v.s += 1
```
In this case, none of these atomic adds should be demoted, but this pr demote the first one.
[gather_uniquely_accessed_pointers.cpp](https://github.com/taichi-dev/taichi/blob/master/taichi/analysis/gather_uniquely_accessed_pointers.cpp) might be helpful to solve this issue.

g1n0st:
> ```python
> @ti.kernel
> def fun():
>     for u in bunny.verts:
>         u.s += 1
>         for i in range(u.verts.size):
>             v = u.verts[i]
>             v.s += 1
> ```
> 
> In this case, none of these atomic adds should be demoted, but this pr demote the first atomic add. [gather_uniquely_accessed_pointers.cpp](https://github.com/taichi-dev/taichi/blob/master/taichi/analysis/gather_uniquely_accessed_pointers.cpp) might be helpful to solve this issue.

Fixed.

ailzhang (approved):
LGTM!

PR #3925, opened 2022-01-01
[opt] [ir] [refactor] Remove exceptions from offload pass

mzmzm:
Related issue = #1059

### DIFF
--- a/taichi/transforms/offload.cpp
+++ b/taichi/transforms/offload.cpp
@@ -464,7 +464,6 @@ class PromoteIntermediateToGlobalTmp : public BasicStmtVisitor {
       auto ptr = stmt->insert_after_me(
           Stmt::make<GlobalTemporaryStmt>(offset, stmt->ret_type));
       ptr->insert_after_me(Stmt::make<GlobalStoreStmt>(ptr, stmt));
-      throw IRModified();
     }
   }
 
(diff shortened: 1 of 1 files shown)


strongoier (changes requested):
It seems that your change doesn't touch places catching `IRModified`: https://github.com/taichi-dev/taichi/blob/51f557ee3d5a8943e2aaae9b6e34e419827f43e1/taichi/transforms/offload.cpp#L696

strongoier (changes requested):
There's still one place catching `IRModified` :-)

taichi/transforms/offload.cpp
```diff
@@ -653,24 +646,22 @@ class FixCrossOffloadReferences : public BasicStmtVisitor {
       } else {
         // For other cases like ArgLoadStmt UnaryOpStmt which needs to load.
         auto load = Stmt::make<GlobalLoadStmt>(global_temporary.get());
+        auto pload = load.get();
         stmt_to_offloaded_[load.get()] = offloaded;
         stmt->set_operand(index, load.get());
         stmt->insert_before_me(std::move(global_temporary));
         stmt->insert_before_me(std::move(load));
+        generic_visit(pload);
```
strongoier, line 654: IIUC a global load from a global temporary doesn't need to be visited again, because no other stmts are involved.

strongoier (changes requested):

strongoier (approved):
LGTM!

PR #3940, opened 2022-01-05
[Error] Do not show body in exceptions in nodes with body

### LINKED ISSUE #3784: Error message too long when error occurs in For/FunctionDef nodes
**Describe the bug**
When an error occurs in a node with `body` (for example, For, If, FunctionDef nodes), the error will contain the body as well, which is not necessary. 

**To Reproduce**

```py
@ti.kernel
def foo():
    for i in range(1, 2, 3):
        a = 1
        b = 1
        c = 1
        d = 1
        e = 1
```

**Log/Screenshots**
```
taichi.lang.exception.TaichiCompilationError: On line 5 of file "/home/lin/tmp/test.py":
    for i in range(1, 2, 3):
    ^^^^^^^^^^^^^^^^^^^^^^^^
        a = 1
        ^^^^^
        b = 1
        ^^^^^
        c = 1
        ^^^^^
        d = 1
        ^^^^^
        e = 1
        ^^^^^
TaichiSyntaxError: Range should have 1 or 2 arguments, found 3
```

lin-hitonami:
Related issue = fixes #3784 
`max(node.body[0].lineno - 1, node.lineno)` is to deal with circumstances when `node.body[0]` starts on the same line as `node` and ends on another line. It makes sure at least one line can be printed. For example:
```python
for i in range(1, 2, 3):print(
i)
```

### DIFF
--- a/python/taichi/lang/ast/ast_transformer_utils.py
+++ b/python/taichi/lang/ast/ast_transformer_utils.py
@@ -201,7 +201,14 @@ class ASTTransformerContext:
             hint = ' ' * col_offset + '^' * (end_col_offset - col_offset)
             msg += gen_line(self.src[node.lineno - 1], hint)
         else:
-            for i in range(node.lineno - 1, node.end_lineno):
+            node_type = node.__class__.__name__
+
+            if node_type in ["For", "While", "FunctionDef", "If"]:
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Nice!

PR #3945, opened 2022-01-05
[misc] [cuda] Set the toolkit used by KernelProfiler at runtime

yolo2themoon:
Related issue = #2922


python/taichi/profiler/kernelprofiler.py
```diff
@@ -72,6 +73,19 @@ def get_kernel_profiler_mode(self):
         """Get status of :class:`~taichi.profiler.kernelprofiler.KernelProfiler`."""
         return self._profiling_mode
 
+    def set_toolkit(self, toolkit_name='default'):
```
qiao-bo, line 76: Can we replace all the `default` string to `event`? also in the above `set_kernel_profiler_toolkit` function. It is clear the argument has default value, but not quite clear what `default` actually is :) WDYT?

qiao-bo (commented):
Thanks!

python/taichi/profiler/kernelprofiler.py
yolo2themoon: The idea here is that backends have different `'default'` toolkits. 
CUDA has `cuEvent`, while for OpenGL and Vulkan, we will use the `timestamp` API. 
If we want to use profiling toolkits other than 'default', we need to specify the name (now we have `'cupti'` ).

qiao-bo (approved):
LGTM!

PR #3946, opened 2022-01-05
[refactor] Remove ndarray_use_torch from pybind

qiao-bo:
Avoid query from pybind to save some host overhead.

### DIFF
--- a/python/taichi/lang/__init__.py
+++ b/python/taichi/lang/__init__.py
@@ -398,7 +398,7 @@ class _SpecialConfig:
         self.gdb_trigger = False
         self.experimental_real_function = False
         self.short_circuit_operators = False
-        self.ndarray_use_torch = False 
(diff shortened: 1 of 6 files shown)


python/taichi/lang/__init__.py
```diff
@@ -398,6 +398,7 @@ def __init__(self):
         self.gdb_trigger = False
         self.experimental_real_function = False
         self.short_circuit_operators = False
+        self.ndarray_use_torch = False
```
ailzhang, line 401: Hmm this is not the best place for ndarray_use_torch, since it's going away soon I'm fine with this as a temporary location. :D

ailzhang (approved):
LGTM!

python/taichi/lang/__init__.py
qiao-bo: ok~

PR #3947, opened 2022-01-05
[perf] [refactor] Reduce kernel launch context construction overhead

turbo0628:
Related issue = #3915 

This PR shortens the kernel context building overhead in the `func__` body.

Briefings of code changes:

1) Add new `set_arg_external_array_w_shape` method for `launch_ctx` in order to diminish a small `shape` loop in Python scope.
2) Add shortcut code path for Ndarray so that we can reduce the isinstance overhead
3) move the ndarray_use_torch out of the loop, just follows the runtime materialization -- This change may conflict with #3946 

Effect:
For CPU backends on my laptop, the overhead of 10000 launches with 6 ndarray reduces from 2.841s to 1.196s. However, it cannot be such significant when the kernel run time is at milliseconds.


taichi/program/kernel.h
```diff
@@ -41,6 +41,13 @@ class Kernel : public Callable {
                                 uint64 size,
                                 bool is_device_allocation);
 
+    void set_arg_external_array_w_shape(int arg_id,
+                                        uint64 ptr,
```
k-ye, line 45:
```suggestion
                                        uintptr_t ptr,
```

taichi/program/kernel.h
```diff
@@ -41,6 +41,13 @@ class Kernel : public Callable {
...
+                                        uint64 elem_count,
+                                        uint64 elem_size,
+                                        std::vector<int64> shape,
```
k-ye, line 48:
```suggestion
                                        const std::vector<int64> &shape,
```

taichi/program/kernel.h
```diff
@@ -41,6 +41,13 @@ class Kernel : public Callable {
                                 uint64 size,
                                 bool is_device_allocation);
 
+    void set_arg_external_array_w_shape(int arg_id,
```
k-ye, line 44: nit: suggest to either use `with_` or just remove `w_` (because shape is already one of the params) :-)

taichi/program/kernel.h
turbo0628: I used a consistent interface with `set_arg_external_array`, which takes `uint64` as type identifier for ptr. Shall we change them all?

k-ye: SG, thx!

taichi/program/kernel.h
turbo0628: Thank you! I'll change the name in a moment :)

taichi/program/kernel.cpp
```diff
@@ -266,6 +266,22 @@ void Kernel::LaunchContextBuilder::set_arg_external_array(
   ctx_->set_device_allocation(arg_id, is_device_allocation);
 }
 
+void Kernel::LaunchContextBuilder::set_arg_external_array_w_shape(
+    int arg_id,
+    uint64 ptr,
+    uint64 elem_size,
```
ailzhang, line 272: can we pass in elem_size * elem_count directly?

ailzhang (approved):
LGTM!

python/taichi/lang/kernel_impl.py
```diff
@@ -489,6 +490,17 @@ def func__(*args):
                 if isinstance(needed, template):
                     continue
                 provided = type(v)
+                # Shortened code path for ndarray.
```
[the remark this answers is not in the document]
strongoier: I'm also curious here. IMHO this modification might make people confused when reading this part of code :-)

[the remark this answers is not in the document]
turbo0628: Yeah that's for sure. I tried to refactor the conditions but got stuck in tests, there were come complex logics that should be simplified. Shall we open another issue to track this?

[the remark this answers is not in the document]
strongoier: Sounds good.

taichi/program/kernel.cpp
turbo0628: I pass them separately for the reason that multiplication in python scope has noticeable overhead. 
The overhead would be accumulated to significance in the case of frequent kernel launch, so I made a slightly different interface. 

Ideally I should pass the "Ndarray" object directly to the C++ function in order to eliminate the extra three or four pybind invokes (`get_data_ptr``get_nelement``get_element_size` are just pybind calls). However, I encountered a pybind problem of reference passing from the python world is still a little verbose and painful. I'll leave it here if I can not fix this at noon.

taichi/program/kernel.h
turbo0628: Resolved :)

taichi/program/kernel.h
turbo0628: I finally keep the `with_` as it reads better.

python/taichi/lang/kernel_impl.py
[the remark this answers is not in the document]
turbo0628: About perf gain, this shortcut will not evaluate `match_ext_arr` and other `isinstance`, so the effect is observable. After all this function body is critical to launch overhead, we should try best to be efficient.

About readability, the first `if` clause in this loop takes a shortcut convention, while the later clauses are nested `if-elif-else`. So I think extract out the Ndarray in a shortcut won't cause confusion. However there's repeated logics later in the clausees, this might be confusing. I'd like to clean this up in anthoer PR.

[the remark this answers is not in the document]
strongoier: Regarding the logic I think `isinstance(needed, any_arr)` is needed for `Ndarray`s.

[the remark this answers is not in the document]
turbo0628: I see, have amended this issue!

taichi/program/kernel.cpp
turbo0628: Solved by 1) pass in elem_size * elem_count directly.
 2) add set function for Ndarray to further reduce overhead.

taichi/program/kernel.cpp
```diff
@@ -266,6 +266,33 @@ void Kernel::LaunchContextBuilder::set_arg_external_array(
   ctx_->set_device_allocation(arg_id, is_device_allocation);
 }
 
+void Kernel::LaunchContextBuilder::set_arg_external_array_with_shape(
+    int arg_id,
+    uintptr_t ptr,
+    uint64 size,
+    const std::vector<int64> &shape,
+    bool is_device_allocation) {
```
ailzhang, line 274: Actually if we are going to separate external array and ndarray, is_device_allocation is always false for external array. So we can safely set it to false and remove this arg

ailzhang (approved):
Awesome thanks!

taichi/program/kernel.h
```diff
@@ -37,10 +38,18 @@ class Kernel : public Callable {
...
                                 uint64 size,
                                 bool is_device_allocation);
 
+    void set_arg_external_array_with_shape(int arg_id,
+                                           uintptr_t ptr,
+                                           uint64 size,
+                                           const std::vector<int64> &shape,
+                                           bool is_device_allocation);
+
+    void set_arg_external_ndarray(int arg_id, const Ndarray &arr);
```
ailzhang, line 51: oh sorry one more nitpicking: let's name this `set_arg_ndarray` since there no internal/external concept ndarrays ;)

turbo0628: Done :)

taichi/program/kernel.cpp
turbo0628: Good, didn't know that this flag is specific to Ndarray. Then this function looks neat.

PR #3949, opened 2022-01-06
[refactor] Get global vars by using __globals__

### LINKED ISSUE #3944: Get global variables of taichi kernels and functions using __globals__
We currently use `inspect.getclosurevars` to get global variables used in taichi kernels and functions. It can minimize the size of the global variable dict. However, it has many limitations. As mentioned in #3709, it ignores the variables in the annotations. It also ignores variables in listcomps and dictcomps. 

We can change back to the method in #282 to solve the problem.

lin-hitonami:
Related issue = close #3944

### DIFF
--- a/python/taichi/lang/ast/ast_transformer_utils.py
+++ b/python/taichi/lang/ast/ast_transformer_utils.py
@@ -178,9 +178,8 @@ class ASTTransformerContext:
         for s in reversed(self.local_scopes):
             if name in s:
                 return s[name]
-        var = self.global_vars.get(name)
-        if var is not None:
-            return var
+        if name in self.global_vars:
+            return self.global_vars[name]
         return getattr(builtins, name, None)
 
     def get_pos_info(self, node):
--- a/python/taichi/lang/ast/ast_transformer_utils.py
+++ b/python/taichi/lang/ast/ast_transformer_utils.py
@@ -1,4 +1,5 @@
 import ast
+import builtins
 import traceback
 from enum import Enum
 from sys import version_info
@@ -177,7 +178,10 @@ class ASTTransformerContext:
         for s in reversed(self.local_scopes):
             if name in s:
                 return s[name]
-        return self.global_vars.get(name)
+        var = self.global_vars.get(name)
+        if var:
+            return var
+        return getattr(builtins, name, None)
 
     def get_pos_info(self, node):
         msg = f'On line {node.lineno + self.lineno_offset} of file "{self.file}":\n'
--- a/python/taichi/lang/ast/ast_transformer_utils.py
+++ b/python/taichi/lang/ast/ast_transformer_utils.py
@@ -179,7 +179,7 @@ class ASTTransformerContext:
             if name in s:
                 return s[name]
         var = self.global_vars.get(name)
-        if var:
+        if var is not None:
             return var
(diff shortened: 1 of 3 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -331,12 +332,17 @@ def __init__(self, pos, needed, provided):
...
 
 def _get_global_vars(_func):
-    closure_vars = inspect.getclosurevars(_func)
-    return {
-        **closure_vars.globals,
-        **closure_vars.nonlocals,
-        **closure_vars.builtins
-    }
+    # Discussions: https://github.com/taichi-dev/taichi/issues/282
+    global_vars = copy.copy(_func.__globals__)
```
yihong0618, line 336: for sollow copy 
```python
d = {}
c = d.copy() is work I think.

# so
global_vars = _func.__globals__.copy() # maybe OK?

```

python/taichi/lang/ast/ast_transformer_utils.py
```diff
@@ -177,7 +178,10 @@ def get_var_by_name(self, name):
         for s in reversed(self.local_scopes):
             if name in s:
                 return s[name]
-        return self.global_vars.get(name)
+        var = self.global_vars.get(name)
+        if var:
+            return var
+        return getattr(builtins, name, None)
```
yihong0618, line 184:
```python
return var or getattr(builtins, name, None) # maybe better?
```

yihong0618 (approved):

python/taichi/lang/ast/ast_transformer_utils.py
lin-hitonami: This might be buggy if var equals to 0?

yihong0618: yes I did not see your second commit(^-^)

python/taichi/lang/ast/ast_transformer_utils.py
```diff
@@ -177,7 +178,10 @@ def get_var_by_name(self, name):
         for s in reversed(self.local_scopes):
             if name in s:
                 return s[name]
-        return self.global_vars.get(name)
+        var = self.global_vars.get(name)
+        if var is not None:
+            return var
+        return getattr(builtins, name, None)
```
frostming, line 184: Is there any case the value is `None`?

Maybe changed to:
```python
if name in self.global_vars:
    return self.global_vars[name]
return getattr(builtins, name, None)
```

Further, is returning `None` correct when name is not found?

yihong0618: > Is there any case the value is `None`?
> 
> Maybe changed to:
> 
> ```python
> if name in self.global_vars:
>     return self.global_vars[name]
> return getattr(builtins, name, None)
> ```
> 
> Further, is returning `None` correct when name is not found?

maybe?
```python
return self.global_vars.get(name, getattr(builtins, name, None))
```

lin-hitonami: @frostming It should raise error when nothing is found, and it will be done in #3951 which depends on this PR (this PR fixes the bug which causes the test of #3951 to fail).

strongoier (approved):
LGTM!

PR #3951, opened 2022-01-06
[Error] Raise exception when encountering undefined name

### LINKED ISSUE #3941: Better error reporting when a variable is not found
As mentioned in #3802, the error when a variable is not found is confusing, like
```
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

We should raise an error if ast transformer fails to find the variable.

lin-hitonami:
Related issue = close #3941


ailzhang (approved):
LGTM!

PR #3960, opened 2022-01-07
[refactor] [ci] Enable identifier naming in clang-tidy

qiao-bo:
Let's enable the `readability-identifier-naming` check to make use of the rules defined in `.clang-tidy`. Currently, checks only enabled for the `taichi/` directory.

Local checks and fixes can be applied with `python scripts/run_clang_tidy.py $PWD/taichi -clang-tidy-binary clang-tidy-10 -clang-apply-replacements-binary clang-apply-replacements-10 -checks=-*,performance-inefficient-string-concatenation,readability-identifier-naming -header-filter=$PWD/taichi -p $PWD/build -j4 -fix`

### DIFF
--- a/.github/workflows/scripts/check_clang_tidy.sh
+++ b/.github/workflows/scripts/check_clang_tidy.sh
@@ -9,4 +9,4 @@ rm -rf build && mkdir build && cd build
 cmake $CI_SETUP_CMAKE_ARGS ..
 
 cd ..
-python3 ./scripts/run_clang_tidy.py $PWD/taichi -clang-tidy-binary clang-tidy-10 -checks=-*,performance-inefficient-string-concatenation -header-filter=$PWD/taichi -p $PWD/build -j2
+python3 ./scripts/run_clang_tidy.py $PWD/taichi -clang-tidy-binary clang-tidy-10 -checks=-*,performance-inefficient-string-concatenation,readability-identifier-naming -header-filter=$PWD/taichi -p $PWD/build -j2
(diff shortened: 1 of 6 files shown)


.github/workflows/scripts/check_clang_tidy.sh
```diff
@@ -9,4 +9,4 @@ rm -rf build && mkdir build && cd build
 cmake $CI_SETUP_CMAKE_ARGS ..
 
 cd ..
-python3 ./scripts/run_clang_tidy.py $PWD/taichi -clang-tidy-binary clang-tidy-10 -checks=-*,performance-inefficient-string-concatenation -header-filter=$PWD/taichi -p $PWD/build -j2
+python3 ./scripts/run_clang_tidy.py $PWD/taichi -clang-tidy-binary clang-tidy-10 -checks=-*,performance-inefficient-string-concatenation,readability-identifier-naming -header-filter=$PWD/taichi -p $PWD/build -j2
```
ailzhang, line 12: OOC: is `=$PWD/taichi` necessary after the rule?

qiao-bo: clang-tidy checks the directory specified here. Our cpp files are mainly in this folder, right?

ailzhang (approved):
Thanks!

.github/workflows/scripts/check_clang_tidy.sh
ailzhang: sorry I was dumb :P

Issue #3938, opened 2022-01-05
[RFC] Standardizing Taichi Error Types


lin-hitonami:
# Motivation
In the taichi code base, there are too many kinds of error types like `KernelDefError`, `KernelArgError`, `InvalidOperationError`, `TaichiSyntaxError`, `RuntimeError`, `TypeError`, `ValueError`, and even `Exception`. Moreover, there is no standard on which error should be raised in which situation. So, we should standardize the error types.

# Goals
* Make rules on when should we use which error type.
* Let the error types inherit the corresponding python errors.
* Change the errors in the code base to comply with the rules.

# Proposal
* `TaichiCompilationError`
  * The base class of all errors generated in the compilation process.
  * Used when an error in the compilation process occurs that does not fit in other errors.
* `TaichiTypeError`
  * Inherits `TaichiCompilationError` and `TypeError`
  * Used when a type error is detected when performing type checks.
* `TaichiSyntaxError`
  * Inherits `TaichiCompilationError` and `SyntaxError`
  * Used when encountering a syntax that taichi does not support.
* `TaichiRuntimeError`
  * Used when a compiled taichi program fails to execute.

k-ye:
Looks great!

I suggest not to `TaichiTypeError` inherit from `TaichiCompilationError`. It is possible to produce a type error at compile time. E.g. your kernel takes in a `vec3`, but a `vec4` is passed in.

strongoier:
> Looks great!
> 
> I suggest not to `TaichiTypeError` inherit from `TaichiCompilationError`. It is possible to produce a type error at compile time. E.g. your kernel takes in a `vec3`, but a `vec4` is passed in.

I would consider this case as a `TaichiRuntimeError` because it doesn't belong to the Taichi compilation process. Instead, the situation is that the compiled program fails to execute.

lin-hitonami:
Maybe we can add `TaichiRuntimeTypeError` that inherits `TaichiRuntimeError` and `TypeError` for this case?

strongoier:
> Maybe we can add `TaichiRuntimeTypeError` that inherits `TaichiRuntimeError` and `TypeError` for this case?

Sounds good to me.

PR #3964, opened 2022-01-07
[Error] Add TaichiTypeError

lin-hitonami:
Related issue = #3938

### DIFF
--- a/python/taichi/lang/exception.py
+++ b/python/taichi/lang/exception.py
@@ -1,3 +1,6 @@
+from taichi._lib import core
+
+
 class TaichiCompilationError(Exception):
     pass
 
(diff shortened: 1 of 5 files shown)


python/taichi/lang/exception.py
```diff
@@ -1,3 +1,6 @@
+from taichi._lib import core
```
ailzhang, line 1: How about `from taichi._lib.core import TaichiTypeError`?

ailzhang (approved):
LGTM!

python/taichi/lang/exception.py
lin-hitonami: It seems that I can't do that... maybe because `taichi._lib.core` is a dynamic lib?

Issue #3785, opened 2021-12-13
Re-raised exceptions generates redundant tracebacks in AST transformer


lin-hitonami:
**Describe the bug**
The current error reporting mechanism of the AST transformer catches exceptions, adds the tracebacks in taichi program and re-raises the exception again and again until it is thrown outside the transformer. This makes python generate tracebacks to each exception and shows `During handling of the above exception, another exception occurred` because the program throws exceptions when processing the exception. 

This can be solved by adding`from None` to `raise TaichiCompilationError(msg)`. In this way, only the last traceback will be displayed. However, we may want to preserve the initial traceback if the error is raised from outside the AST transformer (for example, a python function outside the transformer or some malfunctioning taichi internal functions) like the following part from the example.
```
Traceback (most recent call last):
  File "/home/lin/taichi2/python/taichi/lang/ast/ast_transformer_utils.py", line 23, in __call__
    return method(ctx, node)
  File "/home/lin/taichi2/python/taichi/lang/ast/ast_transformer.py", line 353, in build_Call
    node.ptr = node.func.ptr(*args, **keywords)
  File "/home/lin/tmp/test.py", line 4, in baz
    raise RuntimeError("an error")
RuntimeError: an error
```

**To Reproduce**
```py
import taichi as ti
ti.init()

def baz():
    raise RuntimeError("an error")

@ti.func
def bar():
    baz()

@ti.kernel
def foo():
    bar()

foo()

```

**Log/Screenshots**
```
[Taichi] version 0.8.8, llvm 10.0.0, commit 6ef37747, linux, python 3.9.7
[Taichi] Starting on arch=x64
Traceback (most recent call last):
  File "/home/lin/taichi2/python/taichi/lang/ast/ast_transformer_utils.py", line 23, in __call__
    return method(ctx, node)
  File "/home/lin/taichi2/python/taichi/lang/ast/ast_transformer.py", line 353, in build_Call
    node.ptr = node.func.ptr(*args, **keywords)
  File "/home/lin/tmp/test.py", line 4, in baz
    raise RuntimeError("an error")
RuntimeError: an error

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/lin/taichi2/python/taichi/lang/ast/ast_transformer_utils.py", line 23, in __call__
    return method(ctx, node)
  File "/home/lin/taichi2/python/taichi/lang/ast/ast_transformer.py", line 353, in build_Call
[...]

PR #3965, opened 2022-01-07
[Error] Shorten the length of traceback of TaichiCompilationError

lin-hitonami:
Related issue = #3785

For the following program:
```python
import taichi as ti
ti.init()

@ti.kernel
def foo():
    1 in 2

foo()
```
The traceback looks like this after this PR:
```
  File "/home/lin/tmp/./test.py", line 8, in <module>
    foo()
  File "/home/lin/taichi2/python/taichi/lang/kernel_impl.py", line 724, in wrapped
taichi.lang.exception.TaichiSyntaxError: 
On line 6 of file "/home/lin/tmp/./test.py":
    1 in 2
    ^^^^^^
"In" is only supported inside `ti.static`.
```


strongoier (approved):
LGTM!

Issue #3937, opened 2022-01-05
Build on MSVC failed because of C4819


PGZXB:
Build on MSVC failed. Fail in the step [Build taichi from source](https://docs.taichi.graphics/lang/articles/contribution/dev_install#build-taichi-from-source)(select windows).

### Environment
* Compiler version : VS2022 17.0.0
* Operating system : Windows 10 20H2

### Reproduce steps
* Refer to [Developer installation | Taichi Docs](https://docs.taichi.graphics/lang/articles/contribution/dev_install). Fail in the step [Build taichi from source](https://docs.taichi.graphics/lang/articles/contribution/dev_install#build-taichi-from-source)(select windows).

### Build error
When building extension [SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools), error : 
```shell
Building CXX object external\SPIRV-Tools\...PIRV-Tools-static.dir\val\validate_image.cpp.obj
FAILED: external/SPIRV-Tools/source/CMakeFiles/SPIRV-Tools-static.dir/val/validate_image.cpp.obj
[...]

PR #3967, opened 2022-01-07
[build] Upgrade SPIRV-Headers and SPIRV-Tools to their latest commits

PGZXB:
Related issue = #3937

**Can not just upgrade SPIRV-Tools  because of dependency.**


qiao-bo (commented):
I am not sure about this. Can we update the submodule to a specific commit instead of updating files in that submodule?

qiao-bo (approved):
LGTM!

PR #3968, opened 2022-01-07
[rfc] AOT and prod deployment for sparse / quant / mesh

KuribohG:
Related issue = #

### DIFF
--- a/docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
+++ b/docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
@@ -1,0 +1,106 @@
+# [RFC] AOT and prod deployment for sparse / quant / mesh
+
+## Goals
+
+- Make Taichi deployable with full functionalities (including sparse / quant / mesh)
+- Users can manage the ownership of the system resources (including RAM, VRAM, GPU devices, command lists)
+- The provided Taichi runtime API should be independent of arch as much as possible
+
+## Current Problems
+
+Here is a code snippet in Taichi v0.8.8:
+
+```python
+a = ti.field(ti.i32)
+b = ti.field(ti.f32)
+ti.root.pointer(ti.ij, 16).dense(ti.ij, 16).place(a, b)
+
+@ti.kernel
+def run():
+  for I in ti.grouped(a):
+    b[I] = a[I] * 4.2
+```
+
+It is hard to deploy this code with AOT: `a` and `b` are runtime resources, which will couple kernel invocation of `run` and memory allocation for `a` and `b`.
+
+It is not a good idea to pack all these things into a module. The memory allocation process must be deployed to user applications, which will violate Goal 2.
+
+Notice that a SNodeTree is in fact a type. To solve this problem, we need to avoid memory allocation for `a` and `b`, but only keep their types (which is a SNodeTree). We can bind the actual physical memory after compilation.
+
+## Proposal
+
+### Python API
+
+Taichi AOT modules should only contain static program and data, like this:
+
+```python
+# This program is only for demonstration, it can't run on current version
+vec3 = ti.types.Vector(ti.f32, 3)
+s = ti.types.Struct({'a': ti.f32, 'b': vec3})
+
+# AOS
+sn1 = ti.types.SNodeTreeBuilder()
+                .pointer(ti.ij, 3)
+                .dense(ti.ij, 5)
+                .place(s)  # places s.a, s.b in an AOS way
+                .build()
+
+# SOA
+sn2b = ti.types.SNodeTreeBuilder()
+sn2b
+  .pointer(ti.ij, 3)
+  .dense(ti.ij, 5)
+  .place(s.a)           
+sn2b
+  .pointer(ti.ij, 3)
+  .dense(ti.ij, 5)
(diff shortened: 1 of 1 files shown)


docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
```diff
@@ -0,0 +1,106 @@
...
+
+auto *ab_aos_t = m.GetSNodeTreeType("ab_aos");
+const size_t tree_size = ab_aos_t.GetSize();
+// Allocate CUDA memory/ VkBuffer using |tree_size} however you want
+auto mem = AllocateDeviceMemory(tree_size);
+auto typed_mem = taichi::MakeMemoryResourceTyped(mem, ab_aos_t);
+
+auto *run1 = m.GetKernel("run1");
+run1.signature();  // {args: {ab_aos_t,}, returns: {}}
+run1.Run(/*args=*/{typed_mem});
```
k-ye, line 96: I'm not sure how feasible it is to simply provide a `Run()` method now. From our experience so far, users do need to control each individual shader (e.g. their workgroup size, how to bind the buffers, etc...)

k-ye (commented):
Thanks!

docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
k-ye: cc @ailzhang @turbo0628 @ghuau-innopeak

docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Except `GetDeviceMemory` depends on arch (like `void *` for CUDA, `VkBuffer` for Vulkan), other parts will be independent of arch.
+
+## TBD
+
+- Should the shape of SNodeTree be static?
+- Should the compiled results support multiple arch?
+- Is it possible to support compute graph for host logic?
```
ailzhang, line 105: This question is common to both this proposal and our dense aot solution right? (which might indicate it's a separate problem that we're trying to solve?

ailzhang (approved):

docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
ghuau-innopeak: What type is `typed_mem`? If this is a `DeviceAllocation`, I believe the `Run` function call would just be a wrapper to `launch_kernel` and `HostCtx` which should be fine, as the `typed_mem` could be re-use/bound to other individual shader.

docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
```diff
@@ -0,0 +1,106 @@
...
+
+@ti.kernel
+def run():
+  for I in ti.grouped(a):
+    b[I] = a[I] * 4.2
+```
+
+It is hard to deploy this code with AOT: `a` and `b` are runtime resources, which will couple kernel invocation of `run` and memory allocation for `a` and `b`.
+
+It is not a good idea to pack all these things into a module. The memory allocation process must be deployed to user applications, which will violate Goal 2.
```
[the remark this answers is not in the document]
KuribohG: This "user" should refer to "user of aot solution". They are the developers who want to embed Taichi in there application.

docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
KuribohG: Yes. I will remove this question. It's more relative to heterogeneous computing, which is not the focus of this proposal.

docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
```diff
@@ -0,0 +1,106 @@
...
+
+Taichi AOT modules should only contain static program and data, like this:
+
+```python
+# This program is only for demonstration, it can't run on current version
+vec3 = ti.types.Vector(ti.f32, 3)
+s = ti.types.Struct({'a': ti.f32, 'b': vec3})
+
+# AOS
+sn1 = ti.types.SNodeTreeBuilder()
```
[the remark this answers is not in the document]
KuribohG: I added this question to TBD section.

docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
k-ye: `typed_mem` is basically a DeviceAllocation with a SNodeTree type. This is because DeviceAllocation alone only represents a chunk of raw memory, so I think it's better to annotate it with the actual type.

docs/rfc/AOT_and_prod_deployment_for_sparse_quant_mesh.md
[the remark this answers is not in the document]
k-ye: Then let's explicitly list out the pros & cons here. This is a critical decision point.

PR #3971, opened 2022-01-07
[dx11] Add underlying DX11 device, memory allocation, and some tests

quadpixels:
Related issue = #992

This change adds an underlying DX11 device, memory allocation, and some tests.

The tests for the Device functionalities make it easier to confirm the functionalities are working without having to touch the rest of the code base (such as Program) or having to run a .py Taichi program.

I wonder if backend-specific tests are a good approach?

### DIFF
--- a/taichi/backends/dx/dx_device.cpp
+++ b/taichi/backends/dx/dx_device.cpp
@@ -4,18 +4,6 @@ namespace taichi {
 namespace lang {
 namespace directx11 {
 
-bool kD3d11DebugEnabled = false;  // D3D11 debugging is enabled. For testing.
-bool kD3d11ForceRef = false;      // Force REF device. May be used to
-                                  // force software rendering.
-
-void debug_enabled(bool enabled) {
-  kD3d11DebugEnabled = enabled;
-}
-
-void force_ref(bool force) {
-  kD3d11ForceRef = force;
-}
-
 void check_dx_error(HRESULT hr, const char *msg) {
   if (!SUCCEEDED(hr)) {
     TI_ERROR("Error in {}: {}", msg, hr);
--- a/taichi/backends/dx/dx_device.cpp
+++ b/taichi/backends/dx/dx_device.cpp
@@ -4,6 +4,16 @@ namespace taichi {
 namespace lang {
 namespace directx11 {
 
+bool kD3d11DebugEnabled = false;  // D3D11 debugging is enabled. For testing.
+bool kD3d11ForceRef = false;      // Force REF device. May be used to
+                                  // force software rendering.
+
+void check_dx_error(HRESULT hr, const char *msg) {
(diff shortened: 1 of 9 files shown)


bobcao3:
I'm so sorry I didn't see this! Please for the future PRs directly request a review from me

taichi/backends/dx/dx_device.cpp
```diff
@@ -4,6 +4,24 @@ namespace taichi {
 namespace lang {
 namespace directx11 {
 
+bool kD3d11DebugEnabled = false;  // D3D11 debugging is enabled. For testing.
```
bobcao3, line 7: Will it be fine to just leave these as compile constants? We should avoid global states

bobcao3 (approved):
LGTM other than the nitpick on the global states. Like the added test as well

taichi/backends/dx/dx_device.cpp
quadpixels: I don't know many ways for having compile time constants, so I added a new switch "TI_WITH_DX11_DEBUG", wondering if there is some better alternative?

bobcao3: You can use `constexpr bool kD3d11DebugEnabled = false` and then this variable will become compile time constant, and branches using this variable will be compiled out

quadpixels: I see, my intent is some unit tests are enabled only when the flag is true (otherwise the D3D info queue will not contain any information). This may be a user-facing feature (for example if the user wants to capture a trace with tools like RenderDoc?) but I'm not sure if Taichi is intended to be used in that way so I'm wondering which of the following make sense:
* It is set to `constexpr bool kD3d11DebugEnabled = false` in the .cpp file. Anyone who wants to enable it needs to modify the code manually.
* The flag and the test is controlled by the TI_WITH_DX11_DEBUG flag.
* Or something else?

bobcao3: The unit tests should run if the backend is built & tests are built. The unit test and debug layer should only be used for developers, so leaving it in the code is fine.

quadpixels: Thanks for the explanation, I removed the extra flag and left `kD3d11DebugEnabled` in the code.

taichi/backends/dx/dx_info_queue.h
```diff
@@ -0,0 +1,36 @@
...
+class Dx11InfoQueue {
+ public:
+  struct Entry {
+    std::string type;
+    void *addr;
+    int refcount;
+    int intref;
+  };
+  static std::vector<Entry> parse_reference_count(
+      const std::vector<std::string>);
```
k-ye, line 19:
```suggestion
      const std::vector<std::string>&);
```

k-ye (approved):
LGTM!

Issue #3334, opened 2021-10-31
Unify aot data structures 


ailzhang:
Some of the aot data structures can be shared among all backends, e.g. https://github.com/taichi-dev/taichi/blob/master/taichi/backends/opengl/aot_data.h. 
Unifying them should make serialization easier to manage.

PR #3973, opened 2022-01-08
[aot] Add a generic set of AOT structs

k-ye:
Related issue = #3334

I haven't really switched OpenGL or Vulkan to use these new AOT structs yet, but the migration should be very simple, e.g.,

```cpp
auto ad = AotDataConverter::convert(aot_data_);
TextSerializer ts;
t2.serialize_to_json("aot_data", ad);
t2.print();
```

I've also added/removed/changed a few field names:

* `ScalarArg`
   * `offset_in_bytes_in_args_buf` --> `offset_in_args_buf`
* `ogl::CompiledArrayArg` --> `aot::ArrayArg`
   * `shape_offset_in_bytes_in_args_buf` --> `shape_offset_in_args_buf`
   * Added `bind_index`
* `CompiledTaichiKernel`
   * `arg_count` --> `args_count`
   * `ret_count` --> `rets_count`
   * `args_buf_size` --> `args_buffer_size`
   * `ret_buf_size` --> `rets_buffer_size`
   * Removed `used.arr_arg_to_bind_idx`

### DIFF
--- a/taichi/aot/module_data.h
+++ b/taichi/aot/module_data.h
@@ -30,6 +30,70 @@ struct CompiledFieldData {
             column_num);
 };
 
+struct CompiledOffloadedTask {
+  std::string type;
+  std::string name;
+  // Do we need to inline the source code?
+  std::string source_path;
+  int gpu_block_size{0};
+
+  TI_IO_DEF(type, name, source_path, gpu_block_size);
+};
+
+struct ScalarArg {
+  std::string dtype_name;
+  // Unit: byte
+  size_t offset_in_args_buf{0};
+
+  TI_IO_DEF(dtype_name, offset_in_args_buf);
+};
+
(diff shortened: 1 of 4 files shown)


taichi/aot/module_data.h
```diff
@@ -30,6 +30,70 @@ struct CompiledFieldData {
             column_num);
 };
 
+struct CompiledOffloadedTask {
+  std::string type;
+  std::string name;
+  // Do we need to inline the source code?
```
ailzhang, line 36: IMHO it's fine to keep them as separate files?

ailzhang (approved):
Nice thanks!! 
for the bind_idx, I agree it makes more sense to move from CompiledTaichiKernel to ArrayArg for opengl backend as well :D

Issue #3635, opened 2021-11-28
[Discussion] Expected behavior of logical expressions in Taichi


strongoier:
Currently, there isn't a clear description of expected behavior of logical expressions (`and` / `or`) in Taichi. The implementation, https://github.com/taichi-dev/taichi/blob/71dd17426b62327864a21c65ca570812997c33ba/python/taichi/lang/ops.py#L792-L793, shows that `logical_and` / `logical_or` is simply treated as `bitwise_and` / `bitwise_or`. This leads to the non-short-circuit behavior (#3572) and the non-intuitive computation result:
```python
@ti.kernel
def func(a: ti.i32, b: ti.i32) -> ti.i32:
    return a and b

print(func(2, 3))  # 2
```

In Python, the behavior of logical expressions is like:
```python
print(2 and 3)  # 3
print(2.0 or 0)  # 2.0
```
which is short-circuiting and uses the `True` expression as the result.

Should we make logical expressions in Taichi behave the same as in Python, keep the original behavior in Taichi, or define yet-another behavior? Opinions are welcome.

lin-hitonami:
I think the behavior should be identical to Python's because most Python users would expect taichi's grammar is similar to Python's.

ailzhang:
I second @lin-hitonami 's idea, python frontend's behavior should be close to python convention as much as possible. ;)

strongoier:
According to the discussions in this line, we will make the behavior of logical expressions in Taichi the same as Python.

lin-hitonami:
There is a problem if we make the behavior the same as Python. For this program:
```python
@ti.kernel
def foo(a: ti.i32, b: ti.f32):
    c = a and b
```
What is the type of `c`? Will this confuse the type checker?

strongoier:
Yes this is indeed a problem. We cannot have exactly the same behavior as Python because Taichi is statically typed. In Taichi, all operands of logical expressions should be enforced to have a same type.

Then the core problem is more related to the order of our compilation process - we should do type check first, and then do transformations which may lose type enforcement information.

PR #3976, opened 2022-01-08
[example] [test] Fix misuse of logical operators in examples and tests

strongoier:
Related issue = #3635

In #3635, we determine that the logical operators in Taichi should behave the same as in Python. Therefore, we should no longer misuse logical operators as bitwise operators anymore. This PR aims at fixing these misuses in examples and tests.


yolo2themoon (approved):
LGTM! Thanks!

PR #3979, opened 2022-01-10
[opengl] Do not promote simple ExternalTensorShapeAlongAxisStmt into globals.

ailzhang:
* #3980
* __->__ #3979

Previously we only make sure `for i in range(x)` (here `x` is ndarray)
don't get their end_range from gtmp on opengl. This PR expands the scope
to all simple calculations involving only ConstStmt and
ExternalTensorShapeAlongAxisStmt as leaves, and it's not limited to the
end_stmt in RangeForStmt as well. In most cases this should save an
additional serial offloaded task which only saves the simple
intermediates into gtmp.

### DIFF
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -896,9 +896,12 @@ class KernelGen : public IRVisitor {
     const auto axis = stmt->axis;
     used.buf_args = true;
     used.int32 = true;
-    emit("int {} = _args_i32_[{} + {} * {} + {}];", name,
-         taichi_opengl_extra_args_base / sizeof(int), arg_id,
-         taichi_max_num_indices, axis);
+    if (!loaded_args_.count(name)) {
+      emit("int {} = _args_i32_[{} + {} * {} + {}];", name,
+           taichi_opengl_extra_args_base / sizeof(int), arg_id,
+           taichi_max_num_indices, axis);
+      loaded_args_.insert(name);
+    }
   }
 
   std::string make_kernel_name() {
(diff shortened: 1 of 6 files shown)


ailzhang:
Sorry this PR is a bit harder to review, I'll find some time later today to add more context and examples here. :P

k-ye (approved):
LGTM!

ailzhang:
Note before this PR our first serialized ndarray based mpm88 shader looks like this:
```void substep_c44_00()
{ // serial
  float B = float(0.5);
  float C = float(1.0);
  int D = _args_i32_[16 + 4 * 8 + 0];
  int E = 16;
  _gtmp_i32_[E >> 2] = D;
  float G = float(D);
  int H = 0;
  _gtmp_f32_[H >> 2] = G;
  float J = C / G;
  int K = 12;
  _gtmp_f32_[K >> 2] = J;
  float M = J * B;
  float N = M * M;
  int O = 8;
  _gtmp_f32_[O >> 2] = N;
  float Q = float(0.0002);
  float R = float(0.0078125);
  float S = G * R;
  float T = Q / S;
  float U = float(min(T, Q));
  int V = 4;
  _gtmp_f32_[V >> 2] = U;
}
```
Note the `gtmp` stores some intermediate values used in the following shaders. These intermediates are from
```
  @ti.kernel
  def substep(x: ti.any_arr(element_dim=1), v: ti.any_arr(element_dim=1), J: ti.any_arr(),
              C: ti.any_arr(element_dim=2), grid_v: ti.any_arr(element_dim=1), grid_m: ti.any_arr()):
      dx = 1 / grid_v.shape[0]
      inv_dx = grid_v.shape[0]
      p_vol = (dx * 0.5)**2
      p_mass = p_vol * p_rho
      dt = min(2.0e-4 / (grid_v.shape[0] / 128), 2.0e-4)
```
and they got promoted to globals through https://github.com/taichi-dev/taichi/blob/master/taichi/transforms/offload.cpp#L779. 

However, these intermediates are mostly simple arithmetics related to `ExternalTensorShapeAlongAxisStmt` and reading `gtmp` buffer is as expensive as reading `args` buffer. So it's actually better to recalculate them in each shader, so that we can save a serial shader run. 

Note previously we "solved" a similar issue for `RangeForStmt` when its end_stmt is `ExternalTensorShapeAlongAxisStmt` but this PR reverts some of those changes and did it in a slightly better way. The solution here still need some followup around the end_stmt of `OffloadStmt`.

PR #3980, opened 2022-01-10
[refactor] [opengl] Get rid of some no-ops in linear seek

ailzhang:
* __->__ #3980
* #3979

Reduced p2g shader from 658 lines to 514 lines.

### DIFF
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -506,28 +506,35 @@ class KernelGen : public IRVisitor {
   void visit(GlobalStoreStmt *stmt) override {
     TI_ASSERT(stmt->width() == 1);
     auto dt = stmt->val->element_type();
+    std::string index = stmt->dest->is<ExternalPtrStmt>()
+                            ? stmt->dest->short_name()
+                            : fmt::format("{} >> {}", stmt->dest->short_name(),
+                                          opengl_data_address_shifter(dt));
+
     emit(
-        "_{}_{}_[{} >> {}] = {};",
+        "_{}_{}_[{}] = {};",
         ptr_signats_.at(stmt->dest->id),  // throw out_of_range if not a pointer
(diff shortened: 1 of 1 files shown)


k-ye (approved):
Thanks!

ailzhang:
Before this PR, when we serialize the ndarray version of mpm88, our first substep shader looks like this
```
void substep_c44_00()
{ // serial
  float B = float(0.5);
  float C = float(1.0);
  int D = _args_i32_[16 + 4 * 8 + 0];
  int E = 16;
  _gtmp_i32_[E >> 2] = D;
  float G = float(D);
  int H = 0;
  _gtmp_f32_[H >> 2] = G;
  float J = C / G;
  int K = 12;
  _gtmp_f32_[K >> 2] = J;
  float M = J * B;
  float N = M * M;
  int O = 8;
  _gtmp_f32_[O >> 2] = N;
  float Q = float(0.0002);
  float R = float(0.0078125);
  float S = G * R;
  float T = Q / S;
  float U = float(min(T, Q));
  int V = 4;
  _gtmp_f32_[V >> 2] = U;
}
```
And these `gtmp` values are indeed some intermediate values used in multiple following shaders and they got promoted to globals in https://github.com/taichi-dev/taichi/blob/master/taichi/transforms/offload.cpp#L793. The key reason that we have an additional shader here is the arg load statement is not propagated to each shader. 
Given that reading gtmp buffer is as expensive as reading args buffer, we don't gain too much value from calculating "arg load"(I actually meant `ExternalTensorShapeAlongAxisStmt`) related intermediates in an additional buffer.  So what this PR does is mainly just cloning `ExternalTensorShapeAlongAxisStmt` into each shader. 

Note we previously did "solve" the same problem in #3884 for `RangeForStmt` when its `end_stmt` is only about `ExternalTensorShapeAlongAxisStmt`. This PR reverts some changes from that PR and did it in a slightly better way. 

After the `ExternalTensorShapeAlongAxisStmt` are cloned into each shader, we can actually get rid of the additional serial shader and the whole program runs faster.

PR #3984, opened 2022-01-10
[refactor] Refactor func body to reduce python overhead and improve readability

turbo0628:
Related issue = #3915 

Orientation:

The python code around kernel launch, say `func__`, has a long and verbose section to check `ndarray` and torch related types and build respective contexts. The performance of those if-else clauses, however, is critical for lightweight kernels with multiple ndarrays. The flags are also kinda confusing. In this PR I refactored the clause regarding `any_arr` in order to make it more readable and efficient. 

Briefings of modifications:

1) Pack the torch-related code in a function
2) refactor the if clauses in order to skip `match_ext_arr` for `Ndarray`

Hope the refactored code reads better :)

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -476,7 +476,7 @@ class Kernel:
         assert key not in self.compiled_functions
         self.compiled_functions[key] = self.get_function_body(taichi_kernel)
 
-    def get_torch_callbacks(self, v, is_ndarray, has_torch):
+    def get_torch_callbacks(self, v, has_torch, is_ndarray=True):
         callbacks = []
 
         def get_call_back(u, v):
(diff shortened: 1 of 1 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -476,6 +476,42 @@ def taichi_ast_generator():
         assert key not in self.compiled_functions
         self.compiled_functions[key] = self.get_function_body(taichi_kernel)
 
+    def get_torch_callbacks(self, v, is_ndarray, has_torch):
```
k-ye, line 479: My only nit: put `is_ndarray` in the end, so that we we call it, we can write `get_torch_callbacks(..., is_ndarray=True)`, so that ppl know what `True` `False` means... (Or we can use a more explicit variable name for T/F)

k-ye (approved):
LGTM!

python/taichi/lang/kernel_impl.py
turbo0628: SG, let me update it!

turbo0628: Done :)

Issue #3985, opened 2022-01-10
Taichi Python 3.10 Readiness


qiao-bo:
Here we track the actions to be taken to support python 3.10 in Taichi. Feel free to adapt:

#### Failed test:
- [x] mesh on cpu (fixed by #4275)

#### Package that is not yet supported
- [ ] torch [(tracking issue here)](https://github.com/pytorch/pytorch/issues/66424)

#### OS
- [x] Ubuntu
- [x] Macos 10.15
- [x] Macos m1
- [x] Windows

#### CI Dockerfile:
- [x] Ubuntu 18.04 CPU
- [x] Ubuntu 18.04 GPU and all backends

#### Docs:
- [x] dev_install
- [x] get-started
- [x] README

qiao-bo:
related issue: https://github.com/taichi-dev/taichi/issues/3936

qiao-bo:
Python 3.10 is now supported since release v0.9.2.

PR #3986, opened 2022-01-10
[misc] Upgrade test and docker image to support python 3.10

qiao-bo:
Related issue = #3985 

This PR add test support for python 3.10 in Taichi. Note that torch-related executions are temporally disabled because it does not yet support 3.10.

CPU Linux 3.6 test is replaced with 3.10. For other OS we keep the current python version, but can be easily changed in the yml file.


python/taichi/lang/__init__.py
```diff
@@ -483,6 +483,8 @@ def check_version():
         payload['python'] = 'cp38'
     elif python_version.startswith('3.9.'):
         payload['python'] = 'cp39'
+    elif python_version.startswith('3.10.'):
+        payload['python'] = 'cp310'
```
frostming, line 487: Can be changed to a more concise form:
```python
python_version = sys.version_info
payload['python'] = f'cp{python_version[0]}{python_version[1]}'
```

frostming:
Also need to update the `python_requires` range in `setup.py`

python/taichi/lang/__init__.py
qiao-bo: thanks, i believe you mean this?
```
python_version = platform.python_version().split('.')
payload['python'] = f'cp{python_version[0]}{python_version[1]}'
```

frostming: `sys.version_info` is a tuple of version bits so we don't need to split it ourselves.

qiao-bo: i see, thanks a lot, changed!

k-ye (approved):
Great job, LGTM!

.github/workflows/scripts/unix_test.sh
```diff
@@ -20,7 +20,9 @@ fi
 python3 -m pip install dist/*.whl
 if [ -z "$GPU_TEST" ]; then
     python3 -m pip install -r requirements_test.txt
-    python3 -m pip install torch
+    if [[ $PY != *"3.10"* && $PY != *"py310"* ]]; then
```
frostming, line 23: I suggest making a `requirements_torch.txt` which contains:
```
torch; python_version < '3.10'
```
Let `pip` decide whether to install based on the python version.

frostming: Or more simple, via command line:

```
python3 -m pip install "torch; python_version < '3.10'"
```

qiao-bo:
@frostming thanks a lot! changed made.

.github/workflows/scripts/win_test.ps1
```diff
@@ -7,10 +6,13 @@ ti diagnose
...
-    pip install torch==1.10.1+cu113 -f https://download.pytorch.org/whl/cu113/torch_stable.html
-} else {
-    pip install torch
+# TODO relax this when torch supports 3.10
+if (-not "$env:PYTHON".Contains("3.10")) {
+  if ("$env:TI_WANTED_ARCHS".Contains("cuda")) {
+      pip install torch==1.10.1+cu113 -f https://download.pytorch.org/whl/cu113/torch_stable.html
+  } else {
+      pip install torch
+  }
```
frostming, line 15:
```suggestion
# TODO relax this when torch supports 3.10
if ("$env:TI_WANTED_ARCHS".Contains("cuda")) {
    pip install "torch==1.10.1+cu113; python_version < '3.10'" -f https://download.pytorch.org/whl/cu113/torch_stable.html
} else {
    pip install torch
}
```

PR #3987, opened 2022-01-10
[mesh] Support multiple major relations in one mesh-for loop

g1n0st:
Related issue = #3608
In the previous design principle, we restrict one mesh-for should only have one major relation, like the following case:
``` Python
for v in model.verts:
  for i in range(v.verts.size):
    # do something ...
  for i in range(v.cells.size): # it's not allowed before this PR because the major relation is Verts-Verts
    # do other things ...
```
But we found that supporting multiple major relations in one mesh-for loop is necessary for some scenarios. Otherwise, we must define some extra variables, like this:
``` Python
for v in model.verts:
  a, b = 0, 0
  for i in range(v.verts.size):
    a += v.verts[i].id
  for i in range(v.cells.size):
    b += v.cells[i].id
  v.res = a * b
```
Note that defining `a` and `b` as local variables is far optimized than placing them in the global memory. So we deprecate the single-major-relation restriction rule in this PR and add related test.

### DIFF
--- a/taichi/analysis/gather_meshfor_relation_types.cpp
+++ b/taichi/analysis/gather_meshfor_relation_types.cpp
@@ -29,9 +29,6 @@ class GatherMeshforRelationTypes : public BasicStmtVisitor {
     TI_ASSERT(stmt->minor_relation_types.size() == 0);
     mesh_for = stmt;
     stmt->body->accept(this);
-    TI_ASSERT_INFO(
-        stmt->major_to_types.size() <= 1,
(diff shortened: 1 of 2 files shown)


BillXu2000 (approved):
Looks good to me!

BillXu2000:
Conflict occurs due to new mesh-for tests in #3990.
@g1n0st could you please resolve it?

g1n0st:
> Conflict occurs due to new mesh-for tests in #3990. @g1n0st could you please resolve it?

Fixed

PR #3990, opened 2022-01-10
[mesh] [opt] Support nested mesh-for

g1n0st:
Related issue = #3608 

Before this PR, do the neighbor query loop in mesh-for must resort to nested range-for, like:
``` Python
for u in model.verts:
  for i in u.verts.size:
    u.x += u.verts[i].y
```
This PR implemented the nested mesh-for loop, the code above could be rewritten as:
``` Python
for u in model.verts:
  for v in u.verts:
    u.x += v.y
```
**Note that the nested mesh-for loop would not be un-looped.**


BillXu2000 (approved):
Looks good to me!

PR #3991, opened 2022-01-11
[opengl] [aot] Convert opengl aot to dump ModuleData.

ailzhang:
This PR switch opengl backend to dump `ModuleData` instead of its own structure. 

Followup: need to update the deserializer accordingly for bind_index.

### DIFF
--- a/taichi/aot/module_data.h
+++ b/taichi/aot/module_data.h
@@ -91,6 +91,12 @@ struct ModuleData {
 
   size_t root_buffer_size;
 
+  void dump_json(std::string path) {
+    TextSerializer ts;
+    ts.serialize_to_json("aot_data", *this);
+    ts.write_to_file(path);
+  }
+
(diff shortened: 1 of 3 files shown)


taichi/aot/module_data.h
```diff
@@ -33,11 +33,10 @@ struct CompiledFieldData {
 struct CompiledOffloadedTask {
   std::string type;
   std::string name;
-  // Do we need to inline the source code?
-  std::string source_path;
+  std::string source;
```
k-ye, line 36: Are we directly inlining the shaders now? What about SPIR-V?

ailzhang: @k-ye good point! Hmmm I'll revert this back to source_path, and move writing to "spv/glsl" files to `convert()` instead.

k-ye (approved):
LGTM!

PR #3992, opened 2022-01-11
[opt] Add more strict alias analysis for ExternalPtrStmt

ailzhang:
This is not likely not 100% safe, how about banning "passing the same ndarray as
different arguments to a taichi kernel"?

It helps eliminating identical `GlobalLoadStmt`s (whose source are`ExternalPtrStmt`s) in mpm88 aot example. 

Related issue = #


strongoier (approved):
LGTM! Only need to prevent users from passing same ndarrays into a kernel :-)

PR #3995, opened 2022-01-11
[test] [example] Add a test and a video generator for mpm99

strongoier:
Related issue = #3580

This PR serves as a sample of adding tests and video generators for examples involving GUI following the guidelines in #3580. A script for running all video generators is added by the way.

### DIFF
--- a/misc/generate_example_videos.py
+++ b/misc/generate_example_videos.py
@@ -1,0 +1,16 @@
+import re
+import os
+import subprocess
+import sys
+
+if len(sys.argv) != 2:
+    print('usage: generate_example_videos.py <output_directory>')
+    exit(0)
+
+example_root = os.path.join('..', 'tests', 'python', 'examples')
+for example_dir in os.listdir(example_root):
+    full_dir = os.path.join(example_root, example_dir)
+    for filename in os.listdir(full_dir):
+        match = re.match(r'test_(\w+)\.py', filename)
+        if match:
+            subprocess.run(["python", os.path.join(full_dir, filename), os.path.join(sys.argv[1], match.group(1))])
--- a/misc/generate_example_videos.py
+++ b/misc/generate_example_videos.py
@@ -1,5 +1,5 @@
-import re
 import os
+import re
 import subprocess
 import sys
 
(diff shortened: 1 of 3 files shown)


tests/python/examples/simulation/test_mpm99.py
```diff
@@ -0,0 +1,41 @@
...
+    from taichi.examples.simulation.mpm99 import dt, initialize, substep
+
+    initialize()
+    for i in range(FRAMES):
+        for s in range(int(2e-3 // dt)):
+            substep()
+
+
+def video_mpm99(result_dir):
+    from taichi.examples.simulation.mpm99 import (dt, initialize, material,
```
ailzhang, line 16: importing ti from mpm99 seems a bit suspicious :P Is it the `ti` after initialization? 
Does it work if we move imports to the top level?

strongoier: Do you mean moving `import taichi as ti` to the top?

ailzhang: yup, I wonder it local import here is required. (If it is we need to explain it in the instruction

strongoier: Only moving `import taichi as ti` to the top is OK. For others we have to locally import them because some actual logic of that example is running during the import.

misc/generate_example_videos.py
```diff
@@ -0,0 +1,20 @@
+import os
+import re
+import subprocess
+import sys
+
+if len(sys.argv) != 2:
```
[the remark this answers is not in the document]
strongoier: Updated.

ailzhang (approved):
LGTM!

tests/python/examples/simulation/test_mpm99.py
```diff
@@ -0,0 +1,45 @@
...
+                                                  substep, x)
+
+    video_manager = ti.VideoManager(output_dir=result_dir,
```
k-ye, line 21: nit: Do we really want to keep `VideoManager` in the top level in the future...?

strongoier: This is still under discussion. My personal idea is no :-)

PR #4000, opened 2022-01-12
[vulkan] [aot] Enable aot tests for vulkan backend.

ailzhang:
This PR fixes a typo in std::optional serialization and enabled field
based aot tests for vulkan backend.

Note our `arch` in `ti.aot.Module(arch)` API is ignored, this API is a bit
misleading so we made it clear in tests.

Related issue = #


k-ye (approved):
LGTM!

ailzhang:
synk error is not related, merging :D

PR #4009, opened 2022-01-12
[Lang] Make the kernel return a matrix

mzmzm:
Related issue = #909
`@ti.kernel` cannot return a matrix or vector in the current version. I try to add this support in this PR. 
But there are still some problems here. Due to the current designs of the backend, the result buffer needs to be changed to pass more data.
I have done something about the result buffer in the llvm-backend, and it works fine on my computer. 

- [x] llvm IR
- [x] spirV IR
- [x] codegen_metal
- [x] codegen_cc
- [x] codegen_opengl
- [x] vulkan

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -1,5 +1,6 @@
 import ast
 import collections.abc
+import itertools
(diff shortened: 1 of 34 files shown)


mzmzm:
Related #4028

PR #4011, opened 2022-01-12
[vulkan] [aot] Switch vulkan aot to use taichi::aot::ModuleData.

ailzhang:
* #4012
* __->__ #4011

related #3334

This PR implements a converter to convert vulkan aot data structure to
unified taichi::aot::ModuleData.

### DIFF
--- a/taichi/backends/vulkan/aot_module_builder_impl.cpp
+++ b/taichi/backends/vulkan/aot_module_builder_impl.cpp
@@ -46,9 +46,6 @@ class AotDataConverter {
     for (const auto &arg : in.ctx_attribs.args()) {
       res.scalar_args[arg.index] = visit(arg);
     }
-    for (const auto &ret : in.ctx_attribs.rets()) {
-      res.arr_args[ret.index] = visit(ret);
-    }
     return res;
   }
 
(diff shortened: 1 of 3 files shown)


taichi/backends/vulkan/aot_module_builder_impl.cpp
```diff
@@ -46,9 +105,10 @@ uint32_t AotModuleBuilderImpl::to_vk_dtype_enum(DataType dt) {
 
 void AotModuleBuilderImpl::write_spv_file(
     const std::string &output_dir,
-    const TaskAttributes &k,
+    TaskAttributes &k,
     const std::vector<uint32_t> &source_code) const {
   const std::string spv_path = fmt::format("{}/{}.spv", output_dir, k.name);
+  k.source_path = spv_path;
```
k-ye, line 111: Can we directly store `spv_path` to `aot::CompiledOffloadedTask`, so that `k` remains const?

ailzhang: @k-ye You mean doing it inside `aot::CompiledOffloadedTask AotDataConverter::visit(const TaskAttributes &in)` function right? I thought about it but didn't went down that path mainly because: 
we "need/want" the `output_dir` information in the `spv_path` but I cannot convince myself passing it to converter... (converter should be a pure data structure transformation to me and should have nothing to do with where you dump the files...

k-ye: One dump solution: separate path generation from `write_spv_file`..

ailzhang: sorry I was dumb :P should be working now

AmesingFlank (approved):

PR #4022, opened 2022-01-13
[misc] Skip preview when docs and python are not updated.

ailzhang:
We want to start two different preview jobs, but ignore command can only be set in toml files.


ailzhang:
Switched to vercel.

PR #4024, opened 2022-01-14
[misc] Skip preview unless python dir is not updated.

rexwangcc:
Related issue = #

This is primarily a validation test to make sure we skip preview unless python dir is not updated.


rexwangcc:
Changes to locations other than `python/` is not triggering API docs preview, which is expected. Closing.

PR #4027, opened 2022-01-14
[build] Disable LTO for mac.

### LINKED ISSUE #3939: Linking got slower on my m1 mac
Building from source spent a lot time at `Linking CXX shared library libtaichi_core.dylib` (more than 2min) and I'm pretty sure it was only a few secs before. 

We'll need to further investigate this regression but I'll leave this issue here first.

ailzhang:
Fixes #3939.

I have AppleClang12.0.5 and enabling lto=thin made linking
extremely slow here. Need to check if it's related to my clang version.
But let's disable it first :P.

Related issue = #

### DIFF
--- a/cmake/TaichiCXXFlags.cmake
+++ b/cmake/TaichiCXXFlags.cmake
@@ -17,12 +17,11 @@ if (MINGW)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -static-libstdc++")
 endif ()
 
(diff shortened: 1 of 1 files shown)


turbo0628 (approved):
LGTM!

It also works on my laptop with clang 11.1.0 (conda build), macOS 11.6.1. 
I didn't notice performance regression. We can add this option back when we are strict on compilation speed.

PR #4028, opened 2022-01-14
[ir] [refactor] Make ReturnStmt support a vector of stmts

mzmzm:
Related issue = #909


python/taichi/lang/ast/ast_transformer.py
```diff
@@ -468,9 +469,22 @@ def build_Return(ctx, node):
...
+                        expr.make_expr_group(
+                            ti_ops.cast(expr.Expr(node.value.ptr),
+                                        ctx.func.return_type).ptr))
+                else:
+                    _ti_core.create_kernel_exprgroup_return(
+                        expr.make_expr_group([
+                            ti_ops.cast(
+                                exp, ctx.func.return_type.dtype if isinstance(
+                                    ctx.func.return_type, MatrixType) else
+                                ctx.func.return_type) for exp in list(
```
strongoier, line 483: 1. What does the `else` case mean?
2. `list(` is unnecessary.

python/taichi/lang/ops.py
```diff
@@ -129,6 +129,13 @@ def cast(obj, dtype):
     return expr.Expr(_ti_core.value_cast(expr.Expr(obj).ptr, dtype))
 
 
+def group_cast(group, dtype):
```
strongoier, line 132: Where is it used?

python/taichi/lang/util.py
```diff
@@ -191,6 +191,8 @@ def cook_dtype(dtype):
         return dtype
     if isinstance(dtype, _ti_core.Type):
         return _ti_core.DataType(dtype)
+    if isinstance(dtype, impl.MatrixType):
```
strongoier, line 194: `cook_dtype` is used for transformations like `float` -> `default_fp`. Why declare a tensor type in C++ here?

taichi/python/export_lang.cpp
```diff
@@ -926,6 +926,10 @@ void export_lang(py::module &m) {
               dt, total_dim, shape);
         });
 
+  m.def("decl_tensor_type", [&](std::vector<int> shape, DataType element) {
```
strongoier, line 929: ditto

strongoier (changes requested):

taichi/codegen/codegen_llvm.cpp
```diff
@@ -1074,28 +1074,35 @@ void CodeGenLLVM::visit(ReturnStmt *stmt) {
   if (stmt->ret_type.is_pointer()) {
     TI_NOT_IMPLEMENTED
   } else {
+    TI_ASSERT(
+        stmt->values.size() <= 30 &&
```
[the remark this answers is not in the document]
mzmzm: Due to the limitation of our backend in [taichi/inc/constants.h](https://github.com/taichi-dev/taichi/blob/016feb385ffcba70aeaabd49a172346efcab4007/taichi/inc/constants.h#L20), if we want to support more than one return value, the number of values can't exceed 30.

python/taichi/lang/util.py
mzmzm: This function also transforms python_type to c++_type here. I think this is needed.

python/taichi/lang/ops.py
mzmzm: This function is used in my preliminary version, now it is redundant. I will remove it soon.

python/taichi/lang/ast/ast_transformer.py
mzmzm: It seems unnecessary in this PR, I will delete it.

taichi/codegen/codegen_llvm.cpp
```diff
@@ -1074,28 +1074,35 @@ void CodeGenLLVM::visit(ReturnStmt *stmt) {
   if (stmt->ret_type.is_pointer()) {
     TI_NOT_IMPLEMENTED
   } else {
+    TI_ASSERT(
+        stmt->values.size() <= taichi_max_num_ret_value &&
+        "ERROR: Return list's size must less than or equal to 30 currently!");
```
strongoier, line 1079: 1. Why `&&`?
2. 30 should be replaced with `taichi_max_num_ret_value`.

mzmzm: 3b357e2726e185846d448ff4bd6a00b039c5c79b

strongoier (approved):
LGTM!

Issue #4003, opened 2022-01-12
Misleading Error Message


neozhaoliang:
```python
import taichi as ti
ti.init(arch=ti.gpu)

import numpy as np

# define 10 points as vertices
points = ti.Vector.field(3, ti.f32, shape=(10, 1))
points.from_numpy(np.random.random((10, 1, 3)))

@ti.func
def get_polygon(polygon_indices: ti.template()):
    coords = []
    for ind in ti.static(range(len(polygon_indices))):
        coords.append(points[polygon_indices[ind], 0])

    return coords


@ti.kernel
def main():
    arr = np.array([1, 2, 3, 4], dtype=int)
    get_polygon(arr)


main()
```
The above code reports error
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/taichi/lang/ast/ast_transformer_utils.py", line 23, in __call__
    return method(ctx, node)
  File "/usr/local/lib/python3.8/dist-packages/taichi/lang/ast/ast_transformer.py", line 80, in build_Assign
    ASTTransformer.build_assign_unpack(ctx, node_target,
  File "/usr/local/lib/python3.8/dist-packages/taichi/lang/ast/ast_transformer.py", line 98, in build_assign_unpack
    return ASTTransformer.build_assign_basic(ctx, node_target, values,
  File "/usr/local/lib/python3.8/dist-packages/taichi/lang/ast/ast_transformer.py", line 129, in build_assign_basic
    var = ti.expr_init(value)
  File "/usr/local/lib/python3.8/dist-packages/taichi/lang/util.py", line 216, in wrapped
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/taichi/lang/impl.py", line 61, in expr_init
    return Expr(_ti_core.expr_var(Expr(rhs).ptr))
  File "/usr/local/lib/python3.8/dist-packages/taichi/lang/expr.py", line 33, in __init__
    self.ptr = impl.make_constant_expr(arg).ptr
  File "/usr/local/lib/python3.8/dist-packages/taichi/lang/util.py", line 216, in wrapped
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/taichi/lang/impl.py", line 444, in make_constant_expr
    return Expr(_ti_core.make_const_expr_f32(val))
TypeError: make_const_expr_f32(): incompatible function arguments. The following argument types are supported:
[...]

PR #4030, opened 2022-01-14
[Error] Add error message when non-0d numpy ndarray is given to initialize expression

lin-hitonami:
Related issue = #4003


python/taichi/lang/expr.py
```diff
@@ -17,16 +18,21 @@ def __init__(self, *args, tb=None):
...
-                        arg = arg.dtype(arg)
-                except:
-                    pass
+                if isinstance(arg, np.ndarray):
+                    if arg.size != 1:
+                        raise TaichiTypeError(
+                            "Only numpy array with one element can be used to initialize a scalar expression"
+                        )
+                    try:
+                        arg = arg.dtype.type(arg.reshape(()))
```
ailzhang, line 33: Hmm I'm a little skeptical about this: are we going to allow arbitrary shape with one element to be converted to constant expr? 

IMHO we implicitly lose the shape information by doing this without user awareness. Is it better to let users do `arg.reshape(())` instead of us doing it?

lin-hitonami: There is a case in our tests (tests/python/test_loops.py::test_numpy_loops), should we ban it and fix the test?

ailzhang: Discussed offline: let's make it stricter for now, users should be aware of the numpy shape and make sure it only pass in a scalar :D

ailzhang (approved):
LGTM! Thanks!

PR #4031, opened 2022-01-14
[misc] Add a minimal example for micro-benchmarks

yolo2themoon:
Related issue = #4058

### DIFF
--- a/benchmarks/microbenchmarks/_items.py
+++ b/benchmarks/microbenchmarks/_items.py
@@ -4,9 +4,10 @@ import taichi as ti
 
 
 class BenchmarkItem:
+    name = 'item'
+
     def __init__(self):
         self._items = {}  # {tag: {impl: xxx, ...}}
-        self._name = 'item'
 
     def get(self):
         return self._items  #dict
@@ -19,6 +20,9 @@ class BenchmarkItem:
 
     def tag_in_item(self, tag: str):
         return tag in self._items
+
+    def impl(self, tag: str):
+        return self._items[tag]['impl']
 
 
 class DataType(BenchmarkItem):
--- a/benchmarks/microbenchmarks/_items.py
+++ b/benchmarks/microbenchmarks/_items.py
@@ -1,4 +1,4 @@
-from microbenchmarks._utils import _size2tag
+from microbenchmarks._utils import size2tag
 
 import taichi as ti
 
@@ -7,10 +7,10 @@ class BenchmarkItem:
     name = 'item'
 
     def __init__(self):
-        self._items = {}  # {tag: xxx, ...}
+        self._items = {}  # {'tag': impl, ...}
 
     def get(self):
-        return self._items  #dict
+        return self._items
 
     def get_tags(self):
         return self._items.keys()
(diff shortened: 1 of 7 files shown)


benchmarks/microbenchmarks/_result.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self):
+        self._items = {'kernel_elapsed_time_ms': kernel_executor}
+
+    @staticmethod
+    def init_taichi(arch: str, result_tag: str):
+        if result_tag == 'kernel_elapsed_time_ms':
+            ti.init(kernel_profiler=True, arch=get_ti_arch(arch))
+        else:
+            return False
+        return True
```
lin-hitonami, line 29:
```suggestion
        if result_tag == 'kernel_elapsed_time_ms':
            ti.init(kernel_profiler=True, arch=get_ti_arch(arch))
            return True
        return False
```

benchmarks/microbenchmarks/_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    for dsize, unit in reversed(size_subsection):
+        if size_in_byte >= dsize:
+            return str(int(size_in_byte / dsize)) + unit
+
+
+def tags2name(tag_list):
+    name = ''
+    for tag in tag_list:
+        name += '_' + tag
+    return name.lstrip("_")
```
lin-hitonami, line 19:
```suggestion
    return '_'.join(tag_list)
```

benchmarks/microbenchmarks/_items.py
```diff
@@ -0,0 +1,47 @@
...
+    name = 'item'
+
+    def __init__(self):
+        self._items = {}  # {tag: xxx, ...}
+
+    def get(self):
+        return self._items  #dict
+
+    def get_tags(self):
+        return [key for key in self._items]
```
lin-hitonami, line 16:
```suggestion
        return self._items.keys()
```

benchmarks/microbenchmarks/_plan.py
```diff
@@ -0,0 +1,59 @@
...
+            _ms = self._run_func(tag_list)
+            plan['result'] = _ms
+            print(f'{tag_list}={_ms}')
+            ti.reset()
+        rdict = {'results': self.plan, 'info': self.info}
+        return rdict
+
+    def _get_kwargs(self, tags):
+        kwargs = {}
+        for item in self.items:
```
lin-hitonami, line 45: If the order of `items` and `tags` are the same, maybe we can iterate through `zip(self.items, tags)` directly?

yolo2themoon: bf54c75, fixed by removing case_name from the tag list, thanks! 
```
tags = tags[1:] 
for item, tag in zip(self.items, tags):
```

ailzhang (commented):
Now I have a basic idea how this works, but will leave the final review to @lin-hitonami :D

benchmarks/microbenchmarks/_utils.py
lin-hitonami: `'_'.join(tag_list)` should work exactly as the function here?

yolo2themoon: my bad, verified locally but missed it when submitting 😭

lin-hitonami (approved):
LGTM

Issue #4032, opened 2022-01-14
[refactor] Export ASTBuilder to remove some dependencies on global accessing


PGZXB:
### Background
Some functions exported from C++ side is dependent on accessing to global state, which is bad. Most of them depend on `current_ast_builder()` which is dependent on `get_current_program()`. I'm removing `get_current_proram()` in source code of Taichi(see  #3955), which is important to improving code quality. But lots of global functions like `_ti_core.create_kernel_return` exported from C++ prevent me from continuing to refactor.
### Function List
**A -> B : Function B is called by function A**
1. The functions depending on `current_ast_builder()` should be reset to member of `ASTBuilder`.
2. ~~Because `Expr` depends on `ASTBuilder`, `ast_builder` should be passed in `Expr::make`, which is my next task to remove `get_current_program`.~~ The functions depending on `current_ast_builder()` should be also reset to member of `ASTBuilder`.
3. Some functions like `end_frontend_range_for` should be reset to member of `ASTBuilder`.
- insert_deactivate -> Deactivate -> current_ast_builder
- insert_activate -> Activate -> current_ast_builder
- ~~expr_get_addr -> Expr::make~~
- insert_append -> Append -> Expr::make
- insert_external_func_call -> Expr::make + current_ast_builder
- insert_is_active -> is_active -> Expr::make
- insert_len -> Length -> Expr::make
- create_assert_stmt -> current_ast_builder
- begin_frontend_while -> current_ast_builder
- begin_frontend_struct_for -> current_ast_builder
- begin_frontend_mesh_for -> currnet_ast_builder
- end_frontend_range_for
- end_frontend_struct_for
- end_frontend_mesh_for
- pop_scope
- begin_frontend_if -> current_ast_builder
- begin_frontend_if_true -> current_ast_builder
- begin_frontend_if_false -> current_ast_builder
- insert_break_stmt -> current_ast_builder
- create_kernel_return -> current_ast_builder
- insert_continue_stmt -> currnet_ast_builder
- insert_expr_stmt -> current_ast_builder
- begin_func -> current_ast_builder
- end_func
- ~~make_func_call_expr -> Expr::make~~
- ~~expr_atomic_add -> Expr::make~~
- ~~expr_atomic_sub -> Expr::make~~
- ~~expr_atomic_min -> Expr::make~~
- ~~expr_atomic_max -> Expr::make~~
- ~~expr_atomic_bit_and -> Expr::make~~
[...]

PR #4034, opened 2022-01-15
[refactor] Re-export some functions called directly by ASTTransfomer.* as member of ASTBuilder

PGZXB:
Related issue = #4032

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -520,7 +520,6 @@ class ASTTransformer(Builder):
             ast.BitAnd: lambda l, r: l & r,
             ast.MatMult: lambda l, r: l @ r,
         }.get(type(node.op))
-        print(node.left.ptr, node.right.ptr)
         node.ptr = op(node.left.ptr, node.right.ptr)
         return node.ptr
 
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -19,26 +19,26 @@ from taichi import linalg
 
 class ASTTransformer(Builder):
     @staticmethod
-    def build_Name(ctx, node):
+    def build_Name(ast_builder, ctx, node):
         node.ptr = ctx.get_var_by_name(node.id)
         return node.ptr
 
     @staticmethod
-    def build_AnnAssign(ctx, node):
-        build_stmt(ctx, node.value)
-        build_stmt(ctx, node.annotation)
+    def build_AnnAssign(ast_builder, ctx, node):
+        build_stmt(ast_builder, ctx, node.value)
+        build_stmt(ast_builder, ctx, node.annotation)
 
         is_static_assign = isinstance(
             node.value, ast.Call) and node.value.func.ptr is impl.static
 
         node.ptr = ASTTransformer.build_assign_annotated(
-            ctx, node.target, node.value.ptr, is_static_assign,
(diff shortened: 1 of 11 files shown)


python/taichi/lang/ast/ast_transformer.py
```diff
@@ -514,20 +520,21 @@ def build_BinOp(ctx, node):
             ast.BitAnd: lambda l, r: l & r,
             ast.MatMult: lambda l, r: l @ r,
         }.get(type(node.op))
+        print(node.left.ptr, node.right.ptr)
```
strongoier, line 523:
```suggestion
```

strongoier (commented):
Looks great! The title needs to be modified because this PR only reflects the first step of the issue.

@lin-hitonami Is there any reason we make `build_xxx` static methods?

lin-hitonami:
> Looks great! The title needs to be modified because this PR only reflects the first step of the issue.
> 
> @lin-hitonami Is there any reason we make `build_xxx` static methods? I'm wondering whether we can put `ast_builder` as a member variable of `ASTTransformer`. Another option might be to put it under `ctx`, but I'm not sure if that's proper.

Maybe putting it under `ctx` makes the code cleaner? In fact everything inside `ctx` can be moved to `ASTTransformer` if we don't make `build_xxx` static method. cc @ailzhang

ailzhang:
> Maybe putting it under `ctx` makes the code cleaner? In fact everything inside `ctx` can be moved to `ASTTransformer` if we don't make `build_xxx` static method. cc @ailzhang
Yea makes sense, how about putting it in `ctx` first in this PR? I also don't see any blocker for merging `ctx` into `ASTTransformer` but it better stays as a separate PR ;)

taichi/python/export_lang.cpp
```diff
@@ -896,7 +999,7 @@ void export_lang(py::module &m) {
 
   m.def(
       "create_kernel",
-      [&](const std::function<void()> &body, const std::string &name,
+      [&](const std::function<void(Kernel *)> &body, const std::string &name,
```
strongoier, line 1002: Off-topic comment: is this function overriding the other `create_kernel` above?

taichi/program/program.h
```diff
@@ -303,6 +313,11 @@ class Program {
     return program_impl_->allocate_memory_ndarray(alloc_size, result_buffer);
   }
 
+  // Get current_ast_builder
```
strongoier, line 316:
```suggestion
```

taichi/python/export_lang.cpp
```diff
@@ -265,6 +265,95 @@ void export_lang(py::module &m) {
       .def_readwrite("metric_values",
                      &KernelProfileTracedRecord::metric_values);
 
+  // Export ASTBuilder
```
strongoier, line 268: In the future we should not put actual `ASTBuilder` logic in `export_lang`. But it's OK for this PR.

strongoier (approved):
LGTM!

taichi/python/export_lang.cpp
PGZXB: Yes, I think so. I don't know why it is exported twice. In the future, maybe we can remove one in which the `py::gil_scoped_release` isn't used.

PGZXB: > Off-topic comment: is this function overriding the other `create_kernel` above?

Yes, I think so. I don't know why it is exported twice. In the future, maybe we can remove one in which the `py::gil_scoped_release` isn't used.

PR #4037, opened 2022-01-16
[cuda] [refactor] Use cuda graph in command list to defer executions 1/n

qiao-bo:
This PR kicks off the process of moving cuda device operations into CommandList (Unified Device API). The goal is to reduce host-side overhead using a graph (analogous to CommandBuffer in Vulkan) to avoid unnecessary synchronizations. In this PR, we first simulate an immediate buffer fill operation using the submit_sync.


qiao-bo:
close this and use separate PR if continues

PR #4042, opened 2022-01-17
[vulkan] Fix gtmp type

bobcao3:
Also can we have a SPIR-V tag?


ailzhang (approved):
Confirmed that the generated glsl file compiles for me! Thanks!
Note to self: 
```
spirv-cross init_c58_0_k0000_vk_t00.spv > bla.glsl
glslc -fshader-stage=compute -c bla.glsl
```

PR #4043, opened 2022-01-17
[bug] Fix bug that building with TI_WITH_LLVM=OFF will fail

### LINKED ISSUE #3999: [llvm] build without llvm option fails
**Building option -DTI_WITH_LLVM:BOOL=OFF is broken**
Due to some recent changes, this option raises build errors. At this moment, ndarray and JITSession are using "LlvmProgramImpl", which will raise not defined error in the latest build system.

PGZXB:
Related issue = fix #3999

Taichi isn't at a state that can be built easily without LLVM at all. Currently, the `-DTI_WITH_LLVM=OFF` doesn't mean that Taichi will be built without using LLVM at all. Fixing this problem completely in the future version is that we should consider, which is difficult but necessary.

### DIFF
--- a/taichi/jit/jit_session.cpp
+++ b/taichi/jit/jit_session.cpp
@@ -35,10 +35,4 @@ std::unique_ptr<JITSession> JITSession::create(LlvmProgramImpl *llvm_prog,
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


strongoier (approved):
Thanks!

Issue #4019, opened 2022-01-13
Mouse event response has obvious latency


neozhaoliang:
There is an obvious latency when dragging mouse on the gui window, the circle often moves lagging behind the mouse. Here is the test code:

```python
import taichi as ti
ti.init(arch=ti.gpu)

resolution = (600, 360)  # window resolution
pixels = ti.Vector.field(3, ti.f32, shape=resolution)


@ti.kernel
def render(iTime: ti.f32, mouse_x: ti.f32, mouse_y: ti.f32):
    W, H = resolution
    for ix, iy in pixels:
        uv = ti.Vector([(2. * ix / W - 1) * W / H, 2. * iy / H - 1]) * 1.05
        color = ti.Vector([0., 0., 0.])
        x = 1.0
        for _ in ti.static(range(500)):
            x = x * x + 1
            x = ti.sqrt(x)

        m = ti.Vector([(2. * mouse_x - 1) * W / H, 2. * mouse_y - 1]) * 1.05
        if (uv - m).norm_sqr() < 0.1:
            color = ti.Vector([1, 0, 0])

        pixels[ix, iy] = color


if __name__ == "__main__":
    gui = ti.GUI("Test GUI", res=resolution)

    for i in range(1000000):
        gui.get_event()
        if gui.is_pressed(ti.GUI.ESCAPE):
            exit()

        mouse_x, mouse_y = 0.5, 0.5
        if gui.is_pressed(ti.GUI.LMB):
            mouse_x, mouse_y = gui.get_cursor_pos()

        render(i * 0.03, mouse_x, mouse_y)
        gui.set_image(pixels)
        gui.show()
```

mzmzm:
Maybe you can change `gui.get_event()` to `gui.get_event(ti.GUI.PRESS)`. Note that there are a lot of events GUI passing to us, so we need to filter them. I don't know if this is considered a bug, but I will try to fix it.

neozhaoliang:
The problem is still there. See the example code below:

```
import taichi as ti
ti.init(arch=ti.cpu)

# ----------------------------------------------
# global settings

pqr = (3, 3, 7)
resolution = (600, 360)  # window resolution
inf = -1
inf_cos = 1.1  # =1 for paracompact tiling and >1 for noncompact tiling
pi = 3.141592654
max_iterations = 15
black_region_size = 0.065  # control size of the black region

klein_model = False  # if you want to draw the tiling in Klein model
rotate_pattern = True

[...]

PR #4049, opened 2022-01-18
[Lang] [bug] Stop misusing non-template argument types to determine template reinstantiation

mzmzm:
Related issue = #4019

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -296,7 +296,10 @@ class TaichiCallableTemplateMapper:
                                              element_dim] if layout == Layout.SOA else shape[
                                                  -element_dim:]
             return to_taichi_type(arg.dtype), len(shape), element_shape, layout
-        return anno.to_string(),
+        if id(type(arg)) in primitive_types:
+            return anno.to_string()
+        else:
+            return type(arg).__name__,
 
     def extract(self, args):
         extracted = []
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -296,10 +296,9 @@ class TaichiCallableTemplateMapper:
                                              element_dim] if layout == Layout.SOA else shape[
                                                  -element_dim:]
             return to_taichi_type(arg.dtype), len(shape), element_shape, layout
-        if id(type(arg)) in primitive_types:
+        if id(type(arg)) in primitive_types.type_ids:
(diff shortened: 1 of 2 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -311,6 +313,7 @@ def lookup(self, args):
             )
 
         key = self.extract(args)
+        print(key)
```
strongoier, line 316:
```suggestion
```

python/taichi/lang/kernel_impl.py
```diff
@@ -646,6 +649,8 @@ def match_ext_arr(v):
         return has_array
 
     def ensure_compiled(self, *args):
+        print(self.func)
+        print(args)
```
strongoier, line 653:
```suggestion
```

strongoier (changes requested):
This problem should be solved in a more systematic way - only `ti.template()` and `ti.any_arr()` will involve template instantiation, so others should be directly ignored.

python/taichi/lang/kernel_impl.py
```diff
@@ -296,7 +296,11 @@ def extract_arg(arg, anno):
                                              element_dim] if layout == Layout.SOA else shape[
                                                  -element_dim:]
             return to_taichi_type(arg.dtype), len(shape), element_shape, layout
-        return type(arg).__name__,
+        if anno is int:
```
strongoier, line 299: What about using `cook_dtype(anno)` for it?

mzmzm: `cook_dtype` doesn't deal with all data type now, such as `MatrixType`

python/taichi/lang/kernel_impl.py
```diff
@@ -296,7 +296,8 @@ def extract_arg(arg, anno):
                                              element_dim] if layout == Layout.SOA else shape[
                                                  -element_dim:]
             return to_taichi_type(arg.dtype), len(shape), element_shape, layout
-        return type(arg).__name__,
+        # the '#' is just a placeholder
```
strongoier, line 299:
```suggestion
        # Use '#' as a placeholder because other kinds of arguments are not involved in template instantiation
```

mzmzm: e4219f2a3cdc9c18de5b34dfa8b243ff9eb93660

strongoier (approved):
LGTM!

PR #4051, opened 2022-01-18
[Opengl] [cc] Increase the number of arguments allowed in a kernel

### LINKED ISSUE #4050: Raise the limit of argument numbers on all backends to 64
Currently, the limit of argument numbers in a kernel is 8 on OpenGL and cc backends, and 64 on other backends. 
Can we raise the limit on OpenGL and cc backend to 64 as well?

strongoier:
Related issue = close #4050

This PR is a follow-up of #2886, which increases the number of arguments allowed in a kernel for OpenGL and CC backends as well.

### DIFF
--- a/taichi/inc/constants.h
+++ b/taichi/inc/constants.h
@@ -3,11 +3,9 @@
 #include <cstddef>
 
 constexpr int taichi_max_num_indices = 8;
-// legacy: only used in cc and opengl backends
-constexpr int taichi_max_num_args = 8;
-// used in llvm backend: only the first 16 arguments can be ext_arr/any_arr
 // TODO: refine argument passing
-constexpr int taichi_max_num_args_total = 64;
+constexpr int taichi_max_num_args = 64;
+// only the first 16 arguments can be ext_arr/any_arr
 constexpr int taichi_max_num_args_extra = 16;
 constexpr int taichi_max_num_snodes = 1024;
 constexpr int kMaxNumSnodeTreesLlvm = 512;
--- a/taichi/program/context.h
+++ b/taichi/program/context.h
@@ -19,11 +19,11 @@ struct RuntimeContext {
   // - primitive_types
   // - raw ptrs: for external array, or torch-based ndarray
   // - DeviceAllocation*: for taichi ndaray
-  uint64 args[taichi_max_num_args_total];
+  uint64 args[taichi_max_num_args];
   int32 extra_args[taichi_max_num_args_extra][taichi_max_num_indices];
   int32 cpu_thread_id;
   // |is_device_allocation| is true iff args[i] is a DeviceAllocation*.
-  bool is_device_allocation[taichi_max_num_args_total]{false};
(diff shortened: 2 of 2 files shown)


lin-hitonami (commented):
Maybe add a test to make sure it works for all backends?

ailzhang:
Hmmm this might change the layout of arg buffer for opengl aot. Shall we keep it unchanged if it's not critical? (similar for cc backend, we don't have a strong motivation to update it for cc backend yet :P

lin-hitonami:
> Hmmm this might change the layout of arg buffer for opengl aot. Shall we keep it unchanged if it's not critical? (similar for cc backend, we don't have a strong motivation to update it for cc backend yet :P

But if we allow matrices as the argument, 8 arguments seem a bit small?

strongoier:
> Hmmm this might change the layout of arg buffer for opengl aot. Shall we keep it unchanged if it's not critical? (similar for cc backend, we don't have a strong motivation to update it for cc backend yet :P

Sure :-) I can close it for now. I think the request (#4050) mainly aims at making our description in our doc more uniform. We can explicitly specify that opengl and cc backends can hold at most 8 parameters for now. cc @lin-hitonami

k-ye:
BTW, can we guarantee that the Matrix/Vector lifetime exists during the kernel launch (**excluding** its execution). If so, we can just pass in a pointer to the matrix

strongoier:
> BTW, can we guarantee that the Matrix/Vector lifetime exists during the kernel launch (**excluding** its execution). If so, we can just pass in a pointer to the matrix

I'm not sure if I get what you mean. Currently we only have `ti.field` and `ti.ndarray` as global variables, and I don't think we should add more (for example, to make Python-scope vectors and matrices global variables). Accepting an external pointer is the job of `ti.any_arr()` annotation, but it doesn't support matrix operations. The main problem is still the fundamental one - what is the purpose of our types.

k-ye:
> > BTW, can we guarantee that the Matrix/Vector lifetime exists during the kernel launch (**excluding** its execution). If so, we can just pass in a pointer to the matrix
> 
> I'm not sure if I get what you mean. Currently we only have `ti.field` and `ti.ndarray` as global variables, and I don't think we should add more (for example, to make Python-scope vectors and matrices global variables). Accepting an external pointer is the job of `ti.any_arr()` annotation, but it doesn't support matrix operations. The main problem is still the fundamental one - what is the purpose of our types.

We're talking different things. The purpose of expanding the size to 64 is so that we can pass matrix as arguments as well. If a local matrix is stored as a `float*`, then we can just pass this pointer into `Context` (instead of copying every element)

strongoier:
> We're talking different things. The purpose of expanding the size to 64 is so that we can pass matrix as arguments as well. If a local matrix is stored as a `float*`, then we can just pass this pointer into `Context` (instead of copying every element)

Well I guess I skipped the most important sentence in my previous comment. Where is the matrix allocated? I mention global variables because I am thinking about this question.

PR #4060, opened 2022-01-19
[Doc] Update sparse compuation doc

FantasyVR:
Related issue = #


docs/lang/articles/advanced/sparse.md
```diff
@@ -2,100 +2,116 @@
...
-Figure: A swinging "Taichi" pattern represented with a 512x512 sparse grid. The sparse grid has a multi-level *tree* structure.
-White stands for inactive tree nodes, and active tree nodes are darker.
-
-The sparse grid above has the following structure:
-- The grid is divided into 8x8 `block1` containers;
-- Each `block1` container has 4x4 `block2` cells;
-- Each `block2` container has 4x4 `block3` cells;
-- Each `block3` container has 4x4 pixel cells;
-- Each pixel contains an `i32` value `x[i, j]`.
+# Sparse spatial data structures
```
strongoier, line 5: Shall we consistently use "Sparse spatial data structures" in this doc when we refer to this term?

docs/lang/articles/advanced/sparse.md
```diff
@@ -2,100 +2,116 @@
...
-[[Full source code of this animation]](https://github.com/taichi-dev/taichi/blob/master/python/taichi/examples/features/sparse/taichi_sparse.py)
-
-Intuitively, a sparse grid in Taichi allows you to use memory space more wisely, since only tree nodes involved in computation are allocated.
-Now, let's take a step back and think about *why we need sparse grids, how to define them in Taichi, and how to compute on these data structures*.
-
 ## Motivation
 
 High-resolution 2D/3D grids are often needed in large-scale spatial computation, such as physical simulation, rendering, and 3D reconstruction.
-However, these grids tend to consume a huge amount of memory space and computation.
+However, these grids tend to consume a huge amount of memory space and computation if we use dense data structure (see [field](lang/articles/basic/field.md) and [field advanced](lang/articles/advanced/layout.md)).
```
strongoier, line 14: There are many terms here "grid ... if we use dense data structure", "dense grid", "sparse grid", which might confuse readers. Is the concept "grid" coupled with the data structure or not?

docs/lang/articles/advanced/sparse.md
```diff
@@ -2,100 +2,116 @@
...
 oftentimes, they only care about a small fraction of this dense grid since the rest may be empty space (vacuum or air).
 
-In short, the regions of interest in sparse grids may only occupy a small fraction of the whole bounding box.
+<center>
+
+![BVH](https://github.com/FantasyVR/public_files/blob/sparse_computation/taichi/doc/bvh_small.png)
+
+</center>
+
+As shown above, the regions of interest in sparse grids may only occupy a small fraction of the whole bounding box.
```
strongoier, line 24: Shall we add some transition phrases like "for example" to clearly separate it from the context?

docs/lang/articles/advanced/sparse.md
```diff
@@ -2,100 +2,116 @@
...
+
+In fact, the sparse field is a Snode tree shown in the following figure. You could use the struct-for loop to loop over the different levels of the SNode tree like the `print_active` function in `pointer.py`. `for i, j in block` would loop over all active `pointer` Snodes (green grids). `for i, j in pixel` would loop over all active `dense` Snodes (green grids).
+
+<center>
+
+![Pointer Snode Tree](https://github.com/FantasyVR/public_files/blob/sparse_computation/taichi/doc/pointer_tree_small.png)
+
+</center>
 
 While a null pointer can effectively represent an empty sub-tree, at the leaf level using 64 bits to represent the activity
```
strongoier, line 116: This part should be moved under `bitmasked SNode`.

strongoier (commented):
Some preliminary comments :-)

docs/lang/articles/advanced/sparse.md
FantasyVR: This figure is used to show the pointer SNode tree.

docs/lang/articles/advanced/sparse.md
FantasyVR: Yes, good point.

docs/lang/articles/advanced/sparse.md
FantasyVR: `grid` is a word that is common for computer graphics practitioners.  `grid` is the spatial discretization of the simulation world. In this context, there is no specific data structure coupled with `grid`.

docs/lang/articles/advanced/sparse.md
strongoier: Oh I mean paragraphs starting from L116. GitHub shows lines above it, which might be a little confusing :-)

FantasyVR: I see. Actually, this paragraph applies to both `pointer + dense` and `pointer + bitmasked` examples.

docs/lang/articles/advanced/sparse.md
```diff
@@ -2,316 +2,281 @@
...
+    #         field x[3, 2] = 0.000000
+    #         field x[3, 3] = 0.000000
+    #         field x[2, 4] = 2.000000
+    #         field x[2, 5] = 0.000000
+    #         field x[3, 4] = 0.000000
+    #         field x[3, 5] = 0.000000
 ```
+The code snippet above creates an 8x8 sparse grid, with the top-level being 4x4 pointer arrays (line 2 of `pointer.py`),
+and each pointer points to a 2x2 dense block. The `activate` function could automatically activates `block[1,1]` which includes `x[2,3]` and `block[1,2]` which includes `x[2, 4]`. Other pixels of `block[1,1]` (`x[2,2], x[3, 2], x[3, 3]`) and `block[1,2]` (`x[2, 5], x[3, 4], x[3, 5]`) are also implicitly activated if `pixel` is a dense node. Because all pixels in the dense block share the same activity value.
+As the below figure shows, the active blocks and pixels are green. You can write and read the sparse field like a dense field using indices.
```
strongoier, line 97:
```suggestion
You can write and read the sparse field like a dense field using indices. The below figure shows the active blocks and pixels in green.
```

docs/lang/articles/advanced/sparse.md
```diff
@@ -2,316 +2,281 @@
...
-![image](https://raw.githubusercontent.com/taichi-dev/public_files/master/taichi/doc/sparse_grids_2d.png)
+<center>
 
-## Computation on sparse data structures
+![Pointer](https://raw.githubusercontent.com/FantasyVR/public_files/sparse_computation/taichi/doc/pointer_small.png)
 
-### Activation on write
+</center>
 
-When writing to an inactive cell on a sparse data structure, Taichi automatically populates the data structure.
```
strongoier, line 105:
```suggestion
Executing the `activate()` function automatically activates `block[1,1]`, which includes `x[2,3]`, and `block[1,2]`, which includes `x[2,4]`. Other pixels of `block[1,1]` (`x[2,2], x[3,2], x[3,3]`) and `block[1,2]` (`x[2,5], x[3,4], x[3,5]`) are also implicitly activated because all pixels in the dense block share the same activity value.

```

docs/lang/articles/advanced/sparse.md
```diff
@@ -2,316 +2,281 @@
...
 
-For example, when executing `x[2, 3] = 2` on the aforementioned sparse grid `x`,
-Taichi automatically activates `block[1, 1]` so that `pixel[2, 3]` is allocated.
+In fact, the sparse field is a Snode tree shown in the following figure. You could use the struct-for loop to loop over the different levels of the SNode tree like the `print_active` function in `pointer.py`. `for i, j in block` would loop over all active `pointer` Snodes (green grids). `for i, j in pixel` would loop over all active `dense` Snodes (green grids).
```
strongoier, line 106:
```suggestion
In fact, the sparse field is a SNode tree shown in the following figure. You could use the struct-for loop to loop over the different levels of the SNode tree like the `print_active()` function in `pointer.py`. `for i, j in block` would loop over all active `pointer` SNodes. `for i, j in pixel` would loop over all active `dense` SNodes.
```

`Snode -> SNode` should also be applied to the figures.
As they are all green, explicitly stating "green grids" out doesn't seem helpful.

docs/lang/articles/advanced/sparse.md
strongoier: Yes but the main purpose is to introduce `bitmasked`. So I suggest moving all of them under `bitmasked` instead of splitting them in two places.

docs/lang/articles/advanced/sparse.md
```diff
@@ -2,316 +2,281 @@
...
+[Octrees](https://en.wikipedia.org/wiki/Octree) (3D). Since dereferencing pointers is relatively costly on modern computer architectures, compared to quadtrees and octrees, it is more performance-friendly to use shallower trees with larger branching factors.
+[VDB](https://www.openvdb.org/) and [SPGrid](http://pages.cs.wisc.edu/~sifakis/papers/SPGrid.pdf) are such examples.
+In Taichi, programmers can compose data structures similar to VDB and SPGrid with SNodes. The advantage of Taichi sparse spatial data structure are
+1. The sparse data could be used like a dense data structure (accessing using indices)
+2. The sparse data could be looped in parallel.
+3. Automatic memory access optimization.
 
-(Note the changing distribution of active blocks throughout the simulation.)
+![image](https://raw.githubusercontent.com/taichi-dev/public_files/master/taichi/doc/sparse_grids_3d.jpg)
+Figure: A 3D fluid simulation that uses both particles and grids. Left to right: particles, 1x1x1 voxels, 4x4x4 blocks, 16x16x16 blocks.
```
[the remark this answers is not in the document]
FantasyVR: I'm not sure if we should leave this figure.

strongoier:
Would you pull master so that the docsite preview can work properly?

FantasyVR:
> Would you pull master so that the docsite preview can work properly?

After merging the master branch, docsite preview still doesn't work. lol

docs/lang/articles/advanced/sparse.md
```diff
@@ -2,316 +2,288 @@
...
+```python
+print(ti.rescale_index(x, block1, ti.Vector([7, 3]))) # output: [1, 0]
+print(ti.rescale_index(x, block2, [7, 3]))            # output: [3, 1]
+print(ti.rescale_index(x, pixel,  [7, 3]))            # output: [7, 3]
+print(ti.rescale_index(block1, block2, [3, 1]))       # output: [1, 0]
+```
 
-Use `ti.rescale_index` to avoid hard-coding internal information of data structures.
-:::
+As noticed, the output of `ti.rescale_index(x, block1, ti.Vector([7, 3]))` is `[1, 0]`. You could also compute the `block1` index given `pixel` index `[7, 3]` as `[7//2//2, 3//2//2]`. However, doing so couples computation code with the internal configuration of data structures (in this case, the size of `block1` containers). Use `ti.rescale_index` to avoid hard-coding internal information of data structures.
```
strongoier, line 281:
```suggestion
Regarding line 1, you can also compute the `block1` index given `pixel` index `[7, 3]` as `[7//2//2, 3//2//2]`. However, doing so couples computation code with the internal configuration of data structures (in this case, the size of `block1` containers). By using `ti.rescale_index()`, you can avoid hard-coding internal information of data structures.
```

strongoier (approved):
Looks great! Only some writing nits..

BTW you may need to upload the figures to https://github.com/taichi-dev/public_files?

PR #4069, opened 2022-01-20
[opengl] Make sure ndarray arg bind indices are sorted.

ailzhang:
We have reserved bind idx 0~3 for root, gtmp, args and runtime,
so let's force generated bind_idx for array args to be `arg_id + 4`.

Related issue = #


k-ye (approved):
Hah, LGTM!

PR #4070, opened 2022-01-20
[aot] [opengl] Provide range_hint for range_for offloaded tasks in

ailzhang:
* #4071
* __->__ #4070

opengl backend.

For range_for offloaded tasks, we try our best to provide some hints
about the range.

- If the range is known at compile time, it shows as

```
"type": "range_for",
"range_hint": "8192",
```

- If the range is range of a ndarray, it shows as

```
"type": "range_for",
"range_hint": "arg <X>",  # X is ndarray's arg_id
```

For other ranges that's only known at runtime, we currently don't handle
them.

note to self: I actually hacked a version to print out the IR of the
end_stmt of OffloadStmt. Although more accurate, it's not easy to parse
so I just abandoned it. prototype at
https://github.com/taichi-dev/taichi/compare/master...ailzhang:range_hint_aot?expand=1

### DIFF
--- a/taichi/aot/module_data.h
+++ b/taichi/aot/module_data.h
@@ -30,12 +30,13 @@ struct CompiledFieldData {
 
 struct CompiledOffloadedTask {
   std::string type;
+  std::string range_hint;
   std::string name;
   // Do we need to inline the source code?
   std::string source_path;
   int gpu_block_size{0};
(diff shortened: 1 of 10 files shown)


k-ye (approved):
LGTM!

PR #4079, opened 2022-01-20
[bug] Fix warnings on external functions on windows

### LINKED ISSUE #4074: Many taichi APIs are non-taichi functions
It seems that taichi 0.8.10 produces a warning when external functions are called. The problem is that some taichi APIs like `ti.pow` and `ti.random` are non-taichi functions and using them in a kernel results in warnings. Is this an intended behavior?

For example, for the code:
```python
import taichi as ti

ti.init(arch=ti.gpu)

a = ti.field(ti.f32, shape=3)


@ti.kernel
def f():
    for i in a:
        a[i] = ti.pow(ti.random(), 2)


if __name__ == '__main__':
    f()
```
The following warnings are produced:
```
a.py:11: UserWarning: Calling non-taichi function "ti.random". Scope inside the function is not processed by the Taichi AST transformer. The function may not work as expected. Proceed with caution! Maybe you can consider turning it into a @ti.func?
  a[i] = ti.pow(ti.random(), 2)
a.py:11: UserWarning: Calling non-taichi function "ti.pow". Scope inside the function is not processed by the Taichi AST transformer. The function may not work as expected. Proceed with caution! Maybe you can consider turning it into a @ti.func?
  a[i] = ti.pow(ti.random(), 2)
```

lin-hitonami:
Related issue = close #4074

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -339,8 +339,9 @@ class ASTTransformer(Builder):
         if hasattr(func, "_is_taichi_function") or hasattr(
                 func, "_is_wrapped_kernel"):  # taichi func/kernel
             return
-        if hasattr(func,
-                   "__module__") and func.__module__.startswith("taichi."):
+        if hasattr(
+                func, "__module__"
+        ) and func.__module__ and func.__module__.startswith("taichi."):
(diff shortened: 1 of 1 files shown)


ailzhang (approved):
Thanks for fixing it!

PR #4080, opened 2022-01-20
[Bug] [lang] Ban redefinition of template and matrix arguments in Taichi kernel

### LINKED ISSUE #4075: Template arguments of Taichi kernels can be redefined once
**Describe the bug**
Template arguments of Taichi kernels can be redefined once because they are stored inside global variable dict.

**To Reproduce**

```python
import taichi as ti
ti.init()

a = ti.field(ti.i32, shape=(10, ))

@ti.kernel
def foo(t: ti.template()):
    t = 123
    print(t)

foo(a)

```

**Log/Screenshots**

```
123
```

lin-hitonami:
Related issue = close #4075 
The matrix type can be redefined as well, so I also fixed it in this PR.


ailzhang (approved):
LGTM!

Issue #4058, opened 2022-01-19
Build benchmark suites for Taichi


yolo2themoon:
**Concisely describe the proposed feature**
We want to build a variety of benchmark suites to analyze the performance of Taichi, beginning with adding micro-benchmarks to cover some of Taichi's features. The migration to a new version with more benchmarks is currently underway.

**Describe the solution you'd like**
### Suite: MicroBenchmarks
- [ ] A minimal example with BenchmarkItem: `DataType`, `DataSize`, `ResultType` and BenchmarkPlan: `Fill` 
#4031
- [ ] Add a new version run.py and suite_microbenchmarks.py
- [ ] BenchmarkItem: `Container`, Benchmark case: `filed_sparse`
- [ ] BenchmarkItem: `Atomic`, BenchmarkPlan: `Reduction`
- [ ] BenchmarkItem: `Scatter`, BenchmarkPlan: `Stencil2D`
- [ ] BenchmarkPlan: `Saxpy`
- [ ] misc cases

**Additional comments**

PR #4081, opened 2022-01-20
[misc] Add containers and end-to-end result to micro-benchmarks

yolo2themoon:
Related issue = #4058 #4031

### DIFF
--- a/benchmarks/microbenchmarks/_plan.py
+++ b/benchmarks/microbenchmarks/_plan.py
@@ -1,9 +1,9 @@
 import itertools
 
-import taichi as ti
-
 from microbenchmarks._result import ResultType
 from microbenchmarks._utils import get_ti_arch, tags2name
+
+import taichi as ti
 
 
 class BenchmarkPlan:
--- a/benchmarks/microbenchmarks/_plan.py
+++ b/benchmarks/microbenchmarks/_plan.py
@@ -1,9 +1,9 @@
 import itertools
 
+import taichi as ti
+
 from microbenchmarks._result import ResultType
 from microbenchmarks._utils import get_ti_arch, tags2name
-
-import taichi as ti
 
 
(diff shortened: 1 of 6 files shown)


benchmarks/microbenchmarks/_result.py
```diff
@@ -18,11 +31,17 @@ class ResultType(BenchmarkItem):
     name = 'get_result'
```
lin-hitonami, line 31: Maybe change to `MeasurerType`?

yolo2themoon: Awesome! How about renaming it as `MetricType`
it is possible to extend the observed  metrics beyond the `elapsed time` (kernel and e2e)

benchmarks/microbenchmarks/_plan.py
```diff
@@ -49,10 +49,13 @@ def _get_kwargs(self, tags):
...
         for tag in tags:
-            if ResultType.init_taichi(arch, tag):
+            if MetricType.init_taichi(arch, tag):
                 return True
         return False
 
     def _run_func(self, tags: list):
-        return self.func(self.arch, self.basic_repeat_times,
-                         **self._get_kwargs(tags))
+        for tag in tags:
```
lin-hitonami, line 57: Maybe we can generate `kwargs` in function `run`, and then define a item `func_item` to find the function in `func_lut`, then we can get the function by `self.func_lut[kwargs[func_item]]`?

benchmarks/microbenchmarks/_plan.py
```diff
@@ -49,10 +49,13 @@ def _get_kwargs(self, tags):
 
     def _init_taichi(self, arch, tags):
         for tag in tags:
```
lin-hitonami, line 51: same here

lin-hitonami (approved):

Issue #4035, opened 2022-01-15
Support type specification and polymorphism in internal functions


re-xyr:
It will be better if we are able to implement more operations (unary/binary ops) as internal functions instead of hard-coded operations. Because these operations are polymorphic on argument types (size & int/float) and perform implicit casts, we need to extend the functionalities of internal functions to allow type specification and polymorphism.

(cc @strongoier)

re-xyr:
Some points I gathered till now:

- Eventually we'd like to support `Unary/Binary/TernaryOp`s with funccalls.
  - Unary and binary operators all have multiple type signatures.
  - Moreover, ternary op's return type is 1) polymorphic and 2) can only be primitive.
  - Both of these suggests that we need polymorphism and also something like "traits" that can be attached to type variables, like `Primitive`, `Integral`, `Floating`, etc.
  - All of these are under the assumption that Taichi will eventually scrap implicit conversion.
- `ExternalTensorShapeAlongAxis` is a very tricky one. *Its typechecking is not checking types.*
  - Instead it requires the operand to be a specific `Expr` construct, namely `ExternalTensorExpression`, which... seems not to have a type.
  - We need to 1) give `ExternalTensorExpression` a type and 2) have an assertion that `ExternalTensorExpression` will be the only construct to ever have this type.
  - Otherwise we could also just make the parameter to be the `arg_id` of the tensor instead of the tensor itself.

In all the current taichi typechecking is a quite atypical one. Some expressions (especially, non-primitive ones) don't even have types and maybe we should fix that first.

re-xyr:
Hot take: If we're unable to give an `Expr` a type, it probably should not be an `Expr` at all, but some separate construct

PR #4090, opened 2022-01-22
[refactor] [ir] Use InternalFuncCall for retrieving thread index

re-xyr:
Related issue = #4035

`global_thread_idx` is the only expr that can be refactored into a funccall without mass refactoring for now.


strongoier (approved):
LGTM!

PR #4095, opened 2022-01-24
[opengl] Use element_size as alignment in root buffer.

### LINKED ISSUE #4089: [opengl] Make sure each field's alignment is equal to its element size in the root SSBO

ailzhang:
Fixes #4089

Also deduplicated the `get_snode_base_address` between compilation and
aot module builder. We can do it only once when calculating root buffer
size.

Related issue = #

### DIFF
--- a/taichi/backends/opengl/struct_opengl.cpp
+++ b/taichi/backends/opengl/struct_opengl.cpp
@@ -18,8 +18,8 @@ OpenglStructCompiler::CompiledResult OpenglStructCompiler::run(SNode &node) {
     generate_types(*n);
   }
   CompiledResult result;
-  result.snode_map = std::move(snode_map_);
   result.root_size = compute_snode_size(node);
+  result.snode_map = std::move(snode_map_);
   result.root_snode_type_name = node.node_type_name;
   return result;
 }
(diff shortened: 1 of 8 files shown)


taichi/backends/opengl/struct_opengl.cpp
```diff
@@ -3,7 +3,37 @@
...
 
 TLANG_NAMESPACE_BEGIN
+namespace {
+int find_children_id(const SNode *snode) {
+  auto parent = snode->parent;
+  for (int i = 0; i < parent->ch.size(); i++) {
+    if (parent->ch[i].get() == snode)
+      return i;
+  }
+  TI_ERROR("Child not found in parent!");
```
k-ye, line 13: return -1; (or this causes a compiler warning?)

taichi/backends/opengl/struct_opengl.cpp
```diff
@@ -94,7 +125,7 @@ size_t OpenglStructCompiler::compute_snode_size(const SNode &snode) {
     ch_size += compute_snode_size(*ch);
   }
   int n = snode.num_cells_per_container;
-  return n * ch_size + opengl_get_snode_meta_size(snode);
+  return alignment_bytes + n * ch_size + opengl_get_snode_meta_size(snode);
```
k-ye, line 128: Can we guarantee that the return value is also aligned with "element_stride"?

ailzhang: I simplified the code a bit - now both mem_offset_in_root for the snode and return value here are guaranteed to be multiple of its element_stride.

k-ye (approved):
LGTM!

PR #4098, opened 2022-01-24
[refactor] [ir] Refactor ExternalFuncCallExpression into a frontend stmt

re-xyr:
Related issue = #4035 

`ExternalFuncCallExpression`s do not have return values and are always wrapped in `FrontendExprStmt`s. Therefore it'll be better if it is just a stmt by itself.


strongoier (approved):
LGTM!

PR #4105, opened 2022-01-25
[javascript] Find and use emscripten cmake toolchain in CMakeLists.txt (JS 2/n)

AmesingFlank:
Related issue = #3781

Assuming that the path to the emscripten is stored in the `EMSCRIPTEN_DIR` environment variable, this PR finds the cmake toolchain files needed to build taichi with emscripten.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,9 @@
 
 cmake_minimum_required(VERSION 3.12)
 
-if((DEFINED TI_EMSCRIPTENED) AND TI_EMSCRIPTENED)
+option(TI_EMSCRIPTENED "Build using emscripten" OFF)
+
+if(TI_EMSCRIPTENED)
   if(DEFINED ENV{EMSCRIPTEN_DIR})
       set(EMSCRIPTEN_DIR $ENV{EMSCRIPTEN_DIR})
       # Tool chain file must be specified before the `project()`
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,17 @@
 
 cmake_minimum_required(VERSION 3.12)
 
+if((DEFINED TI_EMSCRIPTENED) AND TI_EMSCRIPTENED)
+  if(DEFINED ENV{EMSCRIPTEN_DIR})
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -4,6 +4,17 @@
 
 cmake_minimum_required(VERSION 3.12)
 
+if((DEFINED TI_EMSCRIPTENED) AND TI_EMSCRIPTENED)
```
bobcao3, line 7: Maybe use cmake's like `option` command?

CMakeLists.txt
```diff
@@ -4,6 +4,17 @@
...
+  if(DEFINED ENV{EMSCRIPTEN_DIR})
```
bobcao3, line 8: Maybe consider using CMake variables directly? (i.e. `-DEMSCRIPTEN_DIR=...`) instead of environment variables? Environment variables are quite easy to pollute.

CMakeLists.txt
AmesingFlank: Hmm. I added the option in #4093 already.

AmesingFlank: But since `TaichiCore.cmake` is included after the `project(taichi)` call, that `option` isn't declared here. I didn't wanna redeclare the option either.

CMakeLists.txt
AmesingFlank: I like that idea, but on the other hand we use environment variables everywhere else (`TAICHI_CMAKE_ARGS`, `LLVM_DIR`,... ) so I'm not sure. @ailzhang wdyt?

CMakeLists.txt
bobcao3: If that's the case that option should be moved here probably. Since it's really a platform option not a taichi feature option

AmesingFlank: that makes sense. I'll move it here.

CMakeLists.txt
ailzhang: @AmesingFlank sorry for the late reply! IMHO cmake variables are preferred way and we'll try to unify all cmake args into `TAICHI_CMAKE_ARGS` ;)

feisuzhu:
Closing ancient PR

PR #4106, opened 2022-01-25
[javascript] Support TI_EMSCRIPTENED option as an env var (JS 3/n)

AmesingFlank:
Related issue = #3781 

This PR makes it more convenient to turn on TI_EMSCRIPTENED.

Now you can do `TI_EMSCRIPTENED=1 python setup.py develop --user`, so that you won't have to edit the `TAICHI_CMAKE_ARGS` variable.


ailzhang (approved):
LGTM!

PR #4107, opened 2022-01-25
[javascript] Support compiling taichi under x86 (JS 4/n)

AmesingFlank:
Related issue = #3781

When building with emscripten, the target architecture is x86. This PR enables supports for it.

Surprisingly, this is the only problem due to x86 that I've encountered while building and using an emscriptened version of taichi. This tells us that even for the python version of taichi, we can expect a large portion of it to work if we compile it for x86.


bobcao3 (approved):
seems good, not sure why we are limited  to x64 before, so maybe needs an double check

PR #4108, opened 2022-01-25
[javascript] Avoid all usages of glfw/vulkan/volk when TI_EMSCRIPTENED (JS 5/n)

AmesingFlank:
Related issue = #3781

Avoid all usages of glfw/vulkan/volk when TI_EMSCRIPTENED. To build the JS frontend, we will only be using the AOT features in Taichi. However, Taichi's vulkan backend will try to invoke glfw/vulkan/volk stuff even when using AOT. This PR guards that from happening.


bobcao3 (approved):
LGTM

PR #4111, opened 2022-01-25
[test] Replace make_temp_file with tempfile

ailzhang:
* #4114
* #4113
* #4112
* __->__ #4111

Related: #3782

`make_temp_file` can be easily replaced by native python tempfile so
let's just use tempfile.


k-ye (approved):
LGTM!

PR #4115, opened 2022-01-25
[javascript] [cuda] Only spawn memory allocator daemon when CUDA is enabled (JS 8/n)

AmesingFlank:
Related issue = #3781

The daemon is only used for allocation memory in CUDA. (Actually, do we even use that in CUDA anymore, we have switched completely to pre-allocation right?)

### DIFF
--- a/taichi/system/memory_pool.cpp
+++ b/taichi/system/memory_pool.cpp
@@ -134,7 +134,7 @@ void MemoryPool::terminate() {
 }
 
 MemoryPool::~MemoryPool() {
-  if (killed) {
+  if (!killed) {
     terminate();
   }
 }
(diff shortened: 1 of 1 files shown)


qiao-bo:
Memory allocation in CUDA/CPU has been switched to using preallocated. The memory pool is not used IIUC. ;/

AmesingFlank:
> Memory allocation in CUDA/CPU has been switched to using preallocated. The memory pool is not used IIUC. ;/

Right! Thanks for confirming.
Actually I think this memory pool is still used for allocating things like host-side result buffer. But we should consider removing the daemon entirely.
Created #4135 to track this.

ailzhang (approved):
LGTM!

AmesingFlank:
CI on macOS keeps failing.. interesting. Not sure why this change would affect macOS. Will investigate.

feisuzhu:
Closing ancient PR

PR #4116, opened 2022-01-25
[ci] Use GHA workflow to control the concurrency

frostming:
Related issue = #

This replaces the old usage of cancel workflow.

The workflow means only one run with the same `head_ref`, fallbacking to `run_id` will be run at the same time.
See detailed explanation at [GHA doc](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value)


ailzhang (approved):
Awesome! Thanks!

PR #4121, opened 2022-01-25
[spirv] Move external arrays into seperate buffers

bobcao3:

First step towards ndarray on SPIRV backends.

Also fixed a few small things to speed up compilation a small bit.

cc: @xwang186

### DIFF
--- a/taichi/backends/vulkan/runtime.cpp
+++ b/taichi/backends/vulkan/runtime.cpp
@@ -78,7 +78,7 @@ class HostDeviceContextBlitter {
       const auto dt = arg.dt;
       char *device_ptr = device_base + arg.offset_in_mem;
       do {
-        if (arg.is_array) {
+        if (arg.is_array && arg.stride) {
           DeviceAllocation buffer = ext_arrays.at(i);
           char *const device_arr_ptr =
               reinterpret_cast<char *>(device_->map(buffer));
@@ -134,7 +134,7 @@ class HostDeviceContextBlitter {
     if (!require_sync) {
       for (int i = 0; i < ctx_attribs_->args().size(); ++i) {
         const auto &arg = ctx_attribs_->args()[i];
-        if (arg.is_array) {
+        if (arg.is_array && arg.stride) {
           require_sync = true;
         }
       }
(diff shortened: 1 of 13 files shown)


taichi/backends/vulkan/runtime.cpp
```diff
@@ -427,9 +441,24 @@ void VkRuntime::launch_kernel(KernelHandle handle, RuntimeContext *host_ctx) {
       &ti_kernel->ti_kernel_attribs().ctx_attribs, host_ctx, device_,
       host_result_buffer_, ctx_buffer.get(), ctx_buffer_host.get());
 
+  std::unordered_map<int, DeviceAllocation> ext_arrays;
```
ailzhang, line 444: FYI in opengl we don't redo device allocation if the size doesn't change. maybe it helps with perf a little bit? https://github.com/taichi-dev/taichi/blob/master/taichi/backends/opengl/opengl_api.cpp#L400 (can be a followup ofc)

ailzhang (approved):
Awesome thanks!

taichi/program/program.h
```diff
@@ -5,6 +5,7 @@
 #include <functional>
 #include <optional>
 #include <atomic>
+#include <stack>
```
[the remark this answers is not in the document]
bobcao3: It's used in this class

taichi/backends/vulkan/runtime.cpp
bobcao3: Let's leave it for later, the balance between memory usage and allocation performance is hard

taichi/backends/vulkan/runtime.cpp
```diff
@@ -167,8 +178,8 @@ class HostDeviceContextBlitter {
       const auto dt = ret.dt;
       do {
         if (ret.is_array) {
-          void *host_ptr = host_ctx_->get_arg<void *>(i);
-          std::memcpy(host_ptr, device_ptr, ret.stride);
+          // void *host_ptr = host_ctx_->get_arg<void *>(i);
+          // std::memcpy(host_ptr, device_ptr, ret.stride);
```
xwang186, line 182: We could remove useless comments in the codes

taichi/codegen/spirv/kernel_utils.h
```diff
@@ -20,16 +20,11 @@ namespace spirv {
  * Per offloaded task attributes.
  */
 struct TaskAttributes {
-  enum class BufferType {
-    Root,
-    GlobalTmps,
-    Context,
-    ListGen,
-  };
+  enum class BufferType { Root, GlobalTmps, Context, ListGen, ExtArr };
```
k-ye, line 23: OOC, do we really support ListGen buffer in SPIR-V codegen..?

k-ye (approved):
LGTM! Please resolve @xwang186 's comment :-)

taichi/codegen/spirv/kernel_utils.h
bobcao3: We have very low level all dense->bitmasked support

PR #4126, opened 2022-01-25
[misc] Fix the changelog generator to only count current branch commits

### LINKED ISSUE #4118: make_changelog.py doesn't work well with cherry-picked branch
To repro, please checkout and build `rc-v0.8.11` from source and then run
```
python misc/make_changelog.py
```
Then you'll notice that the generated changelog contains commits not in this branch. 

cc: @frostming would you mind help taking a look?

frostming:
Related issue = Fix #4118

### DIFF
--- a/misc/make_changelog.py
+++ b/misc/make_changelog.py
@@ -17,7 +17,7 @@ def load_pr_tags():
     return details
 
 
-def main(ver='master', repo_dir='.'):
+def main(ver=None, repo_dir='.'):
     g = Repo(repo_dir)
     commits_with_tags = set([tag.commit for tag in g.tags])
     commits = list(g.iter_commits(ver, max_count=200))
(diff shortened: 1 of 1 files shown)


misc/make_changelog.py
```diff
@@ -33,11 +33,8 @@ def format(c):
 
     for i, c in enumerate(commits):
         s = format(c)
-        if c in commits_with_tags:
-            if i == 0:
-                continue
```
frostming, line 38: Not sure why it skips the first tagged commit, but it should be included in the changelog. For example:

```
change 1  [v0.8.10]
change 2
change 3
change 4  [v0.8.9]
```
The resulted changelog should be 
```
change 1
change 2
change 3
```

ailzhang: I guess it was used to skipping version bump commits? @strongoier 
but yea it's okay to keep them in the full changelog as well.

frostming: If that is the case we can make the parsing more explicit: only keep commits with tags, and don't include tags in commits like bump version.

ailzhang (approved):
Thanks!!

misc/make_changelog.py
strongoier: I don't really remember that because that code was there before I made changes. But it seems that our changelog doesn't miss any commit.

PR #4127, opened 2022-01-25
[build] Handle empty TAICHI_EMBIND_SOURCE

ailzhang:
Empty `TAICHI_EMBIND_SOURCE` works on cmake 3.22 but not 3.19 so let's
just condition on it.

Related issue = #


qiao-bo (approved):
LGTM! Thanks for the quick fix!

AmesingFlank (approved):
Sorry, my bad. Thanks for the fix

PR #4131, opened 2022-01-25
[refactor] Export some functions which depend on current_ast_builder() as members of ASTBuilder

PGZXB:
Related issue = #4032

Export functions mentioned in #4032 -> Remove old useless global exported functions

In the future I will put actual `ASTBuilder` logic in class `ASTBuilder` body properly instead of export_lang.


strongoier (approved):
LGTM!

PR #4132, opened 2022-01-25
[ci] Fix approx in autodiff example test

ailzhang:
Fixing broken master.

 #3782 didn't catch this since it was on an older master base
commit.

Related issue = #


qiao-bo (approved):
Thanks

PR #4137, opened 2022-01-26
[Refactor] Do not expose internal function in field, exception, expr, any_array , _ndrange, _ndarray

mzmzm:
Related issue = #3782


python/taichi/lang/_ndrange.py
```diff
@@ -49,3 +49,6 @@ def __init__(self, r):
     def __iter__(self):
         for ind in self.r:
             yield _IntermediateMatrix(len(ind), 1, list(ind))
+
+
+__all__ = ['ndrange']
```
k-ye, line 54: OOC, can we make `ndrange` a function? (It's a code smell issue: class should capitalize its first letter)

mzmzm: Do you mean to add a new API `ndrange` and change `class ndrange` to `class Ndrange`? If so, we have `ti.ndrange` and `ti.Ndrange` to use.

k-ye: I mean can we provide `ndrange` as a function, which can be implemented using some internal helper class like `class _Ndrange`. Users would still only need to use `ti.ndrange`

k-ye: Anyway, that's a question for future works :-)

mzmzm: f70cc8ca95706f8bd61c84b5330f8d9c8b5e880d

k-ye (approved):
lgtm!

PR #4140, opened 2022-01-26
[Refactor] Prevent modules in lang being wild imported and exposed

qiao-bo:
Related issue = #3782 

This PR prevents the modules in `lang/` being exposed such as ` 'kernel_arguments', 'kernel_impl', 'ops' etc`, except `ti.tape` and `ti.sort`.


lin-hitonami (approved):
LGTM!

PR #4150, opened 2022-01-26
[refactor] Move functions in __init__ to misc

mzmzm:
Related issue = #3782
There are too many functions in `lang/__init__.py`. I think we should move them to a new python file so that we can clear python api conveniently.


lin-hitonami (approved):
LGTM!

strongoier (approved):
LGTM!

Issue #3847, opened 2021-12-21
Add require_version configuration in ti.init()


strongoier:
There is a legacy `require_version` function in Taichi:
https://github.com/taichi-dev/taichi/blob/7bae9c77f69e4215e809ee6ddcfeb9f45c724842/python/taichi/_lib/utils.py#L127-L139
This is not exposed to users anymore, which breaks the following code in the *taichi_elements* project: https://github.com/taichi-dev/taichi_elements/blob/818b88ef2e65bec474f19b156f49dba3727aa108/engine/mpm_solver.py#L10

In normal Python projects, versions of dependencies are ensured with `requirements.txt`. However, many existing Taichi programs are standalone files. To give these programs an opportunity to set version requirements, we hope to move the functionality of `require_version` into `ti.init()` by adding a `require_version` configuration. After having this configuration, the broken usage of `require_version` in the *taichi_elements* project needs to be fixed accordingly.

tczhangzhi:
This sounds reasonable and I'd like to be assigned to this issue!

tczhangzhi:
My only concern is to remain the old fashion API after modifying `ti.init()`, otherwise it would be a breaking change and we should note it in the document?

strongoier:
Thanks for your interest in this issue! I have assigned it to you.

Regarding your concern, we hope to keep a stable set of APIs starting from Taichi v1.0 and we are on our way to pick that set. In this process, there must be some breaking changes and we will explicitly state them in our future release notes.

ailzhang:
Closing via #4151

PR #4151, opened 2022-01-26
[Refactor] Add require_version configuration in ti.init()

tczhangzhi:
Related issue = #3847


tczhangzhi:
I suppose passing a tuple into `taichi.init` is not in line with the user's intuition to set required_version. Thus, I slightly modify the `required_version` API to permit a version string as input while refactoring `taichi.init`. A use case is as follows:
```
import taichi

taichi.init(require_version='0.8.0')
```
instead of:
```
import taichi

taichi.init(require_version=(0,8,0))
```
If everything is okay, I would finish a unit test and fix the broken usage of require_version in the taichi_elements project accordingly. Otherwise, I would like to improve the current API following other contributors' advice.

python/taichi/lang/__init__.py
```diff
@@ -566,6 +568,39 @@ def init(arch=None,
                                                 daemon=True)
         check_version_thread.start()
 
+    # Check if installed version meets the requirements.
+    if require_version:
+        # Extract version number part (i.e. toss any revision / hash parts).
```
ailzhang, line 573: let's still keep this in a separate function?

ailzhang (approved):
Thanks a lot for contributing! 
I really like the idea of using a string instead of tuple, another really nice to allow user only specifying `<major>.<minor>` as well. ;) 
Btw this file has been refactored to `python/taichi/lang/misc.py` in #4150 but resolving the merge conflict should be easy.

python/taichi/lang/__init__.py
tczhangzhi: okay

tczhangzhi:
I recently finished further commits following your comments. Thanks for your review and happy new year! (*^▽^*)

k-ye:
Thanks for the contribution! Would you resolve the merge conflicts when you have time, happy new year!

tczhangzhi:
DONE. It is clear now (*^▽^*)

python/taichi/lang/misc.py
```diff
@@ -179,11 +179,54 @@ def prepare_sandbox():
...
+        int(_ti_core.get_version_minor()),
+        int(_ti_core.get_version_patch()),
+    ]
+    # Match installed version and required version.
+    match = major == versions[0] and (
+        minor < versions[1] or minor == versions[1] and patch <= versions[2])
+
+    if not match:
+        print(f"Taichi version mismatch. required >= {major}.{minor}.{patch}")
+        print("Installed =", _ti_core.get_version_string())
```
ailzhang, line 220: nit: can we unify the two prints to one and put them directly in the Exception message below?

python/taichi/lang/misc.py
```diff
@@ -208,6 +252,10 @@ def init(arch=None,
     # Check version for users every 7 days if not disabled by users.
     _version_check.start_version_check_thread()
 
+    # Check if installed version meets the requirements.
+    if not require_version is None:
```
ailzhang, line 256:
```suggestion
    if require_version is not None:
```

ailzhang (changes requested):
Thanks! Only have a few nits left, would you mind also fix issues thrown by `pylint python/taichi/ --disable=all --enable=$(python scripts/generate_pylint_tags.py)` to pass the static checks?

tczhangzhi:
Sorry for the late reply. Updated (ง •̀_•́)ง If there are any other comments, I will follow as soon as possible!

ailzhang (approved):
Thanks!

PR #4154, opened 2022-01-26
[autodiff] Refactor the IB identification and optimize the checker for global atomics and purely nested loops

erizmr:
Related issue = #
Refactor the IB identification, remove the abused `gather_statements`.
Optimize the IB checker for global atomics and purely nested loops to avoid stack overflow of `ad_stack`.

### DIFF
--- a/taichi/transforms/auto_diff.cpp
+++ b/taichi/transforms/auto_diff.cpp
@@ -8,6 +8,68 @@
 #include <typeinfo>
 
 TLANG_NAMESPACE_BEGIN
+class IndependentBlocksJudger : public BasicStmtVisitor{
+   public:
+    using BasicStmtVisitor::visit;
+
+    void visit(LocalLoadStmt *stmt) override{
+      for (auto &lane : stmt->src.data) {
+        touched_allocas.insert(lane.var->as<AllocaStmt>());
+      }
+    }
+
+    void visit(LocalStoreStmt *stmt) override{
+      touched_allocas.insert(stmt->dest->as<AllocaStmt>());
+    }
+
+    void visit(AtomicOpStmt *stmt) override{
+      if(is_inside_loop) return;
+      TI_ASSERT(stmt->dest->is<GlobalPtrStmt>());
+      for (const auto &node :
+            stmt->dest->cast<GlobalPtrStmt>()->snodes.data) {
+        if (node->has_grad()) {
+          qualified_atomics = false;
+          break;
+        }
+      }
+    }
+
+    void visit(RangeForStmt *stmt) override {
+      inner_most_loop = false;
+      is_inside_loop = true;
+      stmt->body->accept(this);
+      is_inside_loop = false;
+    }
+
+    static bool run(IRNode *root) {
+    IndependentBlocksJudger Judger;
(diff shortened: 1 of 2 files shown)


taichi/transforms/auto_diff.cpp
```diff
@@ -8,6 +8,80 @@
...
+  }
+
+  static bool run(IRNode *root) {
+    IndependentBlocksJudger Judger;
+    Block *block = root->as<Block>();
+    root->accept(&Judger);
+    for (const auto &alloca : Judger.touched_allocas_) {
+      // Test if the alloca belongs to the current block
+      bool belong_to_this_block = false;
+      for (auto b = alloca->parent; b; b = b->parent_block()) {
```
ailzhang, line 57: IIUC this can be replaced by a std::any? (or even simpler, make it a std::all and merge the logic of `qualified_local_` below?

ailzhang (commented):
Nice work!

ailzhang (commented):
Also it'd be awesome if we can add some tests - python tests would be nice. If it's hard, maybe add a few C++ tests for the new IndependentBlocksJudger behavior?

taichi/transforms/auto_diff.cpp
```diff
@@ -8,6 +8,80 @@
...
+    for (auto &lane : stmt->src.data) {
+      touched_allocas_.insert(lane.var->as<AllocaStmt>());
+    }
+  }
+
+  void visit(LocalStoreStmt *stmt) override {
+    touched_allocas_.insert(stmt->dest->as<AllocaStmt>());
+  }
+
+  void visit(AtomicOpStmt *stmt) override {
```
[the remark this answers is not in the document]
erizmr: Does it mean we try to raise an error when the Global Data Access rule is violated?

erizmr:
> Also it'd be awesome if we can add some tests - python tests would be nice. If it's hard, maybe add a few C++ tests for the new IndependentBlocksJudger behavior?

Sure. I can add some python tests.

taichi/transforms/auto_diff.cpp
erizmr: Have made it simpler.

taichi/transforms/auto_diff.cpp
```diff
@@ -8,6 +8,81 @@
...
+    // local load/store stmt checks
+    for (auto b = block->parent_block(); b; b = b->parent_block()) {
+      if (b)
+        outside_blocks.insert(b);
+    }
+    for (const auto &alloca : Judger.touched_allocas_) {
+      // Test if the alloca belongs to the current block
+      if (outside_blocks.find(alloca->parent) != outside_blocks.end()) {
+        // This block is not an IB since it loads/modifies outside variables
+        Judger.qualified_local_ = false;
```
ailzhang, line 65: Correct me if I'm wrong, do we still need to save qualified_local_ ? ;) we can early return false here right? ;)

erizmr: Good catch! We can definitely removed it. :-P

ailzhang (approved):
Awesome!! Thanks!

PR #4169, opened 2022-01-27
[misc] Add atomic operators to micro-benchmarks

yolo2themoon:
Related issue = #4058


benchmarks/microbenchmarks/_plan.py
```diff
@@ -62,3 +64,26 @@ def _get_kwargs(self, tags):
...
+
+    def _remove_comflict_items(self):
+        remove_list = []
+        item_name_list = [item.name for item in self.items]
+        #logical_atomic with float_type
+        if set([AtomicOps.name, DataType.name]).issubset(item_name_list):
+            atomic_item = self._get_item(AtomicOps.name)
+            dytpe_item = self._get_item(DataType.name)
+            for case, plan in self.plan.items():
+                atomic_op_tag = atomic_item.get_intersection(plan['tags'])
```
lin-hitonami, line 82: Why not use a dictionary instead of list? If we use dictionary then we do not need to do the intersection.

lin-hitonami: IIUC we can use the dictionary from `_get_kwargs` ?

yolo2themoon: Thanks for the suggestion, the code looks much clearer in 42650e1

```
        if set([AtomicOps.name, DataType.name]).issubset(self.items.keys()):
            for name, case in self.plan.items():
                kwargs = self._get_kwargs(case['tags'])
                atomic_tag = self.items[AtomicOps.name].tag(kwargs[AtomicOps.name])
                dtype_tag = self.items[DataType.name].tag(kwargs[DataType.name])
                if not AtomicOps.is_supported_type(atomic_tag, dtype_tag):
                    remove_list.append(name)
```

benchmarks/microbenchmarks/_plan.py
```diff
@@ -59,6 +61,22 @@ def run(self):
     def _get_kwargs(self, tags):
         kwargs = {}
         tags = tags[1:]  # tags = [case_name, item1_tag, item2_tag, ...]
-        for item, tag in zip(self.items, tags):
+        for item, tag in zip(self.items.values(), tags):
             kwargs[item.name] = item.impl(tag)
         return kwargs
+
+    def _remove_comflict_items(self):
```
lin-hitonami, line 68: Should it be `_remove_conflict_items`?

benchmarks/microbenchmarks/_plan.py
```diff
@@ -61,7 +61,7 @@ def run(self):
     def _get_kwargs(self, tags):
         kwargs = {}
         tags = tags[1:]  # tags = [case_name, item1_tag, item2_tag, ...]
-        for item, tag in zip(self.items, tags):
+        for item, tag in zip(self.items.values(), tags):
             kwargs[item.name] = item.impl(tag)
```
lin-hitonami, line 65: If we write a new function like this function with this line changed to `kwargs[item.name] = tag` then we may not need `self.items[xxx].tag()`?

yolo2themoon: Got it, and fixed in fecd219

lin-hitonami (approved):
LGTM!

PR #4171, opened 2022-01-27
[opengl] Use && instead of and in C++ code

AmesingFlank:


ailzhang (approved):
lol thanks for fixing it!

PR #4172, opened 2022-01-28
[Refactor] Move core_vec(i) to gui and hide

strongoier:
Related issue = #3782

This PR serves as the first step towards putting functions under `tools.util` in correct places.


ailzhang (approved):
LGTM!

PR #4174, opened 2022-01-28
[Refactor] Clean up helper functions in tools.util

strongoier:
Related issue = #3782

This PR serves as the second step towards putting functions under tools.util (see #4172) in correct places.


python/taichi/lang/util.py
```diff
@@ -226,4 +228,69 @@ def wrapped(*args, **kwargs):
...
+
+def obsolete(old, new):
+    """
+    Mark an API as obsolete. Usage:
+
+    sqr = obsolete('ti.sqr(x)', 'x**2')
+    """
+    def wrapped(*args, **kwargs):
+        msg = f'{old} is obsolete. Please use {new} instead.'
+        raise SyntaxError(msg)
```
lin-hitonami, line 286: TaichiSyntaxError?

python/taichi/lang/util.py
```diff
@@ -226,4 +228,69 @@ def wrapped(*args, **kwargs):
...
+    """
+    s = traceback.extract_stack()[:-stacklevel]
+    raw = ''.join(traceback.format_list(s))
+    print(Fore.YELLOW + Style.BRIGHT, end='')
+    print(f'{warning_type.__name__}: {msg}')
+    print(f'\n{raw}')
+    print(Style.RESET_ALL, end='')
+
+
+def deprecated(old, new, warning_type=DeprecationWarning):
```
lin-hitonami, line 249: Do we still use this function?

python/taichi/lang/util.py
```diff
@@ -226,4 +228,69 @@ def wrapped(*args, **kwargs):
...
+            msg = f'{old} is deprecated. Please use {new} instead.'
+            warning(msg, warning_type, stacklevel=2)
+            return foo(*args, **kwargs)
+
+        return wrapped
+
+    return decorator
+
+
+def obsolete(old, new):
```
lin-hitonami, line 278: Can we do this in `__getattr__` as well?

python/taichi/lang/util.py
strongoier: done

python/taichi/lang/util.py
strongoier: No... We can remove this.

python/taichi/lang/util.py
strongoier: Yes but actually I think this can be removed. We don't need so many levels for deprecating things for now and this is only used in one place one year ago.

lin-hitonami (approved):
LGTM!

Issue #4097, opened 2022-01-24
[RFC] Support of TI_EXPORT to control public API visibility


ghuau-innopeak:
With the support of AOT module in Taichi, we can start using the libtaichi_export_core.so in external application. By default, on Linux, all the symbols in the library are `visible`, on Windows, they are all `hidden`. Having the default visibility set to `visible` for a library is a problem in our case as Taichi is using very popular third-party libraries such as `glad` or `volk`.
Some projects might also use those libraries linked statically to the executable resulting in conflicting symbols with the one from taichi. In that case, no error is generated, the loader/linker will just use by priority the symbol from the executable letting Taichi in a "undefined behavior" as some of the function pointers or variable will not be initialized properly :).

The idea would be to add the support of controlling the visibility of the public API like many other projects such as Qt5 by introducing a change to the CMake to force the visibility to `hidden` for everything:

`set(CMAKE_CXX_VISIBILITY_PRESET hidden)`

And then defining some macros compiler/platform specific to implement properly `TI_EXPORT`, so we could use it on classes that should export the API to external libraries, for example:

```
-class VkRuntime {
+class TI_EXPORT VkRuntime {
```

or

```
-class DataType {
+class TI_EXPORT DataType {
```

Those macros could be defined as follow:

```
#if defined _WIN32 || defined __CYGWIN__
  #ifdef BUILDING_DLL
    #ifdef __GNUC__
      #define TI_EXPORT __attribute__ ((dllexport))
    #else
      #define TI_EXPORT __declspec(dllexport) // Note: actually gcc seems to also supports this syntax.
    #endif
  #else
    #ifdef __GNUC__
      #define TI_EXPORT __attribute__ ((dllimport))
    #else
      #define TI_EXPORT __declspec(dllimport) // Note: actually gcc seems to also supports this syntax.
    #endif
  #endif
#else
  #if __GNUC__ >= 4
    #define TI_EXPORT __attribute__ ((visibility ("default")))
  #else
    #define TI_EXPORT
  #endif
#endif
```

**Additional comments**
This will involves changing a lot of files in the project so one question would be: do we already have somewhere an idea of the classes/files that should be public API? Or maybe a list of the folder that should not be there? :), right now, I'm thinking about the following modules:

- `taichi/common`
- `taichi/ir`
- `taichi/system`
- `taichi/aot`
- `taichi/backends`
- `taichi/gui`
- `taichi/ui`
- `taichi/program`

k-ye:
[...]

PR #4177, opened 2022-01-28
[refactor] Add TI_DLL_EXPORT to control symbol visibility

k-ye:
Related issue = #4097

* Renaming `TI_EXPORT` to `TI_DLL_EXPORT` to make its semantics more obvious.
* Changed cmake files so that it is configurable to use `default` or `hidden` symbol visibility.
* Exported `Program`, `Kernel` and `Callable` as a start.

@ghuau-innopeak @xwang186: Note that this PR still uses the `default` visibility. Once this is merged, could you folks help flip to `hidden`, and see if it still works on Android :-)?

### DIFF
--- a/taichi/program/callable.cpp
+++ b/taichi/program/callable.cpp
@@ -4,6 +4,10 @@
 namespace taichi {
 namespace lang {
 
+Callable::Callable() = default;
+
+Callable::~Callable() = default;
+
 int Callable::insert_arg(const DataType &dt, bool is_array) {
   args.emplace_back(dt->get_compute_type(), is_array);
   return (int)args.size() - 1;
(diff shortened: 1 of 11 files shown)


taichi/program/callable.cpp
```diff
@@ -4,6 +4,10 @@
 namespace taichi {
 namespace lang {
 
+Callable::Callable() = default;
```
k-ye, line 7: Necessary to make MSVC happy, because `Callable` has a `std::unique_ptr<FrontendContext>` member, yet `FrontendContext` is forward declared.

bobcao3 (approved):
LGTM

ghuau-innopeak (approved):
LGTM!

Issue #4145, opened 2022-01-26
Towards a strongly-typed taichi


re-xyr:
Taichi is *statically typed*, but not strongly typed. What that means is that taichi has implicit conversion. There are many arguments against implicit conversion:

- It encourages users to ignore the difference between different datatypes, creating higher risks of precision loss and over/underflow.
- Maintaining a set of implicit conversion rules is tedious and error-prone.
- Users are often not exactly sure where the conversion will be inserted, so they can't reliably predict program behavior.
- Only "builtin" operators have implicit conversion; other internal/external functions, or (in the future) user-defined functions do not. This creates a distinction between operators and other functions, which complicates the language's AST structure *and* type system. Should there be no implicit conversions, operators and other functions can share a common implementation.
-  Relying on conversion too much is a code smell and suggests refactoring or rethinking the program structure. Allowing implicit conversion is obfuscating that.

Therefore I propose that taichi moves towards strong typing, i.e. removal of implicit conversion. This is a breaking change so it can happen in conjunction with v1.0 release.

*Apparently there may also be many reasons against this change and I'm happy to hear them.

lin-hitonami:
Implicit type conversion is a useful feature and many programmers like it. Maybe we will lose a portion of users if we do not have it. However, I think we can make implicit conversion a warning, and add an option to treat specified warning as error.

k-ye:
I also agree that we should strike a balance between ease-of-use and type system rigor. That said, we should definitely keep it to a minimum of where implicit conversions are allowed to happen. We should consider identifying all kinds of implicit conversions  in Taichi now, define the allowed cases, and bake them into our spec. Thoughts? @strongoier

re-xyr:
Basically the `BinaryOp` casting now happens like this:

- If the operation is *bitwise* and either side is not *integral*, fail.
- If the operation is *true division* and either side is not *real*, cast it to *default float type*.
- If the type of two sides are not the same,
  - If any side is *custom type*, cast it to its *compute type*.
  - Now resolve the *result type* via [C++ implicit cast rules](https://en.cppreference.com/w/cpp/language/operator_arithmetic#Conversions). Cast both sides to the *result type*.
    - The implementation of this is rather... crude. There's apparently one thing called `std::common_type<T1, T2>::type` while we're using the brute-force `decltype(std::declval<T1>() + std::declval<T2>())`. 

Other casting behaviors include:
- Casting value type to destination type in `StoreStmt` and `AtomicOp`s
[...]

Issue #3982, opened 2022-01-10
Simplify frontend IR and CHI IR


strongoier:
Currently, there are some duplicate or unnecessary expressions / statements in frontend IR. For example, ` FrontendExprStmt` and  `FrontendEvalStmt` behave exactly the same. Furthermore, expressions including `GlobalThreadIndexExpression` are simply calling internal functions, which should be covered by `InternalFuncCallExpression`s. Similar cases also happen to CHI IR.

We need to keep a minimum set of IR to make the code easier to maintain. This issue aims at tracking the progress of the simplification.

PR #4180, opened 2022-01-28
[Lang] [ir] Add general operation abstraction and remove InternalFuncCallExpression

### LINKED ISSUE #4035: Support type specification and polymorphism in internal functions

re-xyr:
Related issue = #3982, #4145, fixes #4035

`Operation` is an abstraction for language internal operations, with support of polymorphic type specifications and describing implicit casting behaviors. Each `Operation` can be thought of a generalized function, i.e. a type signature paired with a `flatten` implementation.

The type signature has these parts:

- `Constraint`s, that are type variables occurring in the signature together with `Trait`s they need to satisfy.
  - `Trait`s are predicates on `DataType`s, like `Primitive`, `Integral`, `Real`.
- parameter types and return type that are `TypeExpr`s, i.e. a type expression containing type variables and can be evaluated to a datatype.
  - To model implicit cast behaviors, a *type-level function* `CommonTypeExpression` is used to find the "least upper bound" of two types according to C++ implicit cast rules mentioned in #4145.

Other than internal funccalls, this construct can be used to reimplement Unary/BinaryOps in the (near?) future.

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -895,7 +895,14 @@ def default_cfg():
 
 def call_internal(name, *args):
     return expr_init(
-        _ti_core.insert_internal_func_call(name, make_expr_group(args)))
+        _ti_core.call_op(getattr(_ti_core.Operation, name),
+                         make_expr_group(args)))
+
+
+def call_test_internal(name, *args):
+    return expr_init(
+        _ti_core.call_op(getattr(_ti_core.TestOperation, name),
+                         make_expr_group(args)))
 
 
 @taichi_scope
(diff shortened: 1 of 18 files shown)


re-xyr:
Only the test involving bit vectorization failed, others all passed. But bit vectorization should be a ChiIR pass so frontend flattening should have nothing to do with that. Plus after this refactoring the flattened ChiIR should actually be the same. I have no idea what is happening.

k-ye:
Could you try to break down this giant into smaller PRs? Thx

feisuzhu:
Closing ancient PR

Issue #3955, opened 2022-01-06
[refactor] Get rid of get_current_program() and global variable current_program


PGZXB:
Get rid of get_current_program in Taichi :
(The following C++ files depend on `lang::get_current_program()`)
- [x] backends/cpu/jit_cpu.cpp
- [x] backends/cuda/jit_cuda.cpp
- [x] ir/expr.cpp
- [x] ir/frontend_ir.cpp
- [x] python/export_lang.cpp
- [x] ui/backends/vulkan/renderable.cpp
- [x] ui/backends/vulkan/renderables/set_image.cpp
- [x] ui/backends/vulkan/app_context.cpp
- [x] program/program.h

`lang::current_ast_builder()` calls `get_current_program()`
So, following files also depend on `get_current_program()`:
- [x] ir/frontend.h
- [x] ir/frontend.cpp
- [x] transforms/constant_fold.cpp

PGZXB:
Over!

Issue #4146, opened 2022-01-26
Possibility of transitioning to top-down typechecking


re-xyr:
Currently taichi's typechecking process is very idiosyncratic:

- `Expression`s are typechecked as soon as they are constructed
- Except `IdExpression`s, which do not have a type until a `FrontendAssignStmt` is constructed for it
- Unlike `Expression`s, `FrontendStmt`s aren't typechecked until compilation

It'll be nice if we could:

- Make the typechecking process top-down
- Combine typechecking and AST lowering pass into an *elaboration* pass
- Give `Expression`s types in the elaboration pass instead

This refactoring is a nice-to-have but not a high priority. Before this refactoring can happen, we should see if we can address these difficulties:

- Some expressions don't have types (`GlobalVariableExpression`, `ExternalTensorExpression`, `LoopIndexExpression` in mesh loops)
- The python code of `Matrix` uses expression `ret_type`s before `lower_ast()` happens

PR #4181, opened 2022-01-28
[refactor] Remove dependency on get_current_program() in lang::BinaryOpExpression

PGZXB:
Related issue = #3955

I think that the method of removing dependency on get_current_program() in lang::BinaryOpExpression is not best, which causes that the `expr_truediv` needs a additional `default_fp` param except `lhs` and `rhs`, but I cannot think of a better idea.


strongoier (commented):
Can we add a parameter for `type_check()` instead?

PGZXB:
> Can we add a parameter for `type_check()` instead?
This is a feasible solution. However, because `type_check()` is virtual function inherited from `Expression`, adding a parameter for`type_check()` causes that most of subclasses' implementation of this function have a unnecessary parameter. Will this cause confusion?

strongoier:
> This is a feasible solution. However, because `type_check()` is virtual function inherited from `Expression`, adding a parameter for`type_check()` causes that most of subclasses' implementation of this function have a unnecessary parameter. Will this causes confusion?

This is OK for now because we will switch into an individual pass for typechecking (https://github.com/taichi-dev/taichi/issues/4146), where we will have easy access to the configurations.

PGZXB:
See #4242

PR #4191, opened 2022-01-29
[bug] [opengl] Process child nodes to compute alignment

ailzhang:
We need to process `snode.ch` in the order of their
`mem_offset_in_parent` so that the computed alignment is correct.

Also noticed the minimization autodiff test had a too strict `rel` (fails randomly on master) so
relaxing it a bit.

Related issue = #


k-ye (approved):
lgtm!

PR #4193, opened 2022-01-29
[doc] Update demo code in readme doc

Tiny-Box:
Related issue = #

[Taichi] version 0.8.11, llvm 10.0.0, commit 9d0973e5, osx, python 3.9.5

When I tried to run taichi with the demo in the readme, I found that the demo code in the readme has not been updated for a long time, and this code does not work in the current version.
After prompting I found the code for fractal.py, which is updated in the readme.


k-ye (approved):
Ouch, thanks for catching this. LGTM!

PR #4201, opened 2022-01-29
[Refactor] Rename tools.util to tools.async_utils and hide functions inside

strongoier:
Related issue = #3782

Now all functions in `tools.util` are async-related and need to be hidden.


ailzhang (approved):
LGTM!

PR #4202, opened 2022-01-29
[spirv] SPIR-V / Vulkan NDArray

bobcao3:


cmake/TaichiCXXFlags.cmake
```diff
@@ -30,7 +30,9 @@ if (WIN32)
         set(CMAKE_CXX_FLAGS
             "${CMAKE_CXX_FLAGS} /Zc:__cplusplus /std:c++17 /bigobj /wd4244 /wd4267 /nologo /Zi /D \"_CRT_SECURE_NO_WARNINGS\" /D \"_ENABLE_EXTENDED_ALIGNED_STORAGE\"")
     else()
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -g -gcodeview -fsized-deallocation -target x86_64-pc-windows-msvc")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -fsized-deallocation -target x86_64-pc-windows-msvc")
```
k-ye, line 33: nits: could you use `target_compile_options`? https://cmake.org/cmake/help/latest/command/target_compile_options.html

taichi/backends/vulkan/runtime.cpp
```diff
@@ -440,57 +447,88 @@ void VkRuntime::launch_kernel(KernelHandle handle, RuntimeContext *host_ctx) {
...
     for (auto &arg : args) {
       if (arg.is_array) {
-        if (arg.stride) {
+        if (host_ctx->is_device_allocation[i]) {
+          // NDArray
+          if (host_ctx->args[i]) {
+            ext_arrays[i] = *(DeviceAllocation *)(host_ctx->args[i]);
+          } else {
+            ext_arrays[i] = kDeviceNullAllocation;
+          }
```
k-ye, line 472: Do we need to update `ext_array_size` here?

taichi/backends/vulkan/runtime.cpp
```diff
@@ -440,57 +447,88 @@ void VkRuntime::launch_kernel(KernelHandle handle, RuntimeContext *host_ctx) {
          /*export_sharing=*/false, AllocUsage::Storage});
   }
 
+  // Create context blitter
   auto ctx_blitter = HostDeviceContextBlitter::maybe_make(
       &ti_kernel->ti_kernel_attribs().ctx_attribs, host_ctx, device_,
       host_result_buffer_, ctx_buffer.get(), ctx_buffer_host.get());
 
   std::unordered_map<int, DeviceAllocation> ext_arrays;
+  std::unordered_map<int, size_t> ext_array_size;
```
k-ye, line 456: I guess `ext_` is no longer the accurate term, because it covers NDarray as well?

taichi/backends/vulkan/runtime.cpp
```diff
@@ -440,57 +447,88 @@ void VkRuntime::launch_kernel(KernelHandle handle, RuntimeContext *host_ctx) {
...
+            ext_arrays[i] = *(DeviceAllocation *)(host_ctx->args[i]);
+          } else {
+            ext_arrays[i] = kDeviceNullAllocation;
+          }
+        } else {
+          // Compute ext arr sizes
+          size_t size = arg.stride;
+
+          for (int ax = 0; ax < 8; ax++) {
+            // FIXME: how and when do we determine the size of ext arrs?
```
ailzhang, line 478: FYI opengl got it from `kernel->args[i].size` but IMHO we should remove that in `Args` class. https://github.com/taichi-dev/taichi/blob/master/taichi/program/callable.h#L22 
But the algorithm here also need to handle the a torch tensor with shape `[1, 0, 1]`  (total size is still 0).

ailzhang (commented):
Great work!

cmake/TaichiCXXFlags.cmake
bobcao3: At this point the TaichiCore targets have not been defined (as this is the CXXFlags cmake file), also this aligns with the flags behavior on other platforms.

taichi/backends/vulkan/runtime.cpp
bobcao3: Nope, `ext_array_size` is only for ext arrs (for host device context blitter). For the runtime, with NDArrays, it does not need to know the buffer size. I will add a comment to clarify this.

taichi/backends/vulkan/runtime.cpp
```diff
@@ -181,8 +188,8 @@ class HostDeviceContextBlitter {
       const auto dt = ret.dt;
       do {
         if (ret.is_array) {
-          // void *host_ptr = host_ctx_->get_arg<void *>(i);
-          // std::memcpy(host_ptr, device_ptr, ret.stride);
+          void *host_ptr = host_ctx_->get_arg<void *>(i);
```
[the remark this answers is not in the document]
bobcao3: Not yet at this point. But we can leave this in: should be easier to add later down the line.

taichi/backends/vulkan/runtime.cpp
k-ye: Can we unify the shape calculation algo here?

k-ye (approved):
LGTM!

Issue #3888, opened 2021-12-27
Add per_vertex_radius for GGUI particles


CitrusDusk:
When drawing particles in a 3D scene using the GGUI system:

`scene.particles(vertices, radius, color, per_vertex_color)`,

currently, it can only draw a group of particles with the same radius, but I want to draw a bunch of particles where each particle has a different radius.

I wish a `per_vertex_radius` parameter could be added so that we can specify the radius for each individual particle (like how we can set the color for each particle using `per_vertex_color`).

qiao-bo:
CC @AmesingFlank

AmesingFlank:
This sounds a like reasonable and useful feature to have. I will try to find some time to add it, and I'd also be happy to review if someone else from the open source community wants to give it a try.

Bl00per72:
I'd like to be assigned to this issue if possible.

AmesingFlank:
> I'd like to be assigned to this issue, if possible.

Sure.

Bl00per72:
Thanks! @AmesingFlank

Bl00per72:
I'm having trouble finding the file in which this code is located. Could you let me know where I should start working? Much appreciated! @AmesingFlank

Bl00per72:
@AmesingFlank I just got around to taking a look at the issue, but I'm not too familiar with the code base and this is my first time contributing to an open source project. So far, I've taken a look at scene.py, where the code for creating particles is, and that led me to take a look at staging_buffer.py, where the utilized functions in particles() are imported from. I understand the structure, but I am not familiar with the concept behind the code. Would you be able to provide me with some guidance on how to approach this problem? It would be greatly appreciated as I would like to provide a solution!

AmesingFlank:
@Bl00per72  Hey! Glad to know that you're getting started and making progress. Here're some suggestions.

Firstly, you'll need modify the vertex shader for rendering particles. (If you aren't familiar with shaders, have a look at this [link](https://learnopengl.com/Getting-started/Shaders)). In taichi, the vertex shader for particles are at 
https://github.com/taichi-dev/taichi/blob/master/python/taichi/shaders/Particles_vk.vert
[...]

PR #4204, opened 2022-01-31
[gui] Implemented per_vertex_radius for particles

emmanuelgarit:
For Particles_vk.vert:
Added uniform input to check if has per vertex radius, added output selected_radius, if has per vertex radius, selected radius is in_texcoord

For renderable_info.h, added has per vertex radius condition

For scene.py, implemented per_vertex_radius into code

Related issue = #3888

### DIFF
--- a/python/taichi/shaders/Particles_vk.vert
+++ b/python/taichi/shaders/Particles_vk.vert
@@ -17,6 +17,7 @@ layout(binding = 0) uniform UBO {
   SceneUBO scene;
   vec3 color;
   int use_per_vertex_color;
+  int has_per_vertex_radius;
   float radius;
   float window_width;
   float window_height;
@@ -26,6 +27,7 @@ ubo;
 
 layout(location = 0) out vec4 pos_camera_space;
 layout(location = 1) out vec4 selected_color;
+layout(location = 2) out vec2 selected_radius;
 
 void main() {
   float distance = length(in_position - ubo.scene.camera_pos);
@@ -41,5 +43,11 @@ void main() {
     selected_color = vec4(ubo.color, 1.0);
   } else {
     selected_color = in_color;
+  }
+
+  if (ubo.has_per_vertex_radius == 0) {
+    selected_radius = vec2(ubo.radius, 0);
+  } else {
+    selected_radius = in_texcoord;
   }
 }
--- a/python/taichi/ui/scene.py
+++ b/python/taichi/ui/scene.py
@@ -7,7 +7,8 @@ from taichi.types.annotations import template
 from taichi.types.primitive_types import f32
 
 from .staging_buffer import (copy_colors_to_vbo, copy_normals_to_vbo,
-                             copy_vertices_to_vbo, get_vbo_field)
+                             copy_vertices_to_vbo, get_vbo_field,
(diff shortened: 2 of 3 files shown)


AmesingFlank:
Hi! Thanks for working on this! 

Have you verified that this works locally on your machine? I think a few things are missing.

For local development installation, please have a look at https://docs.taichi.graphics/lang/articles/contribution/dev_install

python/taichi/shaders/Particles_vk.vert
```diff
@@ -42,4 +44,10 @@ void main() {
   } else {
     selected_color = in_color;
   }
+
+  if (ubo.has_per_vertex_radius == 0) {
+    selected_radius = vec2(ubo.radius, 0);
+  } else {
+    selected_radius = in_texcoord;
+  }
```
AmesingFlank, line 52: Simply outputting `selected_radius` isn't enough. You will also need to actually use that radius (instead of the one in the uniform block) in the fragment shader. You'll also need to edit the .frag file for this.

AmesingFlank, line 52: Simply outputting selected_radius isn't enough. You will also need to actually use that radius (instead of the one in the uniform block) in the fragment shader. You'll also need to edit the .frag file for this.

python/taichi/ui/scene.py
```diff
@@ -125,10 +127,13 @@ def particles(self,
         vbo = get_vbo_field(centers)
         copy_vertices_to_vbo(vbo, centers)
         has_per_vertex_color = per_vertex_color is not None
+        has_per_vertex_radius = per_vertex_radius is not None
         if has_per_vertex_color:
             copy_colors_to_vbo(vbo, per_vertex_color)
+        if has_per_vertex_radius:
+            copy_texcoords_to_vbo(vbo, per_vertex_radius)
```
AmesingFlank, line 134: `copy_texcoords_to_vbo` won't work out of the box here. The `per_vertex_radius` field provided by the user should be a scalar field, but `copy_texcoords_to_vbo` expects a 2D vector field. You'll need to create a new function: something like `copy_per_vertex_radius_to_vbo`.

python/taichi/ui/scene.py
```diff
@@ -125,10 +127,13 @@ def particles(self,
...
         vbo_info = get_field_info(vbo)
-        super().particles(vbo_info, has_per_vertex_color, color, radius)
+        super().particles(vbo_info, has_per_vertex_color, has_per_vertex_radius, color, radius)
```
AmesingFlank, line 136: You haven't updated the `particles` function on the C++ side. Please have a look at 
https://github.com/taichi-dev/taichi/blob/master/taichi/python/export_ggui.cpp

AmesingFlank:
You're definitely going in the right direction, but a few more things need to be changed. Please build taichi locally and verify that your solution works on your machine. Let me know if you need more help.

Bl00per72:
For sure! I didn't build it locally but I will do so once I implement your suggestions. Thanks! @AmesingFlank

k-ye:
Any update here :-)?

feisuzhu:
Closing ancient PR

Issue #3030, opened 2021-09-28
Release the Vulkan backend to macOS


k-ye:

k-ye:
Disengaging this from https://github.com/taichi-dev/taichi/projects/7. We can just release this whenever the feature is ready

PR #4205, opened 2022-01-31
[vulkan] Fix MoltenVK support

bobcao3:
Related issue = https://github.com/taichi-dev/taichi/issues/3030
Depending PR = https://github.com/taichi-dev/taichi/pull/4202/files  (merged)

This fixes the MoltenVK support on mac.

Fixes & improvements include:
- Auto search for clang-11 / clang-10 when host compiler is clang>=12 (higher version of macOS should work without setting CXX env vars)
- Use `dlopen` to load MoltenVK and use Volk (with custom loader) to handle the rest of loading. This avoids the `@rpath` thing on mac, and makes Molten actually shippable
- Fix vulkan device API memory mapping bug when buffer is suballocated from a larger VulkanMemory
- Avoid buffer binding aliasing.
- Use different binding points for aliases if absolutely required
- Add AcquireRelease barrier before atomics (a compromise for aliasing instead of ptr cast). The impact of this is yet to be seen.
- Fix non 32-bit non-integral arguments on to-source translated platforms
- Relax GGUI test on macOS where Metal does not support wide lines
- Relaxed the element wise power test
- Changed args / rets to compiled SPIR-V struct

In addition to be able to run and ship MoltenVK, this also gurantees the SPIR-V codegen will work when translated to MSL (as that is what MoltenVK do under the hood). This helps our on-going AOT effort.

Vulkan testing is added for M1, and vulkan is added in release for all macOS variants


lin-hitonami (approved):
It works! Thanks!

.github/workflows/release.yml
```diff
@@ -209,13 +210,14 @@ jobs:
 
       - name: Build
         run: |
+          brew install molten-vk
```
lin-hitonami, line 213: Should we add `brew uninstall molten-vk` after we complete the building process to make sure the test can pass without molten-vk?

bobcao3: Sure, if it does not take a lot of time

PR #4206, opened 2022-01-31
[metal] Give random seeds a unique value

k-ye:
Related issue = https://github.com/taichi-dev/taichi/issues/4197

### DIFF
--- a/taichi/backends/metal/codegen_metal.cpp
+++ b/taichi/backends/metal/codegen_metal.cpp
@@ -23,6 +23,7 @@ namespace shaders {
 #define TI_INSIDE_METAL_CODEGEN
 #include "taichi/backends/metal/shaders/ad_stack.metal.h"
 #include "taichi/backends/metal/shaders/helpers.metal.h"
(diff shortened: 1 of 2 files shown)


taichi/backends/metal/codegen_metal.cpp
```diff
@@ -110,6 +124,28 @@ class RootIdsExtractor : public BasicStmtVisitor {
   std::unordered_set<int> roots_;
 };
 
+class TaskPreprocessor final : public BasicStmtVisitor {
```
bobcao3, line 127: Should we collect this into a analysis pass? (i.e. dump out usage and capability information, like random, numerical types used, SNode types used, external arrays, etc.)

bobcao3 (approved):
LGTM!

taichi/backends/metal/codegen_metal.cpp
k-ye: +1, this is very useful!

ailzhang (approved):
Thanks!!

PR #4209, opened 2022-02-04
[doc] Add the step of setting "TI_WITH_VULKAN" for linux

jerrylususu:
It seems currently when building from source, by default, taichi will not enable Vulkan backend. Only when the cmake flag "TI_WITH_VULKAN" set to `on` will it be enabled. See  https://github.com/taichi-dev/taichi/blob/00183bbe9b966727c2a28fdfadb17286ba90eb17/cmake/TaichiCore.cmake#L8 . 
As a result, I was unable to use Vulkan backend for my custom build of taichi, and finding out this took me a long time. For some strange reasons, the step of setting this cmake flag is shown in the Windows section, but not in the Linux section. This commit adds the missing step of setting cmake flags back to the Linux section.

It seems to be caused by https://github.com/taichi-dev/taichi/commit/31a1f72e72f3e8fdb44e350d29a737085f2ff02a#diff-60d6ef959481bb49f8acf52a43314bd4dbfadedb2341043488620077b32bae93 . Perhaps the wording makes it unclear that the cmake flag need to be set for both Windows and Linux.

Related issue = #

### DIFF
--- a/docs/lang/articles/contribution/dev_install.md
+++ b/docs/lang/articles/contribution/dev_install.md
@@ -373,7 +373,7 @@ You must install the Vulkan SDK in order to debug Taichi's Vulkan backend. To pr
 5. If the SDK is properly installed, add an environment variable `TAICHI_CMAKE_ARGS` with the value `-DTI_WITH_VULKAN:BOOL=ON` to enable the Vulkan backend: (Otherwise Vulkan backend is disabled by default when compiling from source.)
 
   ```shell
-  export TAICHI_CMAKE_ARGS=$TAICHI_CMAKE_ARGS:" -DTI_WITH_VULKAN:BOOL=ON"
+  export TAICHI_CMAKE_ARGS="$TAICHI_CMAKE_ARGS -DTI_WITH_VULKAN:BOOL=ON"
   ```
 
 </TabItem>
--- a/docs/lang/articles/contribution/dev_install.md
+++ b/docs/lang/articles/contribution/dev_install.md
@@ -370,7 +370,7 @@ You must install the Vulkan SDK in order to debug Taichi's Vulkan backend. To pr
 
 4. Check if the SDK is properly installed: `vulkaninfo`.
 
-5. If the SDK is properly installed, add an environment variable `TAICHI_CMAKE_ARGS` with the value `-DTI_WITH_VULKAN:BOOL=ON` to enable the Vulkan backend: (Otherwise Vulkan backend is disbaled by default when compling from source.)
+5. If the SDK is properly installed, add an environment variable `TAICHI_CMAKE_ARGS` with the value `-DTI_WITH_VULKAN:BOOL=ON` to enable the Vulkan backend: (Otherwise Vulkan backend is disabled by default when compiling from source.)
 
   ```shell
   export TAICHI_CMAKE_ARGS=$TAICHI_CMAKE_ARGS:" -DTI_WITH_VULKAN:BOOL=ON"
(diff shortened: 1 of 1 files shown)


bobcao3:
@k-ye @qiao-bo I'm not sure, is there a release related reason why Vulkan is not default-on?

jerrylususu:
Some more information: I am trying to dockerize my current environment, in which Taichi uses the Vulkan backend. When installed on physical machine, it seems that the auto detection works and my custom build can use Vulkan backend normally. However, when I use `nvidia/vulkan:1.2.170-470` as base image, taichi's pypi version can invoke the Vulkan backend, but failed to do so when I build from the source following the developer instruction. 

Initially I thought perhaps there is something wrong with my Dockerfile, so I tried to use the one in the CI folder (https://github.com/taichi-dev/taichi/blob/master/ci/Dockerfile.ubuntu.20.04), yet I still could not use Vulkan backend when building from source. Later I foudn out that a cmake flag needs to be set (https://github.com/taichi-dev/taichi/blob/98983260d1e27692b5944b3561ede3dc25a241db/ci/scripts/ubuntu_build_test.sh#L27), and indeed after setting this flag Vulkan backend works.

Considering the popularity of containers, it would be great if such nuanced behavior can be documented appropriately, so other developers will not have the same struggle as I experienced.

ifsheldon:
We also tried compiling Taichi on Arch and indeed, if we don't add the flag `TI_WITH_VULKAN=ON` to `TAICHI_CMAKE_ARGS`, the compiled artifact will not support Vulkan and if we do so, the artifact is enabled with Vulkan backend.

docs/lang/articles/contribution/dev_install.md
```diff
@@ -369,6 +369,12 @@ You must install the Vulkan SDK in order to debug Taichi's Vulkan backend. To pr
   > On Ubuntu, check if a JSON file with a name corresponding to your GPU vendor is in: `/etc/vulkan/icd.d/` or `/usr/share/vulkan/icd.d/`.
 
 4. Check if the SDK is properly installed: `vulkaninfo`.
+  
+5. If the SDK is properly installed, add an environment variable `TAICHI_CMAKE_ARGS` with the value `-DTI_WITH_VULKAN:BOOL=ON` to enable the Vulkan backend: (Otherwise Vulkan backend is disbaled by default when compling from source.)
```
qiao-bo, line 373:
```suggestion
5. If the SDK is properly installed, add an environment variable `TAICHI_CMAKE_ARGS` with the value `-DTI_WITH_VULKAN:BOOL=ON` to enable the Vulkan backend: (Otherwise Vulkan backend is disabled by default when compiling from source.)
```

qiao-bo (changes requested):
Just some nits typos ;//

jerrylususu:
typo fixed

qiao-bo (approved):
LGTM! Thanks

docs/lang/articles/contribution/dev_install.md
```diff
@@ -370,6 +370,12 @@ You must install the Vulkan SDK in order to debug Taichi's Vulkan backend. To pr
 
 4. Check if the SDK is properly installed: `vulkaninfo`.
 
+5. If the SDK is properly installed, add an environment variable `TAICHI_CMAKE_ARGS` with the value `-DTI_WITH_VULKAN:BOOL=ON` to enable the Vulkan backend: (Otherwise Vulkan backend is disabled by default when compiling from source.)
+
+  ```shell
+  export TAICHI_CMAKE_ARGS=$TAICHI_CMAKE_ARGS:" -DTI_WITH_VULKAN:BOOL=ON"
```
ailzhang, line 376:
```suggestion
  export TAICHI_CMAKE_ARGS="$TAICHI_CMAKE_ARGS -DTI_WITH_VULKAN:BOOL=ON"
```

ailzhang (changes requested):
nit: I believe it doesn't work with the colons in the middle.

ailzhang (approved):
Thanks a lot for contributing!

PR #4212, opened 2022-02-06
[gui] Add more camera controls

YuCrazing:
According to the convention of game engines, and also for the user experience, I add two more camera controlling commands :
* `q`: Move camera **up** vertically.
* `e`: Move camera **down** vertically.


neozhaoliang:
I don't think we should bind these keys for special purposes here, just let the users do that. (Also `q` is often used for exiting and `e` for shooting)

YuCrazing:
@neozhaoliang, thanks for your suggestion.

This is not a default setting, this preset is enabled only when users use 
``` python
camera.track_user_inputs(window=..., movement_speed=..., hold_key=...)
```
And in that situation, I think users wish they can control the camera in a simple and flexible way. So I think this change is a preferred way for users.

PR #4218, opened 2022-02-07
[Refactor] Move public APIs of ti.tools outside top level

strongoier:
Related issue = #3782

Docs should be modified accordingly after next release.


lin-hitonami (approved):
LGTM!

PR #4219, opened 2022-02-07
[doc] Add instruction to install clang-format-10 on M1 Mac

### LINKED ISSUE #4196: misc/code_format.py does not format correctly with clang-format-11
**Describe the bug**
I'm using m1 mac, and I've got clang-format 11 on my mac, and misc/code_format.py doesn't format correctly. According to the doc, the required clang-format version is 10 but homebrew does not provide clang-format-10.  It will be great if we can support clang-format-11 as well.

lin-hitonami:
Related issue = close #4196


ailzhang (approved):
Thanks!

Issue #3807, opened 2021-12-15
[RFC] Support ndarray for vulkan backend


ailzhang:
Ndarrays are pretty useful in the AOT scenarios (https://github.com/taichi-dev/taichi/issues/3642) so we'd like to support it for vulkan backend as well. 

* Currently ndarray is supported on cpu & cuda & opengl backend. 

Adding ndarray support for vulkan backend should look a lot similar with the approach we used for opengl backend. 

Some notable milestones are: 
- [ ] Separate external arr from args buf.   (opengl reference #3506 ) 
- [ ] Update codegen to support ndarray (opengl reference https://github.com/taichi-dev/taichi/pull/3555/) 
- [ ] Support ndarray based AOT serialization and deserialization. (opengl reference https://github.com/taichi-dev/taichi/pull/3556)

 cc: @ghuau-innopeak

PR #4221, opened 2022-02-07
[vulkan] Add buffer device address (physical pointers) support & other improvements

bobcao3:
Related issue = https://github.com/taichi-dev/taichi/issues/3807

1. Buffer device address

Buffer device address is a unique Vulkan capability of directly addressing the memory using a pointer, and the ability to cast pointers to uint64 (i.e. generic pointers)

Codegen is modified so that ext arrays and ndarrays can be addressed directly instead of using bindings. This also removes an shifting because the pointers are byte addressed. This also means aliasing is not needed if buffer device address is supported.

2. Uniform buffer / constants buffer args

This change moved the args to a uniform buffer which is loaded and cached in a separate specialized constants buffer on the GPU. The constants buffer supports broadcasting (i.e. a single scalar load is issued and results are broadcasted to all lanes in the wave) This also allows tools like spirv-opt to optimize code involving args (e.g. index calculation) as uniform buffer is known to be constant during execution.


taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1647,12 +1682,18 @@ class TaskCodegen : public IRVisitor {
...
     }
 
+    if (buffer.type == BufferType::Rets) {
+      buffer_binding_map_[key] = 1;
+      buffer_value_map_[key] = ret_buffer_value_;
+      return ret_buffer_value_;
+    }
+
     int binding = binding_head_++;
     buffer_binding_map_[key] = binding;
```
AmesingFlank, line 1698: I don't quite understand whats happening here. If we return early for "Args" and "Rets", doesn't that mean `buffer_argument` won't be called for these two buffers? E.g. they would miss the "Block" decoration. This was also the case in #4205, and I've been getting complaints from tint because of that.

Also, wouldn't this cause Args and Rets to be bound to the same binding point as other buffers?

AmesingFlank: Or maybe the "Block" decoration should be added to `buffer_struct_argument`?

AmesingFlank: Another thing about #4205 and this PR that's a bit problematic:  In 4205, `compile_args_struct()` gets called regardless of whether or not there're any arguments. You fixed that in this PR by returning immediately if num args = 0, but not every offloaded task in the kernel is guaranteed to use those args. So you could end with a offloadedTask/shader that declares the args buffer as an input, but never use it.  This might be fine with Vulkan (i'm not sure), but it's causing a lot of trouble for webgpu stuff.

AmesingFlank: Let's call "compile_args" only when we notice a `ArgLoadStmt`. Same for return values.

AmesingFlank:
Other than the previously mentioned stuff, LGTM. 
Also, thanks for removing aliasing, it helps not only moltenvk but also made life easier for me when dealing with wgsl.

taichi/codegen/spirv/spirv_codegen.cpp
bobcao3: The decoration is added when calling "compile_args_struct" where it calls the “buffer_struct_argument" or "uniform_struct_argument"

bobcao3: Also the `binding_head_` is initialized to 2 so other things won't bind on 0 and 1. It's quite ugly, but it's the simplest solution I have

AmesingFlank: > binding_head_ is initialized to 2 

Right. That works. But please add a comment tho.

> The decoration is added when calling "compile_args_struct" where it calls the “buffer_struct_argument" or "uniform_struct_argument"

Ok I see that you added it in this PR. Interesting that #4205 didn't have it but nothing was crashing though. 

So the only thing is to please move "compile_args_struct" out of `run()` and into `visit(ArgLoadStmt)` and I would be a happy, happy man.

bobcao3: Let's leave it in `run` rn, afterwards we will have a prepass that scans all the required features / structures, and then we can clean up all the initialization code (instead of the one pass approach we are using rn). If we want to move it to `visit(ArgLoadStmt)` rn we will need an if statement to avoid generating multiple parts.

AmesingFlank: And an if statement is bad because?

bobcao3: It is not an elegant solution either. Proper fix will happen when we have a prepass featuer scan.

AmesingFlank: its not that ugly either. we know that a visit(ArgLoad) exists iff you need the arg structs. so please move it :)

taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1676,34 +1791,60 @@ class TaskCodegen : public IRVisitor {
...
+        ir_->buffer_struct_argument(ret_struct_type_, 0, 1, "rets");
   }
 
   std::vector<BufferBind> get_buffer_binds() {
     std::vector<BufferBind> result;
     for (auto &[key, val] : buffer_binding_map_) {
       result.push_back(BufferBind{key.first, int(val)});
     }
-    return result;
+    return std::move(result);
```
k-ye, line 1847: We don't need `move`, RVO is guaranteed :-) https://stackoverflow.com/a/17473874/12003165

k-ye (commented):
Cool! 

If we are using uniform buffer for args, how do we handle writing to an external array? Or is the external arrays implemented as Ndarray already?

bobcao3:
> Cool!
> 
> If we are using uniform buffer for args, how do we handle writing to an external array? Or is the external arrays implemented as Ndarray already?

It's the same as NDarray (truely any_arr)

taichi/codegen/spirv/spirv_codegen.cpp
bobcao3: No I don't think that's the right way. We shouldn't scatter global section initialization and type definitions across all the statements. We will add a prepass soon once the Metal random seed PR is merged.

k-ye (approved):
Thanks LGTM!

PR #4224, opened 2022-02-07
[lang] Annotate constants with dtype without casting.

### LINKED ISSUE #4210: False warnings on unsigned integer overflow
**Describe the bug**
Weird warnings about unsigned integer exceeding the range

**To Reproduce**

```py
import taichi as ti

ti.init(arch=ti.gpu)

@ti.kernel
def u32_test():
    x: ti.u32 = 2891336453
    print("x = ", x)

u32_test()
```

**Log/Screenshots**
```
[Taichi] version 0.8.11, llvm 10.0.0, commit 9d0973e5, win, python 3.7.8
[W 02/04/22 20:57:53.169 24856] [impl.py:_clamp_unsigned_to_range@395] Constant 2891336453 has exceeded the range of 32 int, clamped to -1403630843
x =  2891336453
```

**Additional comments**
It wasn't actually clamped and kernel works correctly as far as I can tell, so basically it's just about false warnings

ailzhang:
Fixes #4210

This PR also simplifies the clamping logic by:
1. Make sure constant is valid in range of specified dtype.
2. Always use the widest int64/float64 to communicate between python and C++ and
converts it to the specified dtype in C++. This is just to avoid
handling each dtype on both python and C++ side.

We might consider further improve it by moving this logic from python to C++ in the future and this PR made it a step closer.

Related issue = #

### DIFF
--- a/python/taichi/lang/expr.py
+++ b/python/taichi/lang/expr.py
@@ -9,7 +9,7 @@ from taichi.lang.util import is_taichi_class
 # Scalar, basic data type
 class Expr(TaichiOperations):
     """A Python-side Expr wrapper, whose member variable `ptr` is an instance of C++ Expr class. A C++ Expr object contains member variable `expr` which holds an instance of C++ Expression class."""
-    def __init__(self, *args, tb=None):
+    def __init__(self, *args, tb=None, dtype=None):
         self.tb = tb
         if len(args) == 1:
             if isinstance(args[0], _ti_core.Expr):
@@ -29,11 +29,8 @@ class Expr(TaichiOperations):
                         raise TaichiTypeError(
                             "Only 0-dimensional numpy array can be used to initialize a scalar expression"
                         )
-                    try:
-                        arg = arg.dtype.type(arg)
-                    except:
-                        pass
-                self.ptr = impl.make_constant_expr(arg).ptr
+                    arg = arg.dtype.type(arg)
+                self.ptr = impl.make_constant_expr(arg, dtype).ptr
         else:
             assert False
         if self.tb:
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -4,7 +4,7 @@ from typing import Iterable
 
 import numpy as np
 from taichi._lib import core as _ti_core
-from taichi._logging import warn
+from taichi._logging import error
 from taichi._snode.fields_builder import FieldsBuilder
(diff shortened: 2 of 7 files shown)


taichi/python/export_lang.cpp
```diff
@@ -777,10 +778,11 @@ void export_lang(py::module &m) {
 
   m.def("make_rand_expr", Expr::make<RandExpression, const DataType &>);
 
-  m.def("make_const_expr_i32", Expr::make<ConstExpression, int32>);
-  m.def("make_const_expr_i64", Expr::make<ConstExpression, int64>);
-  m.def("make_const_expr_f32", Expr::make<ConstExpression, float32>);
-  m.def("make_const_expr_f64", Expr::make<ConstExpression, float64>);
+  m.def("make_const_expr_ip",
```
lin-hitonami, line 781: What does `ip` stand for? IIUC `fp` is the acronym of floating point, and the acronym of integer should be `int`?

python/taichi/lang/expr.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, *args, tb=None):
                         arg = arg.dtype.type(arg)
                     except:
```
lin-hitonami, line 34: I don't know why this try-except block exists, can we remove it?

python/taichi/lang/impl.py
```diff
@@ -381,58 +382,45 @@ def get_runtime():
...
-        assert False
-    else:
-        raise TaichiTypeError(
-            f'Invalid constant scalar data type: {type(val)}')
+        constant_dtype = pytaichi.default_ip if dtype is None else dtype
+        _check_in_range(to_numpy_type(constant_dtype), val)
+        return Expr(
+            _ti_core.make_const_expr_ip(
+                constant_dtype, _clamp_unsigned_to_range(np.int64, val)))
+    if isinstance(val, (float, np.floating, np.ndarray)):
```
lin-hitonami, line 420: IIUC, if we remove the try-except block in `Expr.__init__`, we can remove `np.ndarray` here as well.

taichi/python/export_lang.cpp
ailzhang: good point, I don't know what ip stands for either but copied from default_ip lol. Changed to int here ;)

lin-hitonami (approved):
LGTM!

PR #4225, opened 2022-02-07
[Bug] Ban passing torch view tensors into taichi kernel

### LINKED ISSUE #4208: Views of PyTorch tensors are not observed in `ti.any_arr()` access
### **Describe the bug**
When accessing a PyTorch tensor using `ti.any_arr()`, the tensor is always **indexed using its internal memory layout instead of its viewed shape.** This results in unintuitive behavior in a Taichi kernel operating on PyTorch tensors.

### **To Reproduce**
For example, this piece of test does not pass:
```python
import taichi as ti
import torch
ti.init()
@ti.kernel
def copy(x: ti.any_arr(), y: ti.any_arr()):
    for i, j in x:
        y[i, j] = x[i, j]
x = torch.Tensor([[1,2,3],[4,5,6],[7,8,9]]).T
y = ti.ndarray(int, (3, 3))

copy(x, y)

for i in range(3):
    for j in range(3):
        assert x[i, j] == y[i, j]
```
This is because transposing a PyTorch tensor creates an alternate view of the tensor instead of changing its memory layout, making `x[i, j]` in python scope being column-major (w.r.t. memory). In contrast, `x[i, j]` in the Taichi kernel is still row-major. Changing line of the kernel invocation to `copy(x.contiguous(), y)`, of course, makes the test pass.

### **Additional comments**
I'm not sure whether this needs to be fixed as a bug. Alternatively, we can prompt this behavior to users in the documentation of `ti.any_arr()` so that they are aware that PyTorch tensors are always accessed by the actual memory layout in Taichi kernels.

ailzhang:
fixes #4208

Related issue = #

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -487,6 +487,10 @@ class Kernel:
 
         assert has_torch
         assert isinstance(v, torch.Tensor)
+        if v._is_view():
+            raise ValueError(
+                "Torch view tensors are not supported, please call tensor.clone() before passing it into taichi kernel."
+            )
         tmp = v
         taichi_arch = self.runtime.prog.config.arch
         # Ndarray means its memory is allocated on the specified taichi arch.
--- a/tests/python/test_torch_io.py
+++ b/tests/python/test_torch_io.py
@@ -269,3 +269,19 @@ def test_torch_zero():
     test_torch(torch.zeros((0), dtype=torch.int32))
     test_torch(torch.zeros((0, 5), dtype=torch.int32))
     test_torch(torch.zeros((5, 0, 5), dtype=torch.int32))
+
+
+@pytest.mark.skipif(not has_pytorch(), reason='Pytorch not installed.')
+@ti.test(exclude=[ti.opengl, ti.vulkan])
+def test_torch_view():
+    @ti.kernel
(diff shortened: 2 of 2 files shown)


victoriacity (approved):
LGTM Thanks!

ifsheldon:
This is a breaking change, so I suggest at least bump the version by +0.1 according to semantic versioning

ifsheldon:
Besides, this commit affects `field.from_torch()`, but the impact is weird. Can you test this case?
```pyhon
@ti.kernel
def ti_print(a: ti.template()):
    for i in ti.grouped(a):
        print(f"{i[0]} {i[1]}, {a[i]}")


def test_view():
    ti.init(ti.cpu)
    a = torch.tensor([[1, 2], [3, 4]], dtype=torch.float32)
    a_t = a.T
    assert a_t._is_view()
    a_field = ti.field(ti.f32, shape=a.shape)
    a_field.from_torch(a_t)  # ok, but why?
    ti_print(a_field)
    a_complex = torch.ones(2, 2, dtype=torch.cfloat)
    a_real_view = torch.view_as_real(a_complex)
    assert a_real_view._is_view()
    complex_field = ti.field(ti.f32, shape=a_real_view.shape)
    complex_field.from_torch(a_real_view)  # not ok
    ti_print(complex_field)
```

PR #4227, opened 2022-02-08
[Refactor] Rename and move memory profiler info under ti.profiler

mzmzm:
Related issue = #3782

### DIFF
--- a/python/taichi/__init__.py
+++ b/python/taichi/__init__.py
@@ -21,6 +21,7 @@ __deprecated_names__ = {
     'AOS': 'Layout.AOS',
     'print_profile_info': 'profiler.print_scoped_profiler_info',
     'clear_profile_info': 'profiler.clear_scoped_profiler_info',
+    'print_memory_profile_info': 'profiler.print_memory_profiler_info',
     'CuptiMetric': 'profiler.CuptiMetric',
     'get_predefined_cupti_metrics': 'profiler.get_predefined_cupti_metrics',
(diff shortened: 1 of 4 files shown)


python/taichi/profiler/memory_profiler.py
```diff
@@ -0,0 +1,13 @@
+from taichi.lang.impl import get_runtime
+
+
+def print_memory_profile_info():
```
strongoier, line 4:
```suggestion
def print_memory_profiler_info():
```

python/taichi/__init__.py
```diff
@@ -29,6 +29,7 @@
     'kernel_profiler_total_time': 'profiler.get_kernel_profiler_total_time',
     'set_kernel_profiler_toolkit': 'profiler.set_kernel_profiler_toolkit',
     'set_kernel_profile_metrics': 'profiler.set_kernel_profiler_metrics',
+    'print_memory_profile_info': 'profiler.print_memory_profile_info',
```
strongoier, line 32:
```suggestion
    'print_memory_profile_info': 'profiler.print_memory_profiler_info',
```

strongoier (commented):
We would like to unify the name :-)

python/taichi/__init__.py
strongoier: BTW this line may be moved to L24 to group things better (scope profiler, memory profiler, kernel profiler).

strongoier (approved):
LGTM!

PR #4233, opened 2022-02-08
[Refactor] Move ti.quant & ti.type_factory under ti.types.quantized_types

strongoier:
Related issue = #3782

Because these two APIs are still unstable, I turn them into internal (experimental) APIs in this PR.


ailzhang (approved):
Awesome! Thanks!

PR #4235, opened 2022-02-08
[Error] Raise an error when breaking the outermost for

### LINKED ISSUE #4223: Confusing error message when having break in the outermost loop
**Describe the bug**
For this program,
```python
import taichi as ti
ti.init()

@ti.kernel
def foo():
    for j in range(100):
        break

foo()
```
The error message is like
```
  File "test.py", line 9, in <module>
    foo()
  File "/opt/homebrew/Caskroom/miniforge/base/envs/test38/lib/python3.8/site-packages/taichi-0.8.12-py3.8-macosx-11.0-arm64.egg/taichi/lang/kernel_impl.py", line 736, in wrapped
  File "/opt/homebrew/Caskroom/miniforge/base/envs/test38/lib/python3.8/site-packages/taichi-0.8.12-py3.8-macosx-11.0-arm64.egg/taichi/lang/kernel_impl.py", line 667, in __call__
  File "/opt/homebrew/Caskroom/miniforge/base/envs/test38/lib/python3.8/site-packages/taichi-0.8.12-py3.8-macosx-11.0-arm64.egg/taichi/lang/kernel_impl.py", line 619, in func__
    t_kernel(launch_ctx)
RuntimeError: [ir.h:as@225] Assertion failure: is<T>()
```

lin-hitonami:
Related issue = close #4223

### DIFF
--- a/python/taichi/lang/exception.py
+++ b/python/taichi/lang/exception.py
@@ -30,6 +30,8 @@ class TaichiRuntimeTypeError(TaichiRuntimeError, TypeError):
 def handle_exception_from_cpp(exc):
     if isinstance(exc, core.TaichiTypeError):
         return TaichiTypeError(str(exc))
+    if isinstance(exc, core.TaichiSyntaxError):
+        return TaichiSyntaxError(str(exc))
     return exc
 
 
--- a/taichi/common/exceptions.h
+++ b/taichi/common/exceptions.h
@@ -16,5 +16,16 @@ class TaichiTypeError : public std::exception {
   }
 };
 
+class TaichiSyntaxError : public std::exception {
+  std::string msg_;
+
+ public:
+  TaichiSyntaxError(const std::string msg) : msg_(msg) {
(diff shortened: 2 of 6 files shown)


strongoier:
Ah... Seems like your PR cannot live together with #4237. Are we on our way to move scope guards from Python to C++? I prefer to only keep one of that. If so, you may have to refactor a bit on top of current master...

lin-hitonami:
> Ah... Seems like your PR cannot live together with #4237. Are we on our way to move scope guards from Python to C++? I prefer to only keep one of that. If so, you may have to refactor a bit on top of current master...

This time the check cannot be performed in python because the outermost loop may be in a `ti.func`, and in the current design, the inlined function knows nothing about the function that calls it in Python. So, we have to do the check in C++.

Scope guards in python mainly handle problems about compile-time loop unrolling and static scopes, so I don't think they can be moved to C++.

PR #4242, opened 2022-02-09
[refactor] Remove dependency on get_current_program() in lang::BinaryOpExpression

PGZXB:
Related issue = #3955


strongoier (approved):
LGTM!

k-ye (commented):
It seems weird that the type checking API has to take in a compile config. IIRC, `type_check()` was made part of `Expr`'s method, because you can do the checking as you build the AST. But for API clarity, WDYT if we have a separate AST-pass to run the type checking? @strongoier

k-ye:
> One more thing to mention here is that, only the truediv op needs the compile config...

Ah I see... Yeah I can't think of a good fix for this, either. Let's keep it so for now :-(

BTW, I didn't mean that making type_check a dedicated pass is something urgent, just something we should conider

PR #4244, opened 2022-02-09
[vulkan] Disable buffer device address if int64 is not supported

bobcao3:
And fixing minor issues.


qiao-bo (approved):
This works great! Thanks

PR #4248, opened 2022-02-10
[misc] Update master version to 0.9.0

ailzhang:
We plan to do our next release as v0.9.0 so let's make sure master has
the correct version.

Related issue = #


k-ye (approved):
🚀

Issue #2768, opened 2021-08-22
Deprecate DecoratorRecorder


k-ye:
**Concisely describe the proposed feature**

We need to deprecate `DecoratorRecorder`, at least the global instance in https://github.com/taichi-dev/taichi/blob/126cf217abdc33625d914abf2913805bb0c74442/taichi/ir/frontend_ir.cpp#L59 (Engineering cost: https://github.com/taichi-dev/taichi/pull/2735)

As an alternative, the decorators backed by `DecoratorRecorder` could be interleaved inside the AST itself. I.e. we can have a series of decoration instructions that will only modify the behavior of each offloaded task. With a proper design, we could support a per-kernel config.

PR #4254, opened 2022-02-10
Remove global instance of DecoratorRecorder

PGZXB:
Related issue = #2768

Move `DecoratorRecorder` into `ASTBuilder` -> Use member of `ASTBuilder` instead of global instance
Renaming : `DecoratorRecorder` -> `ForLoopDecoratorRecorder`

### DIFF
--- a/taichi/ir/frontend_ir.h
+++ b/taichi/ir/frontend_ir.h
@@ -170,17 +170,33 @@ class FrontendForStmt : public Stmt {
     }
   }
 
-  FrontendForStmt(const ExprGroup &loop_var, const Expr &global_var, Arch arch);
+  FrontendForStmt(const ExprGroup &loop_var,
+                  const Expr &global_var,
+                  Arch arch,
+                  int bit_vectorize,
+                  int num_cpu_threads,
+                  bool strictly_serialized,
+                  int block_dim,
+                  MemoryAccessOptions &&mem_access_opt);
 
   FrontendForStmt(const ExprGroup &loop_var,
                   const mesh::MeshPtr &mesh,
                   const mesh::MeshElementType &element_type,
-                  Arch arch);
+                  Arch arch,
+                  int bit_vectorize,
+                  int num_cpu_threads,
+                  int block_dim,
+                  MemoryAccessOptions &&mem_access_opt);
 
   FrontendForStmt(const Expr &loop_var,
                   const Expr &begin,
                   const Expr &end,
-                  Arch arch);
+                  Arch arch,
+                  int bit_vectorize,
+                  int num_cpu_threads,
+                  bool strictly_serialized,
+                  int block_dim,
+                  MemoryAccessOptions &&mem_access_opt);
 
   bool is_container_statement() const override {
     return true;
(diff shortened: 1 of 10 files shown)


strongoier (approved):
LGTM!

taichi/ir/frontend_ir.h
```diff
@@ -170,17 +170,33 @@ class FrontendForStmt : public Stmt {
     }
   }
 
-  FrontendForStmt(const ExprGroup &loop_var, const Expr &global_var, Arch arch);
+  FrontendForStmt(const ExprGroup &loop_var,
+                  const Expr &global_var,
+                  Arch arch,
+                  int bit_vectorize,
```
k-ye, line 176: Could you actually create a struct to hold these params? That way `FrontendForStmt` and `ForLoopDecoratorRecorder` can share the struct, instead of repeating yourself. Thx!

PGZXB: You mean that let FrontendForStmt holds ptr to ForLoopDecoratorRecorder?

PGZXB: > You mean that let FrontendForStmt holds ptr to ForLoopDecoratorRecorder?

Oh, sorry, my bad. Holding ptr to ForLoopDecoratorRecorder is not feasible.

k-ye: Nah, I meant something like this:

```cpp

struct ForLoopConfig {
  int bit_vectorize;
  int num_cpu_threads;
  ...
};

class FrontendForStmt {
 public:
  FrontendForStmt(..., const ForLoopConfig &config);
};

class ForLoopDecoratorRecorder {
  ForLoopConfig config;
};
```

What do you think :-)

PGZXB: > What do you think

Got it! But why not use `ForLoopConfig` directly if `ForLoopDecoratorRecorder` only wraps `ForLoopConfig` simply? The 
existence of the wrapper is convenient for us to construct/reset `ForLoopConfig`?

k-ye: Recorder sounds like something with behavior, whereas Config is just a struct. That said, sounds good to me if we just use recorder for simplicity. I'm not sure why "Holding ptr to ForLoopDecoratorRecorder is not feasible.", though.

PGZXB: > Recorder sounds like something with behavior, whereas Config is just a struct. That said, sounds good to me if we just use recorder for simplicity. I'm not sure why "Holding ptr to ForLoopDecoratorRecorder is not feasible.", though.

"Holding ptr to ForLoopDecoratorRecorder" : My **old incorrect** thought is let `FrontendForStmt` and `ASTBuilder` share a `ForLoopDecoratorRecorder` instead of bit_vectorize, num_cpu_threads which are copy-saved by `FrontendForStmt`, which is not feasible because the `ForLoopDecoratorRecorder` which is held by ASTBuilder will be reset after constructing `FrontendForStmt`.

PGZXB: @k-ye FINISHED. Please review #4309

PR #4255, opened 2022-02-10
[vulkan] Use TI_VISIBLE_DEVICE to select vulkan device

qiao-bo:
Add an env var option `TI_VISIBLE_DEVICE` to enable user selected vulkan device. The identifier is device id, can be obtained by querying `vulkaninfo`.

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -874,7 +874,7 @@ def stop_grad(x):
 
 
 def current_cfg():
-    return _ti_core.current_compile_config()
+    return get_runtime().prog.config
 
 
 def default_cfg():
--- a/python/taichi/lang/kernel_arguments.py
+++ b/python/taichi/lang/kernel_arguments.py
@@ -1,5 +1,6 @@
 import taichi.lang
 from taichi._lib import core as _ti_core
+from taichi.lang import impl
 from taichi.lang.any_array import AnyArray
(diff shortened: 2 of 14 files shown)


ailzhang (commented):
sorry this might be a dumb question, how do I list all my vulkan devices before setting? ;)

qiao-bo:
> sorry this might be a dumb question, how do I list all my vulkan devices before setting? ;)

legitimate question! `vulkaninfo` can be used. similar to clinfo or devicequery for cuda. The idea is to have a unique identifier rather than 0, 1 ,2...  Please suggest if there are better options~

bobcao3:
Should probably rename to `TI_USE_VULKAN_DEVICE` since "visible device" is quite confusing.

k-ye:
> Should probably rename to `TI_USE_VULKAN_DEVICE` since "visible device" is quite confusing.

I suggest to remove the specific mentioning of `VULKAN_`, since many other APIs support GPU selection as well.

qiao-bo:
> > Should probably rename to `TI_USE_VULKAN_DEVICE` since "visible device" is quite confusing.
> 
> I suggest to remove the specific mentioning of `VULKAN_`, since many other APIs support GPU selection as well.

this is the intention, same switch could be used for other APIs.

bobcao3 (approved):
LGTM

PR #4258, opened 2022-02-10
[Bug] Only ban passing non contiguous torch tensors to taichi kernels.

ailzhang:
Thanks to @ifsheldon for reporting the bug!

It's too strict to ban all view tensors as Pytorch view tensor return
itself as a new view Tensor when nothing changes.
So what we really wanted to ban was actually non contiguous tensors.

```
In [2]: a = torch.ones(4, 4)

In [4]: b = a.view(4, 4)

In [6]: b._is_view()
Out[6]: True

In [7]: b.is_contiguous()
Out[7]: True

In [8]: a._is_view()
Out[8]: False
```
FYI in `from_torch` we actually explicitly call `tensor.contiguous()` to
make sure it's valid. For ndarrays we want users to do that explicitly
so that they're aware of the memcpy.

Related issue = #

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -487,9 +487,9 @@ class Kernel:
 
         assert has_torch
         assert isinstance(v, torch.Tensor)
-        if v._is_view():
+        if not v.is_contiguous():
             raise ValueError(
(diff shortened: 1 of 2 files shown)


k-ye (approved):
LGTM!

PR #4259, opened 2022-02-11
[vulkan] Test & build macOS 10.15 MoltenVK

bobcao3:
Related issue = #

### DIFF
--- a/taichi/backends/vulkan/vulkan_loader.cpp
+++ b/taichi/backends/vulkan/vulkan_loader.cpp
@@ -29,49 +29,48 @@ bool VulkanLoader::check_vulkan_device() {
   VkInstance instance{VK_NULL_HANDLE};
   VkResult res = vkCreateInstance(&create_info, kNoVkAllocCallbacks, &instance);
   
-  if (res != VK_SUCCESS) {
-    TI_WARN("Can not create Vulkan instance");
-    goto cleanup; 
-  }
-
-  load_instance(instance);
-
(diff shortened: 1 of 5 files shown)


taichi/backends/vulkan/vulkan_loader.cpp
```diff
@@ -11,6 +11,80 @@ namespace vulkan {
...
+        for (auto &queue : queue_families) {
+          if (queue.queueFlags & VK_QUEUE_COMPUTE_BIT) {
+            found_device_with_compute = true;
+          }
+        }
+
+        VkPhysicalDeviceProperties properties{};
+        vkGetPhysicalDeviceProperties(physical_device, &properties);
+
+        TI_INFO("Found Vulkan Device {} ({})", i, properties.deviceName);
```
k-ye, line 76:
```suggestion
        TI_DEBUG("Found Vulkan Device {} ({})", i, properties.deviceName);
```

k-ye (approved):
Cool!

taichi/backends/vulkan/vulkan_loader.cpp
bobcao3: This is in conjunction with #4255 so that the user can know which device corresponds to which ID

bobcao3: Maybe is there a way to display info without including the source file path?

ailzhang (approved):
LGTM!

PR #4262, opened 2022-02-11
[refactor] Remove support for raise statement

### LINKED ISSUE #4261: Exceptions raised in Taichi kernels are thrown at compile time
**Describe the bug**
Currently, exceptions in Taichi kernels are thrown at compile time, i.e., when the transformer traverses to the `raise` statement. This leads to some weird behaviors such as writing a `raise` statement in an `if` block results in compilation error even if the condition is not true. So, I propose that we should remove the support for `raise` statements.

**To Reproduce**

```py
import taichi as ti
ti.init()
@ti.kernel
def test():
    x = 1.0
    if x < 0.0001:
        raise ValueError("oops")
test()
```

**Additional comments**
There are some usages of `raise` statements in https://github.com/taichi-dev/taichi/blob/master/python/taichi/_funcs.py , for example,
```py
@func
def _svd(A, dt):
    """Perform singular value decomposition (A=USV^T) for arbitrary size matrix.
    Mathematical concept refers to https://en.wikipedia.org/wiki/Singular_value_decomposition.
    2D implementation refers to :func:`taichi.svd2d`.
    3D implementation refers to :func:`taichi.svd3d`.
    Args:
        A (ti.Matrix(n, n)): input nxn matrix `A`.
        dt (DataType): date type of elements in matrix `A`, typically accepts ti.f32 or ti.f64.
    Returns:
        Decomposed nxn matrices `U`, 'S' and `V`.
    """
    if static(A.n == 2):  # pylint: disable=R1705
        ret = svd2d(A, dt)
        return ret
    elif static(A.n == 3):
        return svd3d(A, dt)
    else:
        raise Exception("SVD only supports 2D and 3D matrices.")
```
We should move the `raise` statement out of the Taichi function.

lin-hitonami:
Related issue = close #4261


python/taichi/lang/ast/ast_transformer.py
```diff
@@ -1178,10 +1178,6 @@ def build_Continue(ctx, node):
     def build_Pass(ctx, node):
         return None
 
-    @staticmethod
-    def build_Raise(ctx, node):
-        raise build_stmt(ctx, node.exc)
```
ailzhang, line 1183: Can we add a test to show what's the current err msg when users write Raise in their kernel? Thanks!

ailzhang (approved):
Thanks!

PR #4264, opened 2022-02-14
[lang] Make ti.cfg an alias of runtime cfg

ailzhang:
Before this PR we have `ti.cfg` and `ti.lang.impl.current_cfg()` and
they're different objects. It's confusing to have both so let's just
make `ti.cfg` an alias of the other. (it's a bit difficult to hook ti.cfg as function call in python3.6 so let's just skip it for now.)

Note `ti.cfg` used to be default_config before `ti.init()` is called and
now it's `None`. It's safe to change this behavior as nothing can be
done without `ti.init()`.

Related issue = #3782

### DIFF
--- a/python/taichi/__init__.py
+++ b/python/taichi/__init__.py
@@ -47,6 +47,11 @@ if sys.version_info.minor < 7:
 else:
 
     def __getattr__(attr):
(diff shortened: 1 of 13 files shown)


strongoier (approved):
LGTM!

PR #4268, opened 2022-02-14
[refactor] Re-expose important implementation classes

strongoier:
Related issue = #3782

Users need to know these classes in order to query usages.


lin-hitonami (approved):
LGTM!

PR #4273, opened 2022-02-14
[lang] External Ptr alias analysis & demote atomics

bobcao3:
I tried to figure out a way to not use a DecorationStmt, but rather directly look at the index computations. However that feels very hard to do and very clumsy. Since the range fors on external and nd arrays are generated during the lowering pass, this index information is lost in the later stages, the DecorationStmt is meant to be a simple way to preserve these information to the end. In the future we can also put debug information (e.g. source correlation, variable names, etc.) into the statement.

As for the loop-unique check part, I'm not so sure about the logic I used here. Also wonders why the loop-unique access checks for the GlobalPtrStmts are so much more involved.

As for performance benefits, it seems there's a very small improvement on lower end Vulkan devices. ( mpm88_ndarray )


ailzhang (approved):
Nice catch, this is definitely beneficial for mobile perf as well!!
(now that I realize I focus too much on the time consuming substep01 shader and forgot to check opportunities in substep02 & 03 :P

taichi/ir/statements.h
```diff
@@ -99,6 +99,38 @@ class ContinueStmt : public Stmt {
   TI_DEFINE_ACCEPT_AND_CLONE;
 };
 
+/**
+ * A decoration statement. The decorated "operands" will keep this decoration.
+ */
+class DecorationStmt : public Stmt {
+ public:
+  enum class Decoration : uint32_t { unknown, loop_unique };
```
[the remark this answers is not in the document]
k-ye: +1, but it should be `kUnknown` according to https://google.github.io/styleguide/cppguide.html#Enumerator_Names

taichi/transforms/demote_atomics.cpp
```diff
@@ -41,42 +42,55 @@ class DemoteAtomics : public BasicStmtVisitor {
...
-            if (idx->is<LoopIndexStmt>() &&
-                idx->as<LoopIndexStmt>()->is_mesh_index() &&
-                loop_unique_ptr_[stmt->dest->as<GlobalPtrStmt>()
-                                     ->snodes.data[0]] != nullptr) {
-              demote = true;
-            }
+          if (dest_ptr->indices.empty()) {
+            demote = false;
           }
+          // TODO: Is BLS / Mem Access Opt a thing for any_arr?
```
k-ye, line 93: Good question, we might be able to extend BLS to external arrays as well.  However, at this point BLS depends on knowing the struct-for's loop index

taichi/analysis/gather_uniquely_accessed_pointers.cpp
```diff
@@ -55,6 +59,18 @@ class LoopUniqueStmtSearcher : public BasicStmtVisitor {
     }
   }
 
+  void visit(DecorationStmt *stmt) override {
+    if (stmt->decoration.size() == 2 &&
```
k-ye, line 63: Why 2?

k-ye (approved):
LGTM!

k-ye:
I wonder if DecoratorRecorder can be merged with this? @strongoier @PGZXB

taichi/analysis/gather_uniquely_accessed_pointers.cpp
```diff
@@ -55,6 +59,18 @@ class LoopUniqueStmtSearcher : public BasicStmtVisitor {
...
+        stmt->decoration[0] ==
+            uint32_t(DecorationStmt::Decoration::loop_unique)) {
+      if (loop_unique_.find(stmt->operand) == loop_unique_.end()) {
+        loop_unique_[stmt->operand] = stmt->decoration[1];
+        num_different_loop_indices = std::max(loop_unique_[stmt->operand] + 1,
+                                              num_different_loop_indices);
```
xumingkuan, line 69: Could you please give me an example that `loop_unique_[stmt->operand]` is greater than or equal to `num_different_loop_indices` before this statement? I feel a bit worried if such a thing can happen because this analysis should guarantee to produce a correct result even if *some* `DecorationStmt`s are removed.

taichi/analysis/gather_uniquely_accessed_pointers.cpp
```diff
@@ -120,6 +166,9 @@ class UniquelyAccessedSNodeSearcher : public BasicStmtVisitor {
   std::unordered_map<const SNode *, GlobalPtrStmt *> accessed_pointer_;
   std::unordered_map<const SNode *, GlobalPtrStmt *> rel_access_pointer_;
 
+  // Search any_arrs that are uniquely accessed. Maps: ArgID -> ExternalPtrStmt
```
xumingkuan, line 169: What is `any_arrs`? Is it just "any external arrays"?

taichi/ir/analysis.h
```diff
@@ -106,7 +106,8 @@ gather_snode_read_writes(IRNode *root);
 std::vector<Stmt *> gather_statements(IRNode *root,
                                       const std::function<bool(Stmt *)> &test);
 void gather_uniquely_accessed_bit_structs(IRNode *root, AnalysisManager *amgr);
-std::unordered_map<const SNode *, GlobalPtrStmt *>
+std::pair<std::unordered_map<const SNode *, GlobalPtrStmt *>,
```
xumingkuan, line 109: We may need to clarify here: the first component means that the **SNode** is uniquely accessed, but the second component means that the external pointer is uniquely accessed. If I'm not missing something, two `ExternalPtrStmt`s pointing to the same external array doesn't affect whether they are gathered in this function or not. However, two `GlobalPtrStmt`s pointing to the same SNode makes it not uniquely accessed in struct-fors because we are parallel accessing them in the struct-for loop.

xumingkuan (commented):
Cool, thanks!

xumingkuan:
I'm not sure if this is correct -- external pointers might need to be treated like SNodes when it comes to parallel.
Suppose `x` is a SNode and `a` is an external array here:
```python
for i in x:
  a[i] += 1
  a[i + 1] += 1
```
Both accesses `a[i]` and `a[i + 1]` are unique in the loop, but if we demote the atomics, there can be data races: `a[i + 1]` when `i = 0` is the same as `a[i]` when `i = 1`.

Maybe it's better to have a test like this, and we may need an `std::unordered_map` from the external arrays to the external pointers to store the unique access results. (Note: this is assuming no two external arrays overlap here and IIRC we can assume this. If they overlap, things will become much more complicated and there's not much we can do.)

taichi/transforms/demote_atomics.cpp
```diff
@@ -13,6 +13,7 @@ TLANG_NAMESPACE_BEGIN
 class DemoteAtomics : public BasicStmtVisitor {
  private:
   std::unordered_map<const SNode *, GlobalPtrStmt *> loop_unique_ptr_;
+  std::unordered_set<ExternalPtrStmt *> loop_unique_arr_ptr_;
```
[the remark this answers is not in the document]
bobcao3: Since ndarray is also part of this, I don't know what's the proper name for it. Ext arr does make sense since there is ExternalPtrStmt

taichi/analysis/gather_uniquely_accessed_pointers.cpp
bobcao3: This is with the case where the lowering pass has transformed a 2d or 3d range / struct on `ti.any_arr` to a linear index, and generated IR that calculates the ND index. In those case the offloaded stmt use a 1d index, but the original code written by the user is ND.

taichi/analysis/gather_uniquely_accessed_pointers.cpp
bobcao3: ext_arr & ndarrays -> `ti.any_arr`

bobcao3:
> I'm not sure if this is correct -- external pointers might need to be treated like SNodes when it comes to parallel. Suppose `x` is an SNode and `a` is an external array here:
> 
> ```python
> for i in x:
>   a[i] += 1
>   a[i + 1] += 1
> ```
> 
> Both accesses `a[i]` and `a[i + 1]` are unique in the loop, but if we demote the atomics, there can be data races: `a[i + 1]` when `i = 0` is the same as `a[i]` when `i = 1`.
> 
> Maybe it's better to have a test like this, and we may need an `std::unordered_map` from the external arrays to the external pointers to store the unique access results. (Note: this is assuming no two external arrays overlap here and IIRC we can assume this. If they overlap, things will become much more complicated and there's not much we can do.)

Makes sense! So we would need an unordered map from arrays to pointers, similar to SNodes, where if all pointers statement has the same index it is fine.

taichi/analysis/gather_uniquely_accessed_pointers.cpp
xumingkuan: I see. So I wonder if we still have `LoopIndexStmt` in this offloaded stmt using the 1d index? If so, this line https://github.com/taichi-dev/taichi/blob/c231aa3b482672a7f37e781df286380ce6384fd5/taichi/analysis/gather_uniquely_accessed_pointers.cpp#L34 may not be able to detect the uniqueness properly -- note that `num_different_loop_indices` is a variable in this class used both in `is_ptr_indices_loop_unique(GlobalPtrStmt *stmt)` and `is_ptr_indices_loop_unique(ExternalPtrStmt *stmt)`.

And, what if we have multiple ND-arrays with different numbers of dimensions in one offloaded stmt? Maybe we can refactor the `num_different_loop_indices` used here (we may need to rename this variable and keep the original one equal to the original dimensions of the offload) into a map from the external arrays to the number of dimensions.

We may need to add some notes that the `DecorationStmt` is essential for alias analysis and cannot be removed until no more alias analysis is going to be performed (e.g., in codegen phase).

strongoier:
> I wonder if DecoratorRecorder can be merged with this? @strongoier @PGZXB

They are at different levels. `DecoratorRecorder` itself is a frontend thing, which is mainly used for passing decorations of for-loops into our frontend IR. `DecorationStmt` in this PR is a CHI IR stmt similar to `LoopUniqueStmt`.

That said, the way to properly express decorations is still worth discussion. I think one thing to unify here is whether we should couple these decorations with the loop (an offloaded task) or with the actual stmt inside.

taichi/analysis/gather_uniquely_accessed_pointers.cpp
bobcao3: I'm relying on this decoration only being added for loops that we transformed. If different dimensions of NDarrays being used in the loop, the loop itself is essentially only a certain amount of dimensions. The frontend statement does not ever use the linearized index in any ways, and the backend IR neither will, thus I don't think it will cause a problem. If these statements are removed, alias analysis will naturally not happen because the linearized index statement is never used in any way.

taichi/analysis/gather_uniquely_accessed_pointers.cpp
```diff
@@ -172,7 +226,60 @@ class UniquelyAccessedSNodeSearcher : public BasicStmtVisitor {
...
+           */
+          if (accessed_ptr->second) {
+            ExternalPtrStmt *other_ptr = accessed_ptr->second;
+            TI_ASSERT(stmt->indices.size() == other_ptr->indices.size());
+            for (int axis = 0; axis < stmt->indices.size(); axis++) {
+              Stmt *this_index = stmt->indices[axis];
+              Stmt *other_index = other_ptr->indices[axis];
+              // We only compare unique indices here.
+              // Since both pointers are loop-unique, all the unique indices
+              // need to be the same for both to be uniquely accessed
```
xumingkuan, line 265: I think we should also care about other indices... e.g., what if we access `a[i, j, 1]` and `a[3, 2, 1]`? There can be a data race when the `i = 3, j = 2` thread is accessing `a[i, j, 1]` when another thread is accessing `a[3, 2, 1]`.

taichi/analysis/gather_uniquely_accessed_pointers.cpp
xumingkuan: I see, so IIUC there is only one set of decoration statements in one loop, and only if we are looping over an NDarray, right?

We should add some comments at
https://github.com/taichi-dev/taichi/blob/c231aa3b482672a7f37e781df286380ce6384fd5/taichi/analysis/gather_uniquely_accessed_pointers.cpp#L189
 that we are initializing `num_different_loop_indices` here in the NDarray case. It can be confusing because we initialize this variable at two places that are far away.

xumingkuan (commented):
Thanks for the updates!

taichi/analysis/gather_uniquely_accessed_pointers.cpp
bobcao3: That won't pass this test since [3,2,1] is not loop unique (just loop invariant). Once we have one of them being not loop unique, we will have the access to this NDarray marked as not unique.

bobcao3: And for loop unique stmts, we are checking that the unique indicies of those statements have the same value, so if two already loop unique pointer uses the same index for them, they should be uniquely accessed as well

xumingkuan: Makes sense, now I see what's happening here. I think we should rename `is_loop_unique` to something like `is_index_partially_loop_unique` or some better names, because "loop-unique" should refer to the accesses like `a[i, j, 1]`, and `i` itself is not loop-unique (`i` combined with `j` is loop-unique).

BTW, this is a little bit conservative in some corner cases: suppose we have accesses `a[i, j, i]` and `a[i, j, j]`. We can demote the atomics but it's hard to see. If we really want to demote them, we can maintain the set of identical loop-unique indices' positions (indices 0 and 1 in this case, which are `i` and `j`). Then if we have a third access `a[i, i, j]`, we can see that we cannot demote the atomics because although having both `a[i, j, j]` and `a[i, i, j]` is good, the set of identical loop-unique index's position is only index 0 (which is `i`), which is not enough for the whole access to be loop-unique.

PR #4274, opened 2022-02-14
[bug] Disallow function definition inside ti.func/kernel

### LINKED ISSUE #4240: Nested function definition inside @ti.func
One can define a nested function inside a @ti.func, but cannot actually call it. The error message is misleading:

**Example code**:

```python
import taichi as ti
ti.init()

@ti.func
def test():
    def test1():
        pass
    # the funny thing is it's ok if we don't call test1() here ...
    test1()

@ti.kernel
def main():
    test()

main()
```

**The error message**

```
[Taichi] version 0.8.11, llvm 10.0.0, commit 9d0973e5, linux, python 3.8.10
[Taichi] Starting on arch=x64
  File "test.py", line 15, in <module>
    main()
  File "/home/taichigraphics/.local/lib/python3.8/site-packages/taichi/lang/kernel_impl.py", line 736, in wrapped
taichi.lang.exception.TaichiNameError: 
On line 13 of file "test.py":
    test()
    ^^^^^^
On line 9 of file "test.py":
    test1()
    ^^^^^
Name "test1" is not defined
```

Shall we fix this so that nested function declarations are allowed, or we explicitly tell the user not to do so?

lin-hitonami:
Related issue = close #4240

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -6,7 +6,7 @@ from sys import version_info
 
 import astor
 from taichi._lib import core as _ti_core
-from taichi.lang import expr, impl, kernel_arguments, kernel_impl, matrix, mesh
+from taichi.lang import expr, impl, kernel_arguments, matrix, mesh
 from taichi.lang import ops as ti_ops
 from taichi.lang._ndrange import _Ndrange, ndrange
 from taichi.lang.ast.ast_transformer_utils import Builder, LoopStatus
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -403,6 +403,12 @@ class ASTTransformer(Builder):
 
     @staticmethod
     def build_FunctionDef(ctx, node):
+        if ctx.visited_funcdef:
+            raise TaichiSyntaxError(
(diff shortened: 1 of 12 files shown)


python/taichi/lang/ast/ast_transformer.py
```diff
@@ -403,6 +403,12 @@ def build_Call(ctx, node):
 
     @staticmethod
     def build_FunctionDef(ctx, node):
+        if ctx.visited_funcdef:
+            raise TaichiSyntaxError(
+                f"Function definition not allowed in 'ti.{'kernel' if ctx.is_kernel else 'func'}'."
```
mzmzm, line 408:
```suggestion
                f"Function definition is not allowed in 'ti.{'kernel' if ctx.is_kernel else 'func'}'."
```

mzmzm (approved):
LGTM!

tests/python/test_syntax_errors.py
```diff
@@ -317,3 +317,31 @@ def foo():
...
+
+@test_utils.test()
+def test_funcdef_in_kernel():
+    @ti.kernel
+    def foo():
+        def bar():
+            pass
+
+    with pytest.raises(ti.TaichiSyntaxError,
+                       match="Function definition not allowed in 'ti.kernel'"):
```
mzmzm, line 330:
```suggestion
                       match="Function definition is not allowed in 'ti.kernel'"):
```

Issue #4266, opened 2022-02-14
Mesh test symbols not found error in python 3.10


qiao-bo:
Several test cases in test_mesh fail when using python 3.10. `python -m pytest tests/python/test_mesh.py -s`.

To reproduce:
Create python 3.10 env, e.g., using conda

The following is a minimal example:
```
import os
import numpy as np
import taichi as ti

this_dir = os.path.dirname(os.path.abspath(__file__))
model_file_path = os.path.join(this_dir, 'ell.json')

ti.init(arch=ti.x64)

def test_mesh_for(cell_reorder=False, vert_reorder=False, extra_tests=True):
    mesh_builder = ti.Mesh.Tet()
    mesh_builder.verts.place({'t': ti.i32}, reorder=vert_reorder)
    mesh_builder.cells.place({'t': ti.i32}, reorder=cell_reorder)
    mesh_builder.cells.link(mesh_builder.verts)
    mesh_builder.verts.link(mesh_builder.cells)
    mesh_builder.cells.link(mesh_builder.cells)
    mesh_builder.verts.link(mesh_builder.verts)
    model = mesh_builder.build(ti.Mesh.load_meta(model_file_path))

    @ti.kernel
    def cell_vert():
        for c in model.cells:
            for j in range(c.verts.size):
                c.t += c.verts[j].id

    cell_vert()

test_mesh_for(False, False)
```
Output:
```
Internal error occurred. Check out this page for possible solutions:                                                           
https://docs.taichi.graphics/lang/articles/misc/install        
Expected<T> must be checked before access or destruction.                                                                      
Unchecked Expected<T> contained error:                                                                                         
Failed to materialize symbols: { (0x55fea1fc5790, { cell_vert_c52_0_kernel_249_mesh_for }) }[E 02/14/22 10:56:07.821 525265] Re
ceived signal 6 (Aborted)        
```
The error only happens when using `arch=ti.cpu`.

qiao-bo:
CC @g1n0st, thanks!

PR #4275, opened 2022-02-14
[lang] Fix bls_buffer allocation of x64 crashed in py3.10

g1n0st:
Related issue = https://github.com/taichi-dev/taichi/issues/4266


qiao-bo (approved):
Thanks for the quick fix!

PR #4276, opened 2022-02-14
[llvm] Use GEP for array access instead of ptrtoint/inttoptr

strongoier:
Related issue = #2590

Using `ptrtoint/inttoptr` for array access prevents LLVM from demoting array alloca into individual allocas, which further prevents mem2reg optimizations. After this PR, with `dynamic_index=True`, we can have the same runtime performance as with `dynamic_index=False` for old code which can only access matrices with constant indices.

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -69,7 +69,7 @@ class Matrix(TaichiOperations):
                     self.local_tensor_proxy = impl.expr_init_local_tensor(
                         [len(n)], dt,
                         expr.make_expr_group([expr.Expr(x) for x in n]))
-                    self.dynamic_index_stride = ti_core.data_type_size(dt)
+                    self.dynamic_index_stride = 1
                     mat = []
                     for i in range(len(n)):
                         mat.append(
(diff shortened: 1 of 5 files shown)


python/taichi/lang/matrix.py
```diff
@@ -69,7 +69,7 @@ def __init__(self, n=1, m=1, dt=None, suppress_warning=False):
                     self.local_tensor_proxy = impl.expr_init_local_tensor(
                         [len(n)], dt,
                         expr.make_expr_group([expr.Expr(x) for x in n]))
-                    self.dynamic_index_stride = ti_core.data_type_size(dt)
+                    self.dynamic_index_stride = 1
```
k-ye, line 72: Has the meaning of `dynamic_index_stride` been changed (from bytes to index)?

k-ye (approved):
LGTM!

python/taichi/lang/matrix.py
strongoier: After this PR, handling dynamic indexing of matrix fields requires manipulating pointers directly, while handling dynamic indexing of local matrices only requires array indices. However, currently they share the same expression (`TensorElementExpression`) and statement (`PtrOffsetStmt`), so the `stride` actually stands for two different things (bytes & indices) in these two cases. A future step must be to split them up. As it will add even one more type of pointer statement, I suggest doing this while we refactor those pointer statements.

PR #4277, opened 2022-02-14
[spirv] Fix buffer info compare to fix external array bind point

bobcao3:
Otherwise we will bind the same buffer over and over again.


ailzhang (approved):
LGTM!

PR #4282, opened 2022-02-15
[vulkan] Reduce runtime host overhead

bobcao3:
... and removed an unnecessary barrier

### DIFF
--- a/taichi/backends/vulkan/runtime.cpp
+++ b/taichi/backends/vulkan/runtime.cpp
@@ -383,6 +383,7 @@ void CompiledTaichiKernel::generate_command_list(
 VkRuntime::VkRuntime(const Params &params)
     : device_(params.device), host_result_buffer_(params.host_result_buffer) {
   TI_ASSERT(host_result_buffer_ != nullptr);
+  current_cmdlist_pending_since_ = high_res_clock::now();
   init_buffers();
 }
 
@@ -536,6 +537,7 @@ void VkRuntime::launch_kernel(KernelHandle handle, RuntimeContext *host_ctx) {
   // Create new command list if current one is nullptr
   if (!current_cmdlist_) {
     ctx_buffers_.clear();
+    current_cmdlist_pending_since_ = high_res_clock::now();
     current_cmdlist_ = device_->get_compute_stream()->new_command_list();
   }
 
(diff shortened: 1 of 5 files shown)


taichi/backends/vulkan/runtime.cpp
```diff
@@ -560,6 +562,19 @@ void VkRuntime::launch_kernel(KernelHandle handle, RuntimeContext *host_ctx) {
     }
   }
 
+  // If we have accumulated some work but does not require sync
+  // and if the accumulated cmdlist has been pending for some time
+  // launch the cmdlist to start processing.
+  if (current_cmdlist_) {
+    constexpr uint64_t max_pending_time = 2000;  // 3000us = 2ms
```
qiao-bo, line 569:
```suggestion
    constexpr uint64_t max_pending_time = 2000;  // 2000us = 2ms
```

qiao-bo: btw, do we need to elaborate here why 2ms?

bobcao3: It's just a value I just decided on. (A queue submission is on the scale of 0.1ms) Doesn't seem to cause too much overhead & pretty good at overlapping CPU work with GPU especially for the JIT scenario.

taichi/backends/vulkan/vulkan_device.cpp
```diff
@@ -724,13 +744,14 @@ void VulkanCommandList::bind_resources(ResourceBinder *ti_binder) {
...
-    vkCmdBindDescriptorSets(buffer_->buffer, bind_point,
-                            current_pipeline_->pipeline_layout()->layout,
+    vkCmdBindDescriptorSets(buffer_->buffer, bind_point, pipeline_layout,
                             /*firstSet=*/0,
                             /*descriptorSetCount=*/1, &set->set,
                             /*dynamicOffsetCount=*/0,
                             /*pDynamicOffsets=*/nullptr);
     buffer_->refs.push_back(set);
+
+    last_bound_layout_ = pipeline_layout;
```
qiao-bo, line 754: this seems not used anywhere?

qiao-bo (approved):
a few nits, LGTM

taichi/backends/vulkan/vulkan_device.cpp
bobcao3: good catch

ailzhang (approved):
LGTM!

Issue #3903, opened 2021-12-29
element_shape should be treated at compile time information for ndarrays


ailzhang:
Now they're treated at runtime information. 
Repro script:
```
import taichi as ti
import numpy as np
ti.init(arch=ti.opengl, log_level=ti.TRACE)

def _test_compiled_functions():
    @ti.kernel
    def func(a: ti.any_arr(element_dim=1)):
        for i in range(5):
            for j in range(4):
                a[i][j * j] = j * j

    v = ti.Vector.ndarray(10, ti.i32, 5)
    func(v)
    assert ti.get_runtime().get_num_compiled_functions() == 1
    v = ti.Vector.ndarray(11, ti.i32, 5)
    func(v)
    print(ti.get_runtime().get_num_compiled_functions())

_test_compiled_functions()
```
Note that although we compiled twice for the above program, the generated opengl kernels are indeed identical (shown below). In other words, we ought to have used element_shape as compile time information but we didn't. 
This can also be verified in the code snippet below that `_args_i32_[16 + 0 * 8 + 1]` (element_shape=10 or 11), which should have been a constant, but now it's loaded as a runtime variable.  
Side note: the reason we recompile for different element_shape now is that we force that to happen in python. taichi/kernel_impl.py at master · taichi-dev/taichi Unfortunately we only enforce recompile but didn't make use of the compile time information at all. 
```
#version 430 core
layout(local_size_x = 128, local_size_y = 1, local_size_z = 1) in;
precision highp float;
layout(std430, binding = 2) buffer args_i32 { int _args_i32_[];};
layout(std430, binding = 4) buffer arr0_i32 { int _arr0_i32_[];};

const float inf = 1.0f / 0.0f;
const float nan = 0.0f / 0.0f;
void func_c44_00()
{ // range for
  // range known at compile time
  int _sid0 = int(gl_GlobalInvocationID.x);
  for (int _sid = _sid0; _sid < (5); _sid += int(gl_WorkGroupSize.x * gl_NumWorkGroups.x)) {
    int _itv = 0 + _sid;
      int B = _itv;
      int C = int(0);
      int D = int(4);
      for (int E_ = C; E_ < D; E_ += 1) {
        int E = E_;
        int F = E;
        int G = F * F;
        int _li_I = 0;
        int _s0_arr0 = _args_i32_[16 + 0 * 8 + 0];
        int _s1_arr0 = _args_i32_[16 + 0 * 8 + 1];
        { // linear seek
[...]

PR #4284, opened 2022-02-15
[opengl] Use element shape as compile information for OpenGL backend

turbo0628:
Related issue = #3903

Motivation:

Taichi automatically re-compiles the kernel when element shapes change. Therefore, element shape is compile-time static. This PR embed the element shape information into shader code for OpenGL backend.

Briefings:
* Changed `ExternalTensorExpression` and `ExternalPtrStmt` to get layout and element shape information.
* Changed OpenGL codegen to embed the shapes.

### DIFF
--- a/python/taichi/lang/kernel_arguments.py
+++ b/python/taichi/lang/kernel_arguments.py
@@ -61,8 +61,9 @@ def decl_any_arr_arg(dtype, dim, element_shape, layout):
     if layout == Layout.AOS:
         element_dim = -element_dim
     return AnyArray(
-        _ti_core.make_external_tensor_expr(dtype, dim, arg_id, element_dim),
-        element_shape, layout)
+        _ti_core.make_external_tensor_expr(dtype, dim, arg_id, element_dim,
+                                           element_shape), element_shape,
+        layout)
 
 
 def decl_ret(dtype):
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -532,29 +532,29 @@ class KernelGen : public IRVisitor {
     auto element_shape = stmt->element_shape;
     std::vector<std::string> size_var_names;
     std::vector<std::string> element_shape_size_var_names;
-    const int layout_SOA = 1;
-    const int layout_AOS = 0;
-    int layout = stmt->element_dim <= 0 ? layout_AOS : layout_SOA;
+    enum ExternalArrayLayout { layout_AOS = 0, layout_SOA = 1 };
+    auto layout = stmt->element_dim <= 0 ? layout_AOS : layout_SOA;
 
     if (element_shape.size() > 0) {
-      int elem_beg = num_indices - element_shape.size();
-      int elem_end = num_indices;
+      int elem_beg(0), elem_end(0);
       if (layout == layout_SOA) {
         elem_beg = 0;
         elem_end = element_shape.size();
+      } else {
+        elem_beg = num_indices - element_shape.size();
+        elem_end = num_indices;
       }
       for (int i = elem_beg; i < elem_end; i++) {
         used.int32 = true;
         std::string var_name = fmt::format("_s{}_{}{}", i, "arr", arg_id);
         if (!loaded_args_.count(var_name)) {
-          // The passed element shape vector is reversed w.r.t. to args buffer
           emit("int {} = {};", var_name, element_shape[i - elem_beg]);
           loaded_args_.insert(var_name);
         }
         element_shape_size_var_names.push_back(std::move(var_name));
       }
     }
-    // Args buffer arrange dimensions higher to lower
+    // Args buffer arrange dimensions from outer to inner
     // AoS args buffer:   array_shape|element_shape
     // SoA args buffer: element_shape|array_shape
     //
@@ -563,11 +563,13 @@ class KernelGen : public IRVisitor {
     // ti.Matrix.ndarray(3, 2, ti.f32, (5, 4), layout=ti.Layout.SOA)
     // args buffer: 3, 2, 5, 4
 
-    int ind_beg = 0;
-    int ind_end = num_indices - element_shape.size();
+    int ind_beg(0), ind_end(0);
     if (layout == layout_SOA) {
       ind_beg = element_shape.size();
       ind_end = num_indices;
+    } else {
+      ind_beg = 0;
+      ind_end = num_indices - element_shape.size();
     }
     for (int i = ind_beg; i < ind_end; i++) {
       used.buf_args = true;
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -536,7 +536,8 @@ class KernelGen : public IRVisitor {
     auto layout = stmt->element_dim <= 0 ? layout_AOS : layout_SOA;
 
     if (element_shape.size() > 0) {
-      int elem_beg(0), elem_end(0);
+      int elem_beg = 0;
+      int elem_end = 0;
       if (layout == layout_SOA) {
         elem_beg = 0;
         elem_end = element_shape.size();
@@ -562,8 +563,8 @@ class KernelGen : public IRVisitor {
     // args buffer: 5, 4, 3, 2
     // ti.Matrix.ndarray(3, 2, ti.f32, (5, 4), layout=ti.Layout.SOA)
     // args buffer: 3, 2, 5, 4
-
-    int ind_beg(0), ind_end(0);
+    int ind_beg = 0;
+    int ind_end = 0;
     if (layout == layout_SOA) {
       ind_beg = element_shape.size();
       ind_end = num_indices;
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -529,8 +529,47 @@ class KernelGen : public IRVisitor {
     const auto *argload = stmt->base_ptrs[0]->as<ArgLoadStmt>();
     const int arg_id = argload->arg_id;
     const int num_indices = stmt->indices.size();
+    auto element_shape = stmt->element_shape;
     std::vector<std::string> size_var_names;
-    for (int i = 0; i < num_indices; i++) {
+    std::vector<std::string> element_shape_size_var_names;
+    const int layout_SOA = 1;
+    const int layout_AOS = 0;
+    int layout = stmt->element_dim <= 0 ? layout_AOS : layout_SOA;
+
+    if (element_shape.size() > 0) {
+      int elem_beg = num_indices - element_shape.size();
+      int elem_end = num_indices;
+      if (layout == layout_SOA) {
+        elem_beg = 0;
+        elem_end = element_shape.size();
+      }
+      for (int i = elem_beg; i < elem_end; i++) {
+        used.int32 = true;
+        std::string var_name = fmt::format("_s{}_{}{}", i, "arr", arg_id);
+        if (!loaded_args_.count(var_name)) {
+          // The passed element shape vector is reversed w.r.t. to args buffer
+          emit("int {} = {};", var_name, element_shape[i - elem_beg]);
+          loaded_args_.insert(var_name);
+        }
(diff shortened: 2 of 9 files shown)


taichi/ir/statements.h
```diff
@@ -263,9 +263,15 @@ class ExternalPtrStmt : public Stmt {
  public:
   LaneAttribute<Stmt *> base_ptrs;
   std::vector<Stmt *> indices;
+  std::vector<int> element_shape;
```
ailzhang, line 266: FYI ir_printer.cpp also need to be updated to include this info for debugging purpose. ;)

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -529,8 +529,47 @@ class KernelGen : public IRVisitor {
...
+      int elem_end = num_indices;
+      if (layout == layout_SOA) {
+        elem_beg = 0;
+        elem_end = element_shape.size();
+      }
+      for (int i = elem_beg; i < elem_end; i++) {
+        used.int32 = true;
+        std::string var_name = fmt::format("_s{}_{}{}", i, "arr", arg_id);
+        if (!loaded_args_.count(var_name)) {
+          // The passed element shape vector is reversed w.r.t. to args buffer
```
ailzhang, line 550: A more accurate way to say this is `element shape in args buffer starts from outermost to innermost dimension` or sth similar?

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -529,8 +529,47 @@ class KernelGen : public IRVisitor {
...
+    // Args buffer arrange dimensions higher to lower
+    // AoS args buffer:   array_shape|element_shape
+    // SoA args buffer: element_shape|array_shape
+    //
+    // ti.Matrix.ndarray(3, 2, ti.f32, (5, 4), layout=ti.Layout.AOS)
+    // args buffer: 5, 4, 3, 2
+    // ti.Matrix.ndarray(3, 2, ti.f32, (5, 4), layout=ti.Layout.SOA)
+    // args buffer: 3, 2, 5, 4
+
+    int ind_beg = 0;
```
ailzhang, line 566: To make it easier to read, can we set elem_beg/elem_end/ind_beg/ind_end in one if/else block? ;)

taichi/ir/statements.cpp
```diff
@@ -26,9 +26,15 @@ bool UnaryOpStmt::same_operation(UnaryOpStmt *o) const {
   return false;
 }
 
-ExternalPtrStmt::ExternalPtrStmt(const LaneAttribute<Stmt *> &base_ptrs,
-                                 const std::vector<Stmt *> &indices)
-    : base_ptrs(base_ptrs), indices(indices) {
+ExternalPtrStmt::ExternalPtrStmt(
+    const LaneAttribute<Stmt *> &base_ptrs,
+    const std::vector<Stmt *> &indices,
+    const std::vector<int> &element_shape = std::vector<int>(),
```
ailzhang, line 32: Can we separate this into two constructors? ;) Specifically I'm interested in if there's any kernels left only proving element_dim without element_shape (I guess yes, those predefined kernels like ext_arr_to_numpy might be still use them. This is totally fine but it's better that we can get it when we read code.

python/taichi/lang/kernel_arguments.py
```diff
@@ -61,8 +61,9 @@ def decl_any_arr_arg(dtype, dim, element_shape, layout):
     if layout == Layout.AOS:
         element_dim = -element_dim
     return AnyArray(
-        _ti_core.make_external_tensor_expr(dtype, dim, arg_id, element_dim),
-        element_shape, layout)
+        _ti_core.make_external_tensor_expr(dtype, dim, arg_id, element_dim,
```
ailzhang, line 64: btw we wanted to pass in a TensorType instead of a flattened primitive dtype but that can be a followup I think.

ailzhang (commented):
Nice work!!

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -529,8 +529,47 @@ class KernelGen : public IRVisitor {
     const auto *argload = stmt->base_ptrs[0]->as<ArgLoadStmt>();
     const int arg_id = argload->arg_id;
     const int num_indices = stmt->indices.size();
+    auto element_shape = stmt->element_shape;
     std::vector<std::string> size_var_names;
-    for (int i = 0; i < num_indices; i++) {
+    std::vector<std::string> element_shape_size_var_names;
+    const int layout_SOA = 1;
```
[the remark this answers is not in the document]
turbo0628: Good, will change to a local enum

taichi/ir/statements.cpp
turbo0628: Two overloaded constructors were placed at the very beginning. However, I could only find one single usage in the codebase so I merged the constructors. The element_dim is only available in ExternalTensorExpression previously, so they are currently used together for ExternalPtrStmt.

taichi/ir/frontend_ir.h
```diff
@@ -408,11 +408,31 @@ class ExternalTensorExpression : public Expression {
...
+                           int arg_id,
+                           int element_dim,
+                           const std::vector<int> &element_shape)
+      : dt(dt),
+        dim(dim),
+        arg_id(arg_id),
+        element_dim(element_dim),
+        element_shape(element_shape) {
     set_attribute("dim", std::to_string(dim));
   }
```
turbo0628, line 438: @ailzhang What do you think about these two constructors here? I just forgot to merge them :)

taichi/ir/statements.h
turbo0628: Thanks! I'm looking into this!

taichi/backends/opengl/codegen_opengl.cpp
turbo0628: Nice suggestion, thanks!

python/taichi/lang/kernel_arguments.py
turbo0628: Let's make it in a followup PR, shall we leave a tracker?

taichi/ir/frontend_ir.h
ailzhang: a cleaner way to write the second is 
```
...
: ExternalTensorExpression(dt, dim, arg_id, element_dim), element_shape(element_shape) {}
```

ailzhang: Btw I probably put my comment in the wrong place :P Here's the one I care about https://github.com/taichi-dev/taichi/blob/master/taichi/program/program.cpp#L432

taichi/ir/statements.cpp
ailzhang: I guess my concern here is moving defaults to header files ;)

turbo0628: I see :)

taichi/ir/frontend_ir.h
turbo0628: Then let's keep two constructors.

taichi/ir/statements.h
turbo0628: I've encountered some problems here. Print the element shape in IR would crash some place in module.py. Is there any  document for IR print conventions?

turbo0628: Done :)

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -529,8 +529,49 @@ class KernelGen : public IRVisitor {
...
     const int num_indices = stmt->indices.size();
+    auto element_shape = stmt->element_shape;
     std::vector<std::string> size_var_names;
-    for (int i = 0; i < num_indices; i++) {
+    std::vector<std::string> element_shape_size_var_names;
+    enum ExternalArrayLayout { layout_AOS = 0, layout_SOA = 1 };
+    auto layout = stmt->element_dim <= 0 ? layout_AOS : layout_SOA;
+
+    if (element_shape.size() > 0) {
+      int elem_beg(0), elem_end(0);
```
ailzhang, line 539: btw in this case we usually write `int elem_beg = 0` directly :P

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -529,8 +529,49 @@ class KernelGen : public IRVisitor {
...
+    // Args buffer arrange dimensions from outer to inner
+    // AoS args buffer:   array_shape|element_shape
+    // SoA args buffer: element_shape|array_shape
+    //
+    // ti.Matrix.ndarray(3, 2, ti.f32, (5, 4), layout=ti.Layout.AOS)
+    // args buffer: 5, 4, 3, 2
+    // ti.Matrix.ndarray(3, 2, ti.f32, (5, 4), layout=ti.Layout.SOA)
+    // args buffer: 3, 2, 5, 4
+
+    int ind_beg(0), ind_end(0);
```
ailzhang, line 566: ditto

taichi/ir/frontend_ir.h
```diff
@@ -408,11 +408,30 @@ class ExternalTensorExpression : public Expression {
...
 
+  // Fill element shape if compile-time specialization is desired.
+  std::vector<int> element_shape;
+
   ExternalTensorExpression(const DataType &dt,
                            int dim,
                            int arg_id,
                            int element_dim)
-      : dt(dt), dim(dim), arg_id(arg_id), element_dim(element_dim) {
+      : ExternalTensorExpression(dt,
```
ailzhang, line 418: Manual initialization for std::vector isn't needed so a better way to write this is 
```
  ExternalTensorExpression(const DataType &dt,
                           int dim,
                           int arg_id,
                           int element_dim):
      : dt(dt),
        dim(dim),
        arg_id(arg_id),
        element_dim(element_dim) {
    set_attribute("dim", std::to_string(dim));
}
  ExternalTensorExpression(const DataType &dt,
                           int dim,
                           int arg_id,
                           int element_dim,
                           const std::vector<int> &element_shape) :
 ExternalTensorExpression(dt,dim, arg_id, element_dim), element_shape(element_shape) {}
```

taichi/ir/statements.h
```diff
@@ -263,10 +263,19 @@ class ExternalPtrStmt : public Stmt {
...
+  // SOA: element_dim > 0
+  int element_dim;
 
   ExternalPtrStmt(const LaneAttribute<Stmt *> &base_ptrs,
                   const std::vector<Stmt *> &indices);
 
+  ExternalPtrStmt(const LaneAttribute<Stmt *> &base_ptrs,
+                  const std::vector<Stmt *> &indices,
+                  const std::vector<int> &element_shape,
+                  const int element_dim);
```
ailzhang, line 277: const is not necessary

ailzhang (commented):
Thanks! Only had a few nits and then it's good to go! :D

strongoier (approved):
LGTM!

taichi/ir/frontend_ir.h
turbo0628: Well the compiler complains "a delegating constructor cannot have other mem-initializers".

I can just leave the std vector a blank initializer

ailzhang: oh sorry so 
```
ExternalTensorExpression(dt,dim, arg_id, element_dim) {this->element_shape = element_shape;}
```
should work?

taichi/ir/statements.h
turbo0628: Fixed, thanks ;)

taichi/ir/frontend_ir.h
turbo0628: Ah this should work, but does it look better than mem initializers?

turbo0628: Looks better actually :)

Thanks!

taichi/backends/opengl/codegen_opengl.cpp
turbo0628: Fixed

taichi/backends/opengl/codegen_opengl.cpp
turbo0628: Fixed +1 lol

ailzhang (approved):
Thanks!!

PR #4302, opened 2022-02-16
[lang] Hide internal apis about Fields

mzmzm:
Related issue = #3782

### DIFF
--- a/python/taichi/lang/field.py
+++ b/python/taichi/lang/field.py
@@ -55,7 +55,7 @@ class Field:
         """
         return self.snode.name
 
-    def _parent(self, n=1):
+    def parent(self, n=1):
         """Gets an ancestor of the representative SNode in the SNode tree.
 
         Args:
--- a/python/taichi/lang/field.py
+++ b/python/taichi/lang/field.py
@@ -20,6 +20,15 @@ class Field:
         self.grad = None
 
     @property
+    def snode(self):
+        """Gets representative SNode for info purposes.
+
+        Returns:
+            SNode: Representative SNode (SNode of first field member).
+        """
+        return self._snode
+
+    @property
     def _snode(self):
         """Gets representative SNode for info purposes.
 
--- a/python/taichi/lang/field.py
+++ b/python/taichi/lang/field.py
@@ -64,6 +64,15 @@ class Field:
         """
         return self._snode._name
 
+    def parent(self, n=1):
+        """Gets an ancestor of the representative SNode in the SNode tree.
+        Args:
+            n (int): the number of levels going up from the representative SNode.
+        Returns:
+            SNode: The n-th parent of the representative SNode.
+        """
+        return self.snode.parent(n)
+
     def _get_field_members(self):
         """Gets field members.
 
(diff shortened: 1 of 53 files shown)


strongoier (commented):
You may need to cancel modifications to `loop_range`, otherwise it will conflict with #4303.

mzmzm:
> You may need to cancel modifications to `loop_range`, otherwise it will conflict with #4303.

Done

python/taichi/lang/struct.py
```diff
@@ -293,60 +293,60 @@ class StructField(Field):
     def __init__(self, field_dict, name=None):
         # will not call Field initializer
         self.field_dict = field_dict
-        self._name = name
-        self.register_fields()
+        self.__name = name
```
strongoier, line 296: For this we can directly say `self.name = name` and remove the property below.

tests/python/test_struct_for_intermediate.py
```diff
@@ -12,7 +12,7 @@ def _test_nested():
 
     @ti.kernel
     def iterate():
-        for i, j in x.parent():
+        for i, j in x._parent():
```
strongoier, line 15: This is a huge break of user code. Let's go back to `parent()`.

python/taichi/lang/struct.py
mzmzm: done

tests/python/test_struct_for_intermediate.py
mzmzm: done

python/taichi/lang/field.py
```diff
@@ -220,6 +218,26 @@ class ScalarField(Field):
...
+        """Gets corresponding SNode for info purposes.
+
+        Returns:
+            SNode: corresponding SNode (SNode of first field member).
+        """
+        return self._snode
+
+    def parent(self, n=1):
+        """Gets an ancestor of the representative SNode in the SNode tree.
+
```
strongoier, line 232: `snode` and `parent` should be moved back to the base class.

python/taichi/lang/matrix.py
```diff
@@ -1100,6 +1100,15 @@ def __init__(self, _vars, n, m):
         self.m = m
         self.dynamic_index_stride = None
 
+    @property
+    def snode(self):
+        """Gets corresponding SNode for info purposes.
```
strongoier, line 1105: ditto

python/taichi/lang/field.py
```diff
@@ -44,29 +53,27 @@ def dtype(self):
...
 
         Returns:
             str: Field name.
         """
-        return self.snode._name
+        return self._snode._name
 
     def parent(self, n=1):
         """Gets an ancestor of the representative SNode in the SNode tree.
-
```
strongoier, line 60: Why remove this line?

python/taichi/lang/field.py
```diff
@@ -44,29 +53,27 @@ def dtype(self):
...
         Args:
             n (int): the number of levels going up from the representative SNode.
-
```
strongoier, line 63: ditto

python/taichi/lang/field.py
mzmzm: fix

python/taichi/lang/field.py
mzmzm: fix

strongoier (approved):
LGTM!

Issue #3988, opened 2022-01-10
ndarray release blockers


ailzhang:
- [ ] get rid of all deprecated `ndarray_use_torch` code @qiao-bo 
- [ ] #3903 
- [ ] ndarray docs

Please feel free to add more!

ailzhang:
need element_shape info in vulkan backend as well.

ailzhang:
need update android ndarray tutorial as well.

PR #4307, opened 2022-02-16
[refactor] Remove Ndarray torch implementation and tests

qiao-bo:
Related issue = #3988 

tests merge will be followed by a separate PR.

### DIFF
--- a/tests/python/test_ndarray.py
+++ b/tests/python/test_ndarray.py
@@ -9,6 +9,9 @@ from taichi.lang.util import has_pytorch
 import taichi as ti
 from tests import test_utils
 
+if has_pytorch():
+    import torch
+
 # properties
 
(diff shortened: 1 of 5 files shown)


strongoier (approved):
LGTM!

ailzhang (approved):
LGTM！

tests/python/test_ndarray.py
```diff
@@ -567,24 +445,28 @@ def func(a: ti.any_arr(element_dim=1)):
     v = np.zeros((6, 10), dtype=np.int32)
     func(v)
     assert impl.get_runtime().get_num_compiled_functions() == 1
-    import torch
```
strongoier, line 570: Splitting this test violates its original intention. I suggest using a numpy array instead here if torch is not installed.

Issue #2498, opened 2021-07-07
Get whether the field needs gradients or not


ljcc0930:

ljcc0930:
related issue = #2499

ljcc0930:
Temporary solution: `<field>.snode.needs_grad`

PR #4308, opened 2022-02-16
[lang] Hide dtype and needs_grad from SNode

strongoier:
Related issue = #3782

Both of them only apply to leaf SNodes, i.e., fields. So users should get the info from a `Field` instance instead of a `SNode`. After this PR, `dtype` of `SNode` is hidden and `needs_grad` is removed.

### DIFF
--- a/python/taichi/lang/field.py
+++ b/python/taichi/lang/field.py
@@ -44,7 +44,7 @@ class Field:
         Returns:
             DataType: Data type of each individual value.
         """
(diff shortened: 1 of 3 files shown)


ailzhang (approved):
LGTM!

ifsheldon:
@ailzhang @strongoier I need `needs_grad` in my library `stannum`. This property dates back to #2498 and I had the conversation in the old slack channel of Taichi community.

PR #4313, opened 2022-02-17
[refactor] Allow more build types from setup.py

qiao-bo:
At the moment setup.py uses DEBUG to set whether build in `Debug` or `Release` mode. This PR adds two other options `RelWithDebInfo` and `MinSizeRel`. Similar to `DEBUG=1`, use `RELWITHDEBINFO=1` to enable the corresponding build option.


ailzhang (approved):
LGTM! Btw did you check if we have the corresponding symbols in each build type?

PR #4317, opened 2022-02-17
[Doc] Avoid log(0) problem in _funcs._randn() and update primitive_types.py

neozhaoliang:


python/taichi/_funcs.py
```diff
@@ -9,28 +9,38 @@
...
     c = ops.cos(math.tau * u2)
     return r * c
 
 
 def randn(dt=None):
-    """Generates a random number from standard normal distribution.
-
-    Implementation refers to :func:`taichi.lang.random.randn`.
+    """Generate a random float sampled from univariate standard normal
+    (Gaussian) distribution of mean 0 and variance 1, using the
```
ailzhang, line 27: can we also explicitly list the possible range of generated number so that it's straightforward? ;)

ailzhang, line 27: can we also explicitly list the possible range of generated number so that it's straightforward? ;)

PR #4322, opened 2022-02-17
[test] Add a test for autodiff/regression

Tiny-Box:
Related issue = #3580

Add some ut about regression. I observed the values output at each point as well as the original values, and it looks like you can't do simple addition or subtraction less than the threshold. Considering that this is the ut of EXAMPLE, I have directly output the image here to indicate whether it is working properly

### DIFF
--- a/python/taichi/examples/autodiff/regression.py
+++ b/python/taichi/examples/autodiff/regression.py
@@ -55,7 +55,6 @@ def initialize():
 
 
 def regress_raw():
-    # why use constant value...
     use_tape = True
 
     for i in range(1000):
--- a/python/taichi/examples/autodiff/regression.py
+++ b/python/taichi/examples/autodiff/regression.py
@@ -81,7 +81,8 @@ def draw():
     for i in range(number_coeffs):
         curve_ys += coeffs[i] * np.power(curve_xs, i)
 
-    plt.title('Nonlinear Regression with Gradient Descent (3rd order polynomial)')
+    plt.title(
+        'Nonlinear Regression with Gradient Descent (3rd order polynomial)')
     ax = plt.gca()
     ax.scatter(xs, ys, label='data', color='r')
     ax.plot(curve_xs, curve_ys, label='fitted')
--- a/python/taichi/examples/autodiff/regression.py
+++ b/python/taichi/examples/autodiff/regression.py
@@ -37,51 +37,68 @@ def update():
 xs = []
 ys = []
 
-for i in range(N):
-    v = random.random() * 5 - 2.5
-    xs.append(v)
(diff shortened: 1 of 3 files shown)


tests/python/examples/autodiff/test_regression.py
```diff
@@ -0,0 +1,58 @@
+import argparse
+
+
+def mkdir_p(dir_path):
```
Tiny-Box, line 4: Generally speaking, I don't think this kind of tool-like function is very appropriate here. But I didn't find matplotlib with a parameter like create_dir=true, so I put it here, and it might be duplicated in other UTs, any suggestions here?

k-ye: I think we can just place this here for now. In case it's used more widely, maybe put it in places like `tests/python/utils`

python/taichi/examples/autodiff/regression.py
```diff
@@ -37,51 +37,68 @@ def update():
...
-use_tape = True
+    print('y')
+    for i in range(N):
+        y.grad[i] = 1
+        ys.append(y[i])
+    print()
+
+
+def regress_raw():
+    # why use constant value...
```
k-ye, line 58: Should probably change to a cmd flag..

k-ye (approved):
LGTM!

python/taichi/examples/autodiff/regression.py
Tiny-Box: get it!

tests/python/examples/autodiff/test_regression.py
Tiny-Box: yep, I will try to adjust it out

python/taichi/examples/autodiff/regression.py
strongoier: I suggest removing this line :-)

strongoier (approved):
LGTM! Thanks for your contribution!

This PR can be merged as soon as CI is passed. The current failure is not your PR's problem and we will look into that.

strongoier:
Could you merge with our updated master to apply our fix to the CI problem? Thanks!

Tiny-Box:
OK, I will rebase the master to my branch

Yi Xu ***@***.***> 于2022年2月21日周一 10:39写道：

Tiny-Box:
Oh, I messed up, I'll cherry-pick the relevant commits here, wait a minute

Tiny-Box:
Just like regular expressions, no matter how many jobs I change, I wouldn't say I know this fucking shit... 
But the good news is it looks like it's okay this time @strongoier

PR #4323, opened 2022-02-17
[test] Add a test for simple_derivative example

Tiny-Box:
Related issue = #3580

like ut about autodiff/regression, this is also a UT that outputs images directly

### DIFF
--- a/tests/python/examples/autodiff/test_simple_derivative.py
+++ b/tests/python/examples/autodiff/test_simple_derivative.py
@@ -1,0 +1,44 @@
+import argparse
+
+
+def mkdir_p(dir_path):
+    '''Creates a directory. equivalent to using mkdir -p on the command line'''
+    from errno import EEXIST
+    from os import makedirs,path
+
+    try:
+        makedirs(dir_path)
+    except OSError as exc: # Python >2.5
+        if exc.errno == EEXIST and path.isdir(dir_path):
+            pass
+        else: raise
+
+
+def pic_simple_derivative(result_dir):
+    from taichi.examples.autodiff.simple_derivative import (initialize,
+                                                            xs, ys, grad_xs)
+    from matplotlib import pyplot as plt
+
+    initialize()
+
+    plt.title('Auto Diff')
+    ax = plt.gca()
+    ax.plot(xs, ys, label='f(x)')
+    ax.plot(xs, grad_xs, label='f\'(x)')
+    ax.legend()
+    ax.grid(True)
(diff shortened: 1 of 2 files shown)


tests/python/examples/autodiff/test_simple_derivative.py
```diff
@@ -0,0 +1,44 @@
...
+    except OSError as exc: # Python >2.5
+        if exc.errno == EEXIST and path.isdir(dir_path):
+            pass
+        else: raise
+
+
+def pic_simple_derivative(result_dir):
+    from taichi.examples.autodiff.simple_derivative import (initialize,
+                                                            xs, ys, grad_xs)
+    from matplotlib import pyplot as plt
```
ailzhang, line 20: It's likely that we don't have matplotlib in CI test environment yet. So please feel free to add it in https://github.com/taichi-dev/taichi/blob/master/requirements_test.txt if you need it!

tests/python/examples/autodiff/test_simple_derivative.py
```diff
@@ -0,0 +1,44 @@
+import argparse
+
+
+def mkdir_p(dir_path):
```
ailzhang, line 4: Yea this one should be moved to shared https://github.com/taichi-dev/taichi/blob/master/tests/test_utils.py?

ailzhang (commented):
Thanks for working on this!

tests/python/examples/autodiff/test_simple_derivative.py
strongoier: The problem here is that the original example uses matplotlib, so importing from the example will result in error. I suggest adding it to `requirements_tests.txt`.

tests/python/examples/autodiff/test_simple_derivative.py
strongoier: Yea you can fix #4322 first :-)

tests/python/examples/autodiff/test_simple_derivative.py
Tiny-Box: get it, I will add matplotlib to `requirements_test.txt` in #4322 first

strongoier:
@Tiny-Box You can now fix this PR :-)

strongoier (commented):
Did you forget to add `test_simple_derivative`?

Tiny-Box:
Oh, I forget it! I will add it today.

strongoier (approved):
LGTM! Thanks!

PR #4329, opened 2022-02-18
[gui] [refactor] Avoid exposing different APIs with different GGUI_AVAILABLE values

strongoier:
Related issue = #3782

This PR replaces inheritance with composition to avoid inheriting from a class only available when GGUI is available. Now APIs  exposed from the `ui` subpackage will not depend on whether `GGUI_AVAILABLE=True`. Meanwhile, the error checking for using GGUI when `GGUI_AVAILABLE=False` remains the same.

### DIFF
--- a/python/taichi/ui/__init__.py
+++ b/python/taichi/ui/__init__.py
@@ -1,3 +1,2 @@
 from .gui import *
-from .imgui import *
 from .ui import *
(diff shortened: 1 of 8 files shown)


AmesingFlank (approved):
thanks!

PR #4330, opened 2022-02-18
[Vulkan] Enable Vulkan device selection when using cuda

qiao-bo:
Changes made to allow the use of `TI_VISIBLE_DEVICE` to select Vulkan device. Users can use the env var to pick the desired device. 

```
[vulkan_device_creator.cpp:pick_physical_device@340] Found Vulkan Device 0 (Intel(R) Graphics (RKL GT1))
[vulkan_device_creator.cpp:pick_physical_device@340] Found Vulkan Device 1 (NVIDIA GeForce RTX 3080)
[vulkan_device_creator.cpp:pick_physical_device@340] Found Vulkan Device 2 (llvmpipe (LLVM 12.0.0, 256 bits))
```
use export `TI_VISIBLE_DEVICE` to select 0, 1, or 2

### DIFF
--- a/taichi/backends/vulkan/vulkan_device_creator.cpp
+++ b/taichi/backends/vulkan/vulkan_device_creator.cpp
@@ -168,20 +168,6 @@ bool is_device_suitable(VkPhysicalDevice device, VkSurfaceKHR surface) {
   }
 }
 
-bool is_device_visible(VkPhysicalDevice device) {
-  auto device_id = VulkanLoader::instance().visible_device_id;
-  if (device_id.empty()) {
-    // user did not specify, all devices visible
-    return true;
-  }
-
-  VkPhysicalDeviceProperties device_properties{};
-  vkGetPhysicalDeviceProperties(device, &device_properties);
-  std::stringstream id_sstream;
-  id_sstream << std::hex << device_properties.deviceID;
-  return id_sstream.str() == device_id;
-}
-
 }  // namespace
 
 VulkanDeviceCreator::VulkanDeviceCreator(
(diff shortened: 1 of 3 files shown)


bobcao3:
Oh crap, sorry I missed it in the previous commit. DeviceID is the hardware identifier of the device type. What we should use is not device ID but the Index where it appeared in the Vulkan device enumeration. A common setup in research or servers is where you have multiple of the same GPUs installed, in those cases all of them have the same ID (but they always have different index)

qiao-bo:
> Oh crap, sorry I missed it in the previous PR. DeviceID is the hardware identifier of the device type. What we should use is not device ID but the Index where it appeared in the Vulkan device enumeration. (made the mistake myself when testing) A common setup in research or servers is where you have multiple of the same GPUs installed, in those cases all of them have the same ID (but they always have different index)

What you said is probably the deviceType. As far as i see in [VkPhysicalDeviceProperties](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceProperties.html), deviceID is 'a unique identifier for the physical device among devices available from the vendor'. I assume different GPUs have the same vendorID but different deviceID.

For other readers this: the reason we don't use 0, 1, etc. as the identifier is that the output of your `vulkaninfo` may have a different order than the `vkEnumeratePhysicalDevices` we use in the code.

bobcao3:
> > Oh crap, sorry I missed it in the previous PR. DeviceID is the hardware identifier of the device type. What we should use is not device ID but the Index where it appeared in the Vulkan device enumeration. (made the mistake myself when testing) A common setup in research or servers is where you have multiple of the same GPUs installed, in those cases all of them have the same ID (but they always have different index)
> 
> What you said is probably the deviceType. As far as i see in [VkPhysicalDeviceProperties](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceProperties.html), deviceID is 'a unique identifier for the physical device among devices available from the vendor'. I assume different GPUs have the same vendorID but different deviceID.
> 
> Moreover, the reason we don't use 0, 1, etc. as the identifier is that the output of your `vulkaninfo` may have a different order than the `vkEnumeratePhysicalDevices` we use in the code.

Nope, it is unique "among devices available from the vendor". i.e. Nvidia makes RTX3080 and RTX2060, they have different device IDs. Two 3080s have the same device ID. Vendor ID identifies a vendor, e.g. Nvidia or AMD

taichi/backends/vulkan/vulkan_device_creator.cpp
```diff
@@ -346,10 +334,33 @@ void VulkanDeviceCreator::pick_physical_device() {
   vkEnumeratePhysicalDevices(instance_, &device_count, devices.data());
   physical_device_ = VK_NULL_HANDLE;
 
-  for (const auto &device : devices) {
-    if (is_device_suitable(device, surface_) && (is_device_visible(device))) {
-      physical_device_ = device;
-      break;
+  for (int i = 0; i < device_count; i++) {
```
qiao-bo, line 337: we have to give the info here, because `vkEnumeratePhysicalDevices` does NOT always print the available devices in the same order. Hence we have to print the info at the point where the device is actually selected.

bobcao3 (approved):
LGTM

PR #4347, opened 2022-02-21
[ci] Fix generate_example_videos.py

ailzhang:
`generate_example_videos.py` doesn't throw when the subprocess fails. This PR fixes it.

Related issue = #


strongoier (approved):
LGTM!

PR #4348, opened 2022-02-21
[ci] Increase ci test parallelism

qiao-bo:
Split the GPU and Windows tests by per arch to increase parallelism and reduce the total time used. This can reduce our CI from ~40 mins to ~30 mins.

### DIFF
--- a/python/taichi/lang/util.py
+++ b/python/taichi/lang/util.py
@@ -11,7 +11,7 @@ from taichi.types.primitive_types import (f16, f32, f64, i8, i16, i32, i64, u8,
 
 _has_pytorch = False
 
-_env_torch = os.environ.get('TI_ENABLE_TORCH', '0')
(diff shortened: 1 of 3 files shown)


python/taichi/lang/util.py
```diff
@@ -11,7 +11,7 @@
 
 _has_pytorch = False
 
-_env_torch = os.environ.get('TI_ENABLE_TORCH', '0')
+_env_torch = os.environ.get('TI_ENABLE_TORCH', '1')
 if not _env_torch or int(_env_torch):
```
ailzhang, line 15: do we still need this env?

ailzhang (approved):
LGTM!

python/taichi/lang/util.py
qiao-bo: @ailzhang, yes, we have like less than 20 tests for torch io etc. This env provides possibilities to enable or disable these tests. I think it's better to leave it here.

PR #4349, opened 2022-02-21
[lang] Add support for operators "is" and "is not" in static scope and deprecate them

### LINKED ISSUE #4341: Add a deprecate warning for `is` support
An example usage can be found in https://github.com/taichi-dev/taichi_elements/pull/92/files, maybe it's safer to deprecate it first before we remove it?

lin-hitonami:
Related issue = close #4341


mzmzm (approved):
LGTM!

strongoier (approved):
LGTM!

PR #4351, opened 2022-02-21
[ci] Use conda python for m1 jobs

ailzhang:
There seems to be a discrepancy between testing.yml and release.yml.
For example, my opengl PR #4345 fails at vulkan backend on m1. Also today's
nightly failed. So unifying them here to debug...

Related issue = #


qiao-bo (approved):
LGTM!

PR #4352, opened 2022-02-21
[example] Add implicit fem example

BillXu2000:
An implicit fem example, and also an example to show the usage of dynamic indexing.
Notes for reviewer:
1. Some kernels might be too magic or not fully compatible with Taichi syntax concept.
2. `_ti_core.GGUI_AVAILABLE` is necessary for ggui environment detection, but I am not sure whether it is safe to use it in example for long.

### DIFF
--- a/python/taichi/examples/simulation/implicit_fem.py
+++ b/python/taichi/examples/simulation/implicit_fem.py
@@ -7,11 +7,13 @@ import taichi as ti
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--exp', default='implicit')
-#parser.add_argument('--exp', default='explicit')
+# edit args.exp to switch between implicit and explicit intergration:
+# parser.add_argument('--exp', default='explicit')
 parser.add_argument('--dim', type=int, default=3)
 parser.add_argument('--gui', default='auto')
-#parser.add_argument('--gui', default='ggui')
-#parser.add_argument('--gui', default='cpu')
+# edit args.gui to switch the gui type:
+# parser.add_argument('--gui', default='ggui')
+# parser.add_argument('--gui', default='cpu')
 args = parser.parse_args()
 
 ti.init(arch=ti.gpu, dynamic_index=True)
@@ -25,7 +27,7 @@ if args.gui == 'auto':
 E, nu = 5e4, 0.0
 mu, la = E / (2 * (1 + nu)), E * nu / ((1 + nu) * (1 - 2 * nu))  # lambda = 0
 density = 1000.0
-dt = 2e-5
+dt = 2e-4
 
 if args.exp == 'implicit':
     dt = 1e-2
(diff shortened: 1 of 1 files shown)


python/taichi/examples/simulation/implicit_fem.py
```diff
@@ -0,0 +1,353 @@
...
+        for i in range(10):
+            get_force()
+            advect()
+    else:
+        for i in range(1):
+            cg()
+    floor_bound()
+
+
+get_vertices()
```
k-ye, line 285: nit: could you `if __name__ == '__main__':` ?

python/taichi/examples/simulation/implicit_fem.py
```diff
@@ -0,0 +1,353 @@
...
+                                 movement_speed=0.03,
+                                 hold_key=ti.ui.RMB)
+        scene.set_camera(camera)
+
+        scene.ambient_light((0.1, ) * 3)
+
+        scene.point_light(pos=(0.5, 10.0, 0.5), color=(0.5, 0.5, 0.5))
+        scene.point_light(pos=(10.0, 10.0, 10.0), color=(0.5, 0.5, 0.5))
+
+        #scene.particles(x, radius=0.02)
```
strongoier, line 312: nit: we'd better not leave commented code

python/taichi/examples/simulation/implicit_fem.py
```diff
@@ -0,0 +1,353 @@
+import argparse
+
+import numpy as np
+from taichi._lib import core as _ti_core
+
+import taichi as ti
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--exp', default='implicit')
+#parser.add_argument('--exp', default='explicit')
```
strongoier, line 10: ditto

strongoier (approved):
Looks great! Thanks!

BTW using `_ti_core.GGUI_AVAILABLE` is OK for now. We will design & add new APIs for detecting such things in the future.

python/taichi/examples/simulation/implicit_fem.py
BillXu2000: User can comment line 9 and uncomment line 10 to switch to the explicit version, so I think this line can be kept here.
Same for line 13 and 14.

python/taichi/examples/simulation/implicit_fem.py
BillXu2000: I will fix this.

python/taichi/examples/simulation/implicit_fem.py
BillXu2000: I will fix this.

python/taichi/examples/simulation/implicit_fem.py
k-ye: FYI you can use `choices`: https://docs.python.org/3/library/argparse.html#choices

BillXu2000: I will use `choices`.

PR #4356, opened 2022-02-22
[llvm] Remove LLVM functions related to a SNode tree from the module when the SNode tree is destroyed

lin-hitonami:
Related issue = #4023


ifsheldon:
But we will still have performance downgrade when the number of allocated and not destroyed SNodeTree is large. Try this.
```python
import taichi as ti
from tqdm import tqdm

if __name__ == "__main__":
    ti.init(ti.cpu)
    iterations = 2
    snode_tree_num = 1000
    for i in range(iterations):
        snode_trees = []
        for j in tqdm(range(snode_tree_num)):
            x = ti.field(ti.f32)
            fb = ti.FieldsBuilder()
            fb.place(x)
            snodetree = fb.finalize()
            snode_trees.append(snodetree)
        for snodetree in snode_trees:
            snodetree.destroy()
```

taichi/struct/struct_llvm.cpp
```diff
@@ -167,9 +171,11 @@ void StructCompilerLLVM::generate_types(SNode &snode) {
 
   // Create a dummy function in the module with the type stub as return type
   // so that the type is referenced in the module
+  std::string func_name = type_stub_name(&snode) + "_func";
```
ailzhang, line 174: can we wrap these lines into a function so that it's easier to update in the future?

ailzhang (approved):
LGTM!

tests/cpp/codegen/refine_coordinates_test.cpp
```diff
@@ -118,7 +118,7 @@ class RefineCoordinatesTest : public ::testing::Test {
     leaf_snode.dt = PrimitiveType::f32;
 
     auto sc = std::make_unique<StructCompilerLLVM>(
-        arch_, &config_, tlctx_, tlctx_->clone_runtime_module());
+        arch_, &config_, tlctx_, tlctx_->clone_runtime_module(), 0);
```
k-ye, line 121:
```suggestion
        arch_, &config_, tlctx_, tlctx_->clone_runtime_module(), /*snode_tree_id=*/0);
```

k-ye (approved):
LGTM!

PR #4361, opened 2022-02-22
[test] Add test for exposed top-level APIs

strongoier:
Related issue = #3782

Since we have put great efforts on picking user APIs, we need a systematic way to prevent future developers from accidentally breaking the API list. This PR aims at solving this problem by testing exposed top-level APIs (including sub-APIs for top-level classes).


k-ye (commented):
Hmm, I do wonder if this is the best way to control API. Have you found any reference for other Python projects doing similar implementation? I'm a bit concerned about the maintanence cost...

strongoier:
> Hmm, I do wonder if this is the best way to control API. Have you found any reference for other Python projects doing similar implementation? I'm a bit concerned about the maintanence cost...

Most Python projects don't control API. Super projects like numpy and torch put much more efforts on the doc site - manually selecting APIs to write and show, and not responsible for anything not on the doc site. We are currently taking an fully automatic approach because we currently don't have so many people to maintain the doc site.

k-ye (approved):
LGTM, thanks!

tests/python/test_api.py
```diff
@@ -0,0 +1,89 @@
...
+    'copy_from', 'element_shape', 'fill', 'from_numpy', 'to_numpy'
+]
+
+
+@pytest.mark.parametrize('src', user_api.keys())
+@test_utils.test(arch=ti.cpu)
+def test_api(src):
+    # When Python version is below 3.7, deprecated names are
+    # handled as normal names, which will fail this test.
+    assert sys.version_info < (3, 7) or [
```
ailzhang, line 87: We might need a better error message to instruct developers how to tell if their apis should be added or just hidden, and how to update the lists if they really want to expose ;) 
But this can be a followup for sure!

ailzhang (approved):
Awesome thanks!

PR #4367, opened 2022-02-23
[ci] Run vulkan and metal separately on M1

ailzhang:
Related issue = #
This seem to help fix the random failure in M1 CI. My guess was that vulkan and metal backend using M1 chip simultaneously might cause some weird behavior....


qiao-bo (approved):
LGTM!

PR #4375, opened 2022-02-23
[mesh] Constructing mesh from data in memory

BillXu2000:
Related issue = #3608

Make it possible to construct mesh directly from data in memory.


g1n0st (approved):
LGTM!

PR #4382, opened 2022-02-24
[aot] [vulkan] Output shapes/dims to AOT exported module

ghuau-innopeak:
Related issue = #3642

This is needed for NdArray support as well as exporting the DeviceAllocation symbol to external libraries.

### DIFF
--- a/taichi/backends/vulkan/aot_module_builder_impl.cpp
+++ b/taichi/backends/vulkan/aot_module_builder_impl.cpp
@@ -44,7 +44,19 @@ class AotDataConverter {
     res.args_buffer_size = in.ctx_attribs.args_bytes();
     res.rets_buffer_size = in.ctx_attribs.rets_bytes();
     for (const auto &arg : in.ctx_attribs.args()) {
-      res.scalar_args[arg.index] = visit(arg);
+      if (!arg.is_array) {
+        aot::ScalarArg scalar_arg{};
+        scalar_arg.dtype_name = arg.dt.to_string();
+        scalar_arg.offset_in_args_buf = arg.offset_in_mem;
+        res.scalar_args[arg.index] = scalar_arg;
+      } else {
+        aot::ArrayArg arr_arg{};
+        arr_arg.dtype_name = arg.dt.to_string();
+        arr_arg.field_dim = arg.field_dim;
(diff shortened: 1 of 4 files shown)


taichi/backends/vulkan/aot_module_builder_impl.cpp
```diff
@@ -44,7 +44,19 @@ class AotDataConverter {
...
+        aot::ScalarArg scalar_arg{};
+        scalar_arg.dtype_name = arg.dt.to_string();
+        scalar_arg.offset_in_args_buf = arg.offset_in_mem;
+        res.scalar_args[arg.index] = scalar_arg;
+      } else {
+        aot::ArrayArg arr_arg{};
+        arr_arg.dtype_name = arg.dt.to_string();
+        arr_arg.field_dim = arg.field_dim;
+        arr_arg.element_shape = arg.element_shape;
+        arr_arg.shape_offset_in_args_buf = arg.index * sizeof(int);
```
k-ye, line 57: @ailzhang @bobcao3 Do you remember whether this is 32-bit or 64-bit? Either way, we should explicitly specify its bits using `int32_t` or `int64_t`.

k-ye (approved):
One nit, o.w. LGTM!

taichi/backends/vulkan/aot_module_builder_impl.cpp
ailzhang: This is 32 bit IIRC

ailzhang (approved):
LGTM!

taichi/backends/vulkan/aot_module_builder_impl.cpp
bobcao3: If previous args contain f16 it might be different. Better to export the kernel_utils output directly?

bobcao3: If this means the "shape offset" as in the length on each axis, this is actually in extra_args. Each element in extra args is int32, while each arg corresponds to 8 extra args (8 axis)

k-ye:
```suggestion
        arr_arg.shape_offset_in_args_buf = arg.index * sizeof(int32_t);
```

PR #4387, opened 2022-02-25
[aot] [vulkan] Add AotKernel and its Vulkan impl

k-ye:

### DIFF
--- a/taichi/backends/vulkan/aot_module_loader_impl.cpp
+++ b/taichi/backends/vulkan/aot_module_loader_impl.cpp
@@ -3,9 +3,30 @@
 #include <fstream>
 #include <type_traits>
 
+#include "taichi/backends/vulkan/runtime.h"
+
 namespace taichi {
 namespace lang {
 namespace vulkan {
+namespace {
+
+using KernelHandle = VkRuntime::KernelHandle;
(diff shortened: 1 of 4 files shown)


taichi/backends/vulkan/aot_module_loader_impl.cpp
```diff
@@ -3,9 +3,30 @@
...
+      : runtime_(runtime), handle_(handle) {
+  }
+
+  void run(RuntimeContext *ctx) override {
+    runtime_->launch_kernel(handle_, ctx);
+  }
+
+ private:
+  VkRuntime *const runtime_;
+  const KernelHandle handle_;
```
bobcao3, line 27: Out of curiosity, I wonder maybe we can ditch the KernelHandle thing and use some pure data structure to replace it?

bobcao3 (approved):
LGTM!

taichi/backends/vulkan/aot_module_loader_impl.cpp
k-ye: I do prefer `KernelHandle` for type safety. It's identical to an integer, so no runtime overhead :-)

PR #4394, opened 2022-02-25
[ci] Reduce test parallelism for m1

qiao-bo:
Lets reduce the test parallelism for m1 to get rid of the flaky test issue. This won't affect our CI build time as m1 is much faster.

### DIFF
--- a/.github/workflows/scripts/unix_test.sh
+++ b/.github/workflows/scripts/unix_test.sh
@@ -42,17 +42,29 @@ else
         python3 tests/run_tests.py -vr2 -t4 -k "not torch" -a cuda
     fi
     if [[ $TI_WANTED_ARCHS == *"cpu"* ]]; then
-        python3 tests/run_tests.py -vr2 -t8 -k "not torch" -a cpu
+        if [[ $ARCH == *"m1"* ]]; then
+            python3 tests/run_tests.py -vr2 -t4 -k "not torch" -a cpu
+	else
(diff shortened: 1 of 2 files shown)


.github/workflows/testing.yml
```diff
@@ -465,3 +465,4 @@ jobs:
           TI_WANTED_ARCHS: "metal,vulkan,cpu"
           PY: ${{ matrix.python }}
           GPU_TEST: ON
+          ARCH: "m1"
```
ailzhang, line 468: let's remove `GPU_TEST` here and only use `PLATFORM` maybe?

.github/workflows/scripts/unix_test.sh
```diff
@@ -50,7 +59,6 @@ else
     if [[ $TI_WANTED_ARCHS == *"opengl"* ]]; then
         python3 tests/run_tests.py -vr2 -t4 -k "not torch" -a opengl
     fi
-    # Run metal and vulkan separately so that they don't use M1 chip simultaneously.
     if [[ $TI_WANTED_ARCHS == *"metal"* ]]; then
```
ailzhang, line 62: This is no longer needed ;) We won't test metal backend on normal self-hosted gpu runners.

ailzhang (approved):
LGTM! Only had a nit before merging :D

.github/workflows/scripts/unix_test.sh
qiao-bo: right ;/

Issue #4395, opened 2022-02-25
Add a manylinux2014 compatible release with only LLVM-based backends


k-ye:
This will include both CPU and CUDA backends, which could be useful for cloud users. See https://github.com/taichi-dev/taichi/issues/4377, https://github.com/taichi-dev/taichi/issues/3332

bobcao3:
If we directly build on CentOS 7 nvidia docker, can we get a manylinux2014 wheel with proper cuda / vulkan?

qiao-bo:
@k-ye @bobcao3, Let me recap some of our previous efforts here: The problem with CentOS 7 is that it comes with GCC 4.8, this is too old to compile Taichi (not saying impossible, but too many errors to dig). This is the same story for Nvidia CentOS7. The way we circumvent this is to use a magic docker image with Red Hat devtoolset-10. (https://git.centos.org/rpms/devtoolset-10-gcc), which is based on CentOS7 but with gcc 10. By using this image (quay.io/pypa/manylinux2014_x86_64), we still need to recompile clang and llvm but it is possible to build. 

Now to support CUDA, I see three options:
[a] Use Nvidia CentOS 7 image, but fix gcc 4.8 problem. Difficulty is high.
[b] Use the CentOS 7 based image with gcc 10. But install and build Nvidia/CUDA tools. Should be possible.
[c] Still use our Ubuntu 18.04 image, but pin glibc version to 2.2.5 (we already knew how to do it). But ship Taichi without Vulkan. Difficulty is low.

I vote for option [c], WDYT?

k-ye:
[c] sounds good to me!

bobcao3:
c sounds like the best option but I do think we should attempt `a` or `b` so that at least we can confirm vulkan works on the manylinux2014 target. (Which is a lot higher than 2.2.5)

k-ye:
I think our top priority *in this issue* is to unblock CUDA users.. Vulkan will be a nice-to-have, but it's not that urgent :-)

strongoier:
Where can we upload such a wheel without Vulkan?

k-ye:
> Where can we upload such a wheel without Vulkan?

I believe GH provides API to upload additional artifacts to release pages like this (https://github.com/taichi-dev/taichi/releases/tag/v0.9.0)

strongoier:
Yes that's definitely possible. However, the updated version is no longer Taichi v9.0. Are we going to do so only for future releases?

[...]

PR #4406, opened 2022-02-28
[build] Enforce compatibility with manylinux2014 when TI_WITH_VULKAN=OFF

strongoier:
Related issue = #4395

This corresponds to option [c] of #4395.


qiao-bo (approved):
LGTM!

PR #4413, opened 2022-03-01
[Doc] Update docstring for functions in operations

neozhaoliang:


neozhaoliang:
@ailzhang Hi, would you please help review this PR?

python/taichi/lang/ops.py
```diff
@@ -181,68 +206,117 @@ def _ternary_operation(taichi_op, python_op, a, b, c):
...
     Returns:
-        The inverses function of sine of `a`.
+        The inverse sine of each element in `x`, in radians and in the closed \
+            interval `[-pi/2, pi/2]`.
+
+    Example::
+
+        >>> from math import pi
+        >>> ti.asin(ti.Matrix([-1.0, 0.0, 1.0])) * 180 / pi
+        >>> [-90., 0., 90.]
```
ailzhang, line 291: Aha I finally got what feels a bit weird here: you only need `>>>` for commands but not outputs so that users can tell the difference ;) 
Examples: https://github.com/google/jax/blob/4e219220558a2279de2abc9dc4140dedeb61703f/jax/_src/errors.py#L96
https://github.com/pytorch/pytorch/blob/master/torch/functional.py#L139

ailzhang (approved):
LGTM! Please fix `>>>` before merging thanks!

ailzhang (approved):
LGTM! Thanks!

PR #4419, opened 2022-03-01
[vulkan] [aot] Add aot namespace Vulkan

qiao-bo:
Related issue = #3642 

Resolve the following TODO: "Instead of prefixing all classes with "Aot", just put them into the `aot` namespace."

Next step: AotModuleLoader --> aot::Module::load()


ailzhang (approved):
LGTM!

Issue #4257, opened 2022-02-10
Feature request: matrix slicing


xuhao1:
Hi there, my program needs to create an array to store some data, it looks like
```python
#Marching...
...
    vertlist = [[0, 0, 0] for i in range(12)]
    for _i in ti.static(range(12)):
        if edges & 2**_i:
            _p0 = ti.static(edges_grid_xyz)[_i][0]
            _p1 = ti.static(edges_grid_xyz)[_i][1]
            p0 = [i + _p0[0]*step, j + _p0[1]*step, k + _p0[2]*step]
            p1 = [i + _p1[0]*step, j + _p1[1]*step, k + _p1[2]*step]
            p = self.vertexInterp(p0, p1, tsdf[p0], tsdf[p1], isolevel)
            vertlist[_i][0] = p[0]
            vertlist[_i][1] = p[1]
            vertlist[_i][2] = p[2]
    
    _i = 0
    while triTable[cubeindex, _i] != -1:
        self.add_triangle(vertlist[triTable[cubeindex, _i]], 
            vertlist[triTable[cubeindex, _i + 1]], vertlist[triTable[cubeindex, _i +2]], triangle_idx)
        triangle_idx += 1
        _i += 3
```

If a use a normal list here for vertlist, it will raise:
```
TypeError: list indices must be integers or slices, not Expr
```

If I change vertlist to
```
                    vertlist = ti.Matrix([[0, 0, 0] for i in range(12)])
```
It will raise

```
                        self.add_triangle(vertlist[triTable[cubeindex, _i], 0],
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
The 0-th index of a Matrix/Vector must be a compile-time constant integer, got <class 'taichi.lang.expr.Expr'>.
This is because matrix operations will be **unrolled** at compile-time for performance reason.
If you want to *iterate through matrix elements*, use a static range:
  for i in ti.static(range(3)):
    print(i, "-th component is", vec[i])
See https://docs.taichi.graphics/lang/articles/advanced/meta#when-to-use-for-loops-with-tistatic for more details.

```
I hope to parallelize this piece of code so the vertlist should not be a global accessible field, so how should I make this code works?

strongoier:
[...]

PR #4420, opened 2022-03-01
[Lang] Support simple matrix slicing

mzmzm:
Related issue = #4257
Simple implementation of matrix slicing, that only supports the positive constant integer type of slice's elements.


python/taichi/lang/impl.py
```diff
@@ -126,23 +126,34 @@ def wrap_scalar(x):
 @taichi_scope
 def subscript(value, *_indices, skip_reordered=False):
     if isinstance(value, np.ndarray):
-        return value.__getitem__(*_indices)
+        return value.__getitem__(_indices)
```
strongoier, line 129: Why this change?

tests/python/test_simple_matrix_slice.py
```diff
@@ -0,0 +1,24 @@
...
+        c = ti.Vector([0, 1, 2, 3, 4, 5, 6])
+        return c[:5:2]
+
+    @ti.kernel
+    def foo2() -> ti.types.matrix(2, 2, dtype=ti.i32):
+        a = ti.Matrix([[1, 2, 3], [4, 5, 6]])
+        return a[:, :b:2]
+
+    v1 = foo1()
+    assert v1[0] == 0
```
strongoier, line 20: Shall we verify the full matrix/vector instead of a single value?

strongoier (commented):
Nice work!

python/taichi/lang/impl.py
mzmzm: `ndarray.__getitem__` only supports an index or a tuple of indices, the previous one has a bug then deal with multiple indexes

python/taichi/lang/impl.py
```diff
@@ -126,23 +126,34 @@ def wrap_scalar(x):
...
     _indices = tuple(flattened_indices)
     if isinstance(_indices,
                   tuple) and len(_indices) == 1 and _indices[0] is None:
         _indices = ()
+
+    if has_slice:
+        if not isinstance(value, Matrix):
+            raise SyntaxError(
+                f"The type {type(value)} do not support index of slice type")
+        return value._subscript(*_indices)
```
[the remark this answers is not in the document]
mzmzm: fix

tests/python/test_simple_matrix_slice.py
mzmzm: fix

strongoier (approved):
LGTM!

PR #4422, opened 2022-03-01
[llvm] Support real function which has scalar arguments

lin-hitonami:
Related issue = #602

New TODOs:
1. Determine which passes real functions need.
2. (Maybe) let the function body be a serial offloaded statement to enable optimization passes
3. CFG dies when there are recursions. (Recursion works when we turn off CFG)

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -226,34 +226,6 @@ def make_tensor_element_expr(_var, _indices, shape, stride):
                                           shape, stride))
 
 
-@taichi_scope
-def insert_expr_stmt_if_ti_func(ast_builder, func, *args, **kwargs):
-    """This method is used only for real functions. It inserts a
-    FrontendExprStmt to the C++ AST to hold the function call if `func` is a
-    Taichi function.
-
-    Args:
-        func: The function to be called.
-        args: The arguments of the function call.
-        kwargs: The keyword arguments of the function call.
-
-    Returns:
-        The return value of the function call if it's a non-Taichi function.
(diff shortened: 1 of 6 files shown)


python/taichi/lang/impl.py
```diff
@@ -226,34 +226,6 @@ def make_tensor_element_expr(_var, _indices, shape, stride):
                                           shape, stride))
 
 
-@taichi_scope
-def insert_expr_stmt_if_ti_func(ast_builder, func, *args, **kwargs):
```
k-ye, line 230: So, are we completely removing the experimental real function in this PR? (If so, I think the tests can be removed as well)

k-ye (approved):
LGTM!

python/taichi/lang/impl.py
lin-hitonami: No, I will remove it in another PR, this is only a buggy function that prevents using `print` in the real functions. I merged this function with ASTTransformer.build_Expr.

bobcao3:
"Let the function body be a serial offloaded statement to enable optimization passes" iirc in the analysis pass serial offloads will get all atomics demoted, that wouldn't be ideal...?

lin-hitonami:
> "Let the function body be a serial offloaded statement to enable optimization passes" iirc in the analysis pass serial offloads will get all atomics demoted, that wouldn't be ideal...?

So maybe we shouldn't run demote_atomics pass on real functions. Let me check if there are other passes that should run on real functions that require offloaded statements.

ailzhang (approved):
LGTM!!

PR #4437, opened 2022-03-03
[aot] [refactor] Refactor AOT runtime API to use module

qiao-bo:
Towards a simpler AOT runtime API as suggested in issue #3642.

This PR makes the following changes 
- [x] `ModuleLoader` to `Module.load()`
- [x] `module->get_kernel()`
- [x] `kernel->launch()`

### DIFF
--- a/taichi/aot/module_loader.cpp
+++ b/taichi/aot/module_loader.cpp
@@ -11,10 +11,11 @@ std::unique_ptr<Module> Module::load(const std::string &path,
                                      Arch arch,
                                      std::any mod_params) {
   if (arch == Arch::vulkan) {
-    // TODO: unify the types here between vk with metal
-    std::string vk_params = std::any_cast<std::string &>(mod_params);
-    return std::make_unique<vulkan::AotModuleImpl>(vk_params);
+    vulkan::AotModuleParams vulkan_params =
+        std::any_cast<vulkan::AotModuleParams &>(vulkan_params);
+    return vulkan::make_aot_module(vulkan_params);
   } else if (arch == Arch::metal) {
+
     metal::AotModuleParams metal_params =
         std::any_cast<metal::AotModuleParams &>(mod_params);
     return metal::make_aot_module(metal_params);
--- a/taichi/aot/module_loader.cpp
+++ b/taichi/aot/module_loader.cpp
@@ -7,7 +7,7 @@ namespace aot {
 std::unique_ptr<Module> Module::load(const std::string &path,
                                      Arch arch,
                                      std::any mod_params) {
-//  arch_ = arch;
+  //  arch_ = arch;
   if (arch == Arch::vulkan) {
 #ifdef TI_WITH_VULKAN
     return vulkan::make_aot_module(mod_params);
@@ -21,7 +21,7 @@ std::unique_ptr<Module> Module::load(const std::string &path,
   }
 }
 
-//Arch Module::arch() const {
+// Arch Module::arch() const {
 //  return arch_;
 //}
 
--- a/taichi/aot/module_loader.cpp
+++ b/taichi/aot/module_loader.cpp
@@ -12,7 +12,7 @@ std::unique_ptr<Module> Module::load(const std::string &path,
                                      std::any mod_params) {
   if (arch == Arch::vulkan) {
     vulkan::AotModuleParams vulkan_params =
-        std::any_cast<vulkan::AotModuleParams &>(vulkan_params);
+        std::any_cast<vulkan::AotModuleParams &>(mod_params);
     return vulkan::make_aot_module(vulkan_params);
   } else if (arch == Arch::metal) {
 
--- a/taichi/aot/module_loader.cpp
+++ b/taichi/aot/module_loader.cpp
@@ -7,7 +7,6 @@ namespace aot {
 std::unique_ptr<Module> Module::load(const std::string &path,
                                      Arch arch,
                                      std::any mod_params) {
-  //  arch_ = arch;
   if (arch == Arch::vulkan) {
 #ifdef TI_WITH_VULKAN
     return vulkan::make_aot_module(mod_params);
@@ -20,10 +19,6 @@ std::unique_ptr<Module> Module::load(const std::string &path,
     TI_NOT_IMPLEMENTED;
   }
 }
-
-// Arch Module::arch() const {
-//  return arch_;
-//}
 
 }  // namespace aot
 }  // namespace lang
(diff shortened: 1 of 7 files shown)


taichi/aot/module_loader.cpp
```diff
@@ -1,18 +1,34 @@
 #include "taichi/aot/module_loader.h"
 
+#ifdef TI_WITH_VULKAN
```
k-ye, line 3: Just hide these in each backend's impl. It's fine that `xx_backend::make_aot_module()` is **declared** in the header

taichi/aot/module_loader.h
```diff
@@ -37,34 +38,35 @@ class TI_DLL_EXPORT Kernel {
...
+                                      std::any mod_params);
+  // TODO
+  // Module metadata
+  // Arch arch() const;
+  // uint64_t version() const;
+
+  // TODO
+  // APIs to be overriden by each backend.
+  // virtual std::unique_ptr<Field> get_field(const std::string &name) = 0;
+  // virtual std::unique_ptr<Kernel> get_kernel(const std::string &name) = 0;
```
k-ye, line 62: Why not implement this?

tests/cpp/aot/aot_save_load_test.cpp
```diff
@@ -91,19 +95,63 @@ TEST(AotSaveLoad, Vulkan) {
...
 
-  auto ret = aot_loader.get_kernel("init", init_kernel);
-  EXPECT_TRUE(ret);
+  // TODO
+  // auto init_kernel = vk_module->get_kernel("init");
+  // EXPECT_TRUE(init_kernel);
 
-  ret = aot_loader.get_kernel("ret", ret_kernel);
-  EXPECT_TRUE(ret);
+  // auto substep_kernel = vk_module->get_kernel("substep");
```
k-ye, line 139: Why not fix this?

taichi/aot/module_loader.h
qiao-bo: Try to make this PR small to only contain one change, as the title says. This will be implemented to replace the old one in derived classes.

tests/cpp/aot/aot_save_load_test.cpp
qiao-bo: i can fix these all in this PR if that is better

taichi/aot/module_loader.h
k-ye: `get_kernel` already exists in the current codebase, let's just add it back.

tests/cpp/aot/aot_save_load_test.cpp
k-ye: Yep, please do.

taichi/aot/module_loader.cpp
qiao-bo: Just to confirm, you mean to add these declarations in `module_loader.h` is fine? 
```
namespace vulkan {
std::unique_ptr<aot::Module> make_aot_module(std::any mod_params);
}  // namespace vulkan 
 
namespace metal {
std::unique_ptr<aot::Module> make_aot_module(std::any mod_params);
}  // namespace metal 
```

taichi/aot/module_loader.h
qiao-bo: signatures are different. but ok, will do

taichi/aot/module_loader.cpp
qiao-bo: changed

qiao-bo:
maybe it's better to make changes on field such as `module->get_field()` and `field.copy_to()` in a follow-up PR. WDYT @k-ye ?

k-ye:
> maybe it's better to make changes on field such as module->get_field() and field.copy_to() in a follow-up PR. WDYT @k-ye ?

Sure, that makes sense because we don't have Field yet

taichi/aot/module_loader.cpp
```diff
@@ -4,15 +4,20 @@ namespace taichi {
...
 
-Kernel *ModuleLoader::get_kernel(const std::string &name) {
-  auto itr = loaded_kernels_.find(name);
-  if (itr != loaded_kernels_.end()) {
-    return itr->second.get();
+std::unique_ptr<Module> Module::load(const std::string &path,
+                                     Arch arch,
+                                     std::any mod_params) {
+  if (arch == Arch::vulkan) {
+#ifdef TI_WITH_VULKAN
```
k-ye, line 11: Again, you don't have to macro-protect here. `vulkan::make_aot_module` itself can use a `TI_WITH_VULKAN` check

taichi/aot/module_loader.h
```diff
@@ -17,6 +18,17 @@ class RuntimeContext;
 
 namespace aot {
 
+class TI_DLL_EXPORT Field {
```
k-ye, line 21: Thanks for adding this. In the future, please think about these:

* Since Kernel already exists, the refactoring shouldn't remove it.
* Since Field doesn't exist, we don't have to add it during the refactoring.

Let's just keep this one.

taichi/aot/module_loader.h
```diff
@@ -116,5 +127,14 @@ class TargetDevice : public Device {
 };
 
 }  // namespace aot
+
+namespace vulkan {
```
k-ye, line 131: You should place these into their corresponding folders

taichi/backends/metal/aot_module_loader_impl.cpp
```diff
@@ -36,16 +36,26 @@ class AotModuleLoaderImpl : public aot::ModuleLoader {
...
+  std::unique_ptr<aot::Field> get_field(const std::string &name) override {
+    TI_NOT_IMPLEMENTED;
   }
 
   size_t get_root_size() const override {
     return aot_data_.metadata.root_buffer_size;
   }
 
+  // Module metadata
+  Arch arch() const {
```
k-ye, line 52: override

taichi/backends/vulkan/aot_module_loader_impl.cpp
```diff
@@ -26,76 +26,103 @@ class KernelImpl : public aot::Kernel {
...
+  std::unique_ptr<aot::Field> get_field(const std::string &name) override {
+    TI_NOT_IMPLEMENTED;
+  }
+
+  size_t get_root_size() const override {
+    return ti_aot_data_.root_buffer_size;
+  }
+
+  // Module metadata
+  Arch arch() const {
```
k-ye, line 64: ditto

taichi/aot/module_loader.cpp
k-ye: https://github.com/taichi-dev/taichi/issues/4464

k-ye (approved):
LGTM, thanks! (Remember to add `override`)

Issue #3948, opened 2022-01-05
[IR]Lower performance when using vector in the class of Block in IR


mzmzm:
**Describe the bug**
As we can see in 
https://github.com/taichi-dev/taichi/blob/910a4e12c2727737b35088c2dff0081b436d8fd5/taichi/ir/ir.h#L606

The `std::vector` will cause lower performance when we have too much `insert()` and `remove()` in the process of running passes. Usually, we should replace `vector` with `list`. But there is a lot of work when refactoring. Shall we do it?

k-ye:
What if we 1) either preallocate a reasonably large vector capacity, or 2) use `llvm::SmallVector`?

mzmzm:
I mean, when we deal with a very large project which has too many statements in one block, the cost of `insert()` and `remove()` can't be ignored. For example( #1785 ), `newton_ipc_whole.py ` there will be a lot of statements which are in one block after `ast_lower` , `offload`, `ast_access` passes. The process of compilation is time-consuming.

PR #4438, opened 2022-03-03
[ir] Trying to speedup codegen by using list instead of vectors

bobcao3:
Related issue = #3948

Still totally broken, need fix, mark as draft


bobcao3:
Temporary conclusion: insert and erase are not significant enough in our codebase (either mostly inserting at the back or just not the main performance bottleneck)

PR #4439, opened 2022-03-03
[metal] Support device memory allocation/deallocation

k-ye:
Related issue = https://github.com/taichi-dev/taichi/issues/2736

### DIFF
--- a/taichi/backends/metal/device.cpp
+++ b/taichi/backends/metal/device.cpp
@@ -230,7 +230,8 @@ class DeviceImpl : public Device {
     res.device = this;
     res.alloc_id = allocations_.size();
 
-    auto &ialloc = allocations_[res.alloc_id];  // "i" for internal
+    AllocationInternal &ialloc =
+        allocations_[res.alloc_id];  // "i" for internal
     auto mem = std::make_unique<BufferMemoryView>(params.size, mem_pool_);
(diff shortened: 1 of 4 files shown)


qiao-bo (approved):
LGTM

taichi/backends/metal/device.cpp
```diff
@@ -224,11 +226,19 @@ class DeviceImpl : public Device {
   }
 
   DeviceAllocation allocate_memory(const AllocParams &params) override {
-    TI_NOT_IMPLEMENTED;
-    return DeviceAllocation{};
+    DeviceAllocation res;
+    res.device = this;
+    res.alloc_id = allocations_.size();
+
+    auto &ialloc = allocations_[res.alloc_id];  // "i" for internal
```
ailzhang, line 233: super nit: I personally prefer to explicitly spell out the type when it's **short** to help code readability :P But I'm definitely aware that people have different opinions. Maybe good to align on this internally in the future? ;)

ailzhang (approved):
LGTM!

taichi/backends/metal/device.cpp
k-ye: +1 on explicitly writing out the type. Maybe https://google.github.io/styleguide/cppguide.html#Type_deduction ?

Issue #4230, opened 2022-02-08
Please consider adding literals support


igor-elovikov:
**Concisely describe the proposed feature**
Let's have a look on an implementation of popular pcg hash in taichi. Roughly it looks like this:
```c
uint pcg_hash(uint input)
{
    uint state = input * 747796405u + 2891336453u;
    uint word = ((state >> ((state >> 28u) + 4u)) ^ state) * 277803737u;
    return (word >> 22u) ^ word;
}
```

In taichi I should do something like this:
```python
@ti.func
def pcg_hash(inp: ti.u32) -> ti.u32:
    n1: ti.u32 = 747796405
    n2: ti.u32 = 2891336453
    n3: ti.u32 = 277803737

    b28: ti.u32 = 28
    b22: ti.u32 = 22
    b4: ti.u32 = 4

    state: ti.u32 = inp * n1 + n2
    word: ti.u32 = ((state >> ((state >> b28) + b4)) ^ state) * n3

    return (word >> b22) ^ word
```

There are other issues rather than verbosity (mentioned separately here: https://github.com/taichi-dev/taichi/issues/4210, apparently I can't even declare typed const expressions), but here I'd like to point to using literals for convenience.

**Describe the solution you'd like (if any)**

Not sure what is the best way to do that.
I can think of at least two methods.

1. Making dtype callable at least in taichi context. 

```python
@ti.func
def pcg_hash(inp: ti.u32) -> ti.u32:
    state: ti.u32 = inp * ti.u32(747796405) + ti.u32(2891336453)
    word: ti.u32 = ((state >> ((state >> ti.u32(28)) + ti.u32(4))) ^ state) * ti.u32(277803737)

    return (word >> ti.u32(22)) ^ word
```

Much better and with some additional analysis on arguments can be used as an alternative to `ti.cast`

2. Fancy indexing or exploiting `ast.Call`

```python
@ti.func
def pcg_hash(inp: ti.u32) -> ti.u32:
    state: ti.u32 = inp * 747796405[ti.u32] + 2891336453[ti.u32]
    word: ti.u32 = ((state >> ((state >> 28[ti.u32]) + 4[ti.u32])) ^ state) * 277803737[ti.u32]

    return (word >> 22[ti.u32]) ^ word
```

[...]

PR #4440, opened 2022-03-03
[Lang] Support type annotations for literals

strongoier:
Related issue = #4230

This PR implements the approach 1 in #4230. After this PR, the way to give a type to a literal is `ti.u32(12345678)` instead of
```python
a: ti.u32 = 12345678
```
Of course, you can specify both types of the literal and the variable:
```python
a: ti.u32 = ti.u32(12345678)
```

### DIFF
--- a/tests/python/test_literal.py
+++ b/tests/python/test_literal.py
@@ -9,7 +9,8 @@ def test_literal_u32():
     @ti.kernel
     def pcg_hash(inp: ti.u32) -> ti.u32:
         state: ti.u32 = inp * ti.u32(747796405) + ti.u32(2891336453)
-        word: ti.u32 = ((state >> ((state >> ti.u32(28)) + ti.u32(4))) ^ state) * ti.u32(277803737)
+        word: ti.u32 = ((state >> (
+            (state >> ti.u32(28)) + ti.u32(4))) ^ state) * ti.u32(277803737)
         return (word >> ti.u32(22)) ^ word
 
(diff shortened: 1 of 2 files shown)


tests/python/test_literal.py
```diff
@@ -0,0 +1,48 @@
...
+    with pytest.raises(ti.TaichiSyntaxError):
+        keywords_error()
+
+
+@test_utils.test()
+def test_literal_expr_error():
+    @ti.kernel
+    def expr_error():
+        a = 1
+        b = ti.f16(a)
```
lin-hitonami, line 45: Should we allow this as an alias of `ti.cast(a, ti.f16)`?

ailzhang (approved):
LGTM!

tests/python/test_literal.py
```diff
@@ -0,0 +1,48 @@
...
+
+    with pytest.raises(ti.TaichiSyntaxError):
```
ailzhang, line 47: nit: it'd nicer to include `match` keyword arg here so that it's clearer.

tests/python/test_literal.py
strongoier: I suggest not doing that at least for now. The type annotation here is more like `u` of `123u` in C.

tests/python/test_literal.py
strongoier: Will be added in the next PR.

PR #4447, opened 2022-03-04
[bug] [lang] Enable break in the outermost for not in the outermost scope

### LINKED ISSUE #4238: Behavior of break in the outermost for different from the documentation
**Describe the bug**
According to the [documentation](https://docs.taichi.graphics/#parallel-for-loops), only the for loop at the outermost scope is parallelized, and the outermost for loop not at the outermost scope is not parallelized, and parallelized for loops cannot be broken. However, according to my test, error raises when breaking the outermost for loop that is not at the outermost scope. 

**To Reproduce**
```py
import taichi as ti
ti.init()

@ti.kernel
def foo(k: ti.i32):
    if k > 42:
        for i in range(10):
            break

foo(1)
```

**Log/Screenshots**
Please post the **full log** of the program (instead of just a few lines around the error message, unless the log is > 1000 lines). This will help us diagnose what's happening. For example:

```
  File "test.py", line 10, in <module>
    foo(1)
  File "/Users/lin/taichi/python/taichi/lang/kernel_impl.py", line 736, in wrapped
  File "/Users/lin/taichi/python/taichi/lang/kernel_impl.py", line 667, in __call__
  File "/Users/lin/taichi/python/taichi/lang/kernel_impl.py", line 619, in func__
    t_kernel(launch_ctx)
RuntimeError: [ir.h:as@225] Assertion failure: is<T>()
```

### LINKED ISSUE #4324: Loop inside a if statement is identified as an outermost loop
**Describe the bug**
The loop inside a if statement is identified as an outermost loop, which is incompatible with the current doc.
image

According to the current doc, it is the loop at the outermost scope that gets parallelized, not the outermost loop.
https://docs.taichi.graphics/#parallel-for-loops

**To Reproduce**

```py
import taichi as ti
ti.init()

x = ti.field(dtype=int, shape=())
x[None] = 1

@ti.kernel
def test_break():
    if x[None] > 0:
        for i in range(5):
            x[None] += 1
            break
test_break()

```
```py
import taichi as ti
ti.init()

@ti.kernel
def test_break_2(x:int):
    if x > 0:
        for i in range(5):
            x = 1
            break
	
test_break_2(2)
```

**Log/Screenshots**
image

lin-hitonami:
Related issue = close #4324, close #4238

### DIFF
--- a/taichi/ir/frontend_ir.cpp
+++ b/taichi/ir/frontend_ir.cpp
@@ -909,19 +909,19 @@ void ASTBuilder::insert_expr_stmt(const Expr &val) {
 
 void ASTBuilder::create_scope(std::unique_ptr<Block> &list, LoopType tp) {
   TI_ASSERT(list == nullptr);
+  LoopState prev = loop_state_stack_.back();
+  if (tp == NotLoop) {
+    loop_state_stack_.push_back(prev);
+  } else if (tp == For && stack_.size() == 1) {
+    loop_state_stack_.push_back(Outermost);
+  } else {
+    loop_state_stack_.push_back(Inner);
+  }
   list = std::make_unique<Block>();
   if (!stack_.empty()) {
     list->parent_stmt = get_last_stmt();
   }
   stack_.push_back(list.get());
-  LoopState prev = loop_state_stack_.back();
-  if (tp == NotLoop) {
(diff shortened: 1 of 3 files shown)


strongoier (approved):
LGTM!

PR #4448, opened 2022-03-04
[refactor] Move literal construction to expr module

strongoier:
This is a follow-up of #4440 which makes the codebase structure more clear.


lin-hitonami (approved):
LGTM!

PR #4458, opened 2022-03-07
[lang] Add decorator ti.experimental.real_func

lin-hitonami:
Related issue = #602


python/taichi/__init__.py
```diff
@@ -9,7 +9,7 @@
 # Provide a shortcut to types since they're commonly used.
 from taichi.types.primitive_types import *
 
-from taichi import ad, linalg, tools
+from taichi import _wip, ad, linalg, tools
```
ailzhang, line 12: how about ti.experimental.real_func? :P

PR #4476, opened 2022-03-08
[test] Add arch-supporting-checking for some tests

PGZXB:
Related issue = #

Some cases always fail when I run tests with `-a cpu` on my computer because they are initialized forcibly with unsupported archs by my pc like `ti.cuda` or `ti.gpu`. In the pr, I add arch-supporting-checking for them.


strongoier:
I think the problem comes from the fact that we disable arch fallback in tests. Could you try providing an option to allow fallback instead of modifying the tests?

Issue #2398, opened 2021-06-03
One-Year Roadmap


k-ye:
Hi,

We are sharing this one-year road map to let you know what features are planned ahead or have been actively worked on. Hopefully some of these sound interesting/exciting! Let us know what you think so that we can adjust accordingly. Thanks!  

---

## New Features

* [x] Allow adding and deleting new SNodes dynamically (#2418, #2506)
    * Phase 1: LLVM
    * Phase 2: the rest
* [x] Make Taichi's JIT kernels deployable without the Python execution environment (https://github.com/taichi-dev/taichi/issues/2355)
   * https://github.com/taichi-dev/taichi/pull/3908
* [ ] Provide a way for Taichi to pass data with other frameworks (https://github.com/taichi-dev/taichi/issues/2380)
* [x] Support vector/matrix dynamic indexing (https://github.com/taichi-dev/taichi/issues/1004)
* [x] Systematic IR type system
* [ ] (Hard) Real function support (https://github.com/taichi-dev/taichi/issues/602)
* [x] Refactor the compilation workflow so that user-program errors are properly and clearly reported
   * Annotate the IRs with its source code location

## Backends

* [x] Vulkan backend support (https://github.com/taichi-dev/taichi/issues/2298)
* [x] Support `pointer` on Metal (https://github.com/taichi-dev/taichi/issues/1740)
* [ ] Other backends Including DX, WASM, OpenCL, etc (https://github.com/taichi-dev/taichi/issues/992)
* [x] Create a unified backend interface (#2736)
    * [x] Decouple `Program` from each backend
    * [ ] Have a uniform layer for data transfer and device to host synchronization (https://github.com/taichi-dev/taichi/issues/2344)

## Performance

* [ ] Speed up the JIT compilation time by parallel compilation
* [ ] Optimize the GPU backend performance in certain scenarios (https://github.com/taichi-dev/taichi/issues/2324, https://github.com/taichi-dev/taichi/issues/2395)
* [x] Continuous regression tests for the GPU backends
* [ ] Simplify the condition to use BLS
* [x] Accelerate `import taichi`
   * The `ti` CLI should only import `taichi` when necessary

## Documentation

* [x] Add Python docstring for the public APIs (#2579)
* [ ] Add documentation for the essential C++ APIs (e.g. `Program`, `Kernel`)
* [x] Automatically generate the documentation webpages from the Python/C++ API documentation.

## Cleanups

* [ ] Codebase refactoring
[...]

PR #4485, opened 2022-03-08
[misc] Add a convenient script for testing compatibility of Taichi releases.

rexwangcc:
Related issue = #2398 

- Note: this script is not well-polished.
- Some examples within Taichi repo are not included yet, e.g. `tests/python/examples/**`


misc/release_test.sh
```diff
@@ -0,0 +1,244 @@
...
+    local ORG="taichi-dev"
+    local REPO="taichi"
+
+    # divider
+    taichi::utils::line
+    taichi::utils::logger::info "Running GGUI examples"
+    
+    # clone the repo
+    taichi::utils::git_clone "${ORG}" "${REPO}"
+    cd "${REPO}/python/taichi/examples/ggui_examples"
```
knight42, line 93: I think we could leverage `pushd/popd` to switch working dir here, i.e.:
```bash
pushd  "${REPO}/python/taichi/examples/ggui_examples"

# run tests
...

popd
```

and we don't need to pass current working dir if we use `pushd/popd`.

k-ye (commented):
nit: place this in `ci/`?

rexwangcc:
> nit: place this in `ci/`?

Sure.

misc/release_test.sh
```diff
@@ -0,0 +1,244 @@
...
+    taichi::utils::line
+    taichi::utils::logger::info "Running SandyFluid examples"
+    
+    # clone the repo
+    taichi::utils::git_clone "${ORG}" "${REPO}"
+    cd "${REPO}"
+
+    # install dependencies
+    # remove the line contains pinned Taichi version for testing purposes
+    grep -v "taichi" requirements.txt > tmpfile && mv tmpfile requirements.txt
```
ailzhang, line 204: we probably need to be extra careful here as some people might hardcode taichi to a version in the requirements.txt so we may not be testing the new release :P

ailzhang (approved):
This is awesome, thanks @rexwangcc ! 
(btw we also need to run misc/generate_example_videos.py in the release test, shall we include it in the script as well?

misc/release_test.sh
knight42: > pushd/popd aren't portable

The shebang at the beginning is `#!/usr/bin/env bash`, I thought we would always use bash.

> it would change the dirs stack and require us to use +N in the script

I am curious about the reason that we should concert about the dirs stack though. I think we don't need to specify `+N` neither.. For me, `pushd/popd` are convenient commands which temporarily change working dir.

This is not a blocker, the current script is great 👍 !

rexwangcc: > The shebang at the beginning is #!/usr/bin/env bash, I thought we would always use bash.

You are right, I forgot about shebang :P

>  I am curious about the reason that we should concert about the dirs stack though.

It's usually cauz we have to manually `cd`/`pushd` within the dirs for each test block in this script after cloning the repo as we don't know what dependencies we need to install in some external repos.

misc/release_test.sh
```diff
@@ -0,0 +1,244 @@
...
+    # run tests
+    cd "${REPO}/demo" 
+    for match in $(find ./ -name "${PATTERN}"); do
+        python "${match}"
+        taichi::utils::line
+        taichi::utils::pause
+    done
+
+    # run special tests
+    python3 render_particles.py -i ./path/to/particles \
```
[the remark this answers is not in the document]
rexwangcc: Oops, good catch!

misc/release_test.sh
rexwangcc: Yup! That's what this line does, remove the line of `taichi==x.y.z` from `requirements.txt` before installing the dependencies. Do you think it's enough?

misc/release_test.sh
knight42: > we have to manually cd/pushd within the dirs for each test block in this script after cloning the repo

hmmm but we would call `popd` at the end of each test block, so I think the dirs stack would remain the same after each test block.

misc/release_test.sh
rexwangcc: > (btw we also need to run misc/generate_example_videos.py in the release test, shall we include it in the script as well?

we should... I had some issues with running that script while writing this one, e.g. I had to use `pytest` for some of the examples referenced by the script whereas run others within the script. For now I will leave that part for future work if you don't mind :P

misc/release_test.sh
rexwangcc: per offline discussion, cd changes the stack of `dirs`, so I'm going to keep this approach for now :P

misc/release_test.sh
ailzhang: Aha sorry this looks good! Thanks!

PR #4488, opened 2022-03-09
[Lang] Support simple matrix slicing

mzmzm:
Related issue = #4257
Making the matrix slice stronger based on the dynamic index. Support one row or one column slice.


python/taichi/lang/matrix.py
```diff
@@ -269,21 +269,61 @@ def _get_slice(self, a, b):
...
+                    ])
+                return Matrix([[
+                    impl.make_tensor_element_expr(self.local_tensor_proxy,
+                                                  (a, b), (self.n, self.m),
+                                                  self.dynamic_index_stride)
+                    for b in j
+                ] for a in i])
+            if isinstance(i[0], expr.Expr) or isinstance(j[0], expr.Expr):
+                raise TaichiCompilationError(
+                    "It is detected that there is no dynamic index, please consider setting 'dynamic_index = True' in ti.init to fix this Error!"
```
ailzhang, line 324: nit: "Please turn on `ti.init(..., dynamic_index=True)` to support indexing with variables." nit: "Please turn on `ti.init(..., dynamic_index=True)` to support indexing with variables."

python/taichi/lang/matrix.py
```diff
@@ -269,21 +269,61 @@ def _get_slice(self, a, b):
...
+
+        start, stop, step = index.start or 0, index.stop or (
+            self.n if dim is 0 else self.m), index.step or 1
+        if isinstance(start, expr.Expr) and isinstance(start.ptr, int):
+            start = start.ptr
+        if isinstance(stop, expr.Expr) and isinstance(stop.ptr, int):
+            stop = stop.ptr
+        if isinstance(step, expr.Expr) and isinstance(step.ptr, int):
+            step = step.ptr
+        if isinstance(start, expr.Expr) or isinstance(
```
ailzhang, line 282: IIUC you can have a small function to avoid these duplicated if checks: 
```
def helper(x, default):
    if x is None:
        return default
    if isinstance(x, expr.Expr):
        assert isinstance(x.ptr, int)
        return x.ptr
    return x
```
(feel free to edit, this might not be 100% what you want :P

ailzhang (commented):
Nice work!

python/taichi/lang/matrix.py
```diff
@@ -259,21 +260,60 @@ def _get_slice(self, a, b):
...
+            has_slice = True
+
+        if has_slice:
+            if not isinstance(i, list):
+                i = [i]
+            if not isinstance(j, list):
+                j = [j]
+
+            if self.local_tensor_proxy is not None:
+                assert self.dynamic_index_stride is not None
```
strongoier, line 299: The logic of this part seems incomplete - you should directly call `subscript` again for single variable index in order to handle the other cases, where the matrix is an element of a field or ndarray.

strongoier (changes requested):

python/taichi/lang/matrix.py
```diff
@@ -259,21 +260,60 @@ def _get_slice(self, a, b):
...
         return Matrix([[self(i, j) for j in b] for i in a])
 
+    def _cal_slice(self, index, dim):
+        start, stop, step = index.start or 0, index.stop or (
+            self.n if dim is 0 else self.m), index.step or 1
+
+        def helper(x):
+            if isinstance(x, expr.Expr):
+                if isinstance(x.ptr, int):
+                    return x.ptr
```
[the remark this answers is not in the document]
mzmzm: e.g.:`a=1  m[a,:]`

strongoier (approved):
LGTM!

PR #4489, opened 2022-03-09
[ir] Fix a bug in simplify pass

### LINKED ISSUE #4479: An error occurred when using “if-else” statements in the ti.static loop
The k loop is only executed twice when using ti.static 
```py
import taichi as ti
ti.init(arch=ti.cpu, print_ir=True, debug=True, default_fp=ti.f64)

@ti.kernel
def main():
    for i in range(5):
        for k in ti.static(range(4)):
            if i == 3:
                print(k)    #(0,1)
main()
```

The k loop is executed normally when using range()
```py
import taichi as ti
ti.init(arch=ti.cpu, print_ir=True, debug=True, default_fp=ti.f64)

@ti.kernel
def main():
    for i in range(5):
        for k in range(4):
            if i == 3:
                print(k)    #(0,1,2,3)
main()
```

mzmzm:
Related issue = fix #4479

### DIFF
--- a/taichi/transforms/simplify.cpp
+++ b/taichi/transforms/simplify.cpp
@@ -523,8 +523,8 @@ class BasicBlockSimplify : public IRVisitor {
       // TODO: What about IfStmt::true_mask and IfStmt::false_mask?
       if (current_stmt_id < block->size() - 1 &&
           block->statements[current_stmt_id + 1]->is<IfStmt>()) {
-        auto fstmt = block->statements[current_stmt_id + 1]->as<IfStmt>();
-        if (fstmt->cond == if_stmt->cond) {
+        auto bstmt = block->statements[current_stmt_id + 1]->as<IfStmt>();
+        if (bstmt->cond == if_stmt->cond) {
           auto concatenate = [](std::unique_ptr<Block> &clause1,
                                 std::unique_ptr<Block> &clause2) {
             if (clause1 == nullptr) {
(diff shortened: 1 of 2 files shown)


taichi/transforms/simplify.cpp
```diff
@@ -521,21 +521,21 @@ class BasicBlockSimplify : public IRVisitor {
     if (config.advanced_optimization) {
       // Merge adjacent if's with the identical condition.
       // TODO: What about IfStmt::true_mask and IfStmt::false_mask?
-      if (current_stmt_id > 0 &&
-          block->statements[current_stmt_id - 1]->is<IfStmt>()) {
-        auto bstmt = block->statements[current_stmt_id - 1]->as<IfStmt>();
-        if (bstmt->cond == if_stmt->cond) {
+      if (current_stmt_id < block->size() - 1 &&
+          block->statements[current_stmt_id + 1]->is<IfStmt>()) {
+        auto fstmt = block->statements[current_stmt_id + 1]->as<IfStmt>();
```
strongoier, line 526: No need to change the name :-)

strongoier (commented):
Add a test for the bug :-)

taichi/transforms/simplify.cpp
mzmzm: fix

strongoier (approved):
Thanks!

Issue #4401, opened 2022-02-28
Support offline-cache for llvm backend of Taichi


PGZXB:
### Background
Currently, all sources will be recompiled by taichi even when code weren't modified(See #1186), which is inefficient. C++, Java, or even an interpreted language like Python don't do so, which will save compiled-data/interpreted-data to file which will be loaded directly and run if src-code and its dependencies are not modified. Supporting the feature for Taichi is what we should consider, which will decrease the time in which recompiling code. I think that the work is useful and interesting, so I did some experiments for supporting offline-cache. Next step, I will complete implement and start submitting pull requests.

I plan impl it starting from supporting for llvm-backend. The following are my thoughts. **WELCOME** to point out questions or give suggestions.

### My experiment
At present, only support cpu backend.
#### Codebase
My branch for doing the experiment: [PGZXB/taichi at dev-try (github.com)](https://github.com/PGZXB/taichi/tree/dev-try). The comment starting with **"FIXME:WILL_DELETE:"**, **"TODO:WILL_DELETE:"** and **"NOTE:WILL_DELETE:"** is temporary. And logging starting with **"TEMP$$$"** too. The branch is only for do experiment. I will submit pr with opening some new them.
##### i. kernel with return and primitive params 
```python
import taichi as ti

ti.init(arch=ti.cpu)

@ti.kernel
def func4(a: ti.i32, b: ti.i32, c: ti.i32) -> ti.f32:
  return 2 * a + b + c // 2

print(func4(2, 2, 4))
```
* First run (JIT):
```shell
➜  test_$ python3 test.py
[Taichi] version 0.9.0, llvm 10.0.0, commit 2527da40, linux, python 3.8.10
[Taichi] Starting on arch=x64
TEMP$$$ Call runtime_initialize
TEMP$$$ Call runtime_get_mem_req_queue
TEMP$$$ Call LLVMRuntime_initialize_thread_pool
TEMP$$$ Call LLVMRuntime_set_assert_failed
TEMP$$$ Call LLVMRuntime_set_profiler
TEMP$$$ Call LLVMRuntime_set_profiler_start
TEMP$$$ Call LLVMRuntime_set_profiler_stop
TEMP$$$ Mangle kernel name from '__main__.func4' to '__n8__main__5func4i3i32i3i32i3i32'
TEMP$$$ Call runtime_memory_allocate_aligned
TEMP$$$ Call runtime_initialize_snodes
[...]

Issue #2198, opened 2021-03-03
Codebase cleanup


k-ye:
We'd like to prioritize the productionization of the codebase. This is an umbrella issue. Below shows a growing list of the issues we'd like to address, which should bring the codebase to a much better state:

* [ ] Add `googletest` unit test framework #2195 
* [ ] Untangle the dependencies between sub-modules #2196 
* [ ] Identify and remove the unused files as we go
* [ ] Replace `T*_NAMESPACE_BEGIN` (`END`) macros with `namespace taichi { ... }`.
* [ ] IR standardization #2193 
* ... Others TBD

PR #4500, opened 2022-03-10
[opt] Support offline-cache for kernel with arch=cpu

PGZXB:
Related issue = #4401


taichi/llvm/llvm_offline_cache.cpp
```diff
@@ -0,0 +1,87 @@
+#include "llvm_offline_cache.h"
+
+#include "llvm/AsmParser/Parser.h"
+#include "llvm/Support/SourceMgr.h"
+#include "llvm/Support/raw_os_ostream.h"
+#include "llvm/IR/Module.h"
+
+TLANG_NAMESPACE_BEGIN
```
strongoier, line 8: According to #2198, we should at least not introduce new files with `T*_NAMESPACE_BEGIN`.

taichi/llvm/llvm_offline_cache.h
```diff
@@ -0,0 +1,66 @@
+#pragma once
+
+#include "taichi/common/core.h"
+#include "taichi/llvm/llvm_fwd.h"
+#include "taichi/util/io.h"
+
+TLANG_NAMESPACE_BEGIN
```
strongoier, line 7: ditto

PR #4508, opened 2022-03-11
[ci] Reduce the default device memory usage for GPU tests

qiao-bo:


ailzhang (approved):
Thanks!!! (i'll merge this to unblock other PRs, would you mind sending another one to update release.yml as well?

PR #4513, opened 2022-03-11
[build] Option to use LLD

bobcao3:
(Seems to have worked on my end)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,12 @@ if(CCACHE_PROGRAM)
     set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
 endif()
 
+if (USE_LLD)
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fuse-ld=lld")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -56,6 +56,12 @@ if(CCACHE_PROGRAM)
     set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
 endif()
 
+if (USE_LLD)
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fuse-ld=lld")
```
ailzhang, line 60: Can you make USE_LLD an option so that we can set it through TAICHI_CMAKE_ARGS?

ailzhang (approved):
Confirm that this works on my end as well, thanks a lot !!

CMakeLists.txt
bobcao3: I think you can, I did that on my end.

ailzhang: Yea I did try that and it worked :P . I meant we should add it in this PR to make it more usable. 😁

qiao-bo (approved):
LGTM

PR #4514, opened 2022-03-12
[misc] [linux] Implement XDG Base Directory support

### LINKED ISSUE #4504: Follow the XDG Base Directory Specification
## Concisely describe the proposed feature

Currently, Taichi creates a hardcoded `~/.taichi` directory in the home directory. As far as I am concerned, random programs putting their files in my home directory is a cardinal sin. Taichi should follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) instead of polluting the user home directory.

## Describe the solution you'd like

Before choosing the solution, we need to consider 3 questions:

1) Does `~/.taichi` currently contain configuration files?
2) Does `~/.taichi` currently contain data, that **needs** to be persisted across reboots?
3) Will changing the location of `~/.taichi` break anything for current users?

For now, I am going to assume that the answer to all 3 questions is "No". If that is the case, then we can just change the path to `${XDG_CACHE_HOME}/taichi` without any extra compatibility logic. In practice, this would mean changing the non-windows part of the [`get_repo_dir` function](https://github.com/taichi-dev/taichi/blob/570c39d935ac621f4e44eb79fb5fb37c919d6d65/taichi/common/core.cpp#L35) from

```cpp
auto home = std::getenv("HOME");
TI_ASSERT(home != nullptr);
return std::string(home) + "/.taichi/";
```

to

```cpp
auto xdg_cache = std::getenv("XDG_CACHE_HOME");

std::string xdg_cache_str;
if (xdg_cache != nullptr) {
  xdg_cache_str = xdg_cache;
} else {
  // XDG_CACHE_HOME is not defined, defaulting to ~/.cache
  auto home = std::getenv("HOME");
  TI_ASSERT(home != nullptr);
  xdg_cache_str = home;
  xdg_cache_str += "/.cache";
}

return xdg_cache_str + "/taichi/";
```

---

If the answer to (1) or (2) is "Yes", then just use `XDG_CONFIG` or `XDG_DATA` instead of `XDG_CACHE`.

If the answer to (3) is "Yes", then we need to check if `~/.taichi` already exists and either move the folder to the XDG location or just returning the old path from `get_repo_dir`.

RuRo:
Related issue = close #4504

### DIFF
--- a/taichi/common/core.cpp
+++ b/taichi/common/core.cpp
@@ -28,13 +28,23 @@ void set_python_package_dir(const std::string &dir) {
 }
 
 std::string get_repo_dir() {
-  // release mode. Use ~/.taichi as root
 #if defined(TI_PLATFORM_WINDOWS)
   return "C:/taichi_cache/";
 #else
-  auto home = std::getenv("HOME");
-  TI_ASSERT(home != nullptr);
-  return std::string(home) + "/.taichi/";
+  auto xdg_cache = std::getenv("XDG_CACHE_HOME");
+
+  std::string xdg_cache_str;
+  if (xdg_cache != nullptr) {
+    xdg_cache_str = xdg_cache;
+  } else {
+    // XDG_CACHE_HOME is not defined, defaulting to ~/.cache
+    auto home = std::getenv("HOME");
+    TI_ASSERT(home != nullptr);
+    xdg_cache_str = home;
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Awesome thanks!

PR #4517, opened 2022-03-13
[misc] Make result of irpass::print hold more information

PGZXB:
Related issue = #4401 

Result of `irpass::print` will be used as AST-Key (maybe temporary), which needs more information/details to implement [offline-cache](https://github.com/taichi-dev/taichi/issues/4401#issuecomment-1063559899).


strongoier (approved):
LGTM!

PR #4521, opened 2022-03-14
[Lang] [refactor] Remove legacy ways to construct matrices

strongoier:
Related issue = #2880

After this PR, the only way to use `Matrix.__init__()` is by passing initial values.

Hint for reviewers: the git diff looks a bit messy, but the most modifications (L38-L132) in `Matrix.__init__()` simply cancels the indentation after removing the check of `n`.


ailzhang (approved):
LGTM!
I'm a little concerned about breaking existing users with this change, we should at least highlight this change in our next release note.

strongoier:
> LGTM! I'm a little concerned about breaking existing users with this change, we should at least highlight this change in our next release note.

Yes we should highlight that.

PR #4522, opened 2022-03-14
[aot] Add binding_id of root/gtmp/rets/args bufs to CompiledOffloadedTask

### LINKED ISSUE #3961: [aot] Test what happens when both scalar and Ndarray args are used

ailzhang:
Adding binding id of root/gtmp/rets/args to the serialized json file. 
Note this PR doesn't handle multiple root buffers in aot yet. 

Also added a test case with both scalar and ndarray args, which fixes #3961.

Related issue = #

### DIFF
--- a/taichi/aot/module_data.h
+++ b/taichi/aot/module_data.h
@@ -28,6 +28,22 @@ struct CompiledFieldData {
             element_shape);
 };
 
+enum class BufferType { Root, GlobalTmps, Args, Rets };
+
+struct BufferInfo {
+  BufferType type;
+  int id{-1};  // only used if type==Root
+
+  TI_IO_DEF(type, id);
+};
+
+struct BufferBind {
+  BufferInfo buffer;
+  int binding{0};
+
+  TI_IO_DEF(buffer, binding);
+};
+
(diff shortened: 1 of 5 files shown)


bobcao3:
Vulkan will bind the root or gtmp to **multiple** bind points if aliasing is required.

taichi/aot/module_data.h
```diff
@@ -36,7 +36,22 @@ struct CompiledOffloadedTask {
   std::string source_path;
   int gpu_block_size{0};
 
-  TI_IO_DEF(type, range_hint, name, source_path, gpu_block_size);
+  // TODO: Vulkan backend supports multiple root buffers, update this once
+  // support multiple roots in AOT.
+  int root_buffer_bind_id{-1};
```
bobcao3, line 41: Should probably just look at Vulkan's buffer binds struct (a vector of tuple `<buffer, bind_id>`

bobcao3 (changes requested):
Vulkan still won't work fully with this...

bobcao3 (approved):
LGTM!

PR #4524, opened 2022-03-14
[vulkan] Relax a few test precisions for vulkan

ailzhang:
On my workstation this test failed on vulkan backend due to precision
issue. Let's relax it a bit for now.

Related issue = #


qiao-bo (approved):
LGTM

bobcao3 (approved):
LGTM, wonder which device it is? Maybe we can record some precision data. (1e-3 seems like a large error)

ailzhang:
@bobcao3 This is on an Intel UHD Graphics 630.

PR #4525, opened 2022-03-14
[Lang] Add ti.serialize and ti.loop_config

lin-hitonami:
Related issue = #4530


python/taichi/lang/misc.py
```diff
@@ -510,19 +510,36 @@ def loop_unique(val, covers=None):
...
+        get_runtime().prog.current_ast_builder().strictly_serialize()
 
 
+parallelize = _parallelize
+
 serialize = lambda: parallelize(1)
 
 
-def block_dim(dim):
+def _block_dim(dim):
```
k-ye, line 524: Let's not yet remove this functions, just mark them as deprecated?

taichi/transforms/lower_ast.cpp
```diff
@@ -204,9 +204,8 @@ class LowerAST : public IRVisitor {
       auto end = stmt->end;
       flatten_rvalue(begin, &fctx);
       flatten_rvalue(end, &fctx);
-      bool is_good_range_for =
-          current_block_depth_ == 1 || detected_fors_with_break_.find(stmt) ==
-                                           detected_fors_with_break_.end();
+      bool is_good_range_for = detected_fors_with_break_.find(stmt) ==
```
strongoier, line 207: Why this change? I forget what is the behavior of top-level break now...

lin-hitonami: We are now supporting serial fors at the outermost scope, and they can contain break statements, so we should convert them to while statements.

strongoier (approved):
LGTM!

python/taichi/lang/misc.py
```diff
@@ -510,19 +510,32 @@ def loop_unique(val, covers=None):
...
     """Set the number of threads in a block to `dim`.
     """
     get_runtime().prog.current_ast_builder().block_dim(dim)
 
 
+def loop_config(block_dim=None, serialize=None, parallelize=None):
+    if block_dim is not None:
```
ailzhang, line 530: nit: add a docstring since it's user facing?

python/taichi/lang/misc.py
```diff
@@ -510,19 +510,32 @@ def loop_unique(val, covers=None):
...
+        _block_dim(block_dim)
+
+    if serialize:
```
ailzhang, line 533: Looks like serialize should be a boolean defaulting to false instead of None?

ailzhang (approved):
LGTM!

PR #4532, opened 2022-03-15
[Lang] Add taichi gallery command for user to choose and run example in gui

TiGeekMan:
Add `python3 -m taichi gallery` cli command so that the user can choose and run example through a GUI

### DIFF
--- a/python/taichi/_main.py
+++ b/python/taichi/_main.py
@@ -9,6 +9,7 @@ import timeit
 from collections import defaultdict
 from functools import wraps
 from pathlib import Path
+import glob
 
 import numpy as np
 from colorama import Fore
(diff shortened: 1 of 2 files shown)


python/taichi/_main.py
```diff
@@ -135,6 +136,85 @@ def support_choice_with_dot_py(choice):
 
         return support_choice_with_dot_py
 
+    @register
+    def gallery(self, argumets: list = sys.argv[2:]):
```
ailzhang, line 140: Nit:add a docstring for gallery as well?

ailzhang (approved):
This is a great addition to taichi, thanks!

python/taichi/_main.py
TiGeekMan: Sure. Docstring added.

PR #4533, opened 2022-03-15
[vulkan] Fix a memory leak caused by a typo

bobcao3:
This means all of our recycled command buffer is stored into the wrong recycling pool and therefore causing the allocation of new command buffers every time.

TLDR: tiny mistakes can be quite lethal (and hard to find)


k-ye (approved):
Ha, nice catch. We should really enable all warnings...

PR #4536, opened 2022-03-15
[llvm] [lang] Add support for multiple return statements in real function

lin-hitonami:
Related issue = #602 

For reviewers:
This PR inserts return instruction when visiting return statement in the codegen.
 
LLVM requires a basic block to have one and only one terminator instruction (return, branch, etc.) at the end of the block. So, this PR does the following things.
1. Adds a flag `returned` in the codegen, indicating whether the current basic block has inserted a return instruction. 
2. If the current basic block has inserted a return instruction, the codegen stops processing all statements after it in a basic block, and does not add the branch instruction supposed to be added when constructing if/for/while statements. 
3. If the basic block does not have a terminator instruction, insert a return instruction.
4. Restores the `returned` flag to `false` when the current basic block completes.

### DIFF
--- a/taichi/codegen/codegen_llvm.cpp
+++ b/taichi/codegen/codegen_llvm.cpp
@@ -72,9 +72,12 @@ FunctionCreationGuard::FunctionCreationGuard(
 }
 
 FunctionCreationGuard::~FunctionCreationGuard() {
-  mb->builder->CreateRetVoid();
+  if (!mb->returned) {
+    mb->builder->CreateRetVoid();
+  }
   mb->func = old_func;
   mb->builder->restoreIP(ip);
+  mb->returned = false;
 
   {
     llvm::IRBuilderBase::InsertPointGuard gurad(*mb->builder);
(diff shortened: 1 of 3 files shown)


strongoier (commented):
It might help reviewers if you leave some comments in the code or briefly introduce your solution in the PR description :-)

lin-hitonami:
> It might help reviewers if you leave some comments in the code or briefly introduce your solution in the PR description :-)

Updated PR description.

taichi/codegen/codegen_llvm.cpp
```diff
@@ -72,16 +72,20 @@ FunctionCreationGuard::FunctionCreationGuard(
 }
 
 FunctionCreationGuard::~FunctionCreationGuard() {
-  mb->builder->CreateRetVoid();
+  if (!mb->returned) {
```
strongoier, line 75: Does `finalize_offloaded_task_function()` already ensure the existence of return?

strongoier (commented):
Thanks for the clear description!

taichi/codegen/codegen_llvm.cpp
lin-hitonami: They are used in different places. Take range-for for example, `initialize_offloaded_task_function` and `finalize_offloaded_task_function` creates and finalizes a offloaded task function. In the offloaded task function, the codegen uses `FunctionCreationGuard` to create a function that contains the body of the for loop (let's call it the body function). Then, the offloaded task function calls `cpu_parallel_range_for` to execute the body function in parallel.

strongoier (approved):
LGTM!

PR #4537, opened 2022-03-15
[bug] Fix bug that calling std::getenv when cpp-tests running will fail

PGZXB:
Related issue = #

Cpp-tests is called by `subprocess.check_call()`, which causes env-variables missing, so calling `std::getenv()` in cpp-test will fail ( See https://github.com/taichi-dev/taichi/runs/5534004071?check_suite_focus=true
The pr fix the bug.

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -31,6 +31,10 @@ def _test_cpp():
             e: os.getenv(e)
             for e in env_used_by_taichi_core
         }
+        env_used_by_taichi_core = {
+            k: v
+            for k, v in env_used_by_taichi_core.items() if v is not None
+        }
         subprocess.check_call(f'./{cpp_test_filename}',
                               env={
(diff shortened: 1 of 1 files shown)


tests/run_tests.py
```diff
@@ -17,8 +17,25 @@ def _test_cpp():
     curr_dir = os.path.dirname(os.path.abspath(__file__))
     build_dir = os.path.join(curr_dir, '../build')
     if os.path.exists(os.path.join(build_dir, cpp_test_filename)):
+        env_used_by_taichi_core = [
+            'HOME',
```
ailzhang, line 21: Explicitly listing all used envs out seems pretty hard to maintain. As our linux CI is in docker (and windows is moving towards it), maybe it's safe to just copy all envs over to the subprocess?

ailzhang (approved):
Lgtm!

PR #4539, opened 2022-03-15
[misc] Metadata update

Leonz5288:
Related issue = #


python/taichi/_version_check.py
```diff
@@ -43,6 +43,8 @@ def check_version(cur_uuid):
         payload['python'] = 'cp310'
 
     payload['uuid'] = cur_uuid
+    if os.environ['TI_CI'] == '1':
```
rexwangcc, line 46: If `TI_CI` is not set, this will throw, it's it expected behavior? (otherwise we may need to use os.getenv("TI_CI"))

Leonz5288: Got it

rexwangcc (approved):
LGTM

PR #4545, opened 2022-03-16
[build] Default symbol visibility to hidden for all targets

ghuau-innopeak:
Taichi can be used by external applications that has conflicting
dependencies such as ImGui or Vulkan. In order to avoid any conflicts or
nasty bugs, enforce the visibility to 'hidden' by default so only the
public API is exported.

This moves the config to the global property instead of the per target
property


k-ye (approved):
LGTM! It's great to see the linker map finally gone :-)

qiao-bo (approved):
LGTM! Thanks. +1 for linker map goes.

PR #4491, opened 2022-03-09
[ci] Add a Dockerfile for building manylinux2014-compatible Taichi wheels with CUDA backend

strongoier:
Related issue = #4395

PR #4550, opened 2022-03-16
[CI] Generate manylinux2014-compatible wheels with CUDA backend in release workflow

### LINKED ISSUE #4395: Add a manylinux2014 compatible release with only LLVM-based backends

strongoier:
Related issue = close #4395

The docker image produced by #4491 was uploaded as `registry.taichigraphics.com/taichidev-manylinux2014-cuda:v0.0.0`. Then this PR adds a job in release workflow to build and test the new wheel with CPU and CUDA backends in that image.

The new wheel is named as `taichi-0.9.2-cp38-cp38-manylinux2014_x86_64.whl`, while the original one (with OpenGL/Vulkan support as well) is renamed as `taichi-0.9.2-cp38-cp38-manylinux_2_27_x86_64.whl` (the original name with `manylinux1` is wrong). `pip install` will choose the wheel closest to the user platform.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,6 +125,71 @@ jobs:
         run: |
           docker rm taichi_build taichi_test -f
 
+  build_and_test_manylinux2014:
+    name: Build and Upload (manylinux2014)
+    needs: matrix_prep
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.matrix_prep.outputs.matrix) }}
+    runs-on: [self-hosted, cuda, cn, release]
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: "recursive"
+
+      - name: Get sccache cache
+        uses: actions/cache@v2
+        with:
+          path: sccache_cache
+          key: sccache-manylinux-${{ github.sha }}
+          restore-keys: |
+            sccache-manylinux-
+
+      - name: Build
+        run: |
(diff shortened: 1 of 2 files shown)


frostming (commented):
Mostly LGTM but is there any chance to merge it into the matrix of build and test linux?

I see the only difference is the image name and `TI_WANTED_ARCHS`(does it make difference if we make them the same for this env var?)

strongoier:
> Mostly LGTM but is there any chance to merge it into the matrix of build and test linux?
> 
> I see the only difference is the image name and `TI_WANTED_ARCHS`(does it make difference if we make them the same for this env var?)

I agree that there are some similarities, but there are still other differences like:
1. Different `TAICHI_CMAKE_ARGS`
2. No `-v /tmp/.X11-unix:/tmp/.X11-unix`
3. No `-e DISPLAY`

I am not sure if putting these together will make things clearer.

frostming:
@strongoier Ok then that is fine to keep it as it is now. I didn't compare them carefully.

ailzhang (approved):
LGTM!

qiao-bo (approved):
LGTM! In the future, we might need to think about adding to test.yml as well.

frostming (approved):

PR #4551, opened 2022-03-16
[misc] Add TI_CI env to CI/CD

Leonz5288:
Related issue = #


ailzhang (commented):
hmmm why do we need this?

Leonz5288:
> 

We want to also record CI in user data. This is to tag CI records.

frostming (commented):
Should better be put once in the top level, the same as `on` and `jobs`

```yaml
env:
  TI_CI: 1
```

PR #4552, opened 2022-03-16
[refactor] Fix error message

k-ye:
Related issue = #

### DIFF
--- a/python/taichi/aot/module.py
+++ b/python/taichi/aot/module.py
@@ -24,8 +24,8 @@ class KernelTemplate:
             if val is v:
                 key_p += '=' + ky + ','
                 return key_p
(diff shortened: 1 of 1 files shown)


ailzhang (approved):
Thanks!

python/taichi/aot/module.py
```diff
@@ -25,7 +25,7 @@ def keygen(v, key_p, fields):
                 key_p += '=' + ky + ','
                 return key_p
         raise RuntimeError('Arg type must be of type int/float/boolean' +
-                           'or taichi field. Type ' + str(type(v)) +
+                           ' or taichi field. Type ' + str(type(v)) +
                            ' is not supported')
```
frostming, line 29: The `+` is not needed here, it is more efficient and cleaner

frostming (approved):

PR #4554, opened 2022-03-16
[aot] Fix get_kernel API semantics

k-ye:
Related issue = https://github.com/taichi-dev/taichi/pull/4437

We should return a raw pointer, not `unique_ptr`. The module object owns the kernel.


qiao-bo (approved):
Thanks for the fix

PR #4555, opened 2022-03-17
[gui] Hack to make GUI.close() work on macOS

k-ye:
Had to use this sleeping-hack to make closing the window work. Not sure if there's an event handler we can register to control this behavior much more accurately.


yuanming-hu (approved):
Many thanks! LGTM :-)

neozhaoliang (approved):

PR #4567, opened 2022-03-18
[Doc] Format updates

Vissidarte-Herman:
H1 + H2 cause too many transverse lines. I propose H2 + H3. Your call!

BTW, in technical documents, we also use *italic* to indicate results of steps.

Related issue = #

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 import taichi as ti
 ```
 
-# What is Taichi?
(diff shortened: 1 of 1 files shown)


k-ye (approved):
LGTM! Note that if we need to bring back TOC later on, this header might still need to be adjusted... (Or there'll be an extra level of indentation)

PR #4573, opened 2022-03-18
[aot] Remove redundant module_path argument

ailzhang:
We already pass module path inside mod_params so it's redundant for
`load()` to take module_path.

Related issue = #


qiao-bo (approved):
LGTM, thanks!

Issue #4574, opened 2022-03-18
Support (predefined layout) Ndarrays as GGUI's VBO


k-ye:
**Concisely describe the proposed feature**

It would be convenient to be able to use Ndarray as GGUI's VBO. For example, I may want to use an Ndarray of vec3 to represent the vertex positions.

**Describe the solution you'd like (if any)**

Each GGUI renderable already maintains a `vertex_buffer_`: https://github.com/taichi-dev/taichi/blob/d99acb54c4091679610eaf7dbb46147f34dfbbae/taichi/ui/backends/vulkan/renderable.h#L59 Right now the VBO layout is fixed as https://github.com/taichi-dev/taichi/blob/d99acb54c4091679610eaf7dbb46147f34dfbbae/taichi/ui/backends/vulkan/vertex.h#L6-L26

We can selectively register a subset of `Vertex`'s attributes  https://github.com/taichi-dev/taichi/blob/d99acb54c4091679610eaf7dbb46147f34dfbbae/taichi/ui/backends/vulkan/renderable.cpp#L145-L149. Then, everything should just work.

**Additional comments**

k-ye:
Fixed by #4575

PR #4575, opened 2022-03-18
[gui] Make GGUI VBO configurable

k-ye:
Related issue = #4574

Added an enum `VboAttribes`, which has the following VBO profiles:

* position only
* position + normal
* position + normal + UV
* (All) position + normal + UV + color

This is added to `RenderableInfo`, so that all Renderable except for `SetImage` can support this VBO config.

Only Circles is changed to actually use this enum for now. This is to allow this AOT demo to work: https://github.com/damonwy/taichi-aot-tests/blob/0c57df64b78bf46a982f18a424d0ceb83a78a0a2/desktop/mpm88.cpp#L115-L119


taichi/ui/common/renderable_info.h
```diff
@@ -7,7 +9,8 @@ TI_UI_NAMESPACE_BEGIN
 struct RenderableInfo {
   FieldInfo vbo;
   FieldInfo indices;
-  bool has_per_vertex_color;
+  bool has_per_vertex_color{false};
```
AmesingFlank, line 12: I'm guessing this property could be removed in the future, as `vbo_attrs` should be enough to capture this information.
But fine for now.

taichi/ui/backends/vulkan/vertex.cpp
```diff
@@ -0,0 +1,24 @@
...
+    case VboAttribes::kAll:
+      return sizeof(Vertex);
+    case VboAttribes::kPos:
+      return offsetof(Vertex, normal);
+    case VboAttribes::kPosNormal:
+      return offsetof(Vertex, texCoord);
+    case VboAttribes::kPosNormalUv:
+      return offsetof(Vertex, color);
+    default:
+      break;
```
AmesingFlank, line 19: nit, throw an error here?

taichi/ui/backends/vulkan/renderer.cpp
```diff
@@ -43,37 +44,40 @@ void Renderer::set_background_color(const glm::vec3 &color) {
 }
 
 void Renderer::set_image(const SetImageInfo &info) {
-  SetImage *s = get_renderable_of_type<SetImage>();
+  SetImage *s = get_renderable_of_type<SetImage>(VboAttribes::kAll);
```
AmesingFlank, line 47: can probably use `kPosNormalUv` here. Although I don't think it would make any substantial difference.

taichi/ui/backends/vulkan/vertex.h
```diff
@@ -21,9 +23,19 @@ struct Vertex {
...
   vec2 texCoord;
   vec4 color;
 };
 
+enum class VboAttribes {
+  kAll,
+  kPos,
+  kPosNormal,
+  kPosNormalUv,
+};
```
AmesingFlank, line 36: Isn't this still a bit restrictive? 
For example, a user might choose to provides ndarrays for only Pos and Color. In this case, the VBO would still contain Normal and UV, and we'd run into the same problem that we have now.
Would it be better to have something like
```cpp
enum VertexAttribs : int
{
    Pos = 0b1,
    Normal = 0b10,
    Uv = 0b100,
    Color = 0b1000,
}
inline VertexAttribs operator|(VertexAttribs a, VertexAttribs b)
{
    return static_cast<VertexAttribs>(static_cast<int>(a) | static_cast<int>(b));
}
bool hasAttrib(VertexAttribs a, VertexAttribs attrib){
   return static_cast<int>(a) &  static_cast<int>(attrib);
}
```

k-ye: Ah yes, will change to this one, thanks!

taichi/ui/backends/vulkan/renderer.cpp
k-ye: Huh, better not change the behavior here? I feel like other parts (especially `staging_buffer.py`) needs some tweak before this will work?

taichi/ui/backends/vulkan/renderables/set_image.cpp
```diff
@@ -154,19 +154,23 @@ void SetImage::update_vertex_buffer_() {
...
+  // Our actual VBO might only use the first several attributes in `Vertex`,
+  // therefore this slicing & copying for each Vertex.
+  const size_t vbo_size = config_.vbo_size();
   {
-    Vertex *mapped_vbo =
-        (Vertex *)app_context_->device().map(staging_vertex_buffer_);
-
-    memcpy(mapped_vbo, vertices.data(),
-           (size_t)config_.vertices_count * sizeof(Vertex));
+    char *mapped_vbo =
```
k-ye, line 161: I remembered why I didn't go with the bitmask approach in the beginning. This part would require the users to always follow the order Vertex attributes are defined. I.e., you can use {pos, uv}, but not {uv, pos}. Guess this limitation is not that significant, though.

AmesingFlank (approved):

PR #4580, opened 2022-03-19
[Lang] [ir] Move short-circuit boolean logic into AST-to-IR passes

### LINKED ISSUE #4804: Remove IfStmt::true_mask, false_mask
Title; these two members are not used at all.

re-xyr:
Related issue = #3572, #3635, Fixes #4804


re-xyr:
Two tests were failing because I used the new `logical_and` in `ASTTransformer.build_Compare`: `test_element_wise` and `test_bit_array_vectorization`. Some thoughts:

- How does the python `Matrix` handle comparations (or other operators)? I could't find relevant code in `ops.py`,  `matrix.py` or `ast_transformer.py`.
- The bit vectorization pass seems very fragile, as seemingly irrelevant changes will screw it up to the point of the compiler panicking (this is not the first time, see #4180). Maybe we need to revise this pass at some point to make it more robust?

re-xyr:
I have tested locally with `build_Compare` using `logical_and` and `short_circuit_operators` on by default. Turns out all but one of the problems are caused by this one fact: Taichi's logical true is `-1` while Python's is `1`. Prepare for a massive test rewrite...

re-xyr:
@strongoier I think this is good to merge for now.

strongoier (approved):
LGTM!

Issue #4577, opened 2022-03-19
Refactor counters to scoped-counters


PGZXB:
Following #4561, in C++ codebase of Taichi, many counters are used to identify a instance. But the counters are global now, which causes some problems. Using Taichi AST string generated by `irpass::print` as offline-cache-key (#4401) cannot identify a kernel completely because of global counters. We need scoped-counters.

#### An example in which I found problem
```Python
ti.init(arch=ti.cpu, print_preprocessed_ir=True)

@ti.kernel
def f1():
  for i in range(100):
    pass

@ti.kernel
def f2():
  for i in range(100):
    pass
```
**1st:** `f1(); f2()`, Ouput:
```shell
[I 03/19/22 10:47:49.763 17728] [kernel.cpp:lower@79] [f1_c56_0] Preprocessed IR:
kernel {
  $0 : for @tmp0 in range((cast_value<i32> 0), (cast_value<i32> 100)) block_dim=adaptive {
  }
}
[I 03/19/22 10:47:49.792 17728] [kernel.cpp:lower@79] [f2_c58_0] Preprocessed IR:
kernel {
  $0 : for @tmp2 in range((cast_value<i32> 0), (cast_value<i32> 100)) block_dim=adaptive {
  }
}
```
**2nd:** `f2(); f1()`, Output:
```shell
[I 03/19/22 10:49:09.890 17800] [kernel.cpp:lower@79] [f2_c58_0] Preprocessed IR:
kernel {
  $0 : for @tmp0 in range((cast_value<i32> 0), (cast_value<i32> 100)) block_dim=adaptive {
  }
}
[I 03/19/22 10:49:09.919 17800] [kernel.cpp:lower@79] [f1_c56_0] Preprocessed IR:
kernel {
  $0 : for @tmp2 in range((cast_value<i32> 0), (cast_value<i32> 100)) block_dim=adaptive {
  }
}
```
One kernel's ir/ast-string is related with calling order, because the `id`: `Identifier` which is member of `IdExpression` has auto incremental `id`: `int` (depending on static int variable `id_counter`). Global `id_counter` causes the problem. 
I think we can remove this counters, and make them scoped.

#### Counters in C++ Side
(Simply using reg "static.*int.*counter" to search)
- [x] `Identifier::id_counter`, global scope and kernel/func scope
- [ ] `SNode::counter`, program(global) scope
- [ ] `Stmt::instance_id_counter`, kernel/func scope
[...]

PR #4581, opened 2022-03-19
[refactor] Refactor Identifier::id_counter to global and local counter

PGZXB:
Related issue = #4577

### DIFF
--- a/taichi/ir/ir.h
+++ b/taichi/ir/ir.h
@@ -78,16 +78,12 @@ class MemoryAccessOptions {
 
 class Identifier {
  public:
-  static int id_counter;
   std::string name_;
+  int id{0};
 
-  int id;
+  // Identifier() = default;
 
   // Multiple identifiers can share the same name but must have different id's
-  Identifier(const std::string &name_ = "") : name_(name_) {
-    id = id_counter++;
-  }
-
   Identifier(int id, const std::string &name = "") : name_(name), id(id) {
   }
 
@@ -103,10 +99,6 @@ class Identifier {
 
   bool operator==(const Identifier &o) const {
     return id == o.id;
-  }
-
-  static void reset_counter() {
-    id_counter = 0;
   }
 };
 
--- a/taichi/ir/ir.h
+++ b/taichi/ir/ir.h
@@ -88,6 +88,9 @@ class Identifier {
     id = id_counter++;
   }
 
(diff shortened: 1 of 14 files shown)


taichi/ir/ir.h
```diff
@@ -88,6 +88,9 @@ class Identifier {
     id = id_counter++;
```
k-ye, line 88: Should we remove `id_counter` later?

PGZXB: In the pr, the `is_counter` counts global identifier. I think that we can move `id_counter` into `Program` as member of it. WDYT?

k-ye: So, what kind of "global" variables is stamped by this identifier? I'm slightly against putting it into `Program`, as `Program` is more for runtime stuff. I.e., it's better to make AST module self contained.

PGZXB: "global" variables: global vars referenced by kernel, e.g. `a = ti.field(ti.i32, ())`, which don't belongs to a kernel but a `Program`, so maybe the 'id_counter' should be member of `Program` and rename it to `global_id_counter`. 


>it's better to make AST module self contained.

LGTM, but the global taichi objects are not built by `ASTBuilder`.

k-ye: I see. I think it's fine to move it to `Program` as a start, but long term speaking, we should have dedicated counters for each kind of objects. E.g. `SNode::id_counter`, `Field::id_counter`, etc.

PGZXB: Emmm... I have a immature idea. If my changes in the pr are merged, the `Identifier::id_counter` will only be used by `GlobalVariableExpression::GlobalVariableExpression(SNode *snode)` (as  `Identifier::id`). Can we use `snode->id` as `Identifier::id` and remove `Identifier::id_counter`?

k-ye: Hmm... I believe Ndarray also uses GlobalVariableExpression, and I'm not sure whether `GlobalVariableExpression` and `SNode` has a one-to-one mapping. Do  you remember? @strongoier @lin-hitonami

PGZXB: Very sorry, my bad, you are right. `GlobalVariableExpression` is used by `SNode` and others. 
I will move `id_counter` to Program and commit to the pr.

strongoier (approved):
Great! Thanks!

PR #4582, opened 2022-03-19
[bug] Fix bug that caching kernels with same AST will fail

PGZXB:
Related issue = #4401

### DIFF
--- a/taichi/llvm/llvm_program.cpp
+++ b/taichi/llvm/llvm_program.cpp
@@ -717,7 +717,6 @@ void LlvmProgramImpl::cache_kernel(
   if (cache_data_.kernels.find(kernel_key) != cache_data_.kernels.end()) {
     return;
   }
-  return;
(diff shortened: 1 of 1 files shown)


taichi/llvm/llvm_program.cpp
```diff
@@ -714,7 +714,8 @@ void LlvmProgramImpl::cache_kernel(
     const std::string &kernel_key,
     llvm::Module *module,
     std::vector<std::string> &&offloaded_task_name_list) {
-  TI_ASSERT(cache_data_.kernels.find(kernel_key) == cache_data_.kernels.end());
+  if (cache_data_.kernels.find(kernel_key) != cache_data_.kernels.end())
```
k-ye, line 717:
```suggestion
  if (cache_data_.kernels.find(kernel_key) != cache_data_.kernels.end()) {
    return;
  }
```

k-ye (approved):
LGTM!

PR #4589, opened 2022-03-21
[misc] Write version info right after creation of uuid

Leonz5288:
To prevent creation of file only happens after success of server response.
Some Taichi program's execution time won't last to listen to server response.


strongoier (approved):
LGTM!

PR #4593, opened 2022-03-21
[aot] Support return in vulkan aot

ailzhang:
Vulkan aot return didn't work since it was trying to serialize
`Datatype` which was a raw pointer. This PR fixes the issue by changing
it to a PrimitiveTypeID which is a `int`. Note it's possible that we
limit Kernel args and rets to be primitive types only but leaving that
to a followup PR to avoid making this PR too complicated.

Related issue = #


k-ye (approved):
LGTM!

PR #4596, opened 2022-03-22
[misc] Add shape of snode to result of irpass::print()

PGZXB:
Related issue = #4401


strongoier:
IMHO SNode should be serialized separately in a more systematic way - shape is only part of it.

Issue #4602, opened 2022-03-22
[RFC] Taichi Language Reference


strongoier:
**Concisely describe the proposed feature**
A programming language needs a specification. Both users and developers need to refer to it to determine whether a certain behavior is *correct*, *buggy*, or *undefined*.

The specification needs to be documented. It should not only stay in someone's mind. Otherwise, there will always be issues around language behavior, and maintainers have to guess who knows or can make the decision. Different maintainers have different minds, and people forget things from time to time. As a result, the language might go back-and-forth.

To avoid the above problems and standardize Taichi starting from v1.0, I propose to write a specification for Taichi.

**Describe the solution you'd like (if any)**
Because the syntax of Taichi is a subset of that of Python, I propose to model the specification after the [Python language reference](https://docs.python.org/3/reference/index.html). This route is also what [TorchScript](https://pytorch.org/docs/stable/jit_language_reference_v2.html) takes.

Roadmap:

- [x] Introduction
- [x] Basic concepts
    - [x] Values and types
    - [x] Naming and scope
- [x] Expressions
    - [x] Atoms
        - [x] Identifiers (Names)
        - [x] Literals
        - [x] Parenthesized forms
        - [x] List and dictionary displays
    - [x] Primaries
        - [x] Attribute references
        - [x] Subscriptions
        - [x] Slicings
        - [x] Calls
    - [x] The power operator
    - [x] Unary arithmetic and bitwise operations
    - [x] Binary arithmetic operations
    - [x] Shifting operations
    - [x] Binary bitwise operations
    - [x] Comparisons
        - [x] Value comparisons
        - [x] Membership test operations
    - [x] Boolean operations
    - [x] Assignment expressions 
    - [x] Conditional expressions
    - [x] Static expressions
    - [x] Expression lists
- [x] Simple statements
    - [x] Expression statements
    - [x] Assignment statements
        - [x] Augmented assignment statements
        - [x] Annotated assignment statements
    - [x] The `assert` statement
    - [x] The `pass` statement
    - [x] The `return` statement
    - [x] The `break` statement
    - [x] The `continue` statement
- [x] Compound statements
    - [x] The `if` statement
    - [x] The `while` statement
    - [x] The `for` statement

AmesingFlank:
Highly relevant discussion: https://github.com/taichi-dev/taichi/discussions/4594

lin-hitonami:
`6.12. Assignment expressions` seems to be missing, Taichi supports it.

[...]

PR #4603, opened 2022-03-22
[Doc] Set up the basis of Taichi specification

strongoier:
Related issue = #4602

This PR sets up the basic architecture of the specification, and includes the content of the preface, the "Introduction" section, and the "Values and types" subsection of the "Basic concepts" section.

### DIFF
--- a/docs/lang/articles/reference.md
+++ b/docs/lang/articles/reference.md
@@ -8,9 +8,9 @@ This article describes the syntax and semantics of the Taichi programming
 language.
 
 **To users**: If you have gone through user tutorials and still feel uncertain
-about your program behavior, then you are coming to the correct place. If you
-find the actual behavior different from the one described in this article, feel
-free to create an [issue](https://github.com/taichi-dev/taichi/issues/new/choose).
+about your program behavior, then you are in the right place. If you find the
+actual behavior different from the one described in this article, feel free to
+create an [issue](https://github.com/taichi-dev/taichi/issues/new/choose).
 Anything unspecified in this article is subject to change, so you should not
 rely on it in your programs.
 
@@ -32,21 +32,21 @@ subset of that of Python. To make life easier, this article is modeled after
 the [Python language reference](https://docs.python.org/3/reference/). The
 [notation](https://docs.python.org/3/reference/introduction.html#notation) and
 [lexical analysis](https://docs.python.org/3/reference/lexical_analysis.html)
-parts exactly follow Python, so familiarize yourself with them if you are new
-to them.
+parts exactly follow Python. Please familiarize yourself with them if they seem
+new.
 
 ## Basic concepts
 
-Before detailing syntax and semantics in the next few chapters, many important
-basic concepts and general evaluation principles unique to Taichi are
-introduced here.
+Before detailing syntax and semantics in the next few chapters, many basic but
+important concepts and general evaluation principles specific to Taichi are
+listed here.
 
 ### Values and types
 
-Like other programming languages, each expression in Taichi will be evaluated
-to a value, and each value has a type. Because Taichi provides easy interaction
-with Python and [meta-programming](advanced/meta.md) support, there are
-actually two kinds of evaluation: *compile-time evaluation* and *runtime
+Like many other programming languages, each expression in Taichi will be
+evaluated to a value, and each value has a type. Because Taichi provides easy
+interaction with Python and [meta-programming](advanced/meta.md) support, there
+are actually two kinds of evaluation: *compile-time evaluation* and *runtime
 evaluation*. There are also two kinds of values: *Python values* and *Taichi
 values*.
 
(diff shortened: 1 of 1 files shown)


docs/lang/articles/reference.md
```diff
@@ -0,0 +1,172 @@
...
+sidebar_position: 9998
+---
+
+# Language Reference
+
+This article describes the syntax and semantics of the Taichi programming
+language.
+
+**To users**: If you have gone through user tutorials and still feel uncertain
+about your program behavior, then you are coming to the correct place. If you
```
k-ye, line 11:
```suggestion
about your program behavior, then you are in the right place. If you
```

docs/lang/articles/reference.md
```diff
@@ -0,0 +1,172 @@
...
+the Taichi language and the Python language - code in the Taichi scope is
+treated as the former, while code in the Python scope is treated as the latter.
+It should be emphasized that this article is about *the Taichi language*.
+
+That said, because Taichi is embedded in Python, the syntax of Taichi is a
+subset of that of Python. To make life easier, this article is modeled after
+the [Python language reference](https://docs.python.org/3/reference/). The
+[notation](https://docs.python.org/3/reference/introduction.html#notation) and
+[lexical analysis](https://docs.python.org/3/reference/lexical_analysis.html)
+parts exactly follow Python, so familiarize yourself with them if you are new
```
k-ye, line 35: Let's tune down a bit..

```suggestion
parts exactly follow Python, please familiarize yourself with them if they seem new
```

docs/lang/articles/reference.md
```diff
@@ -0,0 +1,172 @@
...
+to them.
+
+## Basic concepts
+
+Before detailing syntax and semantics in the next few chapters, many important
```
k-ye, line 40: many basic but important concepts and general evaluation principles specific to Taichi are listed here.

docs/lang/articles/reference.md
```diff
@@ -0,0 +1,172 @@
...
+basic concepts and general evaluation principles unique to Taichi are
+introduced here.
+
+### Values and types
+
+Like other programming languages, each expression in Taichi will be evaluated
```
k-ye, line 46:
```suggestion
Like many other programming languages, each expression in Taichi will be evaluated
```

docs/lang/articles/reference.md
```diff
@@ -0,0 +1,172 @@
...
+:::note
+For readers familiar with programming language terms, such behavior is inspired
+by [multi-stage programming](https://en.wikipedia.org/wiki/Multi-stage_programming)
+or [partial evaluation](https://en.wikipedia.org/wiki/Partial_evaluation).
+:::
+
+A Python value is simply a
+[Python object](https://docs.python.org/3/reference/datamodel.html),
+which directly comes from the following sources:
+- Free variables
```
k-ye, line 62: nit: I would put `literals` first because that is the simplest

docs/lang/articles/reference.md
```diff
@@ -0,0 +1,172 @@
...
+- Arguments passed via `ti.template()`
+- Literals
+
+Furthermore, as long as all the operands of an operation are Python values,
+compile-time evaluation will take place and a result Python value will be
```
k-ye, line 67:
```suggestion
compile-time evaluation will take place, producing a result Python value.
```

This needs update. If I call `foo(x)`, where both `foo()` and `x` are Python values (i.e., `foo()` is not a Taichi function), all operands are Python value, but I still need to use it inside `ti.static`?

docs/lang/articles/reference.md
```diff
@@ -0,0 +1,172 @@
...
+produced. For meta-programming purposes, Taichi provides an advanced
+environment for compile-time evaluation: `ti.static()`, where more operations
+are supported.
+
+A Python value only exists at compile time. After compile-time evaluation, all
```
k-ye, line 72: I feel like it's more like "transformed into Taichi values at JIT compile time"?

docs/lang/articles/reference.md
```diff
@@ -0,0 +1,172 @@
...
+the remaining expressions will be evaluated to Taichi values at runtime.
+
+A Taichi value has a Taichi type, which is one of the following:
```
k-ye, line 75: I think we need to revamp our type system. At the end of the day, we can say:

which is one of the following:

* A primitive type
* A compound type
* An Ndarray
* ...

You can find more details in Type System.

WDYT?

docs/lang/articles/reference.md
strongoier: No, you can directly use it.

docs/lang/articles/reference.md
strongoier: I understand your point. The accurate way to say this should be "transformed into literals of Taichi AST". However, introducing Taichi AST here may complicate things even further. My purpose is to let users directly think about the final execution behavior based on their source code, and the actual Taichi value is indeed created at runtime.

docs/lang/articles/reference.md
strongoier: I agree. They should be described in a unified way. My writing here is limited by the current situation :-(

k-ye (approved):
LGTM!

PR #4606, opened 2022-03-23
[lang] Replace keywords in python

Leonz5288:
Related issue = #


ailzhang (commented):
Can we prefix origin and replace as well just to avoid future headache :P ?

ailzhang (approved):
Thanks!

PR #4614, opened 2022-03-24
[vulkan] Choose better devices

bobcao3:
Now we will prefer a discrete GPU, and prefer higher version of Vulkan support when multiple GPUs are detected.

### DIFF
--- a/taichi/backends/vulkan/vulkan_device_creator.cpp
+++ b/taichi/backends/vulkan/vulkan_device_creator.cpp
@@ -156,16 +156,31 @@ VulkanQueueFamilyIndices find_queue_families(VkPhysicalDevice device,
   return indices;
 }
 
-bool is_device_suitable(VkPhysicalDevice device, VkSurfaceKHR surface) {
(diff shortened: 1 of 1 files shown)


taichi/backends/vulkan/vulkan_device_creator.cpp
```diff
@@ -156,16 +156,32 @@ VulkanQueueFamilyIndices find_queue_families(VkPhysicalDevice device,
   return indices;
 }
 
-bool is_device_suitable(VkPhysicalDevice device, VkSurfaceKHR surface) {
+size_t is_device_suitable(VkPhysicalDevice device, VkSurfaceKHR surface) {
```
ailzhang, line 159: can we rename this to a better name, e.g. `get_device_score`, it feels a bit strange to return a size_t for `is_device_suitable` :P

ailzhang (approved):
Thanks!! LGTM with a nit about naming :P !

qiao-bo (approved):

PR #4622, opened 2022-03-25
[Lang] [ir] Promote output type of log, exp, and sqrt ops

### LINKED ISSUE #4617: log(x) function automatically casts the result to integers
Example code:

```python
import taichi as ti
ti.init()

@ti.kernel
def test():
      v = ti.Vector([1, 2, 3])
      print(ti.log(v))

test()  # [0, 0, 1]
```
However, we wouldn't expect `log(2) = 0` here.  Can we always return floats here?

victoriacity:
Related issue = close #4617

When an integer is passed to a `ti.log`, `ti.exp`, or `ti.sqrt` function, the output of the function is counterintuitively cast back to an integer although the operand is implicitly cast to float before the calculated.

This PR changes the code behavior as follows:
```python
@ti.kernel
def test():
      v = 2
      u = ti.log(v)
      print(u)
test() # Before: 0; After: 0.693147
```


frostming:
Can you add a test case like in the PR content? Thanks.

strongoier:
Frontend type check should also be updated :-) https://github.com/taichi-dev/taichi/blob/2203b82b50e76915529ea40003a3a7a378bff9e0/taichi/ir/frontend_ir.cpp#L152

strongoier (approved):

PR #4627, opened 2022-03-25
[Lang] Add math module to support glsl-like functions

neozhaoliang:


python/taichi/__init__.py
```diff
@@ -38,9 +38,7 @@
     'imshow': 'tools.imshow',
     'imwrite': 'tools.imwrite',
     'quant': 'types.quantized_types.quant',
-    'type_factory': 'types.quantized_types.type_factory',
```
lin-hitonami, line 41: Why are these deleted? Something went wrong when merging?

PR #4628, opened 2022-03-25
[Lang] Add math module to support glsl-like functions

neozhaoliang:
Sorry for so many file changes, because I rebased my fork with the latest taichi-dev/master branch.


python/taichi/math/mathimpl.py
```diff
@@ -0,0 +1,401 @@
+"""
+Math functions for glsl-like functions and other stuff.
+"""
+import taichi as ti
+
+pi = 3.1415926535897932
+e = 2.718281828459045
```
frostming, line 7: may be better to use the constants from `math` module.

neozhaoliang: Sure, will update this later.

PR #4629, opened 2022-03-25
[Lang] Add vector swizzle feature to math module

TiGeekMan:
This PR adds a new feature "vector swizzle" to the `math` module. Thanks to @lin-hitonami 's great contribution!

The main code is under the `math` module. See the [_VectorType](https://github.com/TiGeekMan/taichi/blob/master/python/taichi/math/vectypes.py#L95) class.

## Feature description

With this feature, you can use glsl-style vectors, like:

+ **Initialize vectors in the glsl way**

```
v = ti.math.vec3(1, 2, 3)
w = ti.math.ivec3(1)  # ivec3(1, 1, 1)
```

+ **Use `rgba`, `xyzw`, `uvw` properties to get and set values**

```
v = ti.math.vec3(1)
w = v.bgr + 2  # [3, 3, 3]
w.xyz = [0, 0, 0]  # w is set to all zero
v = (w + vec3(1)).xyz  # [1, 1, 1]
```
Note operations between two vectors, or between a vector and a matrix/scalar will return vector type.

+ **Multiply a matrix**

```
M = ti.Matrix([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
v = ti.math.vec3(1, 2, 3)
w = (M @ v).xyz  # [1, 2, 3]
```
The vector must follow the matrix, not in the ahead of it. (unlike glsl here)

## Existing problem

Not working when using vector swizzles and inplace operators like  `__iadd__`  **simultaneously**:

```
v = ti.math.vec3(0)
v += 1  # OK, no swizzle and v = [1, 1, 1]

v.xyz += 1  # NOT WORK, v is unchanged
ti.atomic_add(v.xyz, 2)  # NOT WORK, v is unchanged
```

I guess the problem is due to I didn't handle the wrapper function well at [here](https://github.com/TiGeekMan/taichi/blob/master/python/taichi/math/vectypes.py#L95).

## Possible problematic code && Looking for suggestion/help

+ **A hack way to check if another operand is a vector**

In two places I have to check if an operand is a vector:
1. [here](https://github.com/TiGeekMan/taichi/blob/master/python/taichi/lang/matrix.py#L218) I want to make sure the multiplication of a matrix and a vector returns a vector. To avoid importing the hidden class `_VectorType` class (indeed it's deleted at the module level) a hack way `hasattr(other, "_DIM")` is used to check if it's a vector. Not satisfied with this solution. 
2. Also [here](https://github.com/TiGeekMan/taichi/blob/master/python/taichi/lang/impl.py#L39) for the same reason.

+ **Make vector swizzle and inplace operations work together**

`_VectorType` inherits the `Matrix` class and together with its operations, so some operations between vectors will return `ti.Matrix`. Some of them are selected out [here](https://github.com/TiGeekMan/taichi/blob/master/python/taichi/lang/impl.py#L39) and changed their return type by hand.
But this does not work for inplace operators. I'm still trying to figure out how to fix this.


lin-hitonami (approved):
LGTM!

PR #4630, opened 2022-03-25
[gui] Allow to configure the texture data type

ghuau-innopeak:
With AOT modules depending on the device where the kernel is running, we
might support RGBA Float Texture and it's better to pass the data from
the Taichi Compute Kernel directly to the image (normalized between 0
and 1) instead of adding extra step to convert to 0;255

Related issue = #

### DIFF
--- a/taichi/ui/backends/vulkan/renderables/set_image.cpp
+++ b/taichi/ui/backends/vulkan/renderables/set_image.cpp
@@ -37,6 +37,16 @@ void SetImage::update_data(const SetImageInfo &info) {
 
   const FieldInfo &img = info.img;
 
+  // Support configuring the internal image based on the data type of the field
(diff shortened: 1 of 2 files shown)


AmesingFlank (approved):
lgtm

k-ye:
Hmm, weird that all CI jobs are gone in this PR. Could be related to the Github Action maintenance yesterday. Could you re-push the commit and see if that fixes the problem?

ghuau-innopeak:
> Hmm, weird that all CI jobs are gone in this PR. Could be related to the Github Action maintenance yesterday. Could you re-push the commit and see if that fixes the problem?

Just did, we'll see if that works but yeah I did a quick push to remove a duplicated comment due to a merge conflict, and it was at the same time as a Github Maintenance as well, hopefully it will be resolved now.

PR #4643, opened 2022-03-27
[simt] Subgroup reduction primitives

bobcao3:
Related issue = #4631

Also added tests for subgroup reduction. All the SPIR-V reductions are fully works in Vulkan backend now, and tested on M1 macs.


turbo0628 (approved):
LGTM

PR #4644, opened 2022-03-28
[Doc] Update docstring for module misc

neozhaoliang:


python/taichi/lang/impl.py
```diff
@@ -367,13 +367,17 @@ def reset():
 
 @taichi_scope
 def static_print(*args, __p=print, **kwargs):
+    """The print function in Taichi scope. This function is called
```
k-ye, line 370: Note that docstring usually starts with a one line short-line description

python/taichi/ad.py
```diff
@@ -1,3 +1,7 @@
+"""Taichi automatic differentiation module.
```
k-ye, line 1: Here and below: there's a new line after the first short description. See https://github.com/google/styleguide/blob/gh-pages/pyguide.md#382-modules

python/taichi/aot/__init__.py
```diff
@@ -1,2 +1,6 @@
+"""Taichi's AOT (ahead of time) module. Users can use Taichi
```
k-ye, line 1:
```suggestion
"""Taichi's AOT (ahead of time) module. 

Users can use Taichi ...
```

python/taichi/aot/record.py
```diff
@@ -18,10 +18,39 @@ def record_action_config(key, value):
 
 
 def start_recording(filename):
+    """Start recording kernel information to a `yml` file.
```
k-ye, line 21: Here and below: All the verbs should be in third person.

```suggestion
    """Starts recording kernel information to a `yml` file.
```

k-ye (approved):
I got a few nits, but thanks for filling all the docstrings, LGTM!

python/taichi/lang/impl.py
neozhaoliang: Sure. fixed.

k-ye (approved):
lgtm!

PR #4645, opened 2022-03-28
[ci] Add libtaichi_export_core build for desktop in CI

ailzhang:
Related issue = #


ailzhang:
not sure why self hosted macos has the ANDROID_NDK_HOME env set... will take a look tmr.

PR #4650, opened 2022-03-29
[dx11] Constant buffer binding and AtomicIncrement in RAND_STATE

quadpixels:
Related issue = https://github.com/taichi-dev/taichi/issues/992

This makes a bit more progress for the dx11 backend, it now runs the mpm88.py program. This change has to deal with a few updates in the Taichi infrastructure:

- SPIRV-cross now differentiates read-only buffers (`ResourceBinder::buffer`) from read-write buffers (`ResourceBinder::rw_buffer`) differently, so the backend has to handle Constant Buffers in addition to Unordered Access Views.

- SPIRV-cross now uses a different set of default options, so the backend needs to call `remap_num_workgroups_builtin`.

- Using `spv::Op::OpAtomicIIncrement` for the random number generator state increment, so that the generated HLSL will not cause the HLSL compiler to generate an error

Tested: runs mpm88.py


quadpixels:
I just found `SPIRV_Cross_NumWorkgroups` takes up constant buffer slot 1 (while args_t takes slot 0)

The slot is not shown in the HLSL source (if it shows up, it will be `b1`)
```
struct SPIRV_Cross_Input
{
    uint3 gl_GlobalInvocationID : SV_DispatchThreadID;
};
```

To compare this is what slot 0 looks like:
```
cbuffer args_t : register(b0)
{
    int args_arg0 : packoffset(c0);
    int args_extra_args0 : packoffset(c0.y);
    ...
    int args_extra_args127 : packoffset(c32);
};
```

taichi/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -1168,17 +1169,31 @@ void IRBuilder::init_random_function(Value global_tmp_) {
...
-  ib_.begin(spv::OpIAdd)
-      .add_seq(t_uint32_, tmp10, tmp9, _1)
-      .commit(&func_header_);
-  store_var(rand_gtmp_, tmp10);
+
+  enum spv::Op add_op = spv::OpIAdd;
+
+  // use atomic iadd for DX API otherwise the HLSL compiler throws error:
+  // error X3694: race condition writing to shared resource detected
+  if (dynamic_cast<const taichi::lang::directx11::Dx11Device *>(device_)) {
```
bobcao3, line 1177: Some how I remember fixing this problem before but don't know why it is reverted later. My bad. This should work for now!

bobcao3 (approved):
LGTM!

taichi/codegen/spirv/spirv_ir_builder.cpp
bobcao3: Should guard this if statement under `#ifdef TI_WITH_DX11` as well

taichi/backends/dx/dx_device.cpp
```diff
@@ -68,11 +83,40 @@ Dx11CommandList::~Dx11CommandList() {
...
+    cmd->binding = binding;
+    cmd->uav = uav;
+    recorded_commands_.push_back(std::move(cmd));
+  }
+
+  // CBV
+  for (auto &[binding, alloc_id] : binder->cb_binding_to_alloc_id()) {
+    std::unique_ptr<CmdBindConstantBufferToIndex> cmd =
+        std::make_unique<CmdBindConstantBufferToIndex>(this);
+    cmd->desc_ = fmt::format("bind CBV {} to slot b{}", alloc_id, binding);
```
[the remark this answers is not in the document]
quadpixels: I see, it was for temporary debugging and I removed it now.

taichi/codegen/spirv/spirv_ir_builder.cpp
quadpixels: Done, no problem it gave me a good opportunity to learn a bit more about the codegen :D

PR #4657, opened 2022-03-29
[refactor] Remove Expression::serialize and add ExpressionHumanFriendlyPrinter

PGZXB:
Related issue = #

At present, `Expression::serialize` is only for generating human-friendly string.

1. Remove `Expression::serialize` and add `ExpressionHumanFriendlyPrinter`
2. Prepare for adding `OfflineCacheKeyGenerator` instead of using `IRPrinter` to generate cache key


strongoier (approved):
Awesome!

PR #4658, opened 2022-03-29
[doc] Taichi spec: compound statements, if, while

lin-hitonami:
Related issue = #4602


docs/lang/articles/reference.md
```diff
@@ -250,8 +250,89 @@ with `expression2` as the error message.
 
 ## Compound statements
 
+Compound statements contain (groups of) other statements;
+they affect or control the execution of those other statements in some way.
+In general, compound statements span multiple lines,
+although in simple incarnations a whole compound statement may be contained in one line.
```
k-ye, line 256: I know this is from https://docs.python.org/3/reference/compound_stmts.html#compound-statements, but I don't get why incarnations is used here (a sincere question). To me it seems this is just "in simple situations". cc @Vissidarte-Herman

docs/lang/articles/reference.md
```diff
@@ -250,8 +250,89 @@ with `expression2` as the error message.
...
+compound_stmt ::=  if_stmt
+                   | while_stmt
+                   | for_stmt
+
+suite         ::=  stmt_list NEWLINE | NEWLINE INDENT statement+ DEDENT
+statement     ::=  stmt_list NEWLINE | compound_stmt
+stmt_list     ::=  simple_stmt (";" simple_stmt)* [";"]
+```
+
+Note that statements always end in a NEWLINE possibly followed by a DEDENT.
```
k-ye, line 294:
```suggestion
Note that statements always end in a NEWLINE, possibly followed by a DEDENT.
```

k-ye (commented):
If most of these article is from https://docs.python.org/3/reference/compound_stmts.html#compound-statements, can we just refer the readers to the Python reference? cc @strongoier

strongoier:
> If most of these article is from https://docs.python.org/3/reference/compound_stmts.html#compound-statements, can we just refer the readers to the Python reference? cc @strongoier

I agree. We can see how https://pytorch.org/docs/stable/jit_language_reference_v2.html#compound-statements handles this. Our main purpose is to explain the differences (e.g. Is there a compile-time evaluation? Is there a Python value vs Taichi value problem? Is there any special behavior with `ti.static`?).

ailzhang (approved):
LGTM!

docs/lang/articles/reference.md
```diff
@@ -250,8 +250,56 @@ with `expression2` as the error message.
 
 ## Compound statements
 
+This section describes the supported syntax of compound statements in Taichi,
+and it is modeled after [the compound statement chapter of The Python Language Reference](https://docs.python.org/3/reference/compound_stmts.html)
+
+A compound statement consists of one or more ‘clauses’.
```
strongoier, line 256: For writing consistency of this article, let's use *clauses* instead of ‘clauses’.

docs/lang/articles/reference.md
```diff
@@ -250,8 +250,56 @@ with `expression2` as the error message.
...
+A clause consists of a header and a ‘suite’.
+The clause headers of a particular compound statement are all at the same indentation level.
+Each clause header begins with a uniquely identifying keyword and ends with a colon.
+A suite is a group of statements controlled by a clause.
+
```
strongoier, line 261: We can put the following here for completeness:
```
compound_stmt ::= if_stmt | while_stmt | for_stmt
suite         ::= stmt_list NEWLINE | NEWLINE INDENT statement+ DEDENT
statement     ::= stmt_list NEWLINE | compound_stmt
stmt_list     ::= simple_stmt (";" simple_stmt)* [";"]
```

docs/lang/articles/reference.md
```diff
@@ -250,8 +250,56 @@ with `expression2` as the error message.
...
+The difference between the compound statements in Taichi and Python is that Taichi introduces
+compile time evaluation. If the expression in the clause header is a static expression,
+Taichi changes the contents of the compound statement at compile time
```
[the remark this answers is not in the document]
strongoier: replaces the compound statement

docs/lang/articles/reference.md
```diff
@@ -250,8 +250,56 @@ with `expression2` as the error message.
...
+according to the evaluation result of the expression.
+
 ### The `if` statement
 
+The if statement is used for conditional execution:
```
strongoier, line 269:
```suggestion
The `if` statement is used for conditional execution:
```

docs/lang/articles/reference.md
```diff
@@ -250,8 +250,56 @@ with `expression2` as the error message.
...
+```
+if_stmt ::=  "if" (static_expression | assignment_expression) ":" suite
+             ("elif" (static_expression | assignment_expression) ":" suite)*
+             ["else" ":" suite]
+```
+
+It selects exactly one of the suites by evaluating the expressions one by one until one is found to be true
+(see section Boolean operations for the definition of true and false);
```
strongoier, line 277: Add a link to "Boolean operations"?

docs/lang/articles/reference.md
```diff
@@ -250,8 +250,56 @@ with `expression2` as the error message.
 
 ## Compound statements
 
+This section describes the supported syntax of compound statements in Taichi,
+and it is modeled after [the compound statement chapter of The Python Language Reference](https://docs.python.org/3/reference/compound_stmts.html)
```
strongoier, line 254: The "model after" part is already stated in introduction. We can directly say "This section explains the syntax and semantics of compound statements in Taichi".

docs/lang/articles/reference.md
```diff
@@ -250,8 +250,61 @@ with `expression2` as the error message.
...
+
+A *clause* in the `if` statement whose expression is a static expression is called a static `if` *clause*.
+
+The expression of a static `if` *clause* is evaluated at compile time, and it must be a compile-time constant.
+
+The static `if` *clauses* replaces the *clauses* in the compound statement at compile time as below.
+The expressions of static `if` *clauses* are evaluated one by one at compile time until one is found to be true;
+then all *clauses* after that *clause* are removed from the compound statement,
+and that *clause* becomes the last *clause* of the compound statement, with the expression set to `True`.
+The static `if` *clauses* whose expressions are found to be false are removed from the compound statement.
```
strongoier, line 294: This part is a bit heavy. Can we directly state "elif" as a syntax sugar for "else if" and then only describe static/non-static if/else?

lin-hitonami: But python explains `if` statements in this way. If we say `elif` is a syntax sugar for `else if`, we need to rewrite the explanation of normal `if` as well. Also, I think explaining the syntax sugar will make the whole section even longer.

strongoier: I'm proposing this mainly for `static-if`s, because a `static-if` can be easily explained when only a single `if-else` is considered: replace according to `true/false` conditions.

The current explanation is a bit confusing: `and that clause becomes the last clause of the compound statement, with the expression set to True.` It doesn't show the "replace" behavior and doesn't talk about the clauses before that clause.

strongoier (approved):
LGTM!

PR #4663, opened 2022-03-30
[doc] Taichi spec: calls, unary ops, binary ops and comparison

squarefk:
Related issue = #4602

### DIFF
--- a/docs/lang/articles/reference.md
+++ b/docs/lang/articles/reference.md
@@ -366,9 +366,7 @@ are required to have integer types.
 comparison    ::= or_expr (comp_operator or_expr)*
 comp_operator ::= "<" | ">" | "==" | ">=" | "<=" | "!=" | ["not"] "in"
 ```
-A comparison yields a boolean value (True or False), or if one of the operands is a Tensor, a boolean Tensor. Comparisons can be chained arbitrarily as long as they do not yield boolean Tensors that have more than one element. a op1 b op2 c ... is equivalent to a op1 b and b op2 c and ....
-
-All operands need to be primaries. The return type is `i32` for all comparision operations.
+Comparisons can be chained arbitrarily, e.g., `x < y <= z` is equivalent to `(x < y) & (y <= z)`.
 
 #### Value comparisons
 
--- a/docs/lang/articles/reference.md
+++ b/docs/lang/articles/reference.md
@@ -266,7 +266,7 @@ args          ::=  [arg (',' arg)*]
 arg           ::=  identifier
 ```
 
-To favour simplicity, Taichi language doesn't support keyword arguments like Python.
+To favor simplicity, Taichi language doesn't support keyword arguments like Python.
 
 ### The power operator
 
@@ -275,7 +275,7 @@ power ::= 'ti.pow(' primary ',' primary ')' | primary ['**' u_expr]
 ```
 
 Taichi predefines its own exponentiation function 'ti.pow()'. Scalars are broadcast in the case of scalar-tensor/tensor-scalar exponentiation operations, and tensor-tensor exponentiation is done elementwise without any broadcasting.
-Applying Python '**' operator inside Taichi scope is equivalent. In such case, the power operator binds more tightly than unary operators on the left, but less tightly than unary operators on the right; i.e. -3 ** -2 == -(3 ** (-2)).
+Applying Python '**' operator inside Taichi scope is equivalent. In such cases, the power operator binds more tightly than unary operators on the left, but less tightly than unary operators on the right; i.e. -3 ** -2 == -(3 ** (-2)).
 
 ### Unary arithmetic and bitwise operations
 
@@ -323,7 +323,7 @@ The operators <, >, ==, >=, <=, and != compare the values of two objects. The tw
 
 #### Membership test operations
 
-List, dicts and tuples types along with 'in' operations are not supported in Taichi scope. The only execption is to put the whole expression into 'ti.static()' which degenerates into Python expression. More details refer to Python's specification.
+List, dict and tuple types along with 'in' operations, are not supported in Taichi scope. The only exception is to put the whole expression into 'ti.static()' which degenerates into Python expression. More details refer to Python's specifications.
 
 #### Identity comparisons
 
--- a/docs/lang/articles/reference.md
+++ b/docs/lang/articles/reference.md
@@ -260,10 +260,10 @@ stride) inside have to be evaluated to Python values.
 #### Calls
 
 ```
-call          ::=  primary '(' argument_list ')'
-argument_list ::=  args
-args          ::=  [arg (',' arg)*]
-arg           ::=  identifier
+call          ::= primary "(" argument_list ")"
+argument_list ::= args
+args          ::= [arg ("," arg)*]
+arg           ::= identifier
 ```
 
 To favor simplicity, Taichi language doesn't support keyword arguments like Python.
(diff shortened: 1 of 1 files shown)


docs/lang/articles/reference.md
```diff
@@ -176,22 +176,72 @@ compile time.
 
 #### Calls
 
+```
```
strongoier, line 179: 1. Let's use single space and "" for consistence.
2. We support positional arguments as shown in https://docs.python.org/3/reference/expressions.html#calls, so the rules need to be updated.
3. Our call implementation (https://github.com/taichi-dev/taichi/blob/0e715318cb0b6a19fd2313e930114d63e4960283/python/taichi/lang/ast/ast_transformer.py#L409) includes many cases, so a more detailed description is need:
    - Calling a Python function
    - Calling a Taichi function
    - Calling a Taichi builtin function
    - Calling a type annotation for literals

docs/lang/articles/reference.md
```diff
@@ -176,22 +176,72 @@ compile time.
...
+args          ::=  [arg (',' arg)*]
+arg           ::=  identifier
+```
+
+To favor simplicity, Taichi language doesn't support keyword arguments like Python.
+
 ### The power operator
 
+```
+power ::= 'ti.pow(' primary ',' primary ')' | primary ['**' u_expr]
```
strongoier, line 191: `ti.pow` belongs to Taichi's builtin functions, so it doesn't have to be listed as a syntax rule here.

docs/lang/articles/reference.md
```diff
@@ -176,22 +176,72 @@ compile time.
...
+```
+
+Taichi predefines its own exponentiation function 'ti.pow()'. Scalars are broadcast in the case of scalar-tensor/tensor-scalar exponentiation operations, and tensor-tensor exponentiation is done elementwise without any broadcasting.
```
strongoier, line 194: It seems that scalar-tensor/tensor-scalar/tensor-tensor evaluation rules are described multiple times below. Shall we systematically describe it in a new subsection under `Basic concepts`?

docs/lang/articles/reference.md
```diff
@@ -176,22 +176,72 @@ compile time.
...
 ### Unary arithmetic and bitwise operations
 
+```
+u_expr ::=  power | '-' power | '~' power
+```
+The unary - operator yields the negation of its argument. The unary ~ operator yields the bitwise inversion of its argument. - can be used with all scalar and tensor. ~ can only be used with integer scalar (i32, i64, etc.) and interrelated tensor.
+
 ### Binary arithmetic operations
 
+```
```
strongoier, line 206: Let's add more information regarding type check. See https://github.com/taichi-dev/taichi/blob/0e715318cb0b6a19fd2313e930114d63e4960283/taichi/ir/frontend_ir.cpp#L171

docs/lang/articles/reference.md
```diff
@@ -259,22 +263,98 @@ stride) inside have to be evaluated to Python values.
...
+
 ### Comparisons
 
+```
+comparison    ::= or_expr (comp_operator or_expr)*
+comp_operator ::= "<" | ">" | "==" | ">=" | "<=" | "!=" | "is" ["not"] | ["not"] "in"
+```
+A comparison yields a boolean value (True or False), or if one of the operands is a Tensor, a boolean Tensor. Comparisons can be chained arbitrarily as long as they do not yield boolean Tensors that have more than one element. a op1 b op2 c ... is equivalent to a op1 b and b op2 c and ....
+
+All operands need to be primaries. The return type is `i32` for all comparision operations.
```
strongoier, line 346: Comparisons can be chained arbitrarily, e.g., `x < y <= z` is equivalent to `(x < y) & (y <= z)`.

strongoier, line 346: Comparisons can be chained arbitrarily, e.g., `x < y <= z` is equivalent to `(x < y) & (y <= z)`.

docs/lang/articles/reference.md
```diff
@@ -259,22 +263,98 @@ stride) inside have to be evaluated to Python values.
...
+
 #### Value comparisons
 
+The operators <, >, ==, >=, <=, and != compare the values of two objects. The two objects generally need to be of the same type, unless there is an implicit type conversion available between the objects. Built-in Python types like strings, lists, tuples are not supported. They need to be converted into Taichi scalar or tensor in advance.
```
strongoier, line 350:
```suggestion
See [common rules for binary operations](#common-rules-of-binary-operations),
[implicit type casting in binary operations](basic/type.md#implicit-type-casting-in-binary-operations),
and [comparison operators](basic/operator.md#comparison-operators).
```

strongoier, line 350: See [common rules for binary operations](#common-rules-of-binary-operations),
[implicit type casting in binary operations](basic/type.md#implicit-type-casting-in-binary-operations),
and [comparison operators](basic/operator.md#comparison-operators).

docs/lang/articles/reference.md
```diff
@@ -259,22 +263,98 @@ stride) inside have to be evaluated to Python values.
...
+
 #### Membership test operations
 
+List, dict and tuple types along with 'in' operations, are not supported in Taichi scope. The only exception is to put the whole expression into 'ti.static()' which degenerates into Python expression. More details refer to Python's specifications.
```
strongoier, line 354:
```suggestion
The semantics of membership test operations follow
[Python](https://docs.python.org/3/reference/expressions.html#membership-test-operations),
but they are only supported in [static expressions](#static-expressions).
```

docs/lang/articles/reference.md
```diff
@@ -259,22 +263,98 @@ stride) inside have to be evaluated to Python values.
...
+
+```
+
+To favor simplicity, starred arguments and keyword arguments are not supported for Taichi kernels.
+
+When calling a Python function, it needs to be wrapped by `ti.static()`.
+
+The following Taichi builtin functions are supported: `print()`, `min()`, `max()`, `int()`, `float()`, `any()`, `all()`, `abs()`, `pow()`.
+
+Calling a type annotation for literals
```
strongoier, line 289:
```suggestion
The `primary` must be evaluated to one of:
- A [Taichi function](basic/syntax.md#taichi-function).
- A [Taichi builtin function](basic/operator.md#other-arithmetic-functions).
- A Taichi primitive type, which serves as a type annotation for a literal. In this case, the `positional_arguments` must be evaluated to a single Python value, and the Python value will be turned into a Taichi value with that annotated type. 
- A Python callable object. If not inside a [static expression](#static-expressions), a warning is produced.
```

docs/lang/articles/reference.md
```diff
@@ -259,22 +278,107 @@ stride) inside have to be evaluated to Python values.
...
+positional_arguments ::= positional_item ("," positional_item)*
+positional_item      ::= assignment_expression | "*" expression
+starred_and_keywords ::= ("*" expression | keyword_item)
+                         ("," "*" expression | "," keyword_item)*
+keywords_arguments   ::= (keyword_item | "**" expression)
+                         ("," keyword_item | "," "**" expression)*
+keyword_item         ::= identifier "=" expression
+
+
+```
```
strongoier, line 296:
```suggestion
call                 ::= primary "(" [positional_arguments] ")"
positional_arguments ::= positional_item ("," positional_item)*
positional_item      ::= assignment_expression | "*" expression
```

strongoier (approved):

Issue #4659, opened 2022-03-29
`ti.atomic_max` reduction seems to use `f64` instead of `f32`


RuRo:
Consider the following code:
```python
import taichi as ti
    
ti.init(
    ti.gpu,
    print_kernel_nvptx=True,
    print_kernel_llvm_ir_optimized=True,
)
    
@ti.kernel
def reduce():
    for I in ti.grouped(f):
        ti.atomic_max(r[None], f[I])
    
f = ti.field(ti.f32, (2 ** 22,))
r = ti.field(ti.f32, ())
    
reduce()
```

Inspecting the generated `taichi_kernel_cuda_llvm_ir_optimized_0001.ll` and `taichi_kernel_nvptx_0001.ptx` we can see the following lines in the `reduce` function implementation:
```ll
define void @reduce_c56_0_kernel_0_range_for(%struct.RuntimeContext* nocapture readonly byval %context) local_unnamed_addr #3 {
  ; <SNIP>
13:                                               ; preds = %13, %.lr.ph.i
  %lsr.iv = phi i32 [ %lsr.iv.next, %13 ], [ %11, %.lr.ph.i ]
  %.03.i = phi i32 [ %4, %.lr.ph.i ], [ %20, %13 ]
  %.sroa.0.02.i = phi float [ 0x3810000000000000, %.lr.ph.i ], [ %19, %13 ]
  %14 = and i32 %lsr.iv, 16777212
  %15 = zext i32 %14 to i64
  %16 = getelementptr inbounds i8, i8* %8, i64 %15
  %17 = bitcast i8* %16 to float*
  %18 = tail call float @llvm.nvvm.ldg.global.f.f32.p0f32(float* %17, i32 4) #6
  %19 = tail call float @llvm.maxnum.f32(float %.sroa.0.02.i, float %18) #6     ; <--- (4)
  %20 = add nuw nsw i32 %.03.i, %10
  %lsr.iv.next = add i32 %lsr.iv, %12
  %21 = icmp ult i32 %20, 4194304
  br i1 %21, label %13, label %._crit_edge.i

._crit_edge.i:                                    ; preds = %13, %entry
  %.sroa.0.0.lcssa.i = phi float [ 0x3810000000000000, %entry ], [ %19, %13 ]
  %22 = tail call float @llvm.nvvm.shfl.sync.down.f32(i32 -1, float %.sroa.0.0.lcssa.i, i32 16, i32 31) #6
[...]

PR #4664, opened 2022-03-30
[CUDA] Fix the fp32 to fp64 promotion due to incorrect fmax/fmin call

turbo0628:
Related issue = #4659

The precision promotion in fp32 min/max reduction is caused by the `fmax` func which actually has fp64 type. [Reference](https://en.cppreference.com/w/c/numeric/math/fmax)

Fix by changing fmin / fmax to fminf / fmaxf.


ailzhang (approved):
Thanks!!

PR #4665, opened 2022-03-30
[Doc] Taichi spec: list and dictionary displays

strongoier:
Related issue = #4602

The three subsections, `Displays for lists and dictionaries`, `List displays` and `Dictionary displays` in the original plan are merged into one `List and dictionary displays` subsection because we don't have to repeat many parts same as Python.

### DIFF
--- a/docs/lang/articles/reference.md
+++ b/docs/lang/articles/reference.md
@@ -171,7 +171,7 @@ display is one of:
 compute the container items.
 
 ```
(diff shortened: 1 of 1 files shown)


k-ye (approved):
Thanks! WDYT if we also provide a small code snippet?

strongoier:
> Thanks! WDYT if we also provide a small code snippet?

SG! Updated.

PR #4666, opened 2022-03-30
[Doc] Taichi spec: Identity comparisons; Boolean operations

mzmzm:
Related issue = #4602


docs/lang/articles/reference.md
```diff
@@ -188,8 +188,29 @@ Literals are evaluated to Python values at compile time.
...
+`x is not y` yields the inverse truth value.
+
+:::note
+Operator `is` or `is not` in Taichi scope is deprecated. Please avoid using it.
+:::
+
 ### Boolean operations
 
+```
+or_test  ::=  and_test | or_test "or" and_test
```
strongoier, line 203:
```suggestion
or_test  ::= and_test | or_test "or" and_test
```
Let's use one space.

docs/lang/articles/reference.md
```diff
@@ -188,8 +188,29 @@ Literals are evaluated to Python values at compile time.
...
+and_test ::=  not_test | and_test "and" not_test
+not_test ::=  comparison | "not" not_test
+```
+
+User-defined objects can customize their conversion to bool by implementing a `__bool__` method.
+The operator not yields `True` if its operand is false, `False` otherwise.
+The expression `x and y` first evaluates `x`; if it is `False`, its value (`False`) is returned; otherwise, `y` is evaluated and its value is returned (`False` or `True`).
+The expression `x or y` first evaluates `x`; if it is `True`, its value (`True`) is returned; otherwise, `y` is evaluated and its value is returned (`False` or `True`).
+Currently, the type of result value is i32 (Taichi dose not have bool type).
```
strongoier, line 212: We don't have to repeat Python behavior description for Python values. We can say `When all operands of the operator are evaluated to Python values, the evaluation rule of the operator follows Python<link to Python ref here...>.`

Then we need to clarify Taichi behavior in more detail:
- Different behavior with `short_circuit=True/False`
- Operand type requirements (int, Matrix, ...?)
- How does Taichi perform bool operation without bool type?

docs/lang/articles/reference.md
```diff
@@ -188,8 +188,29 @@ Literals are evaluated to Python values at compile time.
 
 #### Identity comparisons
 
+The operators `is` and `is not` test for an object’s identity:
+`x is y` is `true` if and only if `x` and `y` are the same object.
+An Object’s identity is determined using the [id()](https://docs.python.org/3/library/functions.html#id) function.
+`x is not y` yields the inverse truth value.
+
+:::note
+Operator `is` or `is not` in Taichi scope is deprecated. Please avoid using it.
```
strongoier, line 197: We don't have to document deprecated operators. What about simply removing this subsection?

docs/lang/articles/reference.md
mzmzm: fix

docs/lang/articles/reference.md
mzmzm: remove

docs/lang/articles/reference.md
mzmzm: fix

docs/lang/articles/reference.md
```diff
@@ -192,10 +192,21 @@ compile time.
...
+```
+or_test  ::= and_test | or_test "or" and_test
+and_test ::= not_test | and_test "and" not_test
+not_test ::= comparison | "not" not_test
+```
+
+When all operands of the operator are evaluated to Python values,
+the evaluation rule of the operator follows [Python](https://docs.python.org/3/reference/expressions.html#boolean-operations)
+Otherwise, there are different behaviours:
+- There is no short circuit by default in Taichi. It means `x and y` will evaluate both the expression of `x` and `y` rather than only evaluates `x` when `x` is `False`. It also means `x or y` will evaluate both the expression of `x` and `y` rather than only evaluates `x` when `x` is `True`.
```
strongoier, line 206: We need also introduce the case where `short_circuit=True`.

docs/lang/articles/reference.md
```diff
@@ -192,10 +192,21 @@ compile time.
...
+- The type of operator can be `int` and `Matrix`
+- Currently, the type of result value is i32 (Taichi dose not have bool type). So `and` will be `bit-and`, and `or` will be `bit-or`.
```
strongoier, line 208: We need to define what is True/False when we use i32 in place of bool.

PR #4669, opened 2022-03-30
[misc] Temporarily disable a flaky test

strongoier:
https://github.com/taichi-dev/taichi/runs/5750005177?check_suite_focus=true shows that `test_closing_offline_cache` fails on Python 3.10.


qiao-bo (approved):
Thanks!

PR #4680, opened 2022-03-31
[Doc] Taichi spec: conditional expressions and simple statements

mzmzm:
Related issue = #4602

### DIFF
--- a/docs/lang/articles/reference.md
+++ b/docs/lang/articles/reference.md
@@ -207,6 +207,7 @@ Conditional expressions (sometimes called a “ternary operator”) have the low
 
 The expression `x if C else y` first evaluates the condition, `C` rather than `x`.
 If `C` is `true`, `x` is evaluated and its value is returned; otherwise,`y` is evaluated and its value is returned.
+In Taichi, `or_test` in conditional expression will be expanded to `or_test != 0` automatically.
 
 ### Expression lists
 
--- a/docs/lang/articles/reference.md
+++ b/docs/lang/articles/reference.md
@@ -198,6 +198,16 @@ compile time.
 
 ### Conditional expressions
 
+```
+conditional_expression ::=  or_test ["if" or_test "else" expression]
+expression             ::=  conditional_expression | lambda_expr
+```
+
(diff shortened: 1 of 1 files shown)


docs/lang/articles/reference.md
```diff
@@ -198,6 +198,17 @@ compile time.
...
 
+```
+conditional_expression ::=  or_test ["if" or_test "else" expression]
+expression             ::=  conditional_expression | lambda_expr
+```
+
+Conditional expressions (sometimes called a “ternary operator”) have the lowest priority of all Python operations.
+
+The expression `x if C else y` first evaluates the condition, `C` rather than `x`.
+If `C` is `true`, `x` is evaluated and its value is returned; otherwise,`y` is evaluated and its value is returned.
```
strongoier, line 209: We can refer to the `boolean expression` subsection for definition of true/false.

docs/lang/articles/reference.md
```diff
@@ -214,14 +225,70 @@ is evaluated to the value of that expression.
...
+target_list     ::=  target ("," target)* [","]
+target          ::=  identifier
+                     | "(" [target_list] ")"
+                     | "[" [target_list] "]"
+                     | attributeref
+                     | subscription
+                     | "*" target
+```
+
+An assignment statement evaluates the expression list (remember that this can be a single expression or a comma-separated list,
```
strongoier, line 268: We need to explain how a Python value becomes a Taichi value through assignment statement.

docs/lang/articles/reference.md
```diff
@@ -214,14 +225,70 @@ is evaluated to the value of that expression.
...
+In Taichi, some augmented assignments (e.g., `x[i] += 1`) are automatically atomic.
+See [supported-atomic-operations](https://docs.taichi.graphics/lang/articles/operator#supported-atomic-operations) for more details.
+
 #### Annotated assignment statements
 
+```
+annotated_assignment_stmt ::= augtarget ":" expression
+                              ["=" (starred_expression)]
+```
+
```
strongoier, line 289: We need to explain the behavior of annotated assignment: type casting.

PR #4683, opened 2022-03-31
[Lang] Add math module to support glsl-style functions

zemora:


k-ye (approved):
Fantastic LGTM!

python/taichi/math/mathimpl.py
```diff
@@ -0,0 +1,399 @@
+"""
+Math functions for glsl-like functions and other stuff.
+"""
+from math import e, pi
+
+import taichi as ti
+
+
+@ti.func
+def mix(x, y, a):
```
YuCrazing, line 10:
```suggestion
def lerp(x, y, a):
```

YuCrazing: I believe that `lerp` is a more common name. :)

YuCrazing (changes requested):

python/taichi/math/mathimpl.py
neozhaoliang: I guess `lerp` is used mostly for HLSL users? It's not in the GLSL spec: 
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/

Issue #4449, opened 2022-03-04
No iterations in a mixed forward, reverse ndrange


mackrol:
Mixed forward, reverse ndrange yields no iterations regardless of selected architecture
```py
import taichi as ti
ti.init(arch = ti.gpu)

@ti.kernel
def ndrange_test(i1 : ti.i32, i2 : ti.i32, j1 : ti.i32, j2 : ti.i32) -> ti.i32:
    n : ti.i32 = 0
    for _ in range(1): # one thread
        for i, j in ti.ndrange((i1, i2), (j1, j2)):
            n += 1
    return n

print('fwd, fwd:', ndrange_test( 0, 10,  0, 20))
print('fwd, rev:', ndrange_test( 0, 10, 20,  0))
print('rev, fwd:', ndrange_test(10,  0,  0, 20))
print('rev, rev:', ndrange_test(10,  0, 20,  0))
```
gives:
```
[Taichi] version 0.8.11, latest version 0.9.0, llvm 10.0.0, commit 9d0973e5, win, python 3.9.5
[Taichi] Starting on arch=cuda
fwd, fwd: 200
fwd, rev: 0
rev, fwd: 0
rev, rev: 200
```

mackrol:
Results are the same in version 0.9.0
```
[Taichi] version 0.9.0, llvm 10.0.0, commit 0c432a37, win, python 3.9.5
[Taichi] Starting on arch=cuda
fwd, fwd: 200
fwd, rev: 0
rev, fwd: 0
rev, rev: 200
```

qiao-bo:
cc @lin-hitonami

lin-hitonami:
Currently, Taichi does not support reversed range/ndrange fors. `rev, rev` in your program works just by coincident. We will consider adding reversed range/ndrange for support in the future. I will add an error message when there are reversed fors for now.

mackrol:
Yes, I consider `rev, rev` as a bug, it should yield no iterations just like python's `range`

PR #4690, opened 2022-03-31
[bug] Fix ndrange when start > end

lin-hitonami:
Related issue = #4449


ailzhang (approved):
Thanks!! 
nit: add a test?

PR #4695, opened 2022-04-01
[bug] Fix chain assignment

### LINKED ISSUE #4692: Multiple assignments
Consider the following:
```python
import taichi as ti

ti.init(
    ti.cpu,
    cpu_max_num_threads=1,
    debug=True,
    check_out_of_bound=True,
    advanced_optimization=False,
)

@ti.kernel
def bug():
    a = 1
    b = 1

    a = b = a + b

    print(a, b)

bug()
```

The above code prints `2 3` instead of `2 2`. After looking at the generated IR, it seems that taichi interprets the
```
a = b = a + b
```
line as if it were
```
a = a + b
b = a + b
```
which is all kinds of wrong.

lin-hitonami:
Related issue = close #4692


strongoier (approved):
LGTM!

PR #4700, opened 2022-04-01
[Lang] Let assertion error message support f-string

### LINKED ISSUE #4699: Support f-string in assertion message
Currently, we only support formatting assertion message using %-operator, which is not recommended by python. We should add f-string support for assertion as well.

lin-hitonami:
Related issue = close #4699


ailzhang (approved):
LGTM!

PR #4702, opened 2022-04-01
[Doc] Taichi spec: static expressions

lin-hitonami:
Related issue = #4602

### DIFF
--- a/docs/lang/articles/reference.md
+++ b/docs/lang/articles/reference.md
@@ -198,6 +198,24 @@ compile time.
 
 ### Conditional expressions
 
+### Static expressions
+
+```
+static_expression ::= "ti.static(" positional_arguments ")"
+```
+
+Static expressions are expressions that are wrapped by a call to `ti.static()`.
+The `positional_arguments` is evaluated at compile time, and the variables inside it must be compile-time constants.
+
+`ti.static()` receives one or more arguments.
+If one argument it passed to it, it returns the argument.
+If more than one arguments is passed to it, it returns a tuple containing all the arguments in the same order as they are passed.
+
+The static expressions work as hints to trigger many metaprogramming functions in Taichi, 
+such as [compile-time loop unrolling and compile-time branching](lang/articles/advanced/meta.md#compile-time-evaluations).
+
+The static expressions can also be used to [create aliases for Taichi fields and Taichi functions](lang/articles/advanced/syntax_sugars.md#aliases).
+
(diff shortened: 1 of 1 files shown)


docs/lang/articles/reference.md
```diff
@@ -198,6 +198,24 @@ compile time.
...
 ### Conditional expressions
 
+### Static expressions
+
+```
+static_expression ::= "ti.static(" positional_arguments ")"
+```
+
+Static expressions are expressions that are wrapped by a call to `ti.static()`.
+The `positional_arguments` is evaluated at compile time, and the variables inside it must be compile-time constants.
```
k-ye, line 208: > and the variables inside it must be compile-time constants

I think it's more like these variables are Python objects?

docs/lang/articles/reference.md
```diff
@@ -198,6 +198,24 @@ compile time.
...
+
+`ti.static()` receives one or more arguments.
+If one argument it passed to it, it returns the argument.
```
k-ye, line 211: Use a bullet list?

* For  a single argument, `ti.static` returns the evaluated argument.
* Otherwise, if more than one arguments are passed in, it returns ...

docs/lang/articles/reference.md
```diff
@@ -198,6 +198,24 @@ compile time.
...
+If more than one arguments is passed to it, it returns a tuple containing all the arguments in the same order as they are passed.
+
+The static expressions work as hints to trigger many metaprogramming functions in Taichi,
```
k-ye, line 214: Hint can be a bit misleading. It's more like a compile-time evaluation mechanism, just like C++'s `constexpr`.

docs/lang/articles/reference.md
strongoier: Yes. Let's unify the terms and say "and the items inside must be evaluated to Python values".

docs/lang/articles/reference.md
strongoier: `constexpr` is a keyword in C++. `ti.static` is a workaround for not being able to add keywords to Python. IMO "hint" seems a proper word here?

k-ye: I feel like it's a "mechanism", not just a hint. I.e. hint makes me think about `inline` in old-C++, where the compiler can choose to ignore it.

strongoier (approved):
LGTM!

PR #4705, opened 2022-04-01
[wip] [aot] Add support of loading module examples

ghuau-innopeak:
Related issue = #3642


ailzhang (commented):
Thanks a lot for adding this! @ghuau-innopeak I'm thinking about putting the android code in https://github.com/taichi-dev/taichi-aot-demo and adding that repo as a submodule in taich. Wdyt? @k-ye

ghuau-innopeak:
@ailzhang I think for the Android demo, it makes sense to me, it will avoid to have too much different use cases in the CMake files for the build :).
I do think it would be nice to have a small examples for AOT Loading program within Taichi though to test the build and API changes, like `aot_save`, so maybe we could keep `aot_load`? Not sure we will be able to do without `GLFW` though.

ghuau-innopeak:
Will move the MR and code to https://github.com/taichi-dev/taichi-aot-demo

PR #4706, opened 2022-04-02
[ci] Switch to self-hosted PyPI for nightly release

qiao-bo:
We will use self-hosted pypi for nightly release. This PR updates CI and Docs.


misc/upload_release.py
```diff
@@ -45,9 +45,11 @@ def upload_artifact(is_taichi):
         sys.exit(f'Missing password env var {pwd_env}')
     command = [sys.executable, '-m', 'twine', 'upload']
     if not is_taichi:
-        command.extend(['--repository', 'testpypi'])
-    command.extend(
-        ['--verbose', '-u', '__token__', '-p', twine_password, 'dist/*'])
+        command.extend(
+            ['--repository-url', 'https://pypi.taichi.graphics/simple/'])
+    uname = '__token__' if is_taichi else 'admin'
```
frostming, line 50: Does pypicloud supports tokens? If so, we should better use a token.

qiao-bo: @frostming: @rexwangcc and I just checked that Pypi Cloud seems not have access token support (there is registration token though, which does not help). Should we just stick to the user name login?

rexwangcc: Based on what we have learned from the pypicloud setup, I suggest we stick the basic auth login approach (of course we are using a non-admin account for uploading) :P

README.md
```diff
@@ -73,7 +73,7 @@ pip install taichi
 *We also provide a nightly package. Note that nighly packages may crash because they are not fully tested.  We cannot guarantee their validity, and you are at your own risk trying out our latest, untested features.*
 
 ```bash
-pip install -i https://test.pypi.org/simple/ taichi-nightly
+pip install -i https://pypi.taichi.graphics/simple/ taichi-nightly
```
rexwangcc, line 76: nit: would be better to add a comment above this line briefly explain why we are using our self-hosted PyPI for nightly?

misc/upload_release.py
rexwangcc:
```suggestion
    uname = '__token__' if is_taichi else 'taichi-nightly-cloud-pypi-uploader'
```

rexwangcc (approved):
I have one comment otherwise LGTM, thanks!

frostming (approved):

PR #4710, opened 2022-04-02
[aot] Proof of concept C API for AOT

k-ye:
Related issue = #


k-ye:
For Vulkan to run on Desktop, I somehow had to disable LLVM w/ `TI_WITH_LLVM=OFF`, hmmm

setup.py
```diff
@@ -169,7 +169,7 @@ def run(self):
         if get_os_name() != 'win':
             num_threads = os.getenv('BUILD_NUM_THREADS',
                                     multiprocessing.cpu_count())
-            build_args += ['--', f'-j{num_threads}']
+            build_args += ['--', f'-j24']
```
bobcao3, line 172: Accidentally included this?

taichi/backends/device.h
```diff
@@ -400,7 +400,7 @@ class Device {
   void print_all_cap() const;
 
   struct AllocParams {
-    uint64_t size{0};
+    size_t size{0};
```
bobcao3, line 403: Why do we need this change?

PGZXB:
@k-ye Should the C APIs'/structs' name start with unified prefix, like `ti_`, to avoid duplicate name with other libs'?

Examples of other libs' C APIs: 
* LLVM: `LLVM*`
* SDL: `SDL_*`
* leveldb: `leveldb_*`

taichi/backends/device.h
k-ye: Mostly code cleanup: I think size_t is the idiomatic way to represent size. Do you have specific concerns here? (One thing I can think of is that, size_t is platform-dependent)

bobcao3: Yes, shouldn't use platform dependent types here. It can be sometimes dangerous if these types are used (implicit casting, etc.)

setup.py
k-ye: My PC often shuts down due to overheat when I use all the threads... will revert

k-ye:
> @k-ye Should the C APIs'/structs' name start with unified prefix, like `ti_`, to avoid duplicate name with other libs'?
> 
> Examples of other libs' C APIs:
> 
> * LLVM: `LLVM*`
> * SDL: `SDL_*`
> * leveldb: `leveldb_*`

Thanks! Fixed with this style:

* All structs begin with `Taichi_`
* All functions begin with `taichi_`

PR #4715, opened 2022-04-04
[build] Fix non x64 linux builds

bobcao3:
Related issue = #4711

### DIFF
--- a/cmake/TaichiCore.cmake
+++ b/cmake/TaichiCore.cmake
@@ -397,7 +397,7 @@ if (NOT WIN32)
         # Linux
         target_link_libraries(${CORE_LIBRARY_NAME} stdc++fs X11)
         target_link_libraries(${CORE_LIBRARY_NAME} -static-libgcc -static-libstdc++)
-        if (${CMAKE_HOST_SYSTEM_PROCESSOR} STREQL "x86_64")
+        if (${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "x86_64")
             # Avoid glibc dependencies
             if (TI_WITH_VULKAN)
                 target_link_libraries(${CORE_LIBRARY_NAME} -Wl,--wrap=log2f)
--- a/cmake/TaichiCore.cmake
+++ b/cmake/TaichiCore.cmake
@@ -397,7 +397,7 @@ if (NOT WIN32)
         # Linux
         target_link_libraries(${CORE_LIBRARY_NAME} stdc++fs X11)
         target_link_libraries(${CORE_LIBRARY_NAME} -static-libgcc -static-libstdc++)
-        if (TI_ARCH_X64)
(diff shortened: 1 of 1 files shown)


cmake/TaichiCore.cmake
```diff
@@ -397,12 +397,14 @@ if (NOT WIN32)
         # Linux
         target_link_libraries(${CORE_LIBRARY_NAME} stdc++fs X11)
         target_link_libraries(${CORE_LIBRARY_NAME} -static-libgcc -static-libstdc++)
-        # Avoid glibc dependencies
-        if (TI_WITH_VULKAN)
-            target_link_libraries(${CORE_LIBRARY_NAME} -Wl,--wrap=log2f)
-        else()
-            # Enforce compatibility with manylinux2014
-            target_link_libraries(${CORE_LIBRARY_NAME} -Wl,--wrap=log2f -Wl,--wrap=exp2 -Wl,--wrap=log2 -Wl,--wrap=logf -Wl,--wrap=powf -Wl,--wrap=exp -Wl,--wrap=log -Wl,--wrap=pow)
+        if (TI_ARCH_X64)
```
qiao-bo, line 400: Maybe also add `TI_ARCH_X64` as an option (default ON) on the top? so we don't have to specify at every non-aarch64 build. We have too many flags now..

qiao-bo (commented):
Thanks for the quick fix.

cmake/TaichiCore.cmake
bobcao3: It is the same flag as the backend option. (This does complicate things, hmmm...) I should probably change this to the cmake host arch flags

cmake/TaichiCore.cmake
```diff
@@ -397,12 +397,14 @@ if (NOT WIN32)
         # Linux
         target_link_libraries(${CORE_LIBRARY_NAME} stdc++fs X11)
         target_link_libraries(${CORE_LIBRARY_NAME} -static-libgcc -static-libstdc++)
-        # Avoid glibc dependencies
-        if (TI_WITH_VULKAN)
-            target_link_libraries(${CORE_LIBRARY_NAME} -Wl,--wrap=log2f)
-        else()
-            # Enforce compatibility with manylinux2014
-            target_link_libraries(${CORE_LIBRARY_NAME} -Wl,--wrap=log2f -Wl,--wrap=exp2 -Wl,--wrap=log2 -Wl,--wrap=logf -Wl,--wrap=powf -Wl,--wrap=exp -Wl,--wrap=log -Wl,--wrap=pow)
+        if (${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "x86_64")
```
qiao-bo, line 400: This is fine as long as we don't distinguish between `host` vs `target` systems ;/

qiao-bo (approved):

PR #4716, opened 2022-04-05
[misc] Add SNode to offline-cache key

PGZXB:
Related issue = #4401

Temporary:
1. Reusing `ExpressionHumanFriendlyPrinter` to impl `ExpressionOfflineCacheKeyGenerator`
2. `get_hashed_offline_cache_key_of_snode` and `get_hashed_offline_cache_key` are declared/defined in `llvm_offline_cache.*`. I will move them to `offline_cache_util.*`

### DIFF
--- a/taichi/transforms/ir_printer.cpp
+++ b/taichi/transforms/ir_printer.cpp
@@ -55,8 +55,9 @@ class IRPrinter : public IRVisitor {
   std::string *output{nullptr};
   std::stringstream ss;
 
-  IRPrinter(ExpressionPrinter *expr_printer = nullptr, std::string *output = nullptr)
-   : expr_printer_(expr_printer), output(output) {
+  IRPrinter(ExpressionPrinter *expr_printer = nullptr,
+            std::string *output = nullptr)
+      : expr_printer_(expr_printer), output(output) {
   }
 
   template <typename... Args>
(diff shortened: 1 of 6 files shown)


strongoier:
I doubt whether this is the correct way to hash `SNode`s. We need to make sure that:
- All `SNodeTree`s are the same. This can be done by using your hash for the root of a `SNodeTree`.
- In an expression/statement, two `SNode`s can be considered same if and only if they are in the same `SNodeTree` and with same snode id.

taichi/transforms/ir_printer.cpp
```diff
@@ -805,9 +805,9 @@ void print(IRNode *root, std::string *output) {
   return IRPrinter::run(&expr_printer, root, output);
 }
 
-void gen_offline_cache_key(IRNode *root, std::string *output) {
+void gen_offline_cache_key(Program *prog, IRNode *root, std::string *output) {
```
k-ye, line 808: nit: Why is this implemented inside `ir_printer.cpp`? I think it should be in its own file under `analysis`, WDYT?

PGZXB: Agree! But this is temporary. Currently, `gen_offline_cache_key` depends on `IRPrinter` to generate AST-string . Next step, I will create a new class, maybe `ASTOfflineCacheKeyGenerator`, to replace `IRPrinter` to generate AST-Offline-Cache-Key.

But now, because `IRPrinter` is directly defined in ir_printer.cpp without declaring in *.h, I can only put `gen_offline_cache_key` in ir_printer.cpp.

PGZXB: @k-ye WDYT about https://github.com/taichi-dev/taichi/pull/4716#discussion_r846707738

k-ye: Sorry for the late reply, sounds good!

strongoier (approved):
LGTM!

PR #4718, opened 2022-04-05
[SIMT] Add all_sync warp intrinsics

varinic:
Related issue = #4631
Add all_sync intrinsics for cuda backend.

### DIFF
--- a/taichi/runtime/llvm/runtime.cpp
+++ b/taichi/runtime/llvm/runtime.cpp
@@ -1048,6 +1048,14 @@ f32 cuda_shfl_sync_f32(u32 mask, f32 val, i32 delta, int width) {
   return 0;
 }
 
+bool cuda_all_sync(u32 mask, bool bit) {
+  return false;
+}
+
+int32 cuda_all_sync_i32(u32 mask, int32 predicate) {
+  return (int32)cuda_all_sync(mask, (bool)predicate);
+}
+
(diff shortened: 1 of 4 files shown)


qiao-bo:
Hi, thanks for contributing, could you please resolve the conflict first ;/?

taichi/runtime/llvm/runtime.cpp
```diff
@@ -1036,6 +1036,18 @@ f32 cuda_shfl_up_sync_f32(u32 mask, f32 val, i32 delta, int width) {
   return 0;
 }
 
+bool cuda_all_sync(u32 mask, bool bit) {
+  return false;
+}
+
+int32 cuda_all_i32(int32 predicate) {
```
YuCrazing, line 1043: Same question: Is this function redundant? It seems that `cuda_all_i32` can be replaced by `cuda_all_sync_i32`. If you want to keep this function, please expose it to python and add corresponding unit tests. :)

YuCrazing (changes requested):

taichi/runtime/llvm/runtime.cpp
varinic: I will remove `cuda_all_i32`.

YuCrazing (approved):
LGTM! Thanks!

PR #4719, opened 2022-04-05
[SIMT] Add any_sync warp intrinsics

varinic:
Related issue = #4631 
Add any_sync intrinsics for cuda backend.

### DIFF
--- a/taichi/runtime/llvm/runtime.cpp
+++ b/taichi/runtime/llvm/runtime.cpp
@@ -1048,6 +1048,14 @@ f32 cuda_shfl_sync_f32(u32 mask, f32 val, i32 delta, int width) {
   return 0;
 }
 
+bool cuda_any_sync(u32 mask, bool bit) {
+  return false;
+}
+
+int32 cuda_any_sync_i32(u32 mask, int32 predicate) {
+  return (int32)cuda_any_sync(mask, (bool)predicate);
+}
+
 int32 cuda_ballot_sync(int32 mask, bool bit) {
   return 0;
 }
--- a/python/taichi/lang/simt/warp.py
+++ b/python/taichi/lang/simt/warp.py
@@ -7,10 +7,11 @@ def all_nonzero():
     pass
 
 
-def any_nonzero():
-    # TODO
-    pass
-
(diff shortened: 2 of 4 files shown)


taichi/runtime/llvm/runtime.cpp
```diff
@@ -1036,6 +1036,18 @@ f32 cuda_shfl_up_sync_f32(u32 mask, f32 val, i32 delta, int width) {
   return 0;
 }
 
+bool cuda_any_sync(u32 mask, bool bit) {
+  return false;
+}
+
+int32 cuda_any_i32(int32 predicate) {
+  return (int32)cuda_any_sync(UINT32_MAX, (bool)predicate);
```
YuCrazing, line 1044: I am not sure if these functions should return 0 or something else, we need @strongoier  and @qiao-bo to have a look.

qiao-bo: This is fine here. Since only `cuda_any_sync()` is patched later in llvm_context.cpp

qiao-bo:
Hi, can you please fix the conflict first? thanks ;/

taichi/runtime/llvm/runtime.cpp
```diff
@@ -1036,6 +1036,18 @@ f32 cuda_shfl_up_sync_f32(u32 mask, f32 val, i32 delta, int width) {
   return 0;
 }
 
+bool cuda_any_sync(u32 mask, bool bit) {
+  return false;
+}
+
+int32 cuda_any_i32(int32 predicate) {
```
YuCrazing, line 1043: Is this function redundant? It seems that `cuda_any_i32` can be replaced by `cuda_any_sync_i32`. If you want to keep this function, please expose it to python and add corresponding unit tests. :)

YuCrazing (changes requested):

taichi/runtime/llvm/runtime.cpp
varinic: I will remove `cuda_any_i32`.

YuCrazing (approved):
LGTM!Thanks!

PR #4721, opened 2022-04-06
[metal] Tweak Device to support Ndarray

k-ye:


taichi/backends/device.h
```diff
@@ -58,6 +58,7 @@ using DeviceAllocationId = uint32_t;
 struct TI_DLL_EXPORT DeviceAllocation {
   Device *device{nullptr};
   DeviceAllocationId alloc_id{0};
+  // TODO: Shall we include size here?
```
bobcao3, line 61: We shouldn't IMO. This is just an resource handle, size should be queried.

bobcao3 (approved):
LGTM

ailzhang (approved):
Thanks!

PR #4727, opened 2022-04-06
[spirv] Ext arr name should include arg id

ailzhang:
This was the root cause of a weird issue on qualcomm chips on android.


k-ye (approved):
LGTM!

PR #4728, opened 2022-04-06
[Doc] Taichi spec: conditional expressions and simple statements

mzmzm:
Related issue = #4602

### DIFF
--- a/docs/lang/articles/reference.md
+++ b/docs/lang/articles/reference.md
@@ -381,8 +381,6 @@ simple_stmt ::= expression_stmt
 
 ```
 expression_stmt    ::= expression_list
-starred_expression ::= expression | (starred_item ",")* [starred_item]
-starred_item       ::= assignment_expression | "*" or_expr
 ```
 
 An expression statement evaluates the expression list (which may be a single expression).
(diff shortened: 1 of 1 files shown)


docs/lang/articles/reference.md
```diff
@@ -297,14 +308,72 @@ is evaluated to the value of that expression.
...
+starred_expression ::= expression | (starred_item ",")* [starred_item]
+starred_item       ::= assignment_expression | "*" or_expr
+```
+
+An expression statement evaluates the expression list (which may be a single expression).
+
+### Assignment statements {#assignment-statements}
+
+```
+assignment_stmt ::= (target_list "=")+ (starred_expression)
```
strongoier, line 341: ditto

docs/lang/articles/reference.md
```diff
@@ -354,14 +362,70 @@ is evaluated to the value of that expression.
...
+                    | "[" [target_list] "]"
+                    | attributeref
+                    | subscription
+                    | "*" target
+```
+
+An assignment statement evaluates the expression list (remember that this can be a single expression or a comma-separated list,
+the latter yielding a tuple) and assigns the single resulting object to each of the target lists, from left to right.
+
+If the resulting object is a Python value, it will become a Taichi value implicitly by this statement.
```
strongoier, line 406:
```suggestion
The recursive definition of an assignment statement basically follows
[Python](https://docs.python.org/3/reference/simple_stmts.html#assignment-statements),
with the following points to notice:
- According to the [Variables and scope](#variables-and-scope) section, if a
target is an identifier appearing for the first time, a variable is defined
with that name and inferred type from the corresponding right-hand side
expression. If the expression is evaluated to a Python value, it will be turned
into a Taichi value with [default type](basic/type.md#default-primitive-types-for-integers-and-floating-point-numbers).
- If a target is an existing identifier, the corresponding right-hand side
expression must be evaluated to a Taichi value with the type of the
corresponding variable of that identifier. Otherwise, an implicit cast will
happen.
```

docs/lang/articles/reference.md
```diff
@@ -354,14 +362,70 @@ is evaluated to the value of that expression.
...
+
 #### Augmented assignment statements
 
+```
+augmented_assignment_stmt ::= augtarget augop (expression_list)
```
strongoier, line 411:
```suggestion
augmented_assignment_stmt ::= augtarget augop expression_list
```

docs/lang/articles/reference.md
```diff
@@ -354,14 +362,70 @@ is evaluated to the value of that expression.
...
+augtarget                 ::= identifier | attributeref | subscription
+augop                     ::= "+=" | "-=" | "*=" | "/=" | "//=" | "%=" |
+                              "**="| ">>=" | "<<=" | "&=" | "^=" | "|="
+```
+
+In Taichi, some augmented assignments (e.g., `x[i] += 1`) are automatically atomic.
+See [supported-atomic-operations](https://docs.taichi.graphics/lang/articles/operator#supported-atomic-operations) for more details.
```
strongoier, line 418:
```suggestion
Different from [Python](https://docs.python.org/3/reference/simple_stmts.html#augmented-assignment-statements), some augmented assignments (e.g., `x[i] += 1`) are [automatically atomic](basic/operator.md#supported-atomic-operations) in Taichi.
```

docs/lang/articles/reference.md
```diff
@@ -354,14 +362,70 @@ is evaluated to the value of that expression.
...
+
 #### Annotated assignment statements
 
+```
+annotated_assignment_stmt ::= augtarget ":" expression
+                              ["=" (starred_expression)]
```
strongoier, line 424:
```suggestion
annotated_assignment_stmt ::= identifier ":" expression "=" expression
```

docs/lang/articles/reference.md
```diff
@@ -354,14 +362,70 @@ is evaluated to the value of that expression.
...
+```
+The difference from normal [Assignment statements](#assignment-statements) is that only single target is allowed.
+There is also a cast if the type of target is not the same as annotation.
```
strongoier, line 427:
```suggestion
The differences from normal [Assignment statements](#assignment-statements) are:
- Only single identifier target is allowed.
- If the identifier appears for the first time, a variable is defined
with that name and type annotation (the expression after ":"). The right-hand
side expression is cast to a Taichi value with the annotated type.
- If the identifier already exists, the type annotation must be the same as the
type of the corresponding variable of the identifier.
```

docs/lang/articles/reference.md
```diff
@@ -354,14 +362,78 @@ is evaluated to the value of that expression.
...
+```
+
+Different from [Python](https://docs.python.org/3/reference/simple_stmts.html#augmented-assignment-statements), some augmented assignments (e.g., `x[i] += 1`) are [automatically atomic](basic/operator.md#supported-atomic-operations) in Taichi.
+
 #### Annotated assignment statements
 
+```
+annotated_assignment_stmt ::= identifier ":" expression "=" expression
+```
+The differences from normal [Assignment statements](#assignment-statements) are:
```
strongoier, line 429:
```suggestion
The differences from normal [assignment statements](#assignment-statements) are:
```

strongoier (approved):

PR #4733, opened 2022-04-07
[mac] Fix external_func test failures on arm backend

ailzhang:
These tests failed on my M1 mac when building from source. It passed CI
machine mainly because the tests were skipped due to
https://github.com/taichi-dev/taichi/blob/master/tests/python/test_external_func.py#L13.

We actually limit external_func to `x64` backend not `cpu`.
(Although I've tried on arm64 and it worked, I'll leave enabling it on
arm64 to a separate PR)

Related issue = #


qiao-bo (approved):
LGTM!

PR #4734, opened 2022-04-07
[SIMT] [cuda] Use correct source lane offset for warp intrinsics

qiao-bo:
Related issue = #4631 

For warp size of 32, correct lane offset is 0 for `shfl_up*` and 31 for `shfl_down*`. 
ref https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-shfl-sync


YuCrazing (approved):
LGTM! Thanks!

PR #4735, opened 2022-04-08
[Error] Raise an error when non-static condition is passed into ti.static_assert

### LINKED ISSUE #4540: `ti.static_assert` silently succeeds with non-static information
```python
import taichi as ti

ti.init(debug=True)

@ti.kernel
def foo():
    value = False
    ti.static_assert(value, "Oh, no!")

foo()
```

**Actual behaviour:**

Nothing happens.

**Expected behaviour:**

Either one of these would be fine

- `AssertionError: Oh, no!` (if you can automatically infer that `value = False` is a constant/static expression)
- `TaichiCompilationError: static_assert with non-static condition.`

lin-hitonami:
Related issue = close #4540


strongoier (approved):
LGTM!

PR #4740, opened 2022-04-08
[ci] Add python 3.6 and 3.10 to nightly release

qiao-bo:
Since we have switched to our own self-hosted PyPI cloud, Python 3.6 and 3.10 can be enabled to cover more releases in nightly.


ailzhang (approved):
Thanks!

PR #4744, opened 2022-04-08
[bug] [lang] Let matrix initialize to the target type

### LINKED ISSUE #4741: The dtype argument in `ti.Matrix()` not working
Example code:

When declaring a float matrix but using literal integers, the matrix is casted to `int` type:

```python
import taichi as ti
ti.init()

@ti.kernel
def foo():
    a = ti.Matrix([1, 2, 3], ti.f32)
    a /= 2.0
    assert all(a == (0.5, 1., 1.5)), "oops"
    print(a)

foo() # [0, 1, 1]
```

lin-hitonami:
Related issue = close #4741 
[test_f16.py::test_fractal_f16](https://github.com/taichi-dev/taichi/blob/master/tests/python/test_f16.py#L202) fails when dtype of  `z` is f16 as described in #4757 , so I temporarily removed the dtype parameter. When this is fixed, the dtype parameter should be added back.

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -80,7 +80,7 @@ class Matrix(TaichiOperations):
             if in_python_scope():
                 mat = [[x] for x in arr]
             elif not impl.current_cfg().dynamic_index:
-                mat = [[impl.expr_init(x)] for x in arr]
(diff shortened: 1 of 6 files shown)


strongoier (approved):
LGTM!

Issue #4547, opened 2022-03-16
[AOT] Do not re-JIT if the kernel has been added to the AOT module already


k-ye:
**Describe the bug**

If a kernel has been added to an AOT module multiple times, it should only be JIT once. While I found this problem for taichi kernels, we should also check for fields.

**To Reproduce**

Repro platform: macOS + Metal backend

```py
m = ti.aot.Module(ti.metal)
m.add_kernel(foo)
m.add_kernel(foo)
```

ailzhang:
This isn't a v1.1.0 blocker so I'll remove it from the milestone for now.

PR #4748, opened 2022-04-09
[Aot] Support template args in AOT module add_kernel

k-ye:


ailzhang (approved):
Thanks!! 
As discussed offline, we should throw proper error to users if they do
```
m.add_kernel(kernel, template_args={'v': ndarray1}, name='kernel')
m.add_kernel(kernel, template_args={'v': ndarray2}, name='kernel')  # esp. when ndarray2 has different element_shape as ndarray1
```
related: https://github.com/taichi-dev/taichi/issues/4547 Let's make a followup PR to address that together.

PR #4762, opened 2022-04-12
[aot] Provide a convenient API to set devallocation as argument

ailzhang:
Related issue = #


taichi/program/context.h
```diff
@@ -55,6 +56,17 @@ struct RuntimeContext {
   T get_ret(int i) {
     return taichi_union_cast_with_different_sizes<T>(result_buffer[i]);
   }
+
+  void set_arg_devalloc(int arg_id,
+                        DeviceAllocation &alloc,
+                        const std::vector<int> &sizes) {
```
k-ye, line 62: OOC, is this the array shape, or the element shape, or both?

ailzhang: Currently we need both here as the element_shape was only in opengl codegen. Need to port that to vulkan as well.

k-ye: I thought the decision was to remove element_shape?

k-ye: You could provide two APIs here, something like https://github.com/k-ye/taichi/blob/c869ecf7f0f759973620fa9fb2e0a19c4674d5b5/c_api/include/taichi/runtime.h#L39-L50 ? Otherwise this is leaking the implementation to the API.

k-ye (approved):
LGTM!

PR #4765, opened 2022-04-12
[Build] Fix visibility warnings on macOS

FantasyVR:
There are a lot of visibiliti warnings when building Taichi from source on macOS. This PR fixes the visibility warnings on macOS according to https://stackoverflow.com/questions/9894961/strange-warnings-from-the-linker-ld.


k-ye:
I think it's already `hidden` by default: https://github.com/taichi-dev/taichi/blob/a995a91ad4d2f455a6576b91dde795c9cd7d26d5/cmake/TaichiCore.cmake#L12-L18

k-ye:
Ah I see. Could you try setting the visibility in https://github.com/taichi-dev/taichi/blob/master/cmake/TaichiTests.cmake, then?

PR #4769, opened 2022-04-12
[example] Add diff sph demo

erizmr:
Add the diff sph demo.


python/taichi/examples/autodiff/diff_sph/diff_sph_demo.py
```diff
@@ -0,0 +1,632 @@
+import taichi as ti
```
ailzhang, line 1: nit: can we remove the `demo` in the filename? ;)

erizmr: Sure. Have renamed and add some comments explaining what this demo is doing.

ailzhang (approved):
Thanks!

PR #4775, opened 2022-04-13
[doc] Update install instruction with "--upgrade"

yuanming-hu:
So that people use the latest version :-)


Vissidarte-Herman (approved):
LGTM!

PR #4779, opened 2022-04-13
[lang] Store relations with 16-bit type

g1n0st:
Related issue = #3608

Now the relation metadata of `ti.Mesh` stores in `uint16` type instead of `int32` to save the global memory footprint and bandwidth at almost 2x for the relations. 
For the dynamic relations like `VV`, now we store an auxiliary data `patch_offset` to represent the start address of `data` for each patch, and `offset` only represents the local prefix within a patch. So the `offset` data can be packed into `uint16` type.


BillXu2000 (approved):
Looks good to me!

PR #4780, opened 2022-04-13
[Lang] Add basic complex arithmetic and add a mandelbrot example

neozhaoliang:
Add `cconj`, `cmul`, `cdiv`, `csqrt` functions to the `math` module to support basic complex arithmetics, and also add a mandelbrot set animation to the examples list.


python/taichi/math/_complex.py
```diff
@@ -0,0 +1,202 @@
...
+        >>>     z1 = ti.math.vec2(1, 1)
+        >>>     z2 = ti.math.vec2(0, 1)
+        >>>     ti.math.cdiv(z1, z2)  # [1, -1]
+
+    Returns:
+        :class:`~taichi.math.vec2`: the complex division of `z1 / z2`.
+    """
+    x1, y1 = z1[0], z1[1]
+    x2, y2 = z2[0], z2[1]
+    return vec2(x1 * x2 + y1 * y2, -x1 * y2 + x2 * y1) / dot(z1, z2)
```
YuCrazing, line 73:
```suggestion
    return vec2(x1 * x2 + y1 * y2, -x1 * y2 + x2 * y1) / dot(z2, z2)
```

YuCrazing (changes requested):

python/taichi/math/_complex.py
```diff
@@ -0,0 +1,202 @@
...
+        >>>     w = ti.math.cpow(z)  # [-2, 2]
+
+    Returns:
+        :class:`~taichi.math.vec2`: The power :math:`z^a`.
+    """
+    result = vec2(0.)
+    if any(z):
+        r2 = dot(z, z)
+        a = ti.atan2(z[1], z[0]) * n
+        result = ti.pow(r2, n / 2.0) * vec2(ti.cos(a), ti.sin(a))
```
YuCrazing, line 149:
```suggestion
        result = ti.pow(r, n) * vec2(ti.cos(a), ti.sin(a))
```

python/taichi/math/_complex.py
```diff
@@ -0,0 +1,202 @@
...
+        >>> @ti.kernel
+        >>> def test():
+        >>>     z = ti.math.vec2(1, 1)
+        >>>     w = ti.math.clog(z)  # [0.346574, 0.785398]
+
+    Returns:
+        :class:`~taichi.math.vec2`: The logarithm of `z`.
+    """
+    ang = ti.atan2(z[1], z[0])
+    r2 = dot(z, z)
```
YuCrazing, line 198:
```suggestion
    r = z.norm()
```

python/taichi/math/_complex.py
```diff
@@ -0,0 +1,202 @@
...
+    return vec2(ti.log(r2) / 2., ang)
```
YuCrazing, line 199:
```suggestion
    return vec2(ti.log(r), ang)
```

python/taichi/math/_complex.py
```diff
@@ -0,0 +1,202 @@
...
+    Returns:
+        :class:`~taichi.math.vec2`: The power :math:`z^a`
+    """
+    r = ti.exp(z[0])
+    return vec2(r * ti.cos(z[1]), r * ti.sin(z[1]))
+
+
+@ti.func
+def clog(z):
+    """Returns the complex logarithm of `z`.
```
YuCrazing, line 180: Maybe we should point out the range of the result, like `i(-pi, pi] ` or whatever.

python/taichi/math/_complex.py
neozhaoliang: Oh, no, this is a bad mistake.

python/taichi/math/_complex.py
```diff
@@ -0,0 +1,202 @@
...
+
+    Args:
+        z (:class:`~taichi.math.vec2`): The input.
+
+    Example::
+
+        >>> @ti.kernel
+        >>> def test():
+        >>>     z = ti.math.vec2(-1, 0)
+        >>>     w = ti.math.csqrt(z)  # [0, 1]
```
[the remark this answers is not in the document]
neozhaoliang: we don't need to return both roots because the other one is just the negative. This is kept the same with the complex arithmetic in python and numpy.

python/taichi/math/_complex.py
```diff
@@ -0,0 +1,202 @@
...
+
+    Returns:
+        :class:`~taichi.math.vec2`: The complex square root.
+    """
+    result = vec2(0.)
+    if any(z):
```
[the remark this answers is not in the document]
neozhaoliang: `any(z)` will depend on the machine's precision. I think here taking a square root won't cause overflow problems?

python/taichi/math/_complex.py
```diff
@@ -0,0 +1,202 @@
...
+        >>>     w = ti.math.cinv(z)  # [0.5, -0.5]
+
+    Returns:
+        :class:`~taichi.math.vec2`: The reciprocal of `z`.
+    """
+    return cconj(z) / dot(z, z)
+
+
+@ti.func
+def cpow(z, n):
```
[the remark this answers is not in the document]
neozhaoliang: It's a good idea to switch between cartesian and polar representations. Maybe we can add a `polar` function to help user convert a vector to its polar form?

python/taichi/math/_complex.py
```diff
@@ -0,0 +1,202 @@
...
+        >>> def test():
+        >>>     z = ti.math.vec2(1, 1)
+        >>>     w = ti.math.cpow(z)  # [-2, 2]
+
+    Returns:
+        :class:`~taichi.math.vec2`: The power :math:`z^a`.
+    """
+    result = vec2(0.)
+    if any(z):
+        r2 = dot(z, z)
```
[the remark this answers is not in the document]
neozhaoliang: Here we have another `pow` call below. We can avoid taking this square root to gain a bit of speed up~

python/taichi/math/_complex.py
neozhaoliang: This is the same reason as in the above. `devide by 2` is a bit faster than taking square root in the `z.norm()` call.

python/taichi/math/_complex.py
neozhaoliang: Calling `z.norm()` would involve a `square` call, which is less efficient than just divide the logarithm by 2.0.

YuCrazing (approved):
LGTM! Thanks!

PR #4803, opened 2022-04-15
[vulkan] Fix Vulkan device score bug

victoriacity:
Fixes the score calculation of Vulkan devices to correctly use the API version. This makes the Vulkan backend correctly pick the discrete GPU on a laptop.


k-ye (approved):
LGTM!

bobcao3 (approved):

Issue #4777, opened 2022-04-13
Support all kinds of SNodes in AOT


k-ye:
**Concisely describe the proposed feature**

The current AOT solution is good to use for dense Taichi fields. However, to support all kinds of Taichi fields, we probably need to systematically revamp the fields design, so that a full SNode Tree can be baked into the AOT module. I'd like to write an RFC to discuss the solution.

PR #4806, opened 2022-04-16
[Rfc] AOT for all SNodes

k-ye:
Related issue = #4777

For a better reading experience, please go to https://github.com/k-ye/taichi/blob/rfc-aot/docs/rfcs/20220413-aot-for-all-snode.md


k-ye:
@AmesingFlank @xuhao1 @ifsheldon could you provide some feedback on whether you think the `SNodeTreeBuilder` API being intuitive? Thanks!

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,340 @@
...
+  .place(x, y)
+
+# `tree_t` stands for "tree type".
+tree_t = builder.build()
+```
+
+Similarly, `SNodeTreeBuilder.build()` doesn't allocate memory for the tree. It only builds *the type of* a SNode tree. You can later instantiate a tree with `tree_t.instantiate()`. There are a few reasons behind this type-tree decoupling design:
+
+1. We have explicit access to the SNode tree type. This is a must for AOT, but can also be used as type annotations for enhanced language formality.
+2. We can instantiate as many trees as we want from this type.
```
AmesingFlank, line 110: -- and pass them to the same kernel without re-compilation.

This seems to allow SNodes to offer similar functionalities to ndarrays, as it enables the same kernel to operate on different buffers without recompiling. 

How feasible do you think it would be to take this one step further, and allow kernels to operate on different instances of one-layer dense fields without recompilation?

For example, it would be great if we could support
```py
x = builder.add_field(dtype=ti.f32, name='x')
builder.tree()
  .dense(ti.ij, 5)
  .place(x)

tree_t = builder.build()

x_t = tree_t.field_type("x")

@ti.kernel
def f(x: x_t):
  for i in x:
    ....
```
Or, simpler still:
```py
@ti.kernel
def f(x: ti.types.field(ti.f32, (N)) )
  for i in x:
    ...
```

AmesingFlank (commented):
These APIs looks very intuitive to me. I think they'll be big step forward for our AOT tooling.

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,340 @@
...
+Internally, as you are using Taichi's SNode system to construct hierarchies, you are also constructing a `SNodeTree` type at the same time. This is done by Taichi's [`FieldsBuilder`](https://github.com/taichi-dev/taichi/blob/master/python/taichi/_snode/fields_builder.py).
+
+## A working design
+
+We will make the SNode tree and its type explicit by providing `SNodeTreeBuilder`. Each field needs to be registered into the builder via `add_field()`. `add_field()` does *not* actually do any memory allocation. Instead, it just returns a *field ID*, which can be used to retrieve a field from the tree inside the kernel.
+
+```py
+builder = ti.SNodeTreeBuilder()
+
+x = builder.add_field(dtype=ti.f32, name='x')
```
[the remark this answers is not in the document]
k-ye: This is awesome! I think we can have both: `name=None` by default, which, if not explicitly specified, will be the var name.

docs/rfcs/20220413-aot-for-all-snode.md
k-ye: This is an interesting thought. I haven't really figured out how to best describe the type for an individual field. Your suggestion does remind me to make one-layer dense field and Ndarrays interchangeable, i.e. one is the view of the other.

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,340 @@
...
+   # boilerplate to generate tree type and instantiate a tree ...
+
+   @ti.kernel
+   def foo(tr: tree_t):
+     x = ti.static(tr.get_field(x_fid))  # 1
+     for i in x:
+       x[i] = i * 2.0
+   ```
+
+   Note that this design requires that part of the kernel (1) being evaluated inside Python. It also pulls in the global variable `x_fid`, which kind of violates our initial goal. We could require that `x_fid` is passed into the kernel as an argument. But maybe it's fine just to view that as a trivial Python constant?
```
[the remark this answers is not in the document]
k-ye: Good point. I needed this to support this use case: https://github.com/k-ye/taichi/blob/8d6a6706f36322faa8e61181a6163699c2b2a6b1/docs/rfcs/20220413-aot-for-all-snode.md?plain=1#L332

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,340 @@
...
+2. It introduces the notion of "nested field", which Taichi doesn't currently have a good abstraction for. Because of this, it is quite complicated to apply certain kind of optimizations for a composite-typed field. For example, vectorized load/save consumes the same bandwidth as scalar ops on certain platforms. Without a good abstraction, the checking for whether a matrix field is AoS or SoA has to be spread across different passes in CHI IR.
+
+If we further think about the problem, SoA `x` is not really a field. Instead, it is a *grouped view* of three individual scalar fields. This view provides matrix operations, which won't make sense for each individual scalar field.
+
+In addition, because type is currently coupled with Taichi field definition, a Taichi field has to be implemented as individual fields in order to support the SoA scenario. Once we switch to the type builder pattern, we can control how the type is constructed first, and choose the field implementation later.
+
+
+```py
+builder = ti.SNodeTreeBuilder()
+vel = builder.add_field(dtype=ti.vec3, name='vel', layout=ti.Layout.SOA)
```
strongoier, line 178: In fact, the case you are describing is not `SOA`, but any layout other than `AOS`. I suggest rephrasing this section a bit. Using `ti.Layout.SOA` here is especially inaccurate.

strongoier (commented):
The overall idea looks great!

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,340 @@
...
+
+for c in vel.compoenents():
```
ailzhang, line 180:
```suggestion
for c in vel.components():
```

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,340 @@
...
+for f in x._get_field_members():  # `x` consists three scalar f32 fields
+  ti.root.dense(ti.ij).place(f)
+```
+
+This introduces confusion at several spots:
+
+1. Type is not purely decided by `dtype`, but also by how the field is placed.
+2. It introduces the notion of "nested field", which Taichi doesn't currently have a good abstraction for. Because of this, it is quite complicated to apply certain kind of optimizations for a composite-typed field. For example, vectorized load/save consumes the same bandwidth as scalar ops on certain platforms. Without a good abstraction, the checking for whether a matrix field is AoS or SoA has to be spread across different passes in CHI IR.
+
+If we further think about the problem, SoA `x` is not really a field. Instead, it is a *grouped view* of three individual scalar fields. This view provides matrix operations, which won't make sense for each individual scalar field.
```
ailzhang, line 171: nice summary!

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,340 @@
...
+# ...
+tree_t = builder.build()
+
+@ti.kernel
+def foo(tr: tree_t):
+  # ...
+
+m = ti.aot.Module(arch)
+m.add_snode_tree_type(tree_t, name="vel_tree")
+m.add_kernel(foo)
```
ailzhang, line 239: As we discussed offline this RFC is mainly about how to handle the global fields from our python program in aot. As a result the ongoing effort for compute graph will mainly focus on replacing `add_kernel` as `add_graph` (or a similar name) and not worry about fields for now.

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,340 @@
...
+tree = tree_t.instantiate()
+ti._runtime.global_snode_trees.append(tree)
+for ft in ti.root._field_thunks:
+  ft.bind(tree)
+
+# Make `ti.root` a new SNodeTreeBuilder to allow for dynamic fields
+ti.root = SNodeTreeBuilder()
+```
+
+When JIT compiling a Taichi kernel, it transforms `x@old` into `x.tree.get_field(x.field_id)`, where `x` is the `FieldThunk`.
```
ailzhang, line 332: OOC what does field_id map to in our current version?

ailzhang (approved):
Looks great! Thanks!

docs/rfcs/20220413-aot-for-all-snode.md
k-ye: Nothing... It will be a handle into a SNodeTree to identify a field

docs/rfcs/20220413-aot-for-all-snode.md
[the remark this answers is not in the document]
k-ye: So your opinion is that we should always ask the users to be explicit about the field names via `name=`?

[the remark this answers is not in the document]
strongoier: Yep.

docs/rfcs/20220413-aot-for-all-snode.md
strongoier: SG!

k-ye: I just removed this part. `add_field` should not be mixed with `add_field_view`.

ifsheldon:
Overall, the APIs look great to me after your discussion and amendments.

In addition, one that has been in my wish list is to further decouple the `SNodeTree` structure definition from the code of `SNodeTreeBuilder`. We had some discussion long time ago and I think AOT will also benefit from such decoupling.

Specifically, I would love to see an API that allows dynamic `SNodeTree` structure specification, storing this specification somewhere and use it as a blueprint for `SNodeTreeBuilder` to place fields.

For example, some code like this:

```python
depth = calc_tree_depth(some_data)
blueprint = ti.SNodeTreeStructure()
for i in range(depth):
	blueprint.add_dense(ti.ij, 4)
blueprint.save(some_file) # in Python code, store it somewhere

# ....
builder = ti.SNodeTreeBuilder()
blueprint = ti.SNodeTreeStructure(some_file) # load from somewhere, in C code maybe
builder.from_blueprint(blueprint).place(some_field)
builder.finalize()
some_field.from_numpy(some_data)
# maybe field can also have a copy of its structure information?
structure = some_field._structure
another_builder = ti.SNodeTreeBuilder()
another_builder.from_blueprint(structure).place(another_field) # use the same SNodeTree structure
another_builder.finalize()
```

This allows more flexible and dynamic tree structure, and it can also be used in AOT I think. This is just a minor extension to the existing `SNodeTreeBuilder` APIs I think.

ifsheldon:
For me, the first example in AoS vs. SoA is weird because I think when I want to use a field of `vec3`, I know that I want the locality of 3 nested numbers. The semantics of `dtype=ti.vec3` matches well my intention. The code of `ti.Vector.field` is confusing to me because the coder (whoever they are) seems to want the semantics of a field of vector while they want a different kind of locality. So, my question for thor who want to use `ti.Vector.field` like this will be "if you don't want vec.x and vec.y to be close to each other in memory, why do you want to use vector fields?" 

In fact, I'm more in favor of `ti.field(dtype=ti.vec3)` or (maybe) `ti.field(dtype=ti.vec(n))` instead of `ti.Vector.field(n, dtype=ti.f32)`. This should make vectors (and possibly matrices) a first-class dtype in Taichi.

k-ye:
@ifsheldon  thanks!

> Specifically, I would love to see an API that allows dynamic SNodeTree structure specification, storing this specification somewhere and use it as a blueprint for SNodeTreeBuilder to place fields.

Note that `SNodeTreeBuilder` only builds the *type of the SNodeTree*. Then you can instantiate as many trees/fields as you want. From your example code above, I feel like this already satisfies your need?

As a concrete example following https://github.com/k-ye/taichi/blob/rfc-aot/docs/rfcs/20220413-aot-for-all-snode.md#a-working-design

```py
builder = ti.SNodeTreeBuilder()

x = builder.add_field(dtype=ti.f32, name='x')
y = builder.add_field(dtype=ti.i32, name='y')
builder.tree()
  .pointer(ti.ij, 4)
  .dense(ti.ij, 5)
  .place(x, y)

# `tree_t` stands for "tree type".
tree_t = builder.build()

tree1 = tree_t.instantiate()
tree2 = tree_t.instantiate()
# Both `treee1` and `tree2` has fields `x` and `y`.
```

> In fact, I'm more in favor of ti.field(dtype=ti.vec3) or (maybe) ti.field(dtype=ti.vec(n)) instead of ti.Vector.field(n, dtype=ti.f32). This should make vectors (and possibly matrices) a first-class dtype in Taichi.

Oh yeah, that's actually been discussed quite a while back, see https://github.com/taichi-dev/taichi/discussions/4086. The code you are referring to (i.e. `x = ti.Vector.field(3, ti.f32)`) is *the current approach*. I think we will still need to keep this one for compatibility, but we can support `field(dtype=vecn)` as well.

ifsheldon:
Then the dynamic tree structure specification is done now. Almost there. I guess my code example can be written as:
```python
builder = ti.SNodeTreeBuilder()

x = builder.add_field(dtype=ti.f32)
tree_blueprint = builder.tree()
depth = calc_tree_depth(some_data)

for i in range(depth):
	tree_blueprint = tree_blueprint.dense(ti.ij, 4)

builder.build()
```
but this still couples the structure specification with the builder, because I think in the returned object of `.tree()` a reference to `builder` is there since the code is `builder.build()` not `builder.build_from(tree_blueprint)`. If the structure specification is decoupled from the builder, maybe the code can be:
```python
blueprint = ti.SNodeTreeBlueprint()

x = blueprint.add_field(dtype=ti.f32)
depth = calc_tree_depth(some_data)

for i in range(depth):
	blueprint = blueprint.dense(ti.ij, 4)

builder = ti.SNodeTreeBuilder()
builder.build_from(blue_print)
```
The reason I want something like this is to create fields with a structure specification. That specification can be store somewhere, for example, in a field as an attribute.

An useful example is optimizer like those in PyTorch. Given a field, the code in an optimizer should be able to create fields that have the same tree structure to optimize memory access.
```python
def create_adjoint_momentum_field(field):
    """
    create a field that has the same tree structure as the input, which can be used to store optimizer information,
    
    I'm not sure how to design such an API, so the code below is just for illustration
    """
    tree_structure = field._tree_structure # should give a blueprint that can be used in builder
    momentum_field = blueprint.add_field(dtype=ti.f32)
    builder = ti.SNodeTreeBuilder()
    builder.build_from(tree_structure)
    return momentum_field
```

k-ye:
I don't think you should still refer to individual fields in this case. Instead, you should pass around the SNodeTree, which can be easily instantiated from its type.

Also, I feel like the `SNodeTreeBuilder` is already doing something similar to blueprint? Why would we need both blueprint and a builder?

ifsheldon:
> Also, I feel like the `SNodeTreeBuilder` is already doing something similar to blueprint? Why would we need both blueprint and a builder?

Yeah, `SNodeTreeBuilder` is doing something similar in that it allows dynamic tree construction, but it cannot store the tree structure. Like building a house, you can of course tell the builder what you want, the builder sketch out a blueprint and keep the blueprint private as long as the builder can build the house for you, but sometimes you still want to keep the blueprint yourself. Just like the example of optimizer, given an arbitrary field, users (or library developers) may want to copy it with a same structure. You need to store the structure somewhere (i.e., in a blueprint), attach it to a field and then `SNodeTreeBuilder` can based on this blueprint to construct a field with the same structure.

> I don't think you should still refer to individual fields in this case. Instead, you should pass around the SNodeTree, which can be easily instantiated from its type.

I know a SNodeTree may contain multiple fields. But for one field `A`, to clone it and make another field with the same structure, the path from `A` to the root of the tree is enough I think, and this path (or  "*partial* blueprint" of a SNodeTree) can be store in an attribute that is attached to `A`.

k-ye:
Quick notes

1. SNodeTree memory (known at compile time) + memory pool (runtime)/device allocator
3. Ban overlapping field components in the field view
4. BLS/TLS needs type
5. make_memory_typed --> instantiate a tree + memory
6. One SNodeTree per data_oriented?
7. Hide handle, just use name to `place()`

```py
class FieldViewThunk:
  ...

x = ti.field(...)
vel = ti.field(dtype=ti.vec3)  # return `FieldViewThunk`

@ti.kernel
def foo():
  for I in vel:
```

Nested field view
```py
add_field(dtype=ti.vec3, name='pos')
add_field(dtype=ti.vec3, name='normal')
add_field_view(dtype=vertex_t, components=['pos', 'normal'])
```

k-ye:
@ifsheldon Hi, we're back from the holidays now.

> Yeah, SNodeTreeBuilder is doing something similar in that it allows dynamic tree construction, but it cannot store the tree structure.

Hmm, I feel like your request is mostly to make `SNodeTreeBuilder::build` callable multiple times, e.g.

```py
b = ti.SNodeTreeBuilder()
b.add_field(ti.f32, name='x')
b.dense(...).place('x')
tree1_t = b.build()

b.add_field(ti.i32, name='y')
b.pointer(...).place('y')
tree2_t = b.build()
```

This way the builder is just a blueprint? Furthermore, we can serialize and deserialize a builder to a file, if necessary.

---

Cloning a single field in this case probably doesn't make sense anymore, because a field must be defined and used within a tree. We can easily support your use case, though:

```py
# Assuming `tr` has defined fields `x` and `y`
tr = ...  # instantiated from a tree type

def clone_tree(tr):
  return tr.my_type().instantiate()

tr2 = clone_tree(tr)
# `tr2` has `x` and `y` fields, too.
```

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,363 @@
...
+vel_soa[i, j] = vel_soa[i, j] @ some_vec3
+# equivalent to
+vel_tmp = ti.vec3([v0[i, j], v1[i, j], v2[i, j]])
+vel_tmp = vel_tmp @ some_vec3
+v0[i, j] = vel_tmp[0]
+v1[i, j] = vel_tmp[1]
+v2[i, j] = vel_tmp[2]
+```
+
+To make field view even more powerful, we can supported *nested field views*. For example:
```
k-ye, line 216: Added discussion on nested field views @bobcao3

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,363 @@
...
+builder = ti.SNodeTreeBuilder()
+builder.add_field(dtype=ti.vec3, name='pos')
+builder.add_field(dtype=ti.vec3, name='normal')
+builder.add_field(dtype=ti.f32, name='radius')
+builder.add_field_view(dtype=sphere_t, name='spheres', 
+                       compoenents=[['pos', 'normal'], 'radius'])
+###                                 ^^^^^^^^^^^^^^^^^ Note this is nested
+```
+
+## Gradient and AutoDiff
```
k-ye, line 231: Added discussion on gradient fields @ailzhang @strongoier

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,380 @@
...
+};
+
+using TreeType = PointerDense<AB>;
+```
+
+We cannot declare the `run` kernel as `void run(? a, ? b)`. Instead, we have to pass in a `TreeType` instance as a whole into the kernel, i.e., `void run(TreeType &tree)`.
+
+Internally, as you are using Taichi's SNode system to construct hierarchies, you are also constructing a `SNodeTree` type at the same time. This is done by Taichi's [`FieldsBuilder`](https://github.com/taichi-dev/taichi/blob/master/python/taichi/_snode/fields_builder.py).
+
+## A working design
```
k-ye, line 106: Would be great if you could share some feedback :-) @bsavery

ifsheldon:
> I feel like this request is mostly to make SNodeTreeBuilder::build callable multiple times?

This is not the purpose but a result. With a blueprint, we can call a builder to build a field or a SNodeTree for multiple times.

> This way the builder is just a blueprint? Furthermore, we can serialize and deserialize a builder to a file, if necessary.

Not exactly. You will not call the person who knows how to build your house as a blueprint. Likewise, the builder has a blueprint internally, but it's not a blueprint. 

> Cloning a single field in this case probably doesn't make sense anymore, because a field must be defined and used within a tree. WDYT if we support the clone case this way, though?

For the case of building adjoint fields of a given field, cloning a subtree is sufficient and cloning a whole SNodeTree is not so flexible. For example, there's a 2D matrix `Field A` in the diagram. Besides `Field A`, there are `Field B` and `Field C` in the same tree. The user who constructed this tree wants an optimizer for only `Field A`. If the optimizer library copies the whole tree, it's just a waste of memory.
snode_tree
What the optimizer library really needs is the subtree of `Field A` (marked in red), so it'd be better if `Field A` itself has a partial blueprint that stores this subtree. With this partial blueprint, the code of the optimizer library can just ask `ti.SNodeTreeBuilder` to build this subtree and get an adjoint field `Field A adjoint 1`. 

But what if the optimizer library wants to do more on this partial blueprint? Say, creating a field with one more layer of `dense` (this is a fair enough case because the adjoint attributes of an optimizer may be high-dimensional data like vectors and/or matrices). With this partial blueprint, the code of the optimizer library can just manipulate the tree in the blueprint and then ask `ti.SNodeTreeBuilder` to build a new tree like below. The below tree has another field `Field A adjoint 2` that has one more layer of `dense`.
new_snode_tree

Without a blueprint, we cannot pass the tree structure of existing field(s) to other parties and allow them to change a given tree structure. Of course, you can also store the whole tree structure in the first diagram somewhere (possibly also in a field) so that a receiver of a field (e.g., the optimizer library) can have a bigger picture of the memory layout. And this decoupling memory structure from `ti.SNodeTreeBuilder` is more flexible, powerful and intuitive than simply serializing and deserializing a builder. Simply put, the builder is not the point but the structure and sub-structure of a SNodeTree.

docs/rfcs/20220413-aot-for-all-snode.md
strongoier: Are users going to manually place the grad fields? The new way is much cleaner, but I have a feeling that it is not possible to get autodiff backward-compatible..

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,380 @@
...
+```py
+vertex_t = ti.types.struct({'pos': ti.vec3, 'normal': ti.vec3})
+sphere_t = ti.types.struct({'center': vertex_t, 'radius'; ti.f32})
+
+builder = ti.SNodeTreeBuilder()
+builder.add_field(dtype=ti.vec3, name='pos')
+builder.add_field(dtype=ti.vec3, name='normal')
+builder.add_field(dtype=ti.f32, name='radius')
+builder.add_field_view(dtype=sphere_t, name='spheres',
+                       compoenents=[['pos', 'normal'], 'radius'])
```
strongoier, line 244:
```suggestion
                       components=[['pos', 'normal'], 'radius'])
```

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,380 @@
...
+v0[i, j] = vel_tmp[0]
+v1[i, j] = vel_tmp[1]
+v2[i, j] = vel_tmp[2]
+```
+
+To make field view even more powerful, we can supported *nested field views*. For example:
+
+```py
+vertex_t = ti.types.struct({'pos': ti.vec3, 'normal': ti.vec3})
+sphere_t = ti.types.struct({'center': vertex_t, 'radius'; ti.f32})
```
strongoier, line 237:
```suggestion
sphere_t = ti.types.struct({'center': vertex_t, 'radius': ti.f32})
```

docs/rfcs/20220413-aot-for-all-snode.md
```diff
@@ -0,0 +1,380 @@
...
+builder = ti.SNodeTreeBuilder()
+builder.add_field(dtype=ti.f32, name='v0')
+builder.add_field(dtype=ti.f32, name='v1')
+builder.add_field(dtype=ti.f32, name='v2')
+for v in ['v0', 'v1', 'v2']:
+  builder.tree().dense(ti.ij, 4).place(v)
+
+# Checks that
+# 1. `components` and `dtype` are compatible.
+# 2. All the fields in `components` are homogeneous in their SNode hierarchy.
```
strongoier, line 205: I think checking that all components are having the same shape is enough.

strongoier (approved):
LGTM!

docs/rfcs/20220413-aot-for-all-snode.md
ailzhang: Hmmm if we say vec3 implies homogeneous elements, it's fair to say all the element fields must have identical snode hierarchy? I kinda feel enforcing this in the new api until there's a valid request might save us some time worrying about the edge cases.

docs/rfcs/20220413-aot-for-all-snode.md
ailzhang: @strongoier could you elaborate on what'll be bc-breaking? ;) We currently allow manually placing grad fields as well? https://github.com/taichi-dev/taichi/blob/master/python/taichi/examples/autodiff/simple_derivative.py#L10

k-ye:
> With a blueprint, we can call a builder to build a field or a SNodeTree for multiple times.

I think you and I are just referring to the same thing with different terms. IIUC, from your perspective, the blueprint comes with all these SNode-definition ops (e.g. `dense`, `pointer`...) to define a tree. By the end, you just need to feed this blueprint into a builder. Hence the builder needs just a single `build(self, blueprint)` method. What I'm saying is basically to bake these ops into the builder itself. "builder" is supposed to be capable of manipulating the tree structure.

> You will not call the person who knows how to build your house as a blueprint.

Sure, what I'm trying to avoid is to have two concepts talking about the same thing. Let's take this as an example:

```py
blp = ti.SNodeTreeBlueprint()
# many ops on `blp`

# Option 1
ti.build_tree_type(blp)
# Option 2
blp.build()
```

In both cases, we no longer need another concept of builder, and now this blueprint is just the builder itself. Note that "Build" is an action, and needs to be operated on an object (call it blueprint or builder).

>  cloning a subtree is sufficient and cloning a whole SNodeTree is not so flexible... If the optimizer library copies the whole tree, it's just a waste of memory.

This I agree. Meanwhile, you probably need to understand that tree-shape is fixed once built. So to achieve something like "adding a new branch/field to an existing tree", we will need to allocate a new chunk of memory, copy the content from the old memory to the new one (much like how a vector grows its memory). This should be possible once SNodeTree is consolidated. However, things become a lot trickier if we want to support this in the current global fields, because trees (`ti.root`) are implicit in this case.

I like the idea of having a separate blueprint for each field. I believe we can build this utility on top of SNodeTree.

docs/rfcs/20220413-aot-for-all-snode.md
k-ye: Echo what @ailzhang has said. This basically works like before:

1. If you invoke `add_field(dtype, name='x', needs_grad=True, shape=(a, b, c))`. Then you shouldn't `place('x')` anymore. The builder will just generate two new dense SNodes. (This is actually talking about two things. If you define `shape` but not `needs_grad`, the builder will still do so, but just one dense SNode)
2. Otherwise you need to manually place the field and its gradient field?

docs/rfcs/20220413-aot-for-all-snode.md
strongoier: "homogeneous" itself doesn't have a clear meaning, and the criteria can vary significantly. However, the additional check and additional restrictions on user code don't seem to benefit us, especially when we decide to avoid manipulating backend pointers in CHI IR.

k-ye: When I wrote homogeneous, I was thinking of https://github.com/taichi-dev/taichi/issues/3810

k-ye: I think this is a fairly small point. Not enforcing this actually reduces our development efforts. In the long run though, not enforcing this could still result in certain "unexpected" use cases. I'd like to refer to Hyrum's Law here...

> With a sufficient number of users of an API, it does not matter what you promise in the contract: all observable behaviors of your system will be depended on by somebody.

docs/rfcs/20220413-aot-for-all-snode.md
strongoier: In current Taichi, the grad field is always created. `needs_grad` only helps you place it when `shape` is specified. This is different from the description in this RFC, where `needs_grad` is used to determine whether a field has a grad field.

BTW there is another commonly used helper function `lazy_grad()` for placing, which should also be supported somewhere.

docs/rfcs/20220413-aot-for-all-snode.md
ailzhang: If we enforce it we might be able to hear from users who has a valid use case and share with us. But if not we'll always have to assume someone relies on this indeterministic use case :P

strongoier: Yep #3810 is exactly checking "homogeneous" with a relatively relaxed criteria. And behavior like #3810 indeed looks a bit weird to me as a user: why does an API work only when we use a bunch of them in certain patterns?

docs/rfcs/20220413-aot-for-all-snode.md
ailzhang: Good point about `lazy_grad`! maybe that's something we can add to the snodetreebuilder? 

TIL we always create grad field but sometimes don't place it (nice catch!).... But I'd expect nothing should break if we only create the grad field when needs_grad=True or lazy_grad is on in current Taichi? (If so we should probably change our current implementation)

docs/rfcs/20220413-aot-for-all-snode.md
k-ye: Are we talking about vectors or structs here? They deserve different handling IMHO. I don't think homogeneous is necessary for struct views. For matrices, however, it's equally weird why someone would want vectors to take on different sparsity.

strongoier: > If we enforce it we might be able to hear from users who has a valid use case and share with us. But if not we'll always have to assume someone relies on this indeterministic use case :P

This sounds like a good reason for enforcing it!

> They deserve different handling IMHO.

I agree.

docs/rfcs/20220413-aot-for-all-snode.md
strongoier: Sounds good! We can have a `lazy_grad()` API for `SNodeTreeBuilder`. The implementation needs to be changed anyway when we switch to the new API, so I think the behavior that grad fields are always created can be altered then.

k-ye: Thanks for the discussion! Yeah, I didn't know that grad field is always created, either... I feel like we ought to correct this implementation.

k-ye: Fixed

docs/rfcs/20220413-aot-for-all-snode.md
k-ye: Fixed

ifsheldon:
> Meanwhile, you probably need to understand that tree-shape is fixed once built. So to achieve something like "adding a new branch/field to an existing tree", we will need to allocate a new chunk of memory, copy the content from the old memory to the new one (much like how a vector grows its memory).

The intention is not to add a new branch/field to an existing tree but to copy a tree structure from a given tree, modify that structure and build a new tree. The reason why I'd like to separate blueprints and builders is to avoid such confusions. When you build a tree or fields from a blueprint, you concretize the blueprint and allocate memory, while the blueprint is a tentative description of preferred memory layout.

As far as I know, the performance is fine when two fields are accessed "simultaneously" with *the same pattern* even if their memory locations are far away. For example, a kernel accessing two matrices of *the same memory layout* with *the same indices* even if the two matrices are stored far away is fine in terms of performance. In this case, instead of extending an existing tree, cloning a tree with the same memory layout is sufficient. 

Extending an existing tree with memory movement may have other benefits and thus fit into other use cases, though.

k-ye:
> The intention is not to add a new branch/field to an existing tree but to copy a tree structure from a given tree, modify that structure and build a new tree. 

This gets me confused. At one point you said that

>  cloning a subtree is sufficient and cloning a whole SNodeTree is not so flexible. 

So I'd assume that memory is a concern.

>  When you build a tree or fields from a blueprint, you concretize the blueprint and allocate memory, while the blueprint is a tentative description of preferred memory layout.

Builder is for building the layout and tree type. You can then instantiate a tree out of this tree type.

```py
b = ti.SNodeTreeBuilder()
# dense/pointer/place operations on `b`

tree_type = b.build()  # <-- here, weget the type of the tree
tree = tree_type.instantiate()  # <-- here, we allocate the memory for the tree.
```

So I feel like in this case, tree_type is the blueprint, whereas builder is the manipulator of the tree type. For example, let's say we have a tree:

```py
# Get the type
tree_type = tree.get_type()

b = ti.SNodeTreeBuilder.init_from(tree_type)
# Now we can add more fields and SNodes on `b`
```

ifsheldon:
> This gets me confused. At one point you said that ...

I meant cloning only the structure of a subtree, not (the data and its structure). Sorry.

>So I feel like in this case, tree_type is the blueprint, whereas builder is the manipulator of the tree type. For example, let's say we have a tree ...

`tree_type` seems great. But we need to be able to store the tree somewhere, not discarding it after building a tree. And the subtree (a branch if you would) of a field can also be stored in an attribute of a field. I feel like this `tree_type` has not been documented, but maybe I've been missing something. 

In this case, I think builder being the manipulator is confusing because the user need to "import" an existing tree structure (`ti.SNodeTreeBuilder.init_from(tree_type)`) into a builder, and then the builder can further manipulate the tree structure. And then, I guess if the user want to store the new tree structure, he/she will need to "export" that structure from this builder? This is some sort of unnecessary coupling in my opinion. 

From the analogy of building a house, I think it's better to be able to manipulate a blueprint without the help of a builder. Like two cases: (1). you sketch a new blueprint based on a given blueprint and give the new one to a builder to build your house or (2). you ask a builder to look at a given blueprint, ask him/her to modify the blueprint and create a new blueprint, and then the builder builds your house accordingly. I prefer case 1 and I think from the perspective of documentation, it's easier to explain the usages.

k-ye:
Again, I feel like we are just using different terminologies to do the same thing. Builder has nothing to do with an actual tree, where tree is defined as having both a type and a chunk of memory. Here's the relationship:

builder --> tree type --> tree

Builder is not importing the *actual tree*. Instead it's importing the *tree type*. And even so, the tree type being imported is *not* gonna be changed, because the tree type is immutable after built. Instead, this imported tree type serves exactly as the base blueprint you talked about in (1). Builder can keep adding fields. Then once you invoke `build()`, you get a new tree type (the new blueprint). Again, this new tree type cannot be modified now.

ifsheldon:
> Builder has nothing to do with an actual tree, where tree is defined as having both a type and a chunk of memory. Here's the relationship

> But we need to be able to store the tree somewhere, not discarding it after building a tree. And the subtree (a branch if you would) of a field can also be stored in an attribute of a field. 

I should have state this more clearly, sorry. This should be "But we need to be able to store the tree *structure* somewhere, not discarding it after building a tree. And the subtree *structure* of a field can also be stored in an attribute of a field."

> the tree type is immutable after built

If the tree type is the same thing as a blueprint, then I cannot find a reason why this is immutable. Even the house is built, you can still change the blueprint but not the built house. `ti.SNodeTreeBuilder.init_from(tree_type)` is import, and a builder manipulates the tree type and then `.build()` is export, which is like case 2 I mentioned before, which is fine. And I guess a field can have an attribute that stores sub-tree-type (or partial blueprint as I call it)?

> builder --> tree type --> tree

For this dependency is the dependency in case 2. In case 1, which I prefer, it should be like:
```
tree type -> builder -> tree
```
or (depending on how you define dependency)
```
tree type ↘︎
builder   → tree
```
Either way, I think the point is to store and manipulate the structure information and share it with users (conventional users and library developers) so that they can have a clear picture of the memory layout of a field or many fields *without* knowing the code defining the tree type.


Off topic: 
An interesting and educational application is given a tree type (or a blueprint as I call it), to visualize the conceptual and physical memory layouts of field(s), which may be helpful in profiling as well.

k-ye:
> If the tree type is the same thing as a blueprint, then I cannot find a reason why this is immutable.

IMHO, this is a scenario where [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) is a well-suited solution :-) Immutable types are the more natural way of implementing types in programming languages. Let's think about defining a `struct` in Rust -- we aren't allowed to modify that `struct` type once defined. By making them mutable, we are effectively inventing something like a reflection system with dynamic types, which could complicate things a lot. And in general, immutability is a good thing in programming. That's why rust is immutable by default, and you have to use `&mut` to denote that something is mutable.

I think I understand our arguments better now. When I was thinking about types, I was mainly trying to fix the problem where Taichi has to re-JIT the kernels whenever the field being passed in (via `ti.template()`) changes. By explicitly offering users a SNodeTree type, where they can use for arg annotation, we have an intuitive way to avoid re-JIT now. Your focus is more on the mutability of the tree itself. In this regard, we should provide better utilities to manipulate a tree builder (blueprint).

> An interesting and educational application is given a tree type (or a blueprint as I call it), to visualize the conceptual and physical memory layouts of field(s), which may be helpful in profiling as well.

Yeah, 100% on this one.

docs/rfcs/20220413-aot-for-all-snode.md
bsavery: So overall, it seems to make sense.  However, I find the proposed syntax.... a bit "un-pythonic" to be honest.  

What about something (similar to python `dataclasses`) ?

```
@ti.tree_builder
class MySphereTreeBuilder:
    pos: ti.vec3
    radius: ti.f32
```

where you define the fields as a data structure in the "tree builder" class type

k-ye: Nice suggesteion! I'm not exactly sure how field view is gonna be handled here, but I feel like this is definitely possible to do with the current design.

ifsheldon:
> Your focus is more on the mutability of the tree itself. In this regard, we should provide better utilities to manipulate a tree hierarchy (the blueprint)

Yeah, looking forward to it.

> When I was thinking about types, I was mainly trying to fix the problem where Taichi has to re-JIT the kernels whenever the field being passed in (via ti.template()) changes.

Looking forward to seeing this addressed as well.

> By explicitly offering users a SNodeTree type, where they can use for arg annotation, we have an intuitive way to avoid re-JIT now.

This re-jitting now is implicit, even if you add SNodeTree type. But, can you consider add some utilities for explicitly compile or jit a kernel? I was asked by a student of my collaborator whether he can manually compile a kernel and store it somewhere because he wants different kernels whose loops are unrolled with different static values.

k-ye:
> I was asked by a student of my collaborator whether he can manually compile a kernel and store it somewhere (in Python scope, not exporting a kernel to C scope as in AOT) because he wants different kernels whose loops are unrolled with different static values. 

I think this is possible to do, but be mindful about the potential restraints. E.g. by not letting Taichi manage the JIT, you might encounter errors  when a manually-compiled kernel takes in an incompatible input. This would mostly happen when an arg is a template, though.

PR #4809, opened 2022-04-18
[Lang] Add mod function to math module

neozhaoliang:
This PR add [mod](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/mod.xhtml) function to the `math` module.

### DIFF
--- a/python/taichi/math/mathimpl.py
+++ b/python/taichi/math/mathimpl.py
@@ -404,8 +404,29 @@ def cross(x, y):
     return x.cross(y)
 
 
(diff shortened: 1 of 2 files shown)


rexwangcc (approved):
LGTM!

strongoier (commented):
What is the difference between the new `mod` and the original `%`?

TiGeekMan:
@strongoier `mod` always returns a float vector, even for `int` inputs. I add this function since it would be more familiar for glsl users.

PR #4823, opened 2022-04-20
[bug] Fix bug that building with gcc9.4 will fail

PGZXB:
Related issue = #

### DIFF
--- a/taichi/llvm/llvm_offline_cache.cpp
+++ b/taichi/llvm/llvm_offline_cache.cpp
@@ -66,7 +66,7 @@ get_offline_cache_key_of_compile_config(CompileConfig *config) {
   return serializer.data;
 }
 
-static TI_FORCE_INLINE void get_offline_cache_key_of_snode_impl(
+static void get_offline_cache_key_of_snode_impl(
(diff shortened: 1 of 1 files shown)


g1n0st (approved):
Works for me! Thanks!

k-ye:
What's your original purpose for using TI_FORCE_INLINE?

PGZXB:
I used `TI_FORCE_INLINE` at https://github.com/taichi-dev/taichi/blob/ed478046754caf3cd3524e1f7b275dddb4843b8a/taichi/llvm/llvm_offline_cache.cpp#L15
which is only a internal util function. And this don't cause failed building.

Then I used it at https://github.com/taichi-dev/taichi/blob/ed478046754caf3cd3524e1f7b275dddb4843b8a/taichi/llvm/llvm_offline_cache.cpp#L69
**without thinking**, for similar purpose.

**I'm so sorry, my bad**

k-ye:
OK I see... Let's trust the compilers for now, which should be very good at figuring out what to inline these days. We can manually inline when the profiling results indicate this is a bottleneck :-)

PR #4825, opened 2022-04-20
[mesh] Remove link hints from ti.Mesh

BillXu2000:
Related issue = #3608

All the necessary relation information will be detected in ast transformer.
Relations that has not been stored in fields will be fetched from patcher.

I am not sure whether it is appropriate to store mesh information in compile context and update mesh information in kernel_impl.py.

### DIFF
--- a/.github/workflows/issue_open.yml
+++ b/.github/workflows/issue_open.yml
@@ -1,0 +1,16 @@
+name: Add issues to Taichi project
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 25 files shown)


k-ye (commented):
Hmm, could you make this a separate AST processing pass? The current one is purely for transforming that to C++ AST, which isn't very suitable for blending in Mesh logic

BillXu2000:
Updated. There is now a separate AST processing pass to detect mesh relation visiting, though only some simple cases of detections are handled. For complicated situations, users need to manually tell compiler which relations are needed.

k-ye (approved):
Thanks for separating the AST pass. LGTM from my side!

g1n0st (approved):
LGTM!

PR #4834, opened 2022-04-21
[Bug] [ir] Fix implicit cast warning for global stores

### LINKED ISSUE #4833: Type check fails to warn implicit cast for global stores
**Describe the bug**
The following code snippet should be warned about storing a float to an `i32` field:
```py
import taichi as ti

ti.init()

x = ti.field(dtype=ti.i32, shape=())

@ti.kernel
def bug_gen():
    x[None] = 1.0

bug_gen()
```
However, nothing happens when it is run.

strongoier:
Related issue = fix #4833

The previous logic is buggy because `dst_value_type != stmt->val->ret_type` can never happen after inserting the implicit type cast.


k-ye (approved):
LGTM!

lin-hitonami (approved):
LGTM!

PR #4836, opened 2022-04-22
[lang] Group related Matrix methods closer

k-ye:
Related issue = https://github.com/taichi-dev/taichi/issues/4837

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -1,6 +1,7 @@
 import numbers
 from collections.abc import Iterable
 
+import numpy as np
 from taichi._lib import core as ti_core
 from taichi.lang import expr, impl
 from taichi.lang import ops as ops_mod
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -9,11 +9,11 @@ from taichi.lang import runtime_ops
 from taichi.lang._ndarray import Ndarray, NdarrayHostAccess
 from taichi.lang.common_ops import TaichiOperations
 from taichi.lang.enums import Layout
-from taichi.lang.exception import (TaichiCompilationError, TaichiSyntaxError,
+from taichi.lang.exception import (TaichiCompilationError, TaichiRuntimeError, TaichiSyntaxError,
                                    TaichiTypeError)
 from taichi.lang.field import Field, ScalarField, SNodeHostAccess
 from taichi.lang.swizzle_generator import SwizzleGenerator
-from taichi.lang.util import (cook_dtype, in_python_scope, python_scope,
+from taichi.lang.util import (cook_dtype, in_python_scope, in_taichi_scope, python_scope,
                               taichi_scope, to_numpy_type, to_pytorch_type,
                               warning)
 from taichi.types import primitive_types
@@ -340,7 +340,7 @@ class Matrix(TaichiOperations):
         j = 0 if len(indices) == 1 else indices[1]
         if isinstance(i, slice) or isinstance(j, slice):
             return self._get_slice(i, j)
-        return self(i, j)
+        return self._get_entry_and_read_if_in_pyscope([i, j])
 
     @python_scope
     def __setitem__(self, indices, item):
@@ -364,13 +364,26 @@ class Matrix(TaichiOperations):
             self.entries[idx] = item
 
     def __call__(self, *args, **kwargs):
-        # Can be invoked in both Python and Taichi scope. *args must be
-        # compile-time constants (e.g. Python values)
+        # TODO: It's quite hard to search for __call__, consider replacing this
+        # with a method of actual names?
         assert kwargs == {}
-        ret = self._get_entry(*args)
+        return self._get_entry_and_read_if_in_pyscope(args)
+
+    def _get_entry_and_read_if_in_pyscope(self, indices):
+        # Can be invoked in both Python and Taichi scope. `indices` must be
+        # compile-time constants (e.g. Python values)
+        ret = self._get_entry(*indices)
+
+        def ensure_pyscope():
+            if in_taichi_scope():
+                raise TaichiRuntimeError(
+                    'Cannot read host value in Taichi scope')
+
         if isinstance(ret, SNodeHostAccess):
+            ensure_pyscope()
             ret = ret.accessor.getter(*ret.key)
         elif isinstance(ret, NdarrayHostAccess):
+            ensure_pyscope()
             ret = ret.getter()
         return ret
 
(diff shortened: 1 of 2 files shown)


python/taichi/lang/matrix.py
```diff
@@ -311,7 +311,61 @@ def _linearize_entry_id(self, *args):
             f"The 1-th matrix index is out of range: 0 <= {args[1]} < {self.m}"
         return args[0] * self.m + args[1]
 
+    # host access & python scope operation
+    def __len__(self):
```
k-ye, line 315: This is just moving related methods closer

python/taichi/lang/matrix.py
```diff
@@ -311,7 +311,61 @@ def _linearize_entry_id(self, *args):
...
+            indices (Sequence[Expr]): the indices of a element.
+
+        """
+        if not isinstance(indices, (list, tuple)):
+            indices = [indices]
+        assert len(indices) in [1, 2]
+        i = indices[0]
+        j = 0 if len(indices) == 1 else indices[1]
+        idx = self._linearize_entry_id(i, j)
+        if isinstance(self.entries[idx], SNodeHostAccess):
```
k-ye, line 359: Removed `_set_entry()` because it is only used here

python/taichi/lang/matrix.py
```diff
@@ -400,7 +452,7 @@ def _subscript(self, *indices):
             return impl.make_tensor_element_expr(self.entries[0].ptr, (i, j),
                                                  (self.n, self.m),
                                                  self.dynamic_index_stride)
-        return self(i, j)
+        return self._get_entry(i, j)
```
k-ye, line 455: This is always in Taichi scope.

python/taichi/lang/matrix.py
```diff
@@ -311,26 +311,91 @@ def _linearize_entry_id(self, *args):
...
         assert kwargs == {}
-        ret = self._get_entry(*args)
+        return self._get_entry_and_read_if_in_pyscope(args)
+
+    def _get_entry_and_read_if_in_pyscope(self, indices):
+        # Can be invoked in both Python and Taichi scope. `indices` must be
+        # compile-time constants (e.g. Python values)
+        ret = self._get_entry(*indices)
+
+        def ensure_pyscope():
```
lin-hitonami, line 377: Is this necessary? IIUC, `SNodeHostAccess` and `NdarrayHostAccess` are created only when accessing a matrix field/ndarray in python scope, and they never appear in Taichi scope.

lin-hitonami: Alright, it's possible to be passed in as a template argument, and it goes wrong when the matrix is altered after the first kernel execution. However, this problem also happens when the matrix is not an element of a matrix field...

k-ye: > in python scope, and they never appear in Taichi scope.

Yep, that's what `ensure_pyscope()` is for. I'm afraid I didn't really get your concern..

lin-hitonami: I mean they never appear in Taichi scope (except passed as template arguments), so we don't need to check it.

lin-hitonami: When they are passed as template arguments, they work well for the first time, and then the value in the kernel does not change when the values of the matrix changes, but this is the problem of template argument, not the problem of `Matrix`.

k-ye: > I mean they never appear in Taichi scope

If the code falls into this branch, sure. But note the entire function can still be called in Taichi scope. See for example: https://github.com/taichi-dev/taichi/blob/ecc0521fc506b6a5a4155d92f7642aa2c1701ffd/python/taichi/lang/matrix.py#L257-L259. (Note that this function is originally `__call__()`, which gets called in quite a few places in Taichi scope.) I'm adding this check to 1) make sure the function itself isn't miused, 2) as a better documentation to future developer who needs to touch this part.

lin-hitonami: If we add `ensure_pyscope()` here, we ban using a `Matrix` in `MatrixField` as a template argument, but it shares the same problem with a normal python scope `Matrix` which is not banned.

lin-hitonami: > > I mean they never appear in Taichi scope
> 
> If the code falls into this branch, sure. But note the entire function can still be called in Taichi scope. See for example:
> 
> https://github.com/taichi-dev/taichi/blob/ecc0521fc506b6a5a4155d92f7642aa2c1701ffd/python/taichi/lang/matrix.py#L257-L259
> . (Note that this function is originally `__call__()`, which gets called in quite a few places in Taichi scope.) I'm adding this check to 1) make sure the function itself isn't miused, 2) as a better documentation to future developer who needs to touch this part.

But when it is called inside Taichi scope, it does not cause trouble other than the old problem of template arguments. Why should we ban it? 
IIUC the only circumstance `self` can contain `SNodeHostAccess` is that a matrix in Matrix field is passed as a template argument?

k-ye: > Why should we ban it?

Because this just adds more confusion. `SNodeHostAccess` is just not supposed to exist in Taichi scope... If that happens, we should fix that 😂 

> it is a global variable?

Hmm.. how so?

lin-hitonami: Like this.
```
import taichi as ti

ti.init()

a = ti.Matrix.field(n=2, m=2, dtype=ti.f32, shape=(10,))

b = a[0]

@ti.kernel
def foo():
    print(b[0, 0])

foo()
```

k-ye: As discussed, removed `ensure_pyscope`

python/taichi/lang/matrix.py
```diff
@@ -311,26 +310,84 @@ def _linearize_entry_id(self, *args):
...
+            self.entries[idx] = item
+
     def __call__(self, *args, **kwargs):
+        # TODO: It's quite hard to search for __call__, consider replacing this
+        # with a method of actual names?
         assert kwargs == {}
-        ret = self._get_entry(*args)
+        return self._get_entry_and_read_if_in_pyscope(args)
+
+    def _get_entry_and_read_if_in_pyscope(self, indices):
```
lin-hitonami, line 371: Should we rename this function since `ensure_pyscope` is removed?

k-ye: Good point. I'm thinking about calling it `_get_entry_and_read`, which sounds a bit verbose, but I feel like this does have two steps:

1. get means to get `entries[idx]`
2. read means to load the value of that entry, if possible

WDYT?

lin-hitonami: LGTM

lin-hitonami (approved):
LGTM!

PR #4841, opened 2022-04-22
[Lang] Build sparse matrix from ndarray

FantasyVR:
Related issue = #2906

In this draft, The sparse matrix is directly built from `ndarray`.  In the future, we could get rid of `SparseMatrixBuilder`.


taichi/program/sparse_matrix.cpp
```diff
@@ -162,5 +162,19 @@ void SparseMatrix::set_element(int row, int col, float32 value) {
   matrix_.coeffRef(row, col) = value;
 }
 
+SparseMatrix &SparseMatrix::build_from_ndarray(const Ndarray &ndarray) {
+  using T = float32;
+  using V = Eigen::Triplet<T>;
+  std::vector<V> triplets;
+  T *data = reinterpret_cast<T *>(ndarray.get_data_ptr_as_int());
```
k-ye, line 169: Note that this won't really work other than on the CPU backend...

python/taichi/linalg/sparse_matrix.py
```diff
@@ -147,10 +148,22 @@ def __str__(self):
...
+    def build_from_ndarray(self, ndarray):
+        """Build the sparse matrix from a ti.Vector.ndarray.
+
+        Args:
+            ndarray (~taichi.lang.matrix.VectorNdarray): the ndarray to build the sparse matrix from.
+        """
+        if isinstance(ndarray, VectorNdarray):
+            self.matrix.build_from_ndarray(ndarray.arr)
+        else:
+            assert False, "Sparse matrix only supports building from ti.Vector.ndarray"
```
strongoier, line 165: raise `TaichiRuntimeError`?

taichi/program/sparse_matrix.h
```diff
@@ -57,7 +58,9 @@ class SparseMatrix {
   virtual ~SparseMatrix() = default;
 
   virtual void build_triplets(void *triplets_adr){};
-
+  virtual SparseMatrix &build_from_ndarray(const Ndarray &ndarray) {
+    return *this;
+  }
```
strongoier, line 63: This seems a bit weird to me.. Should `SparseMatrix` be an abstract class?

FantasyVR: Thanks,  `build_from_ndarray` is moved outside of `SparseMatrix` in the new commit.

misc/sm_ndarray.py
```diff
@@ -0,0 +1,30 @@
+import taichi as ti
+
+ti.init(arch=ti.cpu, gdb_trigger=True, dynamic_index=True)
```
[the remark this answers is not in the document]
FantasyVR: Thanks. It's removed in the new commit.

misc/sm_ndarray.py
```diff
@@ -0,0 +1,28 @@
+import taichi as ti
+
+ti.init(arch=ti.cpu, gdb_trigger=True)
+
```
Hanke98, line 4: Why do we need `gdb_trigger` here?

taichi/program/sparse_matrix.cpp
```diff
@@ -162,5 +162,34 @@ std::unique_ptr<SparseMatrix> make_sparse_matrix(
...
+  sm.build_triplets(static_cast<void *>(&triplets));
+}
+
+void make_sparse_matrix_from_ndarray(Program *prog,
+                                     SparseMatrix &sm,
+                                     const Ndarray &ndarray) {
+  std::string sdtype = taichi::lang::data_type_name(sm.get_data_type());
+  auto data_ptr = prog->get_ndarray_data_ptr_as_int(&ndarray);
+  auto num_triplets = ndarray.get_nelement() / 3;
+  if (sdtype == "f32") {
```
Hanke98, line 185: What if the ndarray is not in shape of (N, 3)? Is the shape of (1, 3xN) or (3xN, 1) acceptable? Should we check the shape of ndarray before building the matrix? :)

Hanke98 (commented):
Great job and sorry for the late response! Most changes LGTM! I only have two points that are not very clear shown in below.

python/taichi/linalg/sparse_solver.py
```diff
@@ -24,11 +25,15 @@ def __init__(self, dtype=f32, solver_type="LLT", ordering="AMD"):
             self.solver = _ti_core.make_sparse_solver(dtype, solver_type,
                                                       ordering)
         else:
-            assert False, f"The solver type {solver_type} with {ordering} is not supported for now. Only {solver_type_list} with {solver_ordering} are supported."
+            TaichiRuntimeError(
```
strongoier, line 28: You need to `raise` the error.

taichi/program/sparse_matrix.cpp
strongoier: The way to interpret data is subjective. We interpret data as "[row, col, value], [row, col, value]", and users need to follow it. IMO having it documented is enough.

PR #4842, opened 2022-04-22
[Lang] Add single character property setter and getter for Matrix

neozhaoliang:


python/taichi/lang/matrix.py
```diff
@@ -23,12 +24,30 @@
 def _gen_swizzles(cls):
     swizzle_gen = SwizzleGenerator()
     # https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Swizzling
-    KEMAP_SET = ['xyzw', 'rgba', 'stpq']
-    for key_group in KEMAP_SET:
+    KEYMAP_SET = ['xyzw', 'rgba', 'stpq']
+
+    def add_single_swizzle_attrs(cls):
```
k-ye, line 29: Cool! Could you also add some tests for it? E.g. `a[I].r`, `b[I].p`, etc

PR #4843, opened 2022-04-22
[Bug] [ir] Fix and refactor type check for local stores

### LINKED ISSUE #4678: Reassign of vector/variable with different types
**Describe the bug**
When reassigning a variable/vector with a different type, the original type does not change, this may lead to confusion as people are easier to think taichi works similar to python (but not C)

**To Reproduce**

```py
import taichi as ti
@ti.kernel
def test_kernel2():
    p0 = ti.Vector([1, 2, 3], ti.i32)
    p1 = ti.Vector([8, 6, 4], ti.i32)
    print(p0)
    mu = 0.5
    p = p0
    p = p0 + mu*(p1-p0)
    print("interlop", p)
    p = ti.Vector([1.4, 2.4, 3.2], ti.f32)
    print("reassign", p)

    a = 1
    print(a)
    a = 1.5
    print(a)
ti.init(arch=ti.cpu)
test_kernel2()
```

**Log/Screenshots**

```
➜  TaichiSLAM git:(main) ✗ python tests/test.py
[Taichi] version 0.9.3, llvm 10.0.0, commit c902db62, linux, python 3.8.10
[Taichi] Starting on arch=x64
[W 03/30/22 22:38:51.391 3010] [type_check.cpp:visit@145] [$183] Local store may lose precision (target = i32, value = f32), at
183
[W 03/30/22 22:38:51.391 3010] [type_check.cpp:visit@145] [$185] Local store may lose precision (target = i32, value = f32), at
185
[W 03/30/22 22:38:51.391 3010] [type_check.cpp:visit@145] [$187] Local store may lose precision (target = i32, value = f32), at
187
[W 03/30/22 22:38:51.391 3010] [type_check.cpp:visit@145] [$202] Local store may lose precision (target = i32, value = f32), at
202
[W 03/30/22 22:38:51.391 3010] [type_check.cpp:visit@145] [$204] Local store may lose precision (target = i32, value = f32), at
204
[W 03/30/22 22:38:51.391 3010] [type_check.cpp:visit@145] [$206] Local store may lose precision (target = i32, value = f32), at
206
[W 03/30/22 22:38:51.391 3010] [type_check.cpp:visit@145] [$217] Local store may lose precision (target = i32, value = f32), at
217
[1, 2, 3]
interlop [4, 4, 3]
reassign to float [1, 2, 3]
1
1.5 1
```

**Additional comments**
My suggestion is to either of:
- Gives a clear warning or error indication that the line has Implicit converting. now we can't find which line the error is on.
- Convert lvalue to the type of rvalue (like Python)

strongoier:
Related issue = fix #4678

This PR refactors type check for local stores to make the logic concise and clear. The remaining issue in #4678 is caused by printing a wrong argument, which is fixed by the way.

### DIFF
--- a/taichi/transforms/type_check.cpp
+++ b/taichi/transforms/type_check.cpp
@@ -102,48 +102,21 @@ class TypeCheck : public IRVisitor {
   }
 
   void visit(LocalStoreStmt *stmt) override {
-    if (stmt->dest->is<PtrOffsetStmt>() &&
-        stmt->dest->cast<PtrOffsetStmt>()->is_local_ptr()) {
-      auto dst_value_type = stmt->dest->ret_type.ptr_removed();
-      if (dst_value_type->is<CustomIntType>() ||
-          dst_value_type->is<CustomFloatType>()) {
-        // We force the value type to be the compute_type of the bit pointer.
-        // Casting from compute_type to physical_type is handled in codegen.
-        dst_value_type = dst_value_type->get_compute_type();
-      }
-      auto promoted = promoted_type(dst_value_type, stmt->val->ret_type);
-      auto input_type = stmt->val->ret_data_type_name();
-      if (dst_value_type != stmt->val->ret_type) {
-        stmt->val = insert_type_cast_before(stmt, stmt->val, dst_value_type);
-      }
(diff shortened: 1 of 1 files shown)


ailzhang (approved):
LGTM!

PR #4844, opened 2022-04-22
[Error] Show warning when serialize=True is set on a struct for

### LINKED ISSUE #4807: Using `ti.loop_config(serialize=True)` before `ti.grouped()` leads to a wrong result
**Describe the bug**
When using `ti.loop_config(serialize=True)` before `ti.grouped()`, I will get a wrong execution result. If we do not support such usage, maybe we should point out this in the doc, or give an error message at the compilation stage.

**To Reproduce**

A prefix sum example is shown as below:

```py
import taichi as ti
ti.init(arch=ti.cuda)

@ti.kernel
def scan_bad(arr: ti.template(), sum: ti.template()):
    sum[0] = arr[0]
    ti.loop_config(serialize=True)
    for I in ti.grouped(arr):
        if I.x > 0:
            sum[I] = arr[I] + sum[I-1]

@ti.kernel
def scan_good(arr: ti.template(), sum: ti.template()):
    sum[0] = arr[0]
    ti.loop_config(serialize=True)
    for i in range(3):
        if i > 0:
            sum[i] = arr[i] + sum[i-1]

arr = ti.field(ti.i32, 3)
sum = ti.field(ti.i32, 3)
arr[0] = 1
arr[1] = 1
arr[2] = 1

scan_bad(arr, sum)      # sum: [1 2 1]
# scan_good(arr, sum)   # sum: [1 2 3]
print(sum)
```

lin-hitonami:
Related issue = fix #4807

### DIFF
--- a/taichi/ir/frontend_ir.cpp
+++ b/taichi/ir/frontend_ir.cpp
@@ -820,6 +820,9 @@ void ASTBuilder::begin_frontend_range_for(const Expr &i,
 
 void ASTBuilder::begin_frontend_struct_for(const ExprGroup &loop_vars,
                                            const Expr &global) {
+  TI_WARN_IF(for_loop_dec_.config.strictly_serialized,
+             "ti.loop_config(serialize=True) does not have effect on the struct for. "
+             "The execution order is not guaranteed.");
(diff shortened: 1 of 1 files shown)


k-ye (approved):
LGTM!

Issue #4851, opened 2022-04-23
[doc][RFC] Semi-formal Unified Device API specification


bobcao3:
We need a semi-formal specification of this API so that people can extend it or use it and expect it to be consistent on all backends.

It is also a good time to address necessary design and specs changes for unified device API, therefore I'm recording this first in an issue/RFC before we merge all the ideas into a document.

The design listed here are not complete, (thus RFC)

@AmesingFlank @k-ye @qiao-bo @ailzhang 

---

Here is the current design (with some upcoming change regarding Stream synchronization behavior, and the current device API implementation is not yet fully coherent with this design)

### Device

In unified device API, `Device` is the master of all objects.
- A `Device` usually represent a single graphics / compute adapter available on the system, this is mapped usually to a single physical device unless the backend API provides transparent multi-device support (e.g. CPU NUMA, or Nvidia SLI).
- All resources and objects are created from the `Device` instance. These objects include `DeviceAllocation`, `Pipeline`, `Stream`, and `Surface`.
- Calls to Device should be thread-safe unless the backend API is not thread-safe (e.g. OpenGL). For our DirectX, Vulkan, Metal, and possibly CUDA implementations, unified device API should be thread-safe. As for concurrency and dependency management, please refer to the `Stream` section.

### `DeviceAllocation` aka. Resource model

In unified device API, `DeviceAllocation` represents an allocated resource on the computing device. Its actual memory location can be elsewhere or undefined, many objects are entirely opaque.

Currently `DeviceAllocation` represents either a buffer or an image. In the future, this can also represent ray tracing acceleration structures, encoded video packets, etc.

`DeviceAllocation` that represents a buffer can use `get_pointer` call to create a offset pointer into the buffer with type `DevicePointer`

Beware that, `DeviceAllocation` can come from another device. In this case we have interop APIs that can copy contents from another device's `DeviceAllocation` to this device, currently we do not support using foreign `DeviceAllocation` directly.

The allocation and deallocation control of `DeviceAllocation` is explicit. However, if the data is currently in use by the GPU, the deallocation will happen after the GPU work is done. (i.e. resource tracking is done by the API, the user don't need to worry about it)

Note that the backend implementation that tracks the resource usage might depend on the use of synchronization calls such as `Surface::present_image`, `Stream::submit_sync`, `Stream::command_sync`, and `Device::wait_idle`. Without these calls, the backend might fail to track resource usages and thus fail to garbage collect.

### `Stream`

[...]

PR #4852, opened 2022-04-23
[vulkan] Device API explicit semaphores

bobcao3:
Related issue = #4851


taichi/backends/device.h
```diff
@@ -397,13 +397,26 @@ inline bool operator&(AllocUsage a, AllocUsage b) {
   return static_cast<int>(a) & static_cast<int>(b);
 }
 
+class StreamSemaphoreObject {
```
k-ye, line 400: Why Stream in this naming? (Initially I thought we had a stream of semaphores..)

taichi/backends/device.h
```diff
@@ -498,8 +514,9 @@ class Surface {
   virtual ~Surface() {
   }
 
-  virtual DeviceAllocation get_target_image() = 0;
-  virtual void present_image() = 0;
+  virtual std::pair<DeviceAllocation, StreamSemaphore> get_target_image() = 0;
```
k-ye, line 517: not: Better create a small struct instead of std::pair, otherwise it's arcane to future readers when there are many first, second...

taichi/backends/vulkan/vulkan_device.cpp
```diff
@@ -1514,14 +1525,18 @@ void VulkanStream::submit(CommandList *cmdlist_) {
...
 
-  if (last_semaphore_) {
-    submit_info.waitSemaphoreCount = 1;
-    submit_info.pWaitSemaphores = &last_semaphore_->semaphore;
-    submit_info.pWaitDstStageMask = &stage_flag;
+  std::vector<VkSemaphore> vk_wait_semaphores;
+
+  for (const StreamSemaphore &sema_ : wait_semaphores) {
+    auto sema = std::static_pointer_cast<VulkanStreamSemaphoreObject>(sema_);
+    vk_wait_semaphores.push_back(sema->semaphore->semaphore);
```
ailzhang, line 1532: `sema->semaphore->semaphore` is a bit confusing :P

ailzhang (commented):
Thanks!

taichi/backends/device.h
bobcao3: Just that on Windows Semaphore is a system type, will conflict with Win32 API

taichi/backends/vulkan/vulkan_device_creator.cpp
```diff
@@ -18,7 +18,7 @@ namespace vulkan {
 namespace {
 
 // FIXME: NDEBUG is broken, so just manually enable this if necessary.
-constexpr bool kEnableValidationLayers = false;
+constexpr bool kEnableValidationLayers = true;
```
k-ye, line 21: change back?

taichi/backends/vulkan/vulkan_device.cpp
k-ye: Yeah, the inner one could probably be something like `rep`, `ptr`, `raw`, ..

taichi/backends/vulkan/runtime.cpp
```diff
@@ -157,7 +158,7 @@ class HostDeviceContextBlitter {
     }
 
     if (require_sync) {
-      device_->get_compute_stream()->submit_synced(cmdlist);
+      device_->get_compute_stream()->submit_synced(cmdlist, wait_semaphore);
```
k-ye, line 161: Do we need to use the sema returned by `submit_synced`?

taichi/backends/dx/dx_device.cpp
```diff
@@ -805,15 +808,23 @@ std::unique_ptr<CommandList> Dx11Stream::new_command_list() {
...
 
 // No difference for DX11
-void Dx11Stream::submit_synced(CommandList *cmdlist) {
+StreamSemaphore Dx11Stream::submit_synced(
+    CommandList *cmdlist,
+    const std::vector<StreamSemaphore> &wait_semaphores) {
   Dx11CommandList *dx_cmd_list = static_cast<Dx11CommandList *>(cmdlist);
   dx_cmd_list->run_commands();
+
+  return std::make_shared<StreamSemaphoreObject>();
```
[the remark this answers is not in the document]
bobcao3: Makes sense...

taichi/backends/vulkan/runtime.cpp
bobcao3: Not in this case. In the case where we are on multiple queues than `submit_synced` only syncs that one queue, and we need the semaphore for cross queue sync

taichi/backends/vulkan/vulkan_device.cpp
bobcao3: The innermost one is under `vkapi` i.e. our shared ptr Vulkan wrapper, most RHIs have the tradition of using object names for those. The center one should probably be something like `vk_ref`...

taichi/ui/backends/vulkan/renderer.cpp
```diff
@@ -125,11 +126,14 @@ void Renderer::prepare_for_next_frame() {
 }
 
 void Renderer::draw_frame(Gui *gui) {
+  app_context_.prog()->synchronize();
```
ghuau-innopeak, line 129: We need to check if the prog() is not nullptr. In case of AOT module, we don't have a program. I believe this is up to the host application to call synchronize() before the draw() call...

bobcao3: nice catch

taichi/backends/vulkan/vulkan_device.cpp
```diff
@@ -1507,67 +1530,66 @@ void VulkanStream::submit(CommandList *cmdlist_) {
...
+  auto fence = vkapi::create_fence(buffer->device, 0);
 
-  VkSubmitInfo submit_info{};
-  submit_info.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
-  submit_info.commandBufferCount = 1;
-  submit_info.pCommandBuffers = &buffer->buffer;
-
-  VkPipelineStageFlags stage_flag{VK_PIPELINE_STAGE_ALL_COMMANDS_BIT};
+  // Resource tracking, check previously submitted commands
+  // FIXME: Figure out why it doesn't work
```
k-ye, line 1561: OOC, how exactly is it not working?

k-ye (approved):
LGTM!

taichi/backends/vulkan/vulkan_device.cpp
bobcao3: It's crashing randomly after a while, we might have forgotten some references counting somewhere.

k-ye:
BTW, let's put this one on hold for a while until v1.0.1 https://github.com/taichi-dev/taichi/milestone/6 is released (ETA tomorrow)

Issue #4837, opened 2022-04-22
Simplify Taichi's Matrix implementation


k-ye:
**Concisely describe the proposed feature**

Right now Taichi's matrix implementation takes the combination of all these factors into account:

* Python vs Taichi scope
* Local or Global (i.e., Matrix within a field/Ndarray)
* Dynamic indexing or not
* Slicing or not

I'll see if there's a way to decouple the Mat impl. (cc @strongoier )

k-ye:
I'm thinking of instantiating different types of Matrix implementations, depending on whether it is created inside Python or Taichi scope. Here are three ways I have come up with

1. Make `ti.Matrix` a (factory) function, pseudo code below:

```py
def Matrix(*args):
  if in_python_scope():
    return _PyMatImpl(*args)
  else:
    return _TiMatImpl(*args)
```

The downside is that, many places are already doing this check: `isinstance(m, ti.Matrix)`. So changing `Matrix` to a function could potentially break many things.

2. Let `ti.Matrix` have an `_impl` field, whereas `_impl` can be `_PyMatImpl` or `_TiMatImpl`. This can group operations of the same scope to the same class, which is great. Then `Matrix` is mostly a wrapper around `_impl`. The downside is that the code becomes more verbose, in that `Matrix` still has to implement the union of APIs supported in both Python and Taichi scopes. E.g.

```py
class _PyMatImpl:
  def op1_py_only(self):
    ...

class _TiMatImpl:
  def op2_ti_only(self):
    ...

class Matrix:
  def op1(self):
    self._impl.op1_py_only()

  def op2(self):
    self._impl.op2_py_only()
```

3. Adopting `__new__`

Python's `__init__` cannot return anything, but there's an escape hatch: `__new__`. See these issues of how to return subclasses from a seemingly base class's initializer:

* https://stackoverflow.com/questions/18218876/how-to-return-a-subclass-from-new-parameter
* https://stackoverflow.com/questions/28236516/python-subclassing-a-class-with-custom-new

This approach solves both 1 and 2. However, I'm not very familiar with `__new__`, and wonder in general, if this is a good use case. Any comment will be very helpful here! @Zheaoli @frostming

frostming:
@k-ye All three practices can be seen in real-world projects, but have trade-offs.

1. If `isinstance` is widely used then we will discard this option
2. This one looks the cleanest to me.
[...]

PR #4853, opened 2022-04-24
[refactor] Create MatrixImpl to differentiate Taichi and Python scopes

k-ye:
Related issue = #https://github.com/taichi-dev/taichi/issues/4837


k-ye:
By separating Taichi and Python scope's matrix implementation, we've found these places where Py and Taichi matrices are mixed:

1. https://github.com/taichi-dev/taichi/blob/b6408f8ea733c95f0504a3b7bf3befbf0b0891bb/tests/python/test_ast_refactor.py#L179-L206

```
tests/python/test_ast_refactor.py:204: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = ([1 1 2], [2 1 1], <ti.field>), kwargs = {}

    @functools.wraps(_func)
    def wrapped(*args, **kwargs):
        try:
        except (TaichiCompilationError, TaichiRuntimeError) as e:
>           raise type(e)('\n' + str(e)) from None
E           taichi.lang.exception.TaichiCompilationError: 
E           On line 184 of file "/home/dev/tests/python/test_ast_refactor.py", in foo:
E                       c[i * 6] = a[i] == b[i]
E                                  ^^^^
E           Traceback (most recent call last):
E             File "/home/dev/miniconda/envs/py310/lib/python3.10/site-packages/taichi/lang/ast/ast_transformer_utils.py", line 24, in __call__
E               return method(ctx, node)
E             File "/home/dev/miniconda/envs/py310/lib/python3.10/site-packages/taichi/lang/ast/ast_transformer.py", line 186, in build_Subscript
E               node.ptr = impl.subscript(node.value.ptr, *node.slice.ptr)
E             File "/home/dev/miniconda/envs/py310/lib/python3.10/site-packages/taichi/lang/util.py", line 216, in wrapped
E               return func(*args, **kwargs)
E             File "/home/dev/miniconda/envs/py310/lib/python3.10/site-packages/taichi/lang/impl.py", line 134, in subscript
E               return value._subscript(*_indices)
E             File "/home/dev/miniconda/envs/py310/lib/python3.10/site-packages/taichi/lang/util.py", line 216, in wrapped
E               return func(*args, **kwargs)
E             File "/home/dev/miniconda/envs/py310/lib/python3.10/site-packages/taichi/lang/matrix.py", line 550, in _subscript
E               return self._impl._subscript(is_global_mat, indices)
E           AttributeError: '_PyScopeMatrixImpl' object has no attribute '_subscript'
```

This was what @lin-hitonami was talking about, by passing a Python scope matrix into a Taichi kernel as a `ti.template()`

3. https://github.com/taichi-dev/taichi/blob/b6408f8ea733c95f0504a3b7bf3befbf0b0891bb/tests/python/test_mesh.py#L113-L147

```
___________________________________________ test_mesh_reorder[arch=x64-1] ____________________________________________

    @test_utils.test(require=ti.extension.mesh)
    def test_mesh_reorder():
        vec3i = ti.types.vector(3, ti.i32)
        mesh_builder = ti.Mesh.Tet()
        mesh_builder.verts.place({'s': ti.i32, 's3': vec3i}, reorder=True)
        mesh_builder.cells.link(mesh_builder.verts)
        model = mesh_builder.build(ti.Mesh.load_meta(model_file_path))
    
        id2 = np.array([x**2 for x in range(len(model.verts))])
        id123 = np.array([[x**1, x**2, x**3] for x in range(len(model.verts))])
        model.verts.s.from_numpy(id2)
        model.verts.s3.from_numpy(id123)
    
        @ti.kernel
        def foo():
            for v in model.verts:
                assert v.s == v.id**2
                assert v.s3[0] == v.id**1 and v.s3[1] == v.id**2 and v.s3[
                    2] == v.id**3
                v.s = v.id**3
                v.s3 *= v.id
    
        foo()
    
        id3 = model.verts.s.to_numpy()
        id234 = model.verts.s3.to_numpy()
    
        for i in range(len(model.verts)):
            assert model.verts.s[i] == i**3
            assert id3[i] == i**3
>           assert model.verts.s3[i][0] == i**2

tests/python/test_mesh.py:142: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
python/taichi/lang/util.py:226: in wrapped
    return func(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = [0 0 0], indices = 0

    @python_scope
    def __getitem__(self, indices):
        """Access to the element at the given indices in a matrix.
    
        Args:
            indices (Sequence[Expr]): the indices of the element.
    
        Returns:
            The value of the element at a specific position of a matrix.
    
        """
>       return self._impl[indices]
E       TypeError: '_TiScopeMatrixImpl' object is not subscriptable
```

k-ye:
@lin-hitonami I've bumped into the problem where we pass a Python-scope matrix into a Taichi kernel as a `ti.template()`. https://github.com/taichi-dev/taichi/blob/b6408f8ea733c95f0504a3b7bf3befbf0b0891bb/tests/python/test_ast_refactor.py#L179-L206. What in your opinion would be the best approach to convert Python matrix to Taichi matrix? (I feel like it should be somewhere in the AST transformer?)

lin-hitonami:
> . What in your opinion would be the best approach to convert Python matrix to Taichi matrix? (I feel like it should be somewhere in the AST transformer?)

I think we can convert the matrix in `build_FunctionDef` for template arguments. 

Here are some points we should pay attention to:
1. We should make sure the converted matrix is read-only.
2. We should also convert matrices used in the kernel which are global variables.

For the first point, we should distinguish read and write access of the subscription operation, like `__getitem__` and `__setitem__` in Python. 

For the second point, we can convert the matrix which is a global variable in `build_Name`, and register the converted matrix as a variable using `ctx.create_variable`. 

However, this leads to another problem: we now allow creating a variable after reading a global variable with the same name, which Python does not allow (Python throws `UnboundLocalError`). However, if we register the matrix as a variable, we cannot create a variable with the same name as the matrix. Therefore I suggest we ban this as Python does.

strongoier:
Maybe I'm lacking context here. Why do we "convert" a Python matrix to a Taichi matrix in this case? IMO a Python matrix is just like any other global variable, which is read-only and evaluated completely at compile time.

k-ye:
> Therefore I suggest we ban this as Python does.

Until we figure out a better approach, yeah I agree.

> Why do we "convert" a Python matrix to a Taichi matrix in this case?

See https://github.com/taichi-dev/taichi/issues/4837#issuecomment-1106599720. `ti.Matrix` has an `_impl`, which will either be a PyScopeImpl or a TiScopeImpl. When you pass a PyScopeImpl-matrix into Taichi kernel, the correct behavior should be that it makes a copy and converts to a TiScopeImpl.

k-ye:
> Considering that all Python variables (ints, lists, numpy arrays, ...) are handled this way in Taichi, why do we want to handle Mat differently? 

That's a good point. Considering that we can pass matrix by value (via `ti.types.mat(...)), maybe we should just ban this? Then the definition of `ti.template()` becomes "those Python objects that can be used inside Taichi kernel". WDYT?

> TBH if it is possible to break existing code, I would suggest giving PyMat and TiMat different names to get things clear...

I think having separate names, so that users would have to write something like this, is not an option. It simply puts too much on users shoulder.

```py
a = ti.PyMatrix()

@ti.kernel
def foo():
  b = ti.TiMatrix()
```

My initial intention is to separate *the implementaion of `ti.Matrix`*, so that we as devs don't have to jump around the code to know how Matrix works.

strongoier:
> maybe we should just ban this?

Banning this and banning access to global variable `PyMat`s should happen at the same time. However as far as I remember, accessing global variable `PyMat` is a common practice in existing Taichi programs...

lin-hitonami (approved):
LGTM!

PR #4862, opened 2022-04-26
[metal] Complete Device API

k-ye:
Related issue = https://github.com/taichi-dev/taichi/issues/2736

### DIFF
--- a/taichi/backends/metal/device.cpp
+++ b/taichi/backends/metal/device.cpp
@@ -34,7 +34,8 @@ class ResourceBinderImpl : public ResourceBinder {
                  uint32_t binding,
                  DevicePtr ptr,
                  size_t size) override {
-    TI_NOT_IMPLEMENTED;
+    TI_ERROR(
+        "DevicePtr not supported, please use the DeviceAllocation overload");
   }
   void rw_buffer(uint32_t set,
                  uint32_t binding,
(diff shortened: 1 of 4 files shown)


taichi/backends/metal/device.cpp
```diff
@@ -34,7 +34,8 @@ class ResourceBinderImpl : public ResourceBinder {
                  uint32_t binding,
                  DevicePtr ptr,
                  size_t size) override {
-    TI_NOT_IMPLEMENTED;
+    TI_ERROR(
```
bobcao3, line 37: Is this a TODO or a not supported feature in Metal?

k-ye: TODO

k-ye: More specifically, Metal has this API where you can set `offset`, but not `size`... https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/1443126-setbuffer?language=objc

bobcao3: This is actually fine, we can ignore size.

bobcao3: In vulkan when offset+size go over the buffer size it's undefined behavior anyways

k-ye: Fixed

bobcao3 (approved):
LGTM

Issue #4832, opened 2022-04-21
[RFC] Taichi CMake Overhaul


qiao-bo:
We would like to share our proposal for modernizing Taichi's CMake-based build system. By embracing the target-based approach, we can enforce a good modular design in our code base. This brings us benefits such as reduced (re-)compilation time. By being explicit and untangling the dependencies in current code base, code (target) that do not depend on others can be isolated and built independently (target level parallelism). Changes on one target would not incur the rebuild of others. 

This is ongoing work. Previously some efforts have been spent on cleaning up the code base:
* #2196
* #2203
* #2195

## Proposal
The proposed changes mainly consist of two parts: First, we would like to maintain a list of targets that Taichi uses and being built upon. Each of these targets should have its own `CMakeList.txt` that specifies the _build requirements_ as well as the _usage requirements_ of this target. This may require some shuffling in our code base. Second, replacing many of our current CMake functions that have a global scope with target-based APIs. For example, `include_directories` should be replaced with `target_include_directories` to reduce hidden dependencies on header files to its minimum. Here, we share the outcome of some preliminary discussions:

### Targets
Current Taichi C++ code base (https://github.com/taichi-dev/taichi/tree/master/taichi) could be split into the following build targets (names TBD):
1. `Program`: Core module that calls other targets in order. (https://github.com/taichi-dev/taichi/tree/master/taichi/program)
2. `IR`: Include IR, Analysis, Transform. This may depend on `Type`, `SNode` etc.
3. `Codegen`: Current location is (https://github.com/taichi-dev/taichi/tree/master/taichi/backends), We can further divide into `LlvmCodegen` and `SpirvCodegen`. `Codegen` should depend on `Runtime`.
4. `Runtime`: At the moment shares the same location as `Codegen`. Need code shuffling to split out.
5. `Artifact`: This information is required by both `Codegen` and `Runtime`. One example in the context of AOT is KernelAttributes:
https://github.com/taichi-dev/taichi/blob/5b890f90f9edb90c2f1b4841ea7142f4b13e0bb2/taichi/backends/metal/kernel_utils.h#L105-L156
As a first step, we can move this information from `Codegen` to `Runtime` such that `Runtime` does not depend on `Codegen`.
9. `Common`: Include logging, macro and such things. This target is shared among all targets.
10. `Type`: Many targets may depend on this.
[...]

PR #4864, opened 2022-04-26
[Build] [refactor] Use keywords instead of plain target_link_libraries CMake

qiao-bo:
Related issue = #4832 

- The [old style plain function `target_link_libraries` ](https://cmake.org/cmake/help/v3.19/command/target_link_libraries.html#libraries-for-both-a-target-and-its-dependents)could have both immediate and transitive effects (not always behave like PUBLIC). It is recommended to use keywords (namely `PRIVATE`, `INTERFACE`, and `PUBLIC`) to explicitly guard the dependencies and interface. 
- Avoid abusing `target_link_libraries()`. This function should only be used to specify dependencies among targets. CMake provide explicit functions for linking, options, flags, and definitions, etc. Some of them can be implemented using `target_link_libraries()` is merely because of the backward portability CMake tries to provide.


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # The Taichi Programming Language
 #*********************************************************************
 
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.13)
```
k-ye, line 5: Do we need this?

k-ye (approved):
LGTM!

CMakeLists.txt
qiao-bo: yep, [target_link_options](https://cmake.org/cmake/help/latest/command/target_link_options.html) is not available before 3.13. Given 3.13 was more than 3 years old (released in 2019), it is ok to be minimum required ;/

Issue #4824, opened 2022-04-20
[ci] Improve docker image for Linux CPU build and test 


qiao-bo:
Currently, we employ an Ubuntu 18.04 based docker image (Dockerfile [here](https://github.com/taichi-dev/taichi/blob/master/ci/Dockerfile.ubuntu.18.04.cpu)) for building and testing Taichi's CPU backend. The current image is approximately 12GB, we would like improve this image by reducing the image size to its minimal. 

Things to consider include: 
- Familiar with Taichi's build environment and workflows.
- Optimize Dockerfile by following some of the best practices in building (For example enable multi-stage build and minimize number of layers, etc) and making use of optimization tools such as [Dive](https://github.com/wagoodman/dive) and [Docker-slim](https://github.com/docker-slim/docker-slim).
- Ideally we want to reduce the image by 50%, and being tested on Ubuntu and Arch linux.

After exploring the previous steps, we can also move on to optimize other backends:
- Improve docker image for GPU build and test
- Improve docker image for CPU build and test (Windows)
- Combine the multi-stage build approach with our generator scripts.

frostming:
If we switch to the `container` setting in workflow(#4776 ), we can get rid of the conda setup since it can be done by a [Github Action](https://github.com/marketplace/actions/setup-miniconda)

qiao-bo:
> If we switch to the `container` setting in workflow(#4776 ), we can get rid of the conda setup since it can be done by a [Github Action](https://github.com/marketplace/actions/setup-miniconda)

that would be great! conda environment occupies quite some space.

PR #4869, opened 2022-04-27
[ci] Improve docker image for Linux CPU build

xUhEngwAng:
Related issue = #4824 

I checked the ubuntu dockerfile, combined multiple RUN commands into one to reduce the number of image layers, and removed the conda setup phase. The resulting image takes only 2.5GB. As for the python env, the corresponding python version will be installed during the build&test workflow, which can spare a lot of space. 

I didn't apply the multi-stage build mechanism as it seems not suitable for our usage scenario. BTW, in the previous docker image, the conda env takes up to 7GB, due to which it's hard to achieve the 50% optimization goal without touching this part.

Future works include:

+ Switch to Github Action to install the corresponding python env as it seems more flexible and less error-prone.
+ Apply and test the same strategy on other backends.
+ Update the generator script to account for the above changes.


qiao-bo:
Hi, thanks a lot for looking into this and apologize for my late reply.

Indeed the conda env occupies quite some space. However, we cannot move this setup to the build script. Because that will increase significantly the building time. Using the container setting in workflow as suggested [here](https://github.com/taichi-dev/taichi/issues/4824#issuecomment-1103620380) might worth a shot, since workflow installation is generally fast. Nevertheless, we still need to carefully exploit the tradeoff here between image size and build time ;)

Multi stage build effectively do the things like combining multiple RUN command. But perhaps we can also take advantage of what the other native image can offer, such as [this conda base image](https://hub.docker.com/r/conda/miniconda3/tags) is of size 142 MB, I wonder if this is smaller than what we build ourselves. These are interesting things to explore ;)

xUhEngwAng:
> Hi, thanks a lot for looking into this and apologize for my late reply.
> 
> Indeed the conda env occupies quite some space. However, we cannot move this setup to the build script. Because that will increase significantly the building time. Using the container setting in workflow as suggested [here](https://github.com/taichi-dev/taichi/issues/4824#issuecomment-1103620380) might worth a shot, since workflow installation is generally fast. Nevertheless, we still need to carefully exploit the tradeoff here between image size and build time ;)
> 
> Multi stage build effectively do the things like combining multiple RUN command. But perhaps we can also take advantage of what the other native image can offer, such as [this conda base image](https://hub.docker.com/r/conda/miniconda3/tags) is of size 142 MB, I wonder if this is smaller than what we build ourselves. These are interesting things to explore ;)

Thanks for reply. 

Actually I'm a bit confused about the container setting in workflows. I noticed that we need to create two containers, one for building and another for testing. Not sure whether we can specify two containers in a single job and copy files from one to another.

PR #4870, opened 2022-04-27
[bug] Remove redundant AllocStmt when lowering FrontendWhileStmt

jim19930609:
Related issue = #

Simple example:
```
@ti.kernel
def load_while(c: ti.i32):
    while c < 0:
        y[0] = c + 1.0 
````
After lowered to CHI IR, note that the highlighted variable "$0" has never been used: 

image


ailzhang (approved):
Nice catch!!

Issue #4620, opened 2022-03-25
Better error message when slicing a field


ailzhang:
![error](https://user-images.githubusercontent.com/5248122/160044311-2048006a-4ce8-4207-bd6d-59a83c91f34d.png)

We should provide a clear message that slicing is not supported.

PR #4873, opened 2022-04-27
[error] [lang] Improved error messages for illegal slicing or indexing to ti.field

jim19930609:
Related issue = #4620 

illegal behaviors to ti.field:
```
import taichi as ti
ti.init()
f_2d = ti.field(ti.u8, shape=(640, 480))

# illegal indexing
f_2d[0]

# illegal slicing
f_2d[0,4:5]
```

illegal indexing:
image

illegal slicing:
image


tests/python/test_numpy.py
```diff
@@ -187,6 +187,13 @@ def test_index_mismatch():
         val[0, 0] = 1
 
 
+@test_utils.test()
+def test_invalid_slicing():
```
ailzhang, line 191: `test_numpy.py` isn't the best place for this test :P maybe `test_field.py`?

python/taichi/lang/field.py
```diff
@@ -289,7 +295,16 @@ def __setitem__(self, key, value):
...
     def __getitem__(self, key):
         self._initialize_host_accessors()
-        return self.host_accessors[0].getter(*self._pad_key(key))
+        # Check for potential slicing behaviour
+        # for instance: x[0, :]
+        padded_key = self._pad_key(key)
+        for key in padded_key:
+            if not isinstance(key, int):
+                raise TypeError(
+                    f"Detected ilegal element of type: {type(key)}. "
```
ailzhang, line 304: typo: illegal ;)

python/taichi/lang/field.py
```diff
@@ -205,7 +205,13 @@ def _pad_key(self, key):
             key = ()
         if not isinstance(key, (tuple, list)):
             key = (key, )
-        assert len(key) == len(self.shape)
+
+        if len(key) != len(self.shape):
+            raise AssertionError(
+                "Slicing an array or matrix from ti.field is not supported so far."
```
ailzhang, line 211: Shall we simply say Slicing is not supported on ti.field? I'm a bit worried users confuse "matrix" here with the matrix element inside a ti.field ;)

tests/python/test_numpy.py
ailzhang: also maybe good to match the error message as well to make sure it fails at a place that we expected ;)

ailzhang (approved):
Thanks for fixing it!

tests/python/test_numpy.py
jim19930609: Good point, thx!

ailzhang (approved):
Thanks!

PR #4887, opened 2022-04-29
[Build] [refactor] Define Cmake OpenGL runtime target

qiao-bo:
Related issue = #4832 

- Separate opengl runtime codes from the backends folder
- eliminate some warnings

The internal dependencies (which is partly still polluted by global `include_directories()`) will be handled by a follow-up PR.


taichi/program/extension.cpp
```diff
@@ -1,5 +1,5 @@
 #include "extension.h"
-//#include "taichi/backends/opengl/opengl_api.h"
+//#include "taichi/runtime/opengl/opengl_api.h"
```
k-ye, line 2:
```suggestion
```

k-ye (approved):
lgtm!

PR #4889, opened 2022-04-29
[lang] Add reference type support on real functions

lin-hitonami:
Related issue = #602
This PR mainly does the following things: 
1. Add type annotation `ti.ref()`.
2. Add ReferenceExpression and ReferenceStmt to process reference type in the codegen.
3. Add `is_ptr` attribute to `ArgLoadExpression`. When `is_ptr` is `true`, the argument is a pointer to the referenced value, and it can be an lvalue.


k-ye (commented):
LGTM, will wait for @strongoier 's opinion.

strongoier (approved):
LGTM!

PR #4904, opened 2022-05-04
[Lang] Load cuSparse and cuSolver library for preparing sparse matrix on GPU

Hanke98:
Related issue = #2906 

Start to support sparse matrix on CUDA backend from this pr.

We first load the `cuSparse` and `cuSolver` libs via the `DynamicLoader` class in this pr. 

A new class, `CUDADriverBase` is introduced as the base class for the driver classes of these libs, which might also be helpful for the potential libs that we can use in the future.

### DIFF
--- a/taichi/backends/cuda/cuda_driver.cpp
+++ b/taichi/backends/cuda/cuda_driver.cpp
@@ -80,6 +80,7 @@ CUDADriver &CUDADriver::get_instance() {
 
 
 CUSPARSEDriver::CUSPARSEDriver() {
+  // TODO: enable cusparse and cusolver flag env variable.
   auto disabled_by_env_ = (get_environ_config("TI_ENABLE_CUDA", 1) == 0);
   if (disabled_by_env_) {
     TI_TRACE("CUDA driver disabled by enviroment variable \"TI_ENABLE_CUDA\".");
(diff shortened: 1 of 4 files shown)


Hanke98:
Sorry for the delayed update. Now, the `cuSolver` and `cuSpasrse` libraries can be opened via `CUSOLVERDriver` and `CUSPARSEDriver`. And the concrete APIs for accessing those libraries will be updated in the following pr.

BTW, should we add flags to indicate whether we should load those libraries, like `TI_ENABLE_CUSPARSE`?
@FantasyVR could you please have a look? Thanks!

taichi/backends/cuda/cuda_driver.cpp
```diff
@@ -78,4 +61,48 @@ CUDADriver &CUDADriver::get_instance() {
...
+  loader_ = std::make_unique<DynamicLoader>(lib_name);
+  if (!loader_->loaded()) {
+    TI_WARN("{} lib not found.", lib_name);
+    return;
+  } else {
+    TI_TRACE("{} loaded!", lib_name);
+  }
+}
+
+CUSPARSEDriver::CUSPARSEDriver() {
```
k-ye, line 90: nit: I'd suggest not to do this in the constructor, becuase if the lib fails to load, we get a broken/zombie object. A safer approach could be to load DLL in a function. If that fails, returns the exception/error to Python. Otherwise move that loaded lib into `CuSolverDriver`.

Also, we need to make sure not accidentally invoke `CUSPARSEDRiver::get_instance()` when we don't need the sparse matrix, otherwise Taichi could be broken if cuSparse is not installed.

I know this is not a newly introduced problem, so this is just something to think about :-)

k-ye:
> If the libraries are not installed we could generate error messages instead.

SG! TBH I was initially thinking something like torch's tags, e.g. `pip install torch==1.7.1+cu110`. But I think that would cause us to maintain more PyPI pkgs, so probably not worth it at this point..

Issue #2927, opened 2021-09-14
Cleanup build warnings when building with Vulkan


ailzhang:
`TAICHI_CMAKE_ARGS="-DTI_WITH_VULKAN:BOOL=ON"  python setup.py install` has a lot of build warnings, it'd be nice to get rid of most of them (otherwise these might just increase and make it hard to find errors).

PR #4916, opened 2022-05-06
[build] Warning Suppression PR #1: Ignored selected compilation warnings & Fixed clang-specific warnings

jim19930609:
Related issue = #3335, #2927

Changes in compiler options:
1. -Wno-ignored-attributes
[Global] By default, CXX compiler will throw a warning if it decides to ignore an attribute, for example "[[ maybe unused ]]".
However, this behaviour diverges across different compilers (GCC/CLANG), as well as different compiler versions.
Therefore we disable such warnings for now.

2. -Wno-nullability-completeness
[Global] Clang warns if a C++ pointer's nullability wasn't marked explicitly (__nonnull, nullable, ...).
Nullability seems to be a clang-specific feature, thus we disable this warning.

3. -Wno-unused-private-field
[Global] Disable warning for unused-private-field for convenience in development.

4. -Wno-unneeded-internal-declaration
[Global] By evaluating "constexpr", compiler throws a warning for functions known to be dead at compile time.
However, some of these "constexpr" are debug flags and will be manually enabled upon debuging.


taichi/backends/vulkan/vulkan_utils.cpp
```diff
@@ -5,31 +5,7 @@
...
-  vkEnumerateDeviceExtensionProperties(physicalDevice, kNoLayerName, &count,
-                                       nullptr);
-  std::vector<VkExtensionProperties> extensions(count);
-  vkEnumerateDeviceExtensionProperties(physicalDevice, kNoLayerName, &count,
-                                       extensions.data());
-  return extensions;
-}
-
-}  // namespace
+namespace {}  // namespace
```
k-ye, line 8: Feel free to remove this if it's no longer in use!

wyushun, line 8: it's an empty file?

jim19930609: yup, let me remove that

ailzhang (approved):
LGTM!

PR #4921, opened 2022-05-06
[SIMT] Add match_any warp intrinsics

galeselee:
Related issue = [4631](https://github.com/taichi-dev/taichi/issues/4631)#

    1. Add match_any warp intrinsic for cuda backend.
    2. Add match_any unit-test
    Notice: The datatype that supported in match_any is only `i32`.
      a. no `nvvm` intrinsic supports `f32` for match_any
      b. if `i64` is also required, please call me to add it.

### DIFF
--- a/python/taichi/lang/simt/warp.py
+++ b/python/taichi/lang/simt/warp.py
@@ -98,10 +98,10 @@ def shfl_xor_i32(mask, val, offset):
 
 
 def match_any(mask, value):
-    return expr.Expr(
-        _ti_core.insert_internal_func_call("cuda_match_any_sync_i32",
-                                           expr.make_expr_group(mask, value),
-                                           False))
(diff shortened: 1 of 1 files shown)


tests/python/test_simt.py
```diff
@@ -269,8 +269,25 @@ def foo():
...
+
+        for i in range(32):
+            b[i] = ti.simt.warp.match_any(ti.u32(0xFFFFFFFF), a[i])
+
+    foo()
+
+    for i in range(16):
+        assert b[i] == 65535
+    for i in range(16):
+        assert b[i + 16] == (2**32 - 2**16)
```
turbo0628, line 291: A little nit-picking: the test case cannot be easily interpreted. Can you add some comments at the API or in this test to briefly explain why the upper half of 32 is (2**32 - 2**16), document reference are also very helpful!

PR #4925, opened 2022-05-07
[Build] Improved building on Windows

PENGUINLIONG:
- Switched to non-unicode Win32 API because the input is not;
- Fixed that the core export `.dll`'s import `.lib` is not copied to `/build`;
- Fixed a typo in error message about the missing `CLANG_EXECUTABLE`.


turbo0628:
Did it cause any problems in GGUI by using the original API?

PENGUINLIONG:
> Did it cause any problems in GGUI by using the original API?

I haven't tested it for now but it can lead to an encoding problem, that localized text is interpreted as unicode. A minimal reproduction is:

```
#include <string>
void main() {
  std::string a = "土豆";
  std::wstring b(a.begin(), a.end());
  b += L"2";

  std::cout << a << std::endl;
  std::wcout << b << std::endl;

  std::cout << a << std::endl;
  std::wcout << b << std::endl;
}
```

which gives:

图片

And it might be a cause of crash.

ailzhang (commented):
Thanks! I'll leave the ggui window change for @AmesingFlank to review :D

qiao-bo (commented):
Thanks for the fix in CMake

PENGUINLIONG:
It seems the checks failed on something unrelated, is it possible to rerun the checks?

qiao-bo:
> It seems the checks failed on something unrelated, is it possible to rerun the checks?

`/rerun` should work ;/

ailzhang (approved):
LGTM!

PR #4927, opened 2022-05-07
[SIMT] Add uni_sync warp intrinsics

0xzhang:
Related issue = #4631

### DIFF
--- a/python/taichi/lang/simt/warp.py
+++ b/python/taichi/lang/simt/warp.py
@@ -14,9 +14,10 @@ def any_nonzero(mask, predicate):
             "cuda_any_sync_i32", expr.make_expr_group(mask, predicate), False))
 
 
-def unique():
-    # TODO
-    pass
+def unique(mask, predicate):
+    return expr.Expr(
+        _ti_core.insert_internal_func_call(
+            "cuda_uni_sync_i32", expr.make_expr_group(mask, predicate), False))
 
 
 def ballot(predicate):
--- a/taichi/llvm/llvm_context.cpp
+++ b/taichi/llvm/llvm_context.cpp
@@ -363,6 +363,9 @@ std::unique_ptr<llvm::Module> TaichiLLVMContext::clone_module(
     patch_intrinsic("cuda_any", Intrinsic::nvvm_vote_any);
     patch_intrinsic("cuda_any_sync", Intrinsic::nvvm_vote_any_sync);
 
+    patch_intrinsic("cuda_uni", Intrinsic::nvvm_vote_uni);
+    patch_intrinsic("cuda_uni_sync", Intrinsic::nvvm_vote_uni_sync);
+
     patch_intrinsic("cuda_ballot", Intrinsic::nvvm_vote_ballot);
     patch_intrinsic("cuda_ballot_sync", Intrinsic::nvvm_vote_ballot_sync);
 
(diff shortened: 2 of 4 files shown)


0xzhang:
I couldn't find any accurate information about `__uni_sync`. And I misunderstood the meaning of `unique` to Equal. I'll try to fix it.

https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#nvvm-intrin-warp-level-vote

0xzhang:
- GeForce GTX 1650
- CUDA 11.4

It seems `test_simt.py` not testable on my PC.
I use command below.
```shell
python tests/run_tests.py -t1 -k test_simt
```

image

0xzhang:
I only find one description about `__uni_sync()` on whole Internet.
If and only if the `predicate` of all threads which specified by `mask` are all non-zero or all zero, return 1, else return 0.

https://www.cnblogs.com/cuancuancuanhao/p/7841512.html

GaleSeLee:
> * GeForce GTX 1650
> * CUDA 11.4
> 
> It seems `test_simt.py` not testable on my PC. I use command below.
> 
> ```shell
> python tests/run_tests.py -t1 -k test_simt
> ```
> 
> image
> 
> Warp functions have requirements to compute capability of device. Maybe my card doesn't have enough compute capability according the [list](https://developer.nvidia.com/cuda-gpus). Or maybe the WSL environment limits the compute capability.

Do you solve this problem? I have a similar problem when I write `match_all` warp intrinsic

0xzhang:
@GaleSeLee No, I'm not sure the cause of problem. I can only use CI to verify that my test passed.

turbo0628 (approved):
LGTM!

turbo0628:
@GaleSeLee Can you try on that Linux RTX3080 machine?

GaleSeLee:
> No, I'm not sure the cause of problem. I can only use CI to verify that my test passed.

yeah, I'll come back later with the results

GaleSeLee:
> 

It occurs the same error on RTX3080 machine.

turbo0628:
> It occurs the same error on RTX3080 machine.

Could you update more information about the crash in tests? Mine works fine. I think we are using the identical environments.

GaleSeLee:
> > yeah, I'll come back later with the results
> 
> It occurs the same error on RTX3080 machine.

I'm wrong. And there is no error occured on RTX3080 machine.

turbo0628:
@0xzhang I think your guess is right, some legacy GPUs cannot support SIMT instructions. 

However, Taichi should speak out loudly when the GPU cannot run warp-level primitives. Could you file an issue to track this problem?

0xzhang:
> ```shell
> python tests/run_tests.py -t1 -k test_simt
> ```

OK, I create a [gist](https://gist.github.com/0xzhang/1929ad8d044671d05a3fac717c703843).

0xzhang:
> @0xzhang I think your guess is right, some legacy GPUs cannot support SIMT instructions.
> 
> However, Taichi should speak out loudly when the GPU cannot run warp-level primitives. Could you file an issue to track this problem?

Yes, I create a [gist](https://gist.github.com/0xzhang/1929ad8d044671d05a3fac717c703843).

0xzhang:
@turbo0628 I create a gist file [here](https://gist.github.com/0xzhang/1929ad8d044671d05a3fac717c703843).

turbo0628:
Thanks!

PR #4938, opened 2022-05-09
[SIMT] Add match_all warp intrinsics

GaleSeLee:
Related issue = https://github.com/taichi-dev/taichi/issues/4631
1. Add match_all warp intrinsic for Cuda backend.
2. Add match_all unit-test
3. There is still sth wrong, and I am still struggling with the bug.


GaleSeLee:
@qiao-bo

PR #4939, opened 2022-05-09
[Lang] Add more functions to math module

neozhaoliang:


python/taichi/_funcs.py
```diff
@@ -100,22 +100,39 @@ def _matrix_outer_product(self, other):
...
-    c = x * scale
-    s = y * scale
-    r = Matrix([[c, -s], [s, c]], dt=dt)
-    return r, r.transpose() @ A
+    U = Matrix.identity(dt, 2)
+    P = ops.cast(A, dt)
+    zero = ops.cast(0.0, dt)
+    # if A is the zero matrix we simply return the pair (I, A)
+    if (A[0, 0] == zero and A[0, 1] == zero and A[1, 0] == zero
+            and A[0, 0] == zero):
```
FantasyVR, line 118: duplicate `A[0, 0] == zero`

FantasyVR (commented):
Thanks!

python/taichi/_funcs.py
neozhaoliang: Oh, my mistake. It should be A[1, 1] == zero

python/taichi/_funcs.py
```diff
@@ -100,22 +100,39 @@ def _matrix_outer_product(self, other):
...
-    x, y = A(0, 0) + A(1, 1), A(1, 0) - A(0, 1)
-    scale = (1.0 / ops.sqrt(x * x + y * y))
-    c = x * scale
-    s = y * scale
-    r = Matrix([[c, -s], [s, c]], dt=dt)
-    return r, r.transpose() @ A
+    U = Matrix.identity(dt, 2)
+    P = ops.cast(A, dt)
+    zero = ops.cast(0.0, dt)
+    # if A is the zero matrix we simply return the pair (I, A)
```
FantasyVR, line 116:
```suggestion
    # if A is the zero matrix we simply return the pair (I, A)
```
```suggestion
    # if A is a zero matrix we simply return the pair (I, A)
```

FantasyVR (approved):
LGTM！ Just a few nits.

Issue #4949, opened 2022-05-10
Metal backend is kind of broken for external arrays args


k-ye:
The size of an array arg on Metal is still determined at *compile-time*, and is *fixed* after JIT compilation. See https://github.com/taichi-dev/taichi/blob/6d538e171c96193fd2ce27967bd0b17a5797565e/taichi/backends/metal/

With the introduction of Ndarray (`ti.ndarray`), array size is a runtime information. This means that arrays of different size won't re-trigger a kernel JIT. Hence, a Metal kernel with external arrays is now broken, if it's invoked twice with arrays of different size.

The only reason Metal backend still seems work when it's first invoked with an array is reasoned below:

1. [`Kernel::Arg::size`](https://github.com/taichi-dev/taichi/blob/6d538e171c96193fd2ce27967bd0b17a5797565e/taichi/program/callable.h#L22) is zero at arg declaration time: https://github.com/taichi-dev/taichi/blob/6d538e171c96193fd2ce27967bd0b17a5797565e/taichi/program/callable.cpp#L16-L26
2. When setting the external array arg at https://github.com/taichi-dev/taichi/blob/6d538e171c96193fd2ce27967bd0b17a5797565e/python/taichi/lang/kernel_impl.py#L667-L669, it eventually calls into here: https://github.com/taichi-dev/taichi/blob/6d538e171c96193fd2ce27967bd0b17a5797565e/taichi/program/kernel.cpp#L241. Note that `Kernel::Arg::size` is changed at runtime!
3. Then the kernel is invoked via `Kernel::operator()` https://github.com/taichi-dev/taichi/blob/6d538e171c96193fd2ce27967bd0b17a5797565e/taichi/program/kernel.cpp#L106-L108. At this point, Metal's codegen starts working. By the time it gets to https://github.com/taichi-dev/taichi/blob/6d538e171c96193fd2ce27967bd0b17a5797565e/taichi/backends/metal/kernel_utils.cpp#L91, because of step 2, the array size has been set correctly.

---

We should do two things:

1. Stop doing this: https://github.com/taichi-dev/taichi/blob/6d538e171c96193fd2ce27967bd0b17a5797565e/taichi/program/kernel.cpp#L241. The runtime size should be set in https://github.com/taichi-dev/taichi/blob/6d538e171c96193fd2ce27967bd0b17a5797565e/taichi/program/context.h#L16.
[...]

Issue #4800, opened 2022-04-15
AOT for the LLVM backend


k-ye:
* [x] Make CPU backend work
   * [x] Support arg-less kernels
   * [x] Support scalar + Ndarray args
   * [x] Split LlvmProgramImpl into compile-time & runtime parts to form Runtime API
* [x] Make CUDA backend work
   * [x] Support arg-less kernels
   * [x] Support scalar + Ndarray args
* [ ] Provide public interfaces
   * [x] Interface with Compute Graph
   * [ ] Interface with C-API
* [ ] Improvements & Demo
   * [ ] Dense/Pointer SNodes Demo
   * [ ] Support more flexible snode_id, snode_tree_id (determined at runtime)
   * [ ] Implement SNodeTreeType: https://github.com/taichi-dev/taichi/blob/master/docs/rfcs/20220413-aot-for-all-snode.md
   * [ ] Support memory deallocation for CPU backend. Merge `Device::allocate_memory_runtime()` with `Device::allocate_memory()`
   * [x] Verify SNode GC works for Field AOT.
   * [ ] Serialize/deserialize Ndarray/SNodeTree buffers: https://github.com/taichi-dev/taichi/issues/5428

PR #4950, opened 2022-05-10
[refactor] Add ArrayMetadata to store the array runtime size

k-ye:
Related issue = https://github.com/taichi-dev/taichi/issues/4949, https://github.com/taichi-dev/taichi/issues/4800

### DIFF
--- a/taichi/program/context.h
+++ b/taichi/program/context.h
@@ -29,7 +29,7 @@ struct RuntimeContext {
 
   // `array_runtime_size` records the runtime size of the
   // corresponding array arguments.
-  uint64 array_runtime_size[taichi_max_num_args_total]{0};
+  uint64 array_runtime_sizes[taichi_max_num_args_total]{0};
   // `is_device_allocation` is true iff args[i] is a DeviceAllocation*.
   bool is_device_allocation[taichi_max_num_args_total]{false};
   // We move the pointer of result buffer from LLVMRuntime to RuntimeContext
@@ -57,7 +57,7 @@ struct RuntimeContext {
   }
 
   void set_array_runtime_size(int i, uint64 size) {
-    this->array_runtime_size[i] = size;
+    this->array_runtime_sizes[i] = size;
   }
 
   void set_array_is_device_allocation(int i, bool is_device_allocation) {
--- a/taichi/program/context.h
+++ b/taichi/program/context.h
@@ -23,7 +23,11 @@ struct RuntimeContext {
   int32 extra_args[taichi_max_num_args_extra][taichi_max_num_indices];
   int32 cpu_thread_id;
   // |is_device_allocation| is true iff args[i] is a DeviceAllocation*.
-  bool is_device_allocation[taichi_max_num_args_total]{false};
+  struct ArrayMetadata {
+    uint64 runtime_size{0};
+    bool is_device_allocation{false};
+  };
+  ArrayMetadata array_metadata[taichi_max_num_args_total];
   // We move the pointer of result buffer from LLVMRuntime to RuntimeContext
   // because each real function need a place to store its result, but
   // LLVMRuntime is shared among functions. So we moved the pointer to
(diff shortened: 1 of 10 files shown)


taichi/runtime/opengl/opengl_api.cpp
```diff
@@ -400,23 +401,26 @@ void DeviceCompiledTaichiKernel::launch(RuntimeContext &ctx,
...
     TI_ASSERT(args[i].is_array);
-    if (args[i].size == 0 || ctx.is_device_allocation[i])
+    const auto &arr_meta = ctx.array_metadata[i];
+    const auto arr_runtime_sz = arr_meta.runtime_size;
+    if ((arr_runtime_sz == 0) || arr_meta.is_device_allocation) {
       continue;
+    }
     has_ext_arr = true;
-    if (args[i].size != item.second.total_size ||
+    if (arr_runtime_sz != item.second.runtime_size ||
```
k-ye, line 410: I think we should do this only when `arr_runtime_sz > item.second.runtime_size`? @ailzhang @bobcao3

ailzhang: Yup I think that'll work! Note this behavior differs by backend as Vulkan always deallocate the memory allocated for external arrays.  https://github.com/taichi-dev/taichi/blob/master/taichi/runtime/vulkan/runtime.cpp#L547-L554 We can probably align this behavior across them as well.

taichi/program/context.h
```diff
@@ -61,7 +69,7 @@ struct RuntimeContext {
                         DeviceAllocation &alloc,
                         const std::vector<int> &shape) {
     args[arg_id] = taichi_union_cast_with_different_sizes<uint64>(&alloc);
-    set_device_allocation(arg_id, true);
+    set_array_is_device_allocation(arg_id, /*is_device_allocation=*/true);
```
ailzhang, line 72: We also need to `set_array_runtime_size` here?

taichi/program/context.h
```diff
@@ -73,7 +81,7 @@ struct RuntimeContext {
                         const std::vector<int> &shape,
                         const std::vector<int> &element_shape) {
     args[arg_id] = taichi_union_cast_with_different_sizes<uint64>(&alloc);
-    set_device_allocation(arg_id, true);
+    set_array_is_device_allocation(arg_id, /*is_device_allocation=*/true);
```
ailzhang, line 84: ditto

ailzhang (approved):
Thanks for fixing it! This was one source of mixed compile-time and runtime info in the implementation....

taichi/program/context.h
k-ye: But why the old code didn't need this?

taichi/runtime/opengl/opengl_api.cpp
k-ye: +1

taichi/program/context.h
ailzhang: Because we only need the size information to do memcpy for **external arrays** in JIT runtime. This was one of the difference I mentioned between ndarray and external array in runtime that ndarray doesn't need implicit memcpy at this level. Its size information is available at higher `Ndarray` level which is good enough. 
IMHO a better way to do this is creating a class `ExternalArray` which carries its size information and handles the DeviceAllocation alloc/dealloc & memcpy instead of making RuntimeContext more complicated. (That way we can always assume only primitive_types or DeviceAllocation inside RuntimeContext, no need to worry about handling host_ptr inside each backend any more) 
These two functions are primarily used in the AOT runtime where we don't need to worry about external arrays so it was fine :P.

k-ye: > IMHO a better way to do this is creating a class ExternalArray which carries its size information and handles the DeviceAllocation alloc/dealloc & memcpy instead of making RuntimeContext more complicated.

That will be very useful indeed! But my point is that, since the old code doesn't need setting size here yet & this is mostly a refactoring, we probably don't have to `set_array_runtime_size` here?

ailzhang: sure, I was mainly afraid of the newly added size info and shape info in extra_args diverge in the `RuntimeContext` :P. But yea it's fine to ignore that for now :D

PR #4953, opened 2022-05-11
[ci] Add new buildbot with latest driver for Linux test

qiao-bo:
The new buildbot (buildbot10-linux) that has Driver 510 can test Vulkan API version 1.3


qiao-bo:
@frostming @ailzhang Any idea how to neatly solve this syntax issue? Basically I hope to write
```
    strategy:
      matrix:
        gpu_driver: [driver470, driver510]

    runs-on: [self-hosted, cuda, vulkan, cn, ${{ matrix.gpu_driver }}]
```
to run on two self hosted runners. But Github does not like the syntax and complains (https://github.com/taichi-dev/taichi/actions/runs/2306083544). I know the problem is at the line of `runs-on`. If the matrix var is removed, then it is fine ;/.

Any suggestions would be great!

.github/workflows/testing.yml
```diff
@@ -245,9 +245,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        tags: |
```
frostming: Oh, this `|` turns the following into a multiline string I guess.

ailzhang (approved):
LGTM!

PR #4955, opened 2022-05-11
[Build] Improve Windows build script

PENGUINLIONG:
- Support Python package userspace installtion;
- Support downloading prebuilt LLVM and Clang if absent;
- Powershell functions only, no depedency on environmental software, should work with any Windows 10+.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -4,10 +4,18 @@ param(
     [string] $LlvmDir = "",
     [string] $ClangDir = "",
     # Install python package in user-space.
-    [switch] $UserSpace = $false
+    [switch] $UserSpace = $false,
+    # Clean up compilation intermediates instead of building Taichi. Note that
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -1,3 +1,126 @@
...
+}
+
+# Install in userspace?
+$BuildExpr = "python setup.py develop";
+if ($UserSpace) {
+    Write-Host "Taichi Python package will be installed in user-space."
+    $BuildExpr += " --user"
+}
+
+Write-Host
```
qiao-bo, line 120: is this line intended?

qiao-bo (approved):
Thanks for cleaning-up this script!

build.ps1
PENGUINLIONG: Yes, to separate the log a bit from those emitted by the actual building script. 😮

PR #4957, opened 2022-05-11
[bug] [simt] Fix the problem that some intrinsics are never called

### LINKED ISSUE #4954: [cuda] `warp_barrier` intrinsic is never called when using `ti.simt.warp.sync()`
**Describe the bug**
As an expected result, when we use `ti.simt.warp.sync()`,  `warp_barrier` should be invoked at the backend. However, we can not find the `warp_barrier` intrinsic in the IR.

**To Reproduce**

Code:
```py
import taichi as ti

ti.init(arch=ti.cuda, print_ir=True)

@ti.kernel
def test_kernel(arr: ti.template()):
    for I in ti.grouped(arr):
        ti.simt.warp.sync(ti.u32(0xFFFFFFFF))

arr = ti.field(ti.i32, 32)
test_kernel(arr)
```

IR info:
```c++
[Taichi] version 1.0.2, llvm 10.0.0, commit 6d538e17, linux, python 3.8.10
[I 05/11/22 14:23:54.342 7682] [compile_to_offloads.cpp:operator()@22] [test_kernel_c56_0] Initial IR:
kernel {
  $0 : for @tmp0 in S2place<i32> noneblock_dim=adaptive {
    $1 = alloca @tmp1
    @tmp1 = (cast_value<i32> @tmp0)
  }
}
[I 05/11/22 14:23:54.342 7682] [compile_to_offloads.cpp:operator()@22] [test_kernel_c56_0] Lowered:
kernel {
  $0 : struct for in S1dense (bit_vectorize -1) noneblock_dim=adaptive {
    $1 = loop $0 index 0
    $2 = alloca
    $3 = cast_value<i32> $1
    $4 : local store [$2 <- $3]
  }
}
```

strongoier:
Related issue = #4631, fix #4954

Internal func calls are expressions, and they will not be inserted into Taichi AST if they are not assigned to others. The correct and simpler way to insert these intrinsics is to use the `impl.call_internal()` wrapper which includes forced assignment.

BTW in `subgroup.py`, those `dtype`s are removed because they only apply to constants, which are not feasible here.

### DIFF
--- a/python/taichi/lang/simt/subgroup.py
+++ b/python/taichi/lang/simt/subgroup.py
@@ -35,7 +35,9 @@ def broadcast_first(value):
 
 
 def broadcast(value, index):
-    return impl.call_internal("subgroupBroadcast", value, index,
+    return impl.call_internal("subgroupBroadcast",
+                              value,
+                              index,
                               with_runtime_context=False)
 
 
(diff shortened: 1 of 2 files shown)


qiao-bo (approved):
LGTM! Thanks

Issue #4900, opened 2022-05-01
meet some problems of Voxel Challenge on taichi 1.0.1(Received signal 22)


LiujiaHuan:
**Describe the bug**
Running example1.py by using the pycharm with taichi 1.0.1,python 3.10.2,arch=cuda,windows and  got many errors .
(I have tried to switch python version,it still didn't work)
After I switch the taichi to <kbd>1.0.0</kbd> **it did't report errors and program runs correctly.**
So,what is going on here?
I have seen <a href = "https://docs.taichi-lang.org/lang/articles/install">https://docs.taichi-lang.org/lang/articles/install</a>
But it didn't mentioned  "Received signal 22 (SIGABRT)"

**Log/Screenshots**
Please post the **full log** of the program (instead of just a few lines around the error message, unless the log is > 1000 lines). This will help us diagnose what's happening. For example:

```
C:\Users\13956\Desktop\My_taichi-main\venv\Scripts\python.exe C:/Users/13956/Desktop/My_taichi-main/example6.py
[Taichi] version 1.0.1, llvm 10.0.0, commit 1c3619d9, win, python 3.10.2
[Taichi] Starting on arch=cuda

====================================================
Camera:
* Drag with your left mouse button to rotate
* Press W/A/S/D/Q/E to move
====================================================

[I 05/01/22 13:01:55.935 2820] [D:/a/taichi/taichi/taichi/backends/vulkan/vulkan_device_creator.cpp:pick_physical_device@363] Found Vulkan Device 0 (NVIDIA GeForce RTX 3060 Laptop GPU)
[I 05/01/22 13:01:55.935 2820] [D:/a/taichi/taichi/taichi/backends/vulkan/vulkan_device_creator.cpp:pick_physical_device@363] Found Vulkan Device 1 (AMD Radeon(TM) Graphics)
[I 05/01/22 13:01:55.935 2820] [D:/a/taichi/taichi/taichi/backends/vulkan/vulkan_device_creator.cpp:create_logical_device@431] Vulkan Device "NVIDIA GeForce RTX 3060 Laptop GPU" supports Vulkan 0 version 1.3.194
[E 05/01/22 13:01:56.036 2820] Received signal 22 (SIGABRT)


***********************************
* Taichi Compiler Stack Traceback *
***********************************
[...]

Issue #4951, opened 2022-05-11
[vulkan] AMD RX 5600 XT reports Vulkan 1.3 but Core function pointer is absent


bobcao3:
reference #4900

turbo0628:
I'm closing this issue as it should be fixed in v1.0.2

PR #4958, opened 2022-05-11
[vulkan] Set kApiVersion to VK_API_VERSION_1_3

quadpixels:
This change fixes issue #4900. On certain AMD GPUs (such as RX 5500), the Vullkan version needs to be 1.3 in order for Volk to correctly populate the "vkGetDeviceBufferMemoryRequirements" function pointer. If the version is 1.2, the pointer will be NULL and cause an assertion failure.

Tested by running the voxel challenge program and see the program working as expected on a Windows PC.

According to the [manual page](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetDeviceBufferMemoryRequirements.html), this function appears to be from Vulkan 1.3, and as such I guess kApiVersion would need to be set to 1.3 as well. I'm not sure if this means legacy devices would be excluded?

Related issue = #4900 #4951

### DIFF
--- a/taichi/backends/vulkan/vulkan_utils.h
+++ b/taichi/backends/vulkan/vulkan_utils.h
@@ -22,7 +22,7 @@ namespace vulkan {
 class VulkanEnvSettings {
  public:
   static constexpr uint32_t kApiVersion() {
-    return VK_API_VERSION_1_2;
+    return VK_API_VERSION_1_3;
   }
 };
 


k-ye:
Thanks! But what's the implication of this for Vk1.2 users? Note that the majority of our CI bots are still running 1.2. Ideally we should not hardcode this one, and configure it via env var or `ti.init` or something.

quadpixels:
Yes this is a valid concern. I guess the right fix would be in the vulkan_device.cpp plus the version flag, and I'm not very familiar with that part yet.

I think this PR has served its purpose in asking the question about the version and can be closed and be superseded by a better solution.

k-ye:
It's good to know that this simple change fixes the problem on AMD GPUs, thanks!

turbo0628:
I also encountered similar problem on 6900XT, which crashes on vmaCreateAllocator.  This patch can also fix my problem. Thanks!

PR #4962, opened 2022-05-11
[llvm] Create ModuleToFunctionConverter

k-ye:
Related issue = https://github.com/taichi-dev/taichi/issues/4800

Thanks to @PGZXB 's https://github.com/taichi-dev/taichi/blob/master/taichi/llvm/llvm_offline_cache.h, we can now create a callable `void(RuntimeContext&)` directly from an `llvm::Module`, without going through codegen. This would be useful for LLVM AOT. In this PR, I've extracted this conversion part into a separate interface `ModuleToFunctionConverter`.

```cpp
  virtual FunctionType convert(const std::string &kernel_name,
                               const std::vector<ArgInfo> &args,
                               std::unique_ptr<llvm::Module> mod,
                               std::vector<OffloadedTask> &&tasks) const;
```

### DIFF
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -16,8 +16,6 @@
 #include "taichi/codegen/codegen_llvm.h"
 #include "taichi/llvm/llvm_program.h"
 
-#define TI_WITH_CUDA 1
-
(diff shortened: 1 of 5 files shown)


k-ye:
ping..

ailzhang (approved):
LGTM!

PR #4969, opened 2022-05-12
[build] Fixed Ilegal Instruction Error when importing PaddlePaddle module

jim19930609:
Related issue = #4973

**Note:**  this PR may not resolve the problem completely, please check the above issue for diagnose and next steps.

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -110,7 +110,7 @@ def test():
     parser.add_argument('-c',
                         '--cpp',
                         dest='cpp',
-                        default=True,
+                        default=False,
                         action='store_true',
                         help='Run the C++ tests')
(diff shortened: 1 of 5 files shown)


tests/run_tests.py
```diff
@@ -110,7 +110,7 @@ def test():
     parser.add_argument('-c',
                         '--cpp',
                         dest='cpp',
-                        default=True,
+                        default=False,
```
qiao-bo, line 113: why disable cpp test? ;)

jim19930609: It's interesting that with `action='store_true'` + `default=True`, it becomes impossible to turn it off :p

qiao-bo (approved):
LGTM! Thanks for the fix!

PR #4970, opened 2022-05-12
[vulkan] Set kApiVersion to VK_API_VERSION_1_3

turbo0628:
This PR reopens #4958 as the original branch was removed.

According to the [manual](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkApplicationInfo.html), the `apiVersion` to create a Vulkan instance should be the highest version designed to use.

Let's see if this change could pass all tests in CI.


ailzhang (approved):
Thanks!

PR #4972, opened 2022-05-12
[test] Add an ndarray test in C++.

ailzhang:
We only had a test for external array so adding one for ndarray.

Related issue = #


k-ye (approved):
LGTM!

PR #4977, opened 2022-05-13
[cuda] Add block and grid level intrinsic for cuda backend

YuCrazing:
Add `ti.simt.block.sync()` and `ti.simt.grid.memfence()` instrinsics, which equivalent to `__syncthreads()` and `__threadfence()` in cuda.

Related issue = #4631


strongoier (approved):
LGTM!

PR #4982, opened 2022-05-14
[llvm] Use serializer for LLVM cache

k-ye:
Related issue = #4800

@PGZXB FYI, I've switched the LLVM cache writer from the customized solution to Taichi's small serialization library. To see how it works, here are some examples: https://github.com/taichi-dev/taichi/blob/master/tests/cpp/common/serialization_test.cpp


ailzhang (approved):
LGTM!

Issue #4965, opened 2022-05-11
Feature request / request for comment:  Taichi "data structs" decorator


bsavery:
**Concisely describe the proposed feature**
Inspired by python `data classes`.  A similar decorator for data oriented classes for taichi.  This would allow defining a `data struct` class with annotated fields, and a class that is simply a structure of arrays would be made.  

For more background see my thoughts here https://github.com/taichi-dev/taichi/discussions/4784

**Describe the solution you'd like (if any)**
Consider making a struct for a sphere type.  It would be very similar to python data classes  to do:

```
@ti.data_struct
class Sphere:
    center: vec3
    radius: ti.f32
```

Then an array of spheres could just 
` sphere_array = Sphere(n) `

In fact this is _super_ easy to implement, (and I do use this) but I propose this might be more useful to many people.  

```python
import taichi as ti

ti.init(arch=ti.gpu)
vec3 = ti.types.vector(3, ti.f32)


######################## decorator code ##################

@ti.data_oriented
class BaseClass:

    def __init__(self, n):
        self.data = self.struct_type.field(shape=(n,))

    @ti.func
    def __getitem__(self, i):
        # return item as tuple so less dereferencing?
        return self.data[i]

    @ti.func
    def __setitem__(self, i, val_dict):
        # could do some type checking here?
        self.data[i] = self.struct_type(**val_dict)

def ti_datastruct(cls):
    struct_type = ti.types.struct(**cls.__annotations__)
    new_cls = type(cls.__class__.__name__, (BaseClass,), {'struct_type': struct_type})
    return new_cls


######################## end decorator ##################


# now all we need is the decorator 
@ti_datastruct
class Sphere:
    center: vec3
    radius: ti.f32

n = 10
my_spheres = Sphere(n) # maybe should be Sphere(shape=(n,)) for consistency
print(my_spheres.data)

```

**Additional comments**
I really like this pattern from a python point of view vs the current struct syntax.  To create a struct type and array is much cleaner, and allows adding additional methods to structs (#3553).  I posted a simple example, but would like feedback and if this would be something developers would accept as a language feature?

bsavery:
Additionally I should mention this is a simple implementation of this but could be made more robust with a more complete BaseClass.  To do things like:

`list_of_centers = my_spheres.center`

Should be possible

bsavery:
Thinking a bit more about this, let me sum up.  I think this way of using custom structs would be more "pythonic":

```python
@ti_datastruct
[...]

PR #4989, opened 2022-05-16
[Lang] Struct Classes implementation

bsavery:
Related issue = #4965

### TLDR.  Turns this:
```python
@ti.struct_class
class Sphere:
    center: vec3
    radius: ti.f32

    @ti.func
    def area(self):
        return 4 * 3.14 * self.radius * self.radius
```
Into a taichi struct (with methods!)

As discussed in the issue, this is a (draft) Pull request to implement `struct_classes`.  
`struct_class` is a decorator that can be attached to a python class definition that allows two things:

1.  Class field annotations are turned into struct fields (inspired by python Dataclasses)
2.  Class methods are attached to the struct and used like class methods.

IMO this allows a better syntax for structs and of course allows the struct methods.  Additionally it might ease much of what people would do with `ti.data_oriented` for complex arrays of structures in classes.  

**NOTE THIS IS A DRAFT PR**.  I am looking for feedback of the style and naming.  Also note I copied this from my working version so might need tweaking to work on your side.

### **Some technical notes**. 
There are basically two main things added here:
1.  Struct python type https://github.com/taichi-dev/taichi/blob/09de043b1e5fc522d642e7fb14a9dfed3283c8f3/python/taichi/lang/struct.py#L15 now can handle an optional `__struct_methods` object passed to it which contains the methods attached to the struct type.  I am open to other suggestions how to handle this.  But it should not have much overhead as the struct methods are only kept on the python side.  
2.  The new class decorator that turns a class with annotations fields into a struct and passes the methods.

### Example to test
```python
@ti.struct_class
class Sphere:
    center: vec3
    radius: ti.f32

    @ti.func
    def area(self):
        return 4 * 3.14 * self.radius * self.radius

n = 10
my_spheres = Sphere.field(shape=(n,))
print(my_spheres.radius)

# one struct
my_sphere = Sphere(center=vec3(2.0, 1.0, 9.0), radius=4.0)
# add to field
my_spheres[4] = my_sphere
print('done setup')

@ti.kernel
def kernel():
    sphere = my_spheres[4]
    center = sphere.center
    rad = sphere.radius
    area = sphere.area()
    
kernel()
```

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -210,6 +210,18 @@ def make_tensor_element_expr(_var, _indices, shape, stride):
                                           shape, stride))
 
 
+class SrcInfoGuard:
+    def __init__(self, info_stack, info):
+        self.info_stack = info_stack
+        self.info = info
+
+    def __enter__(self):
+        self.info_stack.append(self.info)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.info_stack.pop()
+
+
 class PyTaichi:
     def __init__(self, kernels=None):
         self.materialized = False
@@ -230,6 +242,12 @@ class PyTaichi:
 
     def get_num_compiled_functions(self):
         return len(self.compiled_functions) + len(self.compiled_grad_functions)
+
+    def src_info_guard(self, info):
+        return SrcInfoGuard(self.src_info_stack, info)
+
+    def get_current_src_info(self):
+        return self.src_info_stack[-1]
 
     def set_default_fp(self, fp):
         assert fp in [f16, f32, f64]
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -41,7 +41,7 @@ def expr_init(rhs):
     if isinstance(rhs, Matrix):
         return Matrix(rhs.to_list())
     if isinstance(rhs, Struct):
-        return Struct(rhs.to_dict())
+        return Struct(rhs.to_dict(include_methods=True))
     if isinstance(rhs, list):
         return [expr_init(e) for e in rhs]
     if isinstance(rhs, tuple):
@@ -168,9 +168,10 @@ def subscript(value, *_indices, skip_reordered=False):
         if isinstance(value, MatrixField):
             return _MatrixFieldElement(value, indices_expr_group)
         if isinstance(value, StructField):
+            entries = {k: subscript(v, *_indices) for k, v in value._items}
+            entries['__struct_methods'] = value.struct_methods
             return _IntermediateStruct(
-                {k: subscript(v, *_indices)
-                 for k, v in value._items})
+                entries)
         return Expr(_ti_core.subscript(_var, indices_expr_group))
     if isinstance(value, AnyArray):
         # TODO: deprecate using get_attribute to get dim
@@ -210,18 +211,6 @@ def make_tensor_element_expr(_var, _indices, shape, stride):
                                           shape, stride))
 
 
-class SrcInfoGuard:
-    def __init__(self, info_stack, info):
-        self.info_stack = info_stack
-        self.info = info
-
-    def __enter__(self):
-        self.info_stack.append(self.info)
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        self.info_stack.pop()
-
-
 class PyTaichi:
     def __init__(self, kernels=None):
         self.materialized = False
@@ -242,12 +231,6 @@ class PyTaichi:
 
     def get_num_compiled_functions(self):
         return len(self.compiled_functions) + len(self.compiled_grad_functions)
-
-    def src_info_guard(self, info):
-        return SrcInfoGuard(self.src_info_stack, info)
-
-    def get_current_src_info(self):
-        return self.src_info_stack[-1]
 
     def set_default_fp(self, fp):
         assert fp in [f16, f32, f64]
--- a/python/taichi/lang/struct.py
+++ b/python/taichi/lang/struct.py
@@ -10,7 +10,8 @@ from taichi.lang.util import (cook_dtype, in_python_scope, is_taichi_class,
                               python_scope, taichi_scope)
 from taichi.types import primitive_types
 from taichi.types.compound_types import CompoundType
-from functools import partial
+from taichi.lang.kernel_impl import func
+from types import MethodType
 
 
 class Struct(TaichiOperations):
@@ -56,8 +57,11 @@ class Struct(TaichiOperations):
             )
         self.methods = self.entries.pop("__struct_methods", {})
         for name, method in self.methods.items():
-            # use partial to pass self (this object) to the method
-            setattr(self, name, partial(method, self))
+            # use MethodType to pass self (this object) to the method
+            setattr(self, name, MethodType(method, self))
+            if getattr(method, '_is_taichi_function', False) is True:
+                # mark as a taichi function
+                func(getattr(self, name))
 
         for k, v in self.entries.items():
             if isinstance(v, (list, tuple)):
@@ -362,7 +366,11 @@ class _IntermediateStruct(Struct):
         assert isinstance(entries, dict)
         self.methods = entries.pop('__struct_methods', {})
         for name, method in self.methods.items():
-            setattr(self, name, partial(method, self))
+            # use MethodType to pass self (this object) to the method
+            setattr(self, name, MethodType(method, self))
+            if getattr(method, '_is_taichi_function', False) is True:
+                # mark as a taichi function
+                func(getattr(self, name))
         self.entries = entries
         self._register_members()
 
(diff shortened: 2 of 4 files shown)


python/taichi/lang/struct.py
```diff
@@ -49,6 +56,14 @@ def __init__(self, *args, **kwargs):
             raise TaichiSyntaxError(
                 "Custom structs need to be initialized using either dictionary or keyword arguments"
             )
+        self.methods = self.entries.pop("__struct_methods", {})
+        for name, method in self.methods.items():
+            # use MethodType to pass self (this object) to the method
+            setattr(self, name, MethodType(method, self))
+            if getattr(method, '_is_taichi_function', False) is True:
```
strongoier, line 63: I'm a little curious about why we need this. Doesn't `@ti.func` work?

BTW are we supposed to get these struct methods run in both Python scope and Taichi scope?

python/taichi/lang/struct.py
```diff
@@ -632,7 +664,41 @@ def filled_with_scalar(self, value):
         return Struct(entries)
 
     def field(self, **kwargs):
-        return Struct.field(self.members, **kwargs)
+        return Struct.field(self.members, self.methods, **kwargs)
+
+
+def struct_class(cls):
```
strongoier, line 670: I'm not so sure whether the name `ti.struct_class` is easy enough to understand. Is something like `ti.struct_with_methods` a more straightforward choice? @k-ye

strongoier (commented):
Thanks for the great work!

python/taichi/lang/impl.py
```diff
@@ -210,18 +210,6 @@ def make_tensor_element_expr(_var, _indices, shape, stride):
                                           shape, stride))
 
 
-class SrcInfoGuard:
-    def __init__(self, info_stack, info):
```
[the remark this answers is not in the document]
bsavery: Yes, I'll get that fixed.

python/taichi/lang/struct.py
bsavery: So the fact that we're re-binding the function to a new object with the func() call seems to lose the ti.func bits and gives a warning.  

Python and taichi scope functions should work fine with this change.

python/taichi/lang/struct.py
k-ye: Hmm, I actually feel like `struct_class` OK enough. Do you think it's not straightforward from the C++ perspective..?

k-ye:
Thanks for the PR! I'd like to understand a few things here:

1. Will `ti.struct_class` always AOS?
2. Can we use `ti.field(dtype=StructType, ...)` for this new struct? (We plan to gradually migrate all `ti.Vector.field`, `ti.Matrix.ndarray` to this unified, `dtype` style.)

bsavery:
> Thanks for the PR! I'd like to understand a few things here:
> 
> 1. Will `ti.struct_class` always be AOS?
> 2. Can we use `ti.field(dtype=StructType, ...)` for this new struct? (We plan to gradually migrate all `ti.Vector.field`, `ti.Matrix.ndarray` to this unified, `dtype` style.)

No behavior of creating a struct field is changed.  So you can do anything you can with a normal field.

bsavery:
@strongoier it looks like the CI tests are failing due to an api change?  Assuming because I'm exposing a new name.  Can you point me where to fix that?

strongoier:
> @strongoier it looks like the CI tests are failing due to an api change? Assuming because I'm exposing a new name. Can you point me where to fix that?

Yep. You can fix that in `tests/python/test_api.py`, which restricts all the public APIs.

python/taichi/lang/struct.py
strongoier: To some extent yes :-) Putting `struct` and `class` together confuses me a bit. But it may be OK if we understand it as "turning a Python class into a Taichi struct".

bsavery:
Another point that might be worth pointing out, one could actually do something like this after the change:

```python
def area(self):
    return 4 * 3.14 * self.radius * self.radius
sphere_struct = ti.types.struct(radius=ti.f32, center=vec3, __struct_methods={'area':area})
```

strongoier (commented):
The technical route looks good to me! I think we can now work towards getting the PR in. Some comments besides in-line reviews:
1. I tried your branch locally and tweaked a bit. It seems that `@ti.func` works fine if you remove the manual marking part (`if getattr(method, '_is_taichi_function', False) is True`). However, there's another problem - `ti.func` can only be called in Taichi scope, but not in Python scope. My proposal here is to let users mark nothing and let `struct_class` mark the function as `pyfunc`: https://github.com/taichi-dev/taichi/blob/ac1f06f36f292d800a0f0b6b2f3b38ea66867fd4/python/taichi/lang/kernel_impl.py#L76
2. Could you add some tests in `tests/python/test_custom_struct.py`? Other tests in that file can serve as examples. We need to make sure the new feature actually works as expected. 
3. The last piece is about docstrings and documentation - we can put this off until all other issues are resolved.

bsavery:
> The technical route looks good to me! I think we can now work towards getting the PR in. Some comments besides in-line reviews:
> 
> 1. I tried your branch locally and tweaked a bit. It seems that `@ti.func` works fine if you remove the manual marking part (`if getattr(method, '_is_taichi_function', False) is True`). However, there's another problem - `ti.func` can only be called in Taichi scope, but not in Python scope. My proposal here is to let users mark nothing and let `struct_class` mark the function as `pyfunc`: https://github.com/taichi-dev/taichi/blob/ac1f06f36f292d800a0f0b6b2f3b38ea66867fd4/python/taichi/lang/kernel_impl.py#L76

So I updated the change.  You're right that functions don't need to be re-marked as ti funcs.  I think the user should be able to mark struct class functions as taichi scope or python scope and he can now do that. 
> 2. Could you add some tests in `tests/python/test_custom_struct.py`? Other tests in that file can serve as examples. We need to make sure the new feature actually works as expected.

Added 
> 3. The last piece is about docstrings and documentation - we can put this off until all other issues are resolved.


Let me know what you want here.

strongoier (commented):
> I think the user should be able to mark struct class functions as taichi scope or python scope and he can now do that.

Sounds good!

> Let me know what you want here.

Suggestions for this PR are commented.

BTW some potential future work comes to my mind, which is not required for this PR: it will be pretty helpful if you can also write a short article/tutorial on our main docsite about using the new style to write Taichi programs, so that this feature can get advertised :-)

bsavery:
Will happily write something under the docsite.  Probably a section un "Advanced Topics"?

python/taichi/lang/struct.py
```diff
@@ -233,24 +244,29 @@ def __str__(self):
...
         return str(self.to_dict())
 
-    def to_dict(self):
+    def to_dict(self, include_methods=False):
         """Converts the Struct to a dictionary.
 
         Args:
-
+            include_methods (bool): Whether any struct methods should be included
+                in the result dictionary under the key '__struct_methods'.
```
strongoier, line 252:
```suggestion
                in the result dictionary under the key '__struct_methods'.

```

python/taichi/lang/struct.py
```diff
@@ -336,17 +355,21 @@ def field(cls,
...
-        return StructField(field_dict, name=name)
+        return StructField(field_dict, methods, name=name)
 
 
 class _IntermediateStruct(Struct):
     """Intermediate struct class for compiler internal use only.
 
     Args:
         entries (Dict[str, Union[Expr, Matrix, Struct]]): keys and values for struct members.
+            Any methods included under the key '__struct_methods' will be applied to each
```
strongoier, line 366: This should not be a separate paragraph.

python/taichi/lang/struct.py
```diff
@@ -629,10 +660,53 @@ def filled_with_scalar(self, value):
...
         return Struct(entries)
 
     def field(self, **kwargs):
-        return Struct.field(self.members, **kwargs)
+        return Struct.field(self.members, self.methods, **kwargs)
 
 
-__all__ = ["Struct", "StructField"]
+def struct_class(cls):
+    """ Converts a class with field annotations into a struct type for use as a taichi field
```
strongoier, line 671:
```suggestion
    """Converts a class with field annotations and methods into a taichi struct type.
```

python/taichi/lang/struct.py
```diff
@@ -629,10 +660,53 @@ def filled_with_scalar(self, value):
...
+
+        >>> @ti.stuct_class
+        >>> class Sphere:
+        >>>     center: vec3
+        >>>     radius: ti.f32
+        >>>
+        >>>     @ti.func
+        >>>     def area(self):
+        >>>         return 4 * 3.14 * self.radius * self.radius
+        >>> my_spheres = Sphere.field(shape=(n,))
```
strongoier, line 691:
```suggestion
        >>>
        >>> my_spheres = Sphere.field(shape=(n, ))
```

strongoier (approved):
LGTM!!! Only some nits left. Thanks again for the efforts!

> Will happily write something under the docsite. Probably a section un "Advanced Topics"?

Sounds great! FYI the docs are under `taichi/docs/lang/articles/advanced/`.

bsavery:
Awesome!  I’ll write some documentation this week asap

PR #4996, opened 2022-05-17
[refactor] Program owns allocated ndarrays.

ailzhang:
* #5004
* #5002
* #4999
* #4998
* #4997
* __->__ #4996

The end goal of this refactor is let Ndarray be a simple wrapper around
(DeviceAllocation, dtype, shape) without having to worry about memory
allocation/deallocation. But its current implementation heavily couples
with Program*, so an intermediate state would be:
- If created from Program, Ndarray handles deviceallocation in
ctor/dtor.
- We'll add another ctor simply constructing Ndarray from
(DeviceAllocation, dtype, shape) and update the codebase to it.


k-ye (approved):
LGTM!

PR #4998, opened 2022-05-17
[refactor] Construct ndarray from existing DeviceAllocation.

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #5004
* #5002
* #4999
* __->__ #4998
* #4997
* #4996

The end goal is make this the only ctor for Ndarray class.

PR #4997, opened 2022-05-17
[test] Add test for Ndarray from DeviceAllocation

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #5004
* #5002
* #4999
* #4998
* __->__ #4997
* #4996

PR #4999, opened 2022-05-17
[refactor] Free ndarray's memory when python GC triggers

ailzhang:
* #5004
* #5002
* __->__ #4999
* #4998
* #4997
* #4996

Previously Program manages lifetime of all allocated ndarrays. So when
you call `del ndarray` in python, its memory was not freed. This PR
changes the behavior that `ndarray` memory gets deallocated when python
GC triggers, or its containing `Program` gets destructed, whichever
happens first. There're some quirks around how we handle the async
python GC and manual `ti.reset()`. Thanks to @k-ye, we now added a
`generation` number to track the containing program instance of ndarrays
so that memory deallocation happens correctly.

### DIFF
--- a/tests/python/test_ndarray.py
+++ b/tests/python/test_ndarray.py
@@ -335,6 +335,15 @@ def test_ndarray_numpy_io():
     _test_ndarray_numpy_io()
 
 
+@test_utils.test(arch=supported_archs_taichi_ndarray)
+def test_ndarray_reset():
+    n = 8
+    c = ti.Matrix.ndarray(4, 4, ti.f32, shape=(n))
+    del c
+    d = ti.Matrix.ndarray(4, 4, ti.f32, shape=(n))
+    ti.reset()
+
+
 def _test_ndarray_matrix_numpy_io(layout):
     n = 5
     m = 2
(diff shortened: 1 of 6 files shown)


tests/python/test_ndarray.py
```diff
@@ -607,3 +616,15 @@ def init(d: ti.i32, arr: ti.types.ndarray()):
...
+@test_utils.test(arch=supported_archs_taichi_ndarray)
+def test_generation():
+    curr_arch = ti.lang.impl.current_cfg().arch
+    n1 = 4
+    x = ti.ndarray(dtype=ti.f32, shape=(n1, n1))
+    curr_gen = x._gen
+    ti.reset()  # gen++
+    ti.init(curr_arch)  # calls ti.reset(), gen++
+    y = ti.ndarray(dtype=ti.f32, shape=(n1, ))
+    assert y._gen == curr_gen + 2
```
k-ye, line 630: Super nit, we probably don't care the specific difference between these two gens, so it would be fine to just `assert y._gen != curr_gen`  (O.w. if we change the generation implementation, this test would unexpectedly fail).

Also, would be nice to call a `del x` in the end :-)

k-ye (approved):

PR #5002, opened 2022-05-18
[refactor] Move ndarray fast fill methods to Program

ailzhang:
* #5004
* __->__ #5002

This PR gets rid of `LlvmProgramImpl*` member inside `Ndarray` class,
which is a step closer towards decoupling `Ndarray` and memory
management.


taichi/program/program.cpp
```diff
@@ -569,6 +569,17 @@ void Program::delete_ndarray(Ndarray *ndarray) {
   ndarrays_.erase(ndarray);
 }
 
+void Program::fill_ndarray_fast(Ndarray *ndarray, uint32_t val) {
```
k-ye, line 572: C++ nerd: I'm thinking about tag dispatching here LOL. One way to do it would be have a separate Ndarray class for each backend. (This should work, because by the time we allocate an Ndarray, we know which backend it runs on). But this is nothing urgent at all.

k-ye (approved):
LGTM!

taichi/program/program.cpp
ailzhang: agreed!

PR #5007, opened 2022-05-18
[AOT] Supported inclusion of taichi as subdirectory for AOT modules

PENGUINLIONG:
- Allowed Python not to be built in case of non-Android platforms;
- Skiped byte code generation when `TI_WITH_LLVM=OFF`;
- Fixed include directory path relativity when included as a subdirectory;
- Temporarily disabled warning-as-error on Android.


PENGUINLIONG:
Mirror PR: https://github.com/taichi-dev/taichi-aot-demo/pull/18

ailzhang (approved):
LGTM!

Issue #4986, opened 2022-05-16
Compiled kernel doesn't handle signess right.


feisuzhu:
**Describe the bug**
Subtracting an unsigned number behaves like subtracting signed number.

**To Reproduce**
Please post a **minimal sample code** to reproduce the bug.
The developer team will put a higher priority on bugs that can be reproduced within 20 lines of code. If you want a prompt reply, please keep the sample code **short** and **representative**.

```py
import taichi as ti
ti.init(arch=ti.cpu)

A = ti.field(ti.uint8, shape=())
B = ti.field(ti.uint16, shape=())

def step_raw():
    v: ti.uint16 = 1000
    v -= A[None]
    B[None] = v

step = ti.kernel(step_raw)


def main():
    A[None] = 255
    step()
    print(A[None], B[None])
    step_raw()
    print(A[None], B[None])

if __name__ == "__main__":
    main()
```

**Log/Screenshots**
Please post the **full log** of the program (instead of just a few lines around the error message, unless the log is > 1000 lines). This will help us diagnose what's happening. For example:

PyPI:
```
☁  taichi  python3 r2.py
[Taichi] version 1.0.1, llvm 10.0.0, commit 1c3619d9, linux, python 3.8.10
[Taichi] Starting on arch=x64
255 1001
255 745
```
nightly:
```
☁  taichi  python3 r2.py
[Taichi] version 1.0.2, llvm 10.0.0, commit 22d18951, linux, python 3.8.10
[Taichi] Starting on arch=x64
255 1001
255 745
```

**Additional comments**
If possible, please also consider attaching the output of command `ti diagnose`. This produces the detailed environment information and hopefully helps us diagnose faster.

```
☁  taichi  ti diagnose
[Taichi] version 1.0.2, llvm 10.0.0, commit 22d18951, linux, python 3.8.10

*******************************************
**      Taichi Programming Language      **
*******************************************

Docs:   https://docs.taichi-lang.org/
GitHub: https://github.com/taichi-dev/taichi/
Forum:  https://forum.taichi.graphics/

Taichi system diagnose:

python: 3.8.10 (default, Mar 15 2022, 12:22:08)
[GCC 9.4.0]
system: linux
executable: /usr/bin/python3
platform: Linux-5.4.0-110-generic-x86_64-with-glibc2.29
architecture: 64bit ELF
[...]

PR #5011, opened 2022-05-19
[bug] Fixed numerical error for Atomic-Sub between unsigned values with different number of bits

jim19930609:
Related issue = #4986

1. Issue
```
x: ti.uint16 = 1000
y: ti.uint8 = 1
ti.atomic_sub(x, y)
```
The above code gives `x = 1255` instead of 999. 

2. Diagnose
Taichi transforms `atomic_sub(x, y)` into `neg_y = negate(y); atomic_add(x, neg_y)` during `irpass::lower_ast()`, therefore the following type promotion will bit_cast "neg_y" (1's complement of y) into another unsigned integer, which breaks the interpretation of 1's complement.

3. Solution
We should perform the cast before negating rhs's value.

image

4. Other comments
Frontend IR does not seem to be a good place for transformations. Shall we write a CHI IR pass for `atomic_sub -> atomic_add` to replace the one inside irpass::lower_ast()?

### DIFF
--- a/taichi/ir/frontend_ir.cpp
+++ b/taichi/ir/frontend_ir.cpp
@@ -481,6 +481,11 @@ void AtomicOpExpression::type_check(CompileConfig *) {
 void AtomicOpExpression::flatten(FlattenContext *ctx) {
   // replace atomic sub with negative atomic add
   if (op_type == AtomicOpType::sub) {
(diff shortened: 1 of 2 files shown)


k-ye (approved):
Nice!

ailzhang (approved):
Thanks!

PR #5013, opened 2022-05-19
[Llvm] Add AOT builder and loader

k-ye:
Related issue = #4800

### DIFF
--- a/taichi/backends/cpu/aot_module_loader_impl.cpp
+++ b/taichi/backends/cpu/aot_module_loader_impl.cpp
@@ -1,0 +1,96 @@
+#include "taichi/backends/cpu/aot_module_loader_impl.h"
+
+#include "taichi/llvm/llvm_offline_cache.h"
+#include "taichi/llvm/llvm_program.h"
+#include "taichi/codegen/codegen_llvm.h"
+
+namespace taichi {
+namespace lang {
+namespace cpu {
+namespace {
+
+class KernelImpl : public aot::Kernel {
+ public:
+  explicit KernelImpl(FunctionType fn) : fn_(fn) {
+  }
+
+  void launch(RuntimeContext *ctx) override {
+    fn_(*ctx);
+  }
+
+ private:
+  FunctionType fn_;
+};
+
+class AotModuleImpl : public aot::Module {
+ public:
+  explicit AotModuleImpl(const AotModuleParams &params)
+      : program_(params.program),
+        cache_reader_(LlvmOfflineCacheFileReader::make(params.module_path)) {
+  }
+
+  Arch arch() const override {
+    return Arch::x64;
+  }
+
+  uint64_t version() const override {
+    return 0;
+  }
+
+  size_t get_root_size() const override {
+    return 0;
+  }
+
+ private:
+  std::unique_ptr<aot::Kernel> make_new_kernel(
+      const std::string &name) override {
+    TI_ASSERT(cache_reader_ != nullptr);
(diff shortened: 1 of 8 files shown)


taichi/backends/cpu/aot_module_loader_impl.cpp
```diff
@@ -0,0 +1,96 @@
...
+    return nullptr;
+  }
+
+  LlvmProgramImpl *const program_{nullptr};
+  std::unique_ptr<LlvmOfflineCacheFileReader> cache_reader_{nullptr};
+};
+
+}  // namespace
+
+std::unique_ptr<aot::Module> make_aot_module(std::any mod_params) {
```
jim19930609, line 88: Was wondering what's the benefit of using "std::any + std::any_cast<>" here? 
In this way, if the customer passed an ill-typed `mod_param`, the error message will get postponed from compile time to runtime.

taichi/backends/cpu/aot_module_loader_impl.cpp
```diff
@@ -0,0 +1,96 @@
...
+  auto mod = std::make_unique<AotModuleImpl>(
+      std::any_cast<const AotModuleParams &>(mod_params));
```
jim19930609, line 90: This is more like nitpicking, feel free to ignore this comment:

Looks like AotModuleParams has a public pointer-typed member `LlvmProgramImpl *program`, the underlying data of which will not be protected by the "const reference".

For example, you can still modify the underlying data of `int* p` with `const A&`
```
struct A { 
    A() { p = new int(10); }

    int* p = nullptr;
};

void foo(const A& a) {
    *a.p = 100;
}

int main() {
    
    A a;
    foo(a);
    std::cout << *a.p << std::endl;

    return 0;
}
```

To make your "const reference" bullet-proof, better mark the pointer-typed members as private and provide mutable/read-only interfaces.
```
struct A { 
    A() { p = new int(10); }

    void set_p(int val) { *p=val; }
    int get_p() const { return *p; }
    
 private:
    int* p = nullptr;
};
```

taichi/backends/cpu/aot_module_loader_impl.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  size_t get_root_size() const override {
+    return 0;
+  }
+
+ private:
+  std::unique_ptr<aot::Kernel> make_new_kernel(
+      const std::string &name) override {
+    TI_ASSERT(cache_reader_ != nullptr);
+    auto *tlctx = program_->get_llvm_context(program_->config->arch);
```
jim19930609, line 48: Also check `program_ != nullptr` ?

jim19930609 (commented):
It would be even better if we can have a simple test case along with this PR :)

taichi/backends/cpu/aot_module_loader_impl.cpp
k-ye: Thanks! I fully agree with this point in general, and +100 for encapsulation!

However, note that Params is mostly a simple POD type passing the necessary info to `make_aot_module`, with no behaviors in it. Therefore I usually choose not to use getter/setter for POD data. WDYT?

taichi/backends/cpu/aot_module_loader_impl.cpp
k-ye: Done

k-ye:
> It would be even better if we can have a simple test case along with this PR :)

That's in the coming PR :-)

Issue #4786, opened 2022-04-14
[AOT] Serialize compute graph in python and load in C++ runtime


ailzhang:
Copied from #4615  :P : This would significantly reduce the efforts to porting a demo written in Taichi to non-Python environments. In essence, the AOT module will save not only the Taichi kernels, but also the host logic invoking them.
[Will add more context as we start adding this feature]

ailzhang:
Update 04/27: 
We've converged on the a API style for the simplest case. 
```
@ti.aot.graph
def run_sim():
    kernel1()
    kernel2()

mod = ti.aot.Module()
mod.add_graph(run_sim)

## in C++
# Note that serialization format is implementation detail that we haven't covered
# But it's pretty flexible.
graph = aot_module->load_graph("...");
graph->run(host_ctx);
```
Note it's easy to support static control flow, but we're spending some time exploring how we're supposed to support dynamic control flow. Is it going to be in graph or out-of-graph? (concepts borrowed from "Dynamic Control Flow in Large-Scale Machine Learning" paper) 
- in-graph：encodes control-flow decisions as operations in the dataflow graph
- out-of-graph ：implements control-flow decisions in the client process, using control-flow primitives of a host language like Python. 

1. The out-of-graph approach is more like the traditional tracing approach. 
- Loops are unrolled and branch is selected based on python runtime value. 
- You cannot prevent users from using device to host sync for control flows, which might hurt performance a lot. 

2. For in-graph approach we need to decide whether the control flow support is explicit or implicit: 
- explicit: more like JAX style 
- implicit: more like TorchScript style

We're currently building a toy prototype just to get some sense of usability and coverage using the implicit in-graph approach.

ailzhang:
Update 05/16 (working in progress):

We'll be adding support for construct static compute graph and its execution. 

## Goals:
- The graph can be serialized, then deployed and run in an environment without python runtime. 
- It's be super helpful if we can run the graph in python frontend as if it was deployed, to make the AOT debugging experience better.  
- Normal taichi python frontend users can opt in to use graph execution mode as well. We've noticed pretty heavy overhead from python->C++ communication especially for small kernels on a powerful GPU. Launching a graph instead of individual small kernels from python can dramatically reduce the overhead for these light kernels. 

## Non-goals:
- Graph mode won't be as flexible as current python JIT (dynamic graph) as it won't allow host execution (like returning a value from a taichi kernel to python and condition on it).


## Key terminologies:
[...]

PR #5016, opened 2022-05-20
[aot] Serialize built graph, deserialize and run.

ailzhang:
* #5023
* #5017
* __->__ #5016

related: #4786

[Update]: based on an offline discussion with @k-ye, I've split the
original `Graph` class into `GraphBuilder` and `CompiledGraph` classes
in C++. Note that the implementation didn't follow exactly the builder
design pattern as our builder is slightly simpler as shown below.
The complexity in our problem is more in the need of serialization and
deserialization for the same graph representation intead of its
construction process. So IMHO it's good enough to separate the
GraphBuilder and Runner(`CompiledGraph`) as we discussed. Please feel
free to correct me if I'm wrong!

```
GraphBuilder
    |
 compile()
    |
    |
CompiledGraph <----  serialize/deserialize ----> file
    |
    |
   run()
```

This PR demonstrates a minimal example of serializing a built graph,
deserializing and running it.


tests/cpp/aot/aot_save_load_test.cpp
```diff
@@ -299,4 +300,104 @@ TEST(AotSaveLoad, VulkanNdarray) {
...
+  // where Vulkan is installed but no devices are present
+  if (!vulkan::is_vulkan_api_available()) {
+    return;
+  }
+
+  save_graph();
+
+  // API based on proposal https://github.com/taichi-dev/taichi/issues/3642
+  // Initialize Vulkan program
+  taichi::uint64 *result_buffer{nullptr};
```
k-ye, line 336: Oh, I feel like it would be great if these tests can made the same for all the archs supporting AOT! (cc @turbo0628 @jim19930609 ) A more realistic result probably is that we have different setup routines for each arch. But the graph part stays the same.

taichi/backends/vulkan/aot_module_loader_impl.h
```diff
@@ -27,6 +28,16 @@ class KernelImpl : public aot::Kernel {
     runtime_->launch_kernel(handle, ctx);
   }
 
+  void save_to_module(AotModuleBuilder *builder) override {
+    // This hack exists because ti_aot_data_ is vulkan specific.
+    // We need a generic aot::ModuleData inside AotModuleBuilder.
+    dynamic_cast<AotModuleBuilderImpl *>(builder)->aot_data().kernels.push_back(
```
k-ye, line 34: I think `static_cast` should work here, because we know its type.

taichi/backends/vulkan/aot_module_builder_impl.h
```diff
@@ -21,6 +21,11 @@ class AotModuleBuilderImpl : public AotModuleBuilder {
   void dump(const std::string &output_dir,
             const std::string &filename) const override;
 
+  // FIXME: remove me once TaichiAotData is no longer backend specific.
```
k-ye, line 24: I think it's fine to add `ti_aot_data_` to the module builder base class. It's just that not every impl uses it, yet?

taichi/backends/vulkan/aot_module_builder_impl.cpp
```diff
@@ -147,6 +149,10 @@ void AotModuleBuilderImpl::add_per_backend(const std::string &identifier,
   ti_aot_data_.spirv_codes.push_back(compiled.task_spirv_source_codes);
 }
 
+void AotModuleBuilderImpl::add_compiled_kernel(aot::Kernel *kernel) {
+  kernel->save_to_module(this);
```
k-ye, line 153: I'm just thinking if we can put `save_to_module`'s impl within each module builder impl. That way `aot::Kernel` doesn't need to know anything about the module.

Furthermore, `aot::Kernel` has a misleading namespace now. It's probably more suitable to call this `CompiledKernel` or something...

ailzhang: +1 on the misleading namespace, I'll leave that to a separate search and replace PR!

taichi/aot/module_builder.h
```diff
@@ -37,6 +38,10 @@ class AotModuleBuilder {
   virtual void dump(const std::string &output_dir,
                     const std::string &filename) const = 0;
 
+  void dump_graph(std::string output_dir) const;
```
k-ye, line 41: OOC, why we need a separate `dump_graph`, instead of dumping everything in `dump`?

k-ye (approved):
LGTM!

taichi/aot/module_builder.h
ailzhang: @k-ye Since `dump_graph` is backend independent so I wanted to keep it in the base class. Note it's called in the `dump` in every backend indeed we didn't call them separately.

ailzhang:
Agh got some CLA issue on my newly setup workstation, lemme see if I can hack it around...

taichi/aot/module_builder.h
k-ye: doesnt need to be public?

ailzhang: aha nice catch! it was a legacy public api used in the first prototype. Removed in #5024! :D

PR #5022, opened 2022-05-22
[Lang] [ir] Add short-circuit if-then-else operator

re-xyr:
Related issue = #3572, #3635

This PR introduces a new ternary operator `ifte` at the frontend IR level that supports short-circuit semantics:

```
select(c, x, y) ==> 
  eval c, eval x, eval y, 
  if c 
    then return x 
    else return y

ifte(c, x, y) ==>
  eval c,
  if c
    then eval x, return x
    else eval y, return y
```

On the python side, `IfExp` now uses this operator instead of directly translating into if statements; this guarantees the return type is consistent.

@strongoier: For now `ifte(c, x, y)` does *not* force `x` and `y` to be the same type. Is this desired semantics?

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -495,7 +495,7 @@ class ASTTransformer(Builder):
                 else:
                     ctx.global_vars[
                         arg.arg] = kernel_arguments.decl_scalar_arg(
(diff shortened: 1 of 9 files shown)


strongoier (approved):
LGTM!

> For now ifte(c, x, y) does not force x and y to be the same type. Is this desired semantics?

Yep.

PR #5028, opened 2022-05-23
[llvm] Add LLVM AOT tests

k-ye:
Related issue = #4800


ailzhang (approved):
LGTM except a small nit, definitely not blocking tho

tests/cpp/backends/llvm/generated/metadata.json
```diff
@@ -0,0 +1,14 @@
+{
```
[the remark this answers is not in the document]
k-ye: I agree. But I was thinking about using CMake commands to run this, feel like that's a bit easier..

tests/run_tests.py
```diff
@@ -112,7 +120,7 @@ def test():
                         dest='cpp',
                         default=True,
```
jim19930609, line 121: Just noticed that this argument has `default = True` and `action=store_true`, is it even possible to set it as "false"?

tests/run_tests.py
```diff
@@ -209,6 +217,7 @@ def test():
 
     if args.cpp:
         _test_cpp()
+        return
```
jim19930609, line 220: With `--cpp=True` as the default value, this might turn off all the python tests:

image

k-ye:
Superseded by https://github.com/taichi-dev/taichi/pull/5079, thanks @jim19930609 !

PR #5032, opened 2022-05-23
[Lang] Fix potential precision bug when using math vector and matrix types

neozhaoliang:
Purpose: In the [current implementation](https://github.com/taichi-dev/taichi/blob/master/python/taichi/math/mathimpl.py#L13) of vector and matrix types, the type `float` is indeed determined when the module is imported, hence always `ti.f32`. This PR fixes this problem.

Also added two new functions `determinant` and `inverse` to the math module.

### DIFF
--- a/python/taichi/_kernels.py
+++ b/python/taichi/_kernels.py
@@ -71,10 +71,11 @@ def vector_to_fast_image(img: template(), out: ndarray_type.ndarray()):
         r, g, b = 0, 0, 0
         color = img[i, img.shape[1] - 1 - j]
         if static(img.dtype in [f16, f32, f64]):
-            r, g, b = min(255, max(0, int(color * 255)))
+            r, g, b = min(255, max(0, int(color * 255)))[:3]
         else:
             static_assert(img.dtype == u8)
-            r, g, b = color
+            r, g, b = color[:3]
+
         idx = j * img.shape[0] + i
         # We use i32 for |out| since OpenGL and Metal doesn't support u8 types
         if static(get_os_name() != 'osx'):
--- a/python/taichi/math/mathimpl.py
+++ b/python/taichi/math/mathimpl.py
@@ -633,8 +633,8 @@ def determinant(m):
 
 @ti.func
 def _inverse2x2(m):
-    return mat2(m[1, 1], -m[0, 1], -m[1, 0],
-                m[0, 0]) / (m[0, 0] * m[1, 1] - m[0, 1] * m[1, 0])
+    return mat2([[m[1, 1], -m[0, 1]], [-m[1, 0], m[0, 0]]
+                 ]) / (m[0, 0] * m[1, 1] - m[0, 1] * m[1, 0])
 
 
 @ti.func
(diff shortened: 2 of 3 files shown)


python/taichi/_kernels.py
```diff
@@ -71,10 +71,11 @@ def vector_to_fast_image(img: template(), out: ndarray_type.ndarray()):
         r, g, b = 0, 0, 0
         color = img[i, img.shape[1] - 1 - j]
         if static(img.dtype in [f16, f32, f64]):
-            r, g, b = min(255, max(0, int(color * 255)))
+            r, g, b = min(255, max(0, int(color * 255)))[:3]
```
neozhaoliang, line 74: Shall we simply discard the alpha channel as here, or blend the rgb channel with some predefined background using alpha?

python/taichi/math/mathimpl.py
```diff
@@ -10,44 +10,44 @@
...
-"""3D float vector type.
-"""
 
-vec4 = ti.types.vector(4, float)  # pylint: disable=E1101
-"""4D float vector type.
-"""
+def vec2(*args):
+    """2D floating vector type.
+    """
+    return ti.types.vector(2, impl.get_runtime().default_fp)(*args)  # pylint: disable=E1101
```
strongoier, line 20: Nice solution. In this way it should also be fine to use `float`, though :-)

python/taichi/math/mathimpl.py
```diff
@@ -68,6 +68,19 @@ def uvec4(*args):
     return ti.types.vector(4, _get_uint_ip())(*args)  # pylint: disable=E1101
 
 
+mat2 = ti.types.matrix(2, 2, impl.get_runtime().default_fp)  # pylint: disable=E1101
```
strongoier, line 71: Should `mat2, mat3, mat4` use the same way of definition to avoid the precision bug?

python/taichi/math/mathimpl.py
```diff
@@ -611,12 +624,133 @@ def length(x):
...
+                 ],
+                 [
+                     a11 * b07 - a10 * b09 - a12 * b06, a00 * b09 - a01 * b07 +
+                     a02 * b06, a31 * b01 - a30 * b03 - a32 * b00,
+                     a20 * b03 - a21 * b01 + a22 * b00
+                 ]]) / det
+
+
+@ti.func
+def inverse(mat):  # pylint: disable=R1710
```
strongoier, line 714: What is the difference between this method and the `inverse()` of the `Matrix` class?

python/taichi/math/mathimpl.py
neozhaoliang: Yes you're right, no need to call get_runtime here.

python/taichi/math/mathimpl.py
neozhaoliang: Thanks! I didn't realize this doesn't solve the bug actually.

python/taichi/math/mathimpl.py
neozhaoliang: I forgot that method~
Updated and using the class method now

strongoier (approved):
LGTM!

PR #5035, opened 2022-05-24
[ci] Run cpp tests via run_tests.py

k-ye:
Related issue = #

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -112,7 +112,7 @@ def test():
                         dest='cpp',
                         default=True,
                         action='store_true',
-                        help='Run the C++ tests')
+                        help='Only run the C++ tests')
     parser.add_argument('-s',
                         '--show',
                         dest='show_output',
@@ -209,6 +209,7 @@ def test():
 
     if args.cpp:
         _test_cpp()
+        return
 
     if _test_python(args) != 0:
         exit(1)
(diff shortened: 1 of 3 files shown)


ailzhang (commented):
if we do this, we might need to update `win_build_test.ps1`s as well... (I wonder why we have so many win_build_test.ps1 lol

k-ye:
@ailzhang , hmmm, I don't see `win_build_test.ps1` running the CPP tests :thinking:  https://github.com/taichi-dev/taichi/blob/9fd390b6f06656b8a0f4c06c6e64e179b0435d57/ci/windows/win_build_test.ps1#L61-L63 ?

tests/run_tests.py
```diff
@@ -112,7 +112,7 @@ def test():
                         dest='cpp',
                         default=True,
```
ailzhang, line 113: agh I see what went wrong here: because `--cpp` defaults to true, this PR actually never runs the python tests since it returns immediately after cpp tests finish... (test log in CI can confirm this). This need to be changed to `False` and  cpp tests must be run explicitly in each script. 

Sorry I think it was my bad - we sometimes forgot to run C++ tests in some test scripts so I thought let's just run them by default since they were small....

ailzhang (approved):
LGTM!
FYI I just noticed that we didn't build c++ tests on windows :( but will open a separate issue for it.

ailzhang:
FYI opened an issue at https://github.com/taichi-dev/taichi/issues/5053 but that shouldn't block this PR from merging.

Issue #4857, opened 2022-04-26
Polish and officially release quantized types


strongoier:
Quantized types are an experimental feature introduced in the [QuanTaichi paper](https://yuanming.taichi.graphics/publication/2021-quantaichi/quantaichi.pdf). With this useful feature, users can significantly save memory usage of their Taichi programs. The feature can also enable acceleration of atomic operations on mobile phones.

However, the feature has been neither officially announced nor extensively maintained. As Taichi has come to its 1.0 version, I think it is time to polish the feature and make it available to users. My plan is to refine the API and implementation so that it can fit into current Taichi better, be more user-friendly, and become deployable with Taichi AOT. I would like to write an RFC for it.

strongoier:
Before writing a formal RFC, I would like to briefly summarize some previous discussions on this topic. I think there are still some issues to be solved, and I hope to continue the discussion here.

## Background

A quantized type normally has no native support. Therefore, you need to specify a parent primitive type (e.g. a 32-bit int) and describe how you would like to pack a group of quantized types (e.g. a 15-bit int and a 17-bit int) inside.

In Taichi, this is done by introducing two SNode types, `bit_struct` and `bit_array`. Example usages:
```python
i4 = ti.quant.int(bits=4)
u28 = ti.quant.int(bits=28, signed=False)

p = ti.field(dtype=i4)
q = ti.field(dtype=u28)
ti.root.dense(ti.i, 4).bit_struct(num_bits=32).place(p, q)

r = ti.field(dtype=i4)
ti.root.dense(ti.i, 4).bit_array(ti.i, 8, num_bits=32).place(r)
```

## What are the problems of current APIs?

1. `bit_struct` and `bit_array` are not consistent with other SNode types. A normal SNode specifies two things: how to split the axes, and how are cells stored in the container. Meanwhile, a normal SNode has no limitations on components of its cells. However, `bit_struct` has nothing to do with axes, and both `bit_struct` and `bit_array` must only have `place` SNodes as components of its cells with limitation on total number of bits of all components of its cells. These make the APIs inconsistent.
2. Users cannot use quantized types outside the SNode system. This is especially problematic when it comes to deployment, because ndarrays, which are first-class citizens in Taichi AOT, cannot work with quantized types.

## What are our current thoughts on solving the problems?

As `bit_array` is deeply coupled with the SNode system (it indeed handles axes splitting) while used not that often, we prefer to keep it unchanged. Our main focus is around `bit_struct`.

[...]

PR #5036, opened 2022-05-24
[Lang] [type] Refactor quant type definition APIs

strongoier:
Related issue = https://github.com/taichi-dev/taichi/issues/4857#issuecomment-1130043445

This PR does the following:
- For `ti.types.quant.int`, set the default value of `signed` to `True`.
- Add a `scale` parameter to `ti.types.quant.fixed`.
- Replace all occurrences of `ti.types.quant._custom_float` with `ti.types.quant.float/fixed`.
- Remove redundant logic.

After this PR, users only need to know three public APIs for quant type definition - `ti.types.quant.int/fixed/float`.


ailzhang (approved):
Awesomeeee!

PR #5038, opened 2022-05-25
[opengl] Set host_write to false for opengl ndarray

ailzhang:
As discussed ndarrays can be written through calling write kernels, but
it shouldn't support directly map on host and write to it.

Related issue = #


turbo0628 (approved):
LGTM!

PR #5043, opened 2022-05-25
[Build] Exit CI builds when download of prebuilt packages fails

PENGUINLIONG:
Improve log readability for [this case](https://github.com/taichi-dev/taichi/runs/6585423431?check_suite_focus=true).

### DIFF
--- a/.github/workflows/scripts/win_build.ps1
+++ b/.github/workflows/scripts/win_build.ps1
@@ -47,11 +47,13 @@ Push-Location $libsDir
 if (-not (Test-Path "taichi_llvm")) {
     WriteInfo("Download and extract LLVM")
     curl.exe --retry 10 --retry-delay 5 https://github.com/taichi-dev/taichi_assets/releases/download/llvm10/taichi-llvm-10.0.0-msvc2019.zip -LO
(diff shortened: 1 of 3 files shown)


frostming:
I am not a Powershell genius but I thought it is handled by $ErrorActionPreference = "Stop"

PENGUINLIONG:
> I am not a Powershell genius but I thought it is handled by
> 
> https://github.com/taichi-dev/taichi/blob/e346e9f87b1fcf1151a7135248fd1981d3f477f8/.github/workflows/scripts/win_build.ps1#L11-L12

It would if you didn't install the actual `curl` and `curl` is still an alias of `Invoke-WebRequest`; but it's `curl.exe` which wouldn't be triggering any powershell error. It only handles semantical powershell errors (exceptions).

PENGUINLIONG:
BTW `Invoke-WebRequest` has a PowerShell version requirement so if you already have curl set up I'm somehow reluctant to shift to `Invoke-WebRequest`

frostming:
> but it's curl.exe which wouldn't be triggering any powershell error. It only handles semantical powershell errors (exceptions).

Oh it makes sense, then it LGTM from my side.

PR #5052, opened 2022-05-26
[bug] Ndarray type should include primitive dtype as well

ailzhang:
This PR does three things:
- Switch cgraph `Arg` to take in `ti.f32/i32` instead of string
`f32/i32` as inputs
- Fix a bug that when we produce injected ndarray args for compilation
we only produced f32 ndarrays, which won't work for ndarray of other
primitive dtypes.
- No need to specify `element_shape` if it's scalar arg or scalar
ndarray arg.

Related issue = #4786 

Thanks to @YuCrazing for reporting the bug!

### DIFF
--- a/python/taichi/examples/graph/mpm88_graph.py
+++ b/python/taichi/examples/graph/mpm88_graph.py
@@ -113,12 +113,12 @@ if __name__ == "__main__":
     if not args.baseline:
         print('running in graph mode')
         # Build graph
-        sym_x = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'x', 'f32', element_shape=(2, ))
-        sym_v = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'v', 'f32', element_shape=(2, ))
-        sym_C = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'C', 'f32', element_shape=(2, 2))
-        sym_J = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'J', 'f32', element_shape=())
-        sym_grid_v = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'grid_v', 'f32', element_shape=(2, ))
-        sym_grid_m = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'grid_m', 'f32', element_shape=())
+        sym_x = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'x', ti.f32, element_shape=(2, ))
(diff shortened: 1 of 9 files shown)


python/taichi/examples/graph/mpm88_graph.py
```diff
@@ -113,12 +113,12 @@ def init_particles(x: ti.any_arr(field_dim=1), v: ti.any_arr(field_dim=1),
     if not args.baseline:
         print('running in graph mode')
         # Build graph
-        sym_x = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'x', 'f32', element_shape=(2, ))
-        sym_v = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'v', 'f32', element_shape=(2, ))
-        sym_C = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'C', 'f32', element_shape=(2, 2))
-        sym_J = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'J', 'f32', element_shape=())
-        sym_grid_v = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'grid_v', 'f32', element_shape=(2, ))
-        sym_grid_m = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'grid_m', 'f32', element_shape=())
+        sym_x = ti.graph.Arg(ti.graph.ArgKind.NDARRAY, 'x', ti.f32, element_shape=(2, ))
```
k-ye, line 116: How about having `ScalarArg` and `NdarrayArg` in the future? That way we don't have to have a generic `element_shape`

ailzhang: SG! Btw we have similar structs at https://github.com/taichi-dev/taichi/blob/master/taichi/aot/module_data.h#L60-L83 but they were tweaked for launch kernels individually instead of graph execution. Will send out a followup PR to separate `ScalarArg` and `NdarrayArg` for the graph use case!

taichi/aot/graph_data.h
```diff
@@ -23,10 +24,38 @@ struct Arg {
...
+  // For serialization & deserialization
+  explicit Arg()
+      : tag(ArgKind::kUnknown),
+        name(""),
+        dtype_id(PrimitiveTypeID::unknown),
+        element_shape({}) {
+  }
+
+  explicit Arg(ArgKind tag,
+               std::string name,
```
k-ye, line 39: nit: for all PRs, prefer `const std::string&` (unless you can make sure the callers uses `std::move(str)`).

k-ye (approved):
LGTM!

PR #5058, opened 2022-05-30
[infra] Refactor Vulkan runtime into true Common Runtime

bobcao3:
After this PR, the previous Vulkan runtime is renamed to the "common runtime" for unified device api, under namespace `taichi::lang::gfx` and folder `taichi/runtime/gfx`

In addition, related AOT modules code are moved into the runtime as well, as they do not actually have a Vulkan dependency. This also prepares for moving OpenGL backend to the common runtime. (Currently DX11 is already using the common runtime)

Also this PR fixes a small implementation issue in OpenGL where `rw_buffer` should be used in resource binder for SSBO instead of `buffer` (which is for UBOs). In addition, this PR fixes the memory leak issue in DX11. Although the DX11 device API is still quite lacking in features, it is now as usable as a base Vulkan 1.1 implementation.

This PR also changed linker flags a bit for Windows, now LTO is only enabled for release builds (debug builds will not use LTO)


taichi/gui/gui.h
```diff
@@ -888,7 +888,7 @@ class TI_DLL_EXPORT GUI : public GUIBase {
       redraw();
       process_event();
       if (frame_id % 10 == 0)
-        set_title(fmt::format("{} ({:.2f} FPS)", window_name, real_fps));
+        set_title(fmt::format("{} ({:.2f} FPS)", "", real_fps));
```
ailzhang, line 891: Why's this change?

taichi/aot/module_loader.cpp
```diff
@@ -32,19 +32,22 @@ Kernel *KernelTemplate::get_kernel(
...
-#else
-    TI_NOT_IMPLEMENTED
+    return gfx::make_aot_module(mod_params, arch);
+#endif
+  } else if (arch == Arch::dx11) {
+#ifdef TI_WITH_DX11
+    return gfx::make_aot_module(mod_params, arch);
+#endif
+  } else if (arch == Arch::opengl) {
+#ifdef TI_WITH_OPENGL
```
ailzhang, line 42: opengl is not yet supported? Shall we leave that out of this PR?

ailzhang (commented):
Overall looks good to me! Leaving this to @k-ye to take a second look before merging.

taichi/gui/gui.h
bobcao3: Accidentally included. The text on my Windows 10 is all garbled, tried to fix it but no luck yet

taichi/runtime/gfx/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+# ./taichi/runtime/gfx/CMakeLists.txt
+
+add_library(gfx_runtime)
```
k-ye, line 3: neat!

taichi/runtime/gfx/snode_tree_manager.h
```diff
@@ -8,9 +8,9 @@
...
 namespace taichi {
 namespace lang {
-namespace vulkan {
+namespace gfx {
 
-class VkRuntime;
+class GfxRuntime;
 
 /**
  * @brief Manages the SNodeTrees for the Vulkan backend.
```
k-ye, line 16:
```suggestion
 * @brief Manages the SNodeTrees for the underlying backend.
```

k-ye (approved):
LGTM!

What's the plan for OpenGL? Is it to derive a new runtime from `gfx::GfxRuntime`?

taichi/runtime/gfx/CMakeLists.txt
bobcao3: Yes, that's the plan

qiao-bo (approved):
LGTM

PR #5061, opened 2022-05-30
[Lang] [ir] Disallow reading a whole bit_struct

strongoier:
Related issue = https://github.com/taichi-dev/taichi/issues/4857#issuecomment-1134388812

The ability to read a special intermediate SNode out is unnecessary for users, but adds API inconsistency and additional complexity to the SNode system. According to the discussion linked above, we now disallow it, and will not support it in the future API `bit_struct_wrapper`. `visualize_quant_types.py` is removed by the way because it is designed to illustrate this ability.

After this PR, `GlobalPtrExpression` can only hold a `GlobalVariableExpression` or `ExternalTensorExpression` as its source, which is much cleaner. cc @re-xyr


ailzhang (approved):
Nice!!

PR #5062, opened 2022-05-30
[bug] Stop overloading operator ! for Expr

### LINKED ISSUE #4119: Coexistence of operator! and operator bool in lang::Expr is very likely to cause wrong use
### Background
The `operator!` and `operator bool` coexist in `lang::Expr`. The `operator bool` means if the expr holds a nonnull ptr to `lang::Expression`. The `operator!` will return a new Expr which holds a `logic_not`-`UnaryOpExpression` of `this->expr`.

I used like :
```C++
// expr is a instance of lang::Expr
if (!expr) {
  // do somthing
}
```
The meaning I thought : `if (!expr.operator bool()) { ... }`
True meaning     : `if (expr.operator!().operator bool()) { ... }`

Because of this, I spent nearly half a day in debuging😭.

### My thought
Coexistence of `operator!` and `operator bool` in `lang::Expr` is very likely to cause wrong use when coding. Maybe we can remove one which is not or rarely used, and reset it to a normal member function instead of operator overloading if necessary?

(**At least, the contributors who see the issue will be vigilant when using like `if(!expr)`**:stuck_out_tongue_closed_eyes:.

strongoier:
Related issue = fix #4119

There is already `expr_logic_not()` for the `logic_not` operation. Overloading `!` is redundant, and causes trouble in #4119. This PR removes it.

### DIFF
--- a/taichi/ir/expr.cpp
+++ b/taichi/ir/expr.cpp
@@ -50,10 +50,6 @@ SNode *Expr::snode() const {
   return cast<GlobalVariableExpression>()->snode;
 }
 
-Expr Expr::operator!() {
-  return Expr::make<UnaryOpExpression>(UnaryOpType::logic_not, expr);
-}
-
 void Expr::set_grad(const Expr &o) {
   this->cast<GlobalVariableExpression>()->adjoint.set(o);
 }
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
LGTM!

PR #5066, opened 2022-05-31
[refactor] Pass element_shape and layout to C++ Ndarray

ailzhang:
* #5073
* __->__ #5066
* #5065

Note we still flatten element_shape in the C++ Ndarray, which is blocked by the accessors and will be fixed
in the following PRs.

### DIFF
--- a/taichi/backends/opengl/codegen_opengl.cpp
+++ b/taichi/backends/opengl/codegen_opengl.cpp
@@ -486,13 +486,14 @@ class KernelGen : public IRVisitor {
     const auto &element_shape = stmt->element_shape;
     std::vector<std::string> size_var_names;
     std::vector<std::string> element_shape_size_var_names;
-    enum ExternalArrayLayout { layout_AOS = 0, layout_SOA = 1 };
-    const auto layout = stmt->element_dim <= 0 ? layout_AOS : layout_SOA;
+
+    const auto layout = stmt->element_dim <= 0 ? ExternalArrayLayout::kAOS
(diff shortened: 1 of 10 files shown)


turbo0628:
It saves lots of guess works in the codegen! Looking forward to the changes!

taichi/backends/opengl/codegen_opengl.cpp
```diff
@@ -486,13 +486,14 @@ class KernelGen : public IRVisitor {
...
-    const auto layout = stmt->element_dim <= 0 ? layout_AOS : layout_SOA;
+
+    const auto layout = stmt->element_dim <= 0 ? ExternalArrayLayout::kAOS
+                                               : ExternalArrayLayout::kSOA;
 
     if (element_shape.size() > 0) {
       int elem_beg = 0;
       int elem_end = 0;
-      if (layout == layout_SOA) {
+      if (layout == ExternalArrayLayout::kSOA) {
```
k-ye, line 496: IIUC, once element_shape is no longer passed into extra_args, we can  remove this part one...? (Metal now has the same stuff due to the Ndarray support) cc @turbo0628

k-ye (approved):
LGTM!

taichi/backends/opengl/codegen_opengl.cpp
ailzhang: Yup we briefly discussed this in the afternoon and @turbo0628 will get rid of these hacks later (still need some more cleanup before that

PR #5069, opened 2022-05-31
[Lang] Add more initialization routines for glsl matrix types

neozhaoliang:
This PR is intended for adding support for glsl style matrix constructions.

Feature added:

1. Use n vectors to initialize a matrix.
```python
x = vec3(1, 0, 0)
y = vec3(1, 1, 0)
z = vec3(1, 1, 1)
m = mat3(x, y, z)
```

2. Use a 1d list, or simple n*n scalars to initialize a matrix:

```python
x = mat2(1, 0, 0, 1)
y = mat2([1, 0, 0, 0])
```

3. The usual way: (without this PR)

```python
x = mat2([[1, 0], [0, 1]])
y = mat2(1)
```

### DIFF
--- a/python/taichi/math/mathimpl.py
+++ b/python/taichi/math/mathimpl.py
@@ -68,22 +68,46 @@ def uvec4(*args):
     return ti.types.vector(4, _get_uint_ip())(*args)  # pylint: disable=E1101
 
 
+def _gen_matrix(n, *args):
+    """Supports more matrix construction routines.
+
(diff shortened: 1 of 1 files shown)


YuCrazing (approved):
LGTM!

A small suggestion: `mat3()/mat3()/mat4()` can accept many different kinds of inputs, but actually, users have no knowledge about it. so I think we should provide a very detailed comment or documentation to tell users what kind of inputs they can pass in.

PR #5070, opened 2022-05-31
[llvm] [aot] CUDA-AOT PR #0: Refactored compile_module_to_executable() to CUDAModuleToFunctionConverter

jim19930609:
Related issue = #4800

`compile_module_to_executable()` was used in LLVM - CUDA backend to perform:
1. Transform llvm::Module into CudaJITModule (CudaModule in essense), to make the compiled functions executable.
2. Prepare the execution entrance: a lambda function that does some additional configurations to the RuntimeContext, then execute the offloaded_tasks iteratively.

In this Patch, we refactored `compile_module_to_executable()` to make it decoupled with `Kernel`. In this way, we dont have to recover the entire `Kernel` at AOT runtime. 

`compile_module_to_executable()` was further replaced by `CUDAModuleToFunctionConverter::convert()` to stay consistent with LLVM - CPU backend.

### DIFF
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -15,6 +15,7 @@
 #include "taichi/backends/cuda/cuda_context.h"
 #include "taichi/codegen/codegen_llvm.h"
 #include "taichi/llvm/llvm_program.h"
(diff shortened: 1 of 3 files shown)


k-ye (approved):
LGTM!

Issue #5044, opened 2022-05-25
[aot] remove element shape from extra args


turbo0628:
As element shape is invariant during AOT deployment, we would like to specialize the shape at compile time.
Currently, the specialization is done for OpenGL and Metal backends, but not yet for Vulkan and LLVM.

Tasks
- [x] Specialize element shape for Vulkan backend.
- [x] Specialize element shape for LLVM backend.
- [x] Fix element shape reading problem.
- [x] Assign correct offset value for ExternalTensorShapeAlongAxisStmt in extra args. (Amend struct-fors)
- [x] Exclude the `element_shape` from `shape` for ndarrays.
- [x] Remove element shape in extra args.
Follow-up task:
- [ ] Pass in the layout information for `ExternalPtrStmt` rather than using a negative element dim.

turbo0628:
Another task

- [ ] Exclude the `element_shape` from `shape` for ndarrays.

turbo0628:
The refactored code seems to get the "naked" element shape (e.g. 4 for f32) to the memory allocator. However, the desired behavior should be reading the size of "element packs" (e.g. 16 for (2, 2) f32 elements)

Working on this problem.

turbo0628:
The `ExternalArrayShapeAlongAxisStmt` also reads the shape information in extra args. Need to de-offset the axis id.

turbo0628:
The statement still use negative element dims for layout information. Will fix this hack in a follow-up PR

PR #5071, opened 2022-05-31
[llvm] Specialize element shape for LLVM backend

turbo0628:
Related issue = #5044

### DIFF
--- a/taichi/codegen/codegen_llvm.cpp
+++ b/taichi/codegen/codegen_llvm.cpp
@@ -1643,7 +1643,8 @@ void CodeGenLLVM::visit(ExternalPtrStmt *stmt) {
   int element_shape_idx = 0;
   for (int i = 0; i < num_indices; i++) {
     if (i >= element_shape_begin && i < element_shape_end) {
-      llvm::Value* size_var = tlctx->get_constant(element_shape[element_shape_idx++]);
+      llvm::Value *size_var =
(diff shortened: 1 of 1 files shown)


taichi/codegen/codegen_llvm.cpp
```diff
@@ -1623,8 +1640,15 @@ void CodeGenLLVM::visit(ExternalPtrStmt *stmt) {
       llvm::PointerType::get(tlctx->get_data_type(dt), 0));
 
   auto linear_index = tlctx->get_constant(0);
+  int element_shape_idx = 0;
   for (int i = 0; i < num_indices; i++) {
```
k-ye, line 1644: Just to confirm. In the future `num_indices` should always be equal to `len(array_shape)`, right? @ailzhang @turbo0628

k-ye (approved):
LGTM!

taichi/codegen/codegen_llvm.cpp
turbo0628: If array_shape excludes element_shape, then yes. I plan to remove element shape from the `indices` vector in a follow-up PR.

ailzhang: Yup that's exactly the goal of these refactors.

k-ye: Cool!

Issue #4935, opened 2022-05-09
[cuda] CUDA warp level intrinsics failed on legacy GPU 


0xzhang:
- GeForce GTX 1650
- CUDA 11.4

It seems `test_simt.py` not testable on my PC.
I use command below.
```shell
python tests/run_tests.py -t1 -k test_simt
```

![image](https://user-images.githubusercontent.com/33616362/167296617-43fe9890-07e9-4fe2-826b-bd0bad83889c.png)

Warp functions have requirements to compute capability of device. Maybe my card doesn't have enough compute capability according the [list](https://developer.nvidia.com/cuda-gpus). Or maybe the WSL environment limits the compute capability.

Full error info be archived [here](https://gist.github.com/0xzhang/1929ad8d044671d05a3fac717c703843).

Related PR is #4927.

If taichi could capture this problem in advance would be even better.

ailzhang:
Hey @0xzhang , thanks for reporting the issue! FYI @qiao-bo is OOO this week, so please expect a delay in his reply ;) Thanks a lot!

0xzhang:
> Hey @0xzhang , thanks for reporting the issue! FYI @qiao-bo is OOO this week, so please expect a delay in his reply ;) Thanks a lot!

Thank you for telling me. You are welcome! 😄

qiao-bo:
Hi, sorry for my late reply. We should indeed detect this in advance by inspecting the GPU CC, we will fix this.

qiao-bo:
This should be fixed by 5077. Feel free to reopen if you encounter further problem.

0xzhang:
> This should be fixed by 5077. Feel free to reopen if you encounter further problem.

Thanks!

PR #5077, opened 2022-06-01
[cuda] [simt] Add assertions for warp intrinsics on old GPUs

qiao-bo:
Related issue = #4935  #4631 

Local environment: Quadro P1000 with CC 6.1. Match intrinsics fail due to requirements on CC >= 70.


turbo0628 (approved):
LGTM!

PR #5079, opened 2022-06-01
[llvm] [aot] Add LLVM-CPU AOT tests

jim19930609:
Related issue = #4800

Unlike regular C++ or Python unit tests, an AOT test is composed of:
1. A python script that compiles the Kernels and serialize into file.
2. A C++ test that loads the file then perform execution.

Each AOT test is executed solely and separated from other C++ unit tests.

Due to its unique execution manner, AOT test writers have to specify some config information in `aot_test_cases_`:
```
# { cpp_test_name : [python_test_path, python_function_name] }
"LlvmProgramTest.FullPipeline": [os.path.join('cpp', 'backends', 'llvm', 'cpu_aot.py'), "aot_compile"],
```

The function signature for `python_function_name` has to be `def xxxx(folder_path): `, which will be imported and executed by `test_runner.py`. After that, `test_runner.py` will execute the corresponding C++ function named `cpp_test_name`.


k-ye (approved):
LGTM!

tests/cpp/backends/llvm/cpu_aot.py
```diff
@@ -0,0 +1,23 @@
+import os
+import pathlib
+import shutil
+
+import taichi as ti
+
+
+def aot_compile(dir_name):
+    ti.init(arch=ti.x64)
+    ti.set_logging_level(ti.TRACE)
```
qiao-bo, line 10: is this line intended?

qiao-bo (approved):
LGTM

tests/run_tests.py
```diff
@@ -1,32 +1,69 @@
...
-            "if you want to enable it.")
+
+        cmd = [f'./{cpp_test_filename}']
+        if gtest_option: cmd.append(gtest_option)
+        if extra_env: env_copy.update(extra_env)
+
+        subprocess.check_call(cmd, env=env_copy, cwd=build_dir)
+
+
+def _test_cpp_aot():
```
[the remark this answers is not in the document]
jim19930609: great idea, thanks!

tests/test_utils.py
```diff
@@ -12,6 +12,11 @@
 
 import taichi as ti
 
+aot_test_cases_ = {
```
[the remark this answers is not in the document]
jim19930609: done, thanks!

tests/cpp/backends/llvm/cpu_aot.py
jim19930609: Just removed those debug info, thanks!

tests/test_utils.py
```diff
@@ -12,6 +12,35 @@
...
+AOT test writer will have to configure your test case for "__aot_test_cases",
+the format of which follows:
+
+        "cpp_test_name" : [ "python_test_path", "python_function_name" ]
+
+For example:
+
+        "LlvmProgramTest.FullPipeline": ["cpp/backends/llvm.cpu_aot.py", "aot_compile"]
+
+Note that the function signature for "python_function_name" has to be "def xxxx(folder_path):", which will be imported and executed by test_runner.py. After that, test_runner.py will execute the corresponding C++ test named "cpp_test_name".
```
ailzhang, line 38: is there a `test_runner.py`? ;)

tests/run_tests.py
```diff
@@ -1,32 +1,69 @@
...
+
+        # Run AOT Python codes
+        python_file_path = python_file_path.replace("/", ".")
+        if python_file_path.endswith(".py"):
+            python_file_path = python_file_path[:-3]
+        module = importlib.import_module(python_file_path)
+        assert function_name in module.__dict__.keys()
+
+        function_ptr = module.__dict__[function_name]
+        function_ptr(temp_folderpath)
```
ailzhang, line 50: this part is a bit hacky hmmm

ailzhang (approved):
LGTM!

tests/run_tests.py
jim19930609: It is...Essentially this part is trying to mimic:
```
from cpu_aot.py import aot_compile
aot_compile(folder_path)
```

Do you know if there's any better way of importing a function then execute it? 

The other way around is to fork another subprocess that runs `python3 aot_compile.py --temp_folder=xxx`. However, the forked python process is completely decoupled with the currently python process (the one running run_tests.py started by the user). This could confuse the user due to the diff in environment or even the Python executable (subprocess may not use the same "Python" as the one currently running).

tests/test_utils.py
jim19930609: done, thanks!

tests/run_tests.py
ailzhang: @jim19930609 I personally feel `python3 aot_compile.py --temp_folder=xxx` is slightly cleaner and easier to maintain. For example we can use `exec` to execute a file if you want to avoid using a different python. https://stackoverflow.com/questions/1027714/how-to-execute-a-file-within-the-python-interpreter. 
I feel like running a file (so that we can save the mapping from cpp test to python function will simplify things a lot?

jim19930609: sounds good, let me refactor this part a little bit.

jim19930609: Looks like "exec()" will mess up the Python AST, causing the Taichi compiler to fail. I turned to use "sys.executable" to guarantee Python executable consistency when forking subprocesses.

ailzhang (approved):
LGTM! Thanks!

PR #5081, opened 2022-06-01
[Example] Update visual effects of mass_spring_3d_ggui.py

neozhaoliang:
Add checker pattern to the cloth:

frame056


ailzhang (approved):
Thanks！

PR #5082, opened 2022-06-01
[build] [refactor] Change CMake global include_directories to target based function

qiao-bo:
Related issue = #4832 

This is an intermediate step towards target based CMake. `include_directories` are global, thus polluting all targets. This PR replace these functions by target specific `target_include_directories`. Some repetition cannot be avoided at this stage. Let's first be specific on which include dirs are needed by which target.


PENGUINLIONG (approved):
lgtm

Issue #5055, opened 2022-05-27
[RFC] Add forward mode for autodiff


erizmr:
In this issue, we would like to share a draft implementation plan for the forward mode autodiff.  

# Background

In general, there are two modes for autodiff: reverse mode and forward mode. The two modes have their advantage in different scenarios. The reverse mode is more efficient when the number of inputs is much more than the outputs (e.g., machine learning cases, thousands of trainable parameters and one scalar loss). On the contrary, the forward mode is more efficient. In addtion, the second-order derivatives can be efficiently computed by combining both the forward and reverse mode. 

For a roadmap for the autodiff feature in Taichi, please check out #5050.

# Goals

- Implement forward mode autodiff.
- Design python interface for forward and reverse mode.
- Make it possible to apply both forward/reverse mode iteratively (e.g., `forward(reverse())`), preparing for computing second-order derivatives.

# Implementation Roadmap

- Implement forward mode autodiff.
	- [ ] Implement forward mode routine for operators currently supported by reverse mode
	- [ ] Handle mutation (e.g., for-loops)
	- [ ] Handle control flow
	- [ ] Unify the shared components for reverse and forward mode
	- [ ] Handle the `stop gradients` for `dual` snodes
	

- Design python interface for forward and reverse mode.
	- [ ] Decouple the `adjoint` and `grad`, make the `grad` including both `adjoint` and `dual`
	- [ ] A context manager to switch between reverse and forward mode

- Test cases
	- [ ] Add python test cases for all the operators

# Discussions

- How many kernels we need to compile for forward mode autodiff?

Currently in reverse mode, two kernels (`original kernel` and `grad kernel`) for evaluating function values and compute the gradients respectively. However, in forward mode autodiff, the derivatives are computed eagerly during the function evaluating process, i.e., the functions values and gradients can be computed using only kernel. This raise the question whether need to compile one or two kernels.

victoriacity:
I wonder if [explicitly differentiating a function as in JAX](https://jax.readthedocs.io/en/latest/jax-101/04-advanced-autodiff.html) will be supported, for example,
```python
@ti.func
def f(x): return x**3 + 2*x**2 - 3*x + 1

dfdx = forward(f)

@ti.kernel
def k() -> float:
    return dfdx(1.0)
k() # returns 4.0
```

erizmr:
I think it is possible to support this feature.  A naive current Taichi equivalent is:

```py
import taichi as ti

ti.init()

[...]

PR #5083, opened 2022-06-01
[autodiff] Allocate dual and adjoint snode

erizmr:
Related issue = #5055
Decompose the vague grad to adjoint (for reverse mode) and dual (for forward mode). Support allocating both adjoint and dual SNodes as we need both when computing second-order derivatives.


ailzhang (approved):
LGTM!

PR #5085, opened 2022-06-01
[refactor] Make sure Ndarray shape is field shape

ailzhang:
#5044 
This PR makes sure Ndarray shape is field shape instead of flattened field shape + element shape. It also got rid of `num_active_indices` in Ndarray class as it's just redundant. 

Now that we have two shapes which may be ordered differently depending on layout, a unified `total_shape` still comes in handy in some implementations. But it shouldn't be used by users directly (if they want they should sum up element_shape and field_shape themselves). 

[Update: this is fixed. We still need to update codegen but at runtime we no longer set element_shape in extra_args] ~Note this PR still need to set element_shape in extra_args in the `set_arg_ndarray`. It's blocked by the fact that predefined kernels like `fill_ndarray_matrix` is supposed to work for all element_shapes. In other words element_shape is a runtime information for them. I'll leave that for @turbo0628 ;).~

### DIFF
--- a/taichi/program/kernel.cpp
+++ b/taichi/program/kernel.cpp
@@ -269,8 +269,17 @@ void Kernel::LaunchContextBuilder::set_arg_ndarray(int arg_id,
                                /*is_device_allocation=*/true);
   TI_ASSERT_INFO(arr.shape.size() <= taichi_max_num_indices,
                  "External array cannot have > {max_num_indices} indices");
-  for (uint64 i = 0; i < arr.shape.size(); ++i) {
-    this->set_extra_arg_int(arg_id, i, arr.shape[i]);
+  // TODO: Update the codegen so that we don't reserve slots for element_shape
+  // in extra_args, especially in SOA case.
+  if (arr.layout == ExternalArrayLayout::kAOS) {
+    for (uint64 i = 0; i < arr.shape.size(); ++i) {
+      this->set_extra_arg_int(arg_id, i, arr.shape[i]);
+    }
+  } else {
+    auto element_dim = arr.element_shape.size();
+    for (uint64 i = element_dim; i < arr.total_shape().size(); ++i) {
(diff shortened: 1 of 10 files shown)


taichi/program/kernel.cpp
```diff
@@ -269,8 +269,9 @@ void Kernel::LaunchContextBuilder::set_arg_ndarray(int arg_id,
                                /*is_device_allocation=*/true);
   TI_ASSERT_INFO(arr.shape.size() <= taichi_max_num_indices,
                  "External array cannot have > {max_num_indices} indices");
-  for (uint64 i = 0; i < arr.shape.size(); ++i) {
-    this->set_extra_arg_int(arg_id, i, arr.shape[i]);
+  for (uint64 i = 0; i < arr.total_shape().size(); ++i) {
+    // FIXME: we should use arr.shape instead of arr.total_shape() here!
```
ailzhang, line 273: cc: @turbo0628 I believe this is the last step needed for https://github.com/taichi-dev/taichi/issues/5044. ;)

k-ye:
> predefined kernels like fill_ndarray_matrix is supposed to work for all element_shapes

I thought these are kernel templates. So by the time they are instantiated, they should know the element_shape at compile time?

ailzhang:
> > predefined kernels like fill_ndarray_matrix is supposed to work for all element_shapes
> 
> I thought these are kernel templates. So by the time they are instantiated, they should know the element_shape at compile time?

@k-ye Yea sorry I thought my failed SOA tests required full element_shape + field shape to be set, but then I realized that the reason they fail is because our field shape doesn't start at index 0 of extra_args in SOA case. In this PR I just set field_shape to the expected position in extra_args, but left a todo that we need to update codegen to stop allocating space for element_shape cc: @turbo0628

turbo0628 (approved):
Thanks! LGTM!

Issue #3382, opened 2021-11-04
[refactor] Refactor codegen_llvm.cpp


sjwsl:
Currently, https://github.com/taichi-dev/taichi/blob/master/taichi/codegen/codegen_llvm.cpp is full of if/else, which is ugly and hard to maintain. It's worth spending some time refactoring.

PR #5086, opened 2022-06-01
[llvm] [refactor] Merge AtomicOpStmt codegen in CPU and CUDA backends

strongoier:
Related issue = #3382

AtomicOpStmt codegen in CUDA backend was once refactored in #2490. Same thing should also be done for CPU backends and the result code can actually be merged.


ailzhang (approved):
Nice!

PR #5088, opened 2022-06-02
[autodiff] Extract shared components for reverse and forward mode

erizmr:
Related issue = #5055

### DIFF
--- a/taichi/transforms/auto_diff.cpp
+++ b/taichi/transforms/auto_diff.cpp
@@ -488,74 +488,174 @@ class ReverseOuterLoops : public BasicStmtVisitor {
   }
 };
 
+// Base class for both reverse (make adjoint) and forward (make dual) mode autodiff
+class ADTransform : public IRVisitor{
+  protected:
+    Stmt *constant(float32 x) {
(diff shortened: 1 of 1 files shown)


taichi/transforms/auto_diff.cpp
```diff
@@ -593,17 +693,12 @@ class MakeAdjoint : public IRVisitor {
     }
   }
 
-  Stmt *insert_back(std::unique_ptr<Stmt> &&stmt) {
+  Stmt *insert_ad_transform_stmt(std::unique_ptr<Stmt> &&stmt) override {
```
ailzhang, line 696: nit: shall we keep the old name `insert_back` for this function for now? ;) I guess you changed it likely because forward mode insert in a different place? That could be passed in as an additional arg or simply `insert_grad_stmt`?

ailzhang (approved):
LGTM!

taichi/transforms/auto_diff.cpp
erizmr: Have changed them to `insert_grad_stmt`.

PR #5091, opened 2022-06-02
[llvm] [refactor] Use LLVM native atomic ops if possible

strongoier:
This is a follow-up PR of #5086. `FAdd`, `UMin`, `UMax` are actually supported by LLVM, so there is no need to call manually-implemented functions. The code is also refactored in a cleaner way (no more `real_or_unsigned_type_atomic`).


ailzhang (approved):
LGTM！

PR #5093, opened 2022-06-06
[bug] Minor fix for ndarray element_shape in graph mode

ailzhang:
Now that ndarray's element_shape is separated from shape, this hack can
be removed.

Related issue = #


turbo0628 (approved):
LGTM!

PR #5094, opened 2022-06-06
[aot] Use pre-calculated runtime array size for gfx runtime.

turbo0628:
Related issue = #5044

The current GFX runtime reads from extra args buffer to calculate the shape of ext arrs. If we simply remove the element shape from extra args, the memory allocator would misbehave.

These information has already been stored in the `host_ctx`. Use the pre-calculated information instead.


ailzhang (approved):
Awesome thanks!

PR #5095, opened 2022-06-06
[refactor] Update Ndarray constructor used in AOT runtime.

ailzhang:
This constructor is mainly used to construct an Ndarray out of an
existing device allocation. This PR updates the behavior of this
constructor to seprate element_shape out of shape.

Related issue = #


turbo0628 (approved):
Thanks, LGTM!

PR #5098, opened 2022-06-06
[autodiff] Add forward mode pipeline for autodiff pass

erizmr:
Related issue = #5055

### DIFF
--- a/taichi/ir/transforms.h
+++ b/taichi/ir/transforms.h
@@ -80,7 +80,8 @@ bool lower_access(IRNode *root,
                   const LowerAccessPass::Args &args);
 void auto_diff(IRNode *root,
                const CompileConfig &config,
-               bool use_stack = false);
+               bool use_stack = false,
+               bool reverse_mode = true);
 /**
  * Determine all adaptive AD-stacks' size. This pass is idempotent, i.e.,
  * there are no side effects if called more than once or called when not needed.
@@ -149,6 +150,7 @@ void compile_to_offloads(IRNode *ir,
                          bool verbose,
                          bool grad,
                          bool ad_use_stack,
+                         bool ad_reverse_mode,
                          bool start_from_ast);
 
 void offload_to_executable(IRNode *ir,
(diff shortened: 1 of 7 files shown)


taichi/transforms/compile_to_offloads.cpp
```diff
@@ -35,13 +35,14 @@ void compile_to_offloads(IRNode *ir,
...
                          bool ad_use_stack,
+                         bool ad_reverse_mode,
                          bool start_from_ast) {
   TI_AUTO_PROF;
 
   auto print = make_pass_printer(verbose, kernel->get_name(), ir);
   print("Initial IR");
 
-  if (grad) {
+  if (grad && ad_reverse_mode) {
```
ailzhang, line 45: I wonder if ad_reverse_mode=true implies grad=True? Given that forward mode doesn't really need a grad kernel?

taichi/ir/transforms.h
```diff
@@ -166,6 +168,7 @@ void compile_to_executable(IRNode *ir,
                            Kernel *kernel,
                            bool grad,
                            bool ad_use_stack,
+                           bool ad_reverse_mode,
```
ailzhang, line 171: By any chance we can merge grad and ad_reverse_mode to a enum?

ailzhang (commented):
This PR looks great to me! Just had a minor question about if we can further simplify the implementation. Thanks a lot!

taichi/transforms/compile_to_offloads.cpp
erizmr: Yes. I agree that ad_reverse_mode=true implies grad=True. Though we doesn't need a grad kernel for forward mode, an additional argument is required to tell that the kernel requires a autodiff pass ?

ailzhang (commented):
By looking at the code I feel like it might be cleaner to handle with and without stack separately instead of putting them together in AutodiffMode but happy to discuss this!

erizmr:
> By looking at the code I feel like it might be cleaner to handle with and without stack separately instead of putting them together in AutodiffMode but happy to discuss this!

Does it mean we can keep the argument `use_stack` and only reserve the `REVERSE_MODE` in the `AutodiffMode`?

PR #5099, opened 2022-06-06
[llvm] [refactor] Move load_bit_pointer() to CodeGenLLVM

strongoier:
Related issue = #4857, #3382

`load_bit_pointer()` is a function paired with `create_bit_ptr_struct()`, which belongs to `CodeGenLLVM` instead of `LLVMModuleBuilder`.


ailzhang (approved):
LGTM!

PR #5100, opened 2022-06-06
[refactor] Remove ndarray element shape from extra arg buffer

### LINKED ISSUE #5044: [aot] remove element shape from extra args

turbo0628:
Fix #5044 

This is the final PR to remove element shape form args buffer.

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -656,13 +656,17 @@ class Kernel:
                     is_numpy = isinstance(v, np.ndarray)
                     is_torch = isinstance(v,
                                           torch.Tensor) if has_torch else False
+                    is_soa = needed.layout == Layout.SOA
+                    element_dim = needed.element_dim
+                    array_shape = v.shape[
+                        element_dim:] if is_soa else v.shape[:element_dim]
                     if is_numpy:
                         tmp = np.ascontiguousarray(v)
                         # Purpose: DO NOT GC |tmp|!
                         tmps.append(tmp)
                         launch_ctx.set_arg_external_array_with_shape(
                             actual_argument_slot, int(tmp.ctypes.data),
-                            tmp.nbytes, v.shape)
+                            tmp.nbytes, array_shape)
                     elif is_torch:
                         is_ndarray = False
                         tmp, torch_callbacks = self.get_torch_callbacks(
@@ -670,7 +674,7 @@ class Kernel:
                         callbacks += torch_callbacks
                         launch_ctx.set_arg_external_array_with_shape(
                             actual_argument_slot, int(tmp.data_ptr()),
-                            tmp.element_size() * tmp.nelement(), v.shape)
+                            tmp.element_size() * tmp.nelement(), array_shape)
                     else:
                         # For now, paddle.fluid.core.Tensor._ptr() is only available on develop branch
                         tmp, paddle_callbacks = self.get_paddle_callbacks(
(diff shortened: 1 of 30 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -656,29 +656,35 @@ def func__(*args):
                     is_numpy = isinstance(v, np.ndarray)
                     is_torch = isinstance(v,
                                           torch.Tensor) if has_torch else False
+                    is_soa = needed.layout == Layout.SOA
+                    array_shape = v.shape
+                    element_dim = needed.element_dim
```
ailzhang, line 661: This is not super straightforward for readers ;) - shall we add a note here explaining that we are viewing an external array with specified element_dim etc?

python/taichi/lang/matrix.py
```diff
@@ -1793,7 +1793,6 @@ def __init__(self, n, dtype, shape, layout):
         self.layout = layout
         self.shape = tuple(shape)
         self.element_type = TensorType((n, ), self.dtype)
-        # TODO: pass in element_type, shape, layout directly
```
ailzhang, line 1796: This TODO still applies IMHO : in the ideal world we'll want to pass in the whole element_type  once we have vec3/matrix types in C++. But now we'll have to split it as  dtype + shape :P

taichi/transforms/lower_ast.cpp
```diff
@@ -313,16 +313,9 @@ class LowerAST : public IRVisitor {
...
-        for (int i = tensor->element_dim; i < tensor->dim; i++) {
-          shape.push_back(fctx.push_back<ExternalTensorShapeAlongAxisStmt>(
-              i, tensor->arg_id));
-        }
-      } else {  // Layout.AOS
-        for (int i = 0; i < tensor->dim + tensor->element_dim; i++) {
-          shape.push_back(fctx.push_back<ExternalTensorShapeAlongAxisStmt>(
-              i, tensor->arg_id));
-        }
+      for (int i = 0; i < tensor->dim - abs(tensor->element_dim); i++) {
```
ailzhang, line 316: hmmm do we still have negative element_dim?

taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -565,29 +565,14 @@ class TaskCodegen : public IRVisitor {
       const int num_indices = stmt->indices.size();
       std::vector<std::string> size_var_names;
       const auto &element_shape = stmt->element_shape;
-      enum ExternalArrayLayout { layout_AOS = 0, layout_SOA = 1 };
-      const auto layout = stmt->element_dim <= 0 ? layout_AOS : layout_SOA;
+      const auto layout = stmt->element_dim <= 0 ? ExternalArrayLayout::kAOS
```
ailzhang, line 568: Ideally we'll have a followup PR that properly pass in layout info ;) ?

ailzhang (approved):
Awesommme!! Thanks a ton for removing this!

PR #5102, opened 2022-06-07
[type] [refactor] Remove redundant promotion for custom int in type_check

strongoier:
Related issue = #4857

When loaded, a custom int is guaranteed to get cast into its compute type. Therefore, the type check for `BinaryOpStmt` here does redundant promotion. This part should be removed to avoid misleading developers.


ailzhang (approved):
LGTM!

PR #5104, opened 2022-06-07
[test] Save mpm88 graph in python and load in C++ test.

ailzhang:
This is a simplified version of
https://github.com/ailzhang/taichi-aot-demo/tree/mpm88_cgraph_demo which
strips the GGUI rendering part. Let's add this as a test (as well as
demo ;) ) in the codebase. We used to test the saving part of mpm88
btw and it was replaced with this e2e test.

Huge thanks to @k-ye for help debugging the GGUI rendering issue!

Related issue = #


turbo0628 (approved):
Thanks, the new test looks good!

PR #5106, opened 2022-06-08
[build] [refactor] Move Vulkan runtime out of backends dir

qiao-bo:
Related issue = #4832 

Changes here:
- Split Vulkan files into `runtime/program_impl` and `backends/vulkan`, `backends` folder will later be renamed to `rhi` after all backends' program_impls are moved.
- Build dependencies are followed as: 
`rhi` (device api) < `gfx_runtime` <- `program_impl` <- `core` (ir and stuff,  currently also include codegen, will be further split).


bobcao3:
Sorry I'm a bit confused, why is this moved to runtime? Are we going to move all "Program Impl" to runtimes? What about the Device API implementations? (which the majority of the code here are Device API impl)

qiao-bo:
> Sorry I'm a bit confused, why is this moved to runtime? Are we going to move all "Program Impl" to runtimes? What about the Device API implementations? (which the majority of the code here are Device API impl)

"Program Impl" is regarded as part of runtime (correct me if i am wrong @k-ye). So the idea is to move all runtime to /runtime. Codegen to /codegen. For the device API implementation, since only runtime depends on it (at least it should be), we can put it either within runtime or (later) create another target for this?

bobcao3:
> > Sorry I'm a bit confused, why is this moved to runtime? Are we going to move all "Program Impl" to runtimes? What about the Device API implementations? (which the majority of the code here are Device API impl)
> 
> "Program Impl" is regarded as part of runtime (correct me if i am wrong @k-ye). So the idea is to move all runtime to /runtime. Codegen to /codegen. For the device API implementation, since only runtime depends on it (at least it should be), we can put it either within runtime or (later) create another target for this?

Then the device API code should stay in "backends"? Since afterall the "device.h" is under backends.

qiao-bo:
> > > Sorry I'm a bit confused, why is this moved to runtime? Are we going to move all "Program Impl" to runtimes? What about the Device API implementations? (which the majority of the code here are Device API impl)
> > 
> > 
> > "Program Impl" is regarded as part of runtime (correct me if i am wrong @k-ye). So the idea is to move all runtime to /runtime. Codegen to /codegen. For the device API implementation, since only runtime depends on it (at least it should be), we can put it either within runtime or (later) create another target for this?
> 
> Then the device API code should stay in "backends"? Since afterall the "device.h" is under backends.

I am fine with leaving device API code to backends, then there will be another target called `backend` in our build dependency graph.

k-ye:
Hmm, how about this:

* `rhi`: hosts Unified Device API, this is independent of Taichi.
* `runtime/gfx`: common Taichi runtime. `runtime/gfx` could have backend subfolders, e.g. `opengl`, `vulkan`, ...
* `runtime/program_impls`: Hosts each backend's `ProgramImpl`, will have a subfolder for each backend.

Dependencies:

`rhi` <-- `runtime/gfx` <-- `runtime/program_impls`

AmesingFlank (approved):

PR #5108, opened 2022-06-08
[aot] [llvm] LLVM AOT Field #0: Implemented FieldCacheData & refactored initialize_llvm_runtime_snodes()

jim19930609:
Related issue = #4800 

1. Implemented `FieldCacheData` to serialize necessary compiled information used for field initialization at runtime.

2. Refactored field runtime initialization function "initialize_llvm_runtime_snodes()", so that it only uses compiled information stored in  FieldCacheData.

### DIFF
--- a/taichi/llvm/llvm_offline_cache.h
+++ b/taichi/llvm/llvm_offline_cache.h
@@ -45,7 +45,7 @@ struct LlvmOfflineCache {
   struct FieldCacheData {
     struct SNodeCacheData {
       int id;
-      int type;
+      SNodeType type;
       size_t cell_size_bytes;
       size_t chunk_size;
 
@@ -53,6 +53,7 @@ struct LlvmOfflineCache {
     };
 
     int tree_id;
+    int root_id;
     size_t root_size;
     std::vector<SNodeCacheData> snode_metas;
 
--- a/taichi/llvm/llvm_offline_cache.h
+++ b/taichi/llvm/llvm_offline_cache.h
@@ -45,19 +45,19 @@ struct LlvmOfflineCache {
   struct FieldCacheData {
     struct SNodeCacheData {
       int id;
-      SNodeType type;
-      size_t cell_size_bytes;
-      size_t chunk_size;
+      SNodeType type = SNodeType::undefined;
+      size_t cell_size_bytes{0};
+      size_t chunk_size{0};
 
       TI_IO_DEF(id, type, cell_size_bytes, chunk_size);
     };
 
-    int tree_id;
-    int root_id;
-    size_t root_size;
+    int tree_id{0};
+    int root_id{0};
+    size_t root_size{0};
     std::vector<SNodeCacheData> snode_metas;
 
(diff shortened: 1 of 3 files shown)


taichi/llvm/llvm_offline_cache.h
```diff
@@ -42,7 +42,37 @@ struct LlvmOfflineCache {
     TI_IO_DEF(kernel_key, args, offloaded_task_list);
   };
 
-  std::unordered_map<std::string, KernelCacheData> kernels;
+  struct FieldCacheData {
+    struct SNodeCacheData {
+      int id;
```
k-ye, line 47: nit: I'd always initialize these with `{0}` to prevent any surprises

taichi/llvm/llvm_offline_cache.h
```diff
@@ -42,7 +42,37 @@ struct LlvmOfflineCache {
...
+      SNodeType type;
+      size_t cell_size_bytes;
+      size_t chunk_size;
+
+      TI_IO_DEF(id, type, cell_size_bytes, chunk_size);
+    };
+
+    int tree_id;
```
k-ye, line 55: ditto

taichi/llvm/llvm_program.cpp
```diff
@@ -158,9 +158,9 @@ LlvmProgramImpl::clone_struct_compiler_initial_context(
   return tlctx->clone_runtime_module();
 }
 
-void LlvmProgramImpl::initialize_llvm_runtime_snodes(const SNodeTree *tree,
-                                                     StructCompiler *scomp,
-                                                     uint64 *result_buffer) {
+void LlvmProgramImpl::initialize_llvm_runtime_snodes(
+    const LlvmOfflineCache::FieldCacheData &field_cache_data,
```
k-ye, line 162: nit: I think the naming is a bit weird here, in the sense that init LLVM runtime SNodes shouldn't explicitly depend on some offline cache... Maybe we can extract this struct and give it a more generic naming in the future?

jim19930609: Agreed. Let me rename FieldCacheData and make it a standalone struct. 

btw, does "RuntimeFieldData" or "CompiledFieldData" sound good to you? Or any alternative suggestions?

k-ye: Naming is hard... I don't have a good suggestion at this point, either. Maybe we should change this once SNodeMeta thing is more clear?

jim19930609: Sure, so how about keep it as is for now? 

I'm planning to file another issue, talking about synchronizing the implementations for JIT, AOT and OfflineCache. Maybe this naming thing can be covered in that thread?

taichi/llvm/llvm_offline_cache.h
k-ye: not updated?

k-ye (approved):
lgtm

PR #5111, opened 2022-06-08
[aot] [llvm] LLVM AOT Field #1: Adjust serialization/deserialization logics for FieldCacheData

jim19930609:
Related issue = #4800 

It's interesting that the struct_module will get cloned into each kernel_module, so we dont even need to serialize - deserialize the struct_module for fields. However, this relies on a "recursive clone" mechanism which seems to be pretty fragile. 

Added some comments on how these llvm::Modules get arranged, together with some refactor ideas.


k-ye (approved):
Thanks!

Issue #5114, opened 2022-06-09
[Refactor Plan] Synchronize the implementations for JIT, AOT and OfflineCache


jim19930609:
We should form a unified design for JIT, AOT and OfflineCache, which have lots of concepts/implementations in common. 

Taking LLVM backend as an example, one possible design is using offlinecache to split the entire program into compile time and runtime:

![AOT_JIT_Cache drawio (1)](https://user-images.githubusercontent.com/22334008/172757919-a2cd446f-f915-4e79-85e3-9fa58249044b.png)

nit: refactor the recursive clone of llvm::Modules (marked grey in the diagram)

PR #5120, opened 2022-06-09
[aot] [llvm] LLVM AOT Field #2: Updated LLVM AOTModuleLoader & AOTModuleBuilder to support Fields

jim19930609:
Related issue = #4800

Due to the existence of "recursive clone" for llvm::Modules (#5114 ), ti.kernel compilation cannot start until the finalization of all the SNodeTrees. To enforced this implicit dependency, AOT calls `_root_fb._finalize_for_aot()` to finalize the SNodeTrees upon constructing AOT builder object. By the time `AOTBuilder::add_field()` is called, all the SNodeTrees should have already been compiled. 

Therefore `AOTBuilder::add_field()` never actually compiles any SNodeTrees, instead, it simply grab the necessary compiled info from LlvmProgramImpl to construct its FieldCache. 

In this PR, we reused the `LlvmOfflinelineCache cache_data_` from LlvmProgramImpl to store the compiled info generated from `_root_fb._finalize_for_aot()`. Later on, `AOTBuilder::add_field()` simply copies the designated FieldCache from `LlvmProgramImpl::cache_data_`


taichi/aot/module_builder.h
```diff
@@ -81,6 +88,7 @@ class AotModuleBuilder {
 
  private:
   std::unordered_map<std::string, aot::CompiledGraph> graphs_;
+  Program *prog_ = nullptr;
```
ailzhang, line 91: Instead of saving a `Program*` in the base module, I wonder is saving a `LlvmProgramImpl*` inside the backend module builder enough for this use case? (And I assume we want to get rid of it in the future right? 
Saving a pointer to program is like opening a can of worms so I feel like we should limit its scope as much as possible.

ailzhang (commented):
Overall looks good to me!

taichi/aot/module_builder.h
jim19930609: That's really good point, appreciate it!

taichi/llvm/llvm_aot_module_loader.cpp
```diff
@@ -37,5 +55,42 @@ std::unique_ptr<aot::Kernel> LlvmAotModule::make_new_kernel(
...
+  auto ok = cache_reader_->get_field_cache(loaded, snode_tree_id);
+  TI_ERROR_IF(!ok, "Failed to load field with id={}", snode_tree_id);
+
+  return std::make_unique<FieldImpl>(std::move(loaded));
+}
+
+void finalize_aot_field(aot::Module *aot_module,
+                        aot::Field *aot_field,
+                        uint64 *result_buffer) {
+  auto *llvm_aot_module = dynamic_cast<LlvmAotModule *>(aot_module);
```
ailzhang, line 79: static_cast should be fine here?

ailzhang (approved):
LGTM!

taichi/llvm/llvm_aot_module_loader.cpp
jim19930609: `finalize_aot_field()` is actually a public interface, to be called by the users. So if they made a mistake and passed Vulkan::AotModule to this interface, static_cast might result in some undefined behavior (Not even segmentation fault, if the size of XXX::AotModule is smaller than that of LlvmAotModule)

ailzhang: oh if it's public then sgtm!

PR #5121, opened 2022-06-09
[aot] [llvm] LLVM AOT Field #3: Added AOT tests for Fields - CPU backend

jim19930609:
Related issue = #4800


ailzhang (approved):
Nice work!

PR #5122, opened 2022-06-09
[aot] [bug] Use cached compiled kernel pointer when it's added to graph multiple times

ailzhang:
This bug was triggered when we tried to port stable_fluid demo so this
PR also added a cgraph based stable fluid demo.

```
ti example stable_fluid_graph
```

Note it's not ideal to save both `FunctionType compiled_` as well as
`aot::Kernel compiled_aot_kernel_` inside C++ `Kernel` class. But we plan to clean that
up (likely by getting rid of `FunctionType compiled_`) in #5114.

Related issue = #

### DIFF
--- a/taichi/program/kernel.h
+++ b/taichi/program/kernel.h
@@ -87,7 +87,12 @@ class TI_DLL_EXPORT Kernel : public Callable {
 
   void compile();
 
-  std::unique_ptr<aot::Kernel> compile_to_aot_kernel();
+  void compile_to_aot_kernel();
+
+  aot::Kernel *compiled_aot_kernel() {
+    return compiled_aot_kernel_.get();
+  }
+
   /**
    * Lowers |ir| to CHI IR level
    *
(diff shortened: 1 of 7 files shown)


taichi/program/kernel.h
```diff
@@ -142,6 +147,10 @@ class TI_DLL_EXPORT Kernel : public Callable {
   bool ir_is_ast_{false};
   // The closure that, if invoked, lauches the backend kernel (shader)
   FunctionType compiled_{nullptr};
```
k-ye, line 149: Hmm, are we still using this?

ailzhang: @k-ye Yup it's still used in JIT execution  https://github.com/taichi-dev/taichi/blob/master/taichi/program/kernel.cpp#L124 . Will try to get rid of it in a followup PR! (I feel like that PR will be a bit involved so wanted to fix the cgraph bug first ;)

k-ye (approved):
LGTM!

jim19930609 (approved):

PR #5127, opened 2022-06-10
[Example] Fix mass_spring_3d_ggui backend

neozhaoliang:
Using `ti.gpu` as the default arch may cause problems when cuda is not available and `ti.opengl` is chosen.
Change it to `ti.vulkan` as other ggui examples.


FantasyVR:
LGTM!

FantasyVR (approved):

PR #5138, opened 2022-06-12
[Lang] [ir] Refactor indexing expressions in AST & enforce integer indices

re-xyr:
Related issue = (none)

Before:

| expression type | operand | provide shape & stride | mode of access |
|---|---|---|---|
| `GlobalPtrExpression` | field & extarr | no | global |
| `TensorElementExpression` | local tensor & global tensor | yes | global & local |

After:

| expression type | operand | provide shape & stride | mode of access |
|---|---|---|---|
| `IndexExpression` | field & extarr & local tensor | no | global & local |
| `StrideExpression` | global tensor | yes | global |

This unifies all access modes except for global tensors, because they are essentially a hack at this time.


python/taichi/lang/impl.py
```diff
@@ -199,8 +199,13 @@ def subscript(value, *_indices, skip_reordered=False):
 @taichi_scope
 def make_tensor_element_expr(_var, _indices, shape, stride):
```
strongoier, line 200: Rename it to `make_stride_expr()`?

strongoier (approved):
Awesome!!

BTW could you run examples to see if enforcing integer indices will break things? If so fix those examples plz :-)

strongoier (commented):
nit: resolve conflicts & close local formatter to avoid accidental changes :-)

Issue #5109, opened 2022-06-08
[lang] Texture support


bobcao3:
We wish to support hardware texture & image load-store. In the JIT mode (Python), we will provide a new class called `Texture`

You can create a texture, transfer data from a PIL image, numpy array, or NDArray (or other potential in-memory image formats), and the textures can be passed into Taichi kernels for use.

Initially we will support:
[ ] 2D Textures (1~4 channels)
[ ] Image data from NDArray
[ ] LOD0
[ ] Normalized texture formats (internal int, read out float)
[ ] FP16 / FP32 texture formats (internal ieee754 FP)
[ ] Passing textures as kernel arguments, similar to NDArrays
[ ] Bilinear explicit LOD sampling
[ ] Implicit LOD sampling with gradients (`textureGrad`)
[ ] Texel fetching (`texelFetch`)
[ ] Texel quad gathering (`textureGather`)

Future plans:
[ ] 1D / 3D Textures
[ ] PIL / numpy / tensor bidirectional image IO
[ ] LODs & LOD generation helpers
[ ] Sampler parameter control
[ ] Integer textures (Corresponds to `isamplerND` / `usamplerND` in GLSL)
[ ] AOT support
[ ] Image load-store
[ ] Texture array (Descriptor indexing / descriptor array, not `sampler2DArray`)

k-ye:
Note that we are only supporting texture on Vulkan backend as a start.

lyd405121:
## Will compressed format texture supported in the future?
- Like DXT1-7(windows),pvr(powervr),etc(opengl)
- FYI, Khronos group is working on a convertor or [standard](https://github.com/KhronosGroup/KTX-Software) dealing with all kinds of format

PR #5139, opened 2022-06-12
[lang] Texture support 1/n (Context & Programs)

bobcao3:
Related issue = #5109

### DIFF
--- a/taichi/program/context.h
+++ b/taichi/program/context.h
@@ -62,17 +62,15 @@ struct RuntimeContext {
   template <typename T>
   void set_arg(int i, T v) {
     args[i] = taichi_union_cast_with_different_sizes<uint64>(v);
-    set_array_is_device_allocation(i, /*is_device_allocation=*/false);
+    set_array_device_allocation_type(i, DevAllocType::kNone);
   }
 
   void set_array_runtime_size(int i, uint64 size) {
     this->array_runtime_sizes[i] = size;
   }
 
-  void set_array_is_device_allocation(
-      int i,
-      bool is_device_allocation,
-      DevAllocType usage = DevAllocType::kNdarray) {
+  void set_array_device_allocation_type(
+      int i, DevAllocType usage) {
     this->device_allocation_type[i] = usage;
   }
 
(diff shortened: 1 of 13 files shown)


taichi/program/context.h
```diff
@@ -32,6 +34,8 @@ struct RuntimeContext {
   uint64 array_runtime_sizes[taichi_max_num_args_total]{0};
   // `is_device_allocations` is true iff i-th arg is a `DeviceAllocation*`.
   bool is_device_allocations[taichi_max_num_args_total]{false};
```
k-ye, line 36: can we remove `is_device_allocation` now? (Maybe we need to add something like `DevAllocType::kInvalid`.

taichi/program/context.h
```diff
@@ -14,6 +14,8 @@ struct DeviceAllocation;
 // pointer to the LLVMRuntime struct, kernel arguments, and the thread id (if on
 // CPU).
 struct RuntimeContext {
+  enum class DevAllocType : int8_t { ndarray = 0, texture = 1, image = 2 };
```
k-ye, line 17: nit: Let's use `kNdarray` `kTexture` style, see https://google.github.io/styleguide/cppguide.html#Enumerator_Names

taichi/program/program.cpp
```diff
@@ -564,6 +564,22 @@ Ndarray *Program::create_ndarray(const DataType type,
...
+    textures_.emplace_back(this, type, num_channels, shape[0], 1, 1);
+  } else if (shape.size() == 2) {
+    textures_.emplace_back(this, type, num_channels, shape[0], shape[1], 1);
+  } else if (shape.size() == 3) {
+    textures_.emplace_back(this, type, num_channels, shape[0], shape[1],
+                           shape[2]);
+  } else {
+    TI_ERROR("Texture shape invalid");
+  }
+  return &textures_.back();
```
k-ye, line 580: You don't have pointer stability here, consider using `std::vector<std::unique_ptr<Texture>>`

taichi/program/program.cpp
```diff
@@ -564,6 +564,22 @@ Ndarray *Program::create_ndarray(const DataType type,
   return ndarrays_.back().get();
 }
 
+Texture *Program::create_texture(const DataType type,
+                                 int num_channels,
+                                 const std::vector<int> &shape) {
```
k-ye, line 569: Slightly nit: pass in a copy of `shape`, then pad it to length 3 with `1`? This saves a few if-else checks.

taichi/program/program.cpp
bobcao3: ooo right I forgot about that, was wondering why the ndarrays uses a wrapped pointer

taichi/program/program.cpp
bobcao3: This is intended to support 1D/2D/3D images, so there will always be if checks i think...

taichi/program/context.h
```diff
@@ -60,15 +69,24 @@ struct RuntimeContext {
     this->array_runtime_sizes[i] = size;
   }
 
-  void set_array_is_device_allocation(int i, bool is_device_allocation) {
-    this->is_device_allocations[i] = is_device_allocation;
+  void set_array_is_device_allocation(
+      int i,
+      bool is_device_allocation,
```
k-ye, line 74: nit: We can remove this as well :-) (probably in a separate PR)

k-ye (approved):
LGTM!

taichi/program/texture.h
```diff
@@ -0,0 +1,64 @@
...
+namespace lang {
+
+class Program;
+class Ndarray;
+
+class TI_DLL_EXPORT Texture {
+ public:
+  /* Constructs a Ndarray managed by Program.
+   * Memory allocation and deallocation is handled by Program.
+   * TODO: Ideally Ndarray shouldn't worry about memory alloc/dealloc at all.
```
ailzhang, line 20: nit: update the comments to match texture?

ailzhang (approved):
LGTM!

PR #5146, opened 2022-06-13
[autodiff] Add a context manager for forward mode autodiff

erizmr:
* (to be filled)

The primal kernels inside the context manager will be transofrmed to a forward ad kernel.
They will be recovered to primal kernels after exiting the context manager for futher non-ad use.


python/taichi/lang/ad_fwd_mode_manager.py
```diff
@@ -0,0 +1,30 @@
+from taichi.lang.enums import AutodiffMode
+
+
+class ForwardModeManagerImpl:
```
ailzhang, line 4: Ideally this context manager should live in python/taichi/ad.py and it should be the user facing api instead of ti.fwdAD?

python/taichi/lang/misc.py
```diff
@@ -705,6 +707,85 @@ def Tape(loss, clear_gradients=True):
...
+    if clear_gradients:
+        for ls in loss:
+            ls.dual.fill(0)
+
+    # Clear losses
+    if clear_loss:
+        for ls in loss:
+            ls.fill(0)
+
+    return impl.get_runtime().get_fwd_mode_manager(keep_primal)
```
ailzhang, line 774: Hmmm it feels a bit weird to return a context manager here... I'm happy to discuss this - there might be a better way to organize it.

ailzhang (commented):
Nice work!

python/taichi/lang/ad_fwd_mode_manager.py
erizmr: Let me put it in ad.py

python/taichi/lang/kernel_impl.py
```diff
@@ -486,11 +490,22 @@ def materialize(self, key=None, args=None, arg_features=None):
         if key is None:
             key = (self.func, 0)
         self.runtime.materialize()
+
+        # Transform the primal kernel to forward mode grad kernel
+        # then recover to primal when exiting the forward mode manager
+        if self.runtime.fwd_mode_manager:
+            self.autodiff_mode = AutodiffMode.FORWARD
+            self.runtime.fwd_mode_manager.insert(self)
+            self.compiled_functions = self.runtime.compiled_fwd_mode_grad_functions
```
ailzhang, line 499: why do we need to replace compiled_functions?

python/taichi/ad.py
```diff
@@ -117,3 +122,101 @@ def placeholder(*args, **kwargs):
...
+        self.recover_kernels()
+
+    def insert(self, func):
+        self.calls.append(func)
+
+    def recover_kernels(self):
+        assert self.entered, "Before recover the kernels, fwd mode manager must be entered."
+        for f in self.calls:
+            f.autodiff_mode = AutodiffMode.NONE
+            f.compiled_functions = f.runtime.compiled_functions
```
ailzhang, line 209: Hmmmm this is a bit weird, could you explain why `f.runtime.compiled_functions` is what you wanna recover? It doesn't  seem you set it anywhere. :P

ailzhang (commented):
Thanks for the nice work! 
I have a few questions regarding what need to be recovered when entering/exiting the context manager. 
It would nice to add a few tests for this as well as it's error prone indeed.

python/taichi/lang/kernel_impl.py
erizmr: The reason is to distinguish the primal and forward grad kernel.
The two kernels has the same name and args, thus the func key is the same, i.e., one will overwrite another in a map.

python/taichi/ad.py
erizmr: The `compiled_functions` is set when a kernel is initialized. For primal kernels, its `compiled_functions` is set to `runtime.compiled_functions` by default.

ailzhang: Hmmm suppose different kernels have different `compiled_functions` when entering this context manager, I'd assume when they now all have the same `compiled_functions` when they leave?

erizmr: It is assumed that only the primal kernels will enter the context manager so we set them `compiled_functions` when they leave. One generally way is that recording both the kernel called and its `compiled_functions` and give it back at the exit.

ailzhang (approved):
Thanks!

Issue #5143, opened 2022-06-13
FieldsBuilder raise runtime error on cc backend


erizmr:
**Describe the bug**
SNode writer raise runtime error with FieldsBuilder on cc backend.

**To Reproduce**

```py
import taichi as ti
ti.init(arch=ti.cc, log_level=ti.TRACE)

dual_root = ti.FieldsBuilder()
x_dual = ti.field(float)
dual_root.dense(ti.i, 5).place(x_dual)
dual_root.finalize()
```

**Log/Screenshots**
![image](https://user-images.githubusercontent.com/33411325/173279096-8afce34d-6c5c-4f3a-ad88-539714fb8628.png)
![image](https://user-images.githubusercontent.com/33411325/173279291-38cfbc63-7c54-46a6-9640-3ccc9a5917ca.png)

**Additional comments**
The script works fine on cpu, cuda, vulkan, opengl.

k-ye:
Is there any specific reason that  you want to use the CC backend..?

erizmr:
> Is there any specific reason that you want to use the CC backend..?

There are test cases based on the cc backends.. ioi

k-ye:
OK... If no-one is using it, we should just drop it later. It's not worth the maintenance cost.

PR #5147, opened 2022-06-13
[autodiff] Support basic operations for forward mode autodiff

erizmr:
* __->__ #5147
* #5146

Support cpu and gpu backends. The cc backend has an issue on FieldBuilder ref to #5143.
The opengl backend currently does not support materializing multiple snode trees (see [OpenglProgramImpl::compile_snode_tree_types](https://github.com/taichi-dev/taichi/blob/master/taichi/backends/opengl/opengl_program.cpp#L45)), thus `FieldBuilder` is not supported.


erizmr:
unlinked ghstack

PR #5150, opened 2022-06-13
[AOT] C-APIs for Taichi runtime distribution

PENGUINLIONG:
Because platforms like Android and iOS have their own workflows of library integration, and some users do not have a C++17 toolchain to compile Taichi properly, Taichi need an interface to be correctly linked and/or dynamically loaded by any toolchain, any programming language on any platform. So this PR proposes a set of C-APIs to ease distributed library integration.

This PR introduces the following changes:
- `c_api/taichi.json` which is a machine-readable specification of the first release of Taichi Runtime public APIs;
- `misc/generate-c-api.py` is a Python script to generate C/C++ headers based on the specification;
- `c_api/include/taichi/*.h` are the generated headers.


c_api/include/taichi/taichi_core.h
```diff
@@ -0,0 +1,171 @@
...
+  bool host_write;
+  bool host_read;
+  bool export_sharing;
+  TiMemoryUsageFlagBits usage;
+} TiMemoryAllocateInfo;
+
+// structure.nd_shape
+typedef struct TiNdShape {
+  uint32_t dim_count;
+  uint32_t dims;
```
ailzhang, line 84: Is this a single `uint32_t` or an array?

ailzhang (approved):
LGTM!!

c_api/include/taichi/taichi_core.h
PENGUINLIONG: nice catch 😨

PR #5153, opened 2022-06-13
[cli] Improve user interface for CLI command ti example

neozhaoliang:
I think this should help users better see the full list of examples

Screenshot:
2022-06-14 10-43-11屏幕截图

### DIFF
--- a/python/taichi/_main.py
+++ b/python/taichi/_main.py
@@ -14,6 +14,7 @@ from pathlib import Path
 import numpy as np
 import rich
 from colorama import Fore
+from rich.console import Console
 from taichi._lib import core as _ti_core
 from taichi._lib import utils
 from taichi.tools import cc_compose, diagnose, video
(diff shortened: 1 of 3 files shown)


python/taichi/_main.py
```diff
@@ -230,15 +224,40 @@ def on_mouse_click_callback(example_name):
     @register
     def example(self, arguments: list = sys.argv[2:]):
         """Run an example by name (or name.py)"""
+        def colormap(name):
+            from colorsys import hls_to_rgb  # pylint: disable=C0415
```
ailzhang, line 228: does this introduce a new dependency for users?

neozhaoliang: `colorsys` is a python built-in module, hence won't introduce new dependency here~

ailzhang (approved):
LGTM! 
(In general we should be super cautious about introducing new dependencies but approving since `rich` is fairly small)

PR #5160, opened 2022-06-14
[autodiff] Support basic operations for forward mode autodiff

erizmr:
* __->__ #5160
* #5146

Support cpu and gpu backends.
The cc backend has an issue on FieldBuilder ref to #5143.
The opengl backend currently does not support materializing multiple snode trees (see OpenglProgramImpl::compile_snode_tree_types), thus FieldBuilder is not supported.

Related #5055


ailzhang (approved):
As discussed offline, we'll merge this prototype first and there're a few followups:
- Add a test for fwdmode context manager
- Try getting rid of different compiled_functions inside Kernel class. 
- Enforce checking for `ti.ad` in `test_api.py`

PR #5161, opened 2022-06-14
[cuda] Disable reduction in non-full warps

### LINKED ISSUE #5142: strange behavior using atomic operation with snode on cuda
**Describe the bug**
Using atomic_min when looping through snode gives wrong result on **cuda**.

**To Reproduce**
```py
import taichi as ti

ti.init(arch=ti.gpu)

x = ti.field(ti.f32)
block = ti.root.pointer(ti.ij, (1000,1))
pixel = block.dense(ti.ij, (1,1))
pixel.place(x)

@ti.kernel
def test():
    hit_time = 1000.0
    x[100, 0] = 1.0
    x[200, 0] = 2.0
    for i,j in x:
        # print(hit_time, i, j, x[i,j])
        # hit_time += 0.0
        ti.atomic_min(hit_time, x[i,j])
    print(hit_time)

test()
```

**Log/Screenshots**
The above program should print 1.0. However, it will print 0.0 if we don't uncomment one of the commented lines. This seems to only happen when arch is cuda.

### LINKED ISSUE #5148: Atmoic min get wrong result on cuda backend when block_dim is 16
**Describe the bug**
Atmoic min get wrong result on cuda backend when block_dim is 16.

**To Reproduce**
Run the code below. The expected answer is 1, but I got 0 on my computer.

```py
import taichi as ti

ti.init(arch=ti.cuda)

n = 16
x = ti.field(dtype=ti.f32, shape=n)

@ti.kernel
def test():
    hit_time = 1
    for i in x:
        ti.atomic_min(hit_time, 1)
    print(hit_time)

test()
```
Setting n to any number larger than 16, print value of `hit_time`, or replacing `for i in x` with `for i in range(n)` will get the right answer.
Especially, replacing `for i in x` with `for i in range(n)` will only change the block_dim in the final Taichi IR, and thus I think the bug might be at cuda codegen phase.

bobcao3:
Fixes #5148
Fixes #5142


ailzhang (approved):
nit: can we add the repro example in #5148 as a test to prevent future regression?

PR #5163, opened 2022-06-14
[Doc] Updated odop.md, removing obsolete information

Vissidarte-Herman:
Related issue = #

### DIFF
--- a/docs/lang/articles/advanced/odop.md
+++ b/docs/lang/articles/advanced/odop.md
@@ -43,7 +43,7 @@ a = TiArray(32)
 a.inc()
 ```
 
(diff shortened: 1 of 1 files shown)


docs/lang/articles/advanced/odop.md
```diff
@@ -43,7 +43,7 @@ a = TiArray(32)
 a.inc()
 ```
 
-Programmers used to define Taichi fields in `__init__` functions of `@ti.data_oriented` classes. With the new **Dynamic SNode** feature (released in `v0.8.0`, see [Field (advanced)](layout.md#dynamic-field-allocation-and-destruction) for more details), you can define Taichi fields **at any places** of Python-scope functions. For example,
+Definitions of Taich fields can be made not only in _init_ functions, but also at any place of a Python-scope function in a data-oriented class. For example,
```
strongoier, line 46:
```suggestion
Definitions of Taichi fields can be made not only in `__init__` functions, but also at any place of a Python-scope function in a data-oriented class. For example,
```

strongoier (approved):
Thanks!

Issue #5090, opened 2022-06-02
Throw error for unsupported feature in manylinux wheel


ailzhang:
#5056 
Our manylinux wheel is only built with cpu / cuda backend so for instance, if you want to use GGUI it'll just complain it's not available. We should detect if the installed package is from a manylinux wheel and throw proper error message for it.

PR #5164, opened 2022-06-14
[gui] Detailed error message when GGUI is not available

feisuzhu:
Related issue = #5090

### DIFF
--- a/python/taichi/ui/utils.py
+++ b/python/taichi/ui/utils.py
@@ -1,3 +1,5 @@
+import platform
+import re
 from math import acos, asin, cos, sin
 
 from taichi._lib import core as _ti_core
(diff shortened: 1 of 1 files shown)


python/taichi/ui/utils.py
```diff
@@ -64,8 +67,74 @@ def vec_to_euler(v):
...
+
+            raise GGUINotAvailableException(
+                "GGUI is not available since you have installed a restricted version of taichi."
+            )
+
+    except GGUINotAvailableException:
+        raise
+
+    except Exception:
+        pass
```
ailzhang, line 138: OOC: when do we expect to hit this? ;)

python/taichi/ui/utils.py
```diff
@@ -1,9 +1,12 @@
+import platform
 from math import acos, asin, cos, sin
 
 from taichi._lib import core as _ti_core
 from taichi.lang.impl import default_cfg
 from taichi.lang.matrix import Vector
 
+import taichi
```
ailzhang, line 8: OOC: is this import necessary? ;)

ailzhang (commented):
Nice work!

python/taichi/ui/utils.py
feisuzhu: When the code is buggy :)
This is a defending mechanism, if the code fails, just fallback to the most general situation, or user will see some confusing errors.

python/taichi/ui/utils.py
feisuzhu: Will move it closer to where it's used.

python/taichi/ui/utils.py
ailzhang: Aha sg!

python/taichi/ui/utils.py
```diff
@@ -64,8 +65,75 @@ def vec_to_euler(v):
...
     """Checks if the `GGUI` environment is available.
     """
-    if not _ti_core.GGUI_AVAILABLE:
-        raise Exception("GGUI Not Available")
+    if _ti_core.GGUI_AVAILABLE:
+        return
+
+    try:
+        # Try identifing the reason
+        import taichi
```
ailzhang, line 106: Btw this might not pass the pylint check: https://docs.taichi-lang.org/docs/contributor_guide#enforce-code-style

qiao-bo:
PR title check fails at unrecognized tag, seems like we don't have `ui` tag. Change to `gui` and rebase to trigger the build. @feisuzhu btw, don't forget the CLA ;/

feisuzhu:
> PR title check fails at unrecognized tag, seems like we don't have `ui` tag. Change to `gui` and rebase to trigger the build. @feisuzhu btw, don't forget the CLA ;/

done

ailzhang (approved):
Thanks!

PR #5165, opened 2022-06-14
[AOT] Support Matrix/Vector as graph arguments

turbo0628:
Related issue = #


python/taichi/graph/_graph.py
```diff
@@ -50,23 +51,40 @@ def __init__(self, compiled_graph) -> None:
...
+            elif isinstance(v, Matrix):
+                mat_val_id = 0
+                for a in range(v.n):
+                    for b in range(v.m):
+                        key = f"{k}_mat_{mat_val_id}"
+                        mat_val_id += 1
+                        if isinstance(v[a, b], int):
+                            arg_ints[key] = int(v[a, b])
+                        elif isinstance(v[a, b], float):
+                            arg_floats[key] = float(v[a, b])
```
k-ye, line 78: nit: Why is one going to `arg_doubles` and the other going to `arg_floats`..?

taichi/aot/graph_data.cpp
```diff
@@ -16,27 +17,47 @@ void CompiledGraph::run(
...
     for (int i = 0; i < symbolic_args_.size(); ++i) {
       auto &symbolic_arg = symbolic_args_[i];
       auto found = args.find(symbolic_arg.name);
-      TI_ERROR_IF(found == args.end(), "Missing runtime value for {}",
-                  symbolic_arg.name);
+      size_t mat_arg_count = 0;
+      if (found == args.end()) {
+        // Try find matrix args
+        // TODO replace this logic when C++ matrix is available.
+        std::regex mat_arg_regex(symbolic_arg.name + "_mat_.[0-9]*");
```
k-ye, line 28: This assumes that the symbolic name for matrix args is always of this pattern, which I don't think is the case? I think what we can do here is to store some type inofrmation in `symbolic_args`, thoughts?

python/taichi/graph/_graph.py
turbo0628: The original implementation would cast to `double`. However my test case for Matrix is ti.f32 that should be cast to `float`, so use this branch for new matrix args and avoid break other args. The type needs further refactor, filed an issue on this #5165 .

taichi/aot/graph_data.cpp
turbo0628: `Matrix` args are pure Python implementations, so it's a bit of pain to get type info here.  A `Matrix` arg is always flattened into multiple scalar args with `{user_key}_mat_{id}` suffix. For example, a 3x3  Matrix with "AMat" key will be expanded to 9 scalars with "AMat_mat_[0-8]" keys. 

The flatten behavior is also consistent with the Matrix arg set code in [kernel_impl](https://github.com/taichi-dev/taichi/blob/633c628b5f40f53438863329ed80b8f19ace4a47/python/taichi/lang/kernel_impl.py). So it's easier to refactor when the native Matrix is available.

python/taichi/graph/_graph.py
```diff
@@ -50,23 +51,40 @@ def __init__(self, compiled_graph) -> None:
...
             elif isinstance(v, int):
                 arg_ints[k] = v
             elif isinstance(v, float):
-                arg_floats[k] = v
+                arg_doubles[k] = v
+            elif isinstance(v, Matrix):
+                mat_val_id = 0
+                for a in range(v.n):
+                    for b in range(v.m):
+                        key = f"{k}_mat_{mat_val_id}"
```
turbo0628, line 73: The flatten code @k-ye

tests/python/test_graph.py
```diff
@@ -21,3 +21,84 @@ def test(pos: ti.types.ndarray(field_dim=1)):
...
+    def vector_sum(mat: ti.types.vector(N, dtype),
+                   res: ti.types.ndarray(field_dim=1)):
+        res[0] = mat.sum() + mat[2]
+
+    @ti.kernel
+    def matrix_sum(mat: ti.types.matrix(N, 2, dtype),
+                   res: ti.types.ndarray(field_dim=1)):
+        res[0] = mat.sum()
+
+    sym_A = ti.graph.Arg(ti.graph.ArgKind.MATRIX, 'mat', dtype)
```
ailzhang, line 56: @turbo0628 Matrix argkind can have element shape, I think we can do this flattening all in  python without hacking in C++?

turbo0628: The flattening is all in Python, the MATRIX is actually a scalar. Should I remove this?

taichi/python/export_lang.cpp
```diff
@@ -563,6 +563,9 @@ void export_lang(py::module &m) {
   py::enum_<aot::ArgKind>(m, "ArgKind")
       .value("SCALAR", aot::ArgKind::kScalar)
       .value("NDARRAY", aot::ArgKind::kNdarray)
+      // Using this MATRIX as Scalar alias, we can move to native matrix type
+      // when supported
+      .value("MATRIX", aot::ArgKind::kScalar)
```
turbo0628, line 568: @ailzhang see here

tests/python/test_graph.py
turbo0628: I just think it a bit weird to expose `ArgKind.SCALAR` for Matrix

ailzhang: @turbo0628 There's a slight difference, I'd expect here to be `ti.graph.Arg(ti.graph.ArgKind.Matrix, 'mat', ti.types,matrix(N, 2, dtype))`? and it's flattened in python, so what C++ get is actually `N*2` scalar args.

taichi/aot/graph_data.cpp
k-ye: Hmmm I see. This is quite ad-hoc IMHO, but I get your point. nit: the regex should be `_mat_[0-9]+`?

turbo0628: Exactly...This hack is too ad-hoc, @ailzhang suggest that we can assign the key with suffix earlier in the dispatch stage. The regex should not appear here then. I'll make this modification later today!

tests/python/test_graph.py
turbo0628: Make sense, will make the changes later today

python/taichi/aot/utils.py
```diff
@@ -66,6 +66,8 @@ def produce_injected_args(kernel, symbolic_args=None):
                                   layout=Layout.AOS))
             else:
                 raise RuntimeError('')
+        elif isinstance(anno, MatrixType):
+            injected_args.append(anno)
```
ailzhang, line 70: This is not quite right - when you compile you actually should feed in a dummy `Matrix` instance instead of `MatrixType` instance. (That's also why you need special handling for MatrixType args below...

I believe the right thing to do here is creating a Matrix of `anno` type and put each scalar in the inject_args one by one (once we have matrix type we can just inject one matrix arg but now it'll have to be  N scalars).

ailzhang (commented):
Thanks! left a comment inline where we still need some adjustment.

python/taichi/aot/utils.py
```diff
@@ -66,6 +67,20 @@ def produce_injected_args(kernel, symbolic_args=None):
                                   layout=Layout.AOS))
             else:
                 raise RuntimeError('')
+        elif isinstance(anno, MatrixType):
+            if not isinstance(symbolic_args[i], list):
```
ailzhang, line 71: nit: we can also assert on symbolic_args[i]'s argKind is Matrix here ;)

ailzhang (approved):
Nice work thanks!!

python/taichi/aot/utils.py
turbo0628: The symbolic args are already expanded here, so there are only Args and (2-D) list of Args here

python/taichi/aot/utils.py
```diff
@@ -66,6 +67,20 @@ def produce_injected_args(kernel, symbolic_args=None):
...
+
+            symbolic_mat_m = len(symbolic_args[i])
+            symbolic_mat_n = len(symbolic_args[i][0])
+
+            if not symbolic_mat_m == anno.m and symbolic_mat_n == anno.n:
+                raise RuntimeError(
+                    f'Matrix dimension mismatch, expected ({anno.m}, {anno.n}) '
+                    f'but dispathed shape ({symbolic_mat_m}, {symbolic_mat_n}).'
+                )
+            dummy_arr = np.zeros((anno.m, anno.n))
```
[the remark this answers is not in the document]
turbo0628: Good point!

taichi/python/export_lang.cpp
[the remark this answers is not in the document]
turbo0628: Thanks, just want to hint that matrix type is still absent in native codebase. I'll add the kMatrix type :)

PR #5174, opened 2022-06-14
[lang] Texture support 3/n (Python changes)

bobcao3:
Related issue = #5109

After this PR the initial support will be fully complete. Implementing `textureGrad` and `textureGather` should be straightforward small changes.

After the all the ones outlined in initial support is added with a test case, we should be able to release the API out to the wild.


python/taichi/examples/rendering/simple_texture.py
```diff
@@ -0,0 +1,50 @@
...
+        ret = ti.cast(ret * 255, ti.u8)
+        arr[i, j] = ret
+
+
+make_texture(tex_ndarray)
+tex.from_ndarray(tex_ndarray)
+
+
+@ti.kernel
+def paint(t: ti.f32, tex: ti.types.texture):
```
k-ye, line 28: nit: Should this be `ti.types.texture()` (i.e. a function call)? Because we might need to add arguments to `texture` in the future, e.g. `format`, `dim`, ...

k-ye (approved):
LGTM!

Issue #5170, opened 2022-06-14
Enforce checking for ti.ad in test_api.py


erizmr:
There are two APIs under the `ti.ad` module, i.e., `ti.ad.Tape` and `ti.ad.FwdMode`, which are not covered by the test cases.

erizmr:
Sorted by #5175

PR #5175, opened 2022-06-15
[refactor] [autodiff] Refactor autodiff api and add corresponding tests

erizmr:
Related issue = #5170 
- Make `ad` a submodule instead of a single python file. 
- Deprecate `ti.clear_all_gradients`, use `ti.ad.clear_all_gradients` instead
- Add api test for `ad` module


ailzhang (approved):
Thanks!!

Issue #5053, opened 2022-05-26
Run C++ tests in Windows CI 


ailzhang:
Just realized that we didn't build C++ tests in windows, not sure if it's intended. https://github.com/taichi-dev/taichi/blob/master/.github/workflows/testing.yml#L417 

Also there seems to be a bug in print in windows CPU docker job, 
```
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\_distutils\core.py", line 163, in run_commands
    dist.run_commands()
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\_distutils\dist.py", line 967, in run_commands
    self.run_command(cmd)
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\dist.py", line 1214, in run_command
    super().run_command(command)
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\_distutils\dist.py", line 986, in run_command
    cmd_obj.run()
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\command\install.py", line 74, in run
    self.do_egg_install()
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\command\install.py", line 123, in do_egg_install
    self.run_command('bdist_egg')
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\_distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\dist.py", line 1214, in run_command
    super().run_command(command)
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\_distutils\dist.py", line 986, in run_command
    cmd_obj.run()
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\command\bdist_egg.py", line 165, in run
    cmd = self.call_command('install_lib', warn_dir=0)
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\command\bdist_egg.py", line 151, in call_command
    self.run_command(cmdname)
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\_distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\tools\miniconda3\envs\py37\lib\site-packages\setuptools\dist.py", line 1214, in run_command
    super().run_command(command)
[...]

PR #5176, opened 2022-06-15
[build] Run C++ tests on Windows

feisuzhu:
Related issue = #5053


.github/workflows/scripts/win_test.ps1
```diff
@@ -13,6 +15,10 @@ if ("$env:TI_WANTED_ARCHS".Contains("cuda")) {
     pip install "torch; python_version < '3.10'"
     pip install "paddlepaddle==2.3.0; python_version < '3.10'"
 }
+
+# Run C++ tests
+python tests/run_tests.py --cpp
```
qiao-bo, line 20: without `if (-not $?) { exit 1 }` the ps script seems not terminate upon errors on our CI.
```suggestion
python tests/run_tests.py --cpp
if (-not $?) { exit 1 }
```

tests/run_tests.py
```diff
@@ -112,7 +119,7 @@ def _test_python(args):
 
     try:
         from multiprocessing import cpu_count  # pylint: disable=C0415
-        threads = min(8, cpu_count())  # To prevent running out of memory
+        threads = max(8, cpu_count())  # To prevent running out of memory
```
qiao-bo, line 122: NITS, now we explicitly control the number of threads to launch the test with `-t` in the win_test script. So `threads = args.threads or env_threads or threads` will never evaluated to this part. So change here will not increase the test parallelism.

feisuzhu: Ah, didn't notice it slipped here, will revert it.

.github/workflows/scripts/win_test.ps1
feisuzhu: fixed

qiao-bo (approved):
Thanks

PR #5186, opened 2022-06-16
[ci] Save built wheel when test fails

feisuzhu:


.github/workflows/testing.yml
```diff
@@ -171,6 +172,13 @@ jobs:
         run: |
           docker rm taichi_build taichi_test -f
 
+      - name: Save wheel if test failed
+        if: failure() && steps.test.conclusion == 'failure'
```
qiao-bo, line 176: Seems the github syntax says `${{ failure() && steps.demo.conclusion == 'failure' }}` ? https://docs.github.com/en/actions/learn-github-actions/expressions#example-10
are these the same? ;/

feisuzhu: It says

> When you use expressions in an if conditional, you may omit the expression syntax 

in the same page :)

qiao-bo: cool, didn't know that

qiao-bo (approved):
LGTM!

.github/workflows/testing.yml
frostming: What is the difference between `failure()` and `steps.demo.conclusion == 'failure'`?

qiao-bo: > What is the difference between `failure()` and `steps.demo.conclusion == 'failure'`? Or does one imply the other?

as far as i understand from the doc (link above), the first part is true as long as any previous steps fail, while the second part specifies a condition for that failure, namely `steps.test.conclusion` the test (id) part should fail, not the build step for example.

frostming: Fine, as the example uses that expression.

PR #5187, opened 2022-06-16
[llvm] [refactor] LLVMProgramImpl code clean up: part-2

jim19930609:
Related issue = #4800, #5114

Removed `Program::get_llvm_program_impl()` interface, so we dont have to expose backend-specific LlvmProgramImpl along with Program.

### DIFF
--- a/.github/workflows/scripts/win_test.ps1
+++ b/.github/workflows/scripts/win_test.ps1
@@ -1,5 +1,7 @@
 $ErrorActionPreference = "Stop"
 
+$env:PYTHONUNBUFFERED = 1
+
 . venv\Scripts\activate.ps1
 python -c "import taichi"
 ti diagnose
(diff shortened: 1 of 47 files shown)


ailzhang (commented):
Hmmm I vaguely remember this function - the main motivation of `get_llvm_program_impl`was a "temp" hack and we wanted to remove it in the future (as opposed to exposing a more backend agnostic version). Please correct me if I'm wrong! @k-ye

jim19930609:
> Hmmm I vaguely remember this function - the main motivation of `get_llvm_program_impl`was a "temp" hack and we wanted to remove it in the future (as opposed to exposing a more backend agnostic version). Please correct me if I'm wrong! @k-ye

I guess we'll have to get rid of `get_llvm_program_impl()` since it exposes LlvmProgramImpl in the header file, which couples the entire LLVM backend with the higher-level Base classes (Any file that include "program.h"). 

IMO, removing the backend-agnostic interface `get_program_impl()` could be a second step, which helps keep the higher-level interface clean. This is difficult because we put too many members and implementations in the "Program":

2022-06-17 10-03-53 的屏幕截图

Ideally, we should move most of these members to "ProgramImpl", and make "Program" an interface class. In that way, lower-level objects such as CodeGenLLVM can keep "LlvmProgramImpl" instead of "Program".

k-ye:
I agree that we should remove get_llvm_program_impl() first. One nit: inside codegen, instead of writing `static_cast<LlvmProgramImpl*>(program->get_program_impl())`, we could write a small helper here?

ailzhang (approved):
LGTM!

PR #5188, opened 2022-06-16
[llvm] [refactor] LLVMProgramImpl code clean up: part-3

jim19930609:
Related issue = #4800, #5114

Let's avoid accessing LlvmProgramImpl with "dynamic_cast + macro" in high-level object such as `program.cpp` or `kernel.cpp`, which strongly burdens the macro management. 

For the place where ProgramImpl gets dispatched to each backend-specific XXXProgramImpl, it will stay unchanged for now. Will file another issue to discuss how we can replace that with "template + linking"


ailzhang (approved):
LGTM!

PR #5191, opened 2022-06-17
[lang] Improve error message when literal val is out of range of default dtype

ailzhang:
I tried hacking around to be smart about the literal dtypes depending on
backends or passed in python/numpy value. But those cause various
inconsistencies as well so I decided to improve the error message
first...

Related issue = #

### DIFF
--- a/python/taichi/lang/expr.py
+++ b/python/taichi/lang/expr.py
@@ -3,7 +3,7 @@ from taichi._lib import core as _ti_core
 from taichi.lang import impl
 from taichi.lang.common_ops import TaichiOperations
 from taichi.lang.exception import TaichiTypeError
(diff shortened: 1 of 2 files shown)


strongoier (commented):
I think there are only two cases when `_check_in_range(to_numpy_type(constant_dtype), val) == False`.
- If `dtype is None`, then `default_ip` is used. We need to ask users to explicitly specify the dtype, or set another `default_ip`.
- Otherwise, the literal just falls out of the specified dtype. The old error message is sufficient.

ailzhang:
@strongoier  aha good point - those messy checks were legacy from my attempt to remove  _clamp_unsigned_to_signed (then I realized it's kinda necessary :P 
Also added a few test cases for the newly added check.

strongoier (approved):
LGTM!

PR #5207, opened 2022-06-20
[llvm] [aot] Split LlvmRuntimeExecutor from LlvmProgramImpl

jim19930609:
Related issue = #4800, #5114

There are two major customer-level classes, namely Kernel and FieldsBuilder. Both Kernel and FieldsBuilder rely on Program/ProgramImpl to access compile-time and runtime interfaces.

However, AOT runtime requires a standalone runtime interface decoupled from the compile-time interfaces. To satisfy AOT needs, this PR splits LlvmRuntimeExecutor out from LlvmProgramImpl:
1. LlvmRuntimeExecutor: contains only runtime objects and interfaces
2. LlvmProgramImpl: contains compile-time interfaces and some wrapper functions to access runtime interfaces (for use in JIT)

*Note: Compile - runtime split for TaichiLLVMContext is reserved for successive PRs.

image


taichi/llvm/llvm_runtime.h
```diff
@@ -0,0 +1,155 @@
+#pragma once
```
k-ye, line 1: file nit: better call it `llvm_runtime_executor.h` (in another PR)?

taichi/llvm/llvm_runtime.h
```diff
@@ -0,0 +1,155 @@
...
+}  // namespace cpu
+
+class LlvmRuntimeExecutor {
+ public:
+  LlvmRuntimeExecutor(CompileConfig &config, KernelProfilerBase *profiler);
+
+  /* ------------------------ */
+  /* ---- Initialization ---- */
+  /* ------------------------ */
+  void initialize_llvm_runtime_snodes(
```
k-ye, line 39: I'm still a bit concerned that we are exposing many public interfaces, a lot of which may not be relevant to AOT users. Is it possible to friend this class with `LlvmProgramImpl`?

jim19930609: Yeah, that sounds way better than exposing a lot of JIT-specific public interfaces. 

Will rearrange these public/private interfaces, thanks!

k-ye (approved):
LGTM!

taichi/llvm/llvm_runtime.h
jim19930609: Let's start with: 
1. materialize_runtime()
2. initialize_llvm_runtime_snodes()
3. allocate_memory_ndarray()
4. check_runtime_error() // for AOT tests

Will expose more once necessary

taichi/llvm/llvm_runtime.h
jim19930609: Done~

PR #5212, opened 2022-06-20
[spirv] Fixed `OpLoad` with physical address

PENGUINLIONG:
Compiline SPIR-V with target Vulkan version >1.0 triggers the following error:

```
SPIR-V module not valid: Memory accesses with PhysicalStorageBufferEXT must use Aligned.
```


k-ye (approved):
LGTM!

Issue #5133, opened 2022-06-10
Enable Werror for TI_WITH_CUDA_TOOLKIT=ON


ailzhang:
Performance benchmark job now uses gcc instead of clang-10 as host compiler. One blocker of moving to clang-10 is that we need to enable Werror when TI_WITH_CUDA_TOOLKIT is on. Once it's fixed we should use clang-10 for perf benchmark job as well.

qiao-bo:
this is possible to fix in code. IIUC, the cupti source code is copied into our repo and we can just eliminate warning in the source.

feisuzhu:
issue irrelevant now.

PR #5216, opened 2022-06-21
[build] Fix -Werror errors for TI_WITH_CUDA_TOOLKIT=ON (#5133)

feisuzhu:
Related issue = #5133


qiao-bo (changes requested):
There is a standard way in Cmake to `FindCUDAToolkit` (https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html). Can we make use of that?

taichi/backends/cuda/cupti_toolkit.cpp
```diff
@@ -1045,7 +1045,7 @@ bool CuptiToolkit::update_record(
       range_name += description_ptrs[description_index];
     }
 
-    const bool isolated = true;
+    // const bool isolated = true;
```
qiao-bo, line 1048:
```suggestion
```

taichi/backends/cuda/cupti_toolkit.cpp
```diff
@@ -145,7 +145,7 @@ ScopeExit<T> MoveScopeExit(T t) {
 
 #define SCOPE_EXIT(func)                                      \
   const auto NV_ANONYMOUS_VARIABLE_INDIRECT(EXIT, __LINE__) = \
-      MoveScopeExit([=]() { func; })
+      MoveScopeExit([=]() { (void)(func); })
```
qiao-bo, line 148: thanks, I wonder does this fix alone eliminate the warnings?

feisuzhu: Yes, after the fix all warnings are disappeared.

qiao-bo (approved):
LGTM! Thanks

PR #5219, opened 2022-06-21
[misc] Add filename option to ti.tools.VideoManager.

houkensjtu:
Related issue = #

Currently, the `VideoManager` uses the default filename `video.mp4` when `make_video()` is called. This PR adds an optional filename argument to the `VideoManager` constructor.
1. When no filename is specified, the default output filename is still `video.mp4`.
2. The code will disregard user's illegal file extension input.


python/taichi/tools/video.py
```diff
@@ -103,7 +106,15 @@ def __init__(self,
         self.automatic_build = automatic_build
 
     def get_output_filename(self, suffix):
-        return os.path.join(self.directory, 'video' + suffix)
+        if not self.video_filename:
+            return os.path.join(self.directory, 'video' + suffix)
+        else:
```
strongoier, line 111: Remove the redundant `else` to make pylint happy :-)

strongoier (approved):
LGTM!

Have you signed the CLA? If not plz do so..

python/taichi/tools/video.py
houkensjtu: Thx for the instruction! Hope pylint is happy now with the follow-up commit!

PR #5223, opened 2022-06-21
[type] [refactor] Decouple quant from SNode 2/n: Remove physical_type from QuantIntType

strongoier:
Related issue = #4857

Having a `physical_type` in `QuantIntType` is simply wrong - a `QuantIntType` may be placed under different `bit_struct`s. The `physical_type` needs to be manually passed into functions.

### DIFF
--- a/taichi/backends/cuda/codegen_cuda.cpp
+++ b/taichi/backends/cuda/codegen_cuda.cpp
@@ -539,11 +539,11 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
           // Bit pointer case.
           auto val_type = ptr_type->get_pointee_type();
           if (auto qit = val_type->cast<QuantIntType>()) {
-            dtype = qit->get_physical_type();
+            dtype = get_ch->input_snode->physical_type;
             auto [data_ptr, bit_offset] = load_bit_pointer(llvm_val[stmt->src]);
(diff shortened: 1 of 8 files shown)


strongoier:
> Out of curiosity, does it make sense if we have a `MaterializedQuantInt` (or `PlacedQuantInt`) or something as an abstraction in the future? It basicallys contains a `QuantIntType` + a physical storage type.

Yep. In fact my original thought is to create a `BitStructMemberType` which contains a `BitStructType` + a `member_id` for those placed types. However, I haven't figured out a perfect way to make it also serve as a `QuantIntType` yet. Do you have any suggestions here? Is using inheritance feasible (it seems that we don't use inheritance in Taichi types though..)?

strongoier:
> I can't say that I have enough context here LOL... My current understanding is that, `QuantIntType` (QIT) is a computational type without the knowledge of the storage type. So by combining QIT with a storage type, you have fully materialized info. (I'm not sure if it's still appropriate to call this materialized info a "type")
> 
> But anyhow, I think the cleanup in the current PR looks good!

Your understanding is correct. The only issue is that the storage type is a bit complicated, especially when it comes to floating points. I'll tweak the implementation a bit more to see what is needed in the end.

BTW give me a LGTM plz :-)

k-ye (approved):
LGTM!

PR #5224, opened 2022-06-22
[autodiff] Forward mode generating computation graph adaptively for parameters to compute derivatives

erizmr:
* (to be filled)

### DIFF
--- a/tests/python/test_ad_basics_fwd.py
+++ b/tests/python/test_ad_basics_fwd.py
@@ -15,7 +15,7 @@ def test_ad_fwd_add():
     def ad_fwd_add():
         loss[1] += 2 * x[3]
 
-    with ti.ad.FwdMode(loss=[loss], parameters=x, seed=[0, 0, 0, 1, 0]):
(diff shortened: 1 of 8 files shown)


tests/python/test_ad_basics_fwd.py
```diff
@@ -34,7 +36,46 @@ def test_ad_fwd_multiply():
...
 
-    with ti.ad.FwdMode(loss=[loss], parameters=x, seed=[0, 0, 0, 1, 1]):
+    with ti.ad.FwdMode(loss=loss, parameters=x, seed=[0, 0, 0, 1, 1]):
         ad_fwd_multiply()
 
-    assert loss.grad[1] == 7
+    assert loss.dual[1] == 7
+
+
+@test_utils.test(arch=[ti.cpu, ti.gpu])
```
ailzhang, line 45: OOC: after we switch to lazy_dual approach, do we still need to limit it to cpu &gpu backend?

ailzhang (approved):
Thanks a lot!!

tests/python/test_ad_basics_fwd.py
erizmr: Good catch! Have removed the limit. :)

PR #5226, opened 2022-06-22
[gui] Let gui and renderer manage the resource they own

ailzhang:
* #5227
* __->__ #5226

Tbh it's better to also refactor `Renderer::cleanup()` as well but it's
a bit more involved.


bobcao3 (approved):
LGTM

PR #5227, opened 2022-06-22
[ggui] Fix the device memory leak when GGUI terminates

ailzhang:
* (to be filled)

`ti example diff_sph` usually crashes at termination on my desktop
complaining unfreeed device memory. The root cause for it is that
PyWindow's lifetime is controlled by python, however it shares the same
underlying `VulkanDevice` with `Program` (when ti.arch==vulkan). During
Python shutdown, we don't have a good way to enforce the destruction order of `PyWindow` and
`Program` so it's possible that `VulkanDevice` (uniquely owned by
`Program`) get destructed first and relealized some allocations
allocated for GGUI aren't freed properly. This PR kinda worked around
the problem by enforcing deallocation upon device destruction.
Tbh a better approach to fix it is just let GGUI gets a `shared_ptr` of
device instead of a raw pointer, we had some concern about  perf before
so let's delay that until it's necessary.

### DIFF
--- a/taichi/backends/vulkan/vulkan_device.cpp
+++ b/taichi/backends/vulkan/vulkan_device.cpp
@@ -1250,6 +1250,25 @@ void VulkanDevice::init_vulkan_structs(Params &params) {
 }
 
 VulkanDevice::~VulkanDevice() {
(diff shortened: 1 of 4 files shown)


bobcao3 (approved):
LGTM

PR #5228, opened 2022-06-22
[ci] Workaround DLLs locking issue in Windows build

feisuzhu:
Related issue = #


feisuzhu:
This is irrelevant now, it didn't happen for a long time, and virtualenvs are now reused.

PR #5232, opened 2022-06-22
[type] [refactor] Decouple quant from SNode 3/n: Extend bit pointers

strongoier:
Related issue = #4857

The original bit pointer is like:
```cpp
struct bit_pointer {
    i8 *byte_ptr;
    i32 bit_offset;
};
```

When creating a new bit pointer, the `byte_ptr` is always cast into `i8 *` first. Then a physical type needs to get passed alongside to allow access of its pointee.

As you may have noticed, the `byte_ptr` is not a pointer to a byte, but simply a byte-level pointer, which can point to any primitive type. Therefore, there is actually no need to force `byte_ptr` as `i8 *`. Then info of its physical type can be easily preserved by extending the definition of bit pointers:
```cpp
// X=8/16/32/64
struct bit_pointer_X {
    iX *byte_ptr;
    i32 bit_offset;
}
```

After this change, logic around bit pointers can get simplified a lot.

### DIFF
--- a/taichi/codegen/codegen_llvm.cpp
+++ b/taichi/codegen/codegen_llvm.cpp
@@ -1195,11 +1195,9 @@ llvm::Value *CodeGenLLVM::quant_type_atomic(AtomicOpStmt *stmt) {
 
   auto dst_type = stmt->dest->ret_type->as<PointerType>()->get_pointee_type();
   if (auto qit = dst_type->cast<QuantIntType>()) {
-    return atomic_add_quant_int(
-        stmt, qit, stmt->dest->as<GetChStmt>()->input_snode->physical_type);
+    return atomic_add_quant_int(stmt, qit);
   } else if (auto qfxt = dst_type->cast<QuantFixedType>()) {
-    return atomic_add_quant_fixed(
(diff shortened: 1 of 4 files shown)


taichi/codegen/codegen_llvm.cpp
```diff
@@ -1479,61 +1474,44 @@ void CodeGenLLVM::visit(LinearizeStmt *stmt) {
...
-  //    i32 offset;
+llvm::Value *CodeGenLLVM::create_bit_ptr(llvm::Value *byte_ptr,
+                                         llvm::Value *bit_offset) {
+  // 1. define the bit pointer struct (X=8/16/32/64)
+  // struct bit_pointer_X {
+  //    iX* byte_ptr;
+  //    i32 bit_offset;
   // };
+  TI_ASSERT(bit_offset->getType()->isIntegerTy(32));
   auto struct_type = llvm::StructType::get(
```
k-ye, line 1485: OOC, where do we name `struct_type`?

k-ye (approved):
LGTM!

taichi/codegen/codegen_llvm.cpp
strongoier: We don't name it - this is hashed by its element types in LLVM.

PR #5233, opened 2022-06-22
[lang] 1d/3d texture support

bobcao3:
Related issue = #5109


bobcao3:
3D texture tested in https://github.com/bobcao3/voxel-challenge

k-ye (approved):
LGTM!

Can we test this feature? One way would probably to generate a 2D image on the fly, sample a few known points on it, and compare with the result.

PR #5208, opened 2022-06-20
[build] [refactor] Move metal source files to target locations

qiao-bo:
Related issue = #4832

One of the series of steps to split source files from backends, this PR addresses Metal:

- Move Metal runtime files to `taichi/runtime/metal` for target `metal_runtime`.
- Move Metal program impl files to `taichi/runtime/program_impls/metal` for target `metal_program_impl`.
- The remaining files in `taichi/backends/metal` (Device API files) belongs to the target `metal_rhi`.

The dependency relation is:
metal_rhi <- metal_runtime <- metal_program_impl <- core

In addition, this PR also breaks the usage of global file glob in Cmake core target (all sources were built under this target):
https://github.com/taichi-dev/taichi/blob/71b8acc9d74cd7d4f52d5f4a479224ce2a407603/cmake/TaichiCore.cmake#L94-L96
This allows us to be explicit on which files are needed by which target.

PR #5235, opened 2022-06-23
[build] [refactor] Move wasm and dx11 source files to target locations

qiao-bo:
Related issue = #4832 

- This PR follows #5200 and #5208. By moving `wasm` and `dx11` sources to `runtime`, `codegen`, and `program_impl` location.
- Remove the corresponding sources from `core`.
- Add CMake targets decls for `codegen`


taichi/codegen/metal/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# ./taichi/codegen/metal/CMakeLists.txt
+
+add_library(metal_codegen)
```
k-ye, line 3: I noticed that metal's shaders are spread in both `codegen/` and `runtime/`, so are OpenGL's. Since we plan to make `codegen` depend on `runtime`, we can place these shaders all under runtime.

https://github.com/taichi-dev/taichi/blob/b1ba4572a1c0fef12fe71b63c8e28b289c77e2ad/taichi/codegen/metal/codegen_metal.cpp#L31-L32

k-ye (approved):
LGTM!

bobcao3 (approved):
LGTM

PR #5237, opened 2022-06-23
[bug] Fixed misuse of std::forward

### LINKED ISSUE #5206: Misuses of `std::forward`
There are some misuses of `std::forward` in the C++ codebase. For example

https://github.com/taichi-dev/taichi/blob/5a066fbcc9caddebfcc89c9ec7a4dc0a8902aaa3/taichi/llvm/llvm_program.h#L170
`args` is not passed by reference.

https://github.com/taichi-dev/taichi/blob/5a066fbcc9caddebfcc89c9ec7a4dc0a8902aaa3/c_api/src/taichi_core_impl.cpp#L32

`aot_module` is not passed by universal reference. (A more serious problem here is that `unique_ptr` should be passed by value because you're sure you want to transfer ownership. Users can't know whether the ownership is transferred by the interface if it is passed by rvalue reference).

In general, `std::forward` should only be used for perfect forwarding with universal reference. Although these misuses are not that harmful, I think it's better to eliminate them.

jim19930609:
Related issue = fix #5206

### DIFF
--- a/c_api/src/taichi_core_impl.cpp
+++ b/c_api/src/taichi_core_impl.cpp
@@ -28,10 +28,9 @@ VulkanRuntime *Runtime::as_vk() {
 
 AotModule::AotModule(Runtime &runtime,
                      std::unique_ptr<taichi::lang::aot::Module> &&aot_module)
-    : runtime_(&runtime),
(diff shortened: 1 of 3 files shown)


ailzhang (approved):
LGTM！

PR #5238, opened 2022-06-23
[type] [bug] Refine SNode with quant 1/n: Fix (atomic_)set_mask_b##N

strongoier:
Related issue = #4857

https://github.com/taichi-dev/taichi/blob/b1ba4572a1c0fef12fe71b63c8e28b289c77e2ad/taichi/runtime/llvm/runtime.cpp#L1827-L1838
In the above function, the argument `value` is not guaranteed to contain zeros at non-masked positions. This is especially buggy when `value` is originally a negative number. We need to replace `value` with `(value & mask)` to avoid pollute other positions. A related test case is added.


k-ye (approved):
thx!

PR #5244, opened 2022-06-23
[dx] Fix DX backend with new runtime & Better D3D11 buffer handling

bobcao3:


taichi/backends/device.h
```diff
@@ -385,6 +385,7 @@ struct PipelineSourceDesc {
 
 // FIXME: this probably isn't backend-neutral enough
 enum class AllocUsage : int {
+  Transfer = 0,
```
k-ye, line 388: Actually, let's append to this list and use `16`? `0` should be reserved if we are using it as a bitmask.

bobcao3: This is intended, with usage of none we default it to transfer

bobcao3: Maybe actually should just call it none? As we assume transfer is always possible on all buffers.

.github/workflows/testing.yml
```diff
@@ -442,7 +442,7 @@ jobs:
         run: |
           .\.github\workflows\scripts\win_build.ps1 -installVulkan -install -libsDir C:\
         env:
-          TAICHI_CMAKE_ARGS: -DTI_WITH_OPENGL:BOOL=ON -DTI_WITH_CC:BOOL=OFF -DTI_BUILD_TESTS:BOOL=ON
+          TAICHI_CMAKE_ARGS: -DTI_WITH_OPENGL:BOOL=ON -DTI_WITH_DX11:BOOL=ON -DTI_WITH_CC:BOOL=OFF -DTI_BUILD_TESTS:BOOL=ON
```
k-ye, line 445: Can we support DX in our test env now?  @feisuzhu

k-ye (approved):
LGTM, would be great if @quadpixels  you could help review as well, thanks!

quadpixels:
I just built this change and ran a few test programs, it seems to work well for me

PR #5248, opened 2022-06-24
[ci] Use clang for performance monitoring

feisuzhu:
Related issue = #5133


.github/workflows/performance_monitoring.yml
```diff
@@ -27,7 +26,9 @@ jobs:
           rm -rf performance-monitoring
           git clone git@github.com:taichi-dev/performance-monitoring.git
           cd performance-monitoring
+          export CXX=clang
```
ailzhang, line 29: nit: better put this in the `env` section below before merging, also `CXX` should point to clang++-10 ;)

feisuzhu:
This is irrelevant now, perf mon will be refactored.

PR #5249, opened 2022-06-24
[AOT] Added safety checks to prevent hard crashes on failure

PENGUINLIONG:
This PR added safety checks to Taichi C-APIs, so that a call with apparently invalid arguments won't trigger a hard crash. Human readable warning are logged for debugability.


k-ye (approved):
LGTM!

PR #5253, opened 2022-06-24
[Lang] [type] Refine SNode with quant 2/n: Enable struct for on bit_array with bit_vectorize off

strongoier:
Related issue = #4857

Previous implementation of struct for on `bit_array` assumes `bit_vectorize` is on, and you cannot access a single element inside. However, `bit_vectorize` is designed for 0/1 arrays, which is not suitable for normal cases. This PR turns `bit_vectorize` off by default, and supports reading/writing a single element inside struct for on `bit_array`.

Notes:
- `bit_vectorize == 1` means 1-bit vectorization, which is indeed saying that `bit_vectorize` is off
- The key here is to add `listgen` for the `bit_array` SNode when `bit_vectorize` is off


ailzhang (approved):
LGTM!
btw `bit_vectorized==1 => bit_vectorized is off` isn't that intuitive but I'd assume you'll refactor this in the followup PRs?

strongoier:
> LGTM! btw `bit_vectorized==1 => bit_vectorized is off` isn't that intuitive but I'd assume you'll refactor this in the followup PRs?

Yep :-)

PR #5246, opened 2022-06-24
[opengl] Move OpenGL backend onto Gfx runtime

bobcao3:

PR #5264, opened 2022-06-27
[build] [bug] Fix dependency for opengl_rhi target

qiao-bo:
After #5246, `opengl_api` is moved to target `opengl_rhi`, which introduces dependency on glfw.  This PR fixes the build error `undefined reference to glfwInit/CreateWindows`.


bobcao3 (approved):
LGTM! What about Vulkan which also depends on GLFW?

qiao-bo:
> LGTM! What about Vulkan which also depends on GLFW?

Strange, did not get any error on that. Maybe a better solution is to write the dependency in `gfx_rhi`, and `vulkan/opengl rhi` depend on `gfx_rhi`? WDYT?

PR #5268, opened 2022-06-27
[metal] Support memcpy_internal via buffer_copy

ailzhang:
There isn't a standalone test for this yet but will try to add one in
the followup PRs.

Related issue = #


k-ye (approved):
LGTM!

PR #5271, opened 2022-06-27
[AOT] Added C-API for on-device memory copy

PENGUINLIONG:
This PR introduces a new C-API of, `ti_copy_memory`, to enable on-device memory copy without interrupting sequences of kernel launches to eliminate host-device synchronization. It is especially useful when the user wants to transfer Taichi results to externally allocated device memory (which might not support compute shader write).

### DIFF
--- a/c_api/include/taichi/taichi_core.h
+++ b/c_api/include/taichi/taichi_core.h
@@ -140,10 +140,11 @@ TI_DLL_EXPORT void *TI_API_CALL ti_map_memory(TiRuntime runtime,
 TI_DLL_EXPORT void TI_API_CALL ti_unmap_memory(TiRuntime runtime,
                                                TiMemory memory);
 
-// function.copy_memory
-TI_DLL_EXPORT void TI_API_CALL ti_copy_memory(TiRuntime runtime,
-                                              const TiMemorySlice *dst_memory,
-                                              const TiMemorySlice *src_memory);
+// function.copy_memory_device_to_device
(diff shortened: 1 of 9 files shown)


taichi/program/program.cpp
```diff
@@ -23,6 +22,7 @@
 #include "taichi/math/arithmetic.h"
 #ifdef TI_WITH_LLVM
 #include "taichi/runtime/program_impls/llvm/llvm_program.h"
+#include "taichi/codegen/llvm/struct_llvm.h"
```
PENGUINLIONG, line 25: This blocks compilation when LLVM is not enabled.

c_api/include/taichi/taichi_core.h
```diff
@@ -133,6 +140,11 @@ TI_DLL_EXPORT void *TI_API_CALL ti_map_memory(TiRuntime runtime,
 TI_DLL_EXPORT void TI_API_CALL ti_unmap_memory(TiRuntime runtime,
                                                TiMemory memory);
 
+// function.copy_memory
+TI_DLL_EXPORT void TI_API_CALL ti_copy_memory(TiRuntime runtime,
```
k-ye, line 144: WDYT if this is called `ti_copy_memory_d2d` to highlight this is device copy?

k-ye: Sure, SG!

k-ye (approved):
LGTM!

PR #5272, opened 2022-06-27
[AOT] Refactor C-API codegen

PENGUINLIONG:
This PR introduces a common interface for `taichi.json` parsing, so that language binding generator scripts can save some boilerplates and avoid inconsistent JSON property interpretation between versions of scripts. Existing generators for C/C++ headers and C# marshalling & import have been adapted to load module descriptions from `taichi_json.py`.

The file names and hard-coded relative paths are subject to change.


ailzhang (approved):
Didn't review carefully :P but overall having a taichi json file parser is good!

PR #5273, opened 2022-06-27
[AOT] Unity native plugin interfaces

PENGUINLIONG:
This PR introduces a new C-API module, `taichi/taichi_unity.h` to provide extension APIs for Unity integration as a native rendering plugin. An extra build target `taichi_unity.dll` is added with enabling cmake flag `TI_WITH_UNITY`.

dont merge this tho. im still working on it

### DIFF
--- a/c_api/include/taichi/taichi_unity.h
+++ b/c_api/include/taichi/taichi_unity.h
@@ -51,6 +51,9 @@ tix_copy_memory_host_to_device_unity(TiRuntime runtime,
                                      const void *src,
                                      uint64_t src_offset);
 
+// function.submit_async
+TI_DLL_EXPORT void TI_API_CALL tix_submit_async_unity(TiRuntime runtime);
+
 #ifdef __cplusplus
 }  // extern "C"
 #endif  // __cplusplus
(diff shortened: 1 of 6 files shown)


c_api/include/taichi/taichi_unity.h
```diff
@@ -0,0 +1,54 @@
...
+
+// handle.native_buffer
+typedef struct TixNativeBufferUnity_t *TixNativeBufferUnity;
+
+// function.import_native_runtime
+TI_DLL_EXPORT TiRuntime TI_API_CALL tix_import_native_runtime_unity();
+
+// function.launch_kernel_async
+TI_DLL_EXPORT void TI_API_CALL
+tix_launch_kernel_async_unity(TiRuntime runtime,
```
k-ye, line 16: I'm very curious about this set of APIs. Is it fair to say that this treats Unity itself as a new Taichi backend..?

PENGUINLIONG: Actually not. These APIs' impl merely extract the runtime device from unity's provision; and calls the core APIs (taichi/taichi_core.h). So actually taichi_unity is a user of taichi_c_api.

These async variants only differs from the existing ones by caching all the launch commands in a temporary storage IN GAME THREAD; and actually consume the commands IN RENDER THREAD in the callback function returned by tix_submit_async_unity (note that it returns a void*).

So b

k-ye (approved):
Looking forward to the full Unity demo!

PR #5275, opened 2022-06-27
[Bug] Fix infinite loop when exponent of integer pow is negative

### LINKED ISSUE #4661: Infinite Loop with Exponentiation
It should either do implicit type conversion or trigger error.
```
import taichi as ti

@ti.kernel
def buggy():
	a = ti.Vector([10,10,10])
	print(a ** -2)

@ti.kernel
def correct():
	a = ti.Vector([10.0,10.0,10.0])
	print(a ** -2)

ti.init()
buggy()
```

AD1024:
Related issue = fix #4661

### DIFF
--- a/taichi/transforms/check_out_of_bound.cpp
+++ b/taichi/transforms/check_out_of_bound.cpp
@@ -10,21 +10,14 @@ TLANG_NAMESPACE_BEGIN
 // TODO: also check RangeAssumptionStmt
 
 class CheckOutOfBound : public BasicStmtVisitor {
- private:
-  CompileConfig config_;
-
  public:
   using BasicStmtVisitor::visit;
   std::set<int> visited;
   DelayedIRModifier modifier;
   std::string kernel_name;
 
-  explicit CheckOutOfBound(const std::string &kernel_name,
-                           const CompileConfig &config)
-      : BasicStmtVisitor(),
-        config_(config),
-        visited(),
-        kernel_name(kernel_name) {
+  explicit CheckOutOfBound(const std::string &kernel_name)
+      : BasicStmtVisitor(), visited(), kernel_name(kernel_name) {
   }
 
   bool is_done(Stmt *stmt) {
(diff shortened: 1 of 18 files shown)


AD1024:
After several attempts to deal with the issue of losing accuracy, I decide to change from converting all pow to return floats to throwing exceptions when pow receives negative exponents when lhs is an integer (a la torch).
For some backends such as Vulkan (correct me if I'm wrong), we could not tell from pow with floats from those with ints if we make all pows to return floats. However, those backends depend on this information to decide whether to insert precision correction (e.g. Adding biases). Thus throwing exception is probably the 'right' way to fix this issue.

taichi/transforms/type_check.cpp
```diff
@@ -355,6 +391,32 @@ class TypeCheck : public IRVisitor {
...
+    if (stmt->op_type == BinaryOpType::pow) {
+      if (is_integral(stmt->rhs->ret_type) &&
+          is_integral(stmt->lhs->ret_type)) {
+        if (config_.debug) {
+          auto iter = stmt->parent->find(stmt);
+          auto compare_rhs = Stmt::make<ConstStmt>(TypedConstant(0));
+          auto compare = std::make_unique<BinaryOpStmt>(
+              BinaryOpType::cmp_ge, stmt->rhs, compare_rhs.get());
+          compare->ret_type = PrimitiveType::i32;
+          if (iter == stmt->parent->statements.begin() ||
```
strongoier, line 404: Could you explain a bit about this condition?

strongoier (commented):
The API doc needs to be updated :-) FYI https://github.com/taichi-dev/taichi/blob/0dfd1660ce8f91367fc2fd413f42e00c742fc240/python/taichi/lang/ops.py#L731-L760

taichi/transforms/type_check.cpp
AD1024: We insert an assertion when:
1. The stmt is the first one in the block
2. The stmt before `pow` is not an assertion
3. The stmt before `pow` is not the assertion checking the exponent

strongoier: Why do we check these conditions?

AD1024: Type check will run several times, so if we don't check these, there will be duplicated assertions inserted.

strongoier: This method seems a bit hacky - is it better to put these runtime assertions into a separate pass? wdyt @jim19930609

jim19930609: Yup, `irpass::check_out_of_bound` sounds like a better place for this assertion.

AD1024: Good idea. I noticed that this pass will potentially run more than once:
https://github.com/taichi-dev/taichi/blob/f1068ae3d6265067d4614387bbb232967e9e994b/taichi/transforms/check_out_of_bound.cpp#L112-L119

the workaround can be less hacky tho

taichi/transforms/check_out_of_bound.cpp
```diff
@@ -104,10 +111,37 @@ class CheckOutOfBound : public BasicStmtVisitor {
...
 
+  void visit(BinaryOpStmt *stmt) override {
+    // Insert assertions if debug is on
+    if (is_done(stmt)) {
+      return;
+    }
+    if (stmt->op_type == BinaryOpType::pow) {
+      if (is_integral(stmt->rhs->ret_type) &&
+          is_integral(stmt->lhs->ret_type)) {
+        if (config_.debug) {
```
strongoier, line 122: This line may not be necessary.
FYI
https://github.com/taichi-dev/taichi/blob/d4f4abf289f75ec17663c0dbf2ef690112580e4e/taichi/program/program.cpp#L76-L77
https://github.com/taichi-dev/taichi/blob/d4f4abf289f75ec17663c0dbf2ef690112580e4e/taichi/transforms/compile_to_offloads.cpp#L99-L103

strongoier (approved):
LGTM!

PR #5280, opened 2022-06-28
[Lang] Add errror message for missing init call

neozhaoliang:


python/taichi/lang/impl.py
```diff
@@ -490,7 +490,13 @@ def create_field_member(dtype, name):
     dtype = cook_dtype(dtype)
 
     # primal
-    x = Expr(get_runtime().prog.make_id_expr(""))
+    try:
+        x = Expr(get_runtime().prog.make_id_expr(""))
+    except:
+        raise TaichiRuntimeError(
```
ailzhang, line 496: This fix is a bit hacky, would you mind explaining  which issue are you trying to solve and what was the original error message?

ailzhang: Yea then the right way is to check if `get_runtime().prog is None` instead of this `try...except` block?

neozhaoliang: Yes that's a better solution.

ailzhang (changes requested):
Would you mind splitting these to two PRs? Thanks!
(I feel like the change to `floor` is a bit tricky and this PR tries to fix two things in one place.

neozhaoliang:
> Would you mind splitting these to two PRs? Thanks! (I feel like the change to `floor` is a bit tricky and this PR tries to fix two things in one place.

Sure, it 's now split into two PRS.

ailzhang (approved):
LGTM! 
(Btw you'll have to fix pylint errors manually ;)

neozhaoliang:
Fixed.

Issue #5236, opened 2022-06-23
Cannot close gui window in jupyter notebook


neozhaoliang:
Description:

I want to call `ti.GUI` or `ti.gui.WIndow` in a jupyter notebook to display a window, but then I cannot close that window unless force to kill the server. And for `ti.ui.Window`, it seems the kernel will crash and restart after a while.

Code:

```python
import taichi as ti
ti.init()

gui = ti.ui.Window("hello", res=(400, 400))
can = gui.get_canvas()

img = ti.Vector.field(3, float, (400, 400))
while gui.running:
    can.set_image(img)
    gui.show()
```

neozhaoliang:
I found a solution here:

use `while not gui.get_event(ti.GUI.ESCAPE, ti.GUI.EXIT):` instead.

ailzhang:
Reopening this issue since we either need to figure out why it's freezing with `gui.running` and fix it or update all examples to avoid using gui.running

PENGUINLIONG:
`gui.destroy()` is all you need.

```diff
import taichi as ti
ti.init()

gui = ti.ui.Window("hello", res=(400, 400))
can = gui.get_canvas()

img = ti.Vector.field(3, float, (400, 400))
while gui.running:
    can.set_image(img)
    gui.show()

+gui.destroy()
```

neozhaoliang:
That works, thanks!

PR #5281, opened 2022-06-28
[Lang] Fix fractal gui close warning

neozhaoliang:
When closing the gui window in `fractal.py` it raises lots of warning messages. This PR fixes this problem.

### DIFF
--- a/python/taichi/examples/simulation/fractal.py
+++ b/python/taichi/examples/simulation/fractal.py
@@ -26,7 +26,7 @@ def paint(t: float):
 def main():
     gui = ti.GUI("Julia Set", res=(n * 2, n))
     t = 0.0
-    while gui.running:
+    while not gui.get_event(ti.GUI.ESCAPE, ti.GUI.EXIT):
         paint(t)
         t += 0.03
         gui.set_image(pixels)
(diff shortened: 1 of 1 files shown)


ailzhang (approved):
LGTM!

python/taichi/examples/simulation/fractal.py
```diff
@@ -25,9 +25,10 @@ def paint(t: float):
 
 def main():
     gui = ti.GUI("Julia Set", res=(n * 2, n))
-
-    for i in range(1000000):
-        paint(i * 0.03)
+    t = 0.0
+    while not gui.get_event(ti.GUI.ESCAPE, ti.GUI.EXIT):
```
ailzhang, line 29: Actually maybe it's better  to check `gui.running` and set it properly?

neozhaoliang: I found using `while not gui.get_event(ti.GUI.ESCAPE, ti.GUI.EXIT):`is a better practice because:
1.  the user can press the  `escape` key to exit
2.  Using `while gui.running` to call the gui in jupter notebook will "freeze" the window: the user cannot close the gui unless he kills the notebook thread.

ailzhang: Oh @neozhaoliang that's good to know, btw can you open an issue for this and see if 
- we can fix the `freeze` when using `while gui.running`
- or update our examples to get rid of `while gui.running`
Thanks!

neozhaoliang: A related issuse is here: #5236

ailzhang: Reopened the issue so that we can properly followup :)

Issue #5274, opened 2022-06-27
API to query whether a field needs gradients


ifsheldon:
**Concisely describe the proposed feature**
I would like to request an API to query whether a field needs gradients. 

**Describe the solution you'd like (if any)**
An attribute or property attached to a field object, which states whether a field needs gradients in backward-mode differentiation.

**Additional comments**
This feature request dates back to #2498 whose conversation was in the old Slack community channel. In my use case, I need it to check whether a field passed by the user of my library `stannum` needs gradients or not, and based on that my library creates different tensors. Besides my use case, it is a common need to check if a field needs gradients or not.

erizmr:
Hi @ifsheldon , we are implementing this feature and want to double check:  if the API you request is to check whether a field is annotated with `needs_grad` (but may has been or has not been allocated with a grad field) or to check whether the field has been allocated a grad field?

ifsheldon:
Yeah, it's sufficient I think. As the name `needs_grad` suggests, a field needs gradients may not have been allocated with gradients, which is fine.

PR #5293, opened 2022-06-29
[autodiff] Add has_grad for checking grad field

erizmr:
Related issue = #5274 
Add a public API `field.has_grad()` for checking whether a field has a grad field for reverse mode autodiff.
So that users can avoid using internal functions `field.snode.ptr.has_adjoint() ` to achieve the same goal.


erizmr:
outdated

PR #5297, opened 2022-06-30
[refactor] Switch to using staging buffer for metal/vulkan/opengl

ailzhang:
* #5299
* #5298
* __->__ #5297

ndarray host accesses

### DIFF
--- a/taichi/program/ndarray.cpp
+++ b/taichi/program/ndarray.cpp
@@ -11,6 +11,22 @@
 namespace taichi {
 namespace lang {
 
+namespace {
+
+size_t flatten_index(const std::vector<int> &shapes,
+                     const std::vector<int> &indices) {
+  TI_ASSERT(shapes.size() == indices.size());
+  if (indices.size() == 1) {
+    return indices[0];
+  } else {
+    size_t ind = indices[0];
+    for (int i = 1; i < indices.size(); i++) {
+      ind = ind * shapes[i] + indices[i];
+    }
+    return ind;
+  }
+}
+}  // namespace
 Ndarray::Ndarray(Program *prog,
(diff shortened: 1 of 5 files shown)


taichi/program/ndarray.cpp
```diff
@@ -103,24 +119,96 @@ std::size_t Ndarray::get_nelement() const {
   return nelement_;
 }
 
+template <typename T>
+T Ndarray::read(const std::vector<int> &I) const {
+  prog_->synchronize();
+  size_t index = flatten_index(total_shape_, I);
+  size_t size_ = sizeof(T);
```
bobcao3, line 126: How large would T typically be? If it's limited to primitive types, we can reuse the same buffer for these accessors to save the cost of buffer allocation

tests/cpp/program/graph_test.cpp
```diff
@@ -48,6 +48,10 @@ TEST(GraphTest, SimpleGraphRun) {
   args.insert({"x", aot::IValue::create<int>(2)});
 
   g->run(args);
+  // TODO: this should be synchronize() indeed but it currently has a shortcut
+  // `sync=True` to return which is pretty error-prone. We should change this to
+  // synchronize() after getting rid of the shortcut.
```
bobcao3, line 53: ooc, why can't we change this to `synchronize` rn?

bobcao3 (approved):
LGTM, other than the one nitpicking

taichi/program/ndarray.cpp
ailzhang: Yup in this case it's typically primitive types, will send a followup PR to cache buffer allocation. :D

PR #5298, opened 2022-06-30
[refactor] Use device api for CPU/CUDA ndarray

ailzhang:
* #5299
* __->__ #5298
* #5297

### DIFF
--- a/taichi/backends/cuda/cuda_device.cpp
+++ b/taichi/backends/cuda/cuda_device.cpp
@@ -79,6 +79,31 @@ void CudaDevice::dealloc_memory(DeviceAllocation handle) {
   }
 }
 
+void *CudaDevice::map(DeviceAllocation alloc) {
+  AllocInfo &info = allocations_[alloc.alloc_id];
+  size_t size = info.size;
+  info.mapped = new char[size];
+  // FIXME: there should be a better way to do this...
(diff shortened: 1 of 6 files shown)


taichi/backends/cuda/cuda_device.cpp
```diff
@@ -79,6 +79,31 @@ void CudaDevice::dealloc_memory(DeviceAllocation handle) {
   }
 }
 
+void *CudaDevice::map(DeviceAllocation alloc) {
+  AllocInfo &info = allocations_[alloc.alloc_id];
+  size_t size = info.size;
+  info.mapped = new char[size];
+  // FIXME: there should be a better way to do this...
```
bobcao3, line 86: Don't see a better way without using unified memory. However the host access flags when allocating the memory should inform the behavior when mapping. E.g. when a memory is flagged only with host write, then we don't need to copy data from GPU to host. When we only have host read, we don't need to copy data back when we unmap, etc.

bobcao3 (approved):
LGTM

taichi/backends/cuda/cuda_device.cpp
ailzhang: Because our synchronize don't always synchronize...  https://github.com/taichi-dev/taichi/blob/master/taichi/program/program.cpp#L231 It requires pretty explicit setting of `sync` variable to make sure synchronize works properly.  This  TODO is about simplifying that part before we use it here.

PR #5302, opened 2022-06-30
[refactor] Synchronize should always be called in non-async mode

ailzhang:
Related issue = #
I got tripped over a few times by `sync` flag that calling `prog.synchronize()` didn't really synchronize due to a fake `sync=True`. Having a manual `sync` flag requires us to know exactly every place could potentially submit jobs and set the flag properly. IMHO this is pretty error prone -  when I call `sync` I'd prefer it to be 100% synchronized even though it might cause 1 extra no-op sync compared to the current `sync` flag solution. 

Originally I wanted to get rid of this flag completely but realized that it's used in the async mode so this PR only decouples it with normal mode.

### DIFF
--- a/taichi/program/program.cpp
+++ b/taichi/program/program.cpp
@@ -227,15 +227,14 @@ void Program::check_runtime_error() {
 }
 
 void Program::synchronize() {
-  if (!sync) {
-    if (config.async_mode) {
-      async_engine->synchronize();
-    }
-    if (arch_uses_llvm(config.arch) || config.arch == Arch::metal ||
-        config.arch == Arch::vulkan || config.arch == Arch::opengl) {
-      program_impl_->synchronize();
-    }
+  if (config.async_mode && !sync) {
+    async_engine->synchronize();
     sync = true;
+  }
+  // Normal mode shouldn't be affected by `sync` flag.
+  if (arch_uses_llvm(config.arch) || config.arch == Arch::metal ||
+      config.arch == Arch::vulkan || config.arch == Arch::opengl) {
+    program_impl_->synchronize();
   }
 }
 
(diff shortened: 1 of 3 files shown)


taichi/program/program.cpp
```diff
@@ -227,16 +227,15 @@ void Program::check_runtime_error() {
...
 void Program::synchronize() {
-  if (!sync) {
-    if (config.async_mode) {
-      async_engine->synchronize();
-    }
-    if (arch_uses_llvm(config.arch) || config.arch == Arch::metal ||
-        config.arch == Arch::vulkan || config.arch == Arch::opengl) {
-      program_impl_->synchronize();
-    }
+  if (config.async_mode && !sync) {
```
k-ye, line 230: Is this part of the plan to remove async..?

taichi/runtime/program_impls/vulkan/vulkan_program.h
```diff
@@ -47,7 +47,9 @@ class VulkanProgramImpl : public ProgramImpl {
   void materialize_snode_tree(SNodeTree *tree, uint64 *result_buffer) override;
 
   void synchronize() override {
-    vulkan_runtime_->synchronize();
+    if (vulkan_runtime_) {
```
k-ye, line 50: When is `vulkan_runtime_` not inited?

ailzhang: @k-ye when you have created a program for compilation only (not execution) e.g. https://github.com/taichi-dev/taichi/blob/master/tests/cpp/aot/aot_save_load_test.cpp#L23-L108 
(ofc it works since these kernels didn't require jit evaluation.

taichi/program/program.cpp
ailzhang: Yup! I've tried the only blocker for removing this flag is  async tests so it should be gone after the async removal :D

taichi/runtime/program_impls/vulkan/vulkan_program.h
k-ye: ... sounds brittle

k-ye (approved):
LGTM!

PR #5303, opened 2022-07-01
[gfx] Reduce args buffer size & use push constants when possible

bobcao3:


taichi/rhi/cpu/cpu_device.h
```diff
@@ -49,6 +49,8 @@ class CpuCommandList : public CommandList {
   }
 
   void bind_pipeline(Pipeline *p) override{TI_NOT_IMPLEMENTED};
+  void set_push_constants(size_t constants_size,
+                          void *constants_data) override{TI_NOT_IMPLEMENTED};
```
k-ye, line 53:
```suggestion
                          void *constants_data) override{TI_NOT_IMPLEMENTED;};
```

This coding style looks a bit weird...

taichi/runtime/gfx/runtime.cpp
```diff
@@ -470,6 +469,15 @@ void GfxRuntime::launch_kernel(KernelHandle handle, RuntimeContext *host_ctx) {
   // Record commands
   const auto &task_attribs = ti_kernel->ti_kernel_attribs().tasks_attribs;
 
+  if (args_buffer) {
```
k-ye, line 472: Do I understand correctly that: if `args_buffer` is allocated on the device, use that; otherwise assume this is using push constants?

taichi/rhi/cpu/cpu_device.h
bobcao3: That's an oops

taichi/runtime/gfx/runtime.cpp
bobcao3: Yes. But actually the other way around, args_buffer is allocated iff we are not using push constants (i.e. when the device supported maximum push constants size is smaller than the args)

k-ye (approved):
LGTM!

ailzhang (approved):
LGTM!

k-ye (changes requested):
CI is broken..

PR #5305, opened 2022-07-01
[llvm] [aot] Added CGraph tests for LLVM backend

jim19930609:
Related issue = #4800

1. Modified RuntimeContext construction inside `CompiledGraph::run()`, so that LlvmAotModuleLoader is able to setup `LLVMRuntime runtime_` upon CompiledGraph construction.

2. Rearranged file structure under /cpp/tests/aot, so that "mpm88" test is shared between llvm and vulkan backends. 

3. Added `tests/cpp/aot/llvm/mpm88_graph_aot_test.cpp`, `tests/cpp/aot/llvm/mpm88_graph_aot_cpu.py`, `tests/cpp/aot/llvm/mpm88_graph_aot_cuda.py` as LLVM AOT test cases.

*Will replace "LlvmProgramImpl" with "LlvmRuntimeExecutor" across all the LLVM AOT tests in a following PR


taichi/program/ndarray.cpp
```diff
@@ -194,6 +194,14 @@ void set_runtime_ctx_ndarray(RuntimeContext *ctx,
                              int arg_id,
                              Ndarray *ndarray) {
   ctx->set_arg_devalloc(arg_id, ndarray->ndarray_alloc_, ndarray->shape);
+
+  int total_array_size = 1;
```
ailzhang, line 198:
```suggestion
  uint64 total_array_size = 1;
```

tests/cpp/aot/llvm/mpm88_graph_aot_cpu.py
```diff
@@ -0,0 +1,10 @@
+import os
+import sys
+
+import taichi as ti
+
+sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), ".."))
```
ailzhang, line 6: nit: I'm personally not a big fan of manipulating `sys.path`. I'm not too worried about this particular use case but in general afraid of changing `sys.path` spreading across the codebase. 
Can we try passing in arch information to the script?

ailzhang (approved):
LGTM!!

tests/cpp/aot/llvm/mpm88_graph_aot_cpu.py
jim19930609: Thanks! Switched to pass "PYTHONPATH" as environmental variables during subprocess call (at tests/run_tests.py).

taichi/program/ndarray.cpp
jim19930609: Done, thanks!

tests/cpp/aot/llvm/mpm88_graph_aot_cpu.py
jim19930609: Thanks! Switched to pass in arguments~

ailzhang (approved):
Thanks!!

Issue #5284, opened 2022-06-28
Gradient of loss w.r.t. intermediate variable in kernel function


zhouxian:
I am seeing some unexpected behavior when running the following snippet:
```python
import taichi as ti
ti.init()

x = ti.field(dtype=ti.f32, shape=(), needs_grad=True)
a = ti.field(dtype=ti.f32, shape=(), needs_grad=True)
b = ti.field(dtype=ti.f32, shape=(), needs_grad=True)


@ti.kernel
def func():
   a[None] = 2 * x[None]
   b[None] = a[None] + 1
   
x[None] = 1

with ti.Tape(loss=b):
   func()
print(a.grad[None], x.grad[None])
```

I am expecting to see x.grad[None] = 2.0 and a.grad[None] = 1.0, but it prints a.grad[None] = 0.0. I think the usage of variables in func() does not violate the rules in autodiff.
Is there any explanation of this? It seems autodiff only computes gradient for the variable that's only being read in the kernel function?

erizmr:
Hi @zhouxian , the bug has been fixed by #5315. 
It will be delivered in next release. Or you can try `pip install -i https://pypi.taichi.graphics/simple/ taichi-nightly` to get the lastest commits.

PR #5315, opened 2022-07-03
[autodiff] Skip store forwarding to keep the GlobalLoadStmt alive

erizmr:
Related issue = #5284 
The gradient accumulation to global field only happens when reading it, i.e., GlobalLoadStmt. However, the store forwarding kills some GlobalLoadStmt so that the accumulations for the grad of intermediate variable are skipped (as shown in #5284).

This PR makes the store forwarding skipped when dealing with GlobalLoadStmt who reads a global field with grad.

### DIFF
--- a/taichi/ir/control_flow_graph.cpp
+++ b/taichi/ir/control_flow_graph.cpp
@@ -275,7 +275,19 @@ bool CFGNode::store_to_load_forwarding(bool after_lower_access) {
       }
     } else if (auto global_load = stmt->cast<GlobalLoadStmt>()) {
       if (!after_lower_access) {
-        result = get_store_forwarding_data(global_load->src, i);
+        if (auto global_ptr = global_load->src->cast<GlobalPtrStmt>()) {
+          TI_ASSERT(global_ptr->width() == 1);
+          auto snodes = global_ptr->snodes;
+          if (!snodes[0]->has_adjoint()) {
+            // No adjoint SNode. Do the store forwarding
+            // otherwise skipping the store forwarding to keep the global load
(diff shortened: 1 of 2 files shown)


taichi/ir/control_flow_graph.cpp
```diff
@@ -275,7 +275,20 @@ bool CFGNode::store_to_load_forwarding(bool after_lower_access) {
       }
     } else if (auto global_load = stmt->cast<GlobalLoadStmt>()) {
       if (!after_lower_access) {
-        result = get_store_forwarding_data(global_load->src, i);
+        if (auto global_ptr = global_load->src->cast<GlobalPtrStmt>()) {
+          TI_ASSERT(global_ptr->width() == 1);
+          auto snodes = global_ptr->snodes;
```
ailzhang, line 280:
```suggestion
          auto& snodes = global_ptr->snodes;
```

taichi/ir/control_flow_graph.cpp
```diff
@@ -275,7 +275,20 @@ bool CFGNode::store_to_load_forwarding(bool after_lower_access) {
...
+          if (!snodes[0]->has_adjoint()) {
+            // No adjoint SNode. Do the store forwarding
+            // otherwise skipping the store forwarding to keep the global
+            // load chain,
+            // so that the grad of intermidiate variable can be computed
+            // by GlobalLoadStmt
+            result = get_store_forwarding_data(global_load->src, i);
```
ailzhang, line 287: can we simplify the logic here so that we write `get_store_forwarding_data(global_load->src, i)` only once?

ailzhang (commented):
LGTM!

PR #5316, opened 2022-07-03
[AOT] Save AOT module artifacts as zip archive

PENGUINLIONG:
AOT modules atm are saved as a list of files which leads to extra complexity importing to Unity. Unity only allows the user to load one file at a time via its provided Assets APIs for consistent cross-platform behavior; and files of different extensions will be imported as different C# types, making it frustratingly difficult to write glue codes.

This PR provides a (maybe temporary) solution to pack all baked files and metadata to a single zip. You only need to apply a minor change to shift from the existing `save` API to the zip API.

```diff
mod = ti.aot.Module(ti.vulkan)
mod.add_kernel(fractal,
                template_args={
                    'canvas': canvas,
                })
- mod.save("Assets/Resources/TaichiModules/module_directory", '')
+ mod.archive("Assets/Resources/TaichiModules/module_file.tcm")
```

In the generated zip archive (`.tcm`), you will find the following files:

- `__content__` - A list of names of kernels and compute graphs available in this module archive;
- `__version__` - Taichi compiler version (e.g. `1.0.4` atm);
- Compiled artifacts (`*.tcb`, `*.spv`, ...)

### DIFF
--- a/python/taichi/aot/module.py
+++ b/python/taichi/aot/module.py
@@ -1,5 +1,9 @@
 from contextlib import contextmanager
+from glob import glob
 from pathlib import Path, PurePosixPath
+from random import randint
+from shutil import rmtree
+from zipfile import ZipFile
 
 from taichi.aot.utils import (produce_injected_args,
                               produce_injected_args_from_template)
@@ -88,6 +92,7 @@ class Module:
         rtm = impl.get_runtime()
         rtm._finalize_root_fb_for_aot()
         self._aot_builder = rtm.prog.make_aot_module_builder(arch)
+        self._content = []
 
     def add_field(self, name, field):
         """Add a taichi field to the AOT module.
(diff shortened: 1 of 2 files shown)


python/taichi/aot/module.py
```diff
@@ -193,3 +203,35 @@ def save(self, filepath, filename):
...
+          filepath (str): path to the stored archive of aot artifacts, MUST
+            end with `.tcm`.
+        """
+        assert filepath.endswith(".tcm"), \
+            "AOT module artifact archive must ends with .tcm"
+        tcm_path = Path(filepath).absolute()
+        assert tcm_path.parent.exists(), "Output directory doesn't exist"
+
+        temp_dir = None
+        while temp_dir is None:
```
ailzhang, line 219: use `tempfile.mkdtemp`?

ailzhang (commented):
LGTM!
nit: add a test in test_aot.py?

ailzhang (approved):
LGTM!

PR #5317, opened 2022-07-04
[lang] Texture image load store support

bobcao3:
Related issue = #5109

### DIFF
--- a/python/taichi/lang/_texture.py
+++ b/python/taichi/lang/_texture.py
@@ -60,6 +60,47 @@ class TextureSampler:
                                with_runtime_context=False)
         return ti.Vector([r, g, b, a])
 
+class RWTextureAccessor:
+    def __init__(self, ptr_expr, num_dims) -> None:
+        self.ptr_expr = ptr_expr
+        self.num_dims = num_dims
+
+    @taichi_scope
+    def load(self, index):
+        args_group = ()
+        if self.num_dims == 1:
+            args_group = impl.make_expr_group(index.x)
+        elif self.num_dims == 2:
+            args_group = impl.make_expr_group(index.x, index.y)
+        elif self.num_dims == 3:
+            args_group = impl.make_expr_group(index.x, index.y, index.z)
+        v = _ti_core.make_texture_op_expr(_ti_core.TextureOpType.load,
+                                          self.ptr_expr, args_group)
+        r = impl.call_internal("composite_extract_0",
+                               v,
+                               with_runtime_context=False)
+        g = impl.call_internal("composite_extract_1",
+                               v,
+                               with_runtime_context=False)
+        b = impl.call_internal("composite_extract_2",
+                               v,
+                               with_runtime_context=False)
+        a = impl.call_internal("composite_extract_3",
+                               v,
+                               with_runtime_context=False)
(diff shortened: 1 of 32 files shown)


taichi/ir/statements.h
```diff
@@ -1453,13 +1453,39 @@ class TexturePtrStmt : public Stmt {
  public:
   Stmt *arg_load_stmt{nullptr};
   int dimensions{2};
+  bool is_storage{false};
```
k-ye, line 1456: nit: Put this to the end so that we avoid paddings

taichi/ir/stmt_op_types.h
```diff
@@ -84,7 +84,13 @@ enum class SNodeOpType : int {
 
 std::string snode_op_type_name(SNodeOpType type);
 
-enum class TextureOpType : int { sample_lod, fetch_texel, undefined };
+enum class TextureOpType : int {
+  sample_lod,
+  fetch_texel,
+  load,
+  store,
+  undefined
```
k-ye, line 92: I think the convention is to put `undefined` in the first place? (Also we should rename this to `kXx` style later)

taichi/rhi/device.h
```diff
@@ -148,6 +148,13 @@ class ResourceBinder {
     TI_NOT_IMPLEMENTED
   }
 
+  virtual void rw_image(uint32_t set,
+                        uint32_t binding,
+                        DeviceAllocation alloc,
+                        int lod) {
+    TI_NOT_IMPLEMENTED
```
k-ye, line 155: This can be `=0` since you have overriden each device impl.

python/taichi/lang/_texture.py
```diff
@@ -61,6 +61,53 @@ def fetch(self, index, lod):
         return ti.Vector([r, g, b, a])
 
 
+class RWTextureAccessor:
```
k-ye, line 64: @strongoier @jim19930609 any concerns here for C++ vec types?

taichi/rhi/device.h
bobcao3: This is intended for devices that do not support this doesn't need to have a override functions

taichi/inc/frontend_statements.inc.h
```diff
@@ -13,3 +13,4 @@ PER_STATEMENT(FrontendSNodeOpStmt)  // activate, deactivate, append, clear
 PER_STATEMENT(FrontendAssertStmt)
 PER_STATEMENT(FrontendFuncDefStmt)
 PER_STATEMENT(FrontendReturnStmt)
+PER_STATEMENT(FrontendTextureOpStmt)
```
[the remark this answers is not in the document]
bobcao3: How do we make sure that image store works? It does not return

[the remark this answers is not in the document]
strongoier: Wrap the expr with `expr_init()`, just like https://github.com/taichi-dev/taichi/blob/31f46dd551fdd77a5e22313d0c2dc4d8942916b4/python/taichi/lang/impl.py#L929-L932

k-ye (approved):
LGTM, thanks!

PR #5327, opened 2022-07-05
[ci] Run release tests

feisuzhu:
Related issue = #

### DIFF
--- a/.github/workflows/scripts/unix_test.sh
+++ b/.github/workflows/scripts/unix_test.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -ex
 
+export PYTHONUNBUFFERED=1
+
 check_in_docker() {
     # This is a temporary solution to detect in a docker, but it should work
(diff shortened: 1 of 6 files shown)


ailzhang (changes requested):
Note that release tests didn't actually run on CI due to wrong syntax :P 
```
2022-07-13T07:00:01.5208256Z + echo 'wanted archs: cpu,cuda,vulkan,opengl'
2022-07-13T07:00:01.5208481Z wanted archs: cpu,cuda,vulkan,opengl
2022-07-13T07:00:01.5208671Z + '[' '' -eq 1 ']'
2022-07-13T07:00:01.5208977Z /home/dev/unix_test.sh: line 53: [: : integer expression expected
```

ailzhang (approved):
This is awesome!! 
Btw to double check: once this PR is merged, would you mind sending a draft PR that breaks any of the examples to make sure this CI tests can capture that  as expected? ;)

PR #5331, opened 2022-07-05
[Lang] [type] Refine SNode with quant 3/n: Turn bit_vectorize into an on/off switch

strongoier:
Related issue = #4857

`bit_vectorize` is currently an ad-hoc configuration function specifying how many bits are vectorized together, which leads to confusing API explanations like `bit_vectorize(1)` means off while `bit_vectorize(32)` means on. As the vectorization unit is always the physical type, this PR turns `bit_vectorize()` into an on/off switch, and all internal configuration parameters are renamed to `is_bit_vectorized`.


tests/python/test_bit_array_vectorization.py
```diff
@@ -32,7 +32,7 @@ def init():
 
     @ti.kernel
     def assign_vectorized():
-        get_runtime().prog.current_ast_builder().bit_vectorize(32)
+        get_runtime().prog.current_ast_builder().bit_vectorize()
```
ailzhang, line 35: OOC: is this API user-facing? can we do `ti.loop_config(bit_vectorize=True)` instead? :P

ailzhang (approved):
LGTM!

tests/python/test_bit_array_vectorization.py
strongoier: Yep. This is planned in the upcoming PR :-)

PR #5335, opened 2022-07-05
[AOT] Fixed certain crashes in C-API

PENGUINLIONG:
This PR fixed certain crashes with C-API.
- Initialization call order for imported Vulkan runtime was wrong;
- Named NDArray argument map allocation were not stable.


k-ye (approved):

ailzhang (approved):
OOC: would you mind elaborating how does this PR fix "Named NDArray argument map allocation were not stable." :P?

PENGUINLIONG:
> OOC: would you mind elaborating how does this PR fix "Named NDArray argument map allocation were not stable." :P? 

There is potential reallocation of `std::vector` when new items are pushed into `ndarrays` and the `NDArray*`s in `arg_map` might point to somewhere already freed

PENGUINLIONG:
jeez i just realized there is a commit i failed to push 😨

PR #5339, opened 2022-07-06
[autodiff] Handle field accessing by zero for forward mode

erizmr:
Related issue = #5055 
To handle the case similar to `ti.root.dense(ti.i, 1).place(a, b, a.dual, b.dual)`. In this case, the field shape is one but not none comparing to `a = ti.field(float, shape=())`.


ailzhang (approved):
LGTM!

PR #5346, opened 2022-07-06
[example] Fix simulation/waterwave.py:update race condition

feisuzhu:
Do not update and read neighbors of same field simultaneously


ailzhang (approved):
Nice catch!!

Issue #5252, opened 2022-06-24
Remove AsyncTaichi and move the parallel compilation part to the codegen


lin-hitonami:
**Concisely describe the proposed feature**
Currently, AsyncTaichi is out of maintenance, and it makes us hard to add new features in the LLVM backend. Therefore, we decide to remove it, and move the parallel compilation part to the codegen.

PR #5251, opened 2022-06-24
[llvm] [refactor] Add parallel compilation for CPU backend

lin-hitonami:
Related issue = #5252 
After the refactoring, CodeGenLLVM only compiles an offloaded task to an LLVM module, and KernelCodeGen compiles different offloaded tasks into different LLVM modules in parallel. 

This PR roughly does things as follows:
1. Extract `ParallelExecutor` from AsyncTaichi.
2. Remove tests of AsyncTaichi.
3. Remove the pointer of the codegen from `OffloadedTask` and let it replace `LlvmOfflineCache::OffloadedTaskCacheData`.
4. Rename `CodeGenLLVM::CompiledData` to `LLVMCompiledData` and let it replace `ModuleGenValue`.
5. Move `maybe_read_compilation_from_cache` and `cache_module` from `CodeGenLLVM` to `KernelCodeGen`.
6. Make `ModuleToFunctionConverter` a abstract class, and CPU and CUDA backend have own versions of it.
7. Remove `CodeGenLLVM::gen` function and move the logic to `KernelCodeGen::codegen` function.
8. Let `KernelCodeGen::modulegen` be a virtual function, it calls `CodeGenLLVM::run_compilation`  to generates a module for a offloaded task/kernel.
9. Call `KernelCodeGen::modulegen` in `KernelCodeGen::codegen` in parallel for `n` offloaded task, generating `n` modules in total.
10. Let `ModuleToFunctionConverter::convert` receive multiple modules.

PR #5351, opened 2022-07-07
[refactor] (Decomp of #5251 1/n) Move ParallelExecutor out of async engine

lin-hitonami:
Related issue = #5252 
Decomposition of #5251 

**Feature added: the executor runs serially and synchronizedly when num_threads=0, and no new threads are created.**


strongoier (approved):
LGTM!

PR #5353, opened 2022-07-07
[llvm] [refactor] (Decomp of #5251 2/n) Make modulegen a virtual function and let LLVMCompiledData replace ModuleGenValue

lin-hitonami:
Related issue = #5252
Decomposition of #5251

### DIFF
--- a/taichi/codegen/codegen.h
+++ b/taichi/codegen/codegen.h
@@ -1,9 +1,14 @@
 // Driver class for kernel code generators.
 
 #pragma once
-
+#include "taichi/ir/ir.h"
 #include "taichi/program/program.h"
-
+#ifdef TI_WITH_LLVM
+#include "llvm/IR/Module.h"
+#include "taichi/codegen/llvm/codegen_llvm.h"
+#include "taichi/runtime/llvm/launch_arg_info.h"
+#include "taichi/codegen/llvm/llvm_codegen_utils.h"
+#endif
 TLANG_NAMESPACE_BEGIN
 
 class KernelCodeGen {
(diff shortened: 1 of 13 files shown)


taichi/codegen/codegen.h
```diff
@@ -22,6 +27,11 @@ class KernelCodeGen {
                                                Stmt *stmt = nullptr);
 
   virtual FunctionType codegen() = 0;
+  virtual LLVMCompiledData modulegen(
```
strongoier, line 30: Will this work properly when `TI_WITH_LLVM` is off?

lin-hitonami: No, I forgot to add the #ifdef. However, IIUC all derived classes of KernelCodeGen use LLVM, then why do we only place this function inside #ifdef instead of place the whole KernelCodeGen inside it?

strongoier: lol if possible we can place the whole KernelCodeGen inside.

lin-hitonami: Maybe we can move codegen.h/cpp into the `llvm/` directory after implementing the parallel compilation function. Let's just leave it here for now.

taichi/codegen/llvm/codegen_llvm.h
```diff
@@ -44,6 +44,11 @@ class FunctionCreationGuard {
   ~FunctionCreationGuard();
 };
 
+struct LLVMCompiledData {
```
jim19930609, line 47: We probably wanna merge this class with "OffloadedTaskCacheData" (in some future PRs maybe)?

jim19930609 (approved):
LGTM!

taichi/codegen/llvm/codegen_llvm.h
jim19930609: Just noticed the following PR. plz ignore this one.

PR #5359, opened 2022-07-07
[refactor] Rename the confused needs_grad

erizmr:
Related issue = #

### DIFF
--- a/taichi/python/export_lang.cpp
+++ b/taichi/python/export_lang.cpp
@@ -970,7 +970,6 @@ void export_lang(py::module &m) {
   });
 
   m.def("test_throw", [] { throw IRModified(); });
-  m.def("is_real", is_real);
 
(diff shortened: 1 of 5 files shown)


ailzhang (commented):
how about we just use `is_real`? :P

taichi/python/export_lang.cpp
```diff
@@ -970,7 +970,7 @@ void export_lang(py::module &m) {
   });
 
   m.def("test_throw", [] { throw IRModified(); });
-  m.def("needs_grad", needs_grad);
+  m.def("is_real", is_real);
```
strongoier, line 973: IIRC `is_real` is already exported :-)

erizmr: Let me remove the redundant one.

strongoier (approved):
LGTM!

PR #5360, opened 2022-07-07
[test] Check repeated arg names in cgraph

ailzhang:
* #5363
* #5362
* #5361
* __->__ #5360

Arg name should be an unique identifier in the same graph so adding a
test to prevent two different args having the same name.

### DIFF
--- a/taichi/program/graph_builder.cpp
+++ b/taichi/program/graph_builder.cpp
@@ -37,6 +37,13 @@ GraphBuilder::GraphBuilder() {
 
 Node *GraphBuilder::new_dispatch_node(Kernel *kernel,
                                       const std::vector<aot::Arg> &args) {
+  for (const auto &arg : args) {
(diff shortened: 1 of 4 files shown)


taichi/program/graph_builder.cpp
```diff
@@ -37,6 +37,13 @@ GraphBuilder::GraphBuilder() {
 
 Node *GraphBuilder::new_dispatch_node(Kernel *kernel,
                                       const std::vector<aot::Arg> &args) {
+  for (const auto &arg : args) {
+    if (all_args_.find(arg.name) != all_args_.end()) {
+      TI_ASSERT(all_args_[arg.name] == arg);
```
turbo0628, line 42: nit: the user might get confused about the error message. Use a TI_ERROR instead?

turbo0628 (approved):
LGTM!

PR #5366, opened 2022-07-07
[autodiff] Support unary ops for forward mode

erizmr:
Related issue = #5055

### DIFF
--- a/taichi/transforms/auto_diff.cpp
+++ b/taichi/transforms/auto_diff.cpp
@@ -1115,10 +1115,32 @@ class MakeDual : public ADTransform {
       accumulate(stmt, mul(cos(stmt->operand), dual(stmt->operand)));
     } else if (stmt->op_type == UnaryOpType::cos) {
       accumulate(stmt, negate(mul(sin(stmt->operand), dual(stmt->operand))));
+    } else if (stmt->op_type == UnaryOpType::tan) {
+      TI_NOT_IMPLEMENTED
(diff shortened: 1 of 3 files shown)


taichi/transforms/auto_diff.cpp
```diff
@@ -1115,10 +1115,32 @@ class MakeDual : public ADTransform {
       accumulate(stmt, mul(cos(stmt->operand), dual(stmt->operand)));
     } else if (stmt->op_type == UnaryOpType::cos) {
       accumulate(stmt, negate(mul(sin(stmt->operand), dual(stmt->operand))));
+    } else if (stmt->op_type == UnaryOpType::tan) {
+      TI_NOT_IMPLEMENTED
```
ailzhang, line 1119: OOC: why is this not implemented :P?  nit: semicolon at the end ;)

ailzhang (commented):
Shall we add a few test cases in this PR as well?

taichi/transforms/auto_diff.cpp
erizmr: Sure. Adding test cases (and the semicolon ).

The reason for that is the derivative of `tan (x)` is `1/ cos^2(x)`, there are many singular points for this function which causing nan. lol

ailzhang: aha would you mind adding this as a comment in the code? ;)

ailzhang (approved):
LGTM!

PR #5369, opened 2022-07-08
[Lang] Implement assignment by slicing

### LINKED ISSUE #4677: Failed to assign row values by matrix slicing
**Describe the bug**
Failed to assign a transposed vector to a matrix row by using slicing m[i,:]

**To Reproduce**

```py
import taichi as ti

@ti.kernel
def test_kernel():
    vertlist = ti.Matrix([[0., 0., 0.] for i in range(3)])
    p = ti.Vector([1, 2, 3])
    vertlist[1, :] = p.transpose()
    print("vertlist assign by slicing", vertlist)

    vertlist[1, 0] = p[0]
    vertlist[1, 1] = p[1]
    vertlist[1, 2] = p[2]

    print("assign by indexing", vertlist)

ti.init(arch=ti.cpu)
test_kernel()

```

**Log/Screenshots**

```
➜  TaichiSLAM git:(main) ✗ python tests/test.py
[Taichi] version 0.9.3, llvm 10.0.0, commit c902db62, linux, python 3.8.10
[Taichi] Starting on arch=x64
vertlist assign by slicing [[0.000000, 0.000000, 0.000000], [0.000000, 0.000000, 0.000000], [0.000000, 0.000000, 0.000000]]
assign by indexing [[0.000000, 0.000000, 0.000000], [1.000000, 2.000000, 3.000000], [0.000000, 0.000000, 0.000000]]
```

**Additional comments**

Allowing slicing will accelerate the speed of development.

AD1024:
Related issue = close #4677

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -116,23 +116,10 @@ class ASTTransformer(Builder):
             return ASTTransformer.build_assign_basic(ctx, node_target, values,
                                                      is_static_assign)
         node_target.value = build_stmt(ctx, node_target.value)
-        targets = list(impl.subscript(node_target.value, *indices, get_ref=True))
-        flatten_targets = []
-        for t in targets:
-            if isinstance(t, list):
-                flatten_targets += t
-            else:
-                flatten_targets.append(t)
-        targets = flatten_targets
+        targets = impl.subscript(node_target.value, *indices, get_ref=True)
         if not isinstance(values, (matrix.Matrix, collections.abc.Sequence)):
             raise ValueError(f"Cannot unpack type: {type(values)}")
-        if isinstance(values, matrix.Matrix):
-            values = list(values.entries)
-        if len(values) != len(targets):
-            raise TaichiSyntaxError(
-                "The number of targets is not equal to value length")
-        for (lhs, rhs) in zip(targets, values):
-            lhs._assign(rhs)
+        targets._assign(values)
         return None
 
     @staticmethod
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -117,7 +117,11 @@ class ASTTransformer(Builder):
             else:
                 raise TaichiTypeError(f'{type(target)} cannot be subscripted')
         else:
-            ASTTransformer.build_assign_basic(ctx, node_target, values, is_static_assign)
+            ASTTransformer.build_assign_basic(ctx,
+                                              target,
+                                              values,
+                                              is_static_assign,
+                                              build_target=False)
 
     @staticmethod
     def build_assign_unpack(ctx, node_target, values, is_static_assign):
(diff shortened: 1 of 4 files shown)


python/taichi/lang/ast/ast_transformer.py
```diff
@@ -102,6 +103,26 @@ def build_Assign(ctx, node):
                                                is_static_assign)
         return None
 
+    @staticmethod
+    def build_assign_slice(ctx, node_target, values, is_static_assign):
+        target = ASTTransformer.build_Subscript(ctx, node_target, get_ref=True)
+        if isinstance(node_target.value.ptr, Matrix):
```
jim19930609, line 109: Put `target = ASTTransformer.build_Subscript(ctx, node_target, get_ref=True)` in this scope so as to avoid adding "build_target=False" to `build_assign_basic`

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -139,7 +164,11 @@ def build_assign_unpack(ctx, node_target, values, is_static_assign):
         return None
 
     @staticmethod
-    def build_assign_basic(ctx, target, value, is_static_assign):
+    def build_assign_basic(ctx,
+                           target,
+                           value,
+                           is_static_assign,
+                           build_target=True):
```
jim19930609, line 171: `build_target` can be avoided with some rearrangements, refer to the comment above~

python/taichi/lang/matrix.py
```diff
@@ -198,7 +198,7 @@ def _get_slice(self, a, b):
             b = [b]
         else:
             b = range(b.start or 0, b.stop or self.m, b.step or 1)
-        return Matrix([[self(i, j) for j in b] for i in a])
+        return Matrix([[self._get_entry(i, j) for j in b] for i in a])
```
jim19930609, line 201: Not sure why we need this line of change? Is it trying to support slicing _PyScopeMatrixImpl at ti_scope?

AD1024: No, it's just a bug fix to this function (it complained about `_PyScopeMatrixImpl` is not callable)

python/taichi/lang/ast/ast_transformer.py
AD1024: A follow up question is that `node_target.value.ptr` is not populated before calling `build_Subscript`

jim19930609: You can probably fetch the variable with: 

2022-07-12 10-48-48 的屏幕截图

It's gonna show the object's type:
2022-07-12 10-49-04 的屏幕截图

jim19930609: `node_target.value.id` is parsed from the AST:

2022-07-12 10-50-35 的屏幕截图

AD1024: This sounds good. My only concern is that this approach won't work for expressions that subscripts are not applied to an identifier. For example:
```python3
@ti.kernel
def test_kernel():
    vertlist = ti.Matrix([[0., 0., 0.] for i in range(3)])
    p = ti.Vector([1, 2, 3])
    mat_dict = {
        'x': vertlist
    }
    mat_dict['x'][1, :] = p.transpose()
    print(mat_dict['x'])
```

jim19930609: Good point! I guess we'll have to stick with the `build_target` solution then :p

jim19930609 (approved):
I think we can have this PR merged and file another PR to fix the same issue with MatrixField/MatrixNdarray. 

Btw can you add some unit tests to guard this fix? Thanks!

AD1024:
> I think we can have this PR merged and file another one to fix the same issue with MatrixField/MatrixNdarray. 
> 
> Btw can you add some unit tests to guard this fix? Thanks!

Got it. Another PR sounds good to me.
Will add a test for this.

jim19930609 (approved):

PR #5372, opened 2022-07-08
[llvm] [refactor] (Decomp of #5251 6/n) Let ModuleToFunctionConverter support multiple modules

lin-hitonami:
Related issue = #5252 
Decomposition of #5251

Made ModuleToFunctionConverter an abstract class.

Currently only CPU backend supports multiple modules, the implementation of CUDA backend only supports a single module.

### DIFF
--- a/taichi/codegen/cuda/codegen_cuda.cpp
+++ b/taichi/codegen/cuda/codegen_cuda.cpp
@@ -41,9 +41,9 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
     auto *llvm_prog = get_llvm_program(kernel->program);
     CUDAModuleToFunctionConverter converter{tlctx,
                                             llvm_prog->get_runtime_executor()};
-
-    return converter.convert(this->kernel, std::move(compiled_res.module),
-                             std::move(compiled_res.tasks));
+    std::vector<LLVMCompiledData> data;
+    data.push_back(std::move(compiled_res));
+    return converter.convert(this->kernel, std::move(data));
   }
 
   llvm::Value *create_print(std::string tag,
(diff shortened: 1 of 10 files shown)


taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -41,9 +41,9 @@ class CodeGenLLVMCUDA : public CodeGenLLVM {
     auto *llvm_prog = get_llvm_program(kernel->program);
     CUDAModuleToFunctionConverter converter{tlctx,
                                             llvm_prog->get_runtime_executor()};
-
-    return converter.convert(this->kernel, std::move(compiled_res.module),
-                             std::move(compiled_res.tasks));
+    std::vector<LLVMCompiledData> data;
+    data.push_back(std::move(compiled_res));
```
jim19930609, line 45: emplace_back

taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -738,11 +738,14 @@ FunctionType CodeGenCUDA::codegen() {
   return CodeGenLLVMCUDA(kernel, ir).gen();
 }
 
+#ifdef TI_WITH_LLVM
```
jim19930609, line 741: Think this TI_WITH_LLVM macro is redundant, since the current file is located under `taichi/codegen/cuda/`. 

Can you try removing this macro and see if it still compiles?

taichi/codegen/cuda/codegen_cuda.cpp
lin-hitonami: IIRC these two are equivalent when moving...

taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -738,11 +738,14 @@ FunctionType CodeGenCUDA::codegen() {
...
+
 FunctionType CUDAModuleToFunctionConverter::convert(
     const std::string &kernel_name,
     const std::vector<LlvmLaunchArgInfo> &args,
-    std::unique_ptr<llvm::Module> mod,
-    std::vector<OffloadedTask> &&tasks) const {
+    std::vector<LLVMCompiledData> &&data) const {
+  auto &mod = data[0].module;
```
jim19930609, line 747: Should be a for loop? 

```
for(auto& d : data) {
  auto& mod = d.module;
  ...
}
```

lin-hitonami: This one is fake for now. I haven't added the support for multiple modules for CUDA. I'll add it later.

taichi/codegen/cuda/codegen_cuda.h
```diff
@@ -22,21 +22,21 @@ class CodeGenCUDA : public KernelCodeGen {
   FunctionType codegen() override;
 };
 
+#ifdef TI_WITH_LLVM
```
jim19930609, line 25: ditto

jim19930609 (approved):
LGTM!

PR #5373, opened 2022-07-08
[llvm] (Decomp of #5251 7/n) Change the way to record the time of offline cache

lin-hitonami:
Related issue = #5252
Decomposition of #5251


jim19930609 (approved):
LGTM!

Issue #5276, opened 2022-06-28
[Backend] DirectX (D3D12, DXIL) Backend


python3kgae:
**Concisely describe the proposed feature**
I would like to add a DirectX 12 backend to the compiler so that I can use new features which requires shader model >= 6.0.

**Describe the solution you'd like (if any)**
The plan is in 3 steps.
1. Add llvm 15 support which is required for generate DXIL.
    Code change will be guarded with something like "#ifdef TI_LLVM_15" if llvm 15 require different implementation.
    CMake option TI_LLVM_15 will be added to control it and OFF by default. 

2. Add AOT compile first which could be tested by add c++ directx12 test without backend.
    DXIL support in llvm 15 is still in early state. It will be a long journey to support all features.
    CodeGenLLVM will be used as base class for llvm ir generation when codegen. Then some llvm passes will be added to prepare for DXIL generation and finally use DirectX backend in llvm 15 to generate DXIL.
    A kernel which has more than 1 tasks is a new thing for DXIL, no plan to support this feature at first.

3. Add DirectX 12 backend.
    No an expert on this part :(
    Maybe start support for simple cases first.

**Additional comments**
I'm working on DXIL generation on llvm 15.  Targeting language other then HLSL to DXIL will help the design of DXIL generation.
This is personal work, not official contribution (which requires examine and approve).

k-ye:
Hi @python3kgae ,

Thanks for proposing this, it looks really attractive!

> Add llvm 15 support which is required for generate DXIL.

Note that we are (conservatively) planning to upgrade to LLVM-12. It's probably fine for us to directly go to 15 as well. One thing we've recently noticed, is the installation speed from `apt-get` for `clang-13` (and likely beyond) is super slow. This could potentially hurt our external contributors...

(Ive followed https://apt.llvm.org/ for `clang-13` installation, not sure if you have better suggestions here.)

> Then some llvm passes will be added to prepare for DXIL generation and finally use DirectX backend in llvm 15 to generate DXIL.

I see. I'm not sure how feasible it is to share the generated LLVM IR between CUDA and DXIL. Hopefully this is not a problem :-) (Taking SPIR-V as an example. Although it claims to support both Vulkan and OpenCL, they use very different Execution Model, making it nontrivial to share the same SPIR-V between these two APIs).

> No an expert on this part :( Maybe start support for simple cases first.

If we can have a working codegen, D3D12 runtime API shouldn't be too hard to sort out :-)

[...]

PR #5381, opened 2022-07-10
[llvm] Allow using clang 15 for COMPILE_LLVM_RUNTIME

python3kgae:
Allow using clang 15 for COMPILE_LLVM_RUNTIME.
Remove use of getPointerElementType.

This just allow using clang 15, clang 10 is still used. Not switch yet.

Related issue = #5276


ailzhang (approved):
LGTM！

Issue #5289, opened 2022-06-29
[ci] Implement timeout for single test case


feisuzhu:
Occasionally single test case can stuck whole pipeline for several hours.
Should treat long running cases (e.g. >10min) as failure.

PR #5384, opened 2022-07-11
[ci] Add per test hard timeout limit

feisuzhu:
Related issue = #5289

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -6,3 +6,4 @@ numpy
 autograd
 requests==2.26
 matplotlib
+cffi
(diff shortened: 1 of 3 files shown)


lin-hitonami:
Why not just use the pytest-timeout package?

feisuzhu:
> Why not just use the `pytest-timeout` package?

`pytest-timeout` handles timeout in python code, which requires GIL to be held, which isn't always feasible, nearly all taichi operations don't release GIL (with `Kernel.__call__` as a notable exception).

The code `pytest_hardtle.py` is actually a modified version of `pytest-timeout`

qiao-bo (approved):
Thanks!

PR #5388, opened 2022-07-11
[gui] [vulkan] Surpport for python users to control the start index and count number of particles & meshes data.

Morcki:
Usage:
```python
# For particles
scene.particles(center, radius, 
vertex_offset = user_defined_first_vertex_index, 
vertex_count = user_defined_vertex_count)

# For mesh
# 1. with indices
scene.mesh(vertices, indices, 
index_offset  = user_defined_first_indices_index, 
index_count   = user_defined_index_count, 
vertex_offset = user_defined_vertex_offset)

# 2. without indices
scene.mesh(vertices, 
vertex_offset = user_defined_first_vertex_index, 
vertex_count = user_defined_vertex_count)
```


python/taichi/ui/scene.py
```diff
@@ -120,18 +137,25 @@ def mesh(self,
...
+                        two_sided, index_count, index_offset, vertex_count,
+                        vertex_offset)
 
     def particles(self,
                   centers,
                   radius,
                   color=(0.5, 0.5, 0.5),
-                  per_vertex_color=None):
+                  per_vertex_color=None,
+                  vertex_offset: int = 0,
```
YuCrazing, line 157: Maybe it is better to change the prefix `vertex_*` to `index_*`.

python/taichi/ui/scene.py
```diff
@@ -112,6 +116,19 @@ def mesh(self,
...
+                index of the first vertex to draw.
+            vertex_count: int type(ohterwise float type will be floored to int),
+                only avaliable when `indices` is not provided, which is the number
+                of vertices to draw.
+            index_offset: int type(ohterwise float type will be floored to int),
+                only avaliable when `indices` is provided, which is the base index
+                within the index buffer.
+            index_count: int type(ohterwise float type will be floored to int),
+                only avaliable when `indices` is provided, which is the the number
+                of vertices to draw.
```
PENGUINLIONG, line 131: Maybe add a line or two to warn the users that the change is provisional. We have to sort out how it interacts with normals and vertex colors. Or simply add them all?

python/taichi/ui/scene.py
```diff
@@ -141,16 +165,21 @@ def particles(self,
                 values. If `per_vertex_color` is provided, this is ignored.
             per_vertex_color (Tuple[float]): a taichi 3D vector field, where each
                 element indicate the RGB color of a particle.
-            two_sided (bool): whether or not the triangles should be able to be
```
PENGUINLIONG, line 144: `two_sided` deleted?

taichi/ui/backends/vulkan/renderable.cpp
```diff
@@ -49,22 +49,45 @@ void Renderable::update_data(const RenderableInfo &info) {
...
   } else {
     num_indices = 1;
+    draw_num_indices = 1;
+    draw_first_indices = 0;
   }
 
   config_.vertices_count = num_vertices;
   config_.indices_count = num_indices;
 
+  if (info.has_user_customized_draw) {
```
PENGUINLIONG, line 79: Can we remove this flag and assign to `config_.draw_xxx` directly?

PENGUINLIONG (changes requested):

python/taichi/ui/scene.py
Morcki: the note of two_sided is redundant cause there is not two_sided args in Scene.particles method. Maybe I think somebody misinput the note from Scene.mesh which include two_sided arg.

taichi/ui/backends/vulkan/renderable.cpp
Morcki: No, I didn't use this at the first time, but I found that other part such as line, triangle, circle and so on were not need for this function at present. So I must use this bool value to close this control for they all were draw as renderable. (May lead to bugs when draw lines, triangles and circles)

PENGUINLIONG (approved):
lgtm :)

PR #5393, opened 2022-07-12
[llvm] [aot] Add LLVM to CAPI part 1: Implemented capi::LlvmRuntime class

jim19930609:
Related issue = #4800

OffloadedTask is one of the dependent for LlvmOfflineCache. This class used to be declared in `codegen_llvm.h`, which essentially pollutes `llvm_offline_cache.h` since the entire `codegen_llvm.h` has to be included (together with all the sub-dependencies).

Added a new file `taichi/codegen/llvm/llvm_compiled_data.h` to store OffloadedTask and LLVMCompiledData, so as to purify the dependencies for `llvm_offline_cache.h`.


c_api/src/taichi_llvm_impl.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace capi {
+
+LlvmRuntime::LlvmRuntime(taichi::Arch arch) : Runtime(arch) {
+  taichi::lang::CompileConfig cfg;
+  cfg.arch = arch;
+  executor_ = std::make_unique<taichi::lang::LlvmRuntimeExecutor>(cfg, nullptr);
+
+  taichi::lang::Device *compute_device = executor_->get_compute_device();
+  std::unique_ptr<taichi::lang::MemoryPool> memory_pool =
```
PENGUINLIONG, line 15: Is `memory_pool` safe to destruct here?

taichi/runtime/llvm/llvm_offline_cache.h
```diff
@@ -8,7 +8,8 @@
 #include "taichi/runtime/llvm/launch_arg_info.h"
 #include "taichi/program/kernel.h"
 #include "taichi/util/io.h"
-#include "taichi/codegen/llvm/codegen_llvm.h"
+#include "taichi/codegen/llvm/llvm_compiled_data.h"
```
PENGUINLIONG, line 11: Reference to `codegen` features from the runtime-only C-API?

PENGUINLIONG (changes requested):

taichi/runtime/llvm/llvm_offline_cache.h
jim19930609: lol, OfflineCache is like sth in between (runtime loads all the compiled kernel/field information from it) and we just happened to put it inside runtime directory. 

However, OfflineCache also need to store the compiled information at the end of compilation phase, which is why it has to include a header file from `codegen` directory.

PENGUINLIONG:
BTW, I just noticed that `llvm_offline_cache.h` is not guarded by `TI_WITH_LLVM` and some build configs fail to compile.

PENGUINLIONG (approved):

jim19930609:
> BTW, I just noticed that `llvm_offline_cache.h` is not guarded by `TI_WITH_LLVM` and some build configs fail to compile.

Ah my bad, thanks!

PR #5394, opened 2022-07-12
[llvm] (Decomp of #5251 11/n) Enable parallel compilation on CPU backend

lin-hitonami:
Related issue = #5252 
Decomposition of #5251


jim19930609 (approved):
LGTM!

PR #5395, opened 2022-07-12
[autodiff] Add test for ternary operators in reverse mode autodiff

erizmr:
Currently, there is no test for ternary operators in reverse mode autodiff. This PR adds a test for it.

### DIFF
--- a/tests/python/test_ad_basics.py
+++ b/tests/python/test_ad_basics.py
@@ -281,9 +281,9 @@ def test_select():
     func.grad()
     for i in range(N):
         if i % 2:
-            loss[i] = i
+            assert loss[i] == i
         else:
-            loss[i] = -i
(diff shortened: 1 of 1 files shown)


tests/python/test_ad_basics.py
```diff
@@ -259,6 +259,35 @@ def test_pow_f64(tifunc, npfunc):
...
+    @ti.kernel
+    def func():
+        for i in range(N):
+            loss[i] += ti.select(i % 2, x[i], y[i])
+
+    func()
+    func.grad()
+    for i in range(N):
+        if i % 2:
+            loss[i] = i
```
ailzhang, line 284: OOC: why do we assign to loss here after running the grad kernel?

ailzhang (approved):
Thanks!

PR #5397, opened 2022-07-12
[llvm] Avoid creating new LLVM contexts when updating struct module

lin-hitonami:
Related issue = #5252

Types in modules imported from file/other contexts may be renamed to "original_type.xxx" when the type is imported multiple times. So, we need to insert a pointer cast when the parameter type of a function is renamed.

### DIFF
--- a/taichi/codegen/llvm/llvm_codegen_utils.cpp
+++ b/taichi/codegen/llvm/llvm_codegen_utils.cpp
@@ -10,8 +10,24 @@ std::string type_name(llvm::Type *type) {
   return type_name_str;
 }
 
+
+/*
+ * Determine whether two types are the same
+ * (a type is a renamed version of the other one) based on the type name
+ */
+bool is_same_type(llvm::Type *a, llvm::Type *b) {
+  if (a == b) {
+    return true;
+  }
+  auto a_name = type_name(a);
+  auto b_name = type_name(b);
+  int min_len = std::min(a_name.size(), b_name.size()) - 1;
+  return a_name.substr(0, min_len) == b_name.substr(0, min_len);
+}
+
 void check_func_call_signature(llvm::Value *func,
-                               std::vector<llvm::Value *> arglist) {
+                               std::vector<llvm::Value *> &arglist,
+                               llvm::IRBuilder<> *builder) {
   llvm::FunctionType *func_type = nullptr;
   if (llvm::Function *fn = llvm::dyn_cast<llvm::Function>(func)) {
     func_type = fn->getFunctionType();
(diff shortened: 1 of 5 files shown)


taichi/codegen/llvm/llvm_codegen_utils.cpp
```diff
@@ -6,12 +6,55 @@ namespace lang {
...
+  type->print(rso, /*IsForDebug=*/false, /*NoDetails=*/true);
   return type_name_str;
 }
 
+/*
+ * Determine whether two types are the same
+ * (required type is required renamed version of the other one) based on the
+ * type name. Check recursively if the types are function types.
+ */
+bool is_same_type(llvm::Type *required, llvm::Type *provided) {
```
jim19930609, line 18: Would be nice if we can make some comments on a summary of the renaming rule (examples are more than welcomed~)

taichi/codegen/llvm/llvm_codegen_utils.cpp
```diff
@@ -31,7 +74,17 @@ void check_func_call_signature(llvm::Value *func,
   for (int i = 0; i < num_params; i++) {
     auto required = func_type->getFunctionParamType(i);
     auto provided = arglist[i]->getType();
+    /*
+     * Types in modules imported from files which are not the first appearances
+     * are renamed "original_type.xxx", so we have to create a pointer cast to
+     * the type in the function parameter list when the types in the function
+     * parameter are renamed.
```
jim19930609, line 81: When importing a module from file, the imported llvm::Types can get conflict with the existing ones in the current llvm::Context. In such scenario, the imported types will get renamed from "original_type" to "original_type.xxx", making them separate types in essence. To recover the renamed types, a pointer cast to the original type was performed.

taichi/codegen/llvm/llvm_codegen_utils.cpp
```diff
@@ -31,7 +74,17 @@ void check_func_call_signature(llvm::Value *func,
   for (int i = 0; i < num_params; i++) {
     auto required = func_type->getFunctionParamType(i);
     auto provided = arglist[i]->getType();
+    /*
```
jim19930609, line 77: btw, is there a way to "link" the conflicting types imported from files instead of renaming them?

jim19930609 (approved):
Detailed comments regarding the renaming rule would be nicer~

taichi/codegen/llvm/llvm_codegen_utils.cpp
lin-hitonami: The linker treats same types with different names as one, so the different type names do not confuse the linker. Therefore this may not be a problem after the module splitting.

jim19930609 (approved):
Thanks!

taichi/codegen/llvm/llvm_codegen_utils.cpp
```diff
@@ -6,12 +6,59 @@ namespace lang {
...
+bool is_same_type(llvm::Type *required, llvm::Type *provided) {
+  if (required == provided) {
+    return true;
+  }
+  if (required->isPointerTy() != provided->isPointerTy()) {
+    return false;
+  }
+  if (required->isPointerTy()) {
+    required = required->getPointerElementType();
+    provided = provided->getPointerElementType();
```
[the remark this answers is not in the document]
lin-hitonami: Yeah... I'll send another pr to fix this.

PR #5398, opened 2022-07-12
[ci] Properly detect $DISPLAY

feisuzhu:
Related issue = #


ailzhang (approved):
Thanks for fixing it!

PR #5400, opened 2022-07-12
[AOT] Fixed Vulkan device import capability settings

PENGUINLIONG:
This PR fixed Vulkan device capability settings and solved some crashes when AOT module is prepared for a lower version of SPIR-V; and guarded some LLVM codes with `TI_WITH_LLVM` to tolerate Vulkan-only builds.


jim19930609 (approved):
LGTM!

PR #5407, opened 2022-07-13
[type] Decouple quant from SNode 4/n: Add exponent info to BitStructType

strongoier:
Related issue = #4857

Previously, `BitStructType` only contained member types and member offsets, which were not enough to handle `QuantFloatType`, making quant codegen rely on SNode info. This PR adds exponent info to `BitStructType` so that it becomes possible to handle `QuantFloatType` (with or without shared exponent) without SNode info.

### DIFF
--- a/taichi/ir/type.cpp
+++ b/taichi/ir/type.cpp
@@ -180,20 +180,37 @@ BitStructType::BitStructType(
       member_exponents_(member_exponents),
       member_exponent_users_(member_exponent_users) {
   TI_ASSERT(member_types_.size() == member_bit_offsets_.size());
-  int physical_type_bits = data_type_bits(physical_type);
+  TI_ASSERT(member_types_.size() == member_owns_shared_exponents_.size());
+  TI_ASSERT(member_types_.size() == member_exponents_.size());
+  TI_ASSERT(member_types_.size() == member_exponent_users_.size());
+  int physical_type_bits = data_type_bits(physical_type_);
+  int member_total_bits = 0;
   for (auto i = 0; i < member_types_.size(); ++i) {
     QuantIntType *component_qit = nullptr;
     if (auto qit = member_types_[i]->cast<QuantIntType>()) {
       component_qit = qit;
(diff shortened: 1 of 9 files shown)


taichi/ir/type.cpp
```diff
@@ -166,12 +166,19 @@ bool QuantFloatType::get_is_signed() const {
...
+    const std::vector<int> &member_exponents,
+    const std::vector<std::vector<int>> &member_exponent_users)
     : physical_type_(physical_type),
       member_types_(member_types),
-      member_bit_offsets_(member_bit_offsets) {
+      member_bit_offsets_(member_bit_offsets),
+      member_owns_shared_exponents_(member_owns_shared_exponents),
+      member_exponents_(member_exponents),
+      member_exponent_users_(member_exponent_users) {
   TI_ASSERT(member_types_.size() == member_bit_offsets_.size());
```
ailzhang, line 182: nit: do we need to make check more sizes given that we now have more vectors?

tests/cpp/ir/type_test.cpp
```diff
@@ -18,10 +18,10 @@ TEST(Type, BitTypes) {
   auto qu11 = TypeFactory::get_instance().get_quant_int_type(11, false, i32);
   auto u16 = TypeFactory::get_instance().get_primitive_int_type(16, false);
 
-  auto bs =
-      TypeFactory::get_instance().get_bit_struct_type(u16, {qi5, qu11}, {0, 5});
+  auto bs = TypeFactory::get_instance().get_bit_struct_type(
+      u16, {qi5, qu11}, {0, 5}, {false, false}, {-1, -1}, {{}, {}});
 
-  EXPECT_EQ(bs->to_string(), "bs(qi5@0, qu11@5)");
```
ailzhang, line 24: nit: how about adding an additional check for bit_struct_type with shared exponent?

ailzhang (approved):
LGTM!

taichi/ir/type.cpp
strongoier: done

PR #5411, opened 2022-07-13
[bug] Simplify scalar handling in cgraph and relax field_dim check

ailzhang:
This PR fixes two issues showed up while porting implicit_fem demo:
- field_dim check can be relaxed, for example when users write `for I in
ti.grouped(arr)` it actually works for ndarrays with different
`field_dim`s.
- passing scalars to c++ shouldn't be seperated based on python
type and it was not correct to do so. For example users can pass a `0`
(python int) to a `ti.f32` graph argument.

Related issue = #


turbo0628 (approved):
LGTM!

PR #5412, opened 2022-07-13
[autodiff] Check not placed field.dual and add needs_dual

erizmr:
Related issue = #5295

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -229,6 +229,7 @@ class PyTaichi:
         self.current_kernel = None
         self.global_vars = []
         self.grad_vars = []
+        self.dual_vars = []
         self.matrix_fields = []
         self.default_fp = f32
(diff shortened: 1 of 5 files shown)


python/taichi/lang/impl.py
```diff
@@ -541,15 +560,25 @@ def create_field_member(dtype, name, needs_grad):
         x_dual.ptr.set_name(name + ".dual")
         x_dual.ptr.set_is_primal(False)
         x.ptr.set_dual(x_dual.ptr)
+        if needs_dual:
+            pytaichi.dual_vars.append(x_dual)
     elif needs_grad:
         raise TaichiRuntimeError(
             f'{dtype} is not supported for field with `needs_grad=True`.')
+    elif needs_dual:
```
ailzhang, line 568: ditto, let's merge this to simplify the code logic a bit :P

PR #5414, opened 2022-07-13
[Doc] Updated the docsite structure

Vissidarte-Herman:
Related issue = #


ailzhang (commented):
could you revert the changes in `python` and `taichi` folder so that this PR can skip most of the CI jobs except for the netlify preview?

PR #5415, opened 2022-07-13
[build] [refactor] Move spirv codegen and common targets

qiao-bo:
Related issue = #4832 

- Add spirv codegen and move the build of spirv-tools out of the core target.
- Define a `taichi_common` target
- Move dangling lang_util files to util folder and define `taichi_util` target.


jim19930609 (approved):
LGTM！

PR #5417, opened 2022-07-13
[AOT] Adjusted C-API for nd-array type conformance

PENGUINLIONG:
The following change breaks the current (type-relaxed) implementation. A primitive data type tag is added to `TiNdArray`.

https://github.com/taichi-dev/taichi/blob/master/taichi/aot/graph_data.cpp#L33-L37


ailzhang (approved):
Thanks!!

PR #5419, opened 2022-07-14
[aot] Support specifying vk_api_version in CompileConfig

ailzhang:
Ideally we should specify this in `ti.aot.Module` instead of `ti.init`
but tbh we need more refactor before we can compile to a target other
than the current running one. So let's use this to avoid complicating
other parts of the codebase, e.g. offline cache uses CompileConfig as
part of cache key.

Related issue = #

### DIFF
--- a/c_api/src/taichi_vulkan_impl.cpp
+++ b/c_api/src/taichi_vulkan_impl.cpp
@@ -72,7 +72,7 @@ make_vulkan_runtime_creator_params() {
 
   // FIXME: (penguinliong) Vulkan runtime should be created outside.
   taichi::lang::vulkan::VulkanDeviceCreator::Params params{};
-  params.api_version = VK_API_VERSION_1_3;
+  params.api_version = std::nullopt;
   params.additional_instance_extensions = extensions;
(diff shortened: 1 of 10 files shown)


PENGUINLIONG:
I just pushed some changes letting the user to specify Vulkan API version (`vk_api_version`), instead of SPIR-V spec version (`spirv_version`) because our device creation logic enabled SPIR-V capabilities based on the set of available extensions on the Vulkan device. The change added more semantics to `VulkanDeviceCreator::Params::api_version`. The users are not expected to assign to this field if they want Taichi to automatically select the Vulkan API version and instance + device extensions to enable.

ailzhang:
> I just pushed some changes letting the user to specify Vulkan API version (`vk_api_version`), instead of SPIR-V spec version (`spirv_version`) because our device creation logic enabled SPIR-V capabilities based on the set of available extensions on the Vulkan device. The change added more semantics to `VulkanDeviceCreator::Params::api_version`. The users are not expected to assign to this field if they want Taichi to automatically select the Vulkan API version and instance + device extensions to enable.

ah this is a bit unfortunate.... but yea let's work around this issue by specifying vk_api_version for now.

PENGUINLIONG (approved):

Issue #5338, opened 2022-07-06
[Perf] Shared Array APIs


turbo0628:
## TL;DR
I would like to request a scratch pad (shared memory in CUDA) API that exhibits ~2x n-body simulation performance on CUDA GPUs, which is on-par with the best CUDA implementations.

## The Problem
The Taichi kernel function for n-body simulation is demonstrated as follows:
```Python
import taichi as ti

nBodies = 3072
softening = 1e-9
dt = 0.01
block_dim=128

ti.init(arch=ti.cuda)

bodies = ti.field(shape=(nBodies * 4), dtype=ti.float32)
velocities = ti.field(shape=(nBodies, 3), dtype=ti.float32)
@ti.kernel
def bodyForce():
    block_dim = ti.static(128)
    ti.loop_config(block_dim=block_dim)
    # LOOP 1
    for i in range(nBodies):
        Fx = 0.0
        Fy = 0.0
        Fz = 0.0
        bx = bodies[i * 4 + 0]
        by = bodies[i * 4 + 1]
        bz = bodies[i * 4 + 2]
        # LOOP 2
        for j in range(nBodies / block_dim):
            # FETCH
            # LOOP BLOCK
            for t in range(block_dim / 8):
                # LOOP UNROLL
                for s in ti.static(range(8)):
                    #############
                    # LOADS FROM DRAM
                    dx = bodies[(j * block_dim + t * 8 + s) * 4 + 0] - bx
                    dy = bodies[(j * block_dim + t * 8 + s) * 4 + 1] - by
                    dz = bodies[(j * block_dim + t * 8 + s) * 4 + 2] - bz
                    #############
                    distSqr = dx * dx + dy * dy + dz * dz + softening
                    invDist = 1.0 / ti.sqrt(distSqr)
                    invDist3 = invDist * invDist * invDist
                    Fx += dx * invDist3
                    Fy += dy * invDist3
                    Fz += dz * invDist3
        velocities[i, 0] += dt * Fx
        velocities[i, 1] += dt * Fy
        velocities[i, 2] += dt * Fz

    for i in range(nBodies):
        bodies[i * 4 + 0] = bodies[i * 4 + 0] + velocities[i, 0] * dt
        bodies[i * 4 + 1] = bodies[i * 4 + 1] + velocities[i, 1] * dt
        bodies[i * 4 + 2] = bodies[i * 4 + 2] + velocities[i, 2] * dt

bodyForce()
```
[...]

PR #5429, opened 2022-07-14
[CUDA] Enable shared memory for CDUA

turbo0628:
Related issue = #5338

### DIFF
--- a/python/taichi/lang/simt/block.py
+++ b/python/taichi/lang/simt/block.py
@@ -1,5 +1,17 @@
 from taichi.lang import impl
-
+from taichi.lang.util import taichi_scope
 
 def sync():
     return impl.call_internal("block_barrier", with_runtime_context=False)
+
+
+class SharedArray:
+    _is_taichi_class = True
+    def __init__(self, shape, dtype):
+        self.shape = shape
+        self.dtype = dtype
+        self.shared_array_proxy = impl.expr_init_scratch_pad(shape, dtype)
+
+    @taichi_scope
+    def _subscript(self, indices, get_ref=False):
+        return impl.make_index_expr(self.shared_array_proxy, (indices,))
--- a/taichi/transforms/ir_printer.cpp
+++ b/taichi/transforms/ir_printer.cpp
@@ -119,8 +119,9 @@ class IRPrinter : public IRVisitor {
   }
 
   void visit(FrontendAllocaStmt *alloca) override {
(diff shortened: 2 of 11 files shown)


python/taichi/lang/simt/block.py
```diff
@@ -1,5 +1,19 @@
...
+    _is_taichi_class = True
+
+    def __init__(self, shape, dtype):
+        self.shape = shape
+        self.dtype = dtype
+        self.shared_array_proxy = impl.expr_init_shared_array(shape, dtype)
+
+    @taichi_scope
+    def _subscript(self, indices, get_ref=False):
+        return impl.make_index_expr(self.shared_array_proxy, (indices, ))
```
turbo0628, line 19: @ailzhang I'm not sure if this is the best method to load an element. Furthermore, I wonder if it's possible to manually keep a pointer that refers to a partial position, like this:
```Python
for p in range(...):
    pad = ti.simt.block.SharedArray(...)
    for i in range(..):
         pad_i = pad.offset(128 + 32 * i)
         for j in range(...):
             calc(pad_i)
```

ailzhang: @turbo0628 we cannot support partial position for now IIUC, as that might create aliasing issues?

taichi/transforms/ir_printer.cpp
```diff
@@ -119,8 +119,9 @@ class IRPrinter : public IRVisitor {
   }
 
   void visit(FrontendAllocaStmt *alloca) override {
-    print("{}${} = alloca {}", alloca->type_hint(), alloca->id,
-          alloca->ident.name());
+    std::string shared_suffix = (alloca->is_shared) ? " shared" : "";
+    print("{}${} = alloca {}{}", alloca->type_hint(), alloca->id,
```
lin-hitonami, line 123: Maybe `... = alloca(shared) ...`  or `... = alloca_shared ...` will be better than simply appending `shared` at the end?

taichi/transforms/ir_printer.cpp
```diff
@@ -187,7 +188,8 @@ class IRPrinter : public IRVisitor {
   }
 
   void visit(AllocaStmt *alloca) override {
-    print("{}${} = alloca", alloca->type_hint(), alloca->id);
+    std::string shared_suffix = (alloca->is_shared) ? " shared" : "";
```
lin-hitonami, line 191: Ditto

lin-hitonami (approved):
LGTM! Just a few nits on the printing format

PR #5430, opened 2022-07-14
[opengl] Speedup compilation for Nvidia cards

bobcao3:
Nvidia GLSL compilers will try to extract statically accessed array elements and place all of them directly, and it is a bit too smart for its own good.

e.g.
```
int arr[];

arr[8000] = 3.0;
```
will be detected as static and the array will be replaced with a sized one: ```int arr[8000]```. This slows down the driver's GLSL compiler as it places all 8000 elements. This PR sorts the fields according to the size so that all the statically accessed fields (mostly matrices and global states) come first and the driver will not place extra elements.


k-ye (approved):
LGMT, thanks!

PR #5433, opened 2022-07-15
[AOT] Support on-device event

PENGUINLIONG:
This PR implements an experimental abstraction of on-device event. Event is a type of synchronization primitive available on most modern graphics APIs to synchronize on-device execution. It enables the users to immediately kick off rendering after Taichi kernel launches, without any interaction with the host.

Event is different from semaphore. Event only synchronizes execution flows on device and can be signaled and awaited at any point in the command stream. Semaphores can only be signaled at the end of a stream, and awaited in the beginning. There exists cases that the users don't have the control over command buffer submission, like in Unity, which means that semaphores might not the best primitive to use. So the support for event is considered plausible.


ailzhang (approved):
Nice!!

PR #5441, opened 2022-07-18
[llvm] [bug] Fix llvm_codegen_utils

feisuzhu:
Fixing the crash in release tests introduced by a typo in #5381 where we need a deep copy of arglist.


ailzhang (approved):
Thanks!!

PR #5448, opened 2022-07-18
[type] [refactor] Decouple quant from SNode 6/n: Rewrite extract_quant_float() without SNode

strongoier:
Related issue = #4857

The old `load_quant_float()` can actually be merged into `extract_quant_float()` . To achieve this, related logic has been refactored and (fortunately) greatly simplified.


ailzhang (approved):
Awesome!!

PR #5454, opened 2022-07-19
[bug] Set SNode tree id to all SNodes

lin-hitonami:
Related issue = #4401 
Fixes wrong answers on `test_mesh.py` when offline cache is on which are caused by getting `snode_tree_id` of non-root SNodes whose `snode_tree_id`s are not set.


PGZXB (approved):
Thanks!

strongoier (approved):

PR #5455, opened 2022-07-19
[gui] [vulkan] A faster version of depth copy through ti.field/ti.ndarray (copy directly from vulkan to cuda/gpu/cpu)

Morcki:
## Usage
```python
window_shape = (720, 1080)
window = ti.ui.Window("Test for copy depth memory", window_shape)
canvas = window.get_canvas()
scene = ti.ui.Scene()
camera = ti.ui.make_camera()

# get the shape of window
w, h = window.get_window_shape()
# depth field/ndarray must have ti.f32 data type and 2d shape.
# and shape must be equal to the window shape
scene_depth = ti.ndarray(ti.f32, shape = (w, h))
# scene_depth = ti.field(ti.f32, shape = (w, h))

while window.running:
    render()
    canvas.scene(scene)
    window.get_depth_buffer_faster(scene_depth)
    window.show()
```
## Explanation
+ Q : ~~why use ndarray instead of field?~~ Now support both ndarray and field data structure.
+ A : ti.field struct has a complicated memory layout under unpacked mode(which is default mode for taichi), so the physical address of your index numbers may not directly to the right place(field memory must be padding to power of 2, and there may be some dirty data on the memory). As for ti.ndarray, it does not have this problem, because it's memory layout is what you see, and there is no other designs for it.


python/taichi/_kernels.py
```diff
@@ -179,6 +182,19 @@ def ext_arr_to_matrix(arr: ndarray_type.ndarray(), mat: template(),
                     mat[I][p, q] = arr[I, p, q]
 
 
+@kernel
+def ext_arr_from_vulkan_layout_to_normal_layout(
```
YuCrazing, line 186: Here we should add more comments to explain the purpose of this kernel's existence. And we should indicate that this kernel is only for internal usage.

python/taichi/ui/window.py
```diff
@@ -126,6 +131,19 @@ def write_image(self, filename):
         """
         return self.window.write_image(filename)
 
+    def copy_depth_buffer_to_ndarray(self, depth: Ndarray):
+        """fetch the depth information of current scene to ti.Ndarray.
+           (support copy from vulkan to cuda/cpu which is a faster version)
+        Args:
+            depth (Ndarray): [window_width, window_height] carries depth information.
+        """
+        if not (isinstance(depth, Ndarray) and len(depth.shape) == 2):
```
YuCrazing, line 140: I think we should check the shape of depth ndarrary here.

Morcki: There is no function method to get resolution of window on python scope, so I judge this on the cpp part (window.cpp).

taichi/ui/common/window_base.h
```diff
@@ -41,6 +42,8 @@ class WindowBase {
 
   virtual void write_image(const std::string &filename) = 0;
 
+  virtual void copy_depth_buffer_to_ndarray(taichi::lang::Ndarray *) = 0;
```
ailzhang, line 45: I feel like it's possible to use `const taichi::lang::Ndarray& ndarray)` , could you give that a try?

python/taichi/_kernels.py
```diff
@@ -179,6 +179,29 @@ def ext_arr_to_matrix(arr: ndarray_type.ndarray(), mat: template(),
...
+        vk_image: ndarray_type.ndarray(),
+        normal_image: ndarray_type.ndarray()):
+    '''extract ndarray of raw vulkan memory layout to normal memory
```
ailzhang, line 186: Formatting for this docstring is a bit off - could you  please fix it?

python/taichi/_kernels.py
```diff
@@ -179,6 +179,29 @@ def ext_arr_to_matrix(arr: ndarray_type.ndarray(), mat: template(),
...
+    layout. Explanation of why using ti.ndarray:
+        ti.field struct has a complicated memory layout under unpacked
```
ailzhang, line 188: OOC: why do we want to mention `ti.field` in this docstring? I'd expect this one to explain what's vulkan memory layout   and normal layout means :P

taichi/ui/backends/vulkan/window.cpp
```diff
@@ -102,6 +105,46 @@ void Window::write_image(const std::string &filename) {
...
+  // instead of accessing through the current SNode
+  if (depth_arr->ndarray_alloc_ == taichi::lang::kDeviceNullAllocation) {
+    TI_ERROR("Null Allocation for Depth-Ndarray!");
+  }
+
+  auto arr_dev_ptr = depth_arr->ndarray_alloc_.get_ptr();
+  renderer_->swap_chain().copy_depth_buffer_to_ndarray(arr_dev_ptr);
+
+  if (!config_.show_window) {
+    prepare_for_next_frame();
```
ailzhang, line 144: OOC: if I call copy depth buffer twice without presenting, is it expected behavior to skip them?

python/taichi/ui/window.py
```diff
@@ -126,6 +128,19 @@ def write_image(self, filename):
         """
         return self.window.write_image(filename)
 
+    def copy_depth_buffer_to_ndarray(self, depth: Ndarray):
```
ailzhang, line 131:
```suggestion
    def get_depth_buffer(self, depth: Ndarray):
```

ailzhang (changes requested):
Thanks for adding this feature! Added a few nits in comments. 
Would you mind adding some tests as well?

taichi/ui/backends/vulkan/window.cpp
ailzhang: sorry I was dumb :P

taichi/python/export_ggui.cpp
```diff
@@ -287,10 +288,19 @@ struct PyWindow {
     window = std::make_unique<vulkan::Window>(prog, config);
   }
 
+  py::tuple get_window_shape() {
+    auto [w, h] = window->get_window_shape();
```
ailzhang, line 292: OOC: is there any difference between resolution saved in Appconfig and w&h saved in swap_chain? Looks like the former is easier to get?

ailzhang (commented):
Thanks! I have a few small questions left inline, will leave the final review of this PR to @PENGUINLIONG ;)

taichi/ui/backends/vulkan/swap_chain.cpp
```diff
@@ -45,6 +45,58 @@ void SwapChain::resize(uint32_t width, uint32_t height) {
...
+
+    BufferImageCopyParams copy_params;
+    copy_params.image_extent.x = w;
+    copy_params.image_extent.y = h;
+    copy_params.image_aspect_flag = VK_IMAGE_ASPECT_DEPTH_BIT;
+    cmd_list = stream->new_command_list();
+    cmd_list->image_to_buffer(depth_staging_buffer.get_ptr(), depth_allocation_,
+                              ImageLayout::transfer_src, copy_params);
+    cmd_list->image_transition(depth_allocation_, ImageLayout::transfer_src,
+                               ImageLayout::present_src);
```
PENGUINLIONG, line 78: Depth image is never presented. Do you mean `ImageLayout::transfer_src`->`ImageLayout::depth_attachment`? Not a big issue on desktop platforms yet better to be corrected.

taichi/ui/backends/vulkan/swap_chain.cpp
```diff
@@ -45,6 +45,58 @@ void SwapChain::resize(uint32_t width, uint32_t height) {
...
+
+  if (memcpy_cap == Device::MemcpyCapability::Direct) {
+    Device::AllocParams params{copy_size, /*host_wrtie*/ false,
+                               /*host_read*/ false, /*export_sharing*/ true,
+                               AllocUsage::Uniform};
+
+    auto depth_staging_buffer = device.allocate_memory(params);
+
+    device.image_transition(depth_allocation_, ImageLayout::present_src,
+                            ImageLayout::transfer_src);
```
PENGUINLIONG, line 68: `ImageLayout::depth_attachment`?

PENGUINLIONG (changes requested):
just noticed some improper image transitions but everything else lgtm, nice work

PENGUINLIONG (approved):
lgtm

PR #5457, opened 2022-07-19
[error] Throw proper error message when an Ndarray is passed in via ti.template

### LINKED ISSUE #5452: Got `__getitem__` error when passing ndarray as template.
**Describe the bug**
When passing ti.ndarray as template arg, and try to get item value from the arg, it reports bug as below:
`AssertionError: __getitem__ cannot be called in Taichi-scope`

**Python Script**
the python script is as below:
```py
import taichi as ti

ti.init(arch=ti.cpu)

from taichi.types import ndarray_type
from taichi.types.annotations import template

@ti.kernel
def load_item_error(arr_src : template(), arr_dst : template()):
    for i, j in ti.ndrange(*arr_src.shape):
        arr_dst[i, j] = arr_src[i, j]

@ti.kernel
def load_item_correct(arr_src : ndarray_type.ndarray(), arr_dst : ndarray_type.ndarray()):
    for i, j in ti.ndrange(*arr_src.shape):
        arr_dst[i, j] = arr_src[i, j]

arr_0 = ti.ndarray(ti.f32, shape = (5, 10))
arr_1 = ti.ndarray(ti.f32, shape = (5, 10))

# report bug
load_item_error(arr_0, arr_1)

# correct case
# load_item_correct(arr_0, arr_1)
```

If I choose the `load_item_correct` kernel, it would be right without errors, It seems like there is something wrong in template().

ailzhang:
ti.types.ndarray itself is a template type so let's throw a better error
message when users misuse it.

fixes #5452

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -335,6 +335,10 @@ class TaichiCallableTemplateMapper:
                 return tuple(
                     TaichiCallableTemplateMapper.extract_arg(item, anno)
                     for item in arg)
+            if isinstance(arg, taichi.lang._ndarray.Ndarray):
+                raise TaichiRuntimeTypeError(
+                    'Ndarray shouldn\'t be passed in via `ti.template()`, please annotate your kernel using `ti.types.ndarray(...)` instead'
+                )
             return arg
         if isinstance(anno, texture_type.TextureType):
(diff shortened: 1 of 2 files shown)


Morcki (approved):
LGTM!
Seems friendly to taichi-newer!

PR #5458, opened 2022-07-19
[llvm] [aot] Added field tests for LLVM backend in CGraph

jim19930609:
Related issue = #4800


tests/cpp/aot/llvm/field_cgraph_test.py
```diff
@@ -0,0 +1,17 @@
...
+
+import taichi as ti
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--arch", type=str)
+args = parser.parse_args()
+
+if __name__ == "__main__":
+    if args.arch == "cpu":
+        compile_field_aot(arch=ti.cpu, compile_for_cgraph=True)
```
ailzhang, line 13: aha I feel like this can be merged with `field_aot_test.cpp` with an additional command argument? ;)

ailzhang (approved):
LGTM!

tests/cpp/aot/llvm/field_cgraph_test.py
jim19930609: sounds good~

PR #5473, opened 2022-07-19
[bug] [llvm] Fix GEP when allocating TLS buffer in struct for

lin-hitonami:
Related issue = #4401 

Taichi allocates tls buffer in struct for by finding the alloca `%18 = alloca [1 x i8], align 8` generated by the C++ code https://github.com/taichi-dev/taichi/blob/2c793e4279808ba37462b6962465d49943fc425a/taichi/runtime/llvm/runtime_module/runtime.cpp#L1369, and changing the allocated type of it to `[tls_size x i8]` by `alloca->setAllocatedType`. After allocation, Taichi uses a GEP instruction to get the pointer to its first element. The GEP instruction was `%82 = getelementptr inbounds [1 x i8], [1 x i8]* %18, i64 0, i64 0` before changing the allocated type of the tls buffer. 

`alloca->setAllocatedType` changes the allocated type of the alloca statement, but it does not change the type of the llvm statement. Therefore, when dumping into the `.ll` file, it creates buggy GEP statement `%82 = getelementptr inbounds [1 x i8], [8 x i8]* %18, i64 0, i64 0` which can't be loaded.
image

Therefore, I create new alloca and GEP statement instead of using `alloca->setAllocatedType` in this PR.


taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2147,8 +2146,34 @@ void CodeGenLLVM::create_offload_struct_for(OffloadedStmt *stmt, bool spmd) {
...
+          {
+            llvm::IRBuilderBase::InsertPointGuard guard(*builder);
+            builder->SetInsertPoint(alloca);
+            auto *new_alloca = builder->CreateAlloca(new_type);
+            new_alloca->setAlignment(Align(8));
+            replaced_alloca_types += 1;
+            for (llvm::User *user : alloca->users()) {
+              if (llvm::GetElementPtrInst *gep =
+                      llvm::dyn_cast<llvm::GetElementPtrInst>(user)) {
+                if (gep->getPointerOperand() == alloca) {
```
python3kgae, line 2158: Could the pointerOperand not equal alloca?

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2147,8 +2146,34 @@ void CodeGenLLVM::create_offload_struct_for(OffloadedStmt *stmt, bool spmd) {
...
+                    llvm::cast<llvm::GetElementPtrInst>(new_gep)->setIsInBounds(
+                        true);
+                    gep->replaceAllUsesWith(new_gep);
+                    gep->eraseFromParent();
+                    break;
+                  }
+                }
+              }
+            }
+            alloca->eraseFromParent();
```
python3kgae, line 2174: Are there users other than GEP?

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2147,8 +2146,34 @@ void CodeGenLLVM::create_offload_struct_for(OffloadedStmt *stmt, bool spmd) {
             alloca_type->getArrayNumElements() == 1 &&
             alloca_type->getArrayElementType() == char_type) {
           auto new_type = llvm::ArrayType::get(char_type, stmt->tls_size);
-          alloca->setAllocatedType(new_type);
```
python3kgae, line 2150: Could we keep this and just setSourceElementType on all the GEP users instead of create new alloca and GEP?

lin-hitonami: I tried and it raised error because the type of the AllocaInst is still [1 × i8].

taichi/codegen/llvm/codegen_llvm.cpp
lin-hitonami: Oh maybe I can simply loop over the users of the AllocaInst instead of looping over all the instructions in the function. Thanks.

taichi/codegen/llvm/codegen_llvm.cpp
lin-hitonami: You are right, the pointerOperand must be the alloca, so this branch is redundant.

taichi/codegen/llvm/codegen_llvm.cpp
python3kgae: How about add alloca->mutateType(updatedPointerType)

lin-hitonami: It raises "isa<X>(Val) && \"cast<Ty>() argument of incompatible type!\"" when cloning the AllocaInst when cloning the module. Also, the description of `mutateType` said "Note that this is an extremely dangerous operation which can create completely invalid IR very easily. It is strongly recommended that you recreate IR objects with the right types instead of mutating them in place." so I don't think it is a good idea to use this function...

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2143,37 +2142,52 @@ void CodeGenLLVM::create_offload_struct_for(OffloadedStmt *stmt, bool spmd) {
...
-
+    TI_ASSERT(num_found_alloca == 1 && alloca);
+    auto new_type = llvm::ArrayType::get(char_type, stmt->tls_size);
+    {
+      llvm::IRBuilderBase::InsertPointGuard guard(*builder);
+      builder->SetInsertPoint(alloca);
+      auto *new_alloca = builder->CreateAlloca(new_type);
+      new_alloca->setAlignment(Align(8));
+      TI_ASSERT(alloca->hasOneUse());
+      auto *gep = llvm::dyn_cast<llvm::GetElementPtrInst>(alloca->user_back());
```
python3kgae, line 2182: Maybe change llvm::dyn_cast to llvm::cast which does the assert.

python3kgae (approved):
Would be nice to have a test if possible.

lin-hitonami:
> Would be nice to have a test if possible.

This will be automatically tested when the offline cache feature is enabled.

PR #5481, opened 2022-07-20
[Lang] Better struct initialization

neozhaoliang:
Related issue = #5485 

Currently in Taichi to create an instance of a `StructType` one has to write in either the following way:

1. pass the full set of `key=val` pairs in `kwargs`, or pass a dict contains the full set of struct members.
2. use a  single scalar to fill all struct members.

This PR allows the user to create a `StructType` instance by passing only a subset of struct  memebers, and fill the struct members in the order they are passed:

```python
import taichi as ti
from taichi.math import vec2
ti.init()

vec2_t = ti.types.vector(2, float)

S = ti.types.struct(
    x=float,
    y=float,
    z=vec2_t
)

@ti.kernel
def test():
    s = S(1.0, z=vec2(1))
    print(s.x, s.y, s.z)  # 1.0, 0.0 [1.0, 1.0]

test()
```

This also works for `ti.dataclass`:

```python
import taichi as ti

ti.init(arch=ti.cpu)
vec3 = ti.types.vector(3, float)

@ti.dataclass
class Foo:

    pos: vec3
    vel: vec3
    mass: float


@ti.kernel
def test():
    A = Foo((1, 1, 1), mass=2)
    print(A.pos, A.vel, A.mass)  # [1., 1. ,1.] [0., 0., 0.] 2.0

test()
```


lin-hitonami (approved):
LGTM

PR #5410, opened 2022-07-13
[gui] [vulkan] Surpport for getting depth information for python users.

Morcki:
Usage:
``` python
window = ti.ui.Window("Test for getting depth buffer from ggui", (768, 768), vsync=True)
video = ti.tools.VideoManager("OutputDir")

while window.running:
    render_scene()
    depth = window.get_depth_buffer()
    video.write_frame(depth)
    window.show()
```

PR #5484, opened 2022-07-21
[gui] [vulkan] Modify some internal APIs

Morcki:
ChangeList:
1. Remove old API function `dump_depth_buffer` from swap_chain.h. Still keep `Window.get_depth_buffer_as_numpy` (usage is same with #5410 ). And there is also another API function `get_depth_buffer` which support multi-backends (details refer to #5455 ).
2. Fix a bug about `Scene.mesh` when there is no indices, it would get a bug.


YuCrazing (approved):
LGTM!

python/taichi/ui/scene.py
```diff
@@ -142,7 +142,10 @@ def mesh(self,
         if vertex_count is None:
             vertex_count = vertices.shape[0]
         if index_count is None:
-            index_count = indices.shape[0]
+            if indices is None:
+                index_count = vertex_count  # FIXME : Need to confirm
+            else:
+                index_count = indices.shape[0]
```
PENGUINLIONG, line 148: probably the following?
```python
if indices is not None:
    if index_count is None:
        index_count = indices.shape[0]
```

PENGUINLIONG (approved):
lgtm despite a minor tweak

python/taichi/ui/scene.py
Morcki: I think we must pass a value to index_count.

Morcki: I tried to pass this value using None, but failed to get an error information. It's the reason that pybind donnot support to passing None value to int object.

ailzhang (approved):
LGTM！

PR #5488, opened 2022-07-21
[Vulkan] Added an interface to get accumulated on-device execution time

PENGUINLIONG:
Currently we don't have an interface to query on-device execution time for the Vulkan backend. Having a gut-feeling that the profiling results we have now might be contaminated by synchronization, I have implemented a prototype to get accumulated time for more precise assessment of our SPIR-V program quality. The implementation is only provisional and the interface is subject to change (`ti.get_compute_stream_device_time_elapsed_us()`).


PENGUINLIONG:
As discussed in #5522 there is a major blocker integrating the Vulkan on-device timers into the existing profiler system. I'm gonna let this interface in for a moment and this interface shall be provisional only. cc @turbo0628

turbo0628 (approved):
Thanks, LGTM!

PR #5495, opened 2022-07-22
[gui] Support rendering lines from a part of VBO

Morcki:
Usage:
```python
import taichi as ti

ti.init(arch=ti.cuda)

N = 10

particles_pos = ti.Vector.field(3, dtype=ti.f32, shape = N)
points_pos = ti.Vector.field(3, dtype=ti.f32, shape = N)
points_indices = ti.Vector.field(1, dtype=ti.i32, shape = N)
        
@ti.kernel
def init_points_pos(points : ti.template()):
    for i in range(points.shape[0]):
        points[i] = [i for j in range(3)]
        # points[i] = [ti.sin(i * 1.0), i * 0.2, ti.cos(i * 1.0)]
        
@ti.kernel
def init_points_indices(points_indices : ti.template()):
    for i in range(N):
        points_indices[i][0] = i // 2 + i % 2

init_points_pos(particles_pos)
init_points_pos(points_pos)
init_points_indices(points_indices)

window = ti.ui.Window("Test for Drawing 3d-lines", (768, 768))
canvas = window.get_canvas()
scene = ti.ui.Scene()
camera = ti.ui.make_camera()
camera.position(5, 2, 2)

while window.running:
    camera.track_user_inputs(window, movement_speed=0.03, hold_key=ti.ui.RMB)
    scene.set_camera(camera)
    scene.ambient_light((0.8, 0.8, 0.8))
    scene.point_light(pos=(0.5, 1.5, 1.5), color=(1, 1, 1))
    
    scene.particles(particles_pos, color = (0.68, 0.26, 0.19), radius = 0.1)
    # Here you will get visible part from the 3rd point with (N - 4) points.
    scene.lines(points_pos, color = (0.28, 0.68, 0.99), width = 5.0, vertex_count = N - 4, vertex_offset = 2)
    # Using indices to indicate which vertex to be used
    # scene.lines(points_pos, color = (0.28, 0.68, 0.99), width = 5.0, indices = points_indices)
    # Case 1, vertex_count will be changed to N - 2 when drawing.
    # scene.lines(points_pos, color = (0.28, 0.68, 0.99), width = 5.0, vertex_count = N - 1, vertex_offset = 0)
    # Case 2, vertex_count will be changed to N - 2 when drawing.
    # scene.lines(points_pos, color = (0.28, 0.68, 0.99), width = 5.0, vertex_count = N, vertex_offset = 2)
    canvas.scene(scene)
    window.show()
```
Details:
Add 4 args for users to control visible part of the 3d lines.Check the note of Scene.lines.

Sepecial note about `vertex_count`:
There are 2 cases that we will change your vertex_count. 
1. If the `vertex_count` is a odd number, then we will change it to `vertex_count - 1`. Refer to case 1 in above python script.
2. If `vertex_offset` + `vertex_count` > `vertices.shape[0]`, then we will reduce `vertex_count` to no more than `vertices.shape[0]` (still keep it a even number) as the equation shows:(Refer to case 2 in above python script.)
    ```python
    vertex_count = vertices.shape[0] - vertex_offset
    vertex_count -= vertex_count % 2
    ```

### DIFF
--- a/python/taichi/ui/scene.py
+++ b/python/taichi/ui/scene.py
@@ -101,7 +101,7 @@ class Scene:
         """Declare multi-lines inside the scene.
 
         Note that under current situation, for example, there you have 4 vertices,
-        vertices.shape[0] is 4. So there will be 2 lines, first line's two points
+        vertices.shape[0] is 4. So there will be 2 lines, the first line's two points
         are vertices[0] and vertices[1], and the second line's two points are
         vertices[2] and vertices[3].
 
@@ -117,13 +117,13 @@ class Scene:
             per_vertex_color (Tuple[float]): a taichi 3D vector field, where each
                 element indicate the RGB color of the line.
             vertex_offset: int type(otherwise float type will be floored to int),
-                if 'indices' is provided, this means the value added to the vertex
-                index before indexing into the vertex buffer, else this means the
+                if 'indices' is provided, this refers to the value added to the vertex
+                index before indexing into the vertex buffer, else this refers to the
                 index of the first vertex to draw.
             vertex_count: int type(ohterwise float type will be floored to int),
-                only avaliable when `indices` is not provided, which is the number
+                only available when `indices` is not provided, which is the number
                 of vertices to draw. There are 2 cases that we will change your
-                vertex_count. [1] If the `vertex_count` is a odd number, then we
+                `vertex_count`. [1] If the `vertex_count` is an odd number, then we
                 will change it to `vertex_count` - 1. [2] If `vertex_offset` plus
                 `vertex_count` greater than vertices.shape[0], then we will reduce
                 `vertex_count` to no more than vertices.shape[0].
@@ -192,18 +192,18 @@ class Scene:
                 element indicate the RGB color of a vertex.
             two_sided (bool): whether or not the triangles should be able to be
                 seen from both sides.
-            vertex_offset: int type(ohterwise float type will be floored to int),
-                if 'indices' is provided, this means the value added to the vertex
-                index before indexing into the vertex buffer, else this means the
+            vertex_offset: int type(otherwise float type will be floored to int),
+                if 'indices' is provided, this refers to the value added to the vertex
+                index before indexing into the vertex buffer, else this refers to the
                 index of the first vertex to draw.
             vertex_count: int type(ohterwise float type will be floored to int),
-                only avaliable when `indices` is not provided, which is the number
+                only available when `indices` is not provided, which is the number
                 of vertices to draw.
             index_offset: int type(ohterwise float type will be floored to int),
-                only avaliable when `indices` is provided, which is the base index
+                only available when `indices` is provided, which is the base index
                 within the index buffer.
             index_count: int type(ohterwise float type will be floored to int),
-                only avaliable when `indices` is provided, which is the the number
+                only available when `indices` is provided, which is the the number
                 of vertices to draw.
         """
         vbo = get_vbo_field(vertices)
--- a/python/taichi/ui/scene.py
+++ b/python/taichi/ui/scene.py
@@ -127,7 +127,7 @@ class Scene:
                 will change it to `vertex_count` - 1. [2] If `vertex_offset` plus
                 `vertex_count` greater than vertices.shape[0], then we will reduce
                 `vertex_count` to no more than vertices.shape[0].
-            index_offset: int type(ohterwise float type will be floored to int),
+            index_offset: int type (otherwise float type will be floored to int).
                 only avaliable when `indices` is provided, which is the base index
                 within the index buffer.
             index_count: int type(ohterwise float type will be floored to int),
--- a/python/taichi/ui/scene.py
+++ b/python/taichi/ui/scene.py
@@ -99,10 +99,12 @@ class Scene:
               index_offset: int = 0,
               index_count: int = None):
         """Declare multi-lines inside the scene.
+        
         Note that under current situation, for example, there you have 4 vertices,
         vertices.shape[0] is 4. So there will be 2 lines, first line's two points
         are vertices[0] and vertices[1], and the second line's two points are
         vertices[2] and vertices[3].
+        
         Args:
             vertices: a taichi 3D Vector field, where each element indicate the
                 3D location of points of lines.
@@ -170,9 +172,11 @@ class Scene:
              index_offset: int = 0,
              index_count: int = None):
         """Declare a mesh inside the scene.
+        
         if you indicate the index_offset and index_count, the normals will also
         be sliced by the args, and the shading resultes will not be affected.
         (It is equal to make a part of the mesh visible)
+        
         Args:
             vertices: a taichi 3D Vector field, where each element indicate the
                 3D location of a vertex.
--- a/python/taichi/ui/scene.py
+++ b/python/taichi/ui/scene.py
@@ -131,7 +131,7 @@ class Scene:
                 Only available when `indices` is provided, which is the base index
                 within the index buffer.
             index_count: int type (otherwise float type will be floored to int),
-                only avaliable when `indices` is provided, which is the the number
+                Only available when `indices` is provided, which is the number
                 of vertices to draw.
         """
         if vertex_count is None:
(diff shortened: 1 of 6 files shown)


YuCrazing (approved):
LGTM!

PENGUINLIONG (approved):
lgtm

neozhaoliang (changes requested):
All are great,  except there should be one line break after line 101, and one line break after 105 for our doc formatting :smiley: :

python/taichi/ui/scene.py
```diff
@@ -106,7 +116,39 @@ def lines(self,
...
+                if 'indices' is provided, this means the value added to the vertex
+                index before indexing into the vertex buffer, else this means the
+                index of the first vertex to draw.
+            vertex_count: int type(ohterwise float type will be floored to int),
+                only avaliable when `indices` is not provided, which is the number
+                of vertices to draw. There are 2 cases that we will change your
+                vertex_count. [1] If the `vertex_count` is a odd number, then we
+                will change it to `vertex_count` - 1. [2] If `vertex_offset` plus
+                `vertex_count` greater than vertices.shape[0], then we will reduce
+                `vertex_count` to no more than vertices.shape[0].
```
Vissidarte-Herman, line 129: vertex_count: int type (otherwise float type will be floored to int),
                only available when `indices` is not provided, which is the number
                of vertices to draw. There are 2 cases where we will change your
                vertex_count. [1] If the `vertex_count` is an odd number, then we
                will change it to `vertex_count` - 1. [2] If `vertex_offset` plus
                `vertex_count` is greater than vertices.shape[0], then we will reduce
                `vertex_count` to no more than vertices.shape[0].

python/taichi/ui/scene.py
```diff
@@ -106,7 +116,39 @@ def lines(self,
...
+            index_offset: int type(ohterwise float type will be floored to int),
```
Vissidarte-Herman, line 130:
```suggestion
            index_offset: int type (otherwise float type will be floored to int).
```

python/taichi/ui/scene.py
```diff
@@ -106,7 +116,39 @@ def lines(self,
...
+                only avaliable when `indices` is provided, which is the base index
```
Vissidarte-Herman, line 131:
```suggestion
                Only available when `indices` is provided, which is the base index
```

python/taichi/ui/scene.py
```diff
@@ -106,7 +116,39 @@ def lines(self,
...
+                within the index buffer.
+            index_count: int type(ohterwise float type will be floored to int),
```
Vissidarte-Herman, line 133:
```suggestion
            index_count: int type (otherwise float type will be floored to int),
```

python/taichi/ui/scene.py
```diff
@@ -106,7 +116,39 @@ def lines(self,
...
+                only avaliable when `indices` is provided, which is the the number
```
Vissidarte-Herman, line 134:
```suggestion
                Only available when `indices` is provided, which is the number
```

Vissidarte-Herman (changes requested):
Minor editorial updates. LGTM!

python/taichi/ui/scene.py
```diff
@@ -106,7 +116,39 @@ def lines(self,
                 If `per_vertex_color` is provided, this is ignored.
             per_vertex_color (Tuple[float]): a taichi 3D vector field, where each
                 element indicate the RGB color of the line.
+            vertex_offset: int type(otherwise float type will be floored to int),
+                if 'indices' is provided, this means the value added to the vertex
+                index before indexing into the vertex buffer, else this means the
+                index of the first vertex to draw.
```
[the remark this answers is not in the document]
Morcki: Thanks for you suggestions!

python/taichi/ui/scene.py
Morcki: Thanks for you suggestions!

YuCrazing (approved):
LGTM!

Vissidarte-Herman (approved):

python/taichi/ui/scene.py
```diff
@@ -93,8 +93,18 @@ def lines(self,
               width,
               indices=None,
               color=(0.5, 0.5, 0.5),
-              per_vertex_color=None):
+              per_vertex_color=None,
+              vertex_offset: int = 0,
```
ailzhang, line 97: Is it better to give `vertex_offset` default to None as well, so that you have a way to know it's not set by users?

python/taichi/ui/scene.py
```diff
@@ -93,8 +93,18 @@ def lines(self,
...
+              vertex_count: int = None,
+              index_offset: int = 0,
```
ailzhang, line 99: ditto

python/taichi/ui/scene.py
```diff
@@ -106,7 +116,39 @@ def lines(self,
                 If `per_vertex_color` is provided, this is ignored.
             per_vertex_color (Tuple[float]): a taichi 3D vector field, where each
                 element indicate the RGB color of the line.
+            vertex_offset: int type(otherwise float type will be floored to int),
```
ailzhang, line 119: https://github.com/taichi-dev/taichi/blob/master/python/taichi/ui/gui.py#L20-L30 please follow this format for args
e.g. 
```
vertex_offset (int) : ...

**Review by ailzhang — COMMENTED** [review]
Thanks!

File: python/taichi/ui/scene.py (hunk 7 above)
Comment (Morcki) [reply python/taichi/ui/scene.py:97]: Here user-no-assignment behavior means no vertex offset, which means `vertex_offset` is 0.

File: python/taichi/ui/scene.py (hunk 8 above)
Comment (Morcki) [reply python/taichi/ui/scene.py:99]: The default value is 0 which is a default behavior for the user

File: python/taichi/ui/scene.py [hunk 10]
```diff
@@ -106,7 +116,39 @@ def lines(self,
...
+            vertex_count = vertices.shape[0]
+        if index_count is None:
+            if indices is None:
+                index_count = vertex_count
+            else:
+                index_count = indices.shape[0]
+        if vertex_count % 2:
+            print(
+                "Warning! Odd drawing count will be cut to neast even number")
+            vertex_count -= 1
```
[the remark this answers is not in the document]
Morcki: `floor` may refer to a float number cut to int type? 
For example, floor(3.21) = 3. But here we minus 1, so if we use `floor` may lead to a misunderstanding?

python/taichi/ui/scene.py
Morcki: Okay, I will follow that.

Morcki: Already done.

taichi/python/export_ggui.cpp
```diff
@@ -140,11 +140,20 @@ struct PyScene {
...
+             float draw_index_count,
+             float draw_first_index,
+             float draw_vertex_count,
+             float draw_first_vertex) {
     RenderableInfo renderable_info;
     renderable_info.vbo = vbo;
     renderable_info.indices = indices;
     renderable_info.has_per_vertex_color = has_per_vertex_color;
+    renderable_info.has_user_customized_draw = true;
+    renderable_info.draw_index_count = (int)draw_index_count;
```
ailzhang, line 153: OOC: why are these passed in as floats while they're mainly used as ints?

ailzhang (approved):
LGTM!
nit: good to add the example usage script in tests in a followup PR?

taichi/python/export_ggui.cpp
Morcki: Test may be diffcult to design, I will consider that in followup PR.

Here, I think user may usually use divide caculations to make a part of renderables visible, so they may forget to convert float to int type in python scope, so I convert that to int type in cpp part. In addition, the display of an index is often not that important, and the user can still control the value if they want. But if you want these args' data type to be consistent, it's also okey.
For example, there are 100 verticles which represent some particles, and the user want the middle part(1/3 ~ 2/3) to be visible. They should use like this:
```python
draw_vertex_count = int(100/3)
draw_vertex_offset = int(100/3)
```
vs.
```python
draw_vertex_count = 100/3
draw_vertex_offset = 100/3
```
Under the second situation, I think user usually do not care about if 1 particle is visible or not.

neozhaoliang (approved):
LGTM!

PR #5497, opened 2022-07-22
[bug] Let the arguments in ti.init override the environment variables

### LINKED ISSUE #5496: Environment variables override the configs set in ti.init()
**Describe the bug**
If we set an environment variable like `TI_OFFLINE_CACHE=1`, even if we do `ti.init(offline_cache=False)`, the offline cache is still on. 

**To Reproduce**

(repro.py)
```py
import taichi as ti
ti.init(offline_cache=False)
```

(repro.sh)
```
TI_OFFLINE_CACHE=1 python repro.py
```

**Log/Screenshots**

```
[W 07/22/22 13:56:27.320 727463] ti.init argument "offline_cache" overridden by environment variable TI_OFFLINE_CACHE=1
```

lin-hitonami:
Related issue = #4401, close #5496

### DIFF
--- a/python/taichi/lang/misc.py
+++ b/python/taichi/lang/misc.py
@@ -384,7 +384,7 @@ def init(arch=None,
             _ti_core.warn(
                 f'Environment variable TI_DEFAULT_IP={env_default_ip} overridden by ti.init argument "default_ip"'
             )
-        if env_default_ip == '32':
+        elif env_default_ip == '32':
             default_ip = i32
         elif env_default_ip == '64':
             default_ip = i64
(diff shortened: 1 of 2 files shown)


python/taichi/lang/misc.py
```diff
@@ -383,7 +382,7 @@ def init(arch=None,
     if env_default_ip:
         if default_ip is not None:
             _ti_core.warn(
-                f'ti.init argument "default_ip" overridden by environment variable TI_DEFAULT_IP={env_default_ip}'
+                f'Environment variable TI_DEFAULT_IP={env_default_ip} overridden by ti.init argument "default_ip"'
             )
         if env_default_ip == '32':
```
python3kgae, line 387: elif here too?

PGZXB (approved):
LGTM!

PR #5501, opened 2022-07-23
[misc] Fix deprecation: ti.Tape -> ti.ad.Tape

lucifer1004:
Since `ti.Tape` has been deprecated, we should change the related code in the code base.


ailzhang (approved):
Thanks for the fix!

PR #5505, opened 2022-07-24
[ci] Update macos runner version

SukkaW:
GitHub Action is sunsetting the macOS 10.15 Actions runner. It will stop working intermittently until being completely removed by 2022-8-30: https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/

Note: the `CI_PLATFORM` env variable is used to download the prebuilt version of LLVM, and setting it to `macos` is completely ok.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -204,7 +204,7 @@ jobs:
     needs: matrix_prep
     strategy:
       fail-fast: false
-      matrix: ${{ fromJson(needs.matrix_prep.outputs.matrix) }}
+      matrix: ${{ fromJson(needs.matrix_prep.outputs.matrix_osx) }}
     runs-on: [self-hosted, macos-10.15]
     env:
       PY: ${{ matrix.python }}
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -208,14 +208,12 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: macos-10.15
+          - os: [[self-hosted, macos-10.15], macos-latest]
             python: 3.7
             with_cc: OFF
             with_cpp_tests: ON
             wanted_archs: 'cpu'
-    runs-on:
-    - self-hosted
-    - ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
     env:
       PY: ${{ matrix.python }}
     steps:


feisuzhu:
Hi, I think it didn't work out :(

`actions/setup-python` is for github runners so we have to limit the task not to run on self-hosted runners.

feisuzhu:
Hi, since taichi needs to support macOS 10.15, I moved the CPU build task to internal buildbot, but we also need to build & test on latest macOS, so could you add a new task addressing this (if you are still interested)?

SukkaW:
> Hi, since taichi needs to support macOS 10.15, I moved the CPU build task to internal buildbot, but we also need to build & test on latest macOS, so could you add a new CI task addressing this (if you are still interested)?

Let me see what I can do on this.

SukkaW:
I have rebased the PR to the latest `master`, and modify the `strategy` to support multiple `runs-on`.

.github/workflows/release.yml
```diff
@@ -200,10 +200,12 @@ jobs:
     needs: matrix_prep
     strategy:
       fail-fast: false
-      matrix: ${{ fromJson(needs.matrix_prep.outputs.matrix) }}
-    runs-on: [self-hosted, macos-10.15]
+      matrix:
+       from_matrix_prep: ${{ fromJson(needs.matrix_prep.outputs.matrix) }}
+       runs_on: [[self-hosted, macos-10.15], macos-latest]
```
SukkaW, line 205: GitHub Action only supports multiple `runs-on` using `strategy`. In order to minimize the changes, I change the original `matrix` to `matrix.from_matrix_prep` and `matrix.runs_on`.

However, I don't know if GitHub Action even recognizes `[[self-hosted, macos-10.15], macos-latest]`.

feisuzhu: Hi, sorry for the late reply, the notification is somehow lost for no reason 😅

macOS are generally backward compatible so we don't need to build a whole matrix of wheels here (with 10.14 and the M1 chip transition as exceptions, which already addressed by `build_and_test_1014` and `build_and_test_m1` task respectively), but we do need to ensure built wheels are behaving normally on newer systems, so IMHO these should be moved to `testing.yml`

.github/workflows/testing.yml
```diff
@@ -190,7 +190,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: macos-10.15
+          - os: macos-latest
```
feisuzhu, line 193: Both `10.15` and `latest` should be tested

feisuzhu (changes requested):

SukkaW:
@feisuzhu I have changed the `testing` to run on `[[self-hosted, macos-10.15], macos-latest]` (I am still not sure if this syntax would work. If not, we might need a separate task), while the `release` is now only run on `[self-hosted, macos-10.15]`.

I have also rebased the PR to the latest `master` branch.

feisuzhu:
@SukkaW No, it's not working 😵

Pasting error message here in case you can't see it
2022-08-15-161203_942x246_scrot

feisuzhu:
I'd recommend you to tinker it in your own repo to avoid this extra long trial-and-error process.

SukkaW:
> Pasting error message here in case you can't see it ![2022-08-15-161203_942x246_scrot]
> I'd recommend you to tinker it in your own repo to avoid this extra long trial-and-error process.

I do not have a self-hosted github action runner thus I can't tinker with it in my own repo (the process will fail anyway).

The error message indicates that GitHub Action is not able to parse `[[self-hosted, macos-10.15], macos-latest]`. So here are two solutions:

- Switch to the YAML array to see if GitHub Action can identify it
- Split it into 2 separate tasks with different `runs-on`

.github/workflows/testing.yml
```diff
@@ -208,14 +208,12 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: macos-10.15
+          - os: [[self-hosted, macos-10.15], macos-latest]
             python: 3.7
             with_cc: OFF
             with_cpp_tests: ON
             wanted_archs: 'cpu'
```
feisuzhu, line 215:
```suggestion
          - os: [self-hosted, macos-10.15]
            python: 3.7
            with_cc: OFF
            with_cpp_tests: ON
            wanted_archs: 'cpu'
          - os: macos-latest
            python: 3.7
            with_cc: OFF
            with_cpp_tests: ON
            wanted_archs: 'cpu'
```

Could you please try this?
It might not work though since github hosted runners may not have brew.

feisuzhu:
Closing this PR since it no longer applies to current situation.

Issue #5463, opened 2022-07-19
Numpy and taichi operation error


neozhaoliang:
Update: just found this has already mentioned in an old issue #4052
Adding a `numpy.float64` with a `ti.Matrix` will report error.
To reproduce:

```python
import taichi as ti
import numpy as np
ti.init(arch=ti.cpu)

a = np.sqrt(1)
v = ti.Vector([1., 1.])

@ti.kernel
def test():
    x = a + v
a = np.sqrt(1)
test()
```

Error message:

```
On line 10 of file "order.py", in test:
    x = a + v
    ^^^^^^^^^
Only 0-dimensional numpy array can be used to initialize a scalar expression
```
However, interchanging the order to `v + a` the code works.

This is because the `__add__` method of `a` will return a numpy.float64 object. 
We should let operations between numpy  and taichi  return a taichi object.

PR #5506, opened 2022-07-25
[Lang] Fix numpy and taichi operations problem

neozhaoliang:
Related issue = #5463

Currently, we support:

1. In python scope operations between `np.ndarray/np.scalar` and `ti.Matrix` will return `ti.Maitrx`.
2. In Taichi scope operations between `np.scalar` and `ti.Matrix` will return `ti.Matrix`.
3. In Taichi scope if the first operand is `np.ndarray` and the second operand is `ti.Matrix`, the result is `ti.Matrix`. (The reverse is not supported)


python/taichi/lang/matrix.py
```diff
@@ -403,6 +403,16 @@ class Matrix(TaichiOperations):
         0
     """
     _is_taichi_class = True
+    __arrar_priority__ = 1000
+
+    def __array_ufunc__(self, ufunc, method, *args, **kwargs):
```
lin-hitonami, line 408: I'm not very familiar with numpy, so could you explain what this function does?

python/taichi/lang/matrix.py
```diff
@@ -403,6 +403,16 @@ class Matrix(TaichiOperations):
         0
     """
     _is_taichi_class = True
+    __arrar_priority__ = 1000
```
lin-hitonami, line 406:
```suggestion
    __array_priority__ = 1000
```

python/taichi/lang/matrix.py
neozhaoliang: It's here: https://numpy.org/neps/nep-0018-array-function-protocol.html

If the user's custom array implements this interface, NumPy will call this function to handle the operations between ndarray and user's custom arrays.

tests/python/test_numpy.py
```diff
@@ -213,3 +213,22 @@ def func2(a: ti.types.ndarray()):
...
+
+
+@test_utils.test()
+def test_numpy_op_with_matrix():
+    a = np.cos(1)
+    b = ti.Vector([1, 2])
+    c = np.array([0, 1])
+    x = a + b
+    y = c + b
+    assert isinstance(x, ti.lang.matrix.Matrix) and isinstance(
```
ailzhang, line 225:
```suggestion
    assert isinstance(x, ti.Matrix) and isinstance(
```

tests/python/test_numpy.py
```diff
@@ -213,3 +213,22 @@ def func2(a: ti.types.ndarray()):
...
+        y, ti.lang.matrix.Matrix)
```
ailzhang, line 226:
```suggestion
        y, ti.Matrix)
```

tests/python/test_numpy.py
```diff
@@ -213,3 +213,22 @@ def func2(a: ti.types.ndarray()):
...
+
+    @ti.kernel
+    def test():
+        x = a + b
```
ailzhang, line 230: can we also check on the values in `x`?

python/taichi/lang/matrix.py
```diff
@@ -403,6 +403,16 @@ class Matrix(TaichiOperations):
...
     _is_taichi_class = True
+    __array_priority__ = 1000
+
+    def __array_ufunc__(self, ufunc, method, *args, **kwargs):
+        """Make sure operations between a numpy ndarray and Taichi matrix always return a Matrix.
+        This function is called only when the first operand is a numpy ndarray.
+        """
+        if method == "__call__":
+            name = ufunc.__name__
+            return getattr(self, f"__{name}__")(args[0].tolist(), **kwargs)
```
ailzhang, line 414: would you mind adding a test with 2d numpy array with ti.Matrix?

ailzhang (commented):
Thanks!

python/taichi/lang/matrix.py
neozhaoliang: 2d test added

tests/python/test_numpy.py
```diff
@@ -215,11 +215,26 @@ def func2(a: ti.types.ndarray()):
...
+    arr2d = np.array([[0, 1], [2, 3]])
+    mat = ti.Matrix([[0, 1], [2, 3]])
 
-    x = np.array([1, 2, 3])
-    assert foo(x, 1) == 2
+    assert isinstance(scalar + vec, ti.Matrix) and isinstance(
+        mat + arr2d, ti.Matrix)
+
+    @ti.kernel
+    def test():
```
ailzhang, line 230: https://docs.taichi-lang.org/docs/reference#the-assert-statement 
IIRC assert only works when debug=True?

neozhaoliang: Thanks for this detailed review! I have added `debug=True` in the test.

neozhaoliang:
removed the support for operations between `numpy.ndarray` and `ti.Matrix` since the implementation is too hacky and does not work for binary ops like `div` and `mod`.

lin-hitonami (approved):
LGTM!

neozhaoliang:
Thanks!

PR #5507, opened 2022-07-25
[Vulkan] Fixed depth texture validation error

PENGUINLIONG:
This PR fixed a trivial validation error that storage usage of depth image is not allowed.


ailzhang (approved):
Thanks for the fix!

PR #5508, opened 2022-07-25
[bug] Add missing bracket when build for android

qiao-bo:
Fix missing bracket and resume correct function indent by pre-commit.


feisuzhu (approved):
LGTM

Issue #5462, opened 2022-07-19
Better error message for using ggui on opengl backend


neozhaoliang:
Related issue #5438 

When using `ti.ui.Window` on `arch=opengl`, it reports error

```
RuntimeError: [spirv_ir_builder.cpp:get_primitive_type@292] Type u8 not supported.
GLFW Error 65537: The GLFW library is not initialized
```

Shall we make it more specify like "use ggui on opengl backend is not supported?"

PENGUINLIONG:
Attempted to fix this problem but ended up that `set_image` wouldn't be working properly without a significent fix in the OpenGL backend (mainly to support Vulkan-OpenGL interop). Based on that the GGUI is implemented completely in Vulkan and Vulkan was introduced to replace OpenGL, if the user is able to launch GGUI I think it's better not to use the OpenGL backend. Ideally the OpenGL backend should be used only for deployment on old embedded devices.

But AFAIK there was a plan to implement Vulkan-OpenGL interop, so I'm gonna defer this fix to that point.

PR #5509, opened 2022-07-25
[Lang] Improve error message for ggui on opengl backend

neozhaoliang:
Related issue = #5462

### DIFF
--- a/python/taichi/ui/utils.py
+++ b/python/taichi/ui/utils.py
@@ -107,7 +107,7 @@ def check_ggui_availability():
         wheel_tag = try_get_wheel_tag(taichi)
         if platform.system(
         ) == "Linux" and wheel_tag and 'manylinux2014' in wheel_tag:
-            libc_ver = try_get_loaded_libc_vrsion()
+            libc_ver = try_get_loaded_libc_version()
             if libc_ver and libc_ver < (2, 27):
                 raise GGUINotAvailableException(
                     "GGUI is not available since you have installed a restricted version of taichi. "
--- a/python/taichi/ui/utils.py
+++ b/python/taichi/ui/utils.py
@@ -99,6 +99,9 @@ def check_ggui_availability():
     """Checks if the `GGUI` environment is available.
     """
     if _ti_core.GGUI_AVAILABLE:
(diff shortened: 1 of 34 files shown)


ailzhang (changes requested):
This isn't the best place to put this adhoc check, would you mind taking a look at `check_ggui_availability` and see if you can do anything over there?

neozhaoliang:
Thanks for the suggestion, the check is moved to `check_ggui_availability`.

ailzhang (commented):
Hmmm I'd prefer to have `_ti_core.GGUI_AVAILABLE  == false` in this case... so I feel like it's better to do this check in C++?

neozhaoliang:
Yes that would be better if we set `GGUI_AVAILABLE` in the c++ part since this check is used elsewhere by the python part.
I have move the check to c++, but I'm not sure if it's in the right place?

taichi/python/export_ggui.cpp
```diff
@@ -302,6 +302,10 @@ struct PyWindow {
     // todo: support other ggui backends
     if (!lang::vulkan::is_vulkan_api_available()) {
       throw std::runtime_error("Vulkan must be available for GGUI");
+    } else if (ti_arch == Arch::opengl) {
```
ailzhang, line 305: Is it better to check ti_arch is one of cpu, vulkan and cuda before checking is_vulkan_api_available so that other backends like metal also have proper error messages?

neozhaoliang: Sure, the code is updated now.

python/taichi/ui/utils.py
```diff
@@ -107,7 +107,7 @@ def check_ggui_availability():
         wheel_tag = try_get_wheel_tag(taichi)
         if platform.system(
         ) == "Linux" and wheel_tag and 'manylinux2014' in wheel_tag:
-            libc_ver = try_get_loaded_libc_version()
```
ailzhang, line 110: btw not sure if this is a typo, but if it is we probably want to fix it as well?

neozhaoliang: Sorry it's a typo!

ailzhang (approved):
LGTM!

PR #5512, opened 2022-07-25
[example] Make implicit_fem fallback to CPU when CUDA is not available

### LINKED ISSUE #5499: Exception: Backend Arch.metal doesn't support dynamic index
When running the ti example "implicit_fem", it happened to report this error message:
[Taichi] version 1.0.3, llvm 10.0.0, commit fae94a21, osx, python 3.8.2
[Taichi] Starting on arch=metal

image
image

strongoier:
Related issue = close #5499


ailzhang (approved):
Thanks!

PR #5514, opened 2022-07-25
[GUI] Shifted to a more commonly supported type for set_image

PENGUINLIONG:
This PR partially solved #5462 by replacing `ti.u8` vectors with `ti.u32`. You usually need extensions to support 8-bit or 16-bit load store on non-CUDA GPU backends so it's better to use u32 packs.

### DIFF
--- a/python/taichi/ui/staging_buffer.py
+++ b/python/taichi/ui/staging_buffer.py
@@ -1,4 +1,3 @@
-from numpy import dtype
 from taichi.lang.impl import ndarray
 from taichi.lang.kernel_impl import kernel
 from taichi.lang.matrix import Vector
--- a/python/taichi/ui/staging_buffer.py
+++ b/python/taichi/ui/staging_buffer.py
@@ -92,7 +92,7 @@ def copy_colors_to_vbo(vbo, colors):
 
 
 @ti.kernel
-def copy_image_f32_to_pack32_rgba(src: ti.template(), dst: ti.template(),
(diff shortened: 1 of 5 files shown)


python/taichi/ui/utils.py
```diff
@@ -21,6 +21,8 @@ def get_field_info(field):
         info.field_source = _ti_core.FieldSource.TaichiX64
     elif default_cfg().arch == _ti_core.vulkan:
         info.field_source = _ti_core.FieldSource.TaichiVulkan
+    elif default_cfg().arch == _ti_core.opengl:
+        info.field_source = _ti_core.FieldSource.TaichiOpenGL
```
bobcao3, line 25: Does this work now?

python/taichi/ui/staging_buffer.py
```diff
@@ -92,36 +92,41 @@ def copy_colors_to_vbo(vbo, colors):
...
+        dst[i, j] = pack
 
 
 # ggui renderer always assumes the input image to be u8 RGBA
 # if the user input is not in this format, a staging ti field is needed
 image_field_cache = {}
 
 
-def to_u8_rgba(image):
+def to_pack32_rgba(image):
```
bobcao3, line 129: I'd call it rgba8 or u8vec4 or something more informative

bobcao3 (approved):
Other than the nits, LGTM

PR #5520, opened 2022-07-26
[AOT] Added texture interfaces to C-API

PENGUINLIONG:
This PR introduced texture interfaces to C-API.

**(WIP)** Don't merge!


taichi/rhi/device.h
```diff
@@ -597,14 +556,30 @@ struct SurfaceConfig {
   uint32_t height{1};
 };
 
-struct TI_DLL_EXPORT ImageParams {
+enum class ImageAllocUsage : int {
```
bobcao3, line 559: Maybe we should build a `enum flags` class template?

taichi/rhi/vulkan/vulkan_device.cpp
```diff
@@ -1892,14 +1892,23 @@ DeviceAllocation VulkanDevice::create_image(const ImageParams &params) {
...
   if (is_depth) {
-    image_info.usage |= VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT;
+    if (params.usage & ImageAllocUsage::Attachment) {
+      image_info.usage |= VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT;
+    }
   } else {
-    image_info.usage |=
-        VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT | VK_IMAGE_USAGE_STORAGE_BIT;
+    if (params.usage & ImageAllocUsage::Storage) {
+      image_info.usage |= VK_IMAGE_USAGE_STORAGE_BIT;
```
bobcao3, line 1907: I think we do allow a depth attachment being a storage image in the device API

taichi/rhi/vulkan/vulkan_device.cpp
```diff
@@ -2405,28 +2413,14 @@ void VulkanSurface::create_swap_chain() {
...
-    view_info.subresourceRange.baseArrayLayer = 0;
-    view_info.subresourceRange.layerCount = 1;
-
-    vkapi::IVkImage image = vkapi::create_image(device_->vk_device(), img);
+    vkapi::IVkImage image = vkapi::create_image(
+        device_->vk_device(), img, surface_format.format, VK_IMAGE_TYPE_2D,
+        VkExtent3D{uint32_t(width), uint32_t(height), 1}, 1u, 1u, usage);
     vkapi::IVkImageView view =
-        vkapi::create_image_view(device_->vk_device(), image, &view_info);
+        vkapi::create_image_view(device_->vk_device(), image);
```
bobcao3, line 2420: `vkapi::` is our light-weight Vulkan C++ reference counted header wrapper, so we should make it adhereing as close to the Vulkan API as possible, this includes using the original Vulkan create info struct if it contains more stuff than we currently use. (i.e. Maybe we want to change swizzle flags, maybe we want a view that is not starting from layer 0)

bobcao3: I'd avoid making "smart" wrappers like this one here, also makes the code harder to understand

taichi/runtime/gfx/runtime.cpp
```diff
@@ -559,6 +559,17 @@ void GfxRuntime::buffer_copy(DevicePtr dst, DevicePtr src, size_t size) {
...
 
+void GfxRuntime::copy_image(DeviceAllocation dst,
+                            DeviceAllocation src,
+                            const ImageCopyParams &params) {
+  ensure_current_cmdlist();
+  transition_image(dst, ImageLayout::transfer_dst);
+  transition_image(src, ImageLayout::transfer_src);
+  current_cmdlist_->copy_image(dst, src, ImageLayout::transfer_dst,
+                               ImageLayout::transfer_src, params);
+  submit_current_cmdlist_if_timeout();
```
bobcao3, line 570: don't think we need this timeout check here

bobcao3 (approved):
otherwise LGTM

taichi/rhi/device.h
PENGUINLIONG: Sounds good, will get it refactorized later.

jim19930609:
Super! Btw it would be nice if we can write some comments on each newly-added public interface, to explain its functionality, each argument and return? Structs such as `TiTextureAllocateInfo`, `TiVulkanTextureInteropInfo` might also need some documentaries. 

Some unit tests for each public interface is also welcomed! (maybe in a separate PR)

PENGUINLIONG:
> Super! Btw it would be nice if we can write some comments on each newly-added public interface, to explain its functionality, each argument and return? Structs such as `TiTextureAllocateInfo`, `TiVulkanTextureInteropInfo` might also need some documentaries.
> 
> Some unit tests for each public interface is also welcomed! (maybe in a separate PR)

I'm working on a detailed documentation on C-API interfaces. It might take a while to be published tho. 😉

PR #5521, opened 2022-07-26
[Lang] [bug] Ensure non-i32 compatibility in while statement conditions

### LINKED ISSUE #5466: If condition must be int32 error

re-xyr:
Related issue = Fixes #5466

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -1118,7 +1118,7 @@ class ASTTransformer(Builder):
                 "'else' clause for 'while' not supported in Taichi kernels")
 
         with ctx.loop_scope_guard():
-            ctx.ast_builder.begin_frontend_while(expr.Expr(1).ptr)
+            ctx.ast_builder.begin_frontend_while(expr.Expr(1, dtype=primitive_types.i32).ptr)
             while_cond = build_stmt(ctx, node.test)
             impl.begin_frontend_if(ctx.ast_builder, while_cond)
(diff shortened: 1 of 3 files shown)


python/taichi/lang/ast/ast_transformer.py
```diff
@@ -1118,7 +1118,8 @@ def build_While(ctx, node):
                 "'else' clause for 'while' not supported in Taichi kernels")
 
         with ctx.loop_scope_guard():
-            ctx.ast_builder.begin_frontend_while(expr.Expr(1).ptr)
+            ctx.ast_builder.begin_frontend_while(
+                ti_ops.cast(1, primitive_types.i32).ptr)
```
strongoier, line 1122:
```suggestion
                expr.Expr(1, dtype=primitive_types.i32).ptr)
```
The cast can be avoided here actually.

tests/python/test_compare.py
```diff
@@ -210,5 +210,8 @@ def foo():
         for i in range(100):
             if i == 0:
                 pass
+        i = 100
+        while i != 0:
```
strongoier, line 214: Modify `if` -> `if/while` in the comment of the test?

strongoier (approved):
Thanks!

PR #5525, opened 2022-07-26
[autodiff] [test] Add more for-loop tests for forward mode

erizmr:
Related issue = #5055

### DIFF
--- a/tests/python/test_ad_for_fwd.py
+++ b/tests/python/test_ad_for_fwd.py
@@ -212,3 +212,142 @@ def test_double_for_loops():
 
     for i in range(N):
         assert f.dual[i] == 2 * i
+
+
+@test_utils.test(exclude=[ti.cc])
+def test_double_for_loops_more_nests():
+    N = 6
+    a = ti.field(ti.f32, shape=N, needs_dual=True)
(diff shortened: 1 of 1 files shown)


tests/python/test_ad_for_fwd.py
```diff
@@ -212,3 +212,142 @@ def double_for():
...
+    def complex():
+        for i in range(N):
+            weight = 2.0
+            tot = 0.0
+            tot_weight = 0.0
+            for j in range(c[i]):
+                tot_weight += weight + 1
+                tot += (weight + 1) * a[i]
+                weight = weight + 1
+                weight = weight * 4
```
ailzhang, line 288: These might be removed by compiler? maybe use atomic add instead?

ailzhang (approved):
LGTM!

tests/python/test_ad_for_fwd.py
erizmr: The local atomic add will still be demoted to `load-add-store` before the autodiff pass?

ailzhang: aha you're right!

PR #5526, opened 2022-07-26
[build] [refactor] Utilize find_cuda_toolkit and clean some target dependencies

qiao-bo:
Related issue = #4832 

- Utilize Cmake's `findCUDAToolkit` to detect cuda libs for profilers.
- Clean up some header includes for `taichi_common` target.
- Add a deprecation note to `taichi_exported_core`.
- Break a backward dependency in wasm target.

[Here](https://github.com/taichi-dev/taichi/issues/4832#issuecomment-1195088614) shows an overview of the current current dependency graph among targets in CMake.


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # The Taichi Programming Language
 #*********************************************************************
 
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.17)
```
jim19930609, line 5: `apt-get install cmake` only installs cmake with version 3.16, and I'm a bit concerned this gonna force our users to abandon the package manager...

2022-07-26 16-27-51 的屏幕截图

jim19930609 (approved):

PR #5528, opened 2022-07-26
[aot] Support texture and rwtexture in cgraph

ailzhang:
This PR works as a proof of concept for texture support in cgraph and
there might be a few adjustments and C++ tests to be added in the
followup PRs.

Run `ti example texture_graph` to see an example running cgraph with
texture arguments.

A demo in C++ that loads the saved cgraph module and run it without
python can be found in
https://github.com/ailzhang/taichi-aot-demo/blob/master/texture/desktop/texture.cpp

Related issue = #


PENGUINLIONG (approved):
lgtm

PR #5532, opened 2022-07-26
[Lang] [type] Refine SNode with quant 8/n: Replace bit_struct with ti.BitpackedFields

strongoier:
Related issue = #4857

This PR finishes the API change proposed in https://github.com/taichi-dev/taichi/issues/4857#issuecomment-1173647894. `bit_struct` is now removed from the Python side. There is still some code around `bit_struct` SNode left in C++ side, which is mainly to ease internal implementation and will no longer block future development, so we can just leave them there.


python/taichi/lang/field.py
```diff
@@ -401,4 +401,34 @@ def __init__(self, accessor, key):
...
+            max_num_bits)
+
+    def place(self, *args, shared_exponent=False):
+        """Places a list of fields with quantized types inside.
+
+        Args:
+            *args (List[Field]): A list of fields with quantized types to place.
+            shared_exponent (bool): Whether the fields have a shared exponent.
+        """
+        if shared_exponent:
```
ailzhang, line 422: nit: maybe move this to a context manager in the future? ;)

ailzhang (approved):
Nice!!

PR #5536, opened 2022-07-27
[Error] Raise error when using the struct for in python scope

### LINKED ISSUE #5535: Iterate over field via struct for in Python scope should report as error
```
import taichi as ti

ti.init(arch=ti.cpu)
N=64
temp = ti.field(ti.i32, shape=N)

for i in temp:
    print(i)
```
This code keeps printing 0s.

lin-hitonami:
Related issue = close #5535


qiao-bo (approved):
LGTM!

PR #5538, opened 2022-07-27
[bug] Add glFinish to wait_idle for opengl device

qiao-bo:
Without it, `synchronize()` in GfxRuntime cannot work properly: 
https://github.com/taichi-dev/taichi/blob/0c464321e7ff62ce0710162d12188634dad2f50f/taichi/runtime/gfx/runtime.cpp#L582

Log time in Python code after `ti.sync()` yields wrong result. This PR fixes this.


ailzhang (approved):
LGTM!

PR #5539, opened 2022-07-27
[ci] Move macOS 10.15 workloads to self-hosted runners

feisuzhu:
Related issue = #5505 #5542


ailzhang (approved):
buffering a stamp here to save CI ;)

PR #5543, opened 2022-07-27
[build] Enforce local-scoped symbols when linking with external static libraries

jim19930609:
Related issue = #


qiao-bo (approved):
LGTM! Awesome!

PR #5524, opened 2022-07-26
[misc] Enable offline cache by default (TEST)

PGZXB:
Related issue = #4401

PR #5550, opened 2022-07-27
[bug] Fix potential bug of lang::Program that could be double finalized

PGZXB:
Related issue = #4401, #5524 

### The bug is reflected in:
* https://github.com/taichi-dev/taichi/runs/7540400411?check_suite_focus=true DeviceAllocation, double deallocated
* https://github.com/taichi-dev/taichi/runs/7539283313?check_suite_focus=true Program::num_instances_ == -1, double finalized
* https://github.com/taichi-dev/taichi/runs/7538198742?check_suite_focus=true MemoryPool, double terminated

### **BUT I DON'T KNOW WHY AND WHEN `Program::finalize` WILL BE DOUBLE CALLED.**
(Only when tests are run on **multi-thread** (and with offline_cache=True) is the bug triggered.

cc: @k-ye @strongoier @lin-hitonami

### DIFF
--- a/taichi/program/program.cpp
+++ b/taichi/program/program.cpp
@@ -381,6 +381,9 @@ uint64 Program::fetch_result_uint64(int i) {
 }
 
 void Program::finalize() {
+  if (finalized_) {
+    return;
+  }
   synchronize();
 
   TI_TRACE("Program finalizing...");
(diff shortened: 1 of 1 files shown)


lin-hitonami:
The program::finalize may be called twice here.
https://github.com/taichi-dev/taichi/blob/83091292819065350a145d9509764e19055d1928/python/taichi/lang/impl.py#L370-L375

lin-hitonami (approved):
LGTM!

PR #5553, opened 2022-07-28
[build] Enforce local-scoped symbols in static llvm libs

qiao-bo:


ailzhang (commented):
Nice work! @jim19930609 would you mind double checking if this fixes the issue on your side?

jim19930609 (approved):
Verified that libtaichi_c_api.so works fine with all the LLVM symbols turned into local symbols

Issue #5533, opened 2022-07-26
Make ti.math.vec2/3/4 type notifiers, not merely functions


neozhaoliang:
when writing raytracing-like programs it's often needed to use structs like

```python
@ti.dataclass
class Ray:
    pos: vec3
    dir: vec3
```
However, our current `vec2/3/4` in the `math` module are functions, not types. I tried two ways to work around this:

1. `vec2 = ti.types.vector(2, float)`

But this will make `vec2` have `ti.f32` type, even the user manually specified `default_fp=ti.f64`.

2. Make vec2 a class:
    ```python
    class Vec2:
        def __new__(cls, *args, **kwargs):
            return ti.types.vector(2, float)(*args, **kwargs)
    ```

But then taichi will not be able to use `cook_type` to convert it to a compound type.

Any idea about this?

PR #5556, opened 2022-07-28
[Lang] Change vec2/3/4 from function calls to types

neozhaoliang:
related issue = #5533


lin-hitonami:
To let uvec always be of the unsigned type of type `default_ip`, I added the `default_up` in the `CompileConfig` to be the unsigned type of `default_ip`. (The uvec types are initialized before `ti.init` so we cannot use conditions to set it to `u32` or `u64`.)

strongoier (approved):
LGTM!

PR #5559, opened 2022-07-29
[refactor] Unify ways to set ndarray args

ailzhang:
We have various ways to set ndarray args and this PR aims to unify them
around RuntimeContex.
TODO: cleanup ways to set external array args
Related issue = #


jim19930609 (approved):
Thanks!

PR #5483, opened 2022-07-21
[Build] Remove TI_EMSCRIPTENED related code

qiao-bo:
Based on [this discussion](https://github.com/taichi-dev/taichi/issues/4832#issuecomment-1188850829). We no longer use emscripten to compile taichi. Hence remove related code.
Related issue = #4832

PR #5560, opened 2022-07-29
[bug] Fix android build for taichi-aot-demo

ailzhang:
Related issue = #
#5483 accidentally broke android build so this PR fixes it


qiao-bo (approved):
My bad, thanks for the fix!

PR #5565, opened 2022-07-29
[refactor] Unify ways to set external array args

ailzhang:
Also I introduced a typo in #5559 (`DevAllocType::kNdarray ->
DevAllocType::kNone`) and it's also fixed in this PR.

Related issue = #

### DIFF
--- a/taichi/program/context.h
+++ b/taichi/program/context.h
@@ -88,11 +88,17 @@ struct RuntimeContext {
     set_array_device_allocation_type(arg_id, DevAllocType::kRWTexture);
   }
 
-  void set_arg_external_array(int arg_id, uintptr_t ptr, uint64 size) {
+  void set_arg_external_array(int arg_id,
+                              uintptr_t ptr,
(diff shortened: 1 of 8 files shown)


jim19930609 (approved):

taichi/program/context.h
```diff
@@ -88,11 +88,17 @@ struct RuntimeContext {
     set_array_device_allocation_type(arg_id, DevAllocType::kRWTexture);
   }
 
-  void set_arg_external_array(int arg_id, uintptr_t ptr, uint64 size) {
+  void set_arg_external_array(int arg_id,
+                              uintptr_t ptr,
+                              uint64 size,
```
jim19930609, line 93: Was thinking this "size" a bit misleading. Looks like it means "number of elements" instead of "size in bytes"

ailzhang: Good point! will likely merge this with set_arg_ndarray in a followup PR and will do the batch renames then ;)

PR #5573, opened 2022-07-29
[type] Refine SNode with quant 10/n: Add validity checks and simplify BitStructType

strongoier:
Related issue = #4857

This PR enforces that at least 2 fields need to be placed when `shared_exponent=True`, and removes the redundant `member_owns_shared_exponents_` from `BitStructType`.

### DIFF
--- a/tests/cpp/ir/type_test.cpp
+++ b/tests/cpp/ir/type_test.cpp
@@ -18,18 +18,18 @@ TEST(Type, TypeToString) {
   auto qfl = TypeFactory::get_instance().get_quant_float_type(qi5, qu7, f32);
 
   auto bs1 = TypeFactory::get_instance().get_bit_struct_type(
-      u16, {qi5, qu7}, {0, 5}, {-1, -1}, {{}, {}});
+      /*physical_type=*/u16, /*member_types=*/{qi5, qu7}, /*member_bit_offsets=*/{0, 5}, /*member_exponents=*/{-1, -1}, /*member_exponent_users=*/{{}, {}});
   EXPECT_EQ(bs1->to_string(), "bs(0: qi5@0, 1: qu7@5)");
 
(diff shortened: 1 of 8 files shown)


tests/python/test_bitpacked_fields.py
```diff
@@ -213,3 +213,16 @@ def verify_val():
...
     set_val()
     verify_val()
+
+
+@test_utils.test()
+def test_invalid_place():
+    with pytest.raises(
+            ti.TaichiCompilationError,
+            match=
+            'At least 2 fields need to be placed when shared_exponent=True'):
```
ailzhang, line 223: nit: it'll be nicer to palce the context manager around the line that really throws the error so that people who look at the test cases can know immediately where it's expected to throw ;)

tests/cpp/ir/type_test.cpp
```diff
@@ -18,19 +18,18 @@ TEST(Type, TypeToString) {
   auto qfl = TypeFactory::get_instance().get_quant_float_type(qi5, qu7, f32);
 
   auto bs1 = TypeFactory::get_instance().get_bit_struct_type(
-      u16, {qi5, qu7}, {0, 5}, {false, false}, {-1, -1}, {{}, {}});
+      u16, {qi5, qu7}, {0, 5}, {-1, -1}, {{}, {}});
```
ailzhang, line 21: nit: will be nicer to add some comments around the arguments so that readers can know what they mean without looking up the header file ;)

ailzhang (approved):
LGTM!

tests/python/test_bitpacked_fields.py
strongoier: done

tests/cpp/ir/type_test.cpp
strongoier: done

ailzhang (approved):
Thanks!

PR #5576, opened 2022-07-29
[gui] Add wareframe mode for mesh & mesh_instance, add slider_int for Window.GUI.

Morcki:
**Note:**
New features:
1. Support three display modes: `Fill Mode` (fill faces), `Line Mode` (Wareframe), `Point Mode` (rendering as points).
2. Add a new widget `slider_int` in `window.GUI`, which supports slide int type value from minimum to maximum.
3. Add an arg named `instance_count` to control the count of mesh instances.

**Usage:**
```python

window = ti.ui.Window("Display Mesh", (1024, 1024), vsync=True)
canvas = window.get_canvas()
scene = ti.ui.Scene()
camera = ti.ui.make_camera()

# slider_int usage
some_int_type_value = 0
def show_options():
    global some_int_type_value

    window.GUI.begin("Display Panel", 0.05, 0.1, 0.2, 0.15)
    display_mode = window.GUI.slider_int("Value Range", some_int_type_value, 0, 5)
    window.GUI.end()

while window.running:
    
    ...
    # if to show wareframe 
    scene.mesh_instance(vertices, indices, instance_count = 100 , show_wareframe = True)
    
    canvas.scene(scene)
    show_options()
    window.show()
```


python/taichi/ui/scene.py
```diff
@@ -205,6 +206,9 @@ def mesh(self,
             index_count (int, optional):
                 only available when `indices` is provided, which is the the number
                 of vertices to draw.
+            display_mode (int, optional):
```
ailzhang, line 209: It's really error prone and hard for users to memorize these, try using an enum instead of ints?

python/taichi/ui/scene.py
```diff
@@ -301,14 +312,17 @@ def mesh_instance(self,
         if transforms and (transforms.m != 4 or transforms.n != 4):
             print("Error! Transform matrix must be 4x4 shape")
             exit()
+        if display_mode < 0 or display_mode >= 3:
+            print("Error! display_mode must be 0(Fill), 1(Line), 2(Point)")
+            exit()
```
ailzhang, line 317: usually a better way to do checks is raising errors instead of explicit print + exit

taichi/python/export_ggui.cpp
```diff
@@ -232,6 +238,7 @@ struct PyScene {
     renderable_info.draw_first_index = (int)draw_first_index;
     renderable_info.draw_vertex_count = (int)draw_vertex_count;
     renderable_info.draw_first_vertex = (int)draw_first_vertex;
+    renderable_info.display_mode = taichi::lang::PolygonMode(display_mode);
```
ailzhang, line 241: Yea this is really error prone(and why you should just expose PolygonMode enum to python instead) : imagine what if you inserted a new entry in the beginning of PolygonMode...

ailzhang (changes requested):
Thanks! Had a small suggestion w.r.t using enums instead of raw ints

python/taichi/ui/scene.py
Morcki: Sorry, I will fix that.

python/taichi/ui/scene.py
Morcki: Yeah, I add a new class named DisplayMode to support passing enum numbers in newer updates.

python/taichi/ui/scene.py
Morcki: Have changed to raise exception.

taichi/python/export_ggui.cpp
Morcki: Fixed that.

taichi/ui/common/app_config.h
```diff
@@ -3,6 +3,7 @@
 #include <string>
 #include "taichi/ui/utils/utils.h"
 #include "taichi/rhi/arch.h"
+#include "taichi/rhi/vulkan/vulkan_device.h"
```
AmesingFlank, line 6: is this change needed?

taichi/ui/backends/vulkan/renderable.cpp
```diff
@@ -48,6 +48,16 @@ void Renderable::update_data(const RenderableInfo &info) {
     prog->flush();
   }
 
+  bool is_pipeline_reset = false;
```
AmesingFlank, line 51:
```suggestion
  bool needs_pipeline_reset = false;
```

taichi/ui/backends/vulkan/renderer.h
```diff
@@ -78,7 +78,7 @@ class TI_DLL_EXPORT Renderer {
   taichi::lang::StreamSemaphore get_render_complete_semaphore();
 
  private:
-  glm::vec3 background_color_ = glm::vec3(0.f, 0.f, 0.f);
+  glm::vec3 background_color_ = glm::vec3(0.f, 0.f, 0.0f);
```
AmesingFlank, line 81: why is this needed?

Morcki: Sorry, I want to change `background_color_` to a default color, but I found that some tests without calling `clear_color` failed to pass. So later I cancled this feature. Just handwriting fault.

taichi/ui/common/app_config.h
Morcki: I will drop that change.

taichi/python/export_ggui.cpp
```diff
@@ -232,13 +239,16 @@ struct PyScene {
...
     renderable_info.draw_first_vertex = (int)draw_first_vertex;
+    renderable_info.display_mode = display_mode;
 
     MeshInfo info;
     info.renderable_info = renderable_info;
     info.color = tuple_to_vec3(color);
     info.two_sided = two_sided;
     if (transforms.valid) {
-      info.num_instances = transforms.shape[0];
+      info.num_instances = draw_instance_count > transforms.shape[0]
```
YuCrazing, line 249: Maybe  `draw_instance_count > transforms.shape[0] - instance_offset` ? or something else.

taichi/python/export_ggui.cpp
```diff
@@ -232,13 +239,16 @@ struct PyScene {
...
+                               ? transforms.shape[0]
+                               : (int)draw_instance_count;
```
YuCrazing, line 251: Maybe `transforms.shape[0] - draw_instance_count`.

YuCrazing (changes requested):

python/taichi/ui/scene.py
```diff
@@ -267,9 +294,11 @@ def mesh_instance(self,
             transforms (ti.Matrix.field, optional):
                 The Matrix must be 4x4 size with N instances, and data type should
                 be ti.f32, ti.i32, ti.u32. If None, then it behaves like raw mesh (no copy).
-            draw_first_instance (int, optional):
+            instance_offset (int, optional):
                 Default value is 0 which means no offset to show mesh instances. Otherwise,
-                the mesh instances will show from the `draw_first_instance`.
+                the mesh instances will show from the `instance_offset`.
+            instance_count (int, optional):
+                Default value is None which takes the shape value of transforms.
```
YuCrazing, line 301:
```suggestion
                The default value is None. If this parameter is not provided, instance_count = transforms.shape[0] - instance_offset.
```

YuCrazing (changes requested):

taichi/python/export_ggui.cpp
Morcki: Thanks, have solved this problem.

YuCrazing (approved):
LGTM!

taichi/ui/backends/vulkan/renderable.h
```diff
@@ -40,6 +40,7 @@ struct RenderableConfig {
   std::string fragment_shader_path;
   taichi::lang::TopologyType topology_type{
       taichi::lang::TopologyType::Triangles};
+  taichi::lang::PolygonMode polygon_type{taichi::lang::PolygonMode::Fill};
```
ailzhang, line 43:
```suggestion
  taichi::lang::PolygonMode polygon_mode{taichi::lang::PolygonMode::Fill};
```

taichi/python/export_ggui.cpp
```diff
@@ -218,11 +223,13 @@ struct PyScene {
                      py::tuple color,
                      bool two_sided,
                      FieldInfo transforms,
+                     float draw_instance_count,
                      float draw_first_instance,
                      float draw_index_count,
```
ailzhang, line 228: nit: I somehow still prefer these counts to be of int type (esp. in C++) to avoid confusing developers, but we can discuss those separately and address that in a different PR if applicable

python/taichi/ui/scene.py
```diff
@@ -13,6 +13,27 @@
 normals_field_cache = {}
 
 
+class DisplayMode_:
+    Fill = _ti_core.DisplayMode.Fill if _ti_core.GGUI_AVAILABLE else None
```
ailzhang, line 17: Your DisplayMode shouldn't be dependent on whether GGUI_AVAILABLE IMHO. The enum can be always available, whether it's usable is controlled by GGUI availability.

python/taichi/ui/scene.py
```diff
@@ -13,6 +13,27 @@
...
+    Line = _ti_core.DisplayMode.Line if _ti_core.GGUI_AVAILABLE else None
+    Point = _ti_core.DisplayMode.Point if _ti_core.GGUI_AVAILABLE else None
+
+    @staticmethod
+    def __call__(mode: int):
```
ailzhang, line 22: This doesn't make sense hmmm: it seems like you never used this call method at all...

ailzhang (changes requested):
Looks a lot better! I had a few more nits inline and then this PR is good to go

python/taichi/ui/scene.py
Morcki: I would like to use display mode through an int value with imgui to control it every frame. As below:

```python
display_mode = 0
def show_options(): # You can choose what display mode you like in realtime
    global display_mode    
    window.GUI.begin("Display Panel", 0.05, 0.1, 0.2, 0.15)
    # display_mode (0:Fill Mode, 1:Line Mode, 2:Point Mode)
    display_mode = window.GUI.slider_int("display_mode", display_mode, 0, 2)
    window.GUI.end()

scene.mesh_instance(vertices, indices, instance_count = 100 ,display_mode=ti.ui.DisplayMode(display_mode)))
```
Here, I want to use `ti.ui.DisplayMode(display_mode)` instead of `if else` blablabla.

taichi/python/export_ggui.cpp
Morcki: Ok, maybe we can talk about this deeply later.

taichi/ui/backends/vulkan/renderable.h
Morcki: Ok, I will replace that.

python/taichi/ui/scene.py
Morcki: Still need it because every type in GGUI should check it's availability. And previously, I submitted the code without checking `GGUI_AVAILABLE` failed on some platforms.
ref:
1. https://github.com/taichi-dev/taichi/blob/7a8a9283535fbe9a17ee46132acd2fa4e7555c1f/python/taichi/ui/scene.py#L75-L80
2. https://github.com/taichi-dev/taichi/blob/7a8a9283535fbe9a17ee46132acd2fa4e7555c1f/python/taichi/ui/ui.py#L29
3. https://github.com/taichi-dev/taichi/blob/7a8a9283535fbe9a17ee46132acd2fa4e7555c1f/python/taichi/ui/ui.py#L12-L24

ailzhang: For the second case about ProjectionMode I think we'd better remove the GGUI_AVAILABLE check. 
Scattering this check around isn't the best practice and we only need to check it when we create a GGUI resource like scene or camera you mentioned above (which really depends on whether it's available), but these enums can be available no matter what. 
For example, in the code snippet you showed in the pr description, you should check GGUI availability when you create a window or scene, instead of when creating a displaymode enum.

python/taichi/ui/scene.py
ailzhang: as discussed offline, a more user friendly implementation should show verbose enum name in strings instead of ints. e.g https://github.com/ocornut/imgui/blob/master/imgui_demo.cpp#L738-L747

python/taichi/ui/scene.py
PENGUINLIONG: personally speaking only the checks in `taichi.ui.Window.__init__` is necessary because the entire GGUI is nonsense without it. And if you don't have a clear idea how to pass parameters , `string` is always a go-to solution.

PENGUINLIONG (commented):
again, i still suggest you only to differentiate filled and wireframe draws, it would be a simple boolean instead of all the struggles with `DisplayMode`. I cannot come up with a case that you wanna debug meshes without the connectivity information between vertices and you still don't think `scene.particles` is a first choice. the structure of input field or ndarray ought to be the same either way.

Morcki:
> again, i still suggest you only to differentiate filled and wireframe draws, it would be a simple boolean instead of all the struggles with `DisplayMode`. I cannot come up with a case that you wanna debug meshes without the connectivity information between vertices and you still don't think `scene.particles` is a first choice. the structure of input field or ndarray ought to be the same either way.

I agree with that, I will pass a bool value to enable users to decide to show WareFrame mode or not to.

ailzhang (approved):
LGTM! Thanks!

PR #5582, opened 2022-08-01
[autodiff] [refactor] Rename the parameters to param for forward mode

erizmr:
Related issue = #

### DIFF
--- a/python/taichi/ad/_ad.py
+++ b/python/taichi/ad/_ad.py
@@ -223,13 +223,13 @@ def no_grad(func):
 
 
 class FwdMode:
-    def __init__(self, loss, parameters, seed=None, clear_gradients=True):
+    def __init__(self, loss, param, seed=None, clear_gradients=True):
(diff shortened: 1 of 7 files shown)


python/taichi/ad/_ad.py
```diff
@@ -223,13 +223,13 @@ def placeholder(*args, **kwargs):
...
 
 class FwdMode:
-    def __init__(self, loss, parameters, seed=None, clear_gradients=True):
+    def __init__(self, loss, param, seed=None, clear_gradients=True):
         self.calls = []
         self.entered = False
         self.kernels_recovered = False
         self.runtime = impl.get_runtime()
         self.loss = loss
-        self.parameters = parameters
```
ailzhang, line 232: nit: will be nicer if we can also update the parameters in the FwdMode class ;)

ailzhang (approved):
Thanks!

PR #5584, opened 2022-08-01
[doc] Removed a redundant line break to see if it will fix a CrowdIn issue

Vissidarte-Herman:
Related issue = #


ailzhang:
nit: in general there should be a more reliable way to reproduce crowdin build locally instead of trying things out :P

rexwangcc (approved):
LGTM!

Vissidarte-Herman:
> nit: in general there should be a more reliable way to reproduce crowdin build locally instead of trying things out :P

For now, we can't. But in the longer term, I do think this would be a more efficient solution.

Issue #5527, opened 2022-07-26
Potential Kernel Profiler Memory Leak


AppledoreM:
**Describe the bug**
Kernel Profiler Memory Leak

**To Reproduce**
Here is a sample to reproduce, disabling the kernel profiler fixes the issue

```py
import taichi as ti
ti.init(arch=ti.cuda, random_seed = 0, kernel_profiler=True, device_memory_GB = 4.5, dynamic_index = False)
emptyField = ti.field(ti.i32, shape=())
if __name__ == "__main__":

    frame = 0
    while True:
        if frame == emptyField[None]:
            pass
        frame += 1
```

**Log/Screenshots**
![image](https://user-images.githubusercontent.com/29791930/180954752-f8c1b59f-1ed7-4e33-b230-a2a40dd82f9f.png)

```
$ python repro.py
[Taichi] version 1.0.4, llvm 10.0.0, commit 13983c78, linux, python 3.8.10
```

qiao-bo:
If you print the kernel profiler info with `ti.profiler.print_kernel_profiler_info()` or just clear the data with `ti.profiler.clear_kernel_profiler_info()`, the occupied memory will be reset.

PR #5589, opened 2022-08-01
[cuda] Decouple update from sync function in kernel profiler

qiao-bo:
Related issue = #5527 

This PR first decouples `update()` from `sync()` in cuda kernel profiler. The memory usage issue will be followed by a separate PR.

### DIFF
--- a/python/taichi/profiler/kernel_profiler.py
+++ b/python/taichi/profiler/kernel_profiler.py
@@ -107,7 +107,7 @@ class KernelProfiler:
         #sync first
         impl.get_runtime().prog.sync_kernel_profiler()
         #then clear backend & frontend info
-        impl.get_runtime().prog.clear_kernel_profile_info()
+        impl.get_runtime().prog.clear_kernel_profiler()
         self._clear_frontend()
 
         return None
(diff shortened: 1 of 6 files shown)


python/taichi/profiler/kernel_profiler.py
```diff
@@ -199,6 +199,7 @@ def _clear_frontend(self):
     def _update_records(self):
         """Acquires kernel records from a backend."""
         impl.get_runtime().prog.sync_kernel_profiler()
+        impl.get_runtime().prog.update_kernel_profiler()
```
ailzhang, line 202: Somehow I feel this should be `prog.kernel_profiler().sync()` lol but that's a totally different issue.

ailzhang (approved):
Thanks!

python/taichi/profiler/kernel_profiler.py
qiao-bo: `sync` used to be sync+update... (reuse the old name in code), now `sync` is sole `sync`, and `update` becomes `update` ;// a bit hard to explain

PR #5592, opened 2022-08-01
[autodiff] [test] Add more complex for loop test cases for forward mode

erizmr:
Related issue = #5055


ailzhang (approved):
nit: `archs_excluded_fwd = [ti.cc, ti.opengl]` so that you can just update one line in the future? ;)

PR #5596, opened 2022-08-02
[dx12] Drop code for dx12 aot.

python3kgae:
This change add Arch::dx12 for DirectX12.
rhi, runtime and codegen are added for dx12 with minimal support to run aot.
One AOT test is added to test dxil container generation.

The codegen will not generate legal dxil yet.
More passes are needed to lower taichi generated llvm IR to DXIL in both dx12 codegen and llvm DirectX target.

TaskCodeGenLLVMDX12 inherits from TaskCodeGenLLVM to get llvm module generated.
But KernelCodeGenDX12 inherits from KernelCodeGen instead of KernalCodeGenLLVM.
Because the output should be dx container instead of llvm module.

Related issue = #5276


jim19930609:
Hi python3kgae,
Thank you so much for enabling the DX12 backend together with the AOT support! This is super!

Just for review purpose, was wondering if you may break this PR into separate smaller PRs so that it's easy to review and make comments? Thanks in advance!

jim19930609:
> Hi python3kgae, Thank you so much for enabling the DX12 backend together with the AOT support! This is super!
> 
> Just for review purpose, was wondering if you may break this PR into separate smaller PRs so that it's easy to review and make comments? Thanks in advance!

Just in case of need, you can probably refer to https://github.com/taichi-dev/taichi/issues/4800. However, there's no concrete rule on how we should split our PR.

python3kgae:
Close to split the PR.

> Hi python3kgae, Thank you so much for enabling the DX12 backend together with the AOT support! This is super!
> 
> Just for review purpose, was wondering if you may break this PR into separate smaller PRs so that it's easy to review and make comments? Thanks in advance!

Sure. I'll close this PR and make small PRs.

Issue #5540, opened 2022-07-27
[RFC][AOT] C-API for Textures


PENGUINLIONG:
Texture support is recently added to Taichi and the corresponsing C-APIs are here provided in #5520 . This issue intends to discuss the design and logstics of these interfaces. I won't dive into every aspect of the design but the most confusing and critical parts will be covered.

This issue is open to discuss, including the aspects not covered by the following text. Please feel free to leave your comments here. 🤗

## Explicit Dimension

The proposed interface requires explicitly specified texture dimensions. Zero in image extent and array layer counts is not allowed.

```cpp
// enumeration.texture_dimension
typedef enum TiTextureDimension {
  TI_TEXTURE_DIMENSION_1D = 1,
  TI_TEXTURE_DIMENSION_2D = 2,
  TI_TEXTURE_DIMENSION_3D = 3,
  TI_TEXTURE_DIMENSION_1DARRAY = 4,
  TI_TEXTURE_DIMENSION_2DARRAY = 5,
  TI_TEXTURE_DIMENSION_CUBE = 6,
  TI_TEXTURE_DIMENSION_MAX_ENUM = 0xffffffff,
} TiTextureDimension;
```

It might be okay to allow the user to specify image dimensions implicitly with zeros in image extents, but it's not compatible with cube maps. Note that although cube maps is convertible to 2D texture arrays in Vulkan and DirectX11, Metal doesn't support such usage. Also, 3D textures cannot be arrayed so explicit typing should be important.

## Explicit Format

Texture C-API introduced `TiTextureFormat` for the user to specity texel (pixel) formats. 

```cpp
// enumeration.texture_format
typedef enum TiTextureFormat {
  TI_TEXTURE_FORMAT_R8 = 0,
  TI_TEXTURE_FORMAT_R8G8 = 1,
  TI_TEXTURE_FORMAT_R8G8B8A8 = 2,
  TI_TEXTURE_FORMAT_R10G10B10A2 = 3,
  TI_TEXTURE_FORMAT_R16 = 4,
  TI_TEXTURE_FORMAT_R16G16 = 5,
  TI_TEXTURE_FORMAT_R16G16B16A16 = 6,
  TI_TEXTURE_FORMAT_R11G11B10F = 7,
  TI_TEXTURE_FORMAT_R16F = 8,
  TI_TEXTURE_FORMAT_R16G16F = 9,
  TI_TEXTURE_FORMAT_R16G16B16A16F = 10,
[...]

PR #5597, opened 2022-08-02
[Vulkan] Track image layout internally

PENGUINLIONG:
For a same reason as discussed in https://github.com/taichi-dev/taichi/issues/5540#issuecomment-1200786925 this PR propose a mechanism to track image layouts internally:

1. When an image is created or imported, it's initial state is always `VK_IMAGE_LAYOUT_UNDEFINED`; `VulkanDevice` tracks a global state for each image;
2. When the user inserts image barrier by `VulkanCommandList::image_transition`, the command list tracks the initial layout of the image when it's first referred by any command, and a pending image layout that will be realized after the execution of the command list.
3. When the user submits a command list, the initial image layout records are compared with the current records of `VulkanDevice`; errors are raised if those records mismatch, because a previously submitted command list has invalidated the layouts assumed by the current list.
4. If the layout records perfectly match, the command list is submitted and tracked layouts in `VulkanDevice` are updated with pending layouts in `VulkanCommandList`.

### DIFF
--- a/taichi/rhi/vulkan/vulkan_device.cpp
+++ b/taichi/rhi/vulkan/vulkan_device.cpp
@@ -2500,8 +2500,9 @@ void VulkanSurface::present_image(
     const std::vector<StreamSemaphore> &wait_semaphores) {
   std::vector<VkSemaphore> vk_wait_semaphores;
 
-  device_->image_transition(get_target_image(), ImageLayout::color_attachment,
-                            ImageLayout::present_src);
+  // Already transitioned to `present_src` at the end of the render pass.
+  //device_->image_transition(get_target_image(), ImageLayout::color_attachment,
+  //                          ImageLayout::present_src);
 
   for (const StreamSemaphore &sema_ : wait_semaphores) {
     auto sema = std::static_pointer_cast<VulkanStreamSemaphoreObject>(sema_);
@@ -2579,9 +2580,6 @@ DeviceAllocation VulkanSurface::get_image_data() {
     screenshot_buffer_ = device_->allocate_memory(params);
   }
 
-  device_->image_transition(img_alloc, ImageLayout::present_src,
-                            ImageLayout::transfer_src);
-
   std::unique_ptr<CommandList> cmd_list{nullptr};
 
   /*
@@ -2602,9 +2600,11 @@ DeviceAllocation VulkanSurface::get_image_data() {
   copy_params.image_extent.y = h;
   copy_params.image_aspect_flag = VK_IMAGE_ASPECT_COLOR_BIT;
   cmd_list = stream->new_command_list();
+  cmd_list->image_transition(img_alloc, ImageLayout::present_src, ImageLayout::transfer_src);
   // TODO: directly map the image to cpu memory
   cmd_list->image_to_buffer(screenshot_buffer_.get_ptr(), img_alloc,
                             ImageLayout::transfer_src, copy_params);
+  cmd_list->image_transition(img_alloc, ImageLayout::transfer_src, ImageLayout::present_src);
   /*
   if (config_.window_handle) {
     cmd_list->image_transition(screenshot_image_, ImageLayout::transfer_src,
--- a/taichi/rhi/vulkan/vulkan_device.cpp
+++ b/taichi/rhi/vulkan/vulkan_device.cpp
@@ -1977,6 +1977,10 @@ DeviceAllocation VulkanDevice::create_image(const ImageParams &params) {
         vkapi::create_image_view(device_, alloc.image, &view_info));
   }
 
+  if (params.initial_layout != ImageLayout::undefined) {
+    image_transition(handle, ImageLayout::undefined, params.initial_layout);
+  }
+
 #ifdef TI_VULKAN_DEBUG_ALLOCATIONS
   TI_TRACE("Allocate VK image {}, alloc_id={}", (void *)alloc.image,
            handle.alloc_id);
@@ -2294,6 +2298,7 @@ VulkanSurface::VulkanSurface(VulkanDevice *device, const SurfaceConfig &config)
   } else {
     ImageParams params = {ImageDimension::d2D,
                           BufferFormat::rgba8,
+                          ImageLayout::present_src,
                           config.width,
                           config.height,
                           1,
--- a/taichi/rhi/vulkan/vulkan_device.cpp
+++ b/taichi/rhi/vulkan/vulkan_device.cpp
@@ -980,12 +980,13 @@ vkapi::IVkQueryPool VulkanCommandList::vk_query_pool() {
   return query_pool_;
 }
 
-VkImageLayout VulkanCommandList::get_image_layout(const DeviceAllocation& image) {
+VkImageLayout VulkanCommandList::get_image_layout(
+    const DeviceAllocation &image) {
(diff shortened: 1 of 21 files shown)


bobcao3:
Why do you need to track the layout?

bobcao3:
For the user, you do not need to know about the layout, you can always assume they are in "undefined"

A `layout_transition(undefined, anything)` is always valid. It might be sub-optimal because the underlying access flags and stages can not be specified as detailed as possible with a undefined initial layout, but no one bothers going into that kinds of detail.

Moreover, on desktop GPUs, layouts are not a thing. They are purely here for tiled GPUs such as mobile chips. A layout transition on desktop is simply a memory barrier, so it doesn't matter which layout you specify.

bobcao3:
Furthermore, this tracking is flawed. We have async compute queues in Vulkan, and we can and do submit multiple command buffers instead of just one big one for the entire frame. If other command buffers have their layout transition run first, it will simply break this and be invalid. The order we record the command list is not guaranteed to be the same order when it's executed on the GPU.

PENGUINLIONG:
> For the user, you **do not** need to know about the layout, you can always assume they are in "undefined"
> 
> A `layout_transition(undefined, anything)` is always valid. It might be sub-optimal because the underlying access flags and stages can not be specified as detailed as possible with a undefined initial layout, but no one bothers going into that kinds of detail.
> 
> Moreover, on desktop GPUs, layouts are not a thing. They are purely here for tiled GPUs such as mobile chips. A layout transition on desktop is simply a memory barrier, so it doesn't matter which layout you specify.

I understand your concerns but your reasoning is only valid when the images are always write-only, so there is no need to *preserve existing data* during layout transition. The image layout actually cannot be assumed undefined, as described [in the Vulkan specification](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkImageLayout.html).

> VK_IMAGE_LAYOUT_UNDEFINED specifies that the layout is unknown. [...] **This layout can be used in place of the current image layout in a layout transition, but doing so will cause the contents of the image’s memory to be undefined.**

Logically, regardless of any specific implementation, the layout transition involves an inplace R/W of the existing texel data to relocate the texels in memory to serve certain needs of hardware design. So `oldLayout` guides the driver and the device to collect data from memory correctly. It might not be a problem on the devices we have tested on but we will have to pay for the divergence from the specification sooner or later. We won't be able to preserve memory data if `oldLayout` is always `undefined`. We certainly can give up tracking layouts, but that's only possible if images are always in `VK_IMAGE_LAYOUT_GENERAL`, and there is a price for it in terms of performance.

> Furthermore, this tracking is flawed. We have async compute queues in Vulkan, and we can and do submit multiple command buffers instead of just one big one for the entire frame. If other command buffers have their layout transition run first, it will simply break this and be invalid. The order we record the command list is not guaranteed to be the same order when it's executed on the GPU.

I do have considered this in the design of the tracking mechanism; and that's why I proposed double tracking on both `Device` and `CommandList` levels. The device records are only updated after a successful submit, and command lists can only be successfully submitted if the *initial layouts assumed by them* perfectly matches the device records. If we are going to support multi-queues, which is not yet implemented AFAIK, I have integrated event primitives to synchronize different queues for on-device executions, if needed. If "using `VK_IMAGE_LAYOUT_UNDEFINED` as source layout all the time" is already not conformant to the specification, I don't think it's different from tracking a wrong layout.

PENGUINLIONG:
So I have adapted the impl to track image layouts on a runtime level. @bobcao3 , any comments?

taichi/runtime/gfx/runtime.cpp
```diff
@@ -560,6 +558,13 @@ void GfxRuntime::buffer_copy(DevicePtr dst, DevicePtr src, size_t size) {
   current_cmdlist_->buffer_copy(dst, src, size);
   submit_current_cmdlist_if_timeout();
 }
+void GfxRuntime::transition_image(DeviceAllocation image, ImageLayout layout) {
+  ImageLayout &last_layout = last_image_layouts_[image.alloc_id];
+  ensure_current_cmdlist();
+  current_cmdlist_->image_transition(image, last_layout, layout);
+  submit_current_cmdlist_if_timeout();
```
bobcao3, line 565: Maybe we don't need this

taichi/runtime/gfx/runtime.cpp
```diff
@@ -560,6 +558,12 @@ void GfxRuntime::buffer_copy(DevicePtr dst, DevicePtr src, size_t size) {
   current_cmdlist_->buffer_copy(dst, src, size);
   submit_current_cmdlist_if_timeout();
 }
+void GfxRuntime::transition_image(DeviceAllocation image, ImageLayout layout) {
+  ImageLayout &last_layout = last_image_layouts_[image.alloc_id];
```
bobcao3, line 562: I would use `.at` here to reduce the chance of error. `[]` operator may result in an insert with a silent initial value, which might not be what we want.

taichi/rhi/vulkan/vulkan_device.h
```diff
@@ -669,6 +669,8 @@ class TI_DLL_EXPORT VulkanDevice : public GraphicsDevice {
 
  private:
   friend VulkanSurface;
+  friend VulkanCommandList;
+  friend VulkanStream;
```
bobcao3, line 673: Do we need this?

taichi/runtime/gfx/runtime.cpp
PENGUINLIONG: It's a desired outcome actually. Images that have not been tracked will be initialized with undefined. Or should I make it more explicit with map.find?

taichi/rhi/vulkan/vulkan_device.cpp
```diff
@@ -1977,10 +1977,6 @@ DeviceAllocation VulkanDevice::create_image(const ImageParams &params) {
         vkapi::create_image_view(device_, alloc.image, &view_info));
   }
 
-  if (params.initial_layout != ImageLayout::undefined) {
-    image_transition(handle, ImageLayout::undefined, params.initial_layout);
```
bobcao3, line 1981: Why do we skip this?

bobcao3: Although the Vulkan API always create image in undefined layout, but in our RHI we allow setting an initial layout to make it easier for the programmer

taichi/runtime/gfx/runtime.cpp
bobcao3: I'd prefer making it more explicit so that we make it easier for people down the line to maintain the code

bobcao3: Also I would think we shouldn't do this in here. The GfxRuntime should track all the resources it use. If some resource not owned by the runtime is passed in this function, something might have gone wrong

PENGUINLIONG: I see, I'm gonna delegate the allocation in `GfxRuntime`.

taichi/rhi/vulkan/vulkan_device.h
PENGUINLIONG: nice catch, forgot to bring it back

bobcao3 (approved):
LGTM

PR #5600, opened 2022-08-02
[bug] Allow ti.types.ndarray annotation for ti.func

jim19930609:
Related issue = #


ailzhang (approved):
nit: add a test? ;)

strongoier (commented):
Hmm it may cause mismatch between the type annotation and the actual function. The `ndarray` annotation can have parameters, which is not equivalent to a template.

jim19930609:
Better fix the warning message

PR #5604, opened 2022-08-02
[build] Forcing write file encoding in misc/make_changelog.py

feisuzhu:
Related issue = #


ailzhang (commented):
is the additional file intended?

ailzhang (approved):
Thanks!

PR #5605, opened 2022-08-02
[Lang] Implement index validation for matrices/vectors

### LINKED ISSUE #5588: Problems related to matrix multiplication by scalar
**Describe the bug**
When try to multiply a scalar using one row of matrix, it will get just one result value instead of one row result value.

**To Reproduce**
Here is a test script:
```py
 import taichi as ti
 
 ti.init(arch=ti.cuda)
 
 @ti.kernel
 def test():
     mat = ti.Matrix([[1, 2, 3, 4],
                      [5, 6, 7, 8]])
 
     result_wrong = ti.Matrix([[mat[0] * 2],    # 2
                               [mat[1] * 4]])  # 20

     result_correct = ti.Matrix([[mat[0,:] * 2],   # 2, 4, 6, 8
                                 [mat[1,:] * 4]]) # 20, 24, 28, 32
     print(result_wrong)
     print(result_correct)    
 
 test()
```
In terms of common sense, users may want the result of `mat[0] * 2` to be [2, 4, 6, 8] instead of just [2].

AD1024:
Related issue = fix #5588


AD1024:
Slightly more complicated fix than I thought.... Hopefully this could be done better in the refactor

python/taichi/lang/matrix.py
```diff
@@ -1431,10 +1473,13 @@ class _MatrixFieldElement(_IntermediateMatrix):
...
-            expr.Expr(ti_python_core.subscript(e.ptr, indices))
-            for e in field._get_field_members()
-        ])
+        super().__init__(
+            field.n,
+            field.m, [
+                expr.Expr(ti_python_core.subscript(e.ptr, indices))
+                for e in field._get_field_members()
+            ],
+            ndim=getattr(field, "ndim", 2))
```
jim19930609, line 1482: Was thinking if we can default the value of `ndim` to 2 inside `field` construction, instead of checking `ndim`'s existence every-time before usage?

python/taichi/lang/struct.py
```diff
@@ -698,6 +705,8 @@ def filled_with_scalar(self, value):
         return Struct(entries)
 
     def field(self, **kwargs):
+        if "ndim" in kwargs:
+            kwargs.pop("ndim")
```
jim19930609, line 709: Looks like `ndim` is an unexpected parameter in this case? Maybe we should throw an error instead?

python/taichi/lang/impl.py
```diff
@@ -45,13 +45,13 @@ def expr_init(rhs):
     if rhs is None:
         return Expr(get_runtime().prog.current_ast_builder().expr_alloca())
     if isinstance(rhs, Matrix) and (hasattr(rhs, "_DIM")):
-        return type(rhs)(*rhs.to_list())
+        return type(rhs)(*rhs.to_list(), ndim=rhs.ndim)
```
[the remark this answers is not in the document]
AD1024: Good question... My guess is that someone might want to implement Vector as a subtype of Matrix? but we definitely can assume there's only `Matrix` for now. Good catch!

python/taichi/lang/struct.py
AD1024: This is the case here:
https://github.com/taichi-dev/taichi/blob/9d6adb29ae7fb0b5d15234051d1f2063d8866e98/python/taichi/lang/struct.py#L328-L334

this could be any field
Maybe I could change it to checking whether it's `Struct` field, if so just ignore `ndim`.

python/taichi/lang/matrix.py
```diff
@@ -435,12 +444,15 @@ def __init__(self, arr, dt=None, suppress_warning=False, is_ref=False):
                     dt = initializer.infer_dt(arr)
                 local_tensor_proxy, mat = initializer.with_dynamic_index(
                     arr, dt)
-
         self.n, self.m = len(mat), 1
         if len(mat) > 0:
             self.m = len(mat[0])
         entries = [x for row in mat for x in row]
 
+        if ndim is not None:
```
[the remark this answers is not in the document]
AD1024: Yes, this will result in an implicit reshape.

python/taichi/lang/matrix.py
AD1024: iirc `ndim` is only used in Vector and Matrix fields, so when `field` is a StructField, then there's no need to declare an `ndim` for it.

jim19930609: I see... I think the root reason is that essentially `ndim` is not part of the attributes for a `field` object.

Let's keep it this way for now, and try to resolve it in the refactored Matrix/Vec implementation.

python/taichi/lang/struct.py
jim19930609: Yup, checking `Struct` field sounds better to me since it's a more defined behaviour. Dropping `ndim` regardlessly sounds pretty dangerous though

python/taichi/lang/matrix.py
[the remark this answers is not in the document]
jim19930609: I think I'm more concerned that the `m` and `n` of a Matrix is not consistent with its `ndim`. For example:
```
M = ti.Matrix([[0, 1], [2, 3], [4, 5]], ti.i32, ndim=1)
print(M.n) # 3
print(M.m) # 2
```
Looks like we end up having a 1-dimensional Matrix with `n = 3` and `m = 2`?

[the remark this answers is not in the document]
AD1024: Exactly. Essentially `ndim` is used by `ti.Vector` to propagate dimensional information for the checker in `__getitem__`. This probably couldn't be distinguished from user-level function calls...

jim19930609 (approved):

PR #5606, opened 2022-08-03
[gui] Fix ti.ui.Space in GGUI

### LINKED ISSUE #5599: ti.ui.SPACE might be confusing in GGUI
**Describe the bug**
ti.ui.SPACE does not work to match key events in GGUI.

**To Reproduce**

```py
import taichi as ti

ti.init(arch=ti.cuda)

res = (512, 512)
window = ti.ui.Window("cloth", res, vsync=True)

while window.running:
    for event in window.get_events(ti.ui.PRESS):
        print('event.key =', event.key)
        if event.key in [ti.ui.SPACE]:
            print('space catched')
        if event.key in [' ']:
            print('catched as \' \'')
    window.show()
```

**Log/Screenshots**
When space is pressed, `print('catched as \' \'')` will be executed, but `print('space catched')` will not be executed.

BillXu2000:
Fix #5599


PENGUINLIONG (approved):
lol, lgtm

PR #5615, opened 2022-08-03
[aot] Rewrite mpm88 aot test with C-API

jim19930609:
Related issue = #


cmake/TaichiCAPITests.cmake
```diff
@@ -17,12 +17,25 @@ add_executable(${C_API_TESTS_NAME} ${TAICHI_C_API_TESTS_SOURCE})
...
 target_link_libraries(${C_API_TESTS_NAME} PRIVATE taichi_c_api)
 target_link_libraries(${C_API_TESTS_NAME} PRIVATE gtest_main)
+target_link_libraries(${C_API_TESTS_NAME} PRIVATE ${llvm_libs})
 
 target_include_directories(${C_API_TESTS_NAME}
   PRIVATE
     ${PROJECT_SOURCE_DIR}
     ${PROJECT_SOURCE_DIR}/c_api/include
     ${PROJECT_SOURCE_DIR}/c_api/src
+    ${PROJECT_SOURCE_DIR}/c_api/tests
```
k-ye, line 27: OOC, with C-API, why do we still need these headers?

ailzhang (commented):
@jim19930609 can we keep the original cgraph test as well to prevent it from being broken? We'll eventually migrate that to c-api as well.

cmake/TaichiCAPITests.cmake
jim19930609: hmmm, this is an interesting question. 

To make a long story short, right now GUI is not compatible with C-APi Vulkan backend, therefore I need to use "VulkanDevice" to workaround this issue when writing the Mpm88 Demo: https://github.com/taichi-dev/taichi-aot-demo/tree/master/mpm88_desktop.

The build system for CAPI tests will get better once GUI get merged into C-API

jim19930609: okay I give up, let's not keep the work around in C-API test

c_api/tests/mpm88_test.cpp
```diff
@@ -0,0 +1,237 @@
...
+      ti_launch_kernel(runtime_, k_substep_p2g_, 6, &k_substep_p2g_args_[0]);
+      ti_launch_kernel(runtime_, k_substep_update_grid_v_, 2,
+                       &k_substep_update_grid_v_args_[0]);
+      ti_launch_kernel(runtime_, k_substep_g2p_, 6, &k_substep_g2p_args_[0]);
+    }
+    ti_wait(runtime_);
+  }
+
+ private:
+  class NdarrayAndMem {
```
ailzhang, line 106: for followups: shall we move this into c-api since it's pretty commonly useful? ;)

ailzhang (approved):
Thanks!!

PR #5617, opened 2022-08-03
[Lang] Add parameter 'order' to specify layout for scalar, vector, matrix fields

strongoier:
Related issue = #5491

This PR adds support for the following syntax as described in #5491:
```python
a = ti.field(ti.i32, shape=(128, 64, 32), order='kij')
b = ti.Vector.field(3, ti.i32, shape=(128, 64), order='ji', layout=ti.Layout.SOA)
```

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -1069,6 +1069,7 @@ class Matrix(TaichiOperations):
               m,
               dtype,
               shape=None,
+              order=None,
               name="",
               offset=None,
               needs_grad=False,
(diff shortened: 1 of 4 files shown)


python/taichi/lang/matrix.py
```diff
@@ -1136,43 +1138,68 @@ def field(cls,
         impl.get_runtime().matrix_fields.append(entries)
 
         if shape is None:
-            assert offset is None, "shape cannot be None when offset is being set"
-
-        if shape is not None:
+            if offset is not None:
+                raise TaichiSyntaxError(
+                    'Shape cannot be None when offset is set')
```
k-ye, line 1143: Super nit: `shape`, `offset` are param names, so let's keep it all lowercases.

k-ye (approved):
LGTM!

python/taichi/lang/matrix.py
strongoier: Fixed

yuanming-hu (approved):
Amazing job! LGTM!

PR #5623, opened 2022-08-04
[cuda] Reduce kernel profiler memory usage

qiao-bo:
Related issue = #5527 

As described in the issue, kernel profiler takes substantial ram usage to log execution. ~0.15GB/s increase, which is not acceptable. The core reason is that all the cuda events used to log the time are not destroyed until the last update. This PR fixes this by moving the event destroy immediately after kernel execution. Locally tested the example in the issue no longer increase ram usage.


ailzhang (approved):
Thanks a ton for fixing this!

PR #5626, opened 2022-08-04
[ci] Only run portion of tests on PR draft

feisuzhu:
Related issue = #5545

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -11,8 +11,21 @@ concurrency:
 
 env:
   TI_CI: "1"
+  TI_LITE_TEST: ${{ github.event_name == 'pull_request' && ! contains(github.event.pull_request.labels.*.name, 'error-prone') && '1' || '' }}
 
 jobs:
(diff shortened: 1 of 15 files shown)


.github/workflows/testing.yml
```diff
@@ -427,7 +447,7 @@ jobs:
         env:
           PY: py38
           GPU_TEST: ON
-          TI_WANTED_ARCHS: 'cpu,cuda,vulkan,opengl'
```
qiao-bo, line 430: we don't need cpu here because?

qiao-bo (approved):
LGTM!

qiao-bo:
btw, how much time has been reduced here by enabling the lite option?

.github/workflows/testing.yml
feisuzhu: Because CPU are tested in dedicated cpu tasks

feisuzhu:
> btw, how much time has been reduced here by enabling the lite option?

Halved :)

PR #5640, opened 2022-08-05
[ci] Drop py36 in nightly and release

ailzhang:
We've announced in v1.0.0 release note that py36 will be dropped in
v1.1.0 so let's do it :D
https://github.com/taichi-dev/taichi/releases/tag/v1.0.0

Related issue = #


qiao-bo (approved):
LGTM! A quick grep shows some doc sites need adaption as well...

PR #5641, opened 2022-08-05
[aot] Add helper function to construct Ndarray in C-API

jim19930609:
Related issue = #


PENGUINLIONG:
i wonder if we could ship c++ wrappings directly?

jim19930609:
> i wonder if we could ship c++ wrappings directly?

Moved helper functions/structs to `c_api_test_utils.h` and `c_api_test_utils.cpp`, which won't be included in the release of `libtaichi_c_api.so`

ailzhang (approved):
Awesome thanks!

PR #5646, opened 2022-08-05
[Doc] Add doc about offline cache

PGZXB:
Related issue = #4401

Thank @turbo0628 and @lin-hitonami


docs/lang/articles/performance/performance.md
```diff
@@ -223,3 +223,26 @@ we found BLS to be more effective for caching the destinations of the atomic ope
 
 As a rule of thumb, run benchmarks to decide whether to enable BLS or not.
 :::
+
+## Offline Cache
+
+A taichi kernel is implicitly compiled the first time it is invoked. The compilation yieldings are kept in an "online" in-memory cache to reduce the overhead in subsequent function calls. As long as the kernel function is unchanged, it can be directly loaded and launched. However, the cache is no longer availabe when the entire program runs down to its termination. Taichi re-compiles all the kernel functions and constructs the "online" in-memory cache when the user runs the program again. The first launch of a Taichi function is always slow due to compilation overhead.
+
+We address the problem by introducing the "offline" cache feature, that could dump and save the compilation cache on disk for future runs. The first launch overhead can be drastically reduced in repeated runs. Taichi now automatically constructs and maintains the offline cache by default. We also provide several options in `ti.init` to configure the detailed behaviors of offline cache.
+* `offline_cache: bool`: Enable or disable offline cache. Default: `True`.
```
Vissidarte-Herman, line 232:
```suggestion
* `offline_cache: bool`: Enables or disables offline cache. Default: `True`.
```

docs/lang/articles/performance/performance.md
```diff
@@ -223,3 +223,26 @@ we found BLS to be more effective for caching the destinations of the atomic ope
...
+* `offline_cache_file_path: str`: The offline cache file directory. Default: `'C:\taichi_cache\ticache\'` on Windows and `'~/.cache/taichi/ticache/'` on unix-like systems. Directories are created automatically.
+* `offline_cache_max_size_of_files: int32`: Max size of cache files in bytes. Trigger cleaning process when the cache file size exceeds the given capacity. Default: 100MB.
+* `offline_cache_cleaning_policy: str`: The policy used to replace out-of-date files in the cache. Optional polices: `'never'`, `'version'`, `'lru'` or `'fifo'`. Default: `'lru'`.
```
Vissidarte-Herman, line 235:
```suggestion
* `offline_cache_cleaning_policy: str`: Policy about how to replace outdated files in the cache. Options: `'never'`, `'version'`, `'lru'` and `'fifo'`. Default: `'lru'`.
```

docs/lang/articles/performance/performance.md
```diff
@@ -223,3 +223,26 @@ we found BLS to be more effective for caching the destinations of the atomic ope
...
+  * `'never'`: Never clean, keep all the cache files. Ignore the `offline_cache_max_size_of_files` configuration;
```
Vissidarte-Herman, line 236:
```suggestion
  * `'never'`: Never cleans and keeps all the cached files regardless of the `offline_cache_max_size_of_files` configuration;
```

docs/lang/articles/performance/performance.md
```diff
@@ -223,3 +223,26 @@ we found BLS to be more effective for caching the destinations of the atomic ope
...
+  * `'version'`: Only discard old-version cache files with respect to the kernel function;
+  * `'lru'`: Discard the cache files that are least recently used;
```
Vissidarte-Herman, line 238:
```suggestion
  * `'lru'`: Discards the cached files least used recently;
```

docs/lang/articles/performance/performance.md
```diff
@@ -223,3 +223,26 @@ we found BLS to be more effective for caching the destinations of the atomic ope
...
+  * `'fifo'`: Discard the cache files that are earliest added.
```
Vissidarte-Herman, line 239:
```suggestion
  * `'fifo'`: Discards the cached files added in the earliest.
```

docs/lang/articles/performance/performance.md
```diff
@@ -223,3 +223,26 @@ we found BLS to be more effective for caching the destinations of the atomic ope
...
+* `offline_cache: bool`: Enable or disable offline cache. Default: `True`.
+* `offline_cache_file_path: str`: The offline cache file directory. Default: `'C:\taichi_cache\ticache\'` on Windows and `'~/.cache/taichi/ticache/'` on unix-like systems. Directories are created automatically.
+* `offline_cache_max_size_of_files: int32`: Max size of cache files in bytes. Trigger cleaning process when the cache file size exceeds the given capacity. Default: 100MB.
+* `offline_cache_cleaning_policy: str`: The policy used to replace out-of-date files in the cache. Optional polices: `'never'`, `'version'`, `'lru'` or `'fifo'`. Default: `'lru'`.
+  * `'never'`: Never clean, keep all the cache files. Ignore the `offline_cache_max_size_of_files` configuration;
+  * `'version'`: Only discard old-version cache files with respect to the kernel function;
```
Vissidarte-Herman, line 237:
```suggestion
  * `'version'`: Discards only the old-version cached files with respect to the kernel function;
```

docs/lang/articles/performance/performance.md
```diff
@@ -223,3 +223,26 @@ we found BLS to be more effective for caching the destinations of the atomic ope
 
 As a rule of thumb, run benchmarks to decide whether to enable BLS or not.
 :::
+
+## Offline Cache
+
+A taichi kernel is implicitly compiled the first time it is invoked. The compilation yieldings are kept in an "online" in-memory cache to reduce the overhead in subsequent function calls. As long as the kernel function is unchanged, it can be directly loaded and launched. However, the cache is no longer availabe when the entire program runs down to its termination. Taichi re-compiles all the kernel functions and constructs the "online" in-memory cache when the user runs the program again. The first launch of a Taichi function is always slow due to compilation overhead.
+
+We address the problem by introducing the "offline" cache feature, that could dump and save the compilation cache on disk for future runs. The first launch overhead can be drastically reduced in repeated runs. Taichi now automatically constructs and maintains the offline cache by default. We also provide several options in `ti.init` to configure the detailed behaviors of offline cache.
```
Vissidarte-Herman, line 231:
```suggestion
We address this problem by introducing the *offline* cache feature, which dumps and saves the compilation cache on disk for future runs. The first launch overhead can be drastically reduced in repeated runs. Taichi now constructs and maintains an offline cache by default, as well as providing several options in `ti.init()` for configuring the offline cache behavior.
```

Vissidarte-Herman: I'm not quite certain about "The first launch overhead can be drastically reduced in repeated runs.". Could I ask you to elaborate on this one? I thought the first launch overhead would stay unchanged and would not be reduced?

docs/lang/articles/performance/performance.md
```diff
@@ -223,3 +223,26 @@ we found BLS to be more effective for caching the destinations of the atomic ope
 
 As a rule of thumb, run benchmarks to decide whether to enable BLS or not.
 :::
+
+## Offline Cache
+
+A taichi kernel is implicitly compiled the first time it is invoked. The compilation yieldings are kept in an "online" in-memory cache to reduce the overhead in subsequent function calls. As long as the kernel function is unchanged, it can be directly loaded and launched. However, the cache is no longer availabe when the entire program runs down to its termination. Taichi re-compiles all the kernel functions and constructs the "online" in-memory cache when the user runs the program again. The first launch of a Taichi function is always slow due to compilation overhead.
```
Vissidarte-Herman, line 229: A Taichi kernel is implicitly compiled the first time it is called. The compilation results are kept in an *online* in-memory cache to reduce the overhead in the subsequent function calls. As long as the kernel function is unchanged, it can be directly loaded and launched. The cache, however, is no longer available when the program terminates. Then, if you run the program again, Taichi has to re-compile all kernel functions and reconstruct the *online* in-memory cache. And the first launch of a Taichi function is always slow due to the compilation overhead.

Vissidarte-Herman: Could we just omit "in-memory"?

Vissidarte-Herman (changes requested):
LGTM! Minor editorial updates.

docs/lang/articles/performance/performance.md
PGZXB: The overhead of "the first launch" includes compilation and the truly 1st run. If the launched Taichi kernel's compilation data was saved on disk, the next "the first launch" overhead would be reduced because of **offline cache**.

docs/lang/articles/performance/performance.md
PGZXB: IMO, maybe "in-memory" can underline the online cache that Taichi had is on the memory level, but the new feature, "offline cache," is on the filesystem level? WDYT?

Vissidarte-Herman:
"in-memory" cache makes a good contrast to "offline" cache. Thanks for shedding the light on this.

Vissidarte-Herman:
Re: "The overhead of "the first launch" includes compilation and the truly 1st run. If the launched Taichi kernel's compilation data was saved on disk, the next "the first launch" overhead would be reduced because of offline cache." 

Thanks for the clarification. I see your point. The first launch overhead can be recycled for future first runs with "offline" cache. Would suggest a better version later, but let's stay with the current version for now!

Vissidarte-Herman (approved):
LTGM!

PR #5651, opened 2022-08-05
[bug] Fix bug that kernel names are not correctly captured by the profiler

### LINKED ISSUE #5643: kernel names are not correctly captured by the profiler
**Describe the bug**
kernel names are not correctly captured by the profiler.

**To Reproduce**

```py
def run_taichi_matv_jvp(N_loss, N_param=10, repeats=10):
    import taichi as ti
    ti.init(arch=ti.cuda, device_memory_GB=16, kernel_profiler=True)

    params = ti.field(float, shape=N_param, needs_dual=True)
    mat = ti.field(float, shape=(N_loss, N_param), needs_dual=True)
    losses = ti.field(float, shape=N_loss, needs_dual=True)

    mat.fill(1.0)
    params.fill(1.0)

    @ti.kernel
    def matvec_tall():
        for i in range(N_loss):
            acc = 0.0
            for j in range(N_param):
                acc += mat[i, j] * params[j]
            losses[i] = acc

    with ti.ad.FwdMode(loss=losses, param=params, seed=[1.0 for _ in range(N_param)]):
        matvec_tall()  # compile

    ti.profiler.print_kernel_profiler_info()
    ti.profiler.clear_kernel_profiler_info()

    for _ in range(repeats):
        with ti.ad.FwdMode(loss=losses, param=params, seed=[1.0 for _ in range(N_param)]):
            matvec_tall()

    ti.profiler.print_kernel_profiler_info()

    query_result = ti.profiler.query_kernel_profiler_info(matvec_tall.__name__)
    print("kernel executed times =", query_result.counter)
    print("kernel elapsed time(min_in_ms) =", query_result.min)
    print("kernel elapsed time(max_in_ms) =", query_result.max)
    print("kernel elapsed time(avg_in_ms) =", query_result.avg)

run_taichi_matv_jvp(N_loss=1024)
```

**Log/Screenshots**
image

**Additional comments**

PGZXB:
Related issue = fixes #5643

### DIFF
--- a/taichi/analysis/offline_cache_util.cpp
+++ b/taichi/analysis/offline_cache_util.cpp
@@ -179,9 +179,11 @@ constexpr std::size_t min_mangled_name_length = offline_cache_key_length + 2;
 
 std::string mangle_name(const std::string &primal_name,
                         const std::string &key) {
-  TI_ASSERT(key.size() == offline_cache_key_length);
   // Result: {primal_name}{key: char[65]}_{(checksum(primal_name)) ^
   // checksum(key)}
+  if (key.size() != offline_cache_key_length) {
+    return primal_name;
+  }
   std::size_t checksum1{0}, checksum2{0};
   for (auto &e : primal_name) {
     checksum1 += std::size_t(e);
--- a/taichi/analysis/offline_cache_util.cpp
+++ b/taichi/analysis/offline_cache_util.cpp
@@ -30,6 +30,7 @@ static std::vector<std::uint8_t> get_offline_cache_key_of_compile_config(
   serializer(config->demote_dense_struct_fors);
   serializer(config->advanced_optimization);
   serializer(config->constant_folding);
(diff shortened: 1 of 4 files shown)


lin-hitonami (approved):
Per offline discussion, it's hard to determine whether a task is cached because the profiler cannot access the kernel information. It needs many refactoring to let the profiler know whether the task is cached. This is a temporary solution that enables the profiler to deduce the real name of a kernel from the mangled name. Special characters cannot be used because they cause error on windows. See https://github.com/taichi-dev/taichi/blob/28c7cc8a2910981f5fe0632f7a191c404cb91caf/taichi/runtime/cuda/jit_cuda.cpp#L52-L72 . Therefore for now we use the checksum to determine whether the name is mangled or not.

ailzhang:
@PGZXB could you cherrypick this commit into rc-v1.1.0 and send another PR against that branch as well?  Thanks! :D

PR #5653, opened 2022-08-06
[lang] Make vector a real class

PENGUINLIONG:
This PR makes `Vector` an actual class so that code completion could work. Not sure why it was implemented in such way but it feels better not to look up docs all the time.

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -1203,11 +1203,6 @@ class Matrix(TaichiOperations):
         return entries
 
     @classmethod
-    def _Vector_field(cls, n, dtype, *args, **kwargs):
-        """ti.Vector.field"""
-        return cls.field(n, 1, dtype, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


python/taichi/lang/matrix.py
```diff
@@ -1755,6 +1745,83 @@ def field(self, **kwargs):
...
 
 
+class VectorType(MatrixType):
+    def __init__(self, n, dtype):
+        super().__init__(n, 1, dtype)
+
+    def __call__(self, *args):
+        """Return a vector matching the shape and dtype.
+
+        This function will try to convert the input to a `n x m` matrix, with n, m being
```
strongoier, line 1755: Should we replace `matrix` with `vector` in docstrings?

PENGUINLIONG: nice catch, thx

strongoier (approved):
LGTM!

PR #5655, opened 2022-08-06
[gui] Support set_image with texture

PENGUINLIONG:
This PR supported `set_image` with textures.

### DIFF
--- a/python/taichi/ui/staging_buffer.py
+++ b/python/taichi/ui/staging_buffer.py
@@ -1,6 +1,7 @@
 from taichi.lang.impl import ndarray
 from taichi.lang.kernel_impl import kernel
 from taichi.lang.matrix import Vector
+from taichi.lang._texture import Texture
 from taichi.types.annotations import template
 from taichi.types.primitive_types import f32, u8, u32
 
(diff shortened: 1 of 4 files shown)


python/taichi/ui/staging_buffer.py
```diff
@@ -122,7 +134,9 @@ def copy_image_u8_to_rgba8(src: ti.template(), dst: ti.template(),
 
 
 def to_rgba8(image):
-    if not hasattr(image, 'n') or image.m != 1:
+    is_texture = isinstance(image, Texture)
+
+    if not is_texture and (not hasattr(image, 'n') or image.m != 1):
         raise Exception(
             'the input image needs to be a Vector field (matrix with 1 column)'
```
ailzhang, line 141: update this error message as well? ;)

ailzhang (approved):
nit: can we also add a test without showing ?

qiao-bo:
@PENGUINLIONG @ailzhang Just curious, is this `x.py` intended to be there?(top level on master) ;// and the reason is?

PENGUINLIONG:
> @PENGUINLIONG @ailzhang Just curious, is this `x.py` intended to be there?(top level on master) ;// and the reason is?

@qiao-bo oof no，plz del it 🤡

PR #5658, opened 2022-08-07
[gui] GGUI scene APIs are broken

PENGUINLIONG:
The `ambient_light` example is broken:

```python
scene = ti.ui.Scene()
scene.ambient_light([0.2, 0.2, 0.2])
```

which gives:

```log
ambient_light(): incompatible function arguments. The following argument types are supported:
    1. (self: taichi._lib.core.taichi_python.PyScene, arg0: tuple) -> None

Invoked with: <taichi._lib.core.taichi_python.PyScene object at 0x7f7c810ad7b0>, [0.2, 0.2, 0.2]
```

`point_light` has a similar issue.

Also note that if a scene doesn't have any lighting except for the ambient light, no lighting SSBO will be allocated and `Canvas.scene` will crash attempting to map this SSBO. Tried not to map when `point_lights.size() == 0` but it doesn't work. So workaround atm.

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -25,8 +25,8 @@ from taichi.lang.snode import SNode
 from taichi.lang.struct import Struct, StructField, _IntermediateStruct
 from taichi.lang.util import (cook_dtype, get_traceback, is_taichi_class,
                               python_scope, taichi_scope, warning)
-from taichi.types.primitive_types import (all_types, f16, f32, f64, i32, i64, u8,
-                                          u32, u64)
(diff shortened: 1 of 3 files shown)


PENGUINLIONG:
BTW when there is no point light, scene rendering doesn't work.

k-ye:
Should this be cherrypicked into V1.1? (cc @ailzhang @feisuzhu )

ailzhang (approved):
Thanks! @PENGUINLIONG would you mind sending another PR against rc-v1.1.0 branch as well?

PR #5659, opened 2022-08-07
[gui] Support numpy vertex array for GGUI

PENGUINLIONG:
GGUI caches vertex buffer with the hash of taichi field as key, `np.ndarray` is non-hashable but can be used as kernel argument. This PR added support for `np.ndarray` to be used as a vertex data source for better prototyping experience. After this change, VBOs will be cached by their number of vertices.


taichi-gardener:

PR #5676, opened 2022-08-08
[build] [refactor] Decouple GUI source files from taichi_core target

qiao-bo:
Related issue = #4832 

- Define `taichi_ui` and `taichi_ui_vulkan` targets
- Move `taichi/gui` files into `taichi/ui`
- Move `field_info` source files to `taichi_core` to break circular dependency between `taichi_core` and `taichi_ui`.

At the moment we only decouple source files for ease of AOT testing. The compilation dependency remain unchanged, namely the CPU GUI files (now under `taichi/ui/gui`) is always compiled and linked, while the GPU GGUI files (under `taichi/ui` except `gui`) are guarded by `TI_WITH_GGUI`. Later we can also separate the compilation of GUI files.


jim19930609 (approved):
Many thanks!

jim19930609 (approved):

PR #5677, opened 2022-08-08
[ci] Confine show_environ task to Linux bots

feisuzhu:
Related issue = #


qiao-bo (commented):
ehmm, why do we need this again? ;//

feisuzhu:
> ehmm, why do we need this again? ;//

The task maybe scheduled to a Windows machine and there is no `env` command in PowerShell. It's a bug fix.

qiao-bo (approved):

PR #5679, opened 2022-08-08
Revert "[ci] Only run portion of tests on PR draft"

ailzhang:
Reverts taichi-dev/taichi#5626 since it broke master


qiao-bo:
do we still need to revert since #5670 is fixed?

feisuzhu:
> do we still need to revert since #5670 is fixed?

No :)

PR #5681, opened 2022-08-09
[llvm] Cherry picking `compile serially` (temporary)

feisuzhu:
* [llvm] Compile serially when num_thread=0

* fix test

Related issue = #


feisuzhu:
So M1 test failed, not merging

PR #5694, opened 2022-08-09
[metal] Fixed synchronization for Metal

PENGUINLIONG:
The Metal backend doesn't wait on all submitted command buffers. If a command buffer is not launched via `submit_sync`, the command buffer might not be awaited correctly.


k-ye:
This is a bit weird. IIRC, Metal backend hasn't switched to unified device API impl yet...

Issue #5555, opened 2022-07-28
opengl backend python3 segmentation fault (core dumped)


erizmr:
**Describe the bug**
The program crashed on opengl backend with a python3 segmentation fault (core dumped).

**To Reproduce**

```py
import taichi as ti
ti.init(arch=ti.opengl)

x = ti.field(dtype=ti.f32, shape=())

@ti.kernel
def mixed_inner_loops():
    for i in range(2):
        x[None] += 1
        for j in range(2):
            x[None] += 1

mixed_inner_loops()

```

**Log/Screenshots**

```
[Taichi] version 1.0.5, llvm 10.0.0, commit 36855941, linux, python 3.8.10
[Taichi] Starting on arch=opengl
[1]    2318755 segmentation fault (core dumped)  python3 fwd_opengl_test.py

```

**Additional comments**
gbd debug backtrace
```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
0x00007fffcd199497 in spvtools::opt::VectorDCE::HasVectorResult(spvtools::opt::Instruction const*) const ()
   from /home/mingrui/workspace/taichi_dev/taichi/python/taichi/_lib/core/taichi_python.cpython-38-x86_64-linux-gnu.so

```

neozhaoliang:
@erizmr Simplified your code a bit.

ailzhang:
FYI I've pinpointed the root cause in spv code that triggers this and working on a fix. (in the worst case I can send out a workaround :P

PR #5696, opened 2022-08-09
[autodiff] [test] Recover the forward mode test cases

erizmr:
The test case for forward mode autodiff work after #5555 fixed. 
Related issue #5708


ailzhang (approved):
Nice thanks!

PR #5697, opened 2022-08-10
[lang] Add parallel scan prefix sum utility

qiao-bo:
This is an single-buffer implementation of parallel scan (prefix sum).

Reference: 
https://developer.download.nvidia.com/compute/cuda/1.1-Beta/x86_website/projects/scan/doc/sc
https://github.com/NVIDIA/cuda-samples/blob/master/Samples/2_Concepts_and_Techniques/shfl_sc

This will be useful for certain Taichi implementations such as PBD.


python/taichi/_kernels.py
```diff
@@ -308,3 +310,143 @@ def parallel_sort(keys, values=None):
...
+        ele_num = int((ele_num + BLOCK_SZ - 1) / BLOCK_SZ)
+        ele_nums.append(ele_num)
+        start_pos += BLOCK_SZ * ele_num
+        ele_nums_pos.append(start_pos)
+
+    if input_arr.dtype != i32:
+        raise RuntimeError("Only ti.i32 type is supported for prefix sum.")
+
+    large_arr = field(i32, shape=start_pos)
+    smem = field(i32, shape=(int(GRID_SZ), 64))
```
k-ye, line 426: OOC, does this only work on field?

qiao-bo: I think Ndarray should also work. Perhaps we can relax these limitations in a follow-up PR? Including support for float types etc.

k-ye: I see, we need to be careful about these implicit fields.

YuCrazing (approved):
LGTM!

PR #5710, opened 2022-08-10
[bug] Properly delete functions of a SNode Tree in CUDA backend

### LINKED ISSUE #5711: Functions of deleted SNodeTrees not deleted in CUDA backend
**Describe the bug**
In https://github.com/taichi-dev/taichi/blob/ca5bf7d138bfe615cff8018c00586e786f5fdc39/taichi/runtime/llvm/llvm_runtime_executor.cpp#L600 , the function deletes the functions of SNodeTrees in the host context instead of the device context (and the deletions are ignored because the functions are not in the host context), so the functions are not deleted at all.

lin-hitonami:
Related issue = fixes #5711


ailzhang (approved):
Thanks!

PR #5712, opened 2022-08-10
[bug] Support indexing via np.integer in field

ailzhang:
Related issue = #

This was used in difftaichi. `python examples/diffmpm.py`


strongoier (commented):
Should the modification apply to `MatrixField` and ndarrays as well?

PR #5717, opened 2022-08-10
[vulkan] Fixed query pool invalid usage

PENGUINLIONG:
This PR fixed a validation issue that query pools are not correctly reset before use (for on-device execution time measurement). Also fixed a bit of compilation issue.

### DIFF
--- a/taichi/rhi/vulkan/vulkan_device.cpp
+++ b/taichi/rhi/vulkan/vulkan_device.cpp
@@ -1722,6 +1722,7 @@ void VulkanStream::command_sync() {
     vkGetQueryPoolResults(device_.vk_device(), cmdbuf.query_pool->query_pool, 0,
                           2, sizeof(uint64_t) * 2, &t, sizeof(uint64_t),
                           VK_QUERY_RESULT_64_BIT | VK_QUERY_RESULT_WAIT_BIT);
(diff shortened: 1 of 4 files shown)


taichi/rhi/vulkan/vulkan_device.cpp
```diff
@@ -782,6 +782,7 @@ VulkanCommandList::VulkanCommandList(VulkanDevice *ti_device,
   info.flags = VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT;
 
   vkBeginCommandBuffer(buffer->buffer, &info);
+  vkCmdResetQueryPool(buffer->buffer, query_pool_->query_pool, 0, 2);
```
ailzhang, line 785: Nit: IIRC the length of query_pool is always 2 which stands for start time and end time, is this correct understanding? 
If so, would you mind putting this into a comment so that it's easier to understand ;) ? Thanks!

ailzhang (approved):
Thanks!

taichi/rhi/vulkan/vulkan_device.cpp
PENGUINLIONG: Thats correct, Imma work on the profiler system integration soon after the C-API docs so I'll solve this after that. The magic number should be completely removed actually.

Issue #5718, opened 2022-08-10
global data access rule checker doesn't allocate memory correctly


erizmr:
**Describe the bug**
global data access rule checker doesn't allocate memory correctly.

**Log/Screenshots**

```
RuntimeError: [snode_tree_buffer_manager.cpp:allocate@42] LLVM backend supports up to 512 snode trees
```

erizmr:
Fixed by #5719 and #5801.

PR #5719, opened 2022-08-10
[autodiff] Move the global data access rule checker to experimental

erizmr:
Related issue = #5718 
Move the global data access rule checker to experimental.

### DIFF
--- a/taichi/program/compile_config.h
+++ b/taichi/program/compile_config.h
@@ -11,6 +11,7 @@ struct CompileConfig {
   bool debug;
   bool cfg_optimization;
   bool check_out_of_bound;
+  bool check_autodiff_valid;
   int simd_width;
   bool lazy_compilation;
   int opt_level;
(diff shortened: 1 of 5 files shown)


taichi/program/compile_config.h
```diff
@@ -11,6 +11,7 @@ struct CompileConfig {
   bool debug;
   bool cfg_optimization;
   bool check_out_of_bound;
+  bool check_autodiff_valid;
```
ailzhang, line 14:
```suggestion
  bool validate_autodiff;
```

ailzhang (approved):
Hmmm please correct me if I'm wrong, this compile config is a temp workaround and  our final solution is still controlled in forward AD context manager right?

taichi/program/compile_config.h
erizmr: Yes. The final solution is controlled in the `ti.ad.Tape`.

PR #5720, opened 2022-08-10
[ci] Fix TI_SKIP_CPP_TESTS

feisuzhu:
Related issue = #


qiao-bo (commented):
Thx for fixing the cpp part. I will leave the sph type part to @jim19930609 ;/

jim19930609 (approved):

PR #5734, opened 2022-08-11
[ci] Prevent using the offline cache of previous jobs

lin-hitonami:
Related issue = #4401 
The release test seems to be using the offline cache of previous jobs on Windows and M1.
Moving the offline cache directory into the repository directory solves this because the repository is deleted after the job finishes.


feisuzhu (approved):
LGTM!

PR #5749, opened 2022-08-12
[gui] Less sync overhead for GGUI

bobcao3:
When we have the source buffer coming from the same device as the GGUI device (i.e. Vulkan compute + GGUI), use "compute op lambda" to directly put the transfer commands into the program's queue, avoiding a bunch of sync copy operations.


PENGUINLIONG (approved):
looks a bit of hacky but it feels like there is no better solution atm. some day `Device` creation should be apart from `Program` so we don't have to wire through the program instance. lgtm

PR #5754, opened 2022-08-12
[Doc] Update field.md

neozhaoliang:


ailzhang (commented):
I wish there was an example of image and how they looks like in each framework :P Maybe put this in FAQ instead?

neozhaoliang:
Added to FAQ list.

PR #5755, opened 2022-08-12
[Error] [lang] Add error when breaking/continuing a static for inside non-static if

### LINKED ISSUE #5729: condition statements in static loop fails silently
**Version**
`taichi-nightly 1.0.5.post20220810`

**Describe the bug**
A clear and concise description of what the bug is, ideally within 20 words.
```py
import taichi as ti

@ti.func
def test_static_loop():
    for i in ti.static(range(5)):
        x = 0.1
        if x == 0.0:
            print("consition statement")
            break
        print("after condition")

@ti.kernel
def test_kernel():
    test_static_loop()

ti.init(arch=ti.cuda)

test_kernel()

```

Expected behavior:
Prints 5 `after condition` or raises an error.

Actual behavior:
Print nothing and fails silently.

lin-hitonami:
Related issue = fixes #5729 
Added `nearest_non_static_if` to the loop attribute. The `NonStaticIfGuard` maintains the `nearest_non_static_if` of the current loop.

For this kernel:
```
import taichi as ti

@ti.kernel
def test_static_loop():
    for i in ti.static(range(5)):
        x = 0.1
        if x == 0.0:
            print("consition statement")
            break
        print("after condition")

ti.init(arch=ti.cuda)

test_static_loop()
```

The error is like:
```
[Taichi] version 1.1.1, llvm 10.0.0, commit 6f9f5549, linux, python 3.10.4
  File "/home/hitonami/test/test.py", line 14, in <module>
    test_static_loop()
  File "/home/hitonami/taichi/python/taichi/lang/kernel_impl.py", line 924, in wrapped
taichi.lang.exception.TaichiSyntaxError: 
On line 9 of file "/home/hitonami/test/test.py", in test_static_loop:
            break
            ^^^^^
On line 7 of file "/home/hitonami/test/test.py", in test_static_loop:
        if x == 0.0:
           ^^^^^^^^
You are trying to `break` a static `for` loop, but the `break` statement is inside a non-static `if`. 
```


lin-hitonami:
I don't think there are any cases anyone should break/continue a static for loop inside non-static, so I made it an error.
Is this error message OK?

PR #5766, opened 2022-08-15
[doc] Fix docs and docs sidebar conflict issue

rexwangcc:
Related issue = #

- [x] Fix the issue that docs sidebar cannot expand when folder and doc have the exact same name, by explicitly giving the folder a id.


rexwangcc:
Close since we have switched to use `slug`.

Issue #5728, opened 2022-08-11
Support shift ptr in dynamic index for autodiff


erizmr:
The IR `shift ptr` related to dynamic index should be supported in current autodiff.

erizmr:
Fixed by #5770

PR #5770, opened 2022-08-15
[autodiff] Support shift ptr in dynamic index

erizmr:
Related issue = #5728

### DIFF
--- a/tests/python/test_ad_dynamic_index.py
+++ b/tests/python/test_ad_dynamic_index.py
@@ -1,0 +1,28 @@
+import taichi as ti
+from tests import test_utils
+
+
+@test_utils.test(require=ti.extension.dynamic_index,
+                 dynamic_index=True,
+                 debug=True)
+def test_matrix_non_constant_index():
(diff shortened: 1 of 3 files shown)


tests/python/test_ad_dynamic_index.py
```diff
@@ -0,0 +1,28 @@
+import taichi as ti
+from tests import test_utils
+
+
+@test_utils.test(require=ti.extension.dynamic_index,
```
ailzhang, line 5: nit: since you also handled ptroffset in MakeDual, maybe it's good to add another fwd mode test with dynamic index=true?

ailzhang (approved):
Awesome!!

tests/python/test_ad_dynamic_index.py
erizmr: Sure. The current API of fwd mode doesn't support vector or matrix field. I can submit another PR to enable vector or matrix field in fwd mode and attach a fwd mode test.

PR #5773, opened 2022-08-15
[opengl] Fix target device for opengl aot

ailzhang:
* #5775
* #5774
* __->__ #5773

Note we used to generate invalid spirv shaders during opengl aot and
failed to run spirv-opt passes. But it only generated a warning instead
of an error so we didn't catch this in tests.

This was the second bug triggered by invalid spirv shaders. We
probably should just throw when spirv-opt fails. Alternatively we can
run spirv-val separately if we really want to ignore spirv-opt failures.

Note we don't properly handle gles target device yet, will do that in a
followup PR.

### DIFF
--- a/taichi/aot/module_loader.h
+++ b/taichi/aot/module_loader.h
@@ -117,7 +117,7 @@ class TargetDevice : public Device {
   }
 
   void set_default_caps(Arch arch) {
-    if (arch == Arch::vulkan) {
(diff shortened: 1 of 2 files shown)


PENGUINLIONG (approved):
lgtm. i'm kinda surprised it doesn't cost you any hack to use spirv 1.3 shaders to xcompile to opengl

PR #5778, opened 2022-08-15
[ci] Skip C++ tests on macOS + CPU

feisuzhu:
Related issue = #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -250,6 +250,7 @@ jobs:
         run: .github/workflows/scripts/unix_test.sh
         env:
           TI_WANTED_ARCHS: 'cpu'
+          TI_SKIP_CPP_TESTS: Disabled because Vulkan is supported but not working on buildbot4
 
   build_and_test_m1:
     name: Build and Upload (Apple M1)


.github/workflows/release.yml
```diff
@@ -250,6 +250,7 @@ jobs:
         run: .github/workflows/scripts/unix_test.sh
         env:
           TI_WANTED_ARCHS: 'cpu'
+          TI_SKIP_CPP_TESTS: Disabled because Vulkan is supported but not working on buildbot4
```
ailzhang, line 253: hmmm I thought this was guarded by https://github.com/taichi-dev/taichi/blob/master/tests/cpp/aot/vulkan/aot_save_load_test.cpp#L220? Is there any particular reason that we want to disable it explicitly?

feisuzhu: 1. On buildbot4 vulkan IS available but if you use it wherever it will crash.
2. Nightly builds are currently broken because of this.

feisuzhu: This is a temporary workaround and I'll try fix it by https://github.com/taichi-dev/taichi/issues/5544

ailzhang (approved):
Thanks！

PR #5551, opened 2022-07-28
[Lang] Enable matrix representation in Frontend/CHI IR

AD1024:
Related issue = #5478

Issue #5478, opened 2022-07-20
[RFC] [Lang] Representing matrices/vectors in Frontend and CHI IR to enable SIMD


AD1024:
# Background
Many matrix-matrix / matrix-vector operations such as load/store, elementwise-add, dot product, etc. could gain performance benefits from SIMD instructions. The current implementation of Taichi matrices and vectors is only visible at Python frontend - all the operations over these data structures are being unrolled prior to further transformations (e.g. optimizations on CHI IR), blocking us from applying any further optimizations. This issue proposes a plan to implement matrices and vectors in CHI IR to expand the space for more optimizations.

# Use cases
Operations over matrices and vectors may benefit from the new implementation. For example,
```python
ti.init(real_matrix=True)
@ti.kernel
def test_kernel(i : ti.i32):
    x = ti.Matrix(...)      # same interface as before
    weights = ...
    result = x <op> weights # same operator as before
```
Our design aims to be compatible with current user-level API while enabling SIMD optimizations depending on hardware capability. We will ensure that there is no performance regression before and after this optimization.
We will conduct a more detailed study on performance benefits of this optimization later.

# Representation local matrices at Python, Frontend IR, and CHI IR level
Here, we mainly consider two major objects: Matrix and indexed Matrix, both should stay as a whole till code generation:
1. Python level:
Add a flag `real_matrix` to `ti.init()`, which controls the transformation for local matrices during AST Transformation.
- Local Matrix: Transformed into FrontendAllocaStmt(DataType::Tensor) and FrontendAssignStmt during AST Transformation.
- Indexed Matrix: Transform into IndexedExpression
2. Frontend IR level:
- Local Matrix: Verify FrontendAllocaStmt(DataType::Tensor) is supported.
- Indexed Matrix: Verify IndexExpression(IdExpression, indices={}) is supported.
3. CHI IR level:
- Local Matrix: Adjust OPs (Load/Store) to support operands of AllocaStmt(DataType::Tensor).
- Indexed Matrix: Add new statement "IndexStmt". Adjust existing Ops accordingly.

# Lowering and Codegen
For prototyping purpose, LLVM backend will be the major code generation target. SIMD data types and instructions in LLVM are described in [the LLVM documentation](https://llvm.org/docs/LangRef.html#vector-type).

## Fallback
[...]

PR #5782, opened 2022-08-15
[Lang] Enable definition of local matrices/vectors

AD1024:
Related issue = #5478 
A part of PR #5551


taichi/transforms/type_check.cpp
```diff
@@ -553,6 +553,26 @@ class TypeCheck : public IRVisitor {
...
+                   "Matrix should have tensor type, got {}",
+                   stmt->ret_type->to_string());
+    auto tensor_type = stmt->ret_type->as<TensorType>();
+    auto element_dtype = tensor_type->get_element_type();
+    for (auto elt : stmt->values) {
+      element_dtype = promoted_type(element_dtype, elt->ret_type);
+    }
+    for (int i = 0; i < stmt->values.size(); ++i) {
+      if (element_dtype != stmt->values[i]->ret_type) {
+        cast(stmt->values[i], element_dtype);
```
jim19930609, line 568: Just complaining: 

I know this get fused with CHI IR Passes thus cause no side-effects. However, it is a bit stupid to initialize the constant values with a "default type" then cast it to the dtype specified by the Matrix (this also happens to some other data structures). 

In the future we might want to directly initialize the constants to it's specified dtype.

taichi/transforms/alg_simp.cpp
```diff
@@ -245,7 +251,13 @@ class AlgSimp : public BasicStmtVisitor {
                  (fast_math || is_integral(stmt->ret_type)) &&
                  irpass::analysis::same_value(stmt->lhs, stmt->rhs)) {
         // fast_math or integral operands: a -^ a -> 0
-        replace_with_zero(stmt);
+        if (stmt->lhs->ret_type->is<PrimitiveType>() &&
```
jim19930609, line 254: ditto

taichi/transforms/alg_simp.cpp
```diff
@@ -164,8 +164,14 @@ class AlgSimp : public BasicStmtVisitor {
     if ((fast_math || is_integral(stmt->ret_type)) &&
         irpass::analysis::same_value(stmt->lhs, stmt->rhs)) {
       // fast_math or integral operands: a / a -> 1
-      replace_with_one(stmt);
-      return true;
+      if (stmt->lhs->ret_type->is<PrimitiveType>() &&
```
jim19930609, line 167: If we just wanna skip TensorType, then better do it this way:
```
if (stmt->ret_type->is<TensorType>() ||
            stmt->rhs->ret_type->is<TensorType>()) {
  return false;
}
```
So we don't overkill. Just in case someone adds another DataType aside from TensorType and PrimitiveType in the future.

taichi/runtime/llvm/llvm_context.cpp
```diff
@@ -135,6 +135,10 @@ llvm::Type *TaichiLLVMContext::get_data_type(DataType dt) {
     return llvm::Type::getInt64Ty(*ctx);
   } else if (dt->is_primitive(PrimitiveTypeID::f16)) {
     return llvm::Type::getHalfTy(*ctx);
+  } else if (dt->is<TensorType>()) {
+    auto vectorty = dt->as<TensorType>();
+    auto dtype = this->get_data_type(vectorty->get_element_type());
+    return llvm::VectorType::get(dtype, vectorty->get_num_elements(), false);
```
jim19930609, line 141: better add a comment, so we know it's codegen to fixed-sized VectorType 
`llvm::VectorType::get(dtype, vectorty->get_num_elements(), false/*scalable*/);`

taichi/python/export_lang.cpp
```diff
@@ -285,6 +286,7 @@ void export_lang(py::module &m) {
       .def("insert_external_func_call", &ASTBuilder::insert_external_func_call)
       .def("expr_alloca", &ASTBuilder::expr_alloca)
       .def("expr_alloca_local_tensor", &ASTBuilder::expr_alloca_local_tensor)
+      .def("expr_alloca_matrix", &ASTBuilder::expr_alloca_local_matrix)
```
jim19930609, line 289: Is `expr_alloca_matrix` doing the same thing as `expr_alloca_local_tensor`? Then can we make `expr_alloca_matrix` an alternative code branch controlled by "real_matrix" inside `expr_alloca_local_tensor`?

This way, we hide implementation details to Python

taichi/ir/type.h
```diff
@@ -175,6 +175,8 @@ class TensorType : public Type {
     return shape_;
   }
 
+  int vector_width() const;
```
jim19930609, line 178: mark `override`? (overrides Type::vector_width())

taichi/ir/type.cpp
```diff
@@ -87,6 +87,14 @@ std::string TensorType::to_string() const {
...
+int TensorType::vector_width() const {
+  int vw = 1;
+  for (auto dim : shape_) {
+    vw *= dim;
+  }
+  return vw;
+}
+
 int Type::vector_width() const {
   return 1;  // TODO: CPU vectorization
```
jim19930609, line 99: @strongoier correct me if I were wrong, but I think we can probably remove this "TODO" since we have a dedicated TensorType now?

taichi/ir/statements.h
```diff
@@ -1807,5 +1807,17 @@ class MeshPatchIndexStmt : public Stmt {
   TI_DEFINE_ACCEPT_AND_CLONE
 };
 
+class MatrixInitStmt : public Stmt {
```
jim19930609, line 1810: Let's add some comments to explain what this `Stmt` represents

taichi/ir/frontend_ir.h
```diff
@@ -504,6 +504,26 @@ class GlobalVariableExpression : public Expression {
   TI_DEFINE_ACCEPT_FOR_EXPRESSION
 };
 
+class MatrixExpression : public Expression {
```
jim19930609, line 507: Some comments to plain what this `Expression` represents. 

Since Frontend IR is fairly close to Python codes, it would be nice if we can further introduce what python code gets lowered to this Expression.

taichi/ir/control_flow_graph.cpp
```diff
@@ -501,7 +501,8 @@ bool CFGNode::dead_store_elimination(bool after_lower_access) {
       }
     }
     auto load_ptrs = irpass::analysis::get_load_pointers(stmt);
-    if (load_ptrs.size() == 1 && store_ptrs.empty() && stmt->width() == 1) {
+    if (load_ptrs.size() == 1 && store_ptrs.empty() && stmt->width() == 1 &&
+        !stmt->is<PtrOffsetStmt>()) {
```
jim19930609, line 505: Was wondering how "PtrOffsetStmt" get related to MatrixType change?

taichi/analysis/data_source_analysis.cpp
```diff
@@ -37,6 +37,10 @@ std::vector<Stmt *> get_load_pointers(Stmt *load_stmt) {
     return external_func->arg_stmts;
   } else if (auto ref = load_stmt->cast<ReferenceStmt>()) {
     return {ref->var};
+  } else if (auto matrix_init = load_stmt->cast<MatrixInitStmt>()) {
+    return matrix_init->values;
+  } else if (auto ptr_offset = load_stmt->cast<PtrOffsetStmt>()) {
```
jim19930609, line 42: Is this change to PtrOffsetStmt relevant to MatrixType?

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -688,6 +688,11 @@ llvm::Type *TaskCodeGenLLVM::llvm_type(DataType dt) {
     return llvm::Type::getDoubleTy(*llvm_context);
   } else if (dt->is_primitive(PrimitiveTypeID::f16)) {
     return llvm::Type::getHalfTy(*llvm_context);
+  } else if (dt->is<TensorType>()) {
+    auto tensor_type = dt->cast<TensorType>();
+    auto element_type = llvm_type(tensor_type->get_element_type());
+    return llvm::VectorType::get(element_type, tensor_type->get_num_elements(),
+                                 false);
```
jim19930609, line 695: Better add a comment to the argument, so it doesn't seem like a magic number:

llvm::VectorType::get(element_type, tensor_type->get_num_elements(), false /\*scalable\*/);

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2515,6 +2528,16 @@ void TaskCodeGenLLVM::visit(MeshPatchIndexStmt *stmt) {
   llvm_val[stmt] = get_arg(2);
 }
 
+void TaskCodeGenLLVM::visit(MatrixInitStmt *stmt) {
+  auto type = tlctx->get_data_type(stmt->ret_type->as<TensorType>());
+  llvm::Value *vec = llvm::UndefValue::get(type);
```
jim19930609, line 2533: Looks like the only reason you need this UndefValue is to pass the type info to `builder->CreateInsertElement()`

To make the code more explicit, can we try switching to this overloaded `builder->CreateInsertElement()` interface, where you directly pass-in the `Type` instead of having to wrap it with some UndefValue?

2022-08-17 11-52-13 的屏幕截图

jim19930609:
Nice work! Just some comments wrt implementation details~

taichi/transforms/type_check.cpp
AD1024: Agree. One possible solution is to enable parent(s) AST nodes to inform their children about how to initialize the most general parts (i.e. fields defined in `Stmt`/`Expression`) when constructing the AST.

taichi/ir/type.h
AD1024: Nice catch. Just realized that this is no longer used. (this was used by LocalLoadStmt for loading multiple values, but now we are wrapping into a `MatrixInitStmt`)

taichi/analysis/data_source_analysis.cpp
AD1024: Yes, PtrOffsetStmt represents matrix indexing.

jim19930609: Ah I see, we might migrate these PtrOffsetStmt to something like a MatrixPtrStmt, so it has more explicit semantic. But that's not urgent.

jim19930609 (approved):
Thanks!

taichi/ir/type.cpp
strongoier: We should actually remove all legacy code related to CPU vectorization.

strongoier (commented):
BTW have you tried to run tests locally to see if things work well with `dynamic_index=True` (the tests are not enabled in CI)?

python/taichi/lang/matrix.py
```diff
@@ -97,6 +97,27 @@ def prop_setter(instance, value):
     return cls
 
 
+def make_matrix(arr, dt=None):
+    if len(arr) == 0:
```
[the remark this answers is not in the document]
AD1024: In the original implementation, there is a case where the matrix is empty. Though we can omit that part here.

python/taichi/lang/impl.py
```diff
@@ -48,6 +54,8 @@ def expr_init(rhs):
     if isinstance(rhs, Matrix) and (hasattr(rhs, "_DIM")):
         return Matrix(*rhs.to_list(), ndim=rhs.ndim)
     if isinstance(rhs, Matrix):
+        if current_cfg().real_matrix:
+            return rhs
```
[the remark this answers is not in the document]
AD1024: I think in this case we extract out the elements and call `make_matrix` again

python/taichi/lang/impl.py
```diff
@@ -35,6 +35,12 @@ def expr_init_local_tensor(shape, element_type, elements):
         shape, element_type, elements)
 
 
+@taichi_scope
+def expr_init_matrix(shape, element_type, elements):
```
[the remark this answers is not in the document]
AD1024: Good catch. We merged this into `expr_init_local_tensor`

taichi/ir/frontend_ir.h
```diff
@@ -876,6 +902,9 @@ class ASTBuilder {
                                  const ExprGroup &args,
                                  const ExprGroup &outputs);
   Expr expr_alloca();
+  Expr expr_alloca_local_matrix(const std::vector<int> &shape,
```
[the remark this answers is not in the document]
AD1024: Same above.

AD1024:
> BTW have you tried to run tests locally to see if things work well with `dynamic_index=True` (the tests are not enabled in CI)?

Will try. I don't think this would be an issue.

taichi/transforms/type_check.cpp
AD1024: In the previous implementation, `dt` is an optional argument to `make_matrix` and will keep as is when passed to FrontendIR. I remember I changed to automatically infer it using some utilities already implemented.

python/taichi/lang/matrix.py
```diff
@@ -97,6 +97,27 @@ def prop_setter(instance, value):
...
+        return impl.expr_init(
+            impl.expr_init_local_tensor([0], ti_python_core.DataType_unknown,
+                                        impl.make_expr_group([])))
+    is_matrix = isinstance(arr[0], Iterable)
+    if dt is None:
+        dt = _make_entries_initializer(is_matrix).infer_dt(arr)
+    if not is_matrix:
+        return impl.expr_init(
+            impl.expr_init_local_tensor([len(arr)], dt,
```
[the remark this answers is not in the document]
AD1024: Sure. I will merge the `expr_init` part into the implementation.

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -124,7 +124,7 @@ void TaskCodeGenLLVM::visit(Block *stmt_list) {
 void TaskCodeGenLLVM::visit(AllocaStmt *stmt) {
   if (stmt->ret_type->is<TensorType>()) {
     auto tensor_type = stmt->ret_type->cast<TensorType>();
-    auto type = tlctx->get_data_type(tensor_type->get_element_type());
+    auto type = tlctx->get_data_type(tensor_type);
```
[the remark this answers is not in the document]
AD1024: Probably missed some changes from the previous implementation. Fixed. Thanks!

python/taichi/lang/matrix.py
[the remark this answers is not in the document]
strongoier: Let's remove the case in both new and old impl.

taichi/transforms/type_check.cpp
strongoier: Then shall we remove the part handling `unknown` in `type_check.cpp`?

taichi/transforms/alg_simp.cpp
```diff
@@ -164,6 +164,11 @@ class AlgSimp : public BasicStmtVisitor {
     if ((fast_math || is_integral(stmt->ret_type)) &&
         irpass::analysis::same_value(stmt->lhs, stmt->rhs)) {
       // fast_math or integral operands: a / a -> 1
+      if (stmt->lhs->ret_type->is<TensorType>() ||
```
strongoier, line 167: Should we simply place a check of `TensorType` at the beginning of this pass?

taichi/ir/type_utils.cpp
```diff
@@ -53,6 +53,36 @@ int data_type_size(DataType t) {
...
+  for (int i = 0; i < shape[dim]; ++i) {
+    if (dim != shape.size() - 1) {
+      fmt += tensor_type_format_helper(shape, format_str, dim + 1);
+    } else {
+      fmt += format_str;
+    }
+    if (i != shape[dim] - 1) {
+      fmt += ", ";
+      if (dim == 0) {
+        fmt += "\n";
```
strongoier, line 69: Do we need this when the tensor is a vector instead of a matrix?

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -688,6 +688,11 @@ llvm::Type *TaskCodeGenLLVM::llvm_type(DataType dt) {
     return llvm::Type::getDoubleTy(*llvm_context);
   } else if (dt->is_primitive(PrimitiveTypeID::f16)) {
     return llvm::Type::getHalfTy(*llvm_context);
+  } else if (dt->is<TensorType>()) {
```
strongoier, line 691: 1. Shall we assert `real_matrix=True` here?
2. It seems that this function is exactly the same as `llvm::Type *TaichiLLVMContext::get_data_type(DataType dt)`. Can we merge them as well (in another PR maybe).

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -800,12 +805,26 @@ void TaskCodeGenLLVM::visit(PrintStmt *stmt) {
...
-      formats += data_type_format(arg_stmt->ret_type);
+      if (arg_stmt->ret_type->is<TensorType>()) {
+        auto dtype = arg_stmt->ret_type->cast<TensorType>();
+        auto elem_type = dtype->get_element_type();
+        for (int i = 0; i < dtype->get_num_elements(); ++i) {
+          auto elem_value = builder->CreateExtractElement(value, i);
+          if (elem_type->is_primitive(PrimitiveTypeID::f32) ||
+              elem_type->is_primitive(PrimitiveTypeID::f16))
+            elem_value = builder->CreateFPExt(
+                elem_value, tlctx->get_data_type(PrimitiveType::f64));
```
strongoier, line 816: We should avoid writing this part twice.

taichi/ir/type_utils.cpp
AD1024: Good point. Let's keep 1-line for Vectors.

taichi/analysis/same_statements.cpp
```diff
@@ -196,6 +196,24 @@ class IRNodeComparator : public IRVisitor {
     basic_check(stmt);
   }
 
+  void visit(MatrixInitStmt *stmt) override {
```
strongoier, line 199: Where is this used?

taichi/transforms/alg_simp.cpp
AD1024: Sure we could do that instead. Just want to place here as a reminder (implement these for TensorType later)

taichi/ir/control_flow_graph.cpp
strongoier: I would also like to understand this change. Why do we exclude `PtrOffsetStmt`? If this only cannot work under `real_matrix=True`, we should check `real_matrix` as well.

taichi/analysis/data_source_analysis.cpp
```diff
@@ -37,6 +37,10 @@ std::vector<Stmt *> get_load_pointers(Stmt *load_stmt) {
     return external_func->arg_stmts;
   } else if (auto ref = load_stmt->cast<ReferenceStmt>()) {
     return {ref->var};
+  } else if (auto matrix_init = load_stmt->cast<MatrixInitStmt>()) {
+    return matrix_init->values;
```
strongoier, line 41: Why is `matrix_init->values` a pointer?

taichi/analysis/same_statements.cpp
AD1024: iirc I was trying to add CSE support for matrix

taichi/ir/control_flow_graph.cpp
AD1024: I was trying to adapt CFG pass with matrix indexing. With the current implementation, this could be removed. (will be in a separate PR)

taichi/analysis/data_source_analysis.cpp
AD1024: Same as below (can be removed according to current cfg pass support for new matrix)

taichi/analysis/same_statements.cpp
strongoier: Got it. Let's make each PR clean :-)

python/taichi/lang/matrix.py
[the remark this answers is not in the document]
strongoier: Yep your understanding is correct. `expr_init_xxx` is meant to be a specialized version of `expr_init`, so variable creation and binding needs to be handled as well.

taichi/codegen/llvm/codegen_llvm.cpp
AD1024: Sounds good to me

taichi/codegen/llvm/codegen_llvm.cpp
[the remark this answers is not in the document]
AD1024: Update: when fixing matrix indexing, this should be changed to allocating a pointer of vector type in order to make StoreInsn work well.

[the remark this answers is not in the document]
AD1024: Fixed: now checking whether real matrix is enabled. If so, allocate pointer to vector type otherwise allocate `[type * array_size]*`

taichi/transforms/alg_simp.cpp
```diff
@@ -100,6 +100,11 @@ class AlgSimp : public BasicStmtVisitor {
 
   bool optimize_multiplication(BinaryOpStmt *stmt) {
     // return true iff the IR is modified
+    if (stmt->lhs->ret_type->is<TensorType>() ||
```
strongoier, line 103: I guess this check is already covered by the one in `void visit(BinaryOpStmt *stmt)`.

taichi/ir/frontend_ir.cpp
```diff
@@ -970,10 +991,25 @@ Expr ASTBuilder::expr_alloca() {
...
                                           const DataType &element_type,
                                           const ExprGroup &elements,
                                           std::string tb) {
+  if (this->use_real_matrix_) {
+    auto matrix_expr = make_local_matrix(shape, element_type, elements.exprs);
+    auto v = this->expr_alloca();
+    this->expr_assign(v, matrix_expr, tb);
+    // type check for variable `v` since
+    // expr_assign couldn't propagate the info
+    v->ret_type = matrix_expr.cast<MatrixExpression>()->dt;
```
strongoier, line 1010: I guess the constructor of `FrontendAssignStmt` called in `expr_assign` propagates the info?

taichi/analysis/data_source_analysis.cpp
```diff
@@ -59,7 +59,7 @@ Stmt *get_store_data(Stmt *store_stmt) {
 
 std::vector<Stmt *> get_store_destination(Stmt *store_stmt) {
   // If store_stmt provides some data sources, return the pointers of the data.
-  if (store_stmt->is<AllocaStmt>() && !store_stmt->ret_type->is<TensorType>()) {
+  if (store_stmt->is<AllocaStmt>()) {
```
strongoier, line 62: This change seems unrelated to this PR. Let's cancel this first.

taichi/transforms/alg_simp.cpp
```diff
@@ -151,6 +151,11 @@ class AlgSimp : public BasicStmtVisitor {
 
   bool optimize_division(BinaryOpStmt *stmt) {
     // return true iff the IR is modified
+    if (stmt->lhs->ret_type->is<TensorType>() ||
```
strongoier, line 154: same redundant check here :-)

python/taichi/lang/impl.py
```diff
@@ -36,6 +37,12 @@ def expr_init_local_tensor(shape, element_type, elements):
         get_runtime().get_current_src_info())
 
 
+@taichi_scope
+def make_local_matrix(shape, element_type, elements):
```
strongoier, line 41: I suggest renaming it to `make_matrix_expr`. The result should be wrapped in `expr.Expr`.
References: https://github.com/taichi-dev/taichi/blob/2ba0343c904fb405086485cd5b853ff606e7b949/python/taichi/lang/impl.py#L224-L228
https://github.com/taichi-dev/taichi/blob/2ba0343c904fb405086485cd5b853ff606e7b949/taichi/python/export_lang.cpp#L974-L975

taichi/ir/frontend_ir.cpp
```diff
@@ -970,6 +990,12 @@ Expr ASTBuilder::expr_alloca() {
   return var;
 }
 
+Expr ASTBuilder::make_local_matrix(const std::vector<int> &shape,
```
strongoier, line 993: Change the name here as well?

taichi/python/export_lang.cpp
```diff
@@ -290,6 +291,7 @@ void export_lang(py::module &m) {
       .def("insert_deactivate", &ASTBuilder::insert_snode_deactivate)
       .def("insert_activate", &ASTBuilder::insert_snode_activate)
       .def("insert_external_func_call", &ASTBuilder::insert_external_func_call)
+      .def("make_local_matrix", &ASTBuilder::make_matrix_expr)
```
strongoier, line 294:
```suggestion
      .def("make_matrix_expr", &ASTBuilder::make_matrix_expr)
```

python/taichi/lang/impl.py
```diff
@@ -36,6 +37,12 @@ def expr_init_local_tensor(shape, element_type, elements):
         get_runtime().get_current_src_info())
 
 
+@taichi_scope
+def make_matrix_expr(shape, element_type, elements):
+    return get_runtime().prog.current_ast_builder().make_local_matrix(
```
strongoier, line 42:
```suggestion
    return get_runtime().prog.current_ast_builder().make_matrix_expr(
```

strongoier (approved):
LGTM!

PR #5783, opened 2022-08-15
[Lang] Indexing for new local matrix implementation

AD1024:
Related issue = #5478 
A part of PR #5551


python/taichi/lang/expr.py
```diff
@@ -39,6 +39,12 @@ def __init__(self, *args, tb=None, dtype=None):
             self.ptr.set_tb(self.tb)
         self.ptr.type_check(impl.get_runtime().prog.config)
 
+    def __getitem__(self, *indices):
```
strongoier, line 42: Shall we add this to `impl.subscript`? Calling `__getitem__` is for non-Taichi types: https://github.com/taichi-dev/taichi/blob/2bf8cef297cd0b617eef6addcd4cb2cd1708a4dc/python/taichi/lang/impl.py#L227-L228

strongoier (commented):
Let's start to add tests. I'll also mark this PR with 'full-ci` to make sure dynamic index is not broken.

taichi/ir/control_flow_graph.cpp
```diff
@@ -240,6 +240,15 @@ void CFGNode::reaching_definition_analysis(bool after_lower_access) {
     // loop in reversed order
     auto stmt = block->statements[i].get();
     auto data_source_ptrs = irpass::analysis::get_store_destination(stmt);
+    if (auto local_store = stmt->cast<LocalStoreStmt>()) {
```
AD1024, line 243: Note the special case here: this is because the granularity of reaching definition and liveness analysis is variable. This means the current implementation only tells us whether the entire matrix is dead or alive at a certain location. Therefore, if we make `get_store_destination` always return `ptr->origin`, during load-store forwarding optimization, a matrix could be replaced by the data loaded into it. For example, if we have
```
$1 = [a, b, c, d]
$2 = alloca
$3 = load [$2 <- $1]
$4 = Ptr shift [$3 + 0]
$5 = load [$4 <- x]
print $2
```
the last `$2` will be replaced by `x` if we directly change `get_store_destination`.

taichi/ir/frontend_ir.cpp
```diff
@@ -424,6 +424,9 @@ Stmt *make_tensor_access(Expression::FlattenContext *ctx,
                          std::vector<int> shape,
                          int stride) {
   flatten_lvalue(var, ctx);
+  if (var->stmt->ret_type->is_primitive(PrimitiveTypeID::unknown)) {
+    var->stmt->ret_type = var->ret_type;
```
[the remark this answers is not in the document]
AD1024: Good point. When lowering to PtrOffsetStmt, the ret_type is expected to be filled for `var->stmt`; however, there could be cases where `var->stmt` is lowered from `IdExpression`, which can result in unknown type.
I will move this assignment to `IdExpression::flatten`.

python/taichi/_kernels.py
```diff
@@ -241,9 +243,15 @@ def fill_matrix(mat: template(), vals: template()):
         for p in static(range(mat.n)):
             for q in static(range(mat.m)):
                 if static(mat[I].ndim == 2):
-                    mat[I][p, q] = vals[p][q]
+                    if static(isinstance(vals[p], Iterable)):
```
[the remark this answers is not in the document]
AD1024: Currently, there are `Matrix` and `_IntermediateMatrix`. The former one can be clearly told apart from `Vector`s; however, unfortunately, the latter one cannot be because there are code paths that constructs `_IntermediateMatrix` from `Vector`s. But after the refactor this could be purely distinguished by # of dimensions. Good point, thanks!

python/taichi/lang/impl.py
```diff
@@ -136,6 +139,9 @@ def subscript(value, *_indices, skip_reordered=False, get_ref=False):
     if isinstance(value, np.ndarray):
         return value.__getitem__(_indices)
 
+    if isinstance(value, Expr):
+        return make_index_expr(value.ptr, _indices)
```
strongoier, line 143: This part seems already covered by https://github.com/taichi-dev/taichi/blob/8c8b74db776ab89deea0aa0dc7bc2c3c27adea10/python/taichi/lang/impl.py#L233-L242

taichi/ir/frontend_ir.cpp
```diff
@@ -1018,7 +1025,9 @@ Expr ASTBuilder::expr_alloca() {
 Expr ASTBuilder::make_matrix_expr(const std::vector<int> &shape,
                                   const DataType &dt,
                                   const std::vector<Expr> &elements) {
-  return Expr(std::make_shared<MatrixExpression>(elements, shape, dt));
+  auto mat = Expr(std::make_shared<MatrixExpression>(elements, shape, dt));
+  mat->ret_type = dt;
```
strongoier, line 1029: Is this necessary? It seems that it has already been covered by https://github.com/taichi-dev/taichi/blob/5774ea62166456a9ec6ce4f69534527ba437bfe8/taichi/ir/frontend_ir.cpp#L434-L440

taichi/ir/frontend_ir.cpp
```diff
@@ -586,6 +590,9 @@ void LoopUniqueExpression::flatten(FlattenContext *ctx) {
 
 void IdExpression::flatten(FlattenContext *ctx) {
   stmt = ctx->current_block->lookup_var(id);
+  if (!ret_type->is_primitive(PrimitiveTypeID::unknown)) {
+    stmt->ret_type = ret_type;
+  }
```
strongoier, line 595: Is this necessary? It seems that it has already been covered by 
https://github.com/taichi-dev/taichi/blob/5774ea62166456a9ec6ce4f69534527ba437bfe8/taichi/ir/frontend_ir.cpp#L27-L33

strongoier (approved):
LGTM!

taichi/ir/frontend_ir.cpp
AD1024: This is a workaround for type checking a FrontendAssignStmt. Since the type info only propagates in constructor, we need its type to propagates the type to the variable being assigned.
https://github.com/taichi-dev/taichi/blob/4b1a619e53ba03cec7b1d2ca4a67139d524cdf30/taichi/ir/frontend_ir.cpp#L27-L33

taichi/ir/frontend_ir.cpp
AD1024: This is a workaround for type checking a FrontendAssignStmt. Since the type info only propagates in the constructor, we need its type to propagate the type to the variable being assigned.

Issue #5767, opened 2022-08-15
id(dtype) of builtin vectors not in real_type_ids


neozhaoliang:
To reproduce:

```python
import taichi as ti

vec3 = ti.math.vec3
print(id(vec3.dtype))
print(ti.types.primitive_types.real_type_ids)
```

The result gives

```
140376791869680
[140376797782384, 140376797803760, 140376797803888, 9489216]
```
The id of the `f32` type of vec3 is not in `ti.types.real_type_ids`.

This mismatch has the effect that the built-in`vec2/vec3/vec4` cannot be used as type hints since [this line of code](https://github.com/taichi-dev/taichi/blob/master/python/taichi/lang/kernel_impl.py#L723) checks the id of the needed type.

Shall we change that line referenced above in `kernel_impl.py` to the following?

```python
if needed.dtype in primitive_types.real_types:
```

strongoier:
Some notes for people interested in the cause of this issue - 

The type of `default_fp` is actually `DataType` in C++, and a `DataType` stores a pointer to its actual `Type`. The actual `Type` is unique, but the `DataType` wrapper can be different.

PR #5792, opened 2022-08-16
[Lang] Improve loop error message and remove the check for real type id

neozhaoliang:
related issue = #5767

### DIFF
--- a/python/taichi/lang/impl.py
+++ b/python/taichi/lang/impl.py
@@ -86,7 +86,7 @@ def expr_init_func(
 def begin_frontend_struct_for(ast_builder, group, loop_range):
     if not isinstance(loop_range, (AnyArray, Field, SNode, _Root)):
         raise TypeError(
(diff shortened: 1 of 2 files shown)


python/taichi/lang/impl.py
```diff
@@ -86,7 +86,7 @@ def expr_init_func(
 def begin_frontend_struct_for(ast_builder, group, loop_range):
     if not isinstance(loop_range, (AnyArray, Field, SNode, _Root)):
         raise TypeError(
-            'Can only iterate through Taichi fields/snodes (via template) or dense arrays (via types.ndarray)'
+            f"Cannont loop over the object {type(loop_range)} in Taichi scope, can only iterate through Taichi fields/snodes (via template) or dense arrays (via types.ndarray)"
```
strongoier, line 89:
```suggestion
            f"Cannot loop over the object {type(loop_range)} in Taichi scope. Only Taichi fields (via template) or dense arrays (via types.ndarray) are supported."
```

strongoier (approved):
Thanks!

PR #5803, opened 2022-08-17
[test] Add a cgraph test with template args in ti.func

### LINKED ISSUE #5771: [feature request] [aot] Support custom struct parameters for AOT
**Describe the bug**
It seems that the aot module does not support custom structs currently. Could we support this feature recently?

**An example**

```py
MyStruct = ti.types.struct(
  id=ti.i32,
  val=ti.f32,
  center=ti.types.vector(3, ti.f32),
  offset=ti.types.vector(4, ti.f32),
  color=ti.types.vector(4, ti.i32),
)


@ti.func
def aot_test_func(x: ti.template()):
  x.id = 0
  x.val = 1.0
  x.center = ti.Vector([0.0, 0.0, 0.0])
  # ...

@ti.kernel
def aot_test_kernel():
  x = MyStruct()
  aot_test_func(x)
```

ailzhang:
Note that we don't support passing `ti.types.struct` to `ti.kernel` yet
so the only place you can pass it as `ti.template()` is inside
`ti.func`. Since `ti.func` is inlined this is "naturally" supported.

fixes #5771

### DIFF
--- a/tests/python/test_graph.py
+++ b/tests/python/test_graph.py
@@ -413,3 +413,38 @@ def test_texture():
     for i in range(res[0]):
         for j in range(res[1]):
             assert test_utils.allclose(pixels[i, j], [0.1, 0.1, 0.1, 1.])
+
+
+@test_utils.test(arch=supported_archs_cgraph)
+def test_ti_func_with_template_args():
+    MyStruct = ti.types.struct(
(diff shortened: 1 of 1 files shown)


YuCrazing (approved):
LGTM!

PR #5805, opened 2022-08-17
[autodiff] Add grad type for SNode

erizmr:
Related issue = #
There are only two grad types for a SNode: `primal` or `not primal (i.e., adjoint)`. However, after incorporating forward mode, there is a new grad type `dual`. The global data access rule checker further requires a `adjoint_checkbit` SNode type. 

In addition, the parents of a `adjoint` or `dual` SNode are not limited to `primal` SNode in the implementation of second-order derivatives, i.e., there are `dual of dual`, `dual of grad` etc. To help manage the gradient SNodes, this PR propose to add a grad types for each SNode. The grad types include: `primal`, `adjoint`, `dual` and `adjoint_checkbit`.


ailzhang (approved):
Thanks!!

PR #5808, opened 2022-08-17
[doc] Updated compilation warnings

Vissidarte-Herman:
Finally find a slot to update it. ~_~

Related issue = #

### DIFF
--- a/docs/lang/articles/contribution/contributor_guide.md
+++ b/docs/lang/articles/contribution/contributor_guide.md
@@ -330,32 +330,32 @@ Below we highlight some of the most widely used styles.
 - Mark the constructor `explicit` to prevent the compiler from doing any implicit conversion.
 - Avoid virtual function calls in the constructors or destructors ([explanation](https://wiki.sei.cmu.edu/confluence/display/cplusplus/OOP50-CPP.+Do+not+invoke+virtual+functions+from+constructors+or+destructors)).
 
-## Compilation warnings
+## Deal with compilation warnings
 
-Taichi implements warning-free code by turning on `-Werror` (which treats warnings as errors) by default. We highly recommend resolving a warning the moment when it is raised.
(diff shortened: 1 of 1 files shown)


jim19930609 (approved):

docs/lang/articles/contribution/contributor_guide.md
```diff
@@ -368,10 +368,15 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-nullability-completeness ")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unneeded-internal-declaration ")
 ```
 
-### How to mute warnings for specific lines of code (NOT RECOMMENDED)
-In rare situations where the warnings cannot be fixed nor muted via regular attempts, one of the last things you can try is to decorate your code with the `#pragma clang diagnostic` macro. Beware that `#pragma`s are not part of the C++ standard and strongly depend on the compiler's implementation. That is to say, this solution is neither stable nor elegant.
+### Mute warnings from specific code blocks
+
+:::caution
+The approach presented here is *not* recommended, because it is *not* reliable and hence considered your last approach.
```
jim19930609, line 374: considered as ...?

jim19930609 (approved):
Many thanks!

docs/lang/articles/contribution/contributor_guide.md
Vissidarte-Herman: Yeah, "considered as" is grammatical. But "as" can usually be omitted.

Vissidarte-Herman: However, I'll change the the sentence a bit: 

The approach presented here is *not* recommended and considered your last approach, because it is *not* reliable.

Thanks for pointing this out.

PR #5812, opened 2022-08-18
[test] Refactor opengl and vulkan cpp aot tests

ailzhang:
And move the macro control to cmake. There're a bit more to cleanup in
the cmake but this PR is getting too large, will leave those to followups.

Also we'll come back and update these tests once we cleanup opengl runtime apis :D 

Related issue = #


jim19930609 (approved):
Thanks!

PR #5813, opened 2022-08-18
[vulkan] Fixed crash in SPIR-V CodeGen when a const is declared twice

### LINKED ISSUE #5802: ti.ndrange not supported under vulkan backend
```py
import taichi as ti
import numpy as np
 
ti.init(arch=ti.vulkan)
 
@ti.kernel
def test_loop_under_vulkan(test_arr : ti.types.ndarray()):
    # Wrong case
    for i, j in ti.ndrange(test_arr.shape[0], test_arr.shape[1]):
    # Correct case
    # for I in ti.grouped(test_arr):
    #    i, j = I[0], I[1]
        pass
 
ti_arr = ti.ndarray(ti.f32, (3,3,3))
test_loop_under_vulkan(ti_arr)
```
Cannot use ti.ndrange to loop in a ti.ndarray under vulkan backend. But using ti.grouped is correct, it's a bug I think.

Error information:
```
[E 08/17/22 14:33:37.292 1069443] [spirv_ir_builder.cpp:register_value@1185] tmp1 is existed.
```

PENGUINLIONG:
Resolved #5802.

### DIFF
--- a/taichi/codegen/spirv/spirv_ir_builder.cpp
+++ b/taichi/codegen/spirv/spirv_ir_builder.cpp
@@ -1181,8 +1181,8 @@ void IRBuilder::store_variable(Value pointer, Value value) {
 
 void IRBuilder::register_value(std::string name, Value value) {
   auto it = value_name_tbl_.find(name);
-  if (it != value_name_tbl_.end()) {
-    TI_ERROR("{} is existed.", name);
+  if (it != value_name_tbl_.end() && it->second.flag != ValueKind::kConstant) {
+    TI_ERROR("{} already exists.", name);
   }
   this->debug(
       spv::OpName, value,


taichi/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -1181,8 +1181,8 @@ void IRBuilder::store_variable(Value pointer, Value value) {
 
 void IRBuilder::register_value(std::string name, Value value) {
   auto it = value_name_tbl_.find(name);
-  if (it != value_name_tbl_.end()) {
-    TI_ERROR("{} is existed.", name);
+  if (it != value_name_tbl_.end() && it->second.flag != ValueKind::kConstant) {
```
ailzhang, line 1184: OOC: does this mean we'll have two constant values with the same name? actually would you mind pasting a snippet of IR before and after this PR? Thanks!!

ailzhang (commented):
Thanks!

taichi/codegen/spirv/spirv_ir_builder.cpp
PENGUINLIONG: > OOC: does this mean we'll have two constant values with the same name? actually would you mind pasting a snippet of IR before and after this PR? Thanks!! 

Well actually it's not quite a problem. OpName is technically a non-semantic op in SPIR-V so duplicated names do no harm. Duplicated constant declarations will be optimized out in a later processing by SPIR-Tools.

ailzhang (approved):
Thanks!

PR #5814, opened 2022-08-18
[bug] Enabled NdarrayType & MatrixType annotation parsing for ti.func

### LINKED ISSUE #5616: Is it possible to keep type annotations in ti.func and ti.kernel consistent?
**Describe the bug**

The problem arises when I try to let a `ti.func` and a `ti.kernel` share the same function body. I try to do this through the following code:

```py
vec3 = ti.types.vector(3, ti.f32)

def common_func(
    x1 : vec3,
    x2 : vec3,
    x3 : vec3,
    x4 : vec3,
) -> ti.f32:
    return (x1-x2).dot(x3-x4)

my_func = ti.func(common_func)
my_kernel = ti.kernel(common_func)
```

However, an error occurs because `ti.func` does not support the `vec3` type annotation:
```
[Taichi] version 1.0.5, llvm 10.0.0, commit eea6f74b, linux, python 3.8.10
  File "test.py", line 347, in <module>
    my_func = ti.func(common_func)
  File "xxx/taichi-1/python/taichi/lang/kernel_impl.py", line 61, in func
    fun = Func(fn, _classfunc=is_classfunc, is_real_function=is_real_function)
  File "/xxx/taichi-1/python/taichi/lang/kernel_impl.py", line 194, in __init__
    self.extract_arguments()
  File "xxx/taichi-1/python/taichi/lang/kernel_impl.py", line 311, in extract_arguments
    raise TaichiSyntaxError(
taichi.lang.exception.TaichiSyntaxError: Invalid type annotation (argument 0) of Taichi function: <taichi.lang.matrix.MatrixType object at 0x7fac675e86a0>
```
So is it possible to keep type annotations in `ti.func` and `ti.kernel` consistent? This will make it easier to let a `ti.func` and a `ti.kernel` share the same function body.

BTW, in this case, `ti.template()` does not solve my problem because I would like to pass the parameter **by value**, not by reference. Moreover, `ti.template()` may take a long compilation time, which is not what I expect.

jim19930609:
Related issue = fix #5616

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -6,7 +6,7 @@ from collections import ChainMap
 from sys import version_info
 
 from taichi._lib import core as _ti_core
-from taichi.lang import expr, impl, kernel_arguments, matrix, mesh
+from taichi.lang import any_array, expr, impl, kernel_arguments, matrix, mesh
 from taichi.lang import ops as ti_ops
 from taichi.lang._ndrange import _Ndrange, ndrange
 from taichi.lang.ast.ast_transformer_utils import (Builder, LoopStatus,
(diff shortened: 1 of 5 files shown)


ailzhang (commented):
LGTM! will leave this PR for @strongoier to take another look :D

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -582,8 +582,58 @@ def transform_as_kernel():
                     # Template arguments are passed by reference.
                     if isinstance(ctx.func.arguments[i].annotation,
                                   annotations.template):
+
+                        ctx.create_variable(ctx.func.arguments[i].name, data)
+                        continue
+
+                    # Ndarray arguments are passed by reference.
```
strongoier, line 589: This part should behave the same as https://github.com/taichi-dev/taichi/blob/365e755190643c9f4504fbe7f990bd167762dfe0/python/taichi/lang/kernel_impl.py#L345-L355

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -582,8 +582,58 @@ def transform_as_kernel():
...
+                                i].annotation.layout is not None and data.layout != ctx.func.arguments[
+                                    i].annotation.layout:
+                            raise TaichiSyntaxError(
+                                f"Argument {arg.arg} of type {ctx.func.arguments[i].annotation} is expected to be a ndarray or matrix with layout {ctx.func.arguments[i].annotation.layout}, but got {data.layout}."
+                            )
+
                         ctx.create_variable(ctx.func.arguments[i].name, data)
                         continue
+
+                    # Matrix arguments are passed by reference.
```
strongoier, line 615: Let's make it pass by value to behave the same as `ti.kernel`.

python/taichi/lang/ast/ast_transformer.py
jim19930609: hmmm, AnyArray doesn't have "check_matched()" and "get_type()" yet, I'll leave dtype check for AnyArray as is for now

strongoier: `check_matched()` is already there. You need to implement `get_type()` though.

jim19930609: done, thanks!

strongoier (commented):
LGTM!

nit: Let's also add tests for invalid cases.

jim19930609:
> LGTM!
> 
> nit: Let's also add tests for invalid cases.

Thanks! Added failure tests to check the invalid cases. 

However, the AnyArray object still uses primitive dtype as its element_type instead of creating an actual TensorType, therefore the failure tests can not cover all the exceptions. This should get fixed along with MatrixType refactor.

strongoier (approved):
LGTM!

PR #5816, opened 2022-08-18
[dx12] Drop code for dx12.

python3kgae:
Add files for dx12.
With TI_NOT_IMPLEMENTED for all methods.

Related issue = #5276


jim19930609 (approved):
Thanks!

PR #5818, opened 2022-08-18
[Bug] [type] Fix wrong type cast in codegen of storing quant floats

### LINKED ISSUE #5817: Codegen of storing quant floats fails when physical type has 64 bits
The following code snippet,
```python
import taichi as ti

ti.init()

qflt = ti.types.quant.float(exp=8, frac=23)
x = ti.field(dtype=qflt)

bitpack = ti.BitpackedFields(max_num_bits=64)
bitpack.place(x)
ti.root.place(bitpack)

@ti.kernel
def foo():
    for i in range(1):
        x[None] = 1

foo()
```
reports
```
python3: /home/yuanming/repos/llvm-10.0.0.src/lib/IR/Instructions.cpp:2741: static llvm::CastInst *llvm::CastInst::Create(Instruction::CastOps, llvm::Value *, llvm::Type *, const llvm::Twine &, llvm::Instruction *): Assertion `castIsValid(op, S, Ty) && "Invalid cast!"' failed.
```
If `max_num_bits` is 32, then everything works fine.

strongoier:
Related issue = #4857, fix #5817

### DIFF
--- a/taichi/codegen/llvm/codegen_llvm_quant.cpp
+++ b/taichi/codegen/llvm/codegen_llvm_quant.cpp
@@ -221,7 +221,7 @@ void TaskCodeGenLLVM::visit(BitStructStoreStmt *stmt) {
           create_call("max_i32", {exponent_bits, tlctx->get_constant(0)});
 
       // Compute the bit pointer of the exponent bits.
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Thanks!

PR #5825, opened 2022-08-18
[vulkan] Support offline cache on Vulkan

PGZXB:
Related issue = #4401
Working in progress: Not multi-thread/process safe now.
I will fix this problem and other potential bugs in next PRs.

Testing on my Win11:


strongoier (approved):
LGTM!

PR #5831, opened 2022-08-19
[Doc] Update type cast

neozhaoliang:


docs/lang/articles/type_system/type.md
```diff
@@ -79,14 +79,16 @@ In the Taichi scope, the type of a variable is *statically typed* upon initializ
...
+- You can also use primitive types like `ti.f32`, `ti.i64` to convert a variable of scalar type to other types:
 
   ```python
   @ti.kernel
   def foo():
       a = 3.14
-      b = int(a)    # 3
-      c = float(b)  # 3.0
+      x = int(a)    # 3
+      y = float(b)  # 3.0
```
strongoier, line 89: What is `b` here?

docs/lang/articles/type_system/type.md
```diff
@@ -79,14 +79,16 @@ In the Taichi scope, the type of a variable is *statically typed* upon initializ
...
+      x1 = ti.i32(x)  # 3
+      y1 = ti.f64(x)  # 3.14
```
strongoier, line 91: Converting from `x` will not get 3.14.

neozhaoliang: Sorry my mistake, it should be fixed now.

docs/lang/articles/type_system/type.md
```diff
@@ -79,14 +79,16 @@ In the Taichi scope, the type of a variable is *statically typed* upon initializ
...
+- You can also use primitive types like `ti.f32`, `ti.i64` to convert a variable of scalar type to other scalar types:
 
   ```python
   @ti.kernel
   def foo():
       a = 3.14
-      b = int(a)    # 3
-      c = float(b)  # 3.0
+      x = int(a)    # 3
+      y = float(a)  # 3.0
```
strongoier, line 89: 3.0 is incorrect.

docs/lang/articles/type_system/type.md
```diff
@@ -79,14 +79,16 @@ In the Taichi scope, the type of a variable is *statically typed* upon initializ
       c = ti.cast(b, ti.f32)  # 3.0
   ```
 
-- You can also use Python's builtin `int()` or `float()` to convert a value to the default integer type or to the default floating-point type:
+- You can also use primitive types like `ti.f32`, `ti.i64` to convert a variable of scalar type to other scalar types:
```
Vissidarte-Herman, line 82:
```suggestion
- As of v1.1.0, you are allowed to use primitive types such as `ti.f32` and `ti.i64` to convert a scalar variable to a different scalar type:
```

Vissidarte-Herman (changes requested):
Minor editorial udpates. Otherwise, LGTM!

Vissidarte-Herman (approved):
LGTM!

PR #5836, opened 2022-08-22
[example] Making gravity option used in stable_fluid_ggui.py

mshoe:
Related issue = #

I was just tinkering with the stable_fluid_ggui.py example, and noticed that the `gravity = True` option wasn't being used, so I added it into the apply_impulse kernel. You can test that it works now by setting `gravity = False`.


ailzhang (approved):
Nice catch! Thank you!

PR #5840, opened 2022-08-22
[Lang] Make python scope comparison return 1 instead of -1

### LINKED ISSUE #5746: vec3 comparison returns -1 in python scope
To reproduce:

```python
import taichi as ti
ti.init()

v = ti.math.vec3(0, 1, 2)
print(v < 1)  # [-1, 0, 0]
```

re-xyr:
Related issue = Fixes #5746

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -898,10 +898,10 @@ class Matrix(TaichiOperations):
             >>> v.any()
             True
         """
-        ret = ops_mod.cmp_ne(self.entries[0], 0)
-        for i in range(1, len(self.entries)):
-            ret = ret + ops_mod.cmp_ne(self.entries[i], 0)
-        return -ops_mod.cmp_lt(ret, 0)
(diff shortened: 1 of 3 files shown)


strongoier (approved):
Thanks!
nit: fix the pylint error

python/taichi/lang/matrix.py
```diff
@@ -898,10 +898,10 @@ def any(self):
             >>> v.any()
             True
         """
-        ret = ops_mod.cmp_ne(self.entries[0], 0)
-        for i in range(1, len(self.entries)):
```
strongoier, line 902: Rename `i` to `e` or something?

re-xyr: I guess `elem`

strongoier: Maybe `entry` is better suited in this file

re-xyr: You're right

Issue #5819, opened 2022-08-18
Taichi Matrix & Vector refactor plan


jim19930609:
# Background and Purpose
Currently, Taichi scalarizes all the Matrix and corresponding operations at Python level. This approach has major **drawbacks** as follow:
1. No actual SIMD support for Matrix data type during CodeGen.
2. Lack of Matrix information at CHI IR level, thus not able to write any optimization pass wrt Matrix operations.
3. Huge overhead to migrate Matrix support to another Frontend Language (Julia, Javascript, etc)

To address the above mentioned drawbacks, we describe our refactor **goal** as follow:
1. Remove early scalarization at Python-level and make full Matrix information visible to CHI IR level.
2. Enable the ability to optimize Matrix operations via CHI IR Passes.
3. Enable SIMD instructions code generation for Matrix operations.
4. Minimize Matrix handling at Python-level to enhance its multiple Frontend Language support
5. No regression wrt functionality, performance and compilation efficiency especially with **dynamic indexing** on

# Design Overview
In general there are three separate code paths with Matrix involved:
1. local Matrix 
2. MatrixField：Field + Matrix
3. MatrixNdarray：Ndarray + Matrix

Each of the code path involves four typical use cases:
1. Indexed Matrix operations：
    - Matrix Indexing
    - Indexed Matrix Access：Load, Store, Add, Mul, ...
3. MatrixType operations:
    - Matrix initialization
    - Matrix operations：Load, Store, matmul, transpose, atomic, ...

# Design Details
## Local Matrix
Implementation plan: https://github.com/taichi-dev/taichi/issues/5478

![image](https://user-images.githubusercontent.com/22334008/185334562-4a24c981-8ad7-4ee7-9417-74366aa750a5.png)

### 1. Representation
ti.Matrix returns a Expr(MatrixExpression) with TensorType
```
@ti.kernel
def example():
    x = ti.Matrix([[1, 2], [3, 4]], dt=ti.f32)
    print(x)
```

### 2. IR changes
[New Frontend IR] MatrixExpression: Initializing a matrix with ti.Matrix will result in the following Frontend IR:
```
kernel {
  $0 = alloca @tmp0
  @tmp0 = [(cast_value<f32> 1.0), (cast_value<f32> 2.0), (cast_value<f32> 3.0), (cast_value<f32> 4.0)] (dt=[Tensor (2, 2) f32])
  $2 = alloca @tmp1
  @tmp1 = @tmp0
[...]

PR #5844, opened 2022-08-23
[Lang] Add @python_scope decorator for selected MatrixNdarray/VectorNdarray methods

jim19930609:
Related issue = #5819

### DIFF
--- a/python/taichi/lang/_ndarray.py
+++ b/python/taichi/lang/_ndarray.py
@@ -75,6 +75,7 @@ class Ndarray:
         else:
             self._fill_by_kernel(val)
 
+    @python_scope
     def _ndarray_to_numpy(self):
         """Converts ndarray to a numpy array.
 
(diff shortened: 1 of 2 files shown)


AD1024:
Looks good. Thanks for adding these!
One question: are these methods supposed to be called by users? Looks like they are for internal uses. 
Another side note (correct me if I'm wrong):
if there are not for users, probably some could be omitted since they are only called by methods that are callable from Python scope. e.g.
https://github.com/taichi-dev/taichi/blob/205007290a61726af7fd66e46614813157f8a5db/python/taichi/lang/_ndarray.py#L245-L247

AD1024 (approved):

jim19930609:
> Looks good. Thanks for adding these! One question: are these methods supposed to be called by users? Looks like they are for internal uses. Another side note (correct me if I'm wrong): if they are not for users, probably some could be omitted since they are only called by methods that are callable from Python scope. e.g.
> 
> https://github.com/taichi-dev/taichi/blob/205007290a61726af7fd66e46614813157f8a5db/python/taichi/lang/_ndarray.py#L245-L247

Yes, I do agree some of these methods are not intended to be called by the users thus does not matter whether it has explicit "ti_scope" or "python_scope" decorators. However, since the "ti_scope" and "python_scope" is essentially exclusive at least for now, marking the decorators explicitly improves the readability a lot~

PR #5846, opened 2022-08-23
[ci] Use local caches on self-hosted runners, and code refactoring.

feisuzhu:
This PR starts as an attempt to remove actions/cache@v2 usage on self-hosted runners, but in the end a lot of refactoring is done here.

Things resolved in this PR:
1. Use local caches on self-hosted runners (of course)
2. Unified Windows CPU & GPU task build scripts, split build and testing process, major code refactoring.
3. Mechanism for probing available resource to determine maximum parallelism (Windows + CPU tasks only for now)
4. Caching pip download and generated virtual environment
5. Extract docker related logic to ci-docker-run fucntion, simpifying yaml file and build scripts.


ailzhang (approved):
Thanks a ton!! 
nit: as discussed offline let's create an issue to track CI/CD pipeline overhaul, it'd be nice to have two images with "existing" and "target" workflow pipeline there ;)

ailzhang:
Also let's make sure this PR only gets merged after v1.1.3 branch cut. :P cc: @strongoier

jim19930609 (approved):

PR #5860, opened 2022-08-23
[Lang] Support linear system solving on GPU with cuSolver

FantasyVR:
Related issue = #2906

We implement a basic linear system solver according to [cuSolverSp_LinearSolver](https://github.com/NVIDIA/cuda-samples/blob/master/Samples/4_CUDA_Libraries/cuSolverSp_LinearSolver/cuSolverSp_LinearSolver.cpp). 

The demo is shown in `misc/test_coo_cusolver.py`.  To solve a linear system, you could follow the steps as below:
1. Create three `ti.ndarrays` according to COO format, the datatype of row/column indices ndarray should be `ti.i32` and the datatype of `value` ndarray should be `ti.f32`. 
2. Create a `ti.linagl.SparseMatrix` as `A`.
3. Build the sparse matrix using `build_coo` function. The sparse matrix will be stored in CSR format.
4. Create a `ti.linalg.SparseSolver`.
5. Create two `ti.ndarrray` to store the right hand side `b` of linear system and the unkonw variables `x`.
6. Solve the linear system `A * x = b` with `solve_cu` function.


misc/test_coo_cusolver.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@ti.kernel
+def print_x(x: ti.types.ndarray(), ncols: ti.i32, length: ti.i32):
+    for i in range(ncols - length, ncols):
+        print(x[i])
+
+
+print(">> load sparse matrix........")
+A_raw_coo = sio.mmread('misc/lap2D_5pt_n100.mtx')
```
strongoier, line 22: The file seems too large. Is it possible to create a smaller one or shall we move it elsewhere?

taichi/program/sparse_solver.cpp
```diff
@@ -67,6 +67,176 @@ bool EigenSparseSolver<EigenSolver, EigenMatrix>::info() {
...
+  }
+
+  // step 4: solve B*z = Q*b
+  float tol = 1e-6;
+  int reorder = 1;
+  int singularity = 0; /* -1 if A is invertible under tol. */
+  CUSOLVERDriver::get_instance().csSpScsrlsvcholHost(
+      handle, nrows, nnz, descrA, (void *)h_val_B, (void *)hrow_offsets_B,
+      (void *)hcol_indices_B, (void *)h_Qb, tol, reorder, (void *)h_z,
+      &singularity);
```
Hanke98, line 207: It seems that the `singularity` never be cheked later?

taichi/rhi/cuda/cusparse_functions.inc.h
```diff
@@ -8,6 +8,10 @@ PER_CUSPARSE_FUNCTION(cpDestroy, cusparseDestroy, cusparseHandle_t);
 PER_CUSPARSE_FUNCTION(cpCreateCoo, cusparseCreateCoo, cusparseSpMatDescr_t*, int, int, int,void*, void*, void*,cusparseIndexType_t, cusparseIndexBase_t,cudaDataType );
 PER_CUSPARSE_FUNCTION(cpCreateCsr, cusparseCreateCsr, cusparseSpMatDescr_t*, int, int, int,void*, void*, void*,cusparseIndexType_t, cusparseIndexType_t, cusparseIndexBase_t,cudaDataType );
 PER_CUSPARSE_FUNCTION(cpCoo2Csr, cusparseXcoo2csr, cusparseHandle_t ,const void*, int, int,void*, cusparseIndexBase_t );
+PER_CUSPARSE_FUNCTION(cpCsrGet, cusparseCsrGet, cusparseSpMatDescr_t ,size_t *,size_t *,size_t *,void**,void**,void**,cusparseIndexType_t* ,cusparseIndexType_t* ,cusparseIndexBase_t* ,cudaDataType*);
+PER_CUSPARSE_FUNCTION(cpCreateMatDescr, cusparseCreateMatDescr, cusparseMatDescr_t*);
+PER_CUSPARSE_FUNCTION(cpSetMatType, cusparseSetMatType, cusparseMatDescr_t, cusparseMatrixType_t);
+PER_CUSPARSE_FUNCTION(cpSetMatIndexBase, cusparseSetMatIndexBase, cusparseMatDescr_t, cusparseIndexBase_t);
 PER_CUSPARSE_FUNCTION(cpDestroySpMat, cusparseDestroySpMat, cusparseSpMatDescr_t);
```
Hanke98, line 15: I also added those functions in my local branch. Let me just rebase your code later~

Hanke98 (commented):
Sorry for the late response! Just flew back from a conference today. I will do a more careful review tomorrow!

misc/test_coo_cusolver.py
FantasyVR: Could I download the file when running this script?  This file is used to show that the results of this solver is identical with [cuSolverSp_LinearSolver](https://github.com/NVIDIA/cuda-samples/blob/master/Samples/4_CUDA_Libraries/cuSolverSp_LinearSolver/cuSolverSp_LinearSolver.cpp).

taichi/program/sparse_solver.cpp
```diff
@@ -67,6 +67,176 @@ bool EigenSparseSolver<EigenSolver, EigenMatrix>::info() {
...
+                                                   sizeof(float) * nrows);
+  for (int row = 0; row < nrows; row++) {
+    h_Qb[row] = h_b[h_Q[row]];
+  }
+
+  // step 4: solve B*z = Q*b
+  float tol = 1e-6;
+  int reorder = 1;
+  int singularity = 0; /* -1 if A is invertible under tol. */
+  CUSOLVERDriver::get_instance().csSpScsrlsvcholHost(
```
Hanke98, line 204:
```suggestion
  // use Cholesky decomposition as defualt
  CUSOLVERDriver::get_instance().csSpScsrlsvcholHost(
```

taichi/program/sparse_solver.cpp
```diff
@@ -67,6 +67,176 @@ bool EigenSparseSolver<EigenSolver, EigenMatrix>::info() {
...
+  float *h_z = (float *)malloc(sizeof(float) * ncols);
+  float *h_x = (float *)malloc(sizeof(float) * ncols);
+  float *h_b = (float *)malloc(sizeof(float) * nrows);
+  float *h_Qb = (float *)malloc(sizeof(float) * nrows);
+
+  int *h_Q = (int *)malloc(sizeof(int) * ncols);
+  int *hrow_offsets_B = (int *)malloc(sizeof(int) * (nrows + 1));
+  int *hcol_indices_B = (int *)malloc(sizeof(int) * nnz);
+  float *h_val_B = (float *)malloc(sizeof(float) * nnz);
+  int *h_mapBfromA = (int *)malloc(sizeof(int) * nnz);
```
Hanke98, line 127: I found that `h_z`, `hrow_offsets_B `, `hcol_indices_B ` and `h_val_B ` are not deleted in the end of the function. Could you re-check it? :-)

Hanke98 (commented):
Thanks a lot! LGTM mostly! I only have few nits:

misc/test_coo_cusolver.py
Hanke98: Or, we can generate a matrix randomly here. Just remember to make sure the matrix is inverteable. :-)

strongoier: Downloading is fine :-)

misc/test_coo_cusolver.py
```diff
@@ -0,0 +1,46 @@
...
+d_val_coo = ti.ndarray(shape=nnz, dtype=ti.f32)
+d_row_coo.from_numpy(A_coo.row)
+d_col_coo.from_numpy(A_coo.col)
+d_val_coo.from_numpy(A_coo.data)
+
+A_ti = ti.linalg.SparseMatrix(n=nrows, m=ncols, dtype=ti.float32)
+A_ti.build_coo(d_row_coo, d_col_coo, d_val_coo)
+x_ti = ti.ndarray(shape=ncols, dtype=ti.float32)
+solver = ti.linalg.SparseSolver()
+solver.solve_cu(A_ti, b, x_ti)
```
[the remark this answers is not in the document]
FantasyVR: Good points. It would be better to use a unified API. Right now, the implemented solver process is different with the Eigen which first `analyse` and `factor` the matrix first. We could unify the API later using the preview API of cusolverSP.

misc/test_coo_cusolver.py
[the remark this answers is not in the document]
FantasyVR: Yeah, it's not a good idea to commit the raw `.mtx` files to Taichi repo. We could download it from the Internet. This matrix is used to check the solution is identical to the Cuda samples which also used this matrix as input.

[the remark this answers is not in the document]
FantasyVR: The testing sparse matrix is replaced using a randomly generated PSD matrix.

Hanke98 (approved):
LGTM! Thanks!

strongoier (approved):

Issue #5873, opened 2022-08-24
[RFC] Implementation plan for MatrixNdarray refactor


jim19930609:
Sub-task of https://github.com/taichi-dev/taichi/issues/5819

![image](https://user-images.githubusercontent.com/22334008/186376004-54cbef7d-3f1f-4608-8721-fdf052130958.png)

PR #5875, opened 2022-08-24
[Lang] MatrixNdarray refactor part0: Support direct TensorType construction in Ndarray and refactor use of element_shape

jim19930609:
Related issue = #5873, #5819

This PR belongs to "Part ②" in #5873, which also covers Part ①.

[Major change]
1. Refactored python-scope `TensorType` to construct & keep a `ptr` with C++ TensorType.
2. Refactored C++ Ndarray to represent matrix-element with `TensorType` instead of the `PrimitiveType + element_shape` combination.
3. Adjust python-scope class `MatrixNdarray` and `VectorNdarray` to adopt refactored `TensorType` and `Ndarray`

### DIFF
--- a/python/taichi/lang/_ndarray.py
+++ b/python/taichi/lang/_ndarray.py
@@ -265,7 +265,8 @@ class ScalarNdarray(Ndarray):
 
 class NdarrayHostAccessor:
     def __init__(self, ndarray):
-        if _ti_core.is_real(ndarray.dtype):
+        dtype = ndarray.element_data_type()
+        if _ti_core.is_real(dtype):
 
             def getter(*key):
                 return ndarray.read_float(key)
@@ -273,7 +274,7 @@ class NdarrayHostAccessor:
             def setter(value, *key):
                 ndarray.write_float(key, value)
         else:
-            if _ti_core.is_signed(ndarray.dtype):
+            if _ti_core.is_signed(dtype):
 
                 def getter(*key):
                     return ndarray.read_int(key)
--- a/python/taichi/lang/_ndarray.py
+++ b/python/taichi/lang/_ndarray.py
@@ -5,7 +5,7 @@ from taichi.lang.enums import Layout
 from taichi.lang.util import cook_dtype, python_scope, to_numpy_type
 from taichi.types import primitive_types
 from taichi.types.ndarray_type import NdarrayTypeMetadata
-from taichi.types.utils import is_real
+from taichi.types.utils import is_real, is_signed
 
 
 class Ndarray:
@@ -275,7 +275,7 @@ class NdarrayHostAccessor:
             def setter(value, *key):
                 ndarray.write_float(key, value)
         else:
-            if _ti_core.is_signed(dtype):
+            if is_signed(dtype):
 
                 def getter(*key):
                     return ndarray.read_int(key)
--- a/python/taichi/lang/_ndarray.py
+++ b/python/taichi/lang/_ndarray.py
@@ -5,6 +5,7 @@ from taichi.lang.enums import Layout
 from taichi.lang.util import cook_dtype, python_scope, to_numpy_type
 from taichi.types import primitive_types
 from taichi.types.ndarray_type import NdarrayTypeMetadata
+from taichi.types.utils import is_real
 
 
 class Ndarray:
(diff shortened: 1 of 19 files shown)


python/taichi/lang/_ndarray.py
```diff
@@ -265,15 +265,16 @@ def __repr__(self):
 
 class NdarrayHostAccessor:
     def __init__(self, ndarray):
-        if _ti_core.is_real(ndarray.dtype):
+        dtype = ndarray.element_data_type()
+        if _ti_core.is_real(dtype):
```
AD1024, line 269: Nice to have: maybe let's put `is_real` in `utils.py` under `types`? It is also exported by pybind.

jim19930609: done, thx!

ailzhang (approved):
nit: IIUC this PR will break our android demo (it doesn't probably due to pinned commit in taichi-aot-demo), would you mind preparing a PR in taichi-aot-demo and make sure we move the pinned commit forward to keep this compatible? Thanks a lot!

taichi/python/export_lang.cpp
```diff
@@ -1106,7 +1107,14 @@ void export_lang(py::module &m) {
            py::return_value_policy::reference)
       .def("get_quant_float_type", &TypeFactory::get_quant_float_type,
            py::arg("digits_type"), py::arg("exponent_type"),
-           py::arg("compute_type"), py::return_value_policy::reference);
+           py::arg("compute_type"), py::return_value_policy::reference)
+      .def(
+          "get_tensor_type",
+          [&](TypeFactory *factory, std::vector<int> shape,
+              const DataType &element_type) {
+            return factory->create_tensor_type(shape, element_type);
```
strongoier, line 1115: Why not use `TypeFactory::get_tensor_type`?

python/taichi/lang/_ndarray.py
```diff
@@ -265,15 +266,16 @@ def __repr__(self):
...
+        if is_real(dtype):
 
             def getter(*key):
                 return ndarray.read_float(key)
 
             def setter(value, *key):
                 ndarray.write_float(key, value)
         else:
-            if _ti_core.is_signed(ndarray.dtype):
+            if _ti_core.is_signed(dtype):
```
strongoier, line 278: Use `is_signed` from `taichi.types.utils`?

taichi/ir/type.cpp
```diff
@@ -64,6 +64,22 @@ DataType DataType::ptr_removed() const {
   }
 }
 
+std::vector<int> DataType::get_element_shape() const {
```
strongoier, line 67: `element_shape` means shape of one's element, which is for `ndarray`. A `TensorType` should only have `shape` instead of `element_shape`.

taichi/python/export_lang.cpp
jim19930609: Oh it's for convenient use at python-scope.

Basically, `TypeFactory::get_tensor_type` takes "Type*" and returns "Type*". However in python-scope, we usually use "DataType", and you have to write sth like:

```
if isinstance(type, _ti_python_core.DataType):
        type = type.get_ptr()
type = _type_factory.get_tensor_type(bits, signed, type)
return cook_type(type)
```

python/taichi/lang/_ndarray.py
jim19930609: done, thx!

taichi/ir/type.cpp
jim19930609: done. thx

jim19930609:
> nit: IIUC this PR will break our android demo (it doesn't probably due to pinned commit in taichi-aot-demo), would you mind preparing a PR in taichi-aot-demo and make sure we move the pinned commit forward to keep this compatible? Thanks a lot!

Per offline discussion with @ailzhang, since there's an overloaded Ndarray constructor which turns PrimitiveType + element_shape arguments into TensorType, there shouldn't be any regression w.r.t taichi-aot-demo.

Added some comments and assertions to clarify this overloaded Ndarray constructor.

taichi/python/export_lang.cpp
strongoier: Got it.

strongoier (approved):
LGTM!

Issue #5627, opened 2022-08-04
CUPTI Profiler requires several updates to work properly


BA-JackHE:
Dear Taichi Development Team,
 
I've been trying to use the CUPTI profiler on MPM samples to analyze the bottlenecks of the generated sub kernels. And I encountered and solved several problems regarding this:

1. kernel_profiler doesn't match the current CuptiMetric class.
https://github.com/taichi-dev/taichi/blob/93fdf4955b790cd7159ff2a057c9cf93899bee3b/python/taichi/profiler/kernel_metrics.py#L56
https://github.com/taichi-dev/taichi/blob/93fdf4955b790cd7159ff2a057c9cf93899bee3b/python/taichi/profiler/kernel_profiler.py#L322
"metric_list[idx]**.format**" should be "metric_list[idx]**.val_format**"

2. The documentation for advanced profiler (CUPTI) is very limited. And some code example still has typos in them, such as:
https://github.com/taichi-dev/taichi/blob/93fdf4955b790cd7159ff2a057c9cf93899bee3b/python/taichi/profiler/kernel_profiler.py#L527
"set_kernel_profile_metrics" should be "set_kernel_**profiler**_metrics"

3. Some predefined CuptiMetric, such as sm_throughput, doesn't work with the latest CUTPI (CUDA 11.7).

Most of the problems mentioned above are easy to fix. I'd like to open a PR to address these issues. Please let me know if it works for you and what should I pay attention to regarding the PR. Thank you!

jim19930609:
Hi BA-JackHE,
Thanks so much for figuring out these issues as well as fixing them for us, this is awesome! A fix PR is always welcomed, and it would be even better if you can turn your use case to an unit test to protect the fix. Thanks!

BA-JackHE:
> Hi BA-JackHE, Thanks so much for figuring out these issues as well as fixing them for us, this is awesome! A fix PR is always welcomed, and it would be even better if you can turn your use case to an unit test to protect the fix. Thanks!

Great! I'll open a fix PR when it's ready. There's still some remaining work for testing. About the unit test, I'll see what I can do.

PR #5879, opened 2022-08-25
[bug] Fix minor issue for CUPTI kernel profiler

BA-JackHE:
Related issue = #5627 

About SM Throughput metric, since "sm__throughput" metric cannot be found in the latest CUPTI document, please let me know if the relative modification is correct.


turbo0628:
LGTM, thanks for the fix!

turbo0628 (approved):
LGTM, thanks for the fix!

PR #5881, opened 2022-08-25
[Lang] MatrixNdarray refactor part1: Refactor Taichi kernel argument to use TensorType

jim19930609:
Related issue = #5873, #5819

This PR belongs to "Part ②" in #5873.


AD1024:
LGTM!
Just wondering is there a test case for passing TensorType arguments to a kernel?

jim19930609:
> LGTM! Just wondering will there be a test case for passing TensorType arguments to a kernel?

Thanks! Basically this is a refactor PR and we're relying on existing unit tests for verification purpose. The entrance to kernel argument construction is `decl_ndarray_arg`:

image

and `decl_ndarray_arg` is called whenever it parses a `ti.types.ndarray()` in the argument annotation.

image

Therefore existing MatrixNdarray test cases such as those in `test_ndarray.py` should be able to cover this change:

image

taichi/program/callable.h
```diff
@@ -16,22 +16,45 @@ class TI_DLL_EXPORT Callable {
...
                  bool is_array = false,
                  std::size_t size_unused = 0,
                  int total_dim = 0,
-                 std::vector<int> element_shape = {})
-        : dt(dt),
-          is_array(is_array),
-          total_dim(total_dim),
-          element_shape(std::move(element_shape)) {
+                 std::vector<int> element_shape = {}) {
+      if (dt->is<PrimitiveType>() && element_shape.size() > 0) {
```
AD1024, line 28: Could you explain this case? Why is it possible that the user declares a value with some primitive type but has non-zero shape? I thought we handle all the cases where `element_shape.size() > 0` at the frontend. Thanks!

jim19930609: Good point, thanks! 

In general it is a historical issue. Taichi used to represent TensorType with a combination of "PrimitiveType" & "element_shape" and there are a bunch of interfaces designed like this (it allows the users to create TensorType by passing in PrimitiveType + element_shape)

In this refactor PR, we mainly wanna remove the "element_shape" member in the objects (class Arg, class ExternalTensorExpression, ...), and force them to use TensorType in thier "dtype" member. However we still wanna keep the interfaces unchanged so as to minimize possible regressions.

AD1024: Got it. Makes sense to me. Thanks for the explanation!

AD1024 (approved):
LGTM!

taichi/program/callable.h
```diff
@@ -16,22 +16,45 @@ class TI_DLL_EXPORT Callable {
...
+
+    int get_element_size() const {
+      return data_type_size(dt_);
     }
+
+    DataType get_dtype() const {
+      return dt_;
+    }
+
+   private:
```
ailzhang, line 56: nit: shall we add a note here (or near DataType class) to explain its difference with primitive types as you put in the comment above? 
That way it's easier to ppl who read the codebase. ;)

ailzhang (approved):
LGTM!

taichi/program/callable.h
jim19930609: comments added~

strongoier (approved):

PR #5882, opened 2022-08-25
[refactor] [ir] Remove legacy stmt width

strongoier:
Related issue = #4096

This is one step of removing legacy incomplete vectorization on CPU for better maintainability of current codebase.


ailzhang (approved):
Cool!

PR #5892, opened 2022-08-26
[refactor] [ir] Remove legacy ElementShuffleStmt

strongoier:
Related issue = #4096

In this PR, `lower_vector_ptr()` is rewritten and renamed to `lower_ptr()` so that the legacy `ElementShuffleStmt` can be fully removed.

### DIFF
--- a/taichi/analysis/value_diff.cpp
+++ b/taichi/analysis/value_diff.cpp
@@ -80,15 +80,6 @@ class ValueDiffLoopIndex : public IRVisitor {
     }
   }
 
-  void visit(ElementShuffleStmt *stmt) override {
-    int old_lane = lane;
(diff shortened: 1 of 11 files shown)


ailzhang (approved):
Nice! 
OOC: what was ElementShuffleStmt used for? :P

jim19930609 (approved):
Thanks! 

Btw, just asking: is "LaneAttribute" also removable?

strongoier:
> Nice! OOC: what was ElementShuffleStmt used for? :P

For legacy vectorization.

strongoier:
> Btw, just asking: is "LaneAttribute" also removable?

Yep. I intend to remove these one by one otherwise a single PR will be too large.

PR #5894, opened 2022-08-26
[doc] Updated forward-mode autodiff

Vissidarte-Herman:
Related issue = #


Vissidarte-Herman:
I'm afraid I've garbled with the intended meaning here and there. Please double check.

erizmr (approved):
LGTM! Thanks!

PR #5899, opened 2022-08-29
[aot] Taichi C-API C++ wrapper

PENGUINLIONG:
This PR proposes a C++ wrapper of the C-APIs to elliminate the majority of the redundant codes in AOT module deployment. It also helps to make C-API tests easier to write.


bobcao3:
Which parts of redundant code is being removed by doing this?

c_api/include/taichi/taichi.hpp
```diff
@@ -0,0 +1,652 @@
+// C++ wrapper of Taichi C-API
```
jim19930609, line 1: Are we gonna release a C++ wrapper together with C-API?

c_api/include/taichi/taichi.hpp
```diff
@@ -0,0 +1,652 @@
...
+  ~Memory() {
+    destroy();
+  }
+
+  Memory &operator=(const Memory &) = delete;
+  Memory &operator=(Memory &&b) {
+    destroy();
+    runtime_ = detail::move_handle(b.runtime_);
+    memory_ = detail::move_handle(b.memory_);
+    should_destroy_ = std::exchange(b.should_destroy_, false);
```
jim19930609, line 102: Can you try if this way of using `std::exchange` also works after C++20? Looks like the result has to be constexpr after C++20: 

2022-08-31 09-31-08 的屏幕截图

Otherwise I'd suggest not using unstable std functions since std::exchange doesn't seem like doing something super hard or fancy.

cmake/TaichiCAPITests.cmake
```diff
@@ -23,6 +23,7 @@ target_include_directories(${C_API_TESTS_NAME}
     ${PROJECT_SOURCE_DIR}
     ${PROJECT_SOURCE_DIR}/c_api/include
     ${PROJECT_SOURCE_DIR}/c_api/src
+    ${PROJECT_SOURCE_DIR}/external/Vulkan-Headers/include
```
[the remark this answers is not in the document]
jim19930609: Wait `taichi/taichi.h` will be released to the users. Does that mean users also have to add this include `${PROJECT_SOURCE_DIR}/external/Vulkan-Headers/include` in their Demo in order to use C-API with Vulkan?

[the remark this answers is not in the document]
jim19930609: I was thinking C-API should be self-contained, is it because we re-used some Vulkan data types in C-API interface?

c_api/include/taichi/taichi.hpp
PENGUINLIONG: `std::exchange` becomes an `constexpr` since  C++ 20 and nothing else changes (See the second entry in your screenshot). It is a preferred way to `std::move` a POD type. Please consult this CppCon talk https://youtu.be/ZG59Bqo7qX4?t=1310 for more detail.

PENGUINLIONG:
> Which parts of redundant code is being removed by doing this?

@bobcao3 It provides an official RAII wrapping over the C-API which saves some boilerplates to create ND-Arrays, filling arguments as well as launching kernel. We have an internal wrapping before to help test writing but this will be one of the released items. Talking about the difference, I think https://github.com/taichi-dev/taichi/pull/5899/files#diff-62eb524acf44d17b06f8a0db012503de707825aaef477718a9101548fa5b6d68 is somehow an example.

PENGUINLIONG:
> In general it looks fine, but I'd suggest put C++ wrapper in a separate directory to avoid confusion

@jim19930609 Sounds good

c_api/include/taichi/taichi.hpp
PENGUINLIONG: Y

jim19930609 (approved):
Thanks!

c_api/tests/c_api_aot_test.cpp
```diff
@@ -11,51 +11,25 @@ static void kernel_aot_test(TiArch arch) {
...
-
-  // Kernel Execution
-  constexpr uint32_t arg_count = 2;
-  TiArgument args[arg_count] = {std::move(arg0), std::move(arg1)};
-
-  ti_launch_kernel(runtime, k_run, arg_count, &args[0]);
-  ti_wait(runtime);
+  k_run[0] = arg0_val;
+  k_run[1] = arg1_array;
+  k_run.launch();
```
ailzhang, line 23: nit: do we need to keep the ti_wait (or similar sync) here?

ailzhang (approved):
Thanks!!

.github/workflows/testing.yml
```diff
@@ -262,7 +262,7 @@ jobs:
           .github/workflows/scripts/unix_build.sh
           brew uninstall molten-vk
         env:
-          TAICHI_CMAKE_ARGS: -DTI_WITH_OPENGL:BOOL=OFF -DTI_WITH_CC:BOOL=${{ matrix.with_cc }} -DTI_WITH_VULKAN:BOOL=ON -DTI_WITH_C_API=ON -DTI_BUILD_TESTS:BOOL=${{ matrix.with_cpp_tests }} -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache
+          TAICHI_CMAKE_ARGS: -DTI_WITH_OPENGL:BOOL=OFF -DTI_WITH_CC:BOOL=${{ matrix.with_cc }} -DTI_WITH_VULKAN:BOOL=OFF -DTI_WITH_C_API=ON -DTI_BUILD_TESTS:BOOL=${{ matrix.with_cpp_tests }} -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache
```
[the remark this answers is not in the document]
PENGUINLIONG: Wasn't able to build without including the vulkan headers. Fixed now.

c_api/tests/c_api_aot_test.cpp
PENGUINLIONG: Added

cmake/TaichiCAPITests.cmake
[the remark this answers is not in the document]
PENGUINLIONG: The user can opt out the use of `taichi/taichi_vulkan.h` if they don't need vulkan interop; but in case they want it, they will have Vulkan SDK installed and the headers are guaranteed forward compatible.

PR #5908, opened 2022-08-30
[aot] Add C API for opengl backend

ailzhang:
* (to be filled)

And c-api tests are enabled for opengl as well

### DIFF
--- a/c_api/src/taichi_llvm_impl.cpp
+++ b/c_api/src/taichi_llvm_impl.cpp
@@ -1,8 +1,6 @@
 #include "taichi_core_impl.h"
 #include "taichi_llvm_impl.h"
 
-#ifdef TI_WITH_LLVM
-
 #include "taichi/program/compile_config.h"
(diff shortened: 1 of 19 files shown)


c_api/src/taichi_llvm_impl.cpp
```diff
@@ -1,8 +1,6 @@
 #include "taichi_core_impl.h"
 #include "taichi_llvm_impl.h"
 
-#ifdef TI_WITH_LLVM
```
PENGUINLIONG, line 4: Woulnd't it cause any compilation problem?

PENGUINLIONG (approved):
lgtm

c_api/src/taichi_llvm_impl.cpp
ailzhang: nah it's handled in cmake, when TI_WITH_LLVM is off this source file won't be added

PR #5909, opened 2022-08-30
[aot] [test] Fix c api aot tests for vulkan and opengl backend

ailzhang:
* __->__ #5909
* #5908


PENGUINLIONG (changes requested):
I think we should keep it on always. The backend should properly ignore this flag if the data is in RAM, directly visible to the host

ailzhang:
@PENGUINLIONG that sounds good, will throw a warning in LLVM backends instead of erroring out :D

jim19930609 (approved):
Thanks!

Issue #5835, opened 2022-08-22
GGUI error if CUDA_VISIBLE_DEVICES is not 0


zhenjia-xu:
**Describe the bug**
I want to use taichi on a multi-gpu server. But GGUI has an error if the `CUDA_VISIBLE_DEVICES` is not 0

**To Reproduce**

```py
import os
os.environ['CUDA_VISIBLE_DEVICES'] = '1'
import taichi as ti

ti.init(arch=ti.cuda, device_memory_GB=1, packed=True)
window = ti.ui.Window("Test", (960, 960), vsync=True, show_window=False)
canvas = window.get_canvas()
scene = ti.ui.Scene()

pos = ti.Vector.field(3, dtype=ti.f32, shape=(1))
scene.particles(pos, color=(0.5, 0.5, 0.5, 1), radius=0.01)

canvas.scene(scene)
```

**Log/Screenshots**
```
-> % python bug.py
[Taichi] version 1.1.2, llvm 10.0.0, commit f25cf4a2, linux, python 3.7.12
[Taichi] Starting on arch=cuda
[I 08/21/22 22:08:31.428 688283] [vulkan_device_creator.cpp:pick_physical_device@372] Found Vulkan Device 0 (llvmpipe (LLVM 12.0.0, 256 bits))
[I 08/21/22 22:08:31.428 688283] [vulkan_device_creator.cpp:pick_physical_device@372] Found Vulkan Device 1 (NVIDIA GeForce RTX 3090)
[I 08/21/22 22:08:31.428 688283] [vulkan_device_creator.cpp:pick_physical_device@372] Found Vulkan Device 2 (NVIDIA GeForce RTX 3090)
[I 08/21/22 22:08:31.428 688283] [vulkan_device_creator.cpp:pick_physical_device@372] Found Vulkan Device 3 (NVIDIA GeForce RTX 3090)
[I 08/21/22 22:08:31.429 688283] [vulkan_device_creator.cpp:pick_physical_device@372] Found Vulkan Device 4 (NVIDIA GeForce RTX 3090)
[I 08/21/22 22:08:31.429 688283] [vulkan_device_creator.cpp:pick_physical_device@372] Found Vulkan Device 5 (NVIDIA GeForce RTX 3090)
[I 08/21/22 22:08:31.429 688283] [vulkan_device_creator.cpp:pick_physical_device@372] Found Vulkan Device 6 (NVIDIA GeForce RTX 3090)
[...]

PR #5910, opened 2022-08-30
[doc] Add explanatation of usage of TI_VISIBLE_DEVICE and CUDA_VISIBLE_DEVICES

Morcki:
Related issue = #5835 

Here we unify the value of `CUDA_VISIBLE_DEVICES` and `TI_VISIBLE_DEVICE` if only one is set.

### DIFF
--- a/docs/lang/articles/reference/global_settings.md
+++ b/docs/lang/articles/reference/global_settings.md
@@ -18,7 +18,7 @@ sidebar_position: 3
 
 :::note
 
-Under the case that you may want to use cuda as taichi kernel backend, and vulkan as GGUI backend. If you have multiple GPU devices, then you should take care of if `CUDA_VISIBLE_DEVICES` is matched with `TI_VISIBLE_DEVICE`. In general, `CUDA_VISIBLE_DEVICES` and `TI_VISIBLE_DEVICE` should indicate the same uuid of a GPU device. Use `nvidia-smi -L` to quiery the details of your GPU devices.
+In case you want to use taichi cuda backend together with GGUI on a machine with multiple GPU cards, please make sure `CUDA_VISIBLE_DEVICES` matches `TI_VISIBLE_DEVICE` if any of them exists. In general, `CUDA_VISIBLE_DEVICES` and `TI_VISIBLE_DEVICE` should point to a GPU device with the same UUID. Use `nvidia-smi -L` to query the details of your GPU devices.
 
 ## Compilation
 
--- a/docs/lang/articles/reference/global_settings.md
+++ b/docs/lang/articles/reference/global_settings.md
@@ -9,10 +9,14 @@ sidebar_position: 3
 - To specify which kind of architecture (Arch) to use: `ti.init(arch=ti.cuda)`.
 - To specify the pre-allocated memory size for CUDA:
   `ti.init(device_memory_GB=0.5)`.
-- To specify which GPU to use for CUDA:
-  `export CUDA_VISIBLE_DEVICES=[gpuid]`.
+- To specify which GPU to use for CUDA(the first one is suggested):
+  `export TI_VISIBLE_DEVICE=[gpuid]` or `export CUDA_VISIBLE_DEVICES=[gpuid]`.
 - To disable a backend (`CUDA`, `METAL`, `OPENGL`) on start up, e.g. CUDA:
   `export TI_ENABLE_CUDA=0`.
+
+:::note
+
+If you set the value of one of the `TI_VISIBLE_DEVICE` and `CUDA_VISIBLE_DEVICES`, then the other's value will be the same. It's not suggested that you set different value of `TI_VISIBLE_DEVICE` and `CUDA_VISIBLE_DEVICES` which will raise a warning message.
 
 ## Compilation
 
--- a/docs/lang/articles/reference/global_settings.md
+++ b/docs/lang/articles/reference/global_settings.md
@@ -9,14 +9,16 @@ sidebar_position: 3
 - To specify which kind of architecture (Arch) to use: `ti.init(arch=ti.cuda)`.
 - To specify the pre-allocated memory size for CUDA:
   `ti.init(device_memory_GB=0.5)`.
-- To specify which GPU to use for CUDA(the first one is suggested):
-  `export TI_VISIBLE_DEVICE=[gpuid]` or `export CUDA_VISIBLE_DEVICES=[gpuid]`.
+- To specify which GPU to use for CUDA:
+  `export CUDA_VISIBLE_DEVICES=[gpuid]`.
+- To specify which GPU to use for VULKAN:
+- `export TI_VISIBLE_DEVICE=[gpuid]`
 - To disable a backend (`CUDA`, `METAL`, `OPENGL`) on start up, e.g. CUDA:
   `export TI_ENABLE_CUDA=0`.
 
 :::note
 
-If you set the value of one of the `TI_VISIBLE_DEVICE` and `CUDA_VISIBLE_DEVICES`, then the other's value will be the same. It's not suggested that you set different value of `TI_VISIBLE_DEVICE` and `CUDA_VISIBLE_DEVICES` which will raise a warning message.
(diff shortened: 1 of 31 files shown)


ailzhang (commented):
Thanks for fixing it! 
nits: 
- add a test to check the expected env variable update? 
- You probably want to raise an error (or at least a warning when both env vars are specified but don't match? 
- It'd be nice to add a faq or section in the docs to tell users how to use this!

Morcki:
> Thanks for fixing it! nits:
> 
> * add a test to check the expected env variable update?
> * You probably want to raise an error (or at least a warning when both env vars are specified but don't match?
> * It'd be nice to add a faq or section in the docs to tell users how to use this!

Not sure if our CI support 2 gpu devices. I will try to add a test. And I will add a warning to warn users to notice that CUDA_VISIBLE_DEVICES is different with TI_VISIBLE_DEVICE

python/taichi/lang/misc.py
```diff
@@ -442,6 +442,15 @@ def init(arch=None,
     print(f'[Taichi] Starting on arch={_ti_core.arch_name(cfg.arch)}')
 
     # user selected visible device
+    if os.environ.get("CUDA_VISIBLE_DEVICES"
+                      ) and not os.environ.get("TI_VISIBLE_DEVICE"):
+        os.environ["TI_VISIBLE_DEVICE"] = os.environ.get(
+            "CUDA_VISIBLE_DEVICES")
+    elif os.environ.get("TI_VISIBLE_DEVICE"
+                        ) and not os.environ.get("CUDA_VISIBLE_DEVICES"):
+        os.environ["CUDA_VISIBLE_DEVICES"] = os.environ.get(
```
jim19930609, line 451: Not sure it works this way ( override CUDA_VISIBLE_DEVICES in Python's os.environ() ). We probably need some test cases to verify/guard this change, but before that we need a CI bot with more than one Nvidia cards...

Morcki:
Unit-test should wait for #5913 fixed

bobcao3 (changes requested):
This will not work. As shown in issue #5835, CUDA devices and Vulkan devices do not use the same indexing method. i.e. You can not just set them to the same number, and you can not assume they appear in the same order as well. In the case from that issue, device 0 in Vulkan is `llvmpipe`, a software implementation.

Morcki:
> This will not work. As shown in issue #5835, CUDA devices and Vulkan devices do not use the same indexing method. i.e. You can not just set them to the same number, and you can not assume they appear in the same order as well. In the case from that issue, device 0 in Vulkan is `llvmpipe`, a software implementation.

Yes, I think so. However there is a solution that may take some work to do while not in this pr.
During cuda context creation, we should store the `uuid` of GPU device being used.
During vulkan context creation, we should fetch the stored `uuid` of GPU device used in CUDA context, and through `vkGetPhysicalDeviceProperties2` to get `VkPhysicalDeviceVulkan11Properties` (also contains `uuid` of each GPU device), and then check if they are matched.

docs/lang/articles/reference/global_settings.md
```diff
@@ -9,11 +9,15 @@ sidebar_position: 3
...
-- To specify which GPU to use for CUDA:
-  `export CUDA_VISIBLE_DEVICES=[gpuid]`.
+- To specify which GPU to use for CUDA(the first one is suggested):
+  `export TI_VISIBLE_DEVICE=[gpuid]` or `export CUDA_VISIBLE_DEVICES=[gpuid]`.
 - To disable a backend (`CUDA`, `METAL`, `OPENGL`) on start up, e.g. CUDA:
   `export TI_ENABLE_CUDA=0`.
 
+:::note
+
+If you set the value of one of the `TI_VISIBLE_DEVICE` and `CUDA_VISIBLE_DEVICES`, then the other's value will be the same. It's not suggested that you set different value of `TI_VISIBLE_DEVICE` and `CUDA_VISIBLE_DEVICES` which will raise a warning message.
```
jim19930609: Setting either `TI_VISIBLE_DEVICE` or `CUDA_VISIBLE_DEVICES` will override the value of the other. Setting different values for `TI_VISIBLE_DEVICE` and `CUDA_VISIBLE_DEVICES` will likely cause undefined behavior thus not recommended.

docs/lang/articles/reference/global_settings.md
```diff
@@ -9,11 +9,15 @@ sidebar_position: 3
 - To specify which kind of architecture (Arch) to use: `ti.init(arch=ti.cuda)`.
 - To specify the pre-allocated memory size for CUDA:
   `ti.init(device_memory_GB=0.5)`.
-- To specify which GPU to use for CUDA:
-  `export CUDA_VISIBLE_DEVICES=[gpuid]`.
+- To specify which GPU to use for CUDA(the first one is suggested):
```
jim19930609: Let's say "(gpuid=0 is recommended)" ?

jim19930609 (commented):
I think we'll need at least one unit test to guard this bug fix?

docs/lang/articles/reference/global_settings.md
Morcki: This is the old docs, I have updated it.

docs/lang/articles/reference/global_settings.md
Morcki: maybe default set is `gpuid=0`?

Morcki:
> I think we'll need at least one unit test to guard this bug fix?

Sure, so I will add a test using our 2-gpu CI bot.

docs/lang/articles/reference/global_settings.md
jim19930609: Oh my bad

jim19930609 (approved):

jim19930609:
Hi @Vissidarte-Herman, do you wanna take a look at the writing part? Thanks!

docs/lang/articles/reference/global_settings.md
```diff
@@ -11,9 +11,15 @@ sidebar_position: 3
...
 - To specify which GPU to use for CUDA:
   `export CUDA_VISIBLE_DEVICES=[gpuid]`.
+- To specify which GPU to use for VULKAN:
+- `export TI_VISIBLE_DEVICE=[gpuid]`
 - To disable a backend (`CUDA`, `METAL`, `OPENGL`) on start up, e.g. CUDA:
   `export TI_ENABLE_CUDA=0`.
 
+:::note
+
+Under the case that you may want to use cuda as taichi kernel backend, and vulkan as GGUI backend. If you have multiple GPU devices, then you should take care of if `CUDA_VISIBLE_DEVICES` is matched with `TI_VISIBLE_DEVICE`. In general, `CUDA_VISIBLE_DEVICES` and `TI_VISIBLE_DEVICE` should indicate the same uuid of a GPU device. Use `nvidia-smi -L` to quiery the details of your GPU devices.
```
jim19930609, line 21: Under the hood, GGUI uses Vulkan to perform rendering. Therefore, in the scenario where CUDA is used as the primary  backend, one should avoid setting different `uuid`s for `CUDA_VISIBLE_DEVICES` and `TI_VISIBLE_DEVICE`, so as to guarantee correct data transfer between CUDA and Vulkan devices. You may use `nvidia-smi -L` to query the details of your GPU devices

Vissidarte-Herman:
I'll take a look earlier next week? @Morcki

docs/lang/articles/reference/global_settings.md
ailzhang, line 21:
```suggestion
In case you want to use taichi cuda backend together with GGUI on a machine with multiple GPU cards, please make sure `CUDA_VISIBLE_DEVICES` matches `TI_VISIBLE_DEVICE` if any of them exists. In general, `CUDA_VISIBLE_DEVICES` and `TI_VISIBLE_DEVICE` should point to a GPU device with the same UUID. Use `nvidia-smi -L` to query the details of your GPU devices.
```

ailzhang (approved):
@Vissidarte-Herman In that case let's maybe get this in first, would you mind sending a new PR for the modification? Thanks a lot!

Issue #5731, opened 2022-08-11
Support to clear adjoint or dual separately 


erizmr:
**Concisely describe the proposed feature**
Current `clear_gradients` clears both adjoint and dual. 

**Describe the solution you'd like (if any)**
Support clear adjoint or dual separately.

erizmr:
Has been fixed by #5911

PR #5911, opened 2022-08-30
[autodiff] Support clear gradients by type

erizmr:
Related issue = #5731

### DIFF
--- a/python/taichi/_kernels.py
+++ b/python/taichi/_kernels.py
@@ -224,7 +224,7 @@ def arr_vulkan_layout_to_field_normal_layout(vk_arr: ndarray_type.ndarray(),
 def clear_gradients(_vars: template()):
     for I in grouped(ScalarField(Expr(_vars[0]))):
         for s in static(_vars):
-            ScalarField(Expr(s))[I] = 0
+            ScalarField(Expr(s))[I] = ops.cast(0, dtype=s.get_dt())
 
 
 @kernel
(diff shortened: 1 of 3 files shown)


python/taichi/_kernels.py
```diff
@@ -224,7 +224,7 @@ def arr_vulkan_layout_to_field_normal_layout(vk_arr: ndarray_type.ndarray(),
 def clear_gradients(_vars: template()):
     for I in grouped(ScalarField(Expr(_vars[0]))):
         for s in static(_vars):
-            ScalarField(Expr(s))[I] = 0
+            ScalarField(Expr(s))[I] = ops.cast(0, dtype=s.get_dt())
```
ailzhang, line 227: OOC: why is this necessary? :P

python/taichi/ad/_ad.py
```diff
@@ -92,7 +95,7 @@ def grad(self):
         self.gradient_evaluated = True
 
 
-def clear_all_gradients():
+def clear_all_gradients(gradient_type=SNodeGradType.ADJOINT):
```
ailzhang, line 98: my gut feeling from this api name is that by default it clears all types of gradients :P

ailzhang (commented):
thanks!!

python/taichi/_kernels.py
erizmr: These are annoying `i32 -> u8 lose precision` warnings if without this.

python/taichi/ad/_ad.py
erizmr: The api name is indeed somehow misleading. I think the initial meaning of the api name might be clearing all the adjoint fields (because there is no other grad type at that time).

ailzhang (approved):
LGTM!

Issue #5517, opened 2022-07-25
Flaky test: [ir.h:as@220] Assertion failure: is<T>()


feisuzhu:
![2022-07-25-235608_2751x1051_scrot](https://user-images.githubusercontent.com/857880/180822345-a749c63b-142a-4992-a3ef-41f0f35f13d6.png)

https://github.com/taichi-dev/taichi/runs/7500311330?check_suite_focus=true

feisuzhu:
Didn't show up for a long time, closing

Issue #5511, opened 2022-07-25
Link modules instead of cloning modules in the codegen in LLVM backends


lin-hitonami:
**Concisely describe the proposed feature**
As described in #5282, the compilation of Taichi kernel becomes slow when there are many SNodeTrees. After profiling, we found that the process of cloning the modules take up most of the time. The cloning happen in three places. 
Two of them happen in the construction process of the struct module as described in #3146. 
https://github.com/taichi-dev/taichi/blob/77398de87e635e4de583de808137e8efe484e8a9/taichi/llvm/llvm_program.cpp#L138-L145
https://github.com/taichi-dev/taichi/blob/77398de87e635e4de583de808137e8efe484e8a9/taichi/llvm/llvm_context.cpp#L455
The other one happens when the LLVM codegen gets the struct module as the initial module of kernel module.
https://github.com/taichi-dev/taichi/blob/d00348c1acccda88c1823e60e151a78b1c4f1b8a/taichi/codegen/llvm/codegen_llvm.cpp#L318-L322

**Describe the solution you'd like (if any)**
We can store the runtime functions, the struct functions of the SNodeTrees, and the kernel functions in different modules, and link the needed modules at the end.

lin-hitonami:
Using an altered version of the program in https://github.com/taichi-dev/taichi/issues/5282:
```
import taichi as ti
from tqdm import tqdm

if __name__ == "__main__":
    ti.init(ti.cpu)
    iterations = 2
    snode_tree_num = 511 # Or 512 without relaxing `constexpr int kMaxNumSnodeTreesLlvm` in taichi/inc/constants.h
    for i in range(iterations):
        snode_trees = []
        for j in tqdm(range(snode_tree_num)):
            x = ti.field(ti.f32)
            fb = ti.FieldsBuilder()
            fb.place(x)
            snodetree = fb.finalize()
            a = x[None]
            snode_trees.append(snodetree)
        for snodetree in snode_trees:
            snodetree.destroy()
```

the module splitting achieved ~7x speed up when the number of SNodeTree is 511.

The current master slows down from ~50iters/s to ~10iters/s, and https://github.com/taichi-dev/taichi/pull/5587 runs stably on ~70iters/s.

The output of master:
```
[Taichi] version 1.0.4, llvm 10.0.0, commit dc93ed8f, linux, python 3.10.4
[Taichi] Starting on arch=x64
[...]

PR #5916, opened 2022-08-30
[misc] Disable parallel compilation

lin-hitonami:
Related issue = #5511 
There are some random failures when parallel compilation is on, like #5517 . So we should disable it before we figure out why.


ailzhang (approved):
Thanks!

PR #5918, opened 2022-08-30
[refactor] [ir] Remove legacy LocalAddress / VectorElement / create_vector_or_scalar_type()

### LINKED ISSUE #4096: Remove `VectorType` and legacy loop vectorization

strongoier:
Related issue = close #4096

As far as I can see, after this PR, the legacy skeleton code for vectorization is fully removed. There might still be missing pieces but I think the codebase is already much cleaner now.


ailzhang (approved):
Thanks!

Issue #5917, opened 2022-08-30
Add unit-tests for GGUI features


Morcki:
**Concisely describe the proposed feature**
Cause there might be problems when using GGUI if we donnot cover the features that we supported and sometimes it would not report bugs. So unit-tests based on visual-checking are needy for CI to automaticlly check if there are some visual bugs here.

**GGUI Unit-test cover range**
- [x] test_fetching_color_attachment ( #5920 )
- [x] test_fetching_depth_attachment (#5947 )
- [x] test_draw_lines (#5948 )
- [x] test_draw_part_of_particles (#5951 )
- [x] test_draw_part_of_mesh (#5955 )
- [x] test_draw_part_of_lines (#5957 )
- [x] test_draw_mesh_instances (#5963 )
- [x] test_draw_part_of_mesh_instances (#5965 )
- [x] test_wireframe_mode (#5967 )

**Notice**
The renderring result on mac is different from other platform, we should handle it separately.

Morcki:
**Codes to generate ground-truth image**
+ test_fetching_color_attachment
```
def test_fetching_color_attachment(filename):
    def write_temp_image_attachment(img, file):
        if img.shape[0] % 2 != 0:
            img = img[:-1]
        if img.shape[1] % 2 != 0:
            img = img[:, :-1]
        ti.tools.imwrite(img, file)
        
    ti.init(arch=ti.vulkan)
    
    window = ti.ui.Window('test', (640, 480), show_window=False)
    canvas = window.get_canvas()

    img = ti.Vector.field(4, ti.f32, (512, 512))

    @ti.kernel
    def init_img():
        for i, j in img:
            img[i, j] = ti.Vector([i, j, 0, 512], dt=ti.f32) / 512

    init_img()

    def render():
        canvas.set_image(img)
        
    render()
    write_temp_image_attachment(window.get_image_buffer_as_numpy(), filename)
    window.destroy()
```

Morcki:
**Codes to generate ground-truth image**
+ test_fetching_depth_attachment
```
def test_fetching_depth_attachment(filename):
    def write_temp_image_attachment(img, file):
        if img.shape[0] % 2 != 0:
            img = img[:-1]
        if img.shape[1] % 2 != 0:
            img = img[:, :-1]
        ti.tools.imwrite(img, file)
        
    window = ti.ui.Window("test", (512, 512), vsync=True, show_window=False)
    canvas = window.get_canvas()
    scene = ti.ui.Scene()
[...]

PR #5920, opened 2022-08-30
[test] [gui] Add a test of fetching color attachment

Morcki:
Related issue = https://github.com/taichi-dev/taichi/issues/5917#issuecomment-1231419913


tests/python/test_ggui.py
```diff
@@ -31,6 +31,20 @@ def write_temp_image(window):
         pass
 
 
+def write_temp_image_array(img, file=None):
+    if img.shape[0] % 2 != 0:
```
ailzhang, line 35: OOC: why clipping?

tests/python/test_ggui.py
```diff
@@ -51,6 +65,30 @@ def verify_image(window, image_name, tolerance=0.1):
         os.remove(actual_name)
 
 
+def verify_image_array(image, image_name, tolerance=0.1):
+    if image.shape[0] % 2 != 0:
```
ailzhang, line 69: nit: this seems to overlap with `verify_image` a lot, is it possible to avoid duplicating utils functions?

tests/python/test_ggui.py
```diff
@@ -51,6 +65,30 @@ def verify_image(window, image_name, tolerance=0.1):
...
+        ti.tools.imwrite(image, actual_name)
+        ground_truth_np = ti.tools.imread(ground_truth_name)
+        actual_np = ti.tools.imread(actual_name)
+        assert len(ground_truth_np.shape) == len(actual_np.shape)
+        for i in range(len(ground_truth_np.shape)):
+            assert ground_truth_np.shape[i] == actual_np.shape[i]
+        diff = ground_truth_np - actual_np
+        mse = np.mean(diff * diff)
+        assert mse <= tolerance  # the pixel values are 0~255
+        os.remove(actual_name)
```
ailzhang, line 89: try `tempfile.TemporaryFile` which automatically does this for you.

tests/python/test_ggui.py
```diff
@@ -31,6 +31,20 @@ def write_temp_image(window):
         pass
 
 
+def write_temp_image_array(img, file=None):
+    if img.shape[0] % 2 != 0:
+        img = img[:-1]
+    if img.shape[1] % 2 != 0:
+        img = img[:, :-1]
+    if not file:
```
ailzhang, line 39: yup tempfile.TemporaryFile might be easier

tests/python/test_ggui.py
Morcki: just to keep consitant with https://github.com/taichi-dev/taichi/blob/367dc7df80feff6976fe96c3380e255b230639d5/python/taichi/tools/video.py#L116, but clipping is not used here cause we always use even number.

tests/python/test_ggui.py
```diff
@@ -16,39 +16,27 @@
...
-    f = get_temp_png()
-    window.save_image(f)
-    try:
-        os.remove(f)
-    except OSError:
-        pass
-
-
-def verify_image(window, image_name, tolerance=0.1):
+def verify_image(image, image_name, tolerance=0.1):
```
ailzhang, line 19: we can possibly move this to test_utils so that it can be used more widely

ailzhang (approved):
LGTM！

PR #5922, opened 2022-08-30
[llvm] Enhance function is_same_type

lin-hitonami:
Related issue = #5511 
After splitting the modules, both the type names can be `xxx.yyy` where `yyy` are digits.

### DIFF
--- a/taichi/codegen/llvm/llvm_codegen_utils.cpp
+++ b/taichi/codegen/llvm/llvm_codegen_utils.cpp
@@ -15,9 +15,12 @@ std::string type_name(llvm::Type *type) {
  * (a type is a renamed version of the other one) based on the
  * type name. Check recursively if the types are function types.
  *
- * The name of a type imported multiple times is added a suffix starting with a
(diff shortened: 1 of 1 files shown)


taichi/codegen/llvm/llvm_codegen_utils.cpp
```diff
@@ -50,10 +49,54 @@ bool is_same_type(llvm::Type *required, llvm::Type *provided) {
     }
     return true;
   }
-  auto req_name = type_name(required);
-  auto prov_name = type_name(provided);
-  int min_len = std::min(req_name.size(), prov_name.size());
-  return req_name.substr(0, min_len) == prov_name.substr(0, min_len);
+
+  auto a_name = type_name(a);
```
ailzhang, line 53: nit: it'd be nice to add a brief note here (or better place if applicable) explaining how the names are produced and formatted. From the comments below I can tell `xxx.111, xxx, xxx.222` are same types but it'd be nice to have a note so that it's easier to understand ;)

ailzhang (approved):
Thanks!

PR #5927, opened 2022-08-31
[refactor] [llvm] Unify llvm_type() and get_data_type()

strongoier:
Related issue = #3382

These functions are repeating each other. Let's only keep one.


lin-hitonami (approved):
LGTM!

PR #5929, opened 2022-08-31
[llvm] Add attributes to LLVMCompiledData

lin-hitonami:
Related issue = #5511
The attributes are not yet in use now, but they will be used in PRs afterwards.


strongoier (approved):
LGTM

PR #5937, opened 2022-09-01
[llvm] [refactor] Move the generation of struct for function to the context

lin-hitonami:
Related issue = #5511 
The generation of the struct for function happens in the codegen currently, but after the module splitting, it will be delayed to linking. Therefore, I moved this part to the TaichiLLVMContext.


ailzhang (approved):
LGTM!

PR #5948, opened 2022-09-01
[test] [gui] Add a test of drawing lines

Morcki:
Related issue = https://github.com/taichi-dev/taichi/issues/5917#issuecomment-1231429518


tests/python/test_ggui.py
```diff
@@ -387,3 +387,48 @@ def render():
     render()
     verify_image(window.get_image_buffer_as_numpy(), 'test_set_image')
     window.destroy()
+
+
+@pytest.mark.skipif(not _ti_core.GGUI_AVAILABLE, reason="GGUI Not Available")
+@test_utils.test(arch=supported_archs)
+def test_draw_lines():
+    ti.init(arch=ti.vulkan)
```
ailzhang, line 395: this is automatically done in test_utils.test
```suggestion
```

tests/python/test_ggui.py
```diff
@@ -387,3 +387,48 @@ def render():
...
+        scene.lines(points_pos, color=(0.28, 0.68, 0.99), width=5.0)
+        canvas.scene(scene)
+
+    for _ in range(RENDER_REPEAT):
+        render()
+        window.get_image_buffer_as_numpy()
+
+    render()
+    if (platform.system() == 'Darwin'):
+        verify_image(window.get_image_buffer_as_numpy(), 'test_draw_lines.mac')
```
ailzhang, line 431: Can you add a comment explaining why the rendered result differs by platform? Thanks!

ailzhang (approved):
LGTM!

jim19930609 (approved):
Interesting, looks like the behavior is platform-dependent. Is this expected?

Morcki:
> Interesting, looks like the behavior is platform-dependent. Is this expected?

Yes, it's a future work that we should fix it on mac.

tests/python/test_ggui.py
Morcki: Just because of platform difference, but I think it's a bug. I will leave a `TODO` for this.

PR #5945, opened 2022-09-01
[gui] Revert "Less sync overhead for GGUI & Device API Examples"

PENGUINLIONG:
Reverts taichi-dev/taichi#5880

The PR unexpectedly broke GGUI mesh instancing which doesn't have any unit test yet. Reverted for now. A fixed version will be re-introduced later based on the previous PR.

PR #5950, opened 2022-09-01
[vulkan] RHI fixes

bobcao3:
Split from #5880
Ref #5945

This contains all the RHI bug fixes discovered from RHI examples
- API Change
  - Removed `clear_color` command as it is not implemented anywhere
  - Changed source pointer type to a const pointer following C++ common practices
  - Consolidated enum defines into a single file and moved more values from device.h to the header
- New features
  - Ability to create device api Surface with a provided native surface handle
  - Enable validation layer through a bool parameter in device creator
- Bug fixes:
  - Fixed the bug that surface created with the wrong / 0 resolution
  - Fixed the bug that surface can fail to create if specified resolution is larger than supported
  - Fixed the bug that geometry drawing segfaults without an index buffer


Morcki (approved):
LGTM, thanks a lot !

PR #5956, opened 2022-09-02
[aot] [opengl] Add GL interop so that we can export GL memory

ailzhang:
See an example usage in:
https://github.com/ailzhang/taichi-aot-demo/blob/gl_demo/mpm88_gl/mpm88.cpp#L158-L159

Related issue = #

### DIFF
--- a/taichi/rhi/opengl/opengl_device.cpp
+++ b/taichi/rhi/opengl/opengl_device.cpp
@@ -552,6 +552,7 @@ GLint GLDevice::get_devalloc_size(DeviceAllocation handle) {
   glGetBufferParameteriv(GL_ARRAY_BUFFER, GL_BUFFER_SIZE, &size);
   check_opengl_error("glGetBufferParameteriv");
   return size;
+  glBindBuffer(GL_ARRAY_BUFFER, 0);
 }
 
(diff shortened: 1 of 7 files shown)


taichi/rhi/opengl/opengl_device.cpp
```diff
@@ -549,6 +549,15 @@ void GLDevice::dealloc_memory(DeviceAllocation handle) {
   check_opengl_error("glDeleteBuffers");
 }
 
+GLint GLDevice::get_devalloc_size(DeviceAllocation handle) {
+  glBindBuffer(GL_ARRAY_BUFFER, handle.alloc_id);
+  check_opengl_error("glBindBuffer");
+  GLint size = 0;
+  glGetBufferParameteriv(GL_ARRAY_BUFFER, GL_BUFFER_SIZE, &size);
+  check_opengl_error("glGetBufferParameteriv");
+  return size;
```
PENGUINLIONG, line 558: nit: conventionally we call the binding function again with ID=0 to release the binding point.
```suggestion
  return size;
  glBindBuffer(GL_ARRAY_BUFFER, 0);
```

PENGUINLIONG (approved):

PR #5958, opened 2022-09-02
[dx12] Drop code for dxil generation.

python3kgae:
Run llvm coden gen for DirectX backend.
No real dxil will output yet.
Need add pass to transform the llvm module into the shape DirectX backend can consume.

Related issue = #5276


jim19930609 (approved):
Thanks!

PR #5962, opened 2022-09-02
[llvm] [refactor] Link modules instead of cloning modules

lin-hitonami:
Related issue = #5511

### DIFF
--- a/taichi/codegen/codegen.cpp
+++ b/taichi/codegen/codegen.cpp
@@ -87,6 +87,8 @@ void KernelCodeGen::cache_module(const std::string &kernel_key,
 }
 
 std::vector<LLVMCompiledData> KernelCodeGen::compile_kernel_to_module() {
+  auto *llvm_prog = get_llvm_program(prog);
+  auto *tlctx = llvm_prog->get_llvm_context(kernel->arch);
   auto &config = prog->config;
   std::string kernel_key = get_hashed_offline_cache_key(&config, kernel);
   kernel->set_kernel_key_for_cache(kernel_key);
@@ -98,6 +100,7 @@ std::vector<LLVMCompiledData> KernelCodeGen::compile_kernel_to_module() {
       TI_DEBUG("Create kernel '{}' from cache (key='{}')", kernel->get_name(),
                kernel_key);
       cache_module(kernel_key, res);
+      TI_ASSERT(res.size() == 1);
       return res;
     }
   }
@@ -110,17 +113,17 @@ std::vector<LLVMCompiledData> KernelCodeGen::compile_kernel_to_module() {
   TI_ASSERT(block);
 
   auto &offloads = block->statements;
-  std::vector<LLVMCompiledData> data(offloads.size());
+  std::vector<std::unique_ptr<LLVMCompiledData>> data(offloads.size());
   using TaskFunc = int32 (*)(void *);
   std::vector<TaskFunc> task_funcs(offloads.size());
   for (int i = 0; i < offloads.size(); i++) {
     auto compile_func = [&, i] {
+      tlctx->fetch_this_thread_struct_module();
       auto offload =
           irpass::analysis::clone(offloads[i].get(), offloads[i]->get_kernel());
       irpass::re_id(offload.get());
       auto new_data = this->compile_task(nullptr, offload->as<OffloadedStmt>());
-      data[i].tasks = std::move(new_data.tasks);
-      data[i].module = std::move(new_data.module);
+      data[i] = std::make_unique<LLVMCompiledData>(std::move(new_data));
     };
     if (kernel->is_evaluator) {
       compile_func();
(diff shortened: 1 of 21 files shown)


taichi/runtime/llvm/llvm_context.h
```diff
@@ -144,14 +136,15 @@ class TaichiLLVMContext {
...
-  void add_function_to_snode_tree(int id, std::string func);
-
-  void delete_functions_of_snode_tree(int id);
+  void delete_snode_tree(int id);
 
   void add_struct_for_func(llvm::Module *module, int tls_size);
 
   static std::string get_struct_for_func_name(int tls_size);
 
+  std::unique_ptr<LLVMCompiledData> link_compile_data(
```
ailzhang, line 145:
```suggestion
  std::unique_ptr<LLVMCompiledData> link_compiled_data(
```

taichi/runtime/llvm/llvm_context.h
```diff
@@ -43,6 +42,8 @@ class TaichiLLVMContext {
   // main_thread is defined to be the thread that runs the initializer
   JITModule *runtime_jit_module{nullptr};
 
+  std::unique_ptr<ThreadLocalData> link_context_data{nullptr};
```
ailzhang, line 45:
```suggestion
  std::unique_ptr<ThreadLocalData> linked_context_data{nullptr};
```

taichi/codegen/codegen.cpp
```diff
@@ -131,11 +134,15 @@ std::vector<LLVMCompiledData> KernelCodeGen::compile_kernel_to_module() {
   if (!kernel->is_evaluator) {
     worker.flush();
   }
+  auto linked = tlctx->link_compile_data(std::move(data));
+  std::vector<LLVMCompiledData> linked_data;
```
ailzhang, line 138: IMHO a more appropriate name for LLVMCompiledData is actually LLVMCompiledKernel? :P

taichi/runtime/cuda/aot_module_builder_impl.cpp
```diff
@@ -10,8 +10,8 @@ namespace lang {
 namespace cuda {
 
 LLVMCompiledData AotModuleBuilderImpl::compile_kernel(Kernel *kernel) {
-  auto cgen = KernelCodeGenCUDA::make_codegen_llvm(kernel, /*ir=*/nullptr);
-  return cgen->run_compilation();
+  auto cgen = KernelCodeGenCUDA(kernel);
+  return std::move(cgen.compile_kernel_to_module()[0]);
```
ailzhang, line 14: OOC: why do we only return std::move the first one? (do we happen to assume vector.size() == 1? if so we should assert it and probably add a note/comment

taichi/runtime/cpu/aot_module_builder_impl.cpp
```diff
@@ -10,8 +10,8 @@ namespace lang {
 namespace cpu {
 
 LLVMCompiledData AotModuleBuilderImpl::compile_kernel(Kernel *kernel) {
-  auto cgen = KernelCodeGenCPU::make_codegen_llvm(kernel, /*ir=*/nullptr);
-  return cgen->run_compilation();
+  auto cgen = KernelCodeGenCPU(kernel);
+  return std::move(cgen.compile_kernel_to_module()[0]);
```
ailzhang, line 14: ditto

taichi/codegen/wasm/codegen_wasm.h
```diff
@@ -23,6 +23,8 @@ class KernelCodeGenWASM : public KernelCodeGen {
   LLVMCompiledData compile_task(
       std::unique_ptr<llvm::Module> &&module = nullptr,
       OffloadedStmt *stmt = nullptr) override;  // AOT Module Gen
+
+  std::vector<LLVMCompiledData> compile_kernel_to_module() override;
```
ailzhang, line 27: returned value isn't really `module` ;) considering changing to a better name?

ailzhang (approved):
Awesome!!!
Only had a fews nits inline. 
Btw good to test if building with `TI_WITH_LLVM=OFF` still works after this PR locally before merging :P

taichi/codegen/codegen.cpp
lin-hitonami: I will split LLVMCompiledData to `LLVMCompiledKernel` and `LLVMCompiledTask` in the future, because we only need `used_snode_trees` and `` when compiling tasks and linking, and after we link the tasks to a kernel, we don't need them anymore.

taichi/runtime/cuda/aot_module_builder_impl.cpp
lin-hitonami: We assume the size=1, and I will refactor it very soon in the future PRs.

taichi/codegen/wasm/codegen_wasm.h
lin-hitonami: After refactoring in the future PRs, this function will return a `LLVMCompiledKernel` which contains only a module and the names of the offloaded task functions.

taichi/runtime/llvm/llvm_context.h
lin-hitonami: It's a ThreadLocalData containing the context for linking so maybe `linking_context_data` sounds better?

taichi/runtime/llvm/llvm_context.h
lin-hitonami: maybe we can call it `link_compiled_tasks`

ailzhang: sg!

taichi/runtime/llvm/llvm_context.h
ailzhang: or simpler `linking_context` :P yea I'm open to either this or `linking_context_data` if you prefer ;)

PR #5966, opened 2022-09-02
[build] Add option to generate dependency graph of cmake targets

ailzhang:
This is off by default, you can turn it on via
`TAICHI_CMAKE_ARGS="-DTI_WITH_GRAPHVIZ:BOOL=ON" python setup.py
develop`. You can find the generated image in `build/` folder.

Here's a generated graph of our current build system: 
ti_targets

Related issue = #


PGZXB (approved):
LGTM!

PR #5969, opened 2022-09-03
[example] Add RHI examples

bobcao3:
Split from https://github.com/taichi-dev/taichi/pull/5880


cpp_examples/rhi_examples/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+macro(make_sample executable_name src_file)
+add_executable(${executable_name})
+set_property(TARGET ${executable_name} PROPERTY CXX_STANDARD 23)
```
ailzhang, line 3: nit: can we cap this at 17 to avoid accidentally using new features?

cpp_examples/rhi_examples/common.h
```diff
@@ -0,0 +1,129 @@
...
+      evd_params.surface_creator = [&](VkInstance instance) -> VkSurfaceKHR {
+        VkSurfaceKHR surface = VK_NULL_HANDLE;
+
+        if (glfwCreateWindowSurface(instance, glfw_window, nullptr, &surface) !=
+            VK_SUCCESS) {
+          TI_ERROR("failed to create window surface!");
+        }
+        return surface;
+      };
+      evd_params.enable_validation_layer = true;
```
ailzhang, line 81: hmmm currently we'll throw if validation layer is requested but not available, but double check before enabling here, or relax the assert in vulkan_device_creator.cpp?

ailzhang (approved):
Thanks a lot!!

cpp_examples/rhi_examples/common.h
bobcao3: This is fine, this is the RHI example. It doesn't need to work on release environment

PR #5971, opened 2022-09-04
[test] [gui] Skip test of drawing mesh instances on Windows platform

Morcki:
Due to drawing mesh instances report bug on Windows platform, skip the test on windows platform for now.
And fix this bug later.


feisuzhu (approved):
LGTM!

PR #5975, opened 2022-09-05
[gui] Unified GGUI 2D shaders

PENGUINLIONG:
Merged all GGUI 2D shaders into a shared one.

This is the initial step of an in-progress effort to support GGUI rendering as an extension to the C-API. The extension gives access to fundamental rasterized graphics capabilities, which could be meaningful to CI integration and POC writing.

(WIP) Don't merge!


python/taichi/shaders/Canvas2D_vk.frag
```diff
@@ -0,0 +1,28 @@
...
+  float radius;
+} ubo;
+
+void main() {
+  float radius = ubo.radius;
+  vec2 invwh = ubo.wh_invwh.zw;
+
+  // Note that `v_position` is interpolated if the primitive is a line or a
+  // triangle so `gl_FragCoord` always touches `v_position`.
+  float r = length(gl_FragCoord.xy - v_position);
```
bobcao3, line 22: do you really need this radius calculation for all 2D shapes?

bobcao3: better guard it under a if

PENGUINLIONG: I intend to use specialization constants but there will be some extra works to do. I will leave this PR open for a moment. I should mark it as a WIP actually.

PR #5976, opened 2022-09-05
[aot] Temporarily allow Vulkan extensions to be automatically enabled in C-API runtimes

PENGUINLIONG:
This PR temprarily allows extensions to be enabled based on device-capability in the C-API. This workaround should be replaced by a more formal device capability mechanism before the next major release.


ailzhang:
@PENGUINLIONG would you mind opening an issue for extension tracking btw ;) ?

PENGUINLIONG:
> @PENGUINLIONG would you mind opening an issue for extension tracking btw ;) ?

@ailzhang will do

Issue #5959, opened 2022-09-02
[RFC] Implementation plan for MatrixField refactor


strongoier:
Sub-task of #5819. The following image shows the ideal state we would like to achieve:

![MatrixField_Plan](https://user-images.githubusercontent.com/3251060/188072878-8a994eaf-5510-4937-8f96-daaa0ecef090.png)

PR #5980, opened 2022-09-05
[ir] MatrixField refactor 1/n: Make a GlobalVariableExpression solely represent a field

strongoier:
Related issue = #5959

Previously, a `GlobalVariableExpression` serves two purposes:
- representing a field;
- being a wrapper of a (maybe intermediate) SNode so that no matter a `FrontendForStmt` is actually a struct for on a SNode or an external tensor, it can always be constructed by passing an `Expr`.

However, the second purpose doesn't make sense - a struct for on a SNode and a struct for on an external tensor are handled differently everywhere (`lower_ast`, `ir_printer`, ...), so it's natural to split the constructors as well.

This PR clearly and consistently distinguishes different types of `FrontendForStmt`:
- a struct for on a SNode (`snode != nullptr`);
- a struct for on an external tensor (`external_tensor != nullptr`);
- a mesh for (`mesh != nullptr`);
- a range for (other).

As a result, the second purpose is eliminated and a `GlobalVariableExpression` serves solely for a field.

### DIFF
--- a/python/taichi/lang/any_array.py
+++ b/python/taichi/lang/any_array.py
@@ -58,8 +58,6 @@ class AnyArray:
     def _loop_range(self):
         """Gets the corresponding taichi_python.Expr to serve as loop range.
 
-        This is not in use now because struct fors on AnyArrays are not supported yet.
-
         Returns:
             taichi_python.Expr: See above.
         """
--- a/python/taichi/lang/field.py
+++ b/python/taichi/lang/field.py
@@ -87,12 +87,12 @@ class Field:
         return self.vars
 
     def _loop_range(self):
-        """Gets representative field member for loop range info.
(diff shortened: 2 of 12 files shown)


ailzhang (approved):
LGTM!
nit: `external_tensor` isn't the best esp. when it comes to `ndarray`s. But this naming convention is kinda misleading but consistent in the current system lol. Shall we consider renaming them all together in a separate PR in the future?

strongoier:
> LGTM! nit: `external_tensor` isn't the best esp. when it comes to `ndarray`s. But this naming convention is kinda misleading but consistent in the current system lol. Shall we consider renaming them all together in a separate PR in the future?

Sounds good. There are other misleading names like `GlobalPtr`, and we might need to systematically rename them all.

PR #5986, opened 2022-09-06
[doc] Add a note about the offline cache in the print_ir part

lucifer1004:
Add a note about the offline cache in the `print_ir` part to help users.


strongoier (approved):
Thanks!

PR #5891, opened 2022-08-26
[cuda] Clear cuda context after init

turbo0628:
fixes #5502 

Briefing:

Currently, Taichi setup a CUDAContext after `ti.init`. However, this might cause problems with third-party frameworks, including PyTorch and Numba.

This PR pop out the CUDAContext at the end of `ti.init`, and push the context back when needed.

Issue #5502, opened 2022-07-24
Torch 1.12.0 and Taichi cannot use CUDA at the same time.


xuan-li:
**Describe the bug**
If taichi is initialized with GPU, Torch cannot execute backward.

PyTorch version: 1.12.0

**To Reproduce**
```py
import taichi as ti
import torch

device = torch.device("cuda:0")
ti.init(arch=ti.gpu)

x = torch.tensor([1.], requires_grad=True, device=device)
loss = x ** 2
loss.backward()
```

**Log/Screenshots**

```
[Taichi] version 1.0.4, llvm 10.0.0, commit 2827db2c, linux, python 3.9.12
[Taichi] Starting on arch=cuda
Traceback (most recent call last):
  File "******", line 10, in <module>
    loss.backward()
  File "******/lib/python3.9/site-packages/torch/_tensor.py", line 396, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)
  File "******/lib/python3.9/site-packages/torch/autograd/__init__.py", line 173, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: Event device type CUDA does not match blocking stream's device type CPU.
```

**Additional comments**
`ti diagnose`:
```
[Taichi] version 1.0.4, llvm 10.0.0, commit 2827db2c, linux, python 3.9.12

*******************************************
**      Taichi Programming Language      **
*******************************************

Docs:   https://docs.taichi-lang.org/
GitHub: https://github.com/taichi-dev/taichi/
Forum:  https://forum.taichi.graphics/

Taichi system diagnose:

python: 3.9.12 (main, Jun  1 2022, 11:38:51) 
[GCC 7.5.0]
system: linux
executable: /home/xuan/miniconda3/envs/dl/bin/python
platform: Linux-5.13.0-51-generic-x86_64-with-glibc2.31
architecture: 64bit ELF
uname: uname_result(system='Linux', node='Wanzi', release='5.13.0-51-generic', version='#58~20.04.1-Ubuntu SMP Tue Jun 14 11:29:12 UTC 2022', machine='x86_64')
locale: en_US.UTF-8
[...]

PR #5992, opened 2022-09-06
[cuda] Use CUDA primary context to work with PyTorch and Numba.

turbo0628:
Related issue = #5502

PR #5891 solved the problem but the fix is fragile. Using primary context instead of `cuCtxCreate` could also work with PyTorch and Numba. The context management code are no longer needed and reverted.


ailzhang (approved):
Nice!!!

PR #5997, opened 2022-09-06
[refactor] Turn on torch_io tests for opengl, vulkan and dx11 backend

ailzhang:
These io are through torch cpu tensors, not really the most efficient solution but should
be working at least.

Related issue = #


turbo0628 (approved):
LGTM!

PR #5998, opened 2022-09-07
[dx12] Drop code for llvm passes which prepare for DXIL generation.

python3kgae:
2 passes are added for DXIL generation.

TaichiIntrinsicLower will translate taichi intrinsic like thread_idx into the form DirectX backend expected.

TaichiRuntimeContextLower will translate the TaichiRuntimeContext parameter for kernel into Buffers/ConstantBuffers. 
TaichiRuntimeContextLower is empty now.
It is added after inline so optimizations reduce the load/store on temp ptr. And it is easier to know a store is on the TaichiRuntimeContext.

Related issue = #5276


ailzhang (approved):
Thanks!!

PR #6000, opened 2022-09-07
[misc] Lazy load spirv code from disk during offline cache

PGZXB:
Related issue = #4401

ps: The above testing was run in an environment with **100MB+** cache files.
(100MB is the default value of `offline_cache_max_size_of_files`

### DIFF
--- a/taichi/runtime/gfx/aot_module_loader_impl.cpp
+++ b/taichi/runtime/gfx/aot_module_loader_impl.cpp
@@ -24,7 +24,7 @@ class FieldImpl : public aot::Field {
 class AotModuleImpl : public aot::Module {
  public:
   explicit AotModuleImpl(const AotModuleParams &params, Arch device_api_backend)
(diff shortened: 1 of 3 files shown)


taichi/runtime/gfx/aot_module_loader_impl.cpp
```diff
@@ -141,19 +147,40 @@ class AotModuleImpl : public aot::Module {
     return std::make_unique<FieldImpl>(runtime_, field);
   }
 
-  std::vector<uint32_t> read_spv_file(const std::string &output_dir,
-                                      const TaskAttributes &k) {
+  bool try_load_spv_code(std::size_t index) {
```
ailzhang, line 150:
```suggestion
  bool try_load_spv_kernel(std::size_t index) {
```

ailzhang (approved):
LGTM!

PR #6005, opened 2022-09-07
[lang] Fixed compilation

PENGUINLIONG:
`KernelMetadata` is recognized by NDK compiler as a nested class in `Metadata`.


PGZXB (approved):
LGTM, Thanks!

PR #6007, opened 2022-09-07
[autodiff] Avoid initializing Field with None

strongoier:
When `x_grad` or `x_dual` is `None`, they should not be used to initialize the `Field` class.


erizmr (approved):
LGTM. Thanks!

PR #6010, opened 2022-09-07
[ir] MatrixField refactor 3/n: Add MatrixFieldExpression

strongoier:
Related issue = #5959

This PR adds definition of `MatrixFieldExpression`, which serves as the representation of a matrix field in C++. Currently only `dynamic_index_stride` has been completely moved there. Others will get migrated in future PRs.


taichi/ir/frontend_ir.h
```diff
@@ -521,6 +521,24 @@ class FieldExpression : public Expression {
   TI_DEFINE_ACCEPT_FOR_EXPRESSION
 };
 
+class MatrixFieldExpression : public Expression {
+ public:
+  std::vector<Expr> fields;
+  std::vector<int> element_shape;
```
jim19930609, line 527: I was thinking we can infer `element_shape` from `ret_type`, whereas `ret_type` for MatrixFieldExpression should have already been set upon construction?

jim19930609 (approved):
LGTM!

taichi/ir/frontend_ir.h
strongoier: This part is a bit tricky - for some use cases (like quant), the fields may have different dtypes, so the result type cannot be simply represented as `TensorType`. I plan to scalarize such tricky cases during `lower_ast`. After `lower_ast` things will work as we discussed.

AD1024 (approved):
LGTM :D

Issue #5915, opened 2022-08-30
pow() with a negative base fails on non-LLVM backends


strongoier:
When `arch=ti.cpu/ti.cuda`, the following script,
```python
import taichi as ti

ti.init()

@ti.kernel
def calc_pow(i: ti.i32) -> ti.f32:
    return pow(-1.0, i)

print(calc_pow(1))
```
gives `-1.0` , which is correct. However, it gives `1.0` when `arch=ti.metal`, and gives `nan` when `arch=ti.opengl`.

This is because the native `pow(x, y)` function in Metal Shading Language/GLSL is implemented with `exp2(y * log2(x))`, so it doesn't support negative bases. See https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf and https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.50.pdf for more details.

The behavior is understandable, because `pow()` with a negative base and a real exponent results in a complex number. What we can do is to follow C++ - specially handle the case where the exponent is an integer, which results in a real number. In this way we can make the behavior of `pow()` consistent across all backends. I'll submit a PR for that.

PR #6018, opened 2022-09-08
[Lang] Support abs(i64)

strongoier:
I found that `abs(i64)` was not supported while fixing #5915. This PR adds support for that.


ailzhang (approved):
LGTM!

lin-hitonami (approved):
LGTM!

PR #6020, opened 2022-09-08
[Bug] Disable vkCmdWriteTimestamp with MacOS to enable tests on Vulkan

### LINKED ISSUE #5888: Investigate vulkan support on macos 10.15
Currently vulkan backend is not working properly on macos 10.15 (with molten-vk installed) so we actually skipped testing ti.vulkan on macos 10.15. We need to investigate the root cause there.

jim19930609:
Related issue = fix #5888


ailzhang (commented):
NIce!! would you mind also enabling macOS 15 Vulkan build and test along with this PR so that we have some tests? (feel free to turn off the ones are still broken ;) https://github.com/taichi-dev/taichi/blob/master/.github/workflows/testing.yml#L222

jim19930609:
Looks like most Vulkan tests are passed, but failed a few ones with precision issue: 

image

Any idea how we can blacklist these tests on MacOS 10.15 with x86_64 arch, or shall I implement a blacklist? @ailzhang @feisuzhu

PENGUINLIONG:
Feel free to disable timestamps on macOS entirely. I found it not working with M1 also.

PENGUINLIONG (approved):
lgtm

ailzhang (approved):
Thanks!!

PR #6028, opened 2022-09-10
[bug] [gui] Fix a bug of drawing mesh instacing that cpu/cuda objects have an offset when copying to vulkan object

Morcki:
This only happend on Windows.


taichi/rhi/interop/vulkan_cpu_interop.cpp
```diff
@@ -15,6 +15,21 @@ namespace lang {
...
+  CpuDevice *cpu_dev = dynamic_cast<CpuDevice *>(src.device);
+
+  DeviceAllocation src_alloc(src);
+
+  CpuDevice::AllocInfo src_alloc_info = cpu_dev->get_alloc_info(src_alloc);
+
+  unsigned char *dst_ptr = (unsigned char *)(vk_dev->map_range(dst, size));
+  unsigned char *src_ptr = (unsigned char *)src_alloc_info.ptr + src.offset;
+
+  memcpy(dst_ptr, src_ptr, size);
```
bobcao3, line 29: Not the best idea... Writing to `dst` here might be very slow because it's on-device memory, or it would be slow for GPU to use. The best solution is to allocate a vulkan buffer that is on the CPU and use that as staging buffer

taichi/ui/backends/vulkan/renderables/mesh.cpp
```diff
@@ -88,21 +88,8 @@ void Mesh::update_data(const MeshInfo &info, const Scene &scene) {
       attr_dev_ptr =
           get_device_ptr(prog, info.mesh_attribute_info.mesh_attribute.snode);
     }
-    Device::MemcpyCapability memcpy_cap = Device::check_memcpy_capability(
```
bobcao3, line 91: Why don't we check capability anymore?

taichi/rhi/interop/vulkan_cpu_interop.cpp
Morcki: Yes, I know that. However here I just want a function to directly passing cpu data to storage buffer which belongs to transfer family and don't need staging buffer cause it is staging buffer. Maybe I should add a annotation that here the `dst` must a staging buffer (host visible buffer) instead of device buffer.

taichi/ui/backends/vulkan/renderables/mesh.cpp
Morcki: Because if we add the function `memcpy_cpu_to_vulkan`, then the `memcpy_direct` includes all kinds of situations as below:
1. copy from cpu data to vulkan staging buffer
2. copy from cuda data to vulkan staging buffer
3. copy from vulkan data to vulkan buffer
4. others will report `ti_not_implemented`

bobcao3: what about situations like `opengl` or `metal`, which should go from opengl->cpu->vulkan before we implement the interop

taichi/rhi/interop/vulkan_cpu_interop.cpp
bobcao3: This is not the best API... If `memcpy_direct` requires `dst` to be of specific attribute, it needs proper checking & validation

Morcki: Ok, so what is the best way of copying from cpu data to the ssbo? And I don't see checking and validation in other place. What can I do to cover this?

taichi/ui/backends/vulkan/renderables/mesh.cpp
Morcki: Yes, I think that's the 4th situation, GGUI doesn't handle opengl and metal for now, and we report an error info. Maybe in future, we should implement that. By the way, I think `check_memcpy_capability` is superfluous, because all other copy functions (like `memcpy via staging` and `memcpy_direct`) have the same checkings of what the `src` and `dst` belong to. Why should we check that twice?

Morcki: Ok, I forgot that, that's right, but for now, we don't support `opengl` and `metal` even using memcpy_via_staging, which is a future work I think? 
By the way I think `check_memcpy_capability` is really confused, because we have to check the `src` and `dst` device twice. Why we should check with dynamic_cast instead of tagging the device backend?

bobcao3: How can we tag the backend?

Morcki: Maybe add an enum class type in child class I think?

bobcao3: I think dynamic cast is not really slower than that?

taichi-gardener:

Morcki:
> /rebase

May I merge this directly and solve the conflicts? It seems that rebase is a bit troublesome.

ailzhang:
@Morcki you can resolve merge conflict locally and the force push into this branch.

bobcao3 (approved):
LGTM

jim19930609 (approved):

PR #6030, opened 2022-09-10
[Lang] Deprecate SOA-layout for NdarrayMatrix/NdarrayVector

### LINKED ISSUE #5849: Delete SOA layout for vector/matrix ndarrays
Talked to @PENGUINLIONG and @YuCrazing that we don't seem to have a use case for this yet, but considering this layout is kinda an overhead implementation wise. Let's delete it for now and this can be added back pretty easily in the future.

jim19930609:
Related issue = fix #5849

After this PR, Ndarray support only one memory layout (ti.AOS), and user will be no longer able to specify layout explicitly.


ailzhang (approved):
Nice! Thanks!
note: although we haven't officially released ndarray, let's still highlight this in the next release note (and in the ndarray doc in the future) so that users know ndarray only supports AOS layout.

PR #2515, opened 2021-07-12
[misc] Output "taichi_cpp_tests.exe" to the "bin" folder on Windows

xumingkuan:
Related issue = #2502

By default, on Windows, Taichi outputs `taichi_cpp_tests.exe` in somewhere like `build\RelWithDebInfo\`, which is hard to find.

This PR moves it to `bin\`, which is in the `PATH` and we can run `taichi_cpp_tests` everywhere.

PR #6038, opened 2022-09-12
[bug] [test] Run C-API tests correctly on Windows

PGZXB:
Related issue = #

Output all exes to build/

> This PR moves it to bin\, which is in the PATH and we can run taichi_cpp_tests everywhere.
>    --from https://github.com/taichi-dev/taichi/pull/2515#issue-941918398

IMO, we should run cpp/capi tests by running `python run_tests.py --cpp` instead of launching `taichi_cpp_tests(.exe)` manually

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -63,26 +63,30 @@ def _test_cpp_aot(test_filename, build_dir, test_info):
 
 def _test_cpp():
     curr_dir = os.path.dirname(os.path.abspath(__file__))
-    if platform.system() == "Windows":
-        cpp_test_filename = 'taichi_cpp_tests.exe'
-        capi_test_filename = 'taichi_c_api_tests.exe'
-        build_dir = os.path.join(curr_dir, '../bin')
+    cpp_test_filename = 'taichi_cpp_tests.exe'
+    capi_test_filename = 'taichi_c_api_tests.exe'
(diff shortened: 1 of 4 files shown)


tests/run_tests.py
```diff
@@ -63,26 +63,30 @@ def _test_cpp_aot(test_filename, build_dir, test_info):
 
 def _test_cpp():
     curr_dir = os.path.dirname(os.path.abspath(__file__))
-    if platform.system() == "Windows":
-        cpp_test_filename = 'taichi_cpp_tests.exe'
-        capi_test_filename = 'taichi_c_api_tests.exe'
-        build_dir = os.path.join(curr_dir, '../bin')
+    cpp_test_filename = 'taichi_cpp_tests.exe'
```
ailzhang, line 66: Did we accidentally skip c api tests on *nix machines by doing this? log from the linux gpu job ;) 
```
2022-09-12T09:28:20.5094707Z Not found /home/dev/tests/../build/taichi_c_api_tests.exe, skipping...
2022-09-12T09:28:20.5095195Z Not found /home/dev/tests/../build/taichi_cpp_tests.exe, skipping...
```

ailzhang (changes requested):
Nice work! Let's make sure it's not accidentally disabled before merging ;)

ailzhang (approved):
Thanks!

PR #6044, opened 2022-09-13
[Lang] Demote pow() with integer exponent

strongoier:
Related issue = #5915

Just as proposed in #5915, we specially handle `pow()` with an integer exponent by demoting it with manual fast-pow implementation in the `demote_operations` pass so that `pow()` can have the same behavior across backends. In this case the return type is the type of base (lhs).


taichi/ir/frontend_ir.cpp
```diff
@@ -192,7 +192,8 @@ void BinaryOpExpression::type_check(CompileConfig *config) {
     ret_type = PrimitiveType::i32;
     return;
   }
-  if (is_shift_op(type)) {
+  if (is_shift_op(type) ||
```
ailzhang, line 195: aha we may gradually add a table in docs/spec to show the expected type of different ops.

ailzhang (approved):
Nice work!

taichi/ir/frontend_ir.cpp
strongoier: Yep. It is currently described here: https://docs.taichi-lang.org/docs/type#implicit-type-casting-in-binary-operations
For `pow()` I think I may need to add something in the API reference, which will be done in a follow-up PR.

PR #6046, opened 2022-09-13
[Doc] Update docstring of pow()

strongoier:
Related issue = #5915

This PR resolves https://github.com/taichi-dev/taichi/pull/6044#discussion_r969192450.

### DIFF
--- a/python/taichi/lang/ops.py
+++ b/python/taichi/lang/ops.py
@@ -732,10 +732,18 @@ def mod(x1, x2):
 def pow(x, a):  # pylint: disable=W0622
     """First array elements raised to powers from second array :math:`x^a`, element-wise.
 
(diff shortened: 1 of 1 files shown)


python/taichi/lang/ops.py
```diff
@@ -732,10 +732,18 @@ def expr_python_mod(a, b):
 def pow(x, a):  # pylint: disable=W0622
```
ailzhang, line 732: Since we are here, would you mind updating this to `base` and `exp` so that they're more self-explanatory? :P

ailzhang (approved):
Awesome thanks!

python/taichi/lang/ops.py
strongoier: Updated.

PR #6048, opened 2022-09-13
[lang] Remove redundant codegen of integer pow

strongoier:
Related issue = #5915

As `pow()` with integer exponent is already demoted (#6044), there's no need for backends to handle integer pows.

### DIFF
--- a/taichi/codegen/cuda/codegen_cuda.cpp
+++ b/taichi/codegen/cuda/codegen_cuda.cpp
@@ -667,10 +667,6 @@ class TaskCodeGenCUDA : public TaskCodeGenLLVM {
         llvm_val[stmt] = create_call("__nv_powf", {lhs, rhs});
       } else if (ret_type->is_primitive(PrimitiveTypeID::f64)) {
         llvm_val[stmt] = create_call("__nv_pow", {lhs, rhs});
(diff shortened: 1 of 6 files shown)


taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -667,10 +667,6 @@ class TaskCodeGenCUDA : public TaskCodeGenLLVM {
         llvm_val[stmt] = create_call("__nv_powf", {lhs, rhs});
       } else if (ret_type->is_primitive(PrimitiveTypeID::f64)) {
         llvm_val[stmt] = create_call("__nv_pow", {lhs, rhs});
-      } else if (ret_type->is_primitive(PrimitiveTypeID::i32)) {
```
ailzhang, line 670: nit: can we add a comment here to redirect readers to the demote pass? :P I'm afraid people might accidentally miss that and add these back in the future lol

ailzhang (approved):
Thanks!

taichi/codegen/cuda/codegen_cuda.cpp
strongoier: SG!

AD1024 (approved):
LGTM!

PR #6050, opened 2022-09-13
[Lang] Sort coo to build correct csr format sparse matrix on GPU

FantasyVR:
Related issue = #2906 

When building a coo format sparse matrix, the indices are not in order. To build a valid csr format sparse matrix. We need to first sort the coo indices arrays.


taichi/program/sparse_matrix.cpp
```diff
@@ -216,9 +245,14 @@ void CuSparseMatrix::build_csr_from_coo(void *coo_row_ptr,
       &matrix_, rows_, cols_, nnz, csr_row_offset_ptr, coo_col_ptr,
       coo_values_ptr, CUSPARSE_INDEX_32I, CUSPARSE_INDEX_32I,
       CUSPARSE_INDEX_BASE_ZERO, CUDA_R_32F);
+  CUSPARSEDriver::get_instance().cpDestroySpVec(vec_permutation);
+  CUSPARSEDriver::get_instance().cpDestroyDnVec(vec_values);
   CUSPARSEDriver::get_instance().cpDestroy(cusparse_handle);
   // TODO: free csr_row_offset_ptr
   // CUDADriver::get_instance().mem_free(csr_row_offset_ptr);
```
Hanke98, line 252: Since we might need the `csr_row_offset_ptr` later, e.g., print the matrix, is it better to remove these two lines?

taichi/program/sparse_matrix.cpp
```diff
@@ -203,11 +203,40 @@ void CuSparseMatrix::build_csr_from_coo(void *coo_row_ptr,
...
+  // Setup permutation vector to identity
+  CUSPARSEDriver::get_instance().cpCreateIdentityPermutation(
+      cusparse_handle, nnz, d_permutation);
+  CUSPARSEDriver::get_instance().cpXcoosortByRow(cusparse_handle, rows_, cols_,
+                                                 nnz, coo_row_ptr, coo_col_ptr,
+                                                 d_permutation, dbuffer);
+  CUSPARSEDriver::get_instance().cpGather(cusparse_handle, vec_values,
+                                          vec_permutation);
+  CUDADriver::get_instance().memcpy_device_to_device(
+      coo_values_ptr, d_values_sorted, nnz * sizeof(float));
```
Hanke98, line 235: Is it ok to directly use `d_value_sorted` later to avoid this copy?

Hanke98 (commented):
Great Job! I only have two little questions:

taichi/program/sparse_matrix.cpp
FantasyVR: Yeah, it makes sense. We should also remember to free the space somewhere.

strongoier (approved):
LGTM!

taichi/program/sparse_matrix.cpp
FantasyVR: It's ok if the coo format ndarrays anymore. But the users may need to reuse the coo values ndarray which is not the right order anymore.

Hanke98: Oh, I see! Thanks.

PR #6051, opened 2022-09-13
[misc] Enable offline cache in frontend instead of C++ Side

PGZXB:
Related issue = #4401 

Enabling offline cache in C++ Side (setting `CompileConfig::offline_cache` as `true` by default) is dangerous. The offline cache will always be enabled when running taichi_cpp_tests because the value of compile config in C++ side is not affected by env variables, which is not expected and can cause some strange bugs because it will unexpectedly load cache files that were not expected to generate.


ailzhang (approved):
Thanks!

PR #6053, opened 2022-09-14
[lang] Reformat source indicator in Python convention

PENGUINLIONG:
Follow the Python stacktrace convention to support click-jump to the exact line of compiler error in VS Code.
Image


lin-hitonami (approved):
LGTM!

PR #6056, opened 2022-09-14
[aot] Dump required device capability in AOT module meta

PENGUINLIONG:
To enable device capability checks during module loads so that the user won't crash in graph or kernel execution, this PR introduces a map of device capabilities assumed by the compiler, so the C-API can lookup module compatibility with the created runtime in a future PR.


ailzhang (approved):
LGTM!

PR #6060, opened 2022-09-14
[aot] Guard C-API interfaces with try-catch

PENGUINLIONG:
To keep the C-API users from unexpected exceptions, the interface implementations are guarded with try-catches.


ailzhang (approved):
Thanks!

PR #6063, opened 2022-09-15
[misc] Simplify PR template

ailzhang:
Now that we use PR title & description as commit message let's simplify this template a bit to avoid polluting commit history.

Related issue = #


feisuzhu (approved):
LGTM!

PR #6074, opened 2022-09-15
[Lang] MatrixField refactor 4/n: Disallow invalid matrix field definition

strongoier:
Related issue = #5959, #4857

Support for different element types of matrix fields was introduced in #2135 for quant. As discussed in https://github.com/taichi-dev/taichi/issues/4857#issuecomment-1123964395, the only case we need to support is different element types with **same compute type**. This PR adds the validity check and removes test cases which are actually not allowed.

### DIFF
--- a/taichi/ir/frontend_ir.h
+++ b/taichi/ir/frontend_ir.h
@@ -534,10 +534,13 @@ class MatrixFieldExpression : public Expression {
     for (auto &field : fields) {
       TI_ASSERT(field.is<FieldExpression>());
     }
-    auto compute_type = fields[0].cast<FieldExpression>()->dt->get_compute_type();
+    auto compute_type =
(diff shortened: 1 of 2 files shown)


taichi/ir/frontend_ir.h
```diff
@@ -531,6 +531,18 @@ class MatrixFieldExpression : public Expression {
   MatrixFieldExpression(const std::vector<Expr> &fields,
                         const std::vector<int> &element_shape)
       : fields(fields), element_shape(element_shape) {
+    for (auto &field : fields) {
+      TI_ASSERT(field.is<FieldExpression>());
+    }
+    auto compute_type =
```
jim19930609, line 537: Assert `fields` not empty before accessing `fields[0]`

jim19930609 (approved):

PR #6075, opened 2022-09-15
[vulkan] Support printing in debug mode on vulkan backend

### LINKED ISSUE #2713: [vulkan] Support print() for Vulkan backend

Currently, the Vulkan backend does not support the print function yet. `print()` will help users debug their codes more easily.

Implementing print is adding an IR visitor for the `PrintStmt` statement, and you also need to create a buffer to store print messages and communicate between the host and the device. The OpenGL backend gives an excellent example to follow with. : )

ailzhang:
Fixes #2713 

This is a long waited debugging utils that we finally get it done. Note since debug printing requires validation layer support, you'll need to have Vulkan SDK installed on your system to use it properly. For example https://vulkan.lunarg.com/doc/sdk/1.3.224.1/linux/getting_started_ubuntu.html. 
Also to enable debug printing on vulkan backend you'll need to make sure `ti.init(..., debug=True)` so that we enable the validation layer properly. 

Nit: our test_print.py doesn't really validate the printed message but I've run them with `-v -s` to verify the outputs visually. 

Note: validation layer is not available on molten-vk so vulkan printing on macos won't work. 

TODO: 
- support on opengl backend.


PENGUINLIONG (approved):
lgtm

PR #6077, opened 2022-09-15
[Lang] Matrix/Vector refactor: support basic matrix ops

AD1024:
Related issue = #5478 
Refactored & Combined PR: #5797 and #5861 
A part of PR #5551


taichi/ir/frontend_ir.cpp
```diff
@@ -173,6 +173,29 @@ void UnaryOpExpression::flatten(FlattenContext *ctx) {
...
+Expr to_broadcast_tensor(const Expr &elt, const DataType &dt) {
+  TI_ASSERT(dt->is<TensorType>());
+  auto tensor_type = dt->as<TensorType>();
+  auto elt_type = tensor_type->get_element_type();
+  TI_ASSERT_INFO(elt_type->is<PrimitiveType>(),
+                 "Only primitive types are supported in Tensors, got {}",
+                 elt_type->to_string());
+  std::vector<Expr> broadcast_values(tensor_type->get_num_elements(), elt);
+  return Expr::make<MatrixExpression>(broadcast_values,
+                                      tensor_type->get_shape(), elt_type);
```
strongoier, line 185:
```suggestion
                                      tensor_type->get_shape(), elt->ret_type);
```
Broadcast should not change the element type?

taichi/ir/frontend_ir.cpp
```diff
@@ -192,17 +215,53 @@ void BinaryOpExpression::type_check(CompileConfig *config) {
 
   if ((lhs_type->is<PrimitiveType>() && rhs_type->is<TensorType>()) ||
       (lhs_type->is<TensorType>() && rhs_type->is<PrimitiveType>())) {
-    TI_NOT_IMPLEMENTED;
+    // convert Tensor/Scalar | Scalar/Tensor operations to broadcasting
+    auto [unified_l, unified_r] = unify_binop_operands(lhs, rhs);
+    lhs = unified_l;
+    rhs = unified_r;
+    if (lhs->ret_type == PrimitiveType::unknown)
```
strongoier, line 222: When does this happen?

taichi/transforms/type_check.cpp
```diff
@@ -277,20 +277,30 @@ class TypeCheck : public IRVisitor {
         stmt->rhs->ret_type->is_primitive(PrimitiveTypeID::unknown))
       error();
     if (stmt->op_type == BinaryOpType::pow &&
-        is_integral(stmt->rhs->ret_type)) {
+        (is_integral(stmt->rhs->ret_type) ||
+         is_integral(stmt->rhs->ret_type))) {
```
strongoier, line 281:
```suggestion
         is_integral_tensor(stmt->rhs->ret_type))) {
```

taichi/transforms/type_check.cpp
```diff
@@ -277,20 +277,30 @@ class TypeCheck : public IRVisitor {
...
       auto default_fp = config_.default_fp;
-      if (!is_real(stmt->lhs->ret_type)) {
-        cast(stmt->lhs, default_fp);
+      if (!is_real(stmt->lhs->ret_type) ||
+          !is_real_tensor(stmt->lhs->ret_type)) {
+        cast(stmt->lhs, make_dt(default_fp));
       }
       if (!is_real(stmt->rhs->ret_type)) {
-        cast(stmt->rhs, default_fp);
+        cast(stmt->rhs, make_dt(default_fp));
```
strongoier, line 303: `is_real_tensor()` should also be checked?

strongoier (commented):
1. Shall we add tests for the modifications?
2. Why both `UnaryOp` and `BinaryOp` are modified in codegen, but only `BinaryOp` is modified in frontend_ir?
3. Shall we have `is_real_scalar()`, `is_real_tensor()`, and a general `is_real()` so that we don't have to use `is_real() || is_real_tensor()` everywhere?

taichi/ir/frontend_ir.cpp
```diff
@@ -173,6 +173,29 @@ void UnaryOpExpression::flatten(FlattenContext *ctx) {
   ctx->push_back(std::move(unary));
 }
 
+Expr to_broadcast_tensor(const Expr &elt, const DataType &dt) {
+  TI_ASSERT(dt->is<TensorType>());
+  auto tensor_type = dt->as<TensorType>();
+  auto elt_type = tensor_type->get_element_type();
```
jim19930609, line 179: Check if `elt` has the same TensorType as `dt`, then we should directly return `elt` here. Otherwise an error should be thrown.

taichi/ir/frontend_ir.cpp
```diff
@@ -173,6 +173,29 @@ void UnaryOpExpression::flatten(FlattenContext *ctx) {
...
+  TI_ASSERT_INFO(elt_type->is<PrimitiveType>(),
+                 "Only primitive types are supported in Tensors, got {}",
+                 elt_type->to_string());
+  std::vector<Expr> broadcast_values(tensor_type->get_num_elements(), elt);
+  return Expr::make<MatrixExpression>(broadcast_values,
+                                      tensor_type->get_shape(), elt_type);
+}
+
+std::tuple<Expr, Expr> unify_binop_operands(const Expr &e1, const Expr &e2) {
+  if ((!e1->ret_type->is<TensorType>() && !e2->ret_type->is<TensorType>()) ||
```
jim19930609, line 189: Let's write `e1->ret_type->is<PrimitiveType>() && e2->ret_type->is<PrimitiveType>()` instead, since BinaryExpression only handles operands with TensorType and PrimitiveType for now.

taichi/ir/frontend_ir.cpp
```diff
@@ -173,6 +173,29 @@ void UnaryOpExpression::flatten(FlattenContext *ctx) {
...
+      (e1->ret_type->is<TensorType>() && e2->ret_type->is<TensorType>())) {
+    return std::tuple(e1, e2);
+  }
+  if (!e1->ret_type->is<TensorType>()) {
```
jim19930609, line 193: ditto. `!e1->ret_type->is<TensorType>()` does not guarantee it's PrimitiveType. It could be unexpected types suchas PointerType or some other types added in the future. In that case we better throw an eror.

taichi/ir/frontend_ir.cpp
```diff
@@ -192,17 +215,53 @@ void BinaryOpExpression::type_check(CompileConfig *config) {
...
-    TI_NOT_IMPLEMENTED;
+    // convert Tensor/Scalar | Scalar/Tensor operations to broadcasting
+    auto [unified_l, unified_r] = unify_binop_operands(lhs, rhs);
+    lhs = unified_l;
+    rhs = unified_r;
+    if (lhs->ret_type == PrimitiveType::unknown)
+      lhs.type_check(config);
+    if (rhs->ret_type == PrimitiveType::unknown)
+      rhs.type_check(config);
+    TI_ASSERT(lhs->ret_type->is<TensorType>());
```
jim19930609, line 226: if `lhs_type->is<TensorType>() && rhs_type->is<PrimitiveType>()`, how come they get unified to TensorType?

jim19930609 (commented):
You can probably separate out the `codegen` part, and add some c++ tests to test the **type promotion** and **broadcasting**. 

Some examples can be found in: `tests/cpp/ir/ir_type_promotion_test.cpp` and `tests/cpp/ir/frontend_type_inference_test.cpp`

taichi/ir/frontend_ir.cpp
AD1024: Unification here means converting to broadcast operations, so the `rhs` will be converted to a tensor of the previous `rhs` value that matches `lhs`'s shape.

taichi/ir/frontend_ir.cpp
AD1024: I suppose this would happen if `lhs`/`rhs` is freshly constructed expressions (via broadcasting conversion)

AD1024:
> 1. Shall we add tests for the modifications?
> 2. Why both `UnaryOp` and `BinaryOp` are modified in codegen, but only `BinaryOp` is modified in frontend_ir?
> 3. Shall we have `is_real_scalar()`, `is_real_tensor()`, and a general `is_real()` so that we don't have to use `is_real() || is_real_tensor()` everywhere?

1. Sure! will add some tests to cover c++ side changes.
2. The reason is we are only covering elementwise operations now at codegen. LLVM has corresponding variant of the common binary ops (+ - * /, etc.)  for vector types so we don't need to change to. For UnaryOps, the major change is on casting because we need to cast every individual elements in a VectorType if a tensor is being cast to another tensor type. 
3. In the first version of the implementation I put them together but this change definitely caused some issue in other parts of the codebase. I tend to leave them as separate implementations.

taichi/ir/frontend_ir.cpp
jim19930609: Oh I misunderstood this part, thanks!

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -448,46 +554,93 @@ void TaskCodeGenLLVM::visit(UnaryOpStmt *stmt) {
...
       llvm_val[stmt] =
           builder->CreateMul(llvm_val[stmt->lhs], llvm_val[stmt->rhs]);
     }
   } else if (op == BinaryOpType::floordiv) {
-    if (is_integral(ret_type))
-      llvm_val[stmt] =
-          create_call(fmt::format("floordiv_{}", data_type_name(ret_type)),
-                      {llvm_val[stmt->lhs], llvm_val[stmt->rhs]});
-    else {
+    if (is_integral_tensor(ret_type) || is_integral(ret_type)) {
```
strongoier, line 607: `floordiv` codegen has been removed in #6135. You need to modify https://github.com/taichi-dev/taichi/blob/79f2a6a9d692d81a6a1b46d7391d58ecf5aed771/taichi/transforms/demote_operations.cpp#L41.

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -355,6 +355,57 @@ TaskCodeGenLLVM::TaskCodeGenLLVM(Kernel *kernel,
 void TaskCodeGenLLVM::visit(DecorationStmt *stmt) {
 }
 
+void TaskCodeGenLLVM::create_value_cast(
```
strongoier, line 358: Can `create_value_cast` and `create_fp_trunc` be unified to something like `create_elementwise_unary`?

python/taichi/lang/expr.py
```diff
@@ -39,6 +41,20 @@ def __init__(self, *args, tb=None, dtype=None):
             self.ptr.set_tb(self.tb)
         self.ptr.type_check(impl.get_runtime().prog.config)
 
+    def __getitem__(self, indices):
```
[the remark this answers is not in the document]
AD1024: When calling matrix-matrix operations to build the program at Python frontend, we will use this to explicitly construct the indexing expression.

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -369,47 +420,102 @@ void TaskCodeGenLLVM::visit(UnaryOpStmt *stmt) {
...
+          return this->builder->CreateFPTrunc(value, type);
+        };
+        create_fp_trunc(stmt, trunc_func, llvm::Type::getHalfTy(*llvm_context),
+                        cast_type->is<TensorType>(), /*trunc_self=*/true);
       }
-    } else if (is_real(from) && is_real(to)) {
-      if (data_type_size(from) < data_type_size(to)) {
-        llvm_val[stmt] = builder->CreateFPExt(
-            llvm_val[stmt->operand], tlctx->get_data_type(stmt->cast_type));
+    } else if ((is_real(from) || is_real_tensor(from)) &&
```
jim19930609, line 469: You can simply check for `is_real(from.get_element_type())`, cause seriously there's no special treatment for `TensorType` here.

Same for most of the checks in `visit(BinaryOpStmt*)`

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -369,47 +420,102 @@ void TaskCodeGenLLVM::visit(UnaryOpStmt *stmt) {
...
     if (from == to) {
       llvm_val[stmt] = llvm_val[stmt->operand];
-    } else if (is_real(from) != is_real(to)) {
-      if (is_real(from) && is_integral(to)) {
-        cast_op = is_signed(to) ? llvm::Instruction::CastOps::FPToSI
-                                : llvm::Instruction::CastOps::FPToUI;
-      } else if (is_integral(from) && is_real(to)) {
-        cast_op = is_signed(from) ? llvm::Instruction::CastOps::SIToFP
-                                  : llvm::Instruction::CastOps::UIToFP;
+    } else if (is_real(from) != is_real(to) ||
```
jim19930609, line 429: DataType has this method of get_element_type() which returns the primitive dtype.

You can simply check for is_real(from.get_element_type()), cause the branch seems agnostic of whether it's TensorType or not.

Same for most of the checks in visit(BinaryOpStmt*)

jim19930609 (commented):
In `codegen_llvm.cpp`, for those LLVM operations that supports both Primitive and TensorType operands, let's simplify the condition from `is_real_tensor(stmt->ret_type) || is_real(ret_type)` to `is_real(ret_type.get_element_type())`?

taichi/codegen/llvm/codegen_llvm.cpp
AD1024: Sure. We can have something like `elementwise_cast` (since we'll need to know the type casting to).

python/taichi/lang/expr.py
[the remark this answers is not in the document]
strongoier: Hmm I thought it was handled here? https://github.com/taichi-dev/taichi/blob/3677d934e3863fb74a7236dc3512ce2c6328b0fc/python/taichi/lang/ast/ast_transformer.py#L241-L250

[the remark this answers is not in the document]
AD1024: This is the case that it's handled by the ASTTransformer, but calling a library function won't be handled (because it's not a kernel, and it inlines the operations).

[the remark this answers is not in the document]
strongoier: 1. Is it possible to make all library functions `ti.func`?
2. Are the changes in this file unrelated to this PR? If so shall we cancel them first?

taichi/ir/type_utils.h
```diff
@@ -185,6 +185,19 @@ inline TypedConstant get_min_value(DataType dt) {
   }
 }
 
+#define DEFINE_TENSOR_CHECK(func_name)                   \
```
strongoier, line 188: Is this still needed?

strongoier (approved):

PR #6080, opened 2022-09-16
[Lang] MatrixNdarray refactor part7: Add scalarization for UnaryOpStmt with TensorType-operand

jim19930609:
Related issue = https://github.com/taichi-dev/taichi/issues/5873, https://github.com/taichi-dev/taichi/issues/5819

This PR is working "Part ④" in https://github.com/taichi-dev/taichi/issues/5873.


taichi/transforms/scalarize.cpp
```diff
@@ -45,9 +45,12 @@ class Scalarize : public IRVisitor {
       // Needs scalarize
       auto dest_tensor_type = dest_dtype->template as<TensorType>();
       auto val_tensor_type = val_dtype->template as<TensorType>();
+
       TI_ASSERT(dest_tensor_type->get_shape() == val_tensor_type->get_shape());
-      TI_ASSERT(dest_tensor_type->get_element_type() ==
-                val_tensor_type->get_element_type());
+      // For sqrt/exp/log with int-type operand, we automatically set the
```
strongoier, line 50: IMO this is normal and it is handled by https://github.com/taichi-dev/taichi/blob/18c946ea67d23265896d3fb967482dfdd4e63795/taichi/transforms/type_check.cpp#L21-L41.

tests/cpp/ir/frontend_type_inference_test.cpp
```diff
@@ -52,31 +53,36 @@ TEST(FrontendTypeInference, UnaryOp) {
   auto prog = std::make_unique<Program>(Arch::x64);
   prog->config.default_fp = PrimitiveType::f64;
   auto const_i16 = value<int16>(-(1 << 10));
-  const_i16->type_check(nullptr);
+
+  CompileConfig dummy_config;
```
strongoier, line 57: Why these changes?

strongoier (commented):
Should `UnaryOpStmt` in `type_check.cpp` also be modified?

taichi/transforms/scalarize.cpp
jim19930609: Yup, it's actually fine for `compile_kernel()` since `type_check()` is called immediately after `lower_ast()`. However, it doesn't make much sense to me as `type_check()` for `compile_function()` is placed at the very end:

2022-09-20 09-38-34 的屏幕截图

tests/cpp/ir/frontend_type_inference_test.cpp
jim19930609: `UnaryOpExpression::type_check(CompileConfig*)` will attempt to obtain `real_matrix` and `default_fp` from the `config` . It will hit an assertion if nullptr is passed in. 

(But I did overkilled a few Expressions other than UnaryOpExpression, will have that fixed, thanks!)

taichi/transforms/scalarize.cpp
strongoier: Ah here I mean `load_store_forwarding()` is unrelated, which should not appear in the comments.

strongoier (approved):
LGTM!

PR #6083, opened 2022-09-16
[aot] Support multi-target builds for Apple M1

PENGUINLIONG:
This PR enables Taichi to be built into multi-target libraries. Recent releases of macOS supports dynamically linking against x86_64 or arm64 of the same `.dylib` on demand, depending on the arch of the parent process. Previously this was not possible because the CMake script forced a `-march=nehalem` that is not a kind of arm64; and `CMAKE_OSX_ARCHITECTURES` is forced `arm64`.

The multi-target feature is only enabled with `-DCMAKE_OSX_ARCHITECTURES="arm64;x86_64"`. Python builds will select one of the archs, based on the arch of the skbuild Python process.

图片

### DIFF
--- a/cmake/TaichiCXXFlags.cmake
+++ b/cmake/TaichiCXXFlags.cmake
@@ -81,8 +81,15 @@ if ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64" OR "${CMAKE_SYSTEM_PROCESSOR}"
     if (MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /D \"TI_ARCH_x64\"")
     else()
-        message("Setting -march=nehalem for x86_64 processors")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=nehalem -DTI_ARCH_x64")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DTI_ARCH_x64")
+        if ("arm64" IN_LIST CMAKE_OSX_ARCHITECTURES)
+            # TODO: (penguinliong) Will probably need this in a future version
+            # of Clang. Clang11 doesn't recognize this.
+            #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mcpu=apple-m1")
+        else()
+            message("Setting -march=nehalem for x86_64 processors")
+            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=nehalem")
+        endif()
     endif()
     set(ARCH "x64")
 elseif ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "arm64")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,20 +6,6 @@ cmake_minimum_required(VERSION 3.17)
 
 project(taichi)
 
-# Taichi does not set target architecture explicitly,
-# but rather rely on CMake to detect the host arch.
-#
-# However on Mac m1, there are two available architectures namely x86_64 and arm64.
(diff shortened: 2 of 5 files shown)


bobcao3:
What about MoltenVK? Do we need to start building and linking to it statically?

PENGUINLIONG:
> What about MoltenVK? Do we need to start building and linking to it statically?

@bobcao3 MoltenVK in VulkanSDK is already multitargeting.

bobcao3:
> > What about MoltenVK? Do we need to start building and linking to it statically?
> 
> @bobcao3 MoltenVK in VulkanSDK is already multitargeting.

What about the version from homebrew, which is the one we are using?

PENGUINLIONG:
> > > What about MoltenVK? Do we need to start building and linking to it statically?
> > 
> > 
> > @bobcao3 MoltenVK in VulkanSDK is already multitargeting.
> 
> What about the version from homebrew, which is the one we are using?

Ah, yes. Homebrew serves on a per-arch basis, and our test runs install MoltenVK from homebrew. It could be an issue if someone installed homebrew molten-vk and a x86_64 Anaconda 3 (like me). We can resolve this issue in a future PR I think. And we can then get rid of the annoying `TI_LIB_DIR`.

bobcao3:
> > > > What about MoltenVK? Do we need to start building and linking to it statically?
> > > 
> > > 
> > > @bobcao3 MoltenVK in VulkanSDK is already multitargeting.
> > 
> > 
> > What about the version from homebrew, which is the one we are using?
> 
> Ah, yes. Homebrew serves on a per-arch basis, and our test runs install MoltenVK from homebrew. It could be an issue if someone installed homebrew molten-vk and a x86_64 Anaconda 3 (like me). We can resolve this issue in a future PR I think. And we can then get rid of the annoying `TI_LIB_DIR`.

Currently we ship a molten dylib, so it will not work

PENGUINLIONG:
> > > > > What about MoltenVK? Do we need to start building and linking to it statically?
> > > > 
> > > > 
> > > > @bobcao3 MoltenVK in VulkanSDK is already multitargeting.
> > > 
> > > 
> > > What about the version from homebrew, which is the one we are using?
> > 
> > 
> > Ah, yes. Homebrew serves on a per-arch basis, and our test runs install MoltenVK from homebrew. It could be an issue if someone installed homebrew molten-vk and a x86_64 Anaconda 3 (like me). We can resolve this issue in a future PR I think. And we can then get rid of the annoying `TI_LIB_DIR`.
> 
> Currently we ship a molten dylib, so it will not work

Shouldn't be a problem atm? As mentioned before the Python wheels are built per-arch so it won't break the existing release pipeline (as the tests do). The multi-target lib can only be built if the user explicitly specify `-DCMAKE_OSX_ARCHITECTURES="arm64;x86_64"`  in the cmake args. We still have a buffer of time to clean things up yet I somehow prefer to keep the PRs small enough to be legible.

cmake/TaichiCXXFlags.cmake
```diff
@@ -81,8 +81,15 @@ if ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64" OR "${CMAKE_SYSTEM_PROCESSOR}"
     if (MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /D \"TI_ARCH_x64\"")
     else()
-        message("Setting -march=nehalem for x86_64 processors")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=nehalem -DTI_ARCH_x64")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DTI_ARCH_x64")
+        if ("arm64" IN_LIST CMAKE_OSX_ARCHITECTURES)
+            # TODO: (penguinliong) Will probably need this in a future version
+            # of Clang. Clang11 doesn't recognize this.
+            #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mcpu=apple-m1")
```
ailzhang, line 88: I wonder if this works if we use AppleClang to build cpp files and only use clang-11 to generate bc files?

ailzhang (approved):
LGTM!

PR #6089, opened 2022-09-17
[bug] Fix crashing when loading old offline cache files

### LINKED ISSUE #6081: Redesign loading offline cache metadata to avoid crashing when loading old version files
Loading old version metadata may cause a crash if the structure has changed.

```
[Taichi] version 1.1.3, llvm 10.0.0, commit e3eb6be7, linux, python 3.9.7
  File "/home/yc/Desktop/mpm-collision/run.py", line 11, in <module>
    ti.init(arch=ti.cuda, dynamic_index=True)
  File "/home/yc/taichi/python/taichi/lang/misc.py", line 455, in init
    impl.get_runtime().create_program()
  File "/home/yc/taichi/python/taichi/lang/impl.py", line 318, in create_program
    self.prog = _ti_core.Program()
MemoryError: std::bad_alloc
```

### To avoid crashing:
Deserialize directly `==>>` Only deserialize the version field of metadata && Version checking && Deserialize

PGZXB:
Issue: #4401, fixes #6081

In future, if necessary, we should maintain a version number for offline cache instead of using the Taichi version directly.

### DIFF
--- a/cmake/TaichiTests.cmake
+++ b/cmake/TaichiTests.cmake
@@ -20,7 +20,8 @@ file(GLOB_RECURSE TAICHI_TESTS_SOURCE
         "tests/cpp/llvm/*.cpp"
         "tests/cpp/program/*.cpp"
         "tests/cpp/struct/*.cpp"
-        "tests/cpp/transforms/*.cpp")
+        "tests/cpp/transforms/*.cpp"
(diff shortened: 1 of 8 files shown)


ailzhang (approved):
Thanks！

PR #6097, opened 2022-09-19
[aot] Support Vulkan layers on MoltenVK platforms

PENGUINLIONG:
Vulkan layers cannot be loaded by directly loading `libMoltenVK.dylib` because it's merely an implementation of Vulkan driver. This issue is a major obstable on the provision of `print` in the Vulkan backend. Link to the ICD loader `libvulkan.dylib` instead.


PENGUINLIONG:
cc @ailzhang You might wanna check if it's still working on a machine with no Vulkan SDK nor MoltenVK installed.

ailzhang:
Thanks! I've verified that this PR enables loading vulkan validation layer on macos, although printing still doesn't work as my M1 machine doesn't support non semantic info extension :(

PR #6098, opened 2022-09-19
[vulkan] Instruct users to install vulkan sdk if they want to use validation layer

ailzhang:
Issue:  #2713


PENGUINLIONG (approved):
lgtm

PR #6115, opened 2022-09-20
[misc] Allow more commits in changelog generation

strongoier:
### Brief Summary

There are more than 200 commits in v1.1.3 (https://github.com/taichi-dev/taichi/releases/tag/v1.1.3) so let's increase the number of commits for changelog generation.


ailzhang (approved):
Thanks!

PR #6119, opened 2022-09-20
[Lang] MatrixField refactor 5/n: Lower access of matrix field element into CHI IR

strongoier:
Issue: #5959

### Brief Summary

This PR lowers access of matrix field element into CHI IR. To verify its functionality, this PR also makes the following path of #5959 running end-to-end (currently only enabled when `real_matrix=True`):

MatrixField_Part5

Key steps:
- Python: when running into subscript of `MatrixField`, create `IndexExpression` in C++ instead of building `_MatrixFieldElement` in Python
- Frontend IR: allow `IndexExpression` to take `MatrixFieldExpression`
- CHI IR: add `MatrixOfGlobalPtrStmt`; add `lower_matrix_ptr` pass, which demotes `PtrOffsetStmt(MatrixOfGlobalPtrStmt)` into `GlobalPtrStmt` in the case of constant index - this is incomplete and only aims at supporting the above path

### DIFF
--- a/taichi/ir/expr.cpp
+++ b/taichi/ir/expr.cpp
@@ -27,13 +27,8 @@ Expr bit_cast(const Expr &input, DataType dt) {
 }
 
 Expr Expr::operator[](const ExprGroup &indices) const {
-  if (is<IndexExpression>()) {
-    // Allow indexing IndexExpression with ret_type = TensorType
-    TI_ASSERT(is_tensor(expr->ret_type));
-  } else {
-    TI_ASSERT(is<FieldExpression>() || is<MatrixFieldExpression>() ||
-              is<ExternalTensorExpression>() || is<IdExpression>());
-  }
+  TI_ASSERT(is<FieldExpression>() || is<MatrixFieldExpression>() ||
+            is<ExternalTensorExpression>() || is_tensor(expr->ret_type));
   return Expr::make<IndexExpression>(*this, indices);
 }
 
--- a/taichi/ir/expr.cpp
+++ b/taichi/ir/expr.cpp
@@ -31,8 +31,8 @@ Expr Expr::operator[](const ExprGroup &indices) const {
     // Allow indexing IndexExpression with ret_type = TensorType
     TI_ASSERT(is_tensor(expr->ret_type));
   } else {
-    TI_ASSERT(is<FieldExpression>() || is<ExternalTensorExpression>() ||
-              is<IdExpression>());
+    TI_ASSERT(is<FieldExpression>() || is<MatrixFieldExpression>() ||
+              is<ExternalTensorExpression>() || is<IdExpression>());
   }
   return Expr::make<IndexExpression>(*this, indices);
 }
--- a/taichi/ir/frontend_ir.cpp
+++ b/taichi/ir/frontend_ir.cpp
@@ -379,13 +379,6 @@ void ExternalTensorExpression::flatten(FlattenContext *ctx) {
   stmt = ctx->back_stmt();
 }
 
-void FieldExpression::flatten(FlattenContext *ctx) {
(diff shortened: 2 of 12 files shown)


taichi/ir/expr.cpp
```diff
@@ -31,8 +31,8 @@ Expr Expr::operator[](const ExprGroup &indices) const {
     // Allow indexing IndexExpression with ret_type = TensorType
     TI_ASSERT(is_tensor(expr->ret_type));
   } else {
-    TI_ASSERT(is<FieldExpression>() || is<ExternalTensorExpression>() ||
-              is<IdExpression>());
+    TI_ASSERT(is<FieldExpression>() || is<MatrixFieldExpression>() ||
```
AD1024, line 34: Would this check disable usage such as `(a @ b)[index]`?

taichi/ir/frontend_ir.cpp
```diff
@@ -401,7 +399,30 @@ Stmt *make_field_access(Expression::FlattenContext *ctx,
...
+  SNode *snode = var.cast<FieldExpression>()->snode;
+  return ctx->push_back(std::make_unique<GlobalPtrStmt>(
+      snode, make_index_stmts(ctx, indices, snode->index_offsets)));
+}
+
+Stmt *make_matrix_field_access(Expression::FlattenContext *ctx,
+                               Expr var,
+                               ExprGroup indices,
+                               DataType ret_type) {
+  auto matrix_field = var.cast<MatrixFieldExpression>();
```
jim19930609, line 417: assert `matrix_field != nullptr` after dynamic pointer cast? Or take `MatrixFieldExpression var` explicitly?

taichi/transforms/lower_matrix_ptr.cpp
```diff
@@ -0,0 +1,60 @@
...
+TLANG_NAMESPACE_BEGIN
+
+class LowerMatrixPtr : public BasicStmtVisitor {
+ public:
+  using BasicStmtVisitor::visit;
+  DelayedIRModifier modifier;
+
+  void visit(PtrOffsetStmt *stmt) override {
+    if (stmt->origin->is<MatrixOfGlobalPtrStmt>()) {
+      TI_ASSERT(stmt->offset->is<ConstStmt>());
```
jim19930609, line 16: What about the situation where a MatrixField is indexed with a LoopIndex? for example:
```
x = ti.Matrix.field(m=2, shape=(5))
@ti.kernel
def test():
  for i in x:
    for j in range(2):
       x[i][j] = 100
```
In this case, x[i] is MatrixOfGlobalPtrStmt, and x[i][j] is a PtrOffsetStmt with offset being `LoopIndexStmt`?

tests/python/test_matrix.py
```diff
@@ -248,8 +248,7 @@ def func4(k: ti.i32):
     func4(10)
 
 
-@test_utils.test(arch=ti.cpu)
-def test_matrix_constant_index():
+def _test_matrix_constant_index():
```
jim19930609, line 251: Maybe add some tests with elementwise operations involved in a following PR~

taichi/transforms/lower_matrix_ptr.cpp
strongoier: This case requires dynamic index, which will be supported in a future PR :-)

taichi/ir/expr.cpp
strongoier: Nice catch. Fixed.

taichi/ir/frontend_ir.cpp
strongoier: Sure. I've chosen the latter.

jim19930609 (approved):
LGTM!

Issue #5769, opened 2022-08-15
Opengl aot support tracking issue


ailzhang:
related: #5634 

We'll use this issue to track progress in opengl aot support. 

To give some context, we removed opengl codegen and moved it on top of spirv codegen in #5246 which relieves our burden to maintain multiple codegens. But we haven't migrated or provided a runtime library (like what we did for vulkan backend) to run our android app yet. 

After migrating on top of spirv, opengl aot actually can benefit a lot from the common gfx runtime and we just need to complete the following milestones:

- [ ] Fix existing buggy opengl aot tests 
- [ ] Expand python aot tests to cover opengl backend
- [ ] Expand cpp aot tests to cover opengl backend
- [ ] Run implicit_fem desktop demo on top of opengl backend
- [ ] Make sure `use_gles` is properly handled and run implicit_fem android demo

And ofc, we will fix whatever pops up when we try to accomplish the milestones above ;)

PR #6121, opened 2022-09-21
[opengl] Fix GLES support

ailzhang:
Issue: #5769 

### Brief Summary
GLES support was broken since the migration to spirv-based codegen and this PR is the first step to bring it back. Also added a bunch of `check_opengl_error` calls to make debugging easier. 
Note thorough testing for GLES backend is currently missing and should be added in followup PRs.


PENGUINLIONG (approved):
lgtm

PR #6124, opened 2022-09-21
[bug] Let every thread has its own CompileConfig

### LINKED ISSUE #6123: Constant folding pass changes the program config for all threads
**Describe the bug**
https://github.com/taichi-dev/taichi/blob/a4b24d1524a82293a4df293e9628acf514d251b9/taichi/transforms/constant_fold.cpp#L221-L242

This part changes the config for all threads, we only want to change the config for this thread.
This is why compiling in parallel will result in random failures.

lin-hitonami:
Issue: fixes #6123 

### Brief Summary
This PR lets every thread has its own CompileConfig. 
I've ran the example `example/simulation/mpm_lagrangian_forces.py` with `num_compile_threads=4`  50 times which usually failed before this PR and it didn't fail this time so I think we can make parallel compilation default on now. 

(This PR contains #6126 and I'll merge this PR after #6126 is merged if this PR is approved)


taichi/program/program.h
```diff
@@ -94,7 +94,8 @@ class TI_DLL_EXPORT Program {
  public:
   using Kernel = taichi::lang::Kernel;
   Callable *current_callable{nullptr};
-  CompileConfig config;
+  std::unordered_map<std::thread::id, CompileConfig> configs;
```
ailzhang, line 97: hmmm I guess we have an assumption here that each thread just mutates its own Compileconfig so that we didn't have any locks (good to put it in the comment) .
I wonder if thread local storage is a better choice than this?
In fact what's even better is that we can force constant folding to use its own copy of Compileconfig to compile?

lin-hitonami: Yeah, but this will include many refactors... maybe we can do it when we implement thread-safe Taichi...

ailzhang: Yep I'm fine with this as a temp workaround, would you mind adding some comments explaining that this why we have this & in the future we'd like to fix constant folding instead? Thanks!

ailzhang (approved):
Thanks!

taichi-gardener:

PR #6126, opened 2022-09-21
[bug] [llvm] Fix is_same_type when the suffix of a type is the prefix of the suffix of the other type

lin-hitonami:
Issue: #5511

### Brief Summary
is_same_type fails when it encounters comparison between `%struct.PhysicalCoordinates.1` and `%struct.PhysicalCoordinates.11` where `1` is the prefix of `11`.

This bug does not affect 1.1.3 because parallel compilation is turned off in 1.1.3.


ailzhang (commented):
LGTM! 
nit: I feel like it's worth adding some unit tests in C++ against `is_same_type` already :P

strongoier (approved):
LGTM!

PR #6128, opened 2022-09-21
[Llvm] Fix codegen for div (unsigned)

### LINKED ISSUE #6118: Wrong result when doing mod on two uint32 values
**Describe the bug**
Hi, I try to use two uint32 values to do a mod but found the result is incorrect.

**To Reproduce**
```py
pos = np.array([660, 241, 838])
pos_field = ti.field(ti.uint32, shape=(3))
pos_field.from_numpy(pos)

primes_field = ti.field(ti.uint32, shape=(3))
primes = np.uint32(np.array([1, 2654435761, 805459861]))
primes_field.from_numpy(primes)

map_size_field = ti.field(ti.uint32, shape=(1,))
size = np.uint32(np.array([524288]))
map_size_field.from_numpy(size)

@ti.kernel
def test_kernel():
    # map_size = ti.uint32(524288)
    for ind in range(1):
        result = ti.uint32(0)
        for i in ti.static(range(3)):
            result ^= (pos_field[i] * primes_field[i])
        
        result = result % map_size_field[0]
        print(result)
```

When  I use the **map_size**  instead of **map_size_field**, the result is correct, and the correct result should be 212107, but the code above gives 4294655115, which is almost an overflow for uint32.

Any idea what kind of bug this is, or I'm making some mistakes here?

strongoier:
Issue: fix #6118

### DIFF
--- a/taichi/codegen/llvm/codegen_llvm.cpp
+++ b/taichi/codegen/llvm/codegen_llvm.cpp
@@ -490,9 +490,12 @@ void TaskCodeGenLLVM::visit(BinaryOpStmt *stmt) {
     if (is_real(stmt->ret_type)) {
       llvm_val[stmt] =
           builder->CreateFDiv(llvm_val[stmt->lhs], llvm_val[stmt->rhs]);
-    } else {
+    } else if (is_signed(stmt->ret_type)) {
       llvm_val[stmt] =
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Thanks!

Issue #5350, opened 2022-07-07
[perf] Enable load store forwarding across loop hierarchy for vulkan performance.


turbo0628:
The `store_to_load_forwarding` transformation in the `cfg_optimization` pass ignores ExternalPtrStmt. It causes performance issue in [the SPH case](https://github.com/taichi-dev/taichi/issues/5046). The CUDA backend is insensitive to this optimization, but Vulkan backend cannot handles this well even on the same hardware.

I think we can get a ~2.2x speed up for the SPH case when this optimization is enabled.

The related source code section is [here](https://github.com/taichi-dev/taichi/blob/master/taichi/ir/control_flow_graph.cpp#L264-L294).

qiao-bo:
After some investigation, this seems not related to Ndarray in the `store_to_load_forwarding` pass. The code block (as mentioned) does consider Ndarray, which is `ExternalPtrStmt`:
https://github.com/taichi-dev/taichi/blob/233f017f00753986b8aea194b2e30ca7a727159e/taichi/ir/control_flow_graph.cpp#L264-L294

Let be Ndarray or Field, the pattern will start with a `GlobalLoadStmt`, hence the else-if clause will be executed, and `store_forwarding` remains true. 

The problem here is that the `get_store_forwarding_data` function cannot find the forwarded data for certain scenarios:
https://github.com/taichi-dev/taichi/blob/233f017f00753986b8aea194b2e30ca7a727159e/taichi/ir/control_flow_graph.cpp#L291

For example, a dummy kernel like this can be optimized by `store_to_load_forwarding`:
```
@ti.kernel
def foo_ndarray(acc: ti.any_arr(field_dim=1)):
    for i in range(16):
        acc[i] = 1.0
        acc[i] -= i 
        acc[i] %= i 
```
Nevertheless, if the access happens e.g., in a loop, the optimization no longer works:
```
@ti.kernel
def foo_field(acc: ti.template()):
    for i in range(16):
        acc[i] = 1.0
        for j in range(8):
            acc[i] -= j 
        for k in range(4):
            acc[i] %= k 

```
This is regardless of Field/Ndarray. The question here becomes how to adapt the `get_store_forwarding_data` function to consider the second case forwardable.

turbo0628:
OK, so this is not a cause of ndarray vs field performance gap.

qiao-bo:
[...]

PR #6129, opened 2022-09-22
[bug] [vulkan] Only enable non_semantic_info cap when validation layer is on

ailzhang:
related: #5350

Thanks @turbo0628 and @lin-hitonami for catching this! It turns out codegening non-semantic info spirv code isn't no-op when validation is off. It slows down the program execution so we should properly guard it.


lin-hitonami (approved):
Thanks!

PR #6132, opened 2022-09-22
[Misc] Add prefix sum executor to avoid multiple field allocations

YuCrazing:
1. Add `PrefixSumExecutor` helper to avoid every prefix sum function call's memory allocation.
2. Move `parallel_sort` and `prefix_sum` utilities from `_kernels.py` to `algorithms.py`.

Maybe we can also integrate @yuanming-hu 's CPU prefix sum into `PrefixSumExecutor` in the future.


tests/python/test_scan.py
```diff
@@ -4,6 +4,10 @@
 
 @test_utils.test(arch=[ti.cuda, ti.vulkan], exclude=[(ti.vulkan, "Darwin")])
 def test_scan():
+
+    # A global prefix sum wrapper, only need to be initialized once.
+    executor = ti.algorithms.PrefixSumExecutor()
```
ailzhang, line 9: nit: shall we use `ti.alg.` so that it's shorter for users? :P

YuCrazing: IMHO, `algorithms` makes it easier to be distinguished from `linalg`. `algorithms` means element-manipulating algorithms (like C++ algorithm library) and the `linalg` is for mathematic linear algebra.

ailzhang: aha I forgot that we have linalg, then all good!

python/taichi/algorithms.py
```diff
@@ -0,0 +1,93 @@
...
+    print(num_stages)
+
+
+# Inclusive In-Place's Parallel Prefix Sum (Scan)
+# Ref[0]: https://developer.download.nvidia.com/compute/cuda/1.1-Beta/x86_website/projects/scan/doc/scan.pdf
+# Ref[1]: https://github.com/NVIDIA/cuda-samples/blob/master/Samples/2_Concepts_and_Techniques/shfl_scan/shfl_scan.cu
+@data_oriented
+class PrefixSumExecutor:
+    def __init__(self, length):
+        self.large_arr = None
```
ailzhang, line 41:
```suggestion
```

python/taichi/algorithms.py
```diff
@@ -0,0 +1,93 @@
...
+        while ele_num > 1:
+            ele_num = int((ele_num + BLOCK_SZ - 1) / BLOCK_SZ)
+            self.ele_nums.append(ele_num)
+            start_pos += BLOCK_SZ * ele_num
+            self.ele_nums_pos.append(start_pos)
+
+        self.large_arr = field(i32, shape=start_pos)
+
+    def run(self, input_arr):
+
```
ailzhang, line 63:
```suggestion
```

ailzhang (approved):
LGTM, thanks! 
Two nits before merging: 
1. docstring ;) 
2. You probably want to track user facing APIs here https://github.com/taichi-dev/taichi/blob/master/tests/python/test_api.py#L146 to avoid accidental changes ;)

python/taichi/algorithms.py
```diff
@@ -0,0 +1,90 @@
...
+from taichi.lang.runtime_ops import sync
+from taichi.lang.simt import subgroup
+from taichi.types.primitive_types import i32
+
+
+# Odd-even merge sort
+# References:
+# https://developer.nvidia.com/gpugems/gpugems2/part-vi-simulation-and-numerical-algorithms/chapter-46-improved-gpu-sorting
+# https://en.wikipedia.org/wiki/Batcher_odd%E2%80%93even_mergesort
+def parallel_sort(keys, values=None):
```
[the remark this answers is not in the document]
YuCrazing: Done. Since I am not familiar with the usage of `parallel_sort`, I left it's docstring blank. Maybe we need @AmesingFlank to complete it later. :)

taichi-gardener:

Issue #6134, opened 2022-09-22
Remove codegen for demoted operations


strongoier:
There's a mandatory `demote_operations` pass (https://github.com/taichi-dev/taichi/blob/master/taichi/transforms/demote_operations.cpp) which demotes high-level operations into low-level ones so that there's no need to handle these high-level ones in codegen of different backends. Those legacy codegen should be removed to keep the codebase clean.

- [x] `pow()` with an integer exponent
- [x] `floordiv`
- [x] `bit_shr`
- [ ] <del>`BitExtractStmt`</del> (See #6141 & #6208)

PR #6135, opened 2022-09-22
[refactor] Remove redundant codegen of floordiv

strongoier:
Issue: #6134

`floordiv` is already demoted and should not appear in codegen.


lin-hitonami (approved):
LGTM

PR #6136, opened 2022-09-22
Revert "[spirv] Fixed OpLoad with physical address"

lin-hitonami:
Reverts taichi-dev/taichi#5212

PR #5212 fixes a warning on Vulkan validation layer, but it introduces performance regression to Vulkan backend.


ailzhang (approved):
Thanks!

PR #6138, opened 2022-09-22
[aot] Track layouts for imported image & tests

PENGUINLIONG:
The image interop APIs doesn't track image layouts atm, which leads to unexpected crash.


ailzhang (approved):
Thanks!

PR #6139, opened 2022-09-22
[vulkan] Fix SPV physical ptr load alignment

bobcao3:
Issue: #6072
Ref #6136

### Brief Summary

We misread the specs, the `aligned` parameter takes a uint literal, not a value. We used to feed in a value, but since SPIR-V wouldn't care, we are actually feeding the value id as an alignment and probably causing performance regression due to that.


lin-hitonami (approved):
LGTM!

PR #6140, opened 2022-09-22
[Do Not Merge] Avoid third-party symbol exportations for libtaichi_c_api.so

### LINKED ISSUE #5872: [aot] libtaichi_c_api.so exports unnecessary third-party symbols
`libtaichi_c_api.so` should only export symbols in need, but now it exports a great many third-party symbols as well. Of such, certain third-party global symbols can potentially cause conflict/override if the same symbol is also exported from some other dynamic libraries. 

For example, the following openGL global symbols:
2022-08-24 15-21-07 的屏幕截图

Make sure we turn those global symbols to local symbols before C-API release.

[log.txt](https://github.com/taichi-dev/taichi/files/9413740/log.txt)

jim19930609:
Issue: fix #5872

*Note: This PR will also remove GGUI symbols from libtaichi_c_api.so on MacOS. Do not have this PR merged until we replaced all the GGUI renderers in taichi-aot-demos.


c_api/version_scripts/export_symbols_mac.lds
```diff
@@ -0,0 +1,3 @@
+# MacOS
+_ti_*
```
ailzhang, line 2: nit: can we add a comment in each file to remind people that these version scripts should match?

.github/workflows/scripts/aot-demo.sh
```diff
@@ -165,4 +165,21 @@ function build-and-test-headless-demo-desktop {
     python3 ci/run_tests.py -l $TAICHI_C_API_INSTALL_DIR
 }
 
+function check-c-api-export-symbols {
+    cd taichi
+    TAICHI_REPO_DIR=$(pwd)
+    TAICHI_C_API_DIR=$(find $TAICHI_REPO_DIR -name libtaichi_c_api.* | head -n 1)
+
+    EXPORT_SYM=" T \| B \| D "
+    CAPI_SYM=" _\?ti_"
```
ailzhang, line 174: ditto: this actually also need to match what's written in version scripts

ailzhang (approved):
LGTM!

PR #6141, opened 2022-09-22
[refactor] Remove redundant codegen of BitExtractStmt

strongoier:
Issue: #6134

### DIFF
--- a/taichi/codegen/cc/codegen_cc.cpp
+++ b/taichi/codegen/cc/codegen_cc.cpp
@@ -74,12 +74,6 @@ class CCTransformer : public IRVisitor {
             stmt->tb);
   }
 
-  void visit(BitExtractStmt *stmt) override {
-    emit("{} = (({} >> {}) & ((1 << {}) - 1));",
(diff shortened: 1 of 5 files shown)


bobcao3:
There is actually hardware support on both Nvidia and AMD for native bit extract. Maybe we should keep them not demoted?

strongoier:
> There is actually hardware support on both Nvidia and AMD for native bit extract (a lot faster than using lower level bit operations). Maybe we should keep them not demoted?

Could you elaborate a bit on what should we do to make use of that? Regarding this PR, `demote_operations` is just an abstract layer of what current codegen of different backends does.

bobcao3:
> > There is actually hardware support on both Nvidia and AMD for native bit extract (a lot faster than using lower level bit operations). Maybe we should keep them not demoted?
> 
> Could you elaborate a bit on what should we do to make use of that? Regarding this PR, `demote_operations` is just an abstract layer of what current codegen of different backends does.

We would want to be able to optionally demote BFE when backend says it's not supported, and if supported the codegen will want to take the BFE CHI-IR and generate the backend native instruction

PR #6146, opened 2022-09-23
[aot] Improve C++ wrapper implementation

PENGUINLIONG:
RT

### DIFF
--- a/c_api/include/taichi/cpp/taichi.hpp
+++ b/c_api/include/taichi/cpp/taichi.hpp
@@ -68,6 +68,7 @@ THandle move_handle(THandle &handle) {
 class Memory {
   TiRuntime runtime_{TI_NULL_HANDLE};
   TiMemory memory_{TI_NULL_HANDLE};
+  size_t size_{0};
   bool should_destroy_{false};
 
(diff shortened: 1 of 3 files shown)


c_api/include/taichi/cpp/taichi.hpp
```diff
@@ -102,10 +107,36 @@ class Memory {
...
   }
 
+  void *map() const {
+    return ti_map_memory(runtime_, memory_);
+  }
+  void unmap() const {
+    ti_unmap_memory(runtime_, memory_);
+  }
+
+  inline void read(void *dst, size_t size) const {
```
ailzhang, line 122: I wonder if read/write should go through staging buffer instead of depending on map/unmap? As some devallocs aren't allocated with host_read/write and map/unmap returns nullptr?

ailzhang (approved):
LGTM!

c_api/include/taichi/cpp/taichi.hpp
PENGUINLIONG: Will ensure every map access is valid by staging buffer

PR #6153, opened 2022-09-23
[Bug] [lang] Fix augmented assign for sar

strongoier:
### Brief Summary

While I was removing redundant codegen for `shr` (#6134), I found that augmented assign for `sar` was accidentally calling `shr`, so I fixed it by the way.

### DIFF
--- a/taichi/transforms/demote_operations.cpp
+++ b/taichi/transforms/demote_operations.cpp
@@ -118,8 +118,8 @@ class DemoteOperations : public BasicStmtVisitor {
                is_signed(lhs->element_type())) {
       // @ti.func
       // def bit_shr(a, b):
-      //     signed_a = ti.cast(a, ti.uXX)
-      //     shifted = ti.bit_sar(signed_a, b)
+      //     unsigned_a = ti.cast(a, ti.uXX)
+      //     shifted = ti.bit_sar(unsigned_a, b)
       //     ret = ti.cast(shifted, ti.iXX)
       //     return ret
       auto unsigned_cast = Stmt::make<UnaryOpStmt>(UnaryOpType::cast_bits, lhs);
--- a/tests/python/test_bit_operations.py
+++ b/tests/python/test_bit_operations.py
@@ -13,8 +13,14 @@ def test_bit_shl():
     def shl(a: ti.i32, b: ti.i32) -> ti.i32:
         return a << b
 
+    @ti.kernel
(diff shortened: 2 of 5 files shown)


taichi/transforms/demote_operations.cpp
```diff
@@ -118,8 +118,8 @@ class DemoteOperations : public BasicStmtVisitor {
                is_signed(lhs->element_type())) {
       // @ti.func
       // def bit_shr(a, b):
-      //     signed_a = ti.cast(a, ti.uXX)
-      //     shifted = ti.bit_sar(signed_a, b)
+      //     unsigned_a = ti.cast(a, ti.uXX)
+      //     shifted = ti.bit_sar(unsigned_a, b)
```
lin-hitonami, line 122: Why is it sar here?

tests/python/test_bit_operations.py
```diff
@@ -13,8 +13,14 @@ def test_bit_shl():
...
 
+    @ti.kernel
+    def shl_assign(a: ti.i32, b: ti.i32) -> ti.i32:
+        c = a
+        c <<= b
+        return c
+
     for i in range(8):
-        assert shl(3, i) == 3 * 2**i
+        assert shl(3, i) == shl_assign(3, i) == 3 * 2**i
```
jim19930609, line 23: `3 <<= 1` gives the same result regardless of whether it's "signed shift" or "unsigned shift". Shall we test with negative number or positive number large enough(`01xxxxxx`) to influence the sign-bit after shift?

taichi/transforms/demote_operations.cpp
strongoier: `sar` for unsigned is equivalent to `shr`.

tests/python/test_bit_operations.py
strongoier: `left shift` gives the same result anyway. Only `right shift` will be influenced by the sign bit.

lin-hitonami: IIRC shl and sal are the same, and they do not preserve the signed bit.

taichi/transforms/demote_operations.cpp
strongoier: I guess the main goal of the demotion is to make codegen simpler (only `sar` needed).

lin-hitonami (approved):
LGTM

Issue #6107, opened 2022-09-20
Add support for initial window position


brent-williams:
Add support for initial window position in GUI and/or GGUI.  

Window position could be in the constructor alongside resolution.  The underlying imgui.ini could as well/instead
be used, currently on macOS Monterey the file is written but always with window name "Debug##Default" and 
the Pos key is sadly ignored if already present.

ailzhang:
Closing via #6156, @brent-williams please feel free to give it a try and let us know if you have any followup questions!

brent-williams:
Cheers @ailzhang, will check it out once 1.1.4 is released.  Wow nice turnaround speed!

PR #6156, opened 2022-09-24
[gui] Support for setting the initial position of GGUI window

Morcki:
Issue: #6107


python/taichi/ui/window.py
```diff
@@ -26,14 +26,14 @@ class Window:
         vsync (bool): whether or not vertical sync should be enabled.
         show_window (bool): where or not display the window after initialization.
     """
-    def __init__(self, name, res, vsync=False, show_window=True):
+    def __init__(self, name, res, pos=(0, 0), vsync=False, show_window=True):
```
PENGUINLIONG, line 29: pos=(0,0) might be a problem for Windows users who put the task bar on top. Consider (100, 100) or something like that.

taichi/ui/utils/utils.h
```diff
@@ -90,6 +92,12 @@ inline GLFWwindow *create_glfw_window_(const std::string &name,
     printf("GLFW reports no Vulkan support\n");
   }
 
+  // Reset the window hints to default
+  glfwDefaultWindowHints();
```
PENGUINLIONG, line 96: Why?

PENGUINLIONG (approved):
otherwise lgtm

python/taichi/ui/window.py
Morcki: Let me try try on Windows.

taichi/ui/utils/utils.h
Morcki: It seems that the Window we create should be invisible before we set the position.

python/taichi/ui/window.py
Morcki: You are right, I will fix that.

ailzhang: If you need to add a kwarg, it'd be better to add it at the end to avoid breaking current api. e.g.  if user had `Window('test', (512, 512), False, True)` it'll be broken by this PR. 
Also `pos` need a proper docstring explaining which window, and where is the origin ;)

ailzhang (commented):
LGTM!

Morcki:
> LGTM!

That's right, I will fix that.

python/taichi/ui/window.py
```diff
@@ -26,14 +26,19 @@ class Window:
         vsync (bool): whether or not vertical sync should be enabled.
         show_window (bool): where or not display the window after initialization.
```
ailzhang, line 27: @Morcki missing docstring here ;)

ailzhang (approved):
LGTM!

python/taichi/ui/window.py
Morcki: oh sorry... thanks for reminding

PR #6160, opened 2022-09-25
[metal] Maintain a print string table per kernel

PGZXB:
Issue: #4401 
* Fix a potential bug in metal AOT
* Prepare for implementing offline cache on metal


ailzhang (approved):
LGTM!

PR #6161, opened 2022-09-26
[lang] Remove print at the end of parallel sort

turbo0628:
Issue: https://forum.taichi.graphics/t/topic/3074


ailzhang (approved):
Thanks!

PR #6163, opened 2022-09-26
[misc] Remove TI_UI namespace macros

### LINKED ISSUE #6147: Eliminate namespace macros
Currently we have macros `TLANG_NAMESPACE_BEGIN/END` which are equivalent to `namespace taichi::lang {}`. The macros may confuse the first-time readers so let's change it to `namespace taichi::lang {}`.

lin-hitonami:
Issue: fixes #6147

### DIFF
--- a/taichi/python/export_ggui.cpp
+++ b/taichi/python/export_ggui.cpp
@@ -25,7 +25,7 @@ namespace py = pybind11;
 #include "taichi/program/ndarray.h"
 #include <memory>
 
-TI_UI_NAMESPACE_BEGIN
+namespace taichi::ui {
 
 using namespace taichi::lang;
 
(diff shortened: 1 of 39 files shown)


taichi/python/export_ggui.cpp
```diff
@@ -25,7 +25,7 @@ namespace py = pybind11;
 #include "taichi/program/ndarray.h"
 #include <memory>
 
-TI_UI_NAMESPACE_BEGIN
+namespace taichi::ui {
```
jim19930609, line 28: Let's do:
```
namespace taichi { 
namespace ui {
```
to preserve the coding style?

jim19930609 (approved):

taichi/python/export_ggui.cpp
lin-hitonami: I don't think it's necessary, and I've replaced 
```
namespace taichi {
namespace lang {
}
}
```
with 
```
namespace taichi::lang {
}
```
in the previous PR...

PR #6165, opened 2022-09-26
[ci] Use beefy GitHub hosted runners for CPU tasks

feisuzhu:
Issue: #


feisuzhu:
Abort due to unexpected high cost.

PR #6166, opened 2022-09-26
[LLVM] Add runtime overflow detection on LLVM-based backends

lin-hitonami:
Issue: #5932 

### Brief Summary
This PR adds runtime overflow detection for addition, subtraction, multiplication and shift left operations on integer types (i/u8/16/32/64). This only works when the compiler is GCC or clang because I've used some built-in functions. This feature is not supported when compiled with MSVC.
The output of the example in #5932 when `debug=True`:
```
[Taichi] version 1.1.4, llvm 10.0.0, commit f892b138, linux, python 3.9.12
[Taichi] Starting on arch=x64
Multiplication overflow detected in File "/home/lin/test/ndrange.py", line 8, in populate:
    for i, j in ti.ndrange(n, n):
                ^^^^^^^^^^^^^^^^
```
There isn't a built-in overflow detection function to `shl`, so I implemented it by checking whether `a << b >> b == a`. 
The multiplication of power of two is optimized to `shl` in `alg_simp`, so sometimes the multiplication will be warned shift left overflow, but I think it should be ok?


jim19930609 (commented):
1. Add a test case to verify whether overflow check is triggered correctly
2. The other concern is that overflow checks like `__builtin_add_overflow` are compiler specific functions (GCC and Clang), and MSVC does not seem to support the same functions (https://developercommunity.visualstudio.com/t/please-implement-integer-overflow-detection/409051). So very likely it's gonna cause compilation failure with compilers other that GCC and Clang.

Shall we wrap it with some compiler-specific macros?

ailzhang (approved):
Awesome!!!

lin-hitonami:
Added `fflush(stdout)` to function `synchronize` of CPU backend to let pytest catch the output message of the program.

TODO: maybe output to stderr in the future.

jim19930609 (approved):
LGTM!

strongoier (approved):
LGTM!

PR #6169, opened 2022-09-27
[llvm] MatrixField refactor 7/n: Simplify codegen for TensorType allocation and access

strongoier:
Issue: #5959

### Brief Summary

This PR enforces that the return value of `AllocaStmt(TensorType)/GlobalTemporaryStmt(TensorType)` is a pointer to a `VectorType/ArrayType` (we no longer use a primitive type ptr for `TensorType`) so that the codegen can be simplified a lot.


jim19930609 (approved):
Super!

PR #6172, opened 2022-09-27
[autodiff] Add grad check feature and related test

PhrygianGates:
to do


python/taichi/ad/_ad.py
```diff
@@ -85,14 +103,133 @@ def __exit__(self, _type, value, tb):
...
     def insert(self, func, args):
+        # print("insert!")
         self.calls.append((func, args))
 
     def grad(self):
         assert self.entered, "Before evaluating gradients tape must be entered."
         assert not self.gradient_evaluated, "Gradients of grad can be evaluated only once."
         for func, args in reversed(self.calls):
-            func.grad(*args)
+            if hasattr(func, 'grad'):
```
erizmr, line 113: Just wondering if there are cases failed without this line?

python/taichi/ad/_ad.py
```diff
@@ -55,6 +67,12 @@ def __init__(self, loss=None, clear_gradients=True, validation=False):
                 Warning)
         self.eval_on_exit = loss is not None
         self.loss = loss
+        self.grad_check = grad_check
+        if self.grad_check:
+            self.watch = watch
```
erizmr, line 72: The `self.watch` seems never used?

python/taichi/ad/_ad.py
```diff
@@ -4,19 +4,31 @@
...
 
 class Tape:
-    def __init__(self, loss=None, clear_gradients=True, validation=False):
+    def __init__(self,
+                 loss=None,
+                 clear_gradients=True,
+                 validation=False,
+                 grad_check=None,
+                 watch=None,
+                 eps_range=2.**np.arange(-3, -30, -1).astype(np.float64)):
```
erizmr, line 31: I think it is not necessary to expose the `eps_range` since it is too tricky for users to tuning this parameter.

python/taichi/ad/_ad.py
```diff
@@ -85,14 +103,133 @@ def __exit__(self, _type, value, tb):
             self.grad()
 
     def insert(self, func, args):
+        # print("insert!")
```
erizmr, line 106: nit: remove this line? :)

python/taichi/lang/field.py
```diff
@@ -377,6 +378,13 @@ def getter(*key):
             def setter(value, *key):
                 assert len(key) == _ti_core.get_max_num_indices()
                 snode.write_float(key, value)
+                if impl.get_runtime().target_tape and impl.get_runtime(
+                ).target_tape.grad_check and not impl.get_runtime(
+                ).grad_replaced:
+                    for x in impl.get_runtime().target_tape.grad_check:
+                        assert snode != x.snode.ptr, "You can not write to input field."
```
erizmr, line 385: I think the error message is not so informative. It might be better to mention that the reason that overwritten is prohibitive is that we are doing grad check.

python/taichi/lang/field.py
```diff
@@ -392,6 +400,13 @@ def getter(*key):
             def setter(value, *key):
                 assert len(key) == _ti_core.get_max_num_indices()
                 snode.write_int(key, value)
+                if impl.get_runtime().target_tape and impl.get_runtime(
+                ).target_tape.grad_check and not impl.get_runtime(
+                ).grad_replaced:
+                    for x in impl.get_runtime().target_tape.grad_check:
+                        assert snode != x.snode.ptr, "You can not write to input field."
```
erizmr, line 407: Same to the error message in `getter`

python/taichi/ad/_ad.py
```diff
@@ -85,14 +103,133 @@ def __exit__(self, _type, value, tb):
...
+            for _i in range(node.ptr.get_num_ch()):
+                ch = node.ptr.get_ch(_i)
+                if not ch.is_place():
+                    visit(SNode(ch))
+                else:
+                    if not ch.is_primal():
+                        continue
+                    if mode == "save":
+                        self.backup_variables.append(
+                            ScalarField(Expr(ch.get_expr())).to_numpy())
```
erizmr, line 222: Have we handled the lose precision warning yet ?

python/taichi/ad/_ad.py
```diff
@@ -85,14 +103,133 @@ def __exit__(self, _type, value, tb):
...
+                x_grad_np = x.grad.to_numpy()
+                extra_dim = x_grad_np.ndim - tangent_np.ndim
+                if extra_dim == 1:
+                    tangent_np = np.expand_dims(tangent_np, axis=-1)
+                if extra_dim == 2:
+                    tangent_np = np.expand_dims(tangent_np, axis=-1)
+
+                ip_autodiff = np.sum(x_grad_np * tangent_np)
+                err = abs(ip_autodiff - ip_numerical)
+                # print("autodiff:", ip_autodiff)
```
erizmr, line 168: nits: remove print?

python/taichi/ad/_ad.py
```diff
@@ -85,14 +103,133 @@ def __exit__(self, _type, value, tb):
...
+                else:
+                    re = err / (abs(ip_autodiff) + 1e-20)
+                re_range.append(re)
+
+                if err * 100 < abs(ip_autodiff):
+                    check_pass = True
+                    break
+
+            self.result[i] = check_pass
+            plt.loglog(np.array(self.eps_range[:len(re_range)]),
```
erizmr, line 181: Maybe we can print the stats out without depending on the matplotlib?

erizmr (commented):
Thanks!!

python/taichi/ad/_ad.py
PhrygianGates: Yes, because now we also record write_float and write_int kernels

python/taichi/ad/_ad.py
PhrygianGates: I see

python/taichi/ad/_ad.py
PhrygianGates: Yes, it will automatically convert taichi type to corresponding numpy type

tests/python/test_ad_grad_check.py
```diff
@@ -0,0 +1,44 @@
...
+    x2[0, 0] = ti.Vector([1, 2, 3])
+    x2[0, 1] = ti.Vector([4, 5, 6])
+    x2[1, 0] = ti.Vector([7, 8, 9])
+    x2[1, 1] = ti.Vector([10, 11, 12])
+
+    @ti.kernel
+    def compute_y2():
+        y2[None] += x2[0, 0][0] + x2[1, 0][1] + x2[1, 1][2]
+
+    t1 = ti.ad.Tape(y1, grad_check=[x1])
```
ailzhang, line 34: nit: why do we use explicit `__enter__` and `__exit__` here?

tests/python/test_ad_grad_check.py
```diff
@@ -0,0 +1,44 @@
+import numpy as np
+import pytest
+
+import taichi as ti
+from tests import test_utils
+
+
+@test_utils.test()
+def test_grad_check():
```
ailzhang, line 9: btw ia nice followup is that we add tests for major ops (which support AD) so that we make sure no regression is happening.

tests/python/test_ad_grad_check.py
```diff
@@ -0,0 +1,44 @@
...
+    ti.init(default_fp=ti.f64)
```
ailzhang, line 10: moving into `test_utils.test()`?

tests/python/test_ad_grad_check.py
```diff
@@ -0,0 +1,44 @@
...
+
+    x1 = ti.field(dtype=float, shape=(2, 2), needs_grad=True)
+    y1 = ti.field(dtype=float, shape=(), needs_grad=True)
+
+    x1.from_numpy(np.array([[1, 2], [3, 4]]))
+
+    @ti.kernel
+    def compute_y1():
+        ti.loop_config(serialize=True)
```
ailzhang, line 19: is `serialize=True` required? ;)

python/taichi/ad/_ad.py
```diff
@@ -4,19 +4,28 @@
...
 from taichi import _snode
 
 
 class Tape:
-    def __init__(self, loss=None, clear_gradients=True, validation=False):
+    def __init__(self,
+                 loss=None,
+                 clear_gradients=True,
+                 validation=False,
+                 grad_check=None):
```
ailzhang, line 28: nit: need a docstring for grad_check and small handling when it's not a list.

python/taichi/ad/_ad.py
```diff
@@ -91,8 +105,112 @@ def grad(self):
...
+                print(i, "grad check pass")
+
+        if all(self.result):
+            print("all grad check pass")
+
+        self.reset(mode="restore")
+        for func, args in self.calls:
+            func(*args)
+
+    def reset(self, mode):
```
ailzhang, line 185: OOC: is this user facing?

ailzhang: btw it's not intuitive to me what is a "save" mode reset and a "restore" mode reset :P

python/taichi/ad/_ad.py
```diff
@@ -91,8 +105,112 @@ def grad(self):
...
         assert not self.gradient_evaluated, "Gradients of grad can be evaluated only once."
         for func, args in reversed(self.calls):
-            func.grad(*args)
+            if hasattr(func, 'grad'):
+                func.grad(*args)
         self.gradient_evaluated = True
+        if self.grad_check:
+            self.check_grad()
+
+    def check_grad(self):
```
ailzhang, line 114: nit: maybe prefix_ with `_`?

python/taichi/ad/_ad.py
```diff
@@ -55,6 +64,11 @@ def __init__(self, loss=None, clear_gradients=True, validation=False):
                 Warning)
         self.eval_on_exit = loss is not None
         self.loss = loss
+        self.grad_check = grad_check
```
ailzhang, line 67: Nit: IMHO it's better to put all grad_check logic in a separate class instead of putting them inside `Tape` class.

python/taichi/ad/_ad.py
```diff
@@ -91,8 +105,112 @@ def grad(self):
...
+                re_range.append(re)
+
+                if err * 100 < abs(ip_autodiff):
+                    check_pass = True
+                    break
+
+            self.result[i] = check_pass
+
+            if not check_pass:
+                print(i, "relative error:", re_range)
```
ailzhang, line 174: hmmm this error message isn't very informative

PR #6173, opened 2022-09-27
[bug] Perform fflush when synchronizing

lin-hitonami:
https://github.com/taichi-dev/taichi/pull/6166
https://github.com/taichi-dev/taichi/actions/runs/3133143633/jobs/5086205296
`fflush` is required on CUDA backend as well.


lin-hitonami:
https://github.com/taichi-dev/taichi/actions/runs/3133530904/jobs/5087008842 It seems that python fails to capture the message even if it is fflushed...

PR #6184, opened 2022-09-27
[aot] Device capability refactorization

PENGUINLIONG:
This PR refactorized `DeviceCapability` control logics in Taichi. `DeviceCapability` controls the compilation of kernels and the execution behavior on device. The current implementation collect capabilities from the device interface but it blocks extension of AOT code generation because it depends on the currently attached compute device whether to enable a set of feature during compilation. AOT users, however, demands an interface to control the availability of `DeviceCapability`.

This PR does several things and it seems large:
- Removed runtime capability form the list of caps, see code comments for details of design decisions;
- Encapsulate the capability settings in `Device` in `DeviceCapabilityConfig`;
- Replace any reference to `Device` with `DeviceCapabilityConfig` in codegens;
- Removed `TargetDevice` for in the SPIR-V codegen;
- Adapt backends (basically gfx backends) to this change.

Python and C-API users will be able to configure `DeviceCapability`s in future PRs.

### DIFF
--- a/python/taichi/examples/simulation/fractal.py
+++ b/python/taichi/examples/simulation/fractal.py
@@ -1,6 +1,6 @@
 import taichi as ti
 
-ti.init(arch=ti.vulkan)
+ti.init(arch=ti.gpu)
 
 n = 320
 pixels = ti.field(dtype=float, shape=(n * 2, n))
--- a/python/taichi/examples/simulation/fractal.py
+++ b/python/taichi/examples/simulation/fractal.py
@@ -1,6 +1,6 @@
 import taichi as ti
 
-ti.init(arch=ti.gpu)
+ti.init(arch=ti.vulkan)
 
 n = 320
 pixels = ti.field(dtype=float, shape=(n * 2, n))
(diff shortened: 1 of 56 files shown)


python/taichi/examples/simulation/fractal.py
```diff
@@ -1,6 +1,6 @@
 import taichi as ti
 
-ti.init(arch=ti.gpu)
+ti.init(arch=ti.vulkan)
```
ailzhang, line 3: accidental change?

taichi/rhi/vulkan/vulkan_device.cpp
```diff
@@ -1329,6 +1329,7 @@ struct VulkanDevice::ThreadLocalStreams {
 VulkanDevice::VulkanDevice()
     : compute_streams_(std::make_unique<ThreadLocalStreams>()),
       graphics_streams_(std::make_unique<ThreadLocalStreams>()) {
+  caps_.set(DeviceCapability::spirv_version, 0x10000);
```
ailzhang, line 1332: Is this newly enforced btw?

ailzhang (approved):
Awesome!!

taichi/rhi/vulkan/vulkan_device.cpp
PENGUINLIONG: that's the minimum requirement. That capacity is required for the backend to work iirc

python/taichi/examples/simulation/fractal.py
PENGUINLIONG: If

k-ye:
Awesome!

PR #6185, opened 2022-09-27
[example] Request to add my code into examples

JiaoLuhuai:
I added my code for ti.example. You can see details of my example here
https://forum.taichi.graphics/t/topic/2924

This is a 1D simulation of "two-stream instability" in Plasma Physicis.


python/taichi/examples/simulation/pic88.py
```diff
@@ -0,0 +1,93 @@
+# Authored by Luhuai Jiao
+# This is a 1D simulation of "two-stream instability" in Plasma Physicis.
+# Some settings of grids and particles are from "Introduction to Computational Plasma Physics"(《计算等离子体物理导论》，谢华生著)
```
neozhaoliang, line 3: We should avoid use mixed English and Chinese here, you can use the ISBN number instead.

python/taichi/examples/simulation/pic88.py
```diff
@@ -0,0 +1,93 @@
+# Authored by Luhuai Jiao
+# This is a 1D simulation of "two-stream instability" in Plasma Physicis.
+# Some settings are from "Introduction to Computational Plasma Physics"(ISBN: 9787030563675)
+
```
JiaoLuhuai, line 4: I have used ISBN instead of Chinese here.

python/taichi/examples/simulation/pic88.py
JiaoLuhuai: Thanks for your advice! There is no Chinese in the new file now.

python/taichi/examples/simulation/pic88.py
```diff
@@ -0,0 +1,93 @@
...
+vb = 1.0  # beam-velocity, one is vb, the other is -vb
+vt = 0.3  # thermal velocity
+wp = 1  # Plasma frequence
+qm = -1  # charge-mass ratio
+q = wp * wp / (qm * np / L)  # charge of a particle
+rho_back = -q * np / L  # background charge density
+dx = L / ng  # grid spacing
+inv_dx = 1.0 / dx
+x = ti.Vector.field(1, ti.f32, np)  # position
+v = ti.Vector.field(1, ti.f32, np)  # velocity
```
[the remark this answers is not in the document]
JiaoLuhuai: Thanks for your advice! I use `ti.Vector.field` here so that we can expand this simulation to 2D/3D easily

neozhaoliang (approved):
LGTM!

neozhaoliang:
@JiaoLuhuai What does the file name "pic88" mean here? Shall we use a more self-explained name?

JiaoLuhuai:
> 



> @JiaoLuhuai What does the file name "pic88" mean here? Shall we use a more self-explained name?

Good idea! I renamed it "two_stream_instablity"

neozhaoliang (approved):
Fine!

ailzhang (approved):
Thanks!

PR #6192, opened 2022-09-28
[llvm] [refactor] Merge create_call and call

lin-hitonami:
Issue: #3382
Currently `create_call` in `TaskCodeGenLLVM` and `call` in `LLVMModuleBuilder` do the same thing so they should be merged.


ailzhang (approved):
Nice!

PR #6194, opened 2022-09-28
[Lang] MatrixField refactor 9/n: Allow dynamic index of matrix field when real_matrix=True

strongoier:
Issue: #5959

### Brief Summary

This PR makes the rightmost path of #5959 running end-to-end.

### DIFF
--- a/tests/python/test_matrix.py
+++ b/tests/python/test_matrix.py
@@ -193,10 +193,7 @@ def test_matrix_non_constant_index_numpy():
     assert v[3][9] == 9
 
 
-@test_utils.test(require=ti.extension.dynamic_index,
-                 dynamic_index=True,
-                 debug=True)
-def test_matrix_non_constant_index():
+def _test_matrix_non_constant_index():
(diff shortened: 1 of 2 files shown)


tests/python/test_matrix.py
```diff
@@ -248,6 +245,20 @@ def func4(k: ti.i32):
...
 
+@test_utils.test(require=ti.extension.dynamic_index,
+                 dynamic_index=True,
+                 debug=True)
+def test_matrix_non_constant_index():
+    _test_matrix_non_constant_index()
+
+
+@test_utils.test(require=ti.extension.dynamic_index,
+                 real_matrix=True,
```
jim19930609, line 256: Turn on `dynamic_index=True` here?

strongoier: Actually we don't need to turn on that, because this code path doesn't rely on that option. That option is mainly to control alloca of local matrix before we have real matrix.

jim19930609 (approved):

Issue #6190, opened 2022-09-28
ti.reset() should not have side effect from previous runs


ailzhang:
Brief context:
Currently when taichi compiler initiates, there're two major source of compiler state & resources. 
- most states & resource are held by `Program` thus they can be destroyed/initialized properly with Program class. 
- some of them live as static resources in the library and these are persistent across `ti.reset()`.  Examples are opengl initialization, vulkan instance creation and so on. 

These global states are problematic and force us to putting hacks around the codebase to workaround the lifetime management of different resources. 

This issue is used to track the progress of making taichi fresh across multiple `ti.reset()`. 
TODO: convert this to an RFC when we have a better refactor plan.

PR #6198, opened 2022-09-29
[opengl] Split gles backend and enable tests

ailzhang:
* __->__ #6198
* #6197

related: #6190

This PR is the minimal changeset so that we can enable tests for the
newly added gles backend.

It does a few things:
- Remove `use_gles` in `ti.init()` and `CompileConfig`, it's now just a
new backend `ti.gles`.
- `ti.reset()` now destructs glfw resources created in the global scope.
Note in the future we'd like these to be managed by `Program` but
that'll make this PR too large to review.
- Properly enable testing for gles backend and skip when necessary. Most of them are just
the same as ones we skipped for opengl backend.

There're also a few followups that I'll send after this PR since we'd
want those to be tested against both Opengl and Gles backends:
- split gles and opengl initialization
- split opengl initialization and destruction
- maybe also cleanup other backends


taichi/program/program.cpp
```diff
@@ -225,7 +232,8 @@ void Program::synchronize() {
   if (arch_uses_llvm(this_thread_config().arch) ||
       this_thread_config().arch == Arch::metal ||
       this_thread_config().arch == Arch::vulkan ||
-      this_thread_config().arch == Arch::opengl) {
+      this_thread_config().arch == Arch::opengl ||
+      this_thread_config().arch == Arch::gles) {
     program_impl_->synchronize();
```
bobcao3, line 237: What does this `if` condition do? It seems to cover everything?

taichi/program/program.cpp
```diff
@@ -461,7 +473,8 @@ std::size_t Program::get_snode_num_dynamically_allocated(SNode *snode) {
   TI_ASSERT(arch_uses_llvm(this_thread_config().arch) ||
             this_thread_config().arch == Arch::metal ||
             this_thread_config().arch == Arch::vulkan ||
-            this_thread_config().arch == Arch::opengl);
+            this_thread_config().arch == Arch::opengl ||
+            this_thread_config().arch == Arch::gles);
```
bobcao3, line 477: Same as above, do we need this assert?

taichi/rhi/opengl/opengl_api.cpp
```diff
@@ -28,13 +28,24 @@ static void glfw_error_callback(int code, const char *description) {
   TI_WARN("GLFW Error {}: {}", code, description);
 }
 
-bool initialize_opengl(bool use_gles, bool error_tolerance) {
-  static std::optional<bool> supported;  // std::nullopt
+bool initialize_opengl(bool use_gles, bool error_tolerance, bool reset) {
+  // TODO: properly split opengl & gles initialization in a followup PR.
+  // TODO: properly split initialization and reset in a followup PR.
+  static std::optional<bool> supported[2];  // std::nullopt
```
bobcao3, line 34: I'd prefer making it two values and use a pointer:

```
static std::optional<bool> gles_supported;
static std::optional<bool> gl_supported;

...

std::optional<bool> *supported = use_gles ? gles_supported : gl_supported;

...
if (supported->has_value())
```

bobcao3: The underlying code generated is going to be basically entirely the same

taichi/rhi/opengl/opengl_api.cpp
```diff
@@ -160,7 +171,7 @@ bool initialize_opengl(bool use_gles, bool error_tolerance) {
   if (!opengl_version) {
     if (error_tolerance) {
       TI_WARN("Can not create OpenGL context");
-      supported = std::make_optional<bool>(false);
+      supported[(int)use_gles] = std::make_optional<bool>(false);
```
bobcao3, line 174: Is there a reason why this one uses `(int)use_gles` instead of `idx`?

taichi/runtime/program_impls/opengl/opengl_program.cpp
```diff
@@ -116,4 +116,13 @@ const std::unique_ptr<gfx::CacheManager>
   return cache_manager_;
 }
 
+void OpenglProgramImpl::finalize() {
+  runtime_.reset();
+  device_.reset();
+  opengl::initialize_opengl(false, false, true);
```
bobcao3, line 122: Nit
```suggestion
  opengl::initialize_opengl(/*use_gles=*/ false, /*error_tolerance=*/ false, /*reset=*/ true);
```

taichi/transforms/offload.cpp
```diff
@@ -112,7 +112,8 @@ class Offloader {
           offloaded->const_end = true;
           offloaded->end_value = val->val.val_int32();
         } else {
-          if ((arch == Arch::opengl || arch == Arch::vulkan) &&
+          if ((arch == Arch::opengl || arch == Arch::vulkan ||
+               arch == Arch::gles) &&
```
bobcao3, line 116: Just noiticing but shouldn't `dx11` also be here? Maybe this behavior should be covered under CompileConfig or a new struct called `CompilationStrategy`, and we will ask the backend (i.e. Program) to set a default set of parameters instead of putting `if` cases everywhere

bobcao3: ^ this should be put into a seperate issue

ailzhang:
@bobcao3 This PR was more like a proof of concept that I haven't got time to properly refactor it for merging... 
Thanks a lot for the review comments tho! Will make sure they're resolved in the rework :D

PR #6200, opened 2022-09-29
[Doc] Update debugging.md

neozhaoliang:


docs/lang/articles/debug/debugging.md
```diff
@@ -173,7 +179,7 @@ def test() -> ti.i32:
 print(test())
 ```
 
-The code snippet above raises a `TaichiAssertionError`, indicating that you are trying to access a field with improper indices.
+The code snippet above raises a `TaichiAssertionError`, because we are trying to access element of indices `[0, 73]` in the field `f`, while the second dimension of `f` is only 32.
```
Olinaaaloompa, line 182:
```suggestion
The code snippet above raises a `TaichiAssertionError` because you are trying to access elements from a field of shape (32, 32) with indices in the range of `[0, 73]`.
```

docs/lang/articles/debug/debugging.md
```diff
@@ -203,7 +209,7 @@ def do_sqrt_all():
         x[i] = ti.sqrt(x[i])
 ```
 
-When you are done with debugging, set `debug=False`, and then the program ignores the subsequent `assert` statements and avoid additional runtime overhead.
+When you are done with debugging, set `debug=False`, and then the program ignores all `assert` statements in the Taichi scope hence avoid additional runtime overhead.
```
Olinaaaloompa, line 212:
```suggestion
When you are done with debugging, set `debug=False`. Then, the program ignores all `assert` statements in the Taichi scope, which can avoid additional runtime overhead.
```

Olinaaaloompa (changes requested):

docs/lang/articles/debug/debugging.md
neozhaoliang: with indices `[0, 73]` is OK, it's not in the range of `[0, 73]`.

Olinaaaloompa: Updated :)

Olinaaaloompa (approved):
LGTM

Issue #6204, opened 2022-09-29
Autodiff failed on cases where the condition of a if depends on the for loop index


erizmr:
**Describe the bug**
Autodiff failed on cases where the condition of a if depends on the for loop index (on cuda backend).

**To Reproduce**

```py
import taichi as ti
ti.init(arch=ti.cuda)

scalar = lambda: ti.field(dtype=ti.f32)
vec = lambda: ti.Vector.field(3, dtype=ti.f32)

pos=vec()
F=vec()
f_bend=scalar()
loss_n=scalar()
ti.root.dense(ti.ij,(10, 10)).place(pos, F)
ti.root.dense(ti.i,1).place(f_bend)
ti.root.place(loss_n)
ti.root.lazy_grad()


@ti.kernel
def simulation(t:ti.i32):
    for i,j in pos:
        coord = ti.Vector([i,j])
        for n in range(12): 
            f = ti.Vector([0.0,0.0,0.0])
            if n < 4:
                f = ti.Vector([1.0, 1.0, 1.0])
            else:
                f = f_bend[0] * pos[coord]
            F[coord] += f
        pos[coord] += 1.0 * t


with ti.ad.Tape(loss=loss_n):
    simulation(5)
 
```

**Log/Screenshots**
Please post the **full log** of the program (instead of just a few lines around the error message, unless the log is > 1000 lines). This will help us diagnose what's happening. For example:

```
[Taichi] version 1.1.4, llvm 10.0.0, commit 50e5e3fd, linux, python 3.8.10
[Taichi] Starting on arch=cuda
[W 09/29/22 16:33:33.523 1504275] [llvm_offline_cache.cpp:load_meta_data@138] Lock /home/mingrui/.cache/taichi/ticache/llvm/metadata.lock failed
[E 09/29/22 16:33:34.432 1504275] [cuda_driver.h:operator()@87] CUDA Error CUDA_ERROR_ILLEGAL_ADDRESS: an illegal memory access was encountered while calling stream_synchronize (cuStreamSynchronize)
...
```

**Additional comments**
Related question on forum:
https://forum.taichi.graphics/t/topic/3138

erizmr:
Fixed by #6207

PR #6207, opened 2022-09-29
[autodiff] Store if condition in adstack

erizmr:
Issue: #6204 

### Brief Summary
Autodiff failed on cases where the condition of a if depends on the for loop index. This PR makes the if condition stored in adstack to handle these cases.


ailzhang (approved):
Thanks!

PR #6208, opened 2022-09-30
[spirv] Generate OpBitFieldUExtract for BitExtractStmt

strongoier:
Issue: https://github.com/taichi-dev/taichi/pull/6141#issuecomment-1255339288, #6134

### Brief Summary

In SPIR-V there's actually `OpBitFieldUExtract` which does the job of `BitExtractStmt` perfectly, so let's stop demoting `BitExtractStmt` and let codegen handle it with the best instruction(s).

### DIFF
--- a/taichi/transforms/demote_operations.cpp
+++ b/taichi/transforms/demote_operations.cpp
@@ -61,25 +61,6 @@ class DemoteOperations : public BasicStmtVisitor {
     return floor;
   }
 
-  void visit(BitExtractStmt *stmt) override {
-    // @ti.func
-    // def bit_extract(input, begin, end):
-    //   return (input >> begin) & ((1 << (end - begin)) - 1)
-    VecStatement statements;
-    auto begin = statements.push_back<ConstStmt>(
-        TypedConstant(stmt->input->ret_type, stmt->bit_begin));
-    auto input_sar_begin = statements.push_back<BinaryOpStmt>(
-        BinaryOpType::bit_sar, stmt->input, begin);
-    auto mask = statements.push_back<ConstStmt>(TypedConstant(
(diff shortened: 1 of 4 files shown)


taichi/transforms/demote_operations.cpp
```diff
@@ -61,25 +61,6 @@ class DemoteOperations : public BasicStmtVisitor {
     return floor;
   }
 
-  void visit(BitExtractStmt *stmt) override {
```
bobcao3, line 64: I wonder in our definition does BitExtractStmt does sign extension? Does it depend on the input type? If so then we should add a case to detect input types and choose one between `OpBitFieldUExtract` and `OpBitFieldSExtract`

bobcao3:
Another thing, bit instructions are not builtin, they are part of the [extension](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_bit_instructions.html). We may need to add an DeviceCapability to detect and optionally enable this

taichi/transforms/demote_operations.cpp
strongoier: Nope. `BitExtractStmt` doesn't perform sign extension.

bobcao3 (approved):
LGTM! Wonder whether this improves performance for addressing heavy situations in Vulkan

strongoier:
> LGTM! Wonder whether this improves performance for addressing heavy situations in Vulkan

I tried to construct a few cases but none of them could make `BitExtractStmt` itself a bottleneck (it is only part of addressing and already fast enough), so for now no performance gain has been observed. But I think the change is still beneficial once we use `BitExtractStmt` in other scenarios in the future.

jim19930609 (approved):

PR #6213, opened 2022-09-30
[lang] Set ret_type for AtomicOpStmt

ailzhang:
Also added a test in ir builder test, not really related to this PR but thought it was good to have more test coverage for it.

Issue: #


jim19930609 (approved):
thanks!

Issue #5167, opened 2022-06-14
Remove `DataType` and consistently use `const Type *`


strongoier:
Currently, types in Taichi are created and owned by:
https://github.com/taichi-dev/taichi/blob/77ec1a03d810e90886e5d55c22294cd6cd230ea4/taichi/ir/type_factory.h#L9

A type with same parameters is only created once, so the pointer to it can serve as its unique identifier, which allows trivial equality check.

Some refactoring tasks:
- [ ] Because a type should not be modified once creation, `const Type *` should be used everywhere.
- [ ] The following class, https://github.com/taichi-dev/taichi/blob/77ec1a03d810e90886e5d55c22294cd6cd230ea4/taichi/ir/type.h#L53-L54, is legacy and should be removed.

Ris-Bali:
Hey, @lin-hitonami are you still working on this ? If not I would like to have a go at solving this issue. Thanks

lin-hitonami:
@Ris-Bali I'm not working on this. Please go ahead if you want to fix the issue.

PR #6223, opened 2022-10-02
[Refactor] [ir] Modifying TypeFactory class functions to be of const type

matthewd28:
Issue: #5167

### Brief Summary
According to the issue #5167, const Type * should be used everywhere because types shouldn't be modified after created.


strongoier:
Hi @matthewd28. Thanks for contributing & welcome to the Taichi community!

Turning `Type *` into `const Type *` will affect all usages of the return value. You need to take care of all of them instead of only those inside `type_factory.h`. You can follow https://docs.taichi-lang.org/docs/dev_install to build taichi locally and see if the compilation succeeds.

ailzhang:
Hey @matthewd28 , are you still working on this PR?

bobcao3:
Closing stale PR

PR #6245, opened 2022-10-08
[autodiff] Add grad check feature and related test

PhrygianGates:
## Motivation 
Currently Taichi uses autodiff as its differentiation system. We compute gradients of variables in a computation procedure through source code transformation, which is to add additional instructions to the IR layer. However, it's hard for users to check whether the autodiff system computes correct results. Thus, we need a grad check system to verify results given by autodiff.
## Proof of concept
We implement grad check in the python frontend, using finite difference method. Users only need to pass variables they want to check to the autodiff tape scope, and then our grad check system will automatically check whether autodiff gives the correct answer, prompting error messages if not. The API can be used in the following way.
```python
import taichi as ti
import numpy as np

ti.init(ti.gpu, default_fp=ti.f64)

# field with shape
x = ti.field(dtype=float, shape=(2, 2), needs_grad=True)
y = ti.field(dtype=float, shape=(), needs_grad=True)

x.from_numpy(np.array([[1, 2], [3, 4]]))

@ti.kernel
def compute_y():
    for i, j in ti.ndrange(2, 2):
        y[None] += ti.cos(x[i, j])

with ti.ad.Tape(y, grad_check=[x]):
    compute_y()
```
## Implementation
Here I list several key points for the implementation.
### Finite difference method
We imitate what JAX does in their grad check system.
Suppose we have a function $f$ ,  $y = f(\vec{x})$ . By math we know the following equation holds,
$$\frac{f(\vec{x}+\epsilon\vec{\delta})-f(\vec{x}-\epsilon\vec{\delta})}{2\epsilon} \approx \frac{\partial f}{\partial \vec{x}} \cdot \vec{\delta}$$
where $\epsilon$ is a small scalar, $\vec{\delta}$ is a small random vector with the same dimension as $\vec{x}$ .
We compute the left hand side by finite difference method and the right hand side by autodiff.
### Backup all variables
Unlike JAX, we often use Taichi's autodiff system by passing in some global field variables and doing in-place modifications to these variables, which means we cannot extract a pure function $f$ .
However, we need to perturb these input variables and use $f$ to get new output value in order to compute finite differences.
Therefore, we can consider the computing block inside autodiff's tape scope as a black box where input values may be modified each time we run it. The key point is that we need to backup all global field variables before we enter the tape scope. Each time we run the tape scope, we have to restore all variables to their original states.

### DIFF
--- a/python/taichi/ad/_ad.py
+++ b/python/taichi/ad/_ad.py
@@ -91,11 +91,13 @@ class GradChecker:
             self.result[i] = check_pass
 
             if not check_pass:
-                print("variable", i, "has relative error", re_range)
+                print("variable", i, "has relative error", min(re_range),
+                      ", expected relative error 0.01")
             else:
                 print("variable", i, "passes grad check")
 
-        assert all(self.result), "Not all variables pass grad check"
+        assert all(self.result
+                   ), "Grad check failed: Not all variables pass grad check"
 
         restore_all_fields(self.all_fields, self.backups)
         for func, args in self.calls:
--- a/python/taichi/ad/_ad.py
+++ b/python/taichi/ad/_ad.py
@@ -30,8 +30,7 @@ class Tape:
                  validation=False,
                  grad_check=None,
                  watch=None,
-                 eps_range=2.**np.arange(-3, -30, -1).astype(np.float64),
-                 test_mode=False):
+                 eps_range=2.**np.arange(-3, -30, -1).astype(np.float64)):
         """A context manager for reverse mode autodiff :class:`~taichi.ad.Tape`. The
         context manager would catching all of the callings of functions that
         decorated by :func:`~taichi.lang.kernel_impl.kernel` or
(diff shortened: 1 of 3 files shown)


python/taichi/ad/_ad.py
```diff
@@ -7,16 +7,134 @@
...
+                re_range.append(re)
+
+                if err * 100 <= abs(ip_autodiff):
+                    check_pass = True
+                    break
+
+            self.result[i] = check_pass
+
+            if not check_pass:
+                print("variable", i, "has relative error", re_range)
```
erizmr, line 94: nit: can we make it a warning e.g., `Grad check failed: variable xxx has relative error xxx, expected relative error xxx `

erizmr (commented):
Thanks!

python/taichi/ad/_ad.py
PhrygianGates: okay！

erizmr (approved):
LGTM

PR #6249, opened 2022-10-09
[Example] Add karman vortex street example

neozhaoliang:


python/taichi/examples/simulation/karman_vortex_street.py
```diff
@@ -0,0 +1,176 @@
...
+                'my_cmap', colors)
+            vor_img = cm.ScalarMappable(norm=matplotlib.colors.Normalize(
+                vmin=-0.02, vmax=0.02),
+                                        cmap=my_cmap).to_rgba(vor)
+            vel_img = cm.plasma(vel_mag / 0.15)
+            img = np.concatenate((vor_img, vel_img), axis=1)
+            gui.set_image(img)
+            gui.show()
+
+    def pass_to_py(self):
```
ailzhang, line 168: this  method is not used?

ailzhang (approved):
Thanks!

python/taichi/examples/simulation/karman_vortex_street.py
neozhaoliang: Oh, yes it's an unused function. Removed.

PR #6253, opened 2022-10-09
[dx12] Disable DX12 for cpu only test.

python3kgae:
Set TI_WITH_DX12 to OFF for CPU only test.


lin-hitonami:
It seems that the full-ci tag is not working. Maybe a new commit is needed to refresh the tag captured by the github actions.

feisuzhu (approved):

Issue #6133, opened 2022-09-22
Is this memory leak?


RitChan:
**Describe the bug**
Data-oriented class with python objects may cause memory leak.

**To Reproduce**

```py
import gc
import os
import psutil
import taichi as ti


@ti.data_oriented
class X:
    def __init__(self):
        self.py_l = [0] * 5242880  # a list containing 5M integers (5 * 2^20)

    @ti.kernel
    def run(self):
        for i in range(1):
            pass


def get_process_memory():
    process = psutil.Process(os.getpid())
    mem_info = process.memory_info()
    return mem_info.rss


def main():
    ti.init(ti.cpu)
    for i in range(20):
        X().run()
        gc.collect()
        print(f"Iteration {i}, memory usage: {get_process_memory() / 1e6} MB")


if __name__ == '__main__':
    main()

```

**Log/Screenshots**
```
[Taichi] version 1.1.2, llvm 10.0.0, commit f25cf4a2, win, python 3.7.9
[Taichi] Starting on arch=x64
Iteration 0, memory usage: 272.740352 MB
Iteration 1, memory usage: 314.687488 MB
Iteration 2, memory usage: 356.634624 MB
Iteration 3, memory usage: 398.58176 MB
Iteration 4, memory usage: 440.528896 MB
Iteration 5, memory usage: 482.476032 MB
Iteration 6, memory usage: 524.423168 MB
Iteration 7, memory usage: 566.370304 MB
Iteration 8, memory usage: 608.31744 MB
Iteration 9, memory usage: 650.264576 MB
Iteration 10, memory usage: 692.211712 MB
Iteration 11, memory usage: 734.158848 MB
Iteration 12, memory usage: 776.105984 MB
Iteration 13, memory usage: 818.05312 MB
Iteration 14, memory usage: 860.000256 MB
Iteration 15, memory usage: 901.947392 MB
Iteration 16, memory usage: 943.894528 MB
Iteration 17, memory usage: 985.841664 MB
Iteration 18, memory usage: 1027.7888 MB
Iteration 19, memory usage: 1069.735936 MB
```

**Additional comments**
Data-oriented class with only numpy objects is fine.

```Python3
import gc
import os

import numpy as np
import psutil
import taichi as ti


@ti.data_oriented
class X:
    def __init__(self):
        self.np_l = np.zeros(shape=5242880 * 20, dtype="f4")  # 100 M
        # self.py_l = [0] * 5242880

    @ti.kernel
    def run(self):
[...]

PR #6256, opened 2022-10-09
[bug] Prevent ti.kernel from directly caching the passed-in arguments to avoid memory leak

jim19930609:
Issue: #6133


python/taichi/lang/kernel_impl.py
```diff
@@ -388,7 +398,7 @@ def lookup(self, args):
             )
 
         key = self.extract(args)
-        if key not in self.mapping:
+        if key not in self.mapping or self.mapping[key] is None:
```
strongoier, line 401:
```suggestion
        if key not in self.mapping:
```
This change is unnecessary?

strongoier (approved):
LGTM!

PR #6265, opened 2022-10-10
[autodiff] Recover kernel autodiff mode after validation

erizmr:
Recover kernel autodiff mode after validation.


python/taichi/ad/_ad.py
```diff
@@ -205,24 +205,31 @@ def __enter__(self):
...
         self.runtime.target_tape = self
+        return self
 
     def __exit__(self, _type, value, tb):
         self.runtime.target_tape = None
         if self.eval_on_exit:
             self.grad()
+        if self.validation:
+            for f, _, mode_original in self.calls:
+                if mode_original == AutodiffMode.VALIDATION:
```
ailzhang, line 216: OOC: why do we only revert to None when original is validation? (if we don't need revert others, we could choose save this mode for only a subset of kernels or simply assert on they are the same??

erizmr: It is because the validation only happens on forward kernels, i.e. kernels with None autodiff mode.

python/taichi/ad/_ad.py
```diff
@@ -205,24 +206,33 @@ def __enter__(self):
...
         self.runtime.target_tape = None
         if self.eval_on_exit:
             self.grad()
+        if self.validation:
+            for f in self.validation_mode_kernels:
+                f.autodiff_mode = AutodiffMode.NONE
 
     def insert(self, func, args):
+        if self.validation:
+            if func.autodiff_mode == AutodiffMode.VALIDATION:
```
ailzhang, line 221: This still looks a bit weird to me: when did we change this to valiadation mode (logically it should happen here and you save their original modes for restoring right? ;)

erizmr: I was intended to change the mode here. However found the autodiff mode is used when creating kernel before the insert happens LOL.

python/taichi/ad/_ad.py
```diff
@@ -205,24 +206,33 @@ def __enter__(self):
 
         # Attach the context manager to runtime
         self.runtime.target_tape = self
+        return self
```
ailzhang, line 209: why returning self here?

python/taichi/ad/_ad.py
ailzhang: Hmmm IIUC the kernel autodiff mode before entering the tape is not always `None`? Does restoring to `None` make sense? (if so please feel free to add a comment ;)

erizmr: The kernel in validation mode always has a `None` mode before entering the tape. And only the kernels in validation mode  are recorded in the `self.validation_mode_kernels`.

python/taichi/ad/_ad.py
erizmr: This enables `with ti.ad.Tape() as t`, make it possible to call some tape function inside the context manager scope.

python/taichi/ad/_ad.py
```diff
@@ -205,24 +206,32 @@ def __enter__(self):
...
     def __exit__(self, _type, value, tb):
         self.runtime.target_tape = None
         if self.eval_on_exit:
             self.grad()
+        for calls, mode in zip(self.calls, self.modes):
+            calls[0].autodiff_mode = mode
 
     def insert(self, func, args):
+        if self.validation and not impl.get_runtime().grad_replaced:
+            self.modes.append(func.autodiff_mode)
```
ailzhang, line 220: this isn't safe since `self.modes` could be a different length from `self.calls` which breaks your `zip` in `__exit__`.

python/taichi/lang/kernel_impl.py
```diff
@@ -850,12 +842,13 @@ def __call__(self, *args, **kwargs):
             # TODO: if we would like to compute 2nd-order derivatives by forward-on-reverse in a nested context manager fashion,
             # i.e., a `Tape` nested in the `FwdMode`, we can transform the kernels with `mode_original == AutodiffMode.REVERSE` only,
             # to avoid duplicate computation for 1st-order derivatives
-            mode_original = self.autodiff_mode
-            self.autodiff_mode = AutodiffMode.FORWARD
-            self.runtime.fwd_mode_manager.insert(self, mode_original)
-        elif self.runtime.target_tape and self.runtime.target_tape.validation and not self.runtime.grad_replaced:
-            # The autodiff valid check happens on forward kernel
-            self.autodiff_mode = AutodiffMode.VALIDATION
+            self.runtime.fwd_mode_manager.insert(self, args)
```
ailzhang, line 845: IIUC you don't need to pass in args to fwd_mode_manager?

tests/python/test_ad_global_data_access_rule_checker.py
```diff
@@ -134,3 +135,38 @@ def kernel_2_grad():
...
+    loss.grad[None] = 1
+
+    func_calls = []
+    with ti.ad.Tape(loss=loss, validation=True) as t:
+        kernel_1()
+        kernel_2()
+        for f, _ in t.calls:
+            assert f.autodiff_mode == AutodiffMode.VALIDATION
+        func_calls = t.calls
+    for f, _ in func_calls:
```
ailzhang, line 171: nit: can we have another test case that autodiff_mode before entering the context manager is not AutodiffMode.NONE to make sure we restore them correctly?

ailzhang (commented):
LGTM!

ailzhang (approved):
Thanks a ton!!

Issue #6263, opened 2022-10-10
Release the Offline Cache in v1.1.4 (Tracker)


PGZXB:
(Part of #4401)
### Current state
see https://github.com/taichi-dev/taichi/issues/4401#issuecomment-1272283442

### Dest state 
1. Supported offline cache on **CPU, CUDA, Vulkan, OpenGL and Metal**.
2. Enabled the feature by default on the above backends.
3. Fixed bugs I found.
4. Updated the document about offline cache.
5. Test offline cache nightly

#### 1. Supported offline cache on **CPU, CUDA, Vulkan, OpenGL and Metal**.
- [x] CPU, CUDA (llvm)
- [x] Vulkan, OpenGL (gfx)
- [x] Metal
#### 2. Enabled the feature by default on the above backends.
- [x] CPU, CUDA (llvm)
- [x] Vulkan, OpenGL (gfx)
- [x] Metal
#### 3. Fixed bugs I found.
- [x] Fix bug reported by [Build and Test · taichi-dev/taichi@a3eb8d1 (github.com)](https://github.com/taichi-dev/taichi/actions/runs/3215045304)
- [x] Fix asserting failed when registering kernels with same name on Metal
- [x] Serialize missing fields of `metal::TaichiKernelAttributes` and `metal::KernelAttributes`
#### 4. Updated the document about offline cache.
- [x] Update supported backends.
#### 5. Test offline cache nightly
- [x] CPU, CUDA (llvm)
- [x] Vulkan, OpenGL (gfx)
- [x] Metal

PR #6271, opened 2022-10-10
[bug] Fix asserting failed when registering kernels with same name on Metal

PGZXB:
Issue: #6263 
Ignoring kernels which have been registered is required by `metal::CacheManager`.
Bug reported by `TI_WIP_OFFLINE_CACHE=1 python3 run_tests.py -a metal -vr2 --with-offline-cache --rerun-with-offline-cache 1 compare cast numpy`


lin-hitonami:
In which circumstance does a kernel register twice? IIUC this does not happen before implementing the offline cache?

lin-hitonami (approved):
LGTM!

PR #6273, opened 2022-10-10
[autodiff] Skip gdar checking for user defined grad kernel

erizmr:
Skip gdar checking for user defined grad kernel.


ailzhang (approved):
LGTM!

PR #6275, opened 2022-10-10
[Bug] [lang] Make dimension check for GlobalPtrStmt aware of whether it is a cell access

### LINKED ISSUE #6274: An error is reported on a valid ti.is_active() call
Source: https://forum.taichi.graphics/t/topic/3148

In the following code snippet, `test_pf()`(pointer) and `test_dppf()`(dense=>pointer=>pointer) work well, but `test_dpf()`(dense=>pointer) raises an error saying the number of indices is wrong:
```py
import taichi as ti
ti.init(arch=ti.gpu)
pf = ti.field(ti.f64)
dpf = ti.field(ti.f64)
dppf = ti.field(ti.f64)
# pointer case
p = ti.root.pointer(ti.i, 2)
p.place(pf)
# dense => pointer case
d = ti.root.dense(ti.i, 2)
dp = d.pointer(ti.j, 2)
dp.place(dpf)
# dense => pointer => pointer case
dpp = dp.pointer(ti.j, 2)
dpp.place(dppf)

@ti.kernel
def test_pf():
    for i in pf:
        if ti.is_active(p, i):
            print(i, "active")
            
@ti.kernel
def test_dpf():
    for i, j in dpf:
        if ti.is_active(dp, [i,j]):
            print(i,j, "active")

@ti.kernel
def test_dppf():
    for i, j in dppf:
        if ti.is_active(dpp, [i,j]):
            print(i,j, "active")
test_pf()
# test_dpf() # RuntimeError: [$38] S3 has 1 indices. Indexed with 2.
test_dppf()
```

strongoier:
Issue: fix #6274

### Brief Summary

`GlobalPtrStmt` actually has two semantics:
- When it is part of an activation-related `SNodeOpStmt` (`ti.activate(), ti.deactivate(), ti.is_active()`) and the SNode is not of type `dynamic`, the indices are actually targeting a SNode **cell**;
- Otherwise, the indices are targeting the SNode **container**.

The original dimension check in the `type_check()` pass only targets the second case, which results in the false alarm in #6274.

### DIFF
--- a/taichi/ir/frontend_ir.cpp
+++ b/taichi/ir/frontend_ir.cpp
@@ -806,7 +806,9 @@ void SNodeOpExpression::flatten(FlattenContext *ctx) {
     flatten_rvalue(indices[i], ctx);
     indices_stmt.push_back(indices[i]->stmt);
   }
-  auto ptr = ctx->push_back<GlobalPtrStmt>(snode, indices_stmt);
+  auto is_cell_access = SNodeOpStmt::activation_related(op_type) &&
+                        snode->type != SNodeType::dynamic;
+  auto ptr = ctx->push_back<GlobalPtrStmt>(snode, indices_stmt, true, is_cell_access);
   ptr->tb = tb;
   if (op_type == SNodeOpType::is_active) {
     TI_ERROR_IF(snode->type != SNodeType::pointer &&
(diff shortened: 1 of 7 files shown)


ailzhang (approved):
Thanks!

PR #6279, opened 2022-10-10
[Vulkan] Add overflow detection on vulkan when debug=True

lin-hitonami:
Issue: #5932 

### Brief Summary
Added overflow detection of addition, subtraction, multiplication and shift left operators on Vulkan. 

Known bug: 64-bit multiplication overflow detection works on NVIDIA driver 510 with Vulkan version 1.3.194 but does not work on NVIDIA driver 470 with Vulkan version 1.2.175. See https://github.com/taichi-dev/taichi/issues/6303

TODO: Support for OpenGL should be automatically available when the print support of OpenGL is available. Add OpenGL to the test at that time.

### DIFF
--- a/taichi/codegen/spirv/spirv_codegen.cpp
+++ b/taichi/codegen/spirv/spirv_codegen.cpp
@@ -940,21 +940,21 @@ class TaskCodegen : public IRVisitor {
 
     bool debug = device_->get_cap(DeviceCapability::spirv_has_non_semantic_info);
 
-    if (debug && op_type == BinaryOpType::add) {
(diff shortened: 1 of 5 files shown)


ailzhang (approved):
Nice! 
OOC: in the future is it possible to lift these to CHI IR level to save us from implementing it per backend? ;)

lin-hitonami:
> Nice! OOC: in the future is it possible to lift these to CHI IR level to save us from implementing it per backend? ;)

It's possible but I'm afraid we can't use the backend-specific instructions if we lift these to CHI IR level...

lin-hitonami:
Don't merge yet. I've spotted a bug on smul

PR #6290, opened 2022-10-11
[lang] MatrixType bug fix: Fix array indexing with MatrixType-index

jim19930609:
Issue: #5819

### Brief Summary

Taichi supports indexing field/ndarray/... using MatrixTyped index. For example:
```
x = ti.field(shape=(2, 4), ti.i32)

@ti.kernel
def test():
  base = ti.Vector([1, 1])
  x[base] = 100
```

After this PR, the above test case works under `real_matrix` and `real_matrix_scalarize`


tests/python/test_indices.py
```diff
@@ -52,3 +52,35 @@ def prefix_sum():
     prefix_sum()
     for i in range(n):
         assert (val[i] == i + 1)
+
+
+@test_utils.test(arch=[ti.cpu, ti.cuda],
+                 real_matrix=True,
+                 real_matrix_scalarize=True)
+def test_indices_with_matrix():
```
strongoier, line 60: Shall we create a simpler case? It seems a bit heavy for this change.

taichi/ir/frontend_ir.cpp
```diff
@@ -471,13 +471,45 @@ void ExternalTensorExpression::flatten(FlattenContext *ctx) {
...
+  if (indices.exprs.size() == 1 &&
+      indices.exprs[0].get_ret_type()->is<TensorType>()) {
+    flatten_rvalue(indices.exprs[0], ctx);
+
+    auto dtype = indices.exprs[0].get_ret_type()->as<TensorType>();
+    auto alloca_stmt = ctx->push_back<AllocaStmt>(dtype);
+    ctx->push_back<LocalStoreStmt>(alloca_stmt, indices.exprs[0]->stmt);
+
+    for (int i = 0; i < dtype->get_num_elements(); i++) {
+      // TODO: fill in expanded_indices
```
strongoier, line 489: nit: outdated comment?

taichi/ir/frontend_ir.cpp
```diff
@@ -471,13 +471,45 @@ void ExternalTensorExpression::flatten(FlattenContext *ctx) {
...
+
+  // Special case: single index with ret_type=MatrixType, where we have to
+  // expand this MatrixType.
+  if (indices.exprs.size() == 1 &&
+      indices.exprs[0].get_ret_type()->is<TensorType>()) {
+    flatten_rvalue(indices.exprs[0], ctx);
+
+    auto dtype = indices.exprs[0].get_ret_type()->as<TensorType>();
+    auto alloca_stmt = ctx->push_back<AllocaStmt>(dtype);
+    ctx->push_back<LocalStoreStmt>(alloca_stmt, indices.exprs[0]->stmt);
```
strongoier, line 486: We don't need this if expr is already an l-value?

taichi/ir/frontend_ir.cpp
```diff
@@ -667,11 +696,19 @@ void IndexExpression::type_check(CompileConfig *) {
   for (int i = 0; i < indices.exprs.size(); i++) {
     auto &expr = indices.exprs[i];
     TI_ASSERT_TYPE_CHECKED(expr);
-    if (!is_integral(expr->ret_type))
+
+    auto dtype = expr->ret_type;
+    if (indices.exprs.size() == 1 && expr->ret_type->is<TensorType>()) {
```
strongoier, line 701:
```suggestion
    if (indices.exprs.size() == 1 && dtype->is<TensorType>()) {
```

taichi/ir/frontend_ir.cpp
```diff
@@ -667,11 +696,19 @@ void IndexExpression::type_check(CompileConfig *) {
...
+      // Special case: single index with ret_type=MatrixType,
+      // the MatrixType will be expanded during IndexExpression::flatten()
+      dtype = expr->ret_type->cast<TensorType>()->get_element_type();
```
strongoier, line 704:
```suggestion
      dtype = dtype->cast<TensorType>()->get_element_type();
```

taichi/ir/frontend_ir.cpp
```diff
@@ -667,11 +696,19 @@ void IndexExpression::type_check(CompileConfig *) {
   for (int i = 0; i < indices.exprs.size(); i++) {
```
strongoier, line 696: Dimension compatibility checks (before this line) should also be aware of this special case.

python/taichi/lang/impl.py
```diff
@@ -199,10 +199,25 @@ def subscript(value, *_indices, skip_reordered=False, get_ref=False):
                     f"Gradient {_var.get_expr_name()} has not been placed, check whether `needs_grad=True`"
                 )
         field_dim = snode.num_active_indices()
+
         if field_dim != index_dim:
```
strongoier, line 203: Under `field_dim == index_dim` it is still possible that `field_dim == 1` and a 1-element vector is passed in. We need to handle the special case before we compare the dimensions.

PR #6292, opened 2022-10-11
[lang] MatrixType bug fix: Avoid checks for legacy Matrix-class when real_matrix is on

jim19930609:
Issue: #https://github.com/taichi-dev/taichi/issues/5819


jim19930609:
> Is a check still needed to make sure the value has `MatrixType` matching the annotation?

hmmm, good point. I think we still have to check the ret_type once moved to real_matrix. Will fix that

strongoier (approved):
LGTM

PR #6294, opened 2022-10-11
[lang] MatrixType bug fix: Fix MatrixType validations in build_call_if_is_type()

jim19930609:
Issue: #5819

### DIFF
--- a/tests/python/test_literal.py
+++ b/tests/python/test_literal.py
@@ -41,8 +41,7 @@ def test_literal_keywords_error():
         keywords_error()
 
 
-@test_utils.test()
-def test_literal_compound_error():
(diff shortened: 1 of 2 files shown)


tests/python/test_literal.py
```diff
@@ -56,6 +55,16 @@ def expr_error():
         expr_error()
 
 
+@test_utils.test()
+def test_literal_compound_error():
+    _test_literal_compound_error()
+
+
+@test_utils.test(real_matrix=True, real_matrix_scalarize=True)
```
strongoier, line 63: Do we need `real_matrix_scalarize` here?

strongoier (approved):
LGTM

PR #6296, opened 2022-10-12
[bug] Fix duplicate runs with 'run_tests.py --cpp -k' when selecting AOT tests

jim19930609:
Issue: #

2022-10-12 10-59-23 的屏幕截图


ailzhang (approved):
Thanks a lot for fixing this!

PR #6304, opened 2022-10-12
[lang] MatrixType bug fix: Allow dynamic_index=True when real_matrix_scalarize=True

strongoier:
Issue: #5819

### Brief Summary

Under both `real_matrix_scalarize=True` and `dynamic_index=True`, local matrix should go through the old code path for dynamic index (`ArrayType`). This PR does the fix and refactors related tests.

### DIFF
--- a/taichi/transforms/scalarize.cpp
+++ b/taichi/transforms/scalarize.cpp
@@ -378,10 +378,12 @@ class ScalarizePointers : public BasicStmtVisitor {
 
 namespace irpass {
 
-void scalarize(IRNode *root) {
+void scalarize(IRNode *root, const CompileConfig &config) {
   TI_AUTO_PROF;
   Scalarize scalarize_pass(root);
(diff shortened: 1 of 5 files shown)


taichi/transforms/scalarize.cpp
```diff
@@ -378,10 +378,12 @@ class ScalarizePointers : public BasicStmtVisitor {
 
 namespace irpass {
 
-void scalarize(IRNode *root) {
+void scalarize(IRNode *root, const CompileConfig &config) {
```
jim19930609, line 381: nit: another way to obtain `CompileConfig`: `root->kernel->program->this_thread_config()`, so you dont need to change the function signature.

jim19930609 (approved):

taichi/transforms/scalarize.cpp
strongoier: Updated. Thanks!

PR #6286, opened 2022-10-11
[ci] Add end-to-end CI tests for meshtaichi

BillXu2000:
History in #6243 looks too dirty.
wip

PR #6310, opened 2022-10-13
[ci] Add end-to-end CI tests for meshtaichi

BillXu2000:
continued from #6286.
Only branch in main repository can access secrets in workflow.


g1n0st (approved):
LGTM!

PR #6311, opened 2022-10-13
[bug] Fix generating LLVM AOT module for the second time failed.

### LINKED ISSUE #6309: LLVM kernel AOT got skipped for the second time
To repro:
Run the following command twice: 
```
 TAICHI_AOT_FOLDER_PATH=/tmp/aot python tests/cpp/aot/python_scripts/taichi_sparse_test.py
```
You'll notice that after the first run everything is normal (/tmp/aot/metadata.json has all kernel info in it) but after the second run kernel infos disappeared in `metadata.json`, also the `.ll` files for kernels are not generated. 

first time: 
Screenshot from 2022-10-13 10-53-02

second: 
Screenshot from 2022-10-13 10-52-14

cc: @jim19930609 @PGZXB

PGZXB:
Issue: fixes #6309


taichi/runtime/llvm/llvm_offline_cache.cpp
```diff
@@ -266,7 +266,7 @@ void LlvmOfflineCacheFileWriter::dump(const std::string &path,
       TI_ASSERT(mod != nullptr);
       if (format & Format::LL) {
         std::string filename = filename_prefix + ".ll";
-        if (try_lock_with_file(filename)) {  // Not exists
+        if (!merge_with_old || try_lock_with_file(filename)) {
```
ailzhang, line 269: OOC: I took a quick look at how merging works, I wonder is there any specific reason that we need to explicitly load old metadata and overwrite instead of just overwriting without looking at the old ones? Thanks!

ailzhang (commented):
nit: would you mind also adding a test that saves aot module twice to prevent this from being broken accidentally in the future? Thanks!

taichi/runtime/llvm/llvm_offline_cache.cpp
PGZXB: Merging with old metadata is required by the offline cache.

For Offline Cache:
* Merging the new metadata with the old metadata is needed, otherwise we will lose old cached kernels.

see 
https://github.com/taichi-dev/taichi/blob/master/taichi/runtime/program_impls/llvm/llvm_program.cpp#L181-L183

For AOT: not needed and causes the bug.

ailzhang (approved):
As discussed offline we'll refactor llvm aot to be more modular instead of coupling with llvm offline cache in the future, so this PR is good to go once test is added ;) Thanks!

PR #6319, opened 2022-10-13
[Lang] Matrix/Vector refactor: Matrix operations part 1

AD1024:
Issue: #5478

### Brief Summary
Implement matrix/vector operations as standard libraries.


python/taichi/lang/matrix_ops.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@preconditions(square_matrix)
+@func
+def trace(x):
+    shape = static(x.get_shape())
+    result = cast(0, x.element_type())
+    # TODO: fix parallelization
+    loop_config(serialize=True)
+    for i in range(shape[0]):
```
strongoier, line 17:
```suggestion
    for i in static(range(shape[0])):
```
Let's match the old implementation first :-)

python/taichi/lang/matrix_ops.py
```diff
@@ -0,0 +1,50 @@
...
+    # CHI IR Tensor repr is ready stable
+    for i in static(range(shape[0])):
+        result += x[i, i]
+    return result
+
+
+@preconditions(arg_at(0, is_tensor))
+@taichi_scope
+def fill(m, val):
+    # capture reference to m
```
strongoier, line 27: Maybe try
```
@func
def fill(m: ti.template(), val):
```

python/taichi/lang/matrix_ops.py
```diff
@@ -0,0 +1,50 @@
...
+from taichi.lang.util import taichi_scope
+
+
+@preconditions(square_matrix)
+@func
+def trace(x):
+    shape = static(x.get_shape())
+    result = cast(0, x.element_type())
+    # TODO: fix parallelization
+    loop_config(serialize=True)
```
strongoier, line 16: Let's remove this to avoid affecting old code.

tests/python/test_eig.py
```diff
@@ -138,7 +138,7 @@ def eigen_solve():
 
 
 @pytest.mark.parametrize("func", [_test_eig2x2_real, _test_eig2x2_complex])
-@test_utils.test(default_fp=ti.f32, fast_math=False)
+@test_utils.test(default_fp=ti.f32, fast_math=False, dynamic_index=True)
```
strongoier, line 141:
```suggestion
@test_utils.test(default_fp=ti.f32, fast_math=False)
```
This shouldn't be changed.

python/taichi/lang/matrix.py
```diff
@@ -718,11 +732,9 @@ def trace(self):
...
             5
         """
-        assert self.n == self.m
-        _sum = self(0, 0)
-        for i in range(1, self.n):
-            _sum = _sum + self(i, i)
-        return _sum
+        # pylint: disable-msg=C0415
+        from taichi.lang.matrix_ops import trace
+        return trace(self)
```
strongoier, line 737: Will this break python-scope `trace()`?

strongoier (commented):
Tests should be added for `real_matrix=True, real_matrix_scalarize=True`.

python/taichi/lang/matrix_ops_utils.py
```diff
@@ -0,0 +1,58 @@
...
+                arg = args[i]
+            except IndexError:
+                raise
+        do_check(fns, arg, **kwargs)
+        return True, None
+
+    return check
+
+
+def is_tensor(m, msg='not tensor type: {}'):
```
jim19930609, line 45: This is more like "assert_tensor"

python/taichi/lang/matrix.py
```diff
@@ -985,10 +997,9 @@ def fill(self, val):
             >>> A
             [-1, -1, -1, -1]
         """
-        def assign_renamed(x, y):
-            return ops_mod.assign(x, y)
-
-        return self._element_wise_writeback_binary(assign_renamed, val)
+        # pylint: disable=C0415
+        from taichi.lang.matrix_ops import fill
+        return fill(self, val)
```
jim19930609, line 1002: Since we have multiple functions named `fill`, let's explicitly call `matrix_ops.fill()` here

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -717,7 +721,15 @@ def build_Attribute(ctx, node):
             node.ptr = lambda val: append(x.parent(), index, val)
         else:
             build_stmt(ctx, node.value)
-            node.ptr = getattr(node.value.ptr, node.attr)
+            if isinstance(node.value.ptr,
+                          Expr) and not hasattr(node.value.ptr, node.attr):
+                # pylint: disable-msg=C0415
+                from taichi.lang import matrix_ops as tensor_ops
```
jim19930609, line 727: let's put all the `import` on the top. Otherwise it's gonna introduce extra efforts to debug module name conflicts or circular imports.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -515,6 +515,10 @@ def build_Call(ctx, node):
         if ASTTransformer.build_call_if_is_type(ctx, node, args, keywords):
             return node.ptr
 
+        if getattr(node.func, 'call_tensor_op', False):
```
jim19930609, line 518: why dont we directly check if `caller` method exists?

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -717,7 +721,15 @@ def build_Attribute(ctx, node):
...
         else:
             build_stmt(ctx, node.value)
-            node.ptr = getattr(node.value.ptr, node.attr)
+            if isinstance(node.value.ptr,
+                          Expr) and not hasattr(node.value.ptr, node.attr):
+                # pylint: disable-msg=C0415
+                from taichi.lang import matrix_ops as tensor_ops
+                node.ptr = getattr(tensor_ops, node.attr)
+                setattr(node, 'call_tensor_op', True)
+                setattr(node, 'caller', node.value.ptr)
```
jim19930609, line 730: Will `caller` method be overriden once we build another Attribute with the same node?

python/taichi/lang/matrix.py
AD1024: Nice catch. Need to decorate as a `pyfunc`.

python/taichi/lang/ast/ast_transformer.py
AD1024: Currently there will be a circular import error
(`ast_transformer` <- `kernel_impl` <- `matrix_ops` <- `ast_transformer`) if we put it on the top.

python/taichi/lang/ast/ast_transformer.py
AD1024: True!

python/taichi/lang/ast/ast_transformer.py
AD1024: Could you give an example? I think a node should only be built once?

python/taichi/lang/matrix.py
```diff
@@ -500,6 +501,19 @@ def __init__(self,
...
+        if self.ndim == 1:
+            return (self.n, )
+        return (self.n, self.m)
+
+    def element_type(self):
+        if self.dt is None:
+            if self._impl.entries:
+                return getattr(self._impl.entries[0], 'element_type',
+                               lambda: None)()
+            return None
```
strongoier, line 514: Can this part replace the use of `self.dt`?

python/taichi/lang/ast/ast_transformer.py
jim19930609: Ah looks like each "Attribute" corresponds to one "attr", plz ignore this~

jim19930609 (commented):
LGTM. Pass to @strongoier ~

python/taichi/lang/matrix_ops.py
```diff
@@ -0,0 +1,39 @@
...
+    # CHI IR Tensor repr is ready stable
+    for i in static(range(shape[0])):
+        result += x[i, i]
+    return result
+
+
+@preconditions(arg_at(0, assert_tensor))
+@func
+def fill(m: Template(), val):
+    s = static(m.get_shape())
```
strongoier, line 25: super nit: the naming seems a little inconsistent. `trace` uses `x` for the matrix and `shape` for the shape, while `fill` uses `m` for the matrix and `s`  for the shape. Shall we unify them so that future developers can easily find a convention to follow? I suggest using `mat` and `shape`.  WDYT

AD1024: Sounds good. Let's use `mat` for matrices, `vec` for vectors and `shape` for shapes.

strongoier (approved):
LGTM!

PR #6323, opened 2022-10-14
[lang] MatrixType bug fix: Fix array indexing with MatrixType-index

jim19930609:
Issue: #5819

### Brief Summary

Taichi supports indexing field/ndarray/... using MatrixTyped index. For example:
```
x = ti.field(shape=(2, 4), ti.i32)

@ti.kernel
def test():
  base = ti.Vector([1, 1])
  x[base] = 100
```

After this PR, the above test case works under `real_matrix` and `real_matrix_scalarize`


taichi/ir/frontend_ir.h
```diff
@@ -981,6 +981,8 @@ class ASTBuilder {
   void insert_snode_activate(SNode *snode, const ExprGroup &expr_group);
   void insert_snode_deactivate(SNode *snode, const ExprGroup &expr_group);
 
+  std::vector<Expr> flatten_indices_expr(std::vector<Expr> &indices);
```
strongoier, line 984:
```suggestion
  std::vector<Expr> flatten_indices_expr(const std::vector<Expr> &indices);
```

taichi/ir/frontend_ir.cpp
```diff
@@ -1330,6 +1330,66 @@ void ASTBuilder::insert_snode_deactivate(SNode *snode,
                                                expr_group));
 }
 
+std::vector<Expr> ASTBuilder::flatten_indices_expr(std::vector<Expr> &indices) {
```
strongoier, line 1333:
```suggestion
std::vector<Expr> ASTBuilder::flatten_indices_expr(const std::vector<Expr> &indices) {
```

strongoier (approved):
LGTM!

PR #6327, opened 2022-10-14
[build] Allow building test when LLVM is off

ailzhang:
This is not a perfect fix since some of the C++ tests won't run without an available cpu backend but at least this PR let it pass the building phase.

Issue: #


yuanming-hu (approved):
LGTM, thank you!

PR #6333, opened 2022-10-15
[build] Enable google-build-explicit-make-pair check in clang-tidy

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #6335
* #6334
* __->__ #6333
* #6332
* #6331
* #6330

PR #6332, opened 2022-10-15
[build] Enable a few bugprone related rules in clang-tidy

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #6335
* #6334
* #6333
* __->__ #6332
* #6331
* #6330

- bugprone-argument-comment
- bugprone-assert-side-effect
- bugprone-dangling-handle
- bugprone-inaccurate-erase
- bugprone-string-integer-assignment
- bugprone-use-after-move
- bugprone-string-constructor
- bugprone-unused-raii

PR #6331, opened 2022-10-15
[build] Enable modernize-use-override in clang-tidy

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #6335
* #6334
* #6333
* #6332
* __->__ #6331
* #6330

PR #6330, opened 2022-10-15
[ci] Use .clang-tidy for check_static_analyzer job

ailzhang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #6335
* #6334
* #6333
* #6332
* #6331
* __->__ #6330

PR #6334, opened 2022-10-15
[build] Enable google-explicit-constructor check in clang-tidy

ailzhang:
* #6335
* __->__ #6334
* #6333
* #6332
* #6331
* #6330

Note that we have to skip the check for a few widely used constructors
to keep the size of this PR reasonable, will fix those in separate PRs.


YuCrazing (approved):
LGTM!

PR #6336, opened 2022-10-16
[lang] Support structured for loop for RW texture

PENGUINLIONG:
It's inconvenient to pass the width, height, and depth separately into the kernels while these values are known internally from the allocations.


ailzhang (commented):
LGTM(would you mind adding a comment in statements.h about reusing ExternalTensorShapeAlongAxisStmt for textures?)!
will leave this for @strongoier to take another look before merging.

taichi/transforms/lower_ast.cpp
```diff
@@ -248,7 +258,7 @@ class LowerAST : public IRVisitor {
       auto &&new_for = std::make_unique<RangeForStmt>(
           begin, end, std::move(stmt->body), stmt->is_bit_vectorized,
           stmt->num_cpu_threads, stmt->block_dim, stmt->strictly_serialized,
-          /*range_hint=*/fmt::format("arg {}", tensor->arg_id));
+          /*range_hint=*/fmt::format("arg {}", arg_id));
```
strongoier, line 261: Is `arg_id` not given the correct value?

python/taichi/lang/impl.py
```diff
@@ -105,17 +106,19 @@ def expr_init_func(
...
 
 def begin_frontend_struct_for(ast_builder, group, loop_range):
-    if not isinstance(loop_range, (AnyArray, Field, SNode, _Root)):
+    if not isinstance(loop_range,
+                      (AnyArray, Field, SNode, RWTextureAccessor, _Root)):
         raise TypeError(
             f"Cannot loop over the object {type(loop_range)} in Taichi scope. Only Taichi fields (via template) or dense arrays (via types.ndarray) are supported."
         )
-    if group.size() != len(loop_range.shape):
+    if (not isinstance(loop_range, RWTextureAccessor)) and group.size() != len(
```
strongoier, line 114: Why not check shape for `RWTextureAccessor`?

taichi/transforms/lower_ast.cpp
strongoier: Hmm I mean you need to assign the value to `arg_id` so that `RangeForStmt` can get the correct hint string.

python/taichi/lang/impl.py
strongoier: Let's also add a test for the invalid case :-)

taichi/transforms/lower_ast.cpp
PENGUINLIONG: ah, i forgot about it lol

python/taichi/lang/impl.py
PENGUINLIONG: Done. :)))

tests/python/test_texture.py
```diff
@@ -1,3 +1,4 @@
+from ast import Index
```
strongoier, line 1:
```suggestion
```
accidental change?

tests/python/test_texture.py
```diff
@@ -150,3 +151,48 @@ def test_from_to_image():
...
+        for i, j in tex:
+            tex.store(ti.Vector([i, j]), ti.Vector([1.0, 0.0, 0.0, 0.0]))
+
+    has_exception = False
+    try:
+        write(tex)
+        pass
+    except ti.TaichiCompilationError:
+        has_exception = True
+    assert has_exception
```
strongoier, line 198: Regarding test cases for invalid situations, let's follow https://github.com/taichi-dev/taichi/blob/639a154d46f24884681baab7ce3c69a588004ee5/tests/python/test_kernel_arg_errors.py#L42-L45

tests/python/test_texture.py
PENGUINLIONG: thx!

strongoier (approved):
LGTM!

PR #6338, opened 2022-10-16
[lang] Texture format for better expressiveness

PENGUINLIONG:
Take texel format and translate the formats into datatype and channel counts internally for better expressiveness (to tolerate formats like `r11_g11_b10`).


python/taichi/lang/_texture.py
```diff
@@ -109,6 +110,56 @@ def store(self, index, value):
             _ti_core.make_texture_op_expr(_ti_core.TextureOpType.kStore,
                                           self.ptr_expr, args_group))
 
+    @taichi_scope
```
ailzhang, line 113: this method isn't for the current PR IIUC?

PR #6339, opened 2022-10-17
[dx12] Only use llvm to compile dx12.

python3kgae:
Add Dx12ProgramImpl.
Prepare change to launch dx12 kernel like dx11 instead of LlvmRuntimeExecutor.
Because dx12 uses buffer instead of pointer directly.

Issue: #5276


ailzhang (approved):
LGTM!

PR #6352, opened 2022-10-17
[test] Adjust rtol for sparse_linear_solver tests

ailzhang:
This test fails consistently on my machine as atol is too strict, switching to rtol instead.

Issue: #


FantasyVR (approved):
LGTM!

PR #6360, opened 2022-10-18
[Error] Add error message when the number of the loop variables does not match the dimension of the ndrange

lin-hitonami:
Issue: #6351


ailzhang (approved):
Thanks!

Issue #6240, opened 2022-10-08
Add docs and more complicated test case for gdar checker


erizmr:
Add docs and more complicated test case for gdar checker.

ailzhang:
Good to close? @erizmr

PR #6361, opened 2022-10-18
[autodiff] Fix mode capture for validation kernels

erizmr:
Issue: #6240 

Handle cases that there are loops inside the `ti.ad.tape`. In these cases, the `autodiff_mode` has not been recovered to `AutodiffMode.NONE` since `__exit__` of the tape has not been reached.

### DIFF
--- a/python/taichi/ad/_ad.py
+++ b/python/taichi/ad/_ad.py
@@ -218,7 +218,7 @@ class Tape:
     def insert(self, func, args):
         # Kernels with mode `AutodiffMode.NONE` and `AutodiffMode.VALIDATION` are all forward kernels.
         # The difference is there are `assert` for global data access rule check in VALIDATION kernels.
-        assert func.autodiff_mode == AutodiffMode.NONE or func.autodiff_mode == AutodiffMode.VALIDATION, "Inserted funcs should be forward kernels."
+        assert func.autodiff_mode in (AutodiffMode.NONE, AutodiffMode.VALIDATION), "Inserted funcs should be forward kernels."
         self.modes.append(func.autodiff_mode)
         if self.validation:
             func.autodiff_mode = AutodiffMode.VALIDATION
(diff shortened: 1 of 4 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -844,7 +846,9 @@ def __call__(self, *args, **kwargs):
         # Both the class kernels and the plain-function kernels are unified now.
         # In both cases, |self.grad| is another Kernel instance that computes the
         # gradient. For class kernels, args[0] is always the kernel owner.
-        if self.autodiff_mode == AutodiffMode.NONE and self.runtime.target_tape and not self.runtime.grad_replaced:
+
+        # No need to capture grad kernels because they are already binded with their primal kernels
```
ailzhang, line 850:
```suggestion
        # No need to capture grad kernels because they are already bound with their primal kernels
```

python/taichi/ad/_ad.py
```diff
@@ -216,7 +216,9 @@ def __exit__(self, _type, value, tb):
             calls[0].autodiff_mode = mode
 
     def insert(self, func, args):
-        assert func.autodiff_mode == AutodiffMode.NONE, "Inserted funcs should be forward kernels."
+        # Kernels with mode `AutodiffMode.NONE` and `AutodiffMode.VALIDATION` are all forward kernels.
+        # The difference is there are `assert` for global data access rule check in VALIDATION kernels.
+        assert func.autodiff_mode == AutodiffMode.NONE or func.autodiff_mode == AutodiffMode.VALIDATION, "Inserted funcs should be forward kernels."
```
ailzhang, line 221:
```suggestion
        assert func.autodiff_mode in (AutodiffMode.NONE, AutodiffMode.VALIDATION), "Inserted funcs should be forward kernels."
```

python/taichi/lang/kernel_impl.py
```diff
@@ -844,7 +846,9 @@ def __call__(self, *args, **kwargs):
...
+        if self.autodiff_mode != AutodiffMode.REVERSE and self.runtime.target_tape and not self.runtime.grad_replaced:
```
ailzhang, line 851:
```suggestion
        if self.autodiff_mode in (AutodiffMode.NONE, AutodiffMode.VALIDATION) and self.runtime.target_tape and not self.runtime.grad_replaced:
```

ailzhang: Let's be more explicit here in case we add a new autodiff mode in the future (to prevent it being accidentally included ;)

ailzhang (approved):
LGTM!

Issue #6281, opened 2022-10-10
Use logging level to config the "Matrix  too big" warning


neozhaoliang:
When declaring a `ti.Matrix` of > 32 elements, Taichi raises the warning

```
UserWarning: Taichi matrices/vectors with 100x1 > 32 entries are not suggested. Matrices/vectors will be automatically unrolled at compile-time for performance. So the compilation time could be extremely long if the matrix size is too big. You may use a field to store a large matrix like this, e.g.:
    x = ti.field(ti.f32, (100, 1)).
 See https://docs.taichi-lang.org/docs/field#matrix-size for more details.
  File "<stdin>", line 1, in <module>
```
This warning can be suppressed via `ti.Matrix(arr, ..., suppress_warning=True)`, but this is not an intuitive way to do so.

Shall we let the value in `suppress_warning` be decided by the current logging level? For example, if the logging level is above `ti.WARNING` (e.g. `ti.ERROR` or `ti.CRITICAL`), then this warning will be automatically suppressed.

PR #6377, opened 2022-10-19
[Lang] Replace matrix warning param by current logging level

neozhaoliang:
Fix: #6281

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -19,6 +19,7 @@ from taichi.lang.util import (cook_dtype, in_python_scope, python_scope,
                               to_pytorch_type, warning)
 from taichi.types import primitive_types
 from taichi.types.compound_types import CompoundType, TensorType
+from taichi._logging import is_logging_effective
 
 
 def _gen_swizzles(cls):
@@ -397,8 +398,6 @@ class Matrix(TaichiOperations):
     Args:
         arr (Union[list, tuple, np.ndarray]): the initial values of a matrix.
         dt (:mod:`~taichi.types.primitive_types`): the element data type.
-        suppress_warning (bool): whether raise warning or not when the matrix contains more \
-            than 32 elements.
         ndim (int optional): the number of dimensions of the matrix; forced reshape if given.
 
     Example::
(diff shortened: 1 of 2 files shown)


python/taichi/lang/matrix.py
```diff
@@ -482,7 +476,7 @@ def __init__(self,
             assert ndim in (0, 1, 2)
             self.ndim = ndim
 
-        if self.n * self.m > 32 and not suppress_warning:
+        if self.n * self.m > 32 and is_logging_effective('warn'):
```
ailzhang, line 479: IIRC `is_logging_effective` is more about our C++ logging but `warning` below isn't controlled by that. Maybe it makes more sense to think about how to make sure `ti.lang.util.warning` respect our log_level set in `ti.init`? 
In the worst case we can replace the `warning` below with `ti.warn` from C++?

neozhaoliang: I have test the code liek this:
```python
import taichi as ti
ti.init(log_level=ti.ERROR)


@ti.kernel
def make():
    v = ti.Vector([0] * 50)
    print(v[0])

make()
```

The warning msg will be printed if the logging level is set to `ti.ERROR` or `ti.CRITICAL`, and will be hidden is set to `ti.TRACE`. I think this is our desired purpose?

ailzhang: Yep it's the desired behavior but spreading `is_logging_effective` checks over python codebase isn't a clean way to do this ;)

neozhaoliang: That makes sense. I have moved the check to the `warn` function, but I'm afraid this may cause unexpected effects, would you please have a look at this?

python/taichi/lang/matrix.py
```diff
@@ -482,7 +475,7 @@ def __init__(self,
             assert ndim in (0, 1, 2)
             self.ndim = ndim
 
-        if self.n * self.m > 32 and not suppress_warning:
+        if self.n * self.m > 32:
             warning(
```
ailzhang, line 479: OOC: what's the advantage of `warning` over  `ti.warn` in this case? ;)

ailzhang (approved):
discussed with @strongoier offline: ideally we will just use python warning system instead of `ti.warn` in python codebase so this should be fine.

PR #6400, opened 2022-10-21
[bug] Fix flaky test with taichi_sparse_test.cpp

jim19930609:
Issue: #


ailzhang (approved):
Thanks! 
nit: would you mind adding a comment linking to the issue so that we can revert this change when it's fixed?

Issue #6295, opened 2022-10-11
The method from_numpy() not working on Mac when using Vulkan backend 


Linyou:
**Describe the bug**
In the latest night-build release, the method `from_numpy()` is still not working on mac with **Vulkan** backend, but functional on **Metal**. 

**To Reproduce**
```py
import taichi as ti
import numpy as np

ti.init(arch=ti.vulkan)

@ti.kernel
def from_numpy(dst: ti.template(), src: ti.types.ndarray()):
    for I in ti.grouped(src):
        dst[I] = src[I]

def from_numpy_python(dst, src):
    for i in range(dst.shape[0]):
        dst[i] = src[i]

def print_array(name, arr, n):
    print(f"{name}: [", end="")
    for i in range(n):
        if i < n - 1:
            print(arr[i], end=", ")
        else:
            print(arr[i], end="]\n")

a = ti.field(dtype=ti.f32, shape=(10,))
float32 = np.ones((10,)).astype(np.float32)
print("float32: ", float32)

a.from_numpy(float32)
print_array('taichi-numpy', a, 10)

from_numpy(a, float32)
print_array('kernel-numpy', a, 10)

from_numpy_python(a, float32)
print_array('python-numpy', a, 10)
```

**Log/Screenshots**
Using **Vulkan** backend
```
$ python taichi_test.py
[Taichi] version 1.1.4, llvm 10.0.0, commit a3eb8d17, osx, python 3.9.13
[mvk-info] MoltenVK version 1.1.11, supporting Vulkan version 1.1.224.

float32:  [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
taichi-numpy: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
kernel-numpy: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
python-numpy: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
```
Using **Metal** backend
```
[Taichi] version 1.1.4, llvm 10.0.0, commit a3eb8d17, osx, python 3.9.13
[Taichi] Starting on arch=metal
float32:  [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
[...]

PR #6415, opened 2022-10-24
[vulkan] [bug] Stop using the buffer device address feature on macOS

strongoier:
See https://github.com/taichi-dev/taichi/issues/6295#issuecomment-1288522905.


ailzhang (approved):
Thanks!!!

PR #6416, opened 2022-10-24
[error] Warn Linux users about manylinux2014 build on startup

feisuzhu:
Issue: #

### DIFF
--- a/python/taichi/_lib/utils.py
+++ b/python/taichi/_lib/utils.py
@@ -1,5 +1,6 @@
 import os
 import platform
+import re
 import sys
 
 from colorama import Fore, Style
@@ -92,6 +93,12 @@ def print_red_bold(*args, **kwargs):
     print(Style.RESET_ALL, end='')
 
 
+def print_yellow_bold(*args, **kwargs):
+    print(Fore.YELLOW + Style.BRIGHT, end='')
+    print(*args, **kwargs)
+    print(Style.RESET_ALL, end='')
+
+
 def check_exists(src):
     if not os.path.exists(src):
         raise FileNotFoundError(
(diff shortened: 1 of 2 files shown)


python/taichi/_lib/utils.py
```diff
@@ -164,3 +171,77 @@ def _print_taichi_header():
...
+                    print()
+
+                print_yellow_bold(
+                    "You can suppress this warning by setting the environment variable TI_MANYLINUX2014_OK=1."
+                )
+        except Exception:
+            pass
+
+
+warn_restricted_version()
```
ailzhang, line 247: nit: is it better to only show this when `ti.vulkan` is requested in `ti.init()`? I'd assume many users who wish to stay on the old OS don't want to see this everytime taichi starts ;)

ailzhang (approved):
only had a nit about when to show this error message. Thanks a ton!

python/taichi/_lib/utils.py
feisuzhu: I'd like to drive users away from old operating systems haha, if this nagging helps it will be great ;)
For those who just don't care or can't, there is a suppressing switch `export TI_MANYLINUX2014_OK=1`

ailzhang: ok sgtm!

feisuzhu: And, end users can use GGUI with cuda backend I think? I have removed the original messages and left a pointer to the startup message, if it don't show up it could be even more confusing.

ailzhang: Yea you're definitely right! cuda with GGUI could potentially be more confusing to users. :D

PR #6418, opened 2022-10-24
[aot] Rename incomplete to truncated

PENGUINLIONG:
The error code `TI_ERROR_INCOMPLETE` is confusing because it actually means that Taichi has successfully done its work while the data is truncated because the user provided buffer is too short.

Also note that `set_last_error` with >0 error codes ("harmless errors") is allowed by designed. There must be a way to allow users to erase existing errors, otherwise the users can never suppressed those errors even they has successfully recovered from them. The restriction on overriding "hard errors" is thus removed.


ailzhang (approved):
Thanks!

PR #6419, opened 2022-10-24
[aot] Support push_arg for Kernel in C++ wrapper

### LINKED ISSUE #6413: [aot] Support push_back for kernel arguments
There're two main motivation behind this feature request:
1. When the number of argument becomes larger, writing `kernel[10] = bla` is kinda confusing, esp. when you insert an argument in the middle, everything after this argument have to be updated in C++ code. 
2. Currently we support vector/matrix argument via a hack (flattened to a list of argument under the hood), using `push_back` can hide this implementation detail well from users and we are free to change the implementation in the future without breaking existing user code. 

cc: @YuCrazing

ailzhang:
fixes #6413 

### Brief Summary
Note we also changed the way to set vec/matrix args the previous `set` didn't save users from counting index manually. cc: @YuCrazing

### DIFF
--- a/c_api/include/taichi/cpp/taichi.hpp
+++ b/c_api/include/taichi/cpp/taichi.hpp
@@ -507,15 +507,25 @@ class Kernel {
     return at(i);
   }
 
-  // Temporary workaround for setting vec/matrix arguments in a flattened way.
   template <typename T>
-  void set(uint32_t i, const std::vector<T> &v) {
-    if (i + v.size() >= args_.size()) {
-      args_.resize(i + v.size());
-    }
+  void push_arg(const std::vector<T> &v) {
+    int idx = args_.size();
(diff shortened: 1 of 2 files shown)


PENGUINLIONG:
do we need `clear_args` in case some scalar args are updated per frame?

ailzhang:
@PENGUINLIONG good point! (I was once hoping they just update the corresponding arg directly but yea that wasn't ideal. 
Now that I think about it, calling clear on vector should be tolerable perf trade-off and we can work around it if necessary in the future.

PENGUINLIONG (approved):
Then lgtm. Thank you!

YuCrazing (approved):
LGTM! Although partial updating of parameters is not so convenient, we can discuss it in the future.

Issue #6320, opened 2022-10-14
Decrease the size of RuntimeContext to save the stack space


lin-hitonami:
**Concisely describe the proposed feature**
Currently `RuntimeContext` serves as the argument of Taichi kernels and real functions. It stores the information about the Taichi context and real arguments of the kernel. However, its size is 2136 bytes now and it takes up too much stack space, especially when we call the real functions recursively. Every time we call it, we need to allocate 2KB of stack space. Stack space on GPU is very limited, so we need to decrease the size.

**Describe the solution you'd like (if any)**
Currently the argument buffer is arrays with fixed size, and there are many wasted space when the number of arguments is small. Maybe we can allocate the argument buffer dynamically to save the space.

lin-hitonami:
After we allocate RuntimeContext on heap, we don't need to decrease the size anymore.

PR #6420, opened 2022-10-24
[llvm] Allocate RuntimeContext on heap

lin-hitonami:
Issue: #6320 

### Brief Summary
This PR mainly does the following things:
1. Add a `NodeManager *runtime_context_buffer_allocator` to `LLVMRuntime` in runtime.cpp to manage the allocation and recycling of `RuntimeContext`.
2. Add function `allocate_runtime_context` and `recycle_runtime_context` in runtime.cpp to allocate and recycle `RuntimeContext` at runtime.
3. Add function `runtime_context_gc`, refactor `gc_parallel_*`, and add `gc_rc_parallel_*` to do the GC on CPU and CUDA backends.
4. Add `gc_rc` to enum `OffloadedTaskType` to represent the offloaded tasks that perform the GC of the `RuntimeContext`s, and add functions `emit_gc_rc` and `emit_cuda_gc_rc` in the codegens to construct the offloaded task in CPU and CUDA backends.
5. Add `gc_rc` offloaded task to the kernel if there is a real function call in the kernel in `offload` pass.
6. Add `context_set_limit` to `CUDADriver` to set the stack limit.


jim19930609 (approved):
LGTM!

PR #6422, opened 2022-10-25
[Error] Deprecate ndrange with number of the loop variables != the dimension of the ndrange

### LINKED ISSUE #6351: `ti.ndrange` does not check for number of parameters

lin-hitonami:
Issue: fixes #6351

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -1011,8 +1011,11 @@ class ASTTransformer(Builder):
                     'Ndrange for loop with number of the loop variables not equal to '
                     'the dimension of the ndrange is deprecated. '
                     'Please check if the number of arguments of ti.ndrange() is equal to '
-                    'the number of the loop variables.', DeprecationWarning,
-                    ctx.file, node.lineno + ctx.lineno_offset, module="taichi")
+                    'the number of the loop variables.',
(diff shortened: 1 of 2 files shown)


jim19930609 (approved):
Thanks!

Issue #6067, opened 2022-09-15
Outer product between vectors not work


BillXu2000:
**Describe the bug**
Matrix multiplication between n*1 and 1*n vector will cause compilation error.

**To Reproduce**

```py
import taichi as ti
ti.init()

@ti.kernel
def fun():
    a = ti.Vector([0.0, 0.0])
    print(a @ a.transpose())

fun()
```

**Log/Screenshots**

```
{15:04}~/playground:dev ✗ ➭ p test_ra.py
[Taichi] version 1.1.3, llvm 10.0.0, commit 9a35e049, linux, python 3.8.5
[Taichi] Starting on arch=x64
Traceback (most recent call last):
  File "test_ra.py", line 9, in <module>
    fun()
  File "/home/bx2k/git/taichi/python/taichi/lang/kernel_impl.py", line 917, in wrapped
    raise type(e)('\n' + str(e)) from None
taichi.lang.exception.TaichiCompilationError: 
File "test_ra.py", line 7, in fun:
    print(a @ a.transpose())
          ^^^^^^^^^^^^^^^^^
Traceback (most recent call last):
  File "/home/bx2k/git/taichi/python/taichi/lang/ast/ast_transformer_utils.py", line 25, in __call__
    return method(ctx, node)
  File "/home/bx2k/git/taichi/python/taichi/lang/ast/ast_transformer.py", line 738, in build_BinOp
    node.ptr = op(node.left.ptr, node.right.ptr)
  File "/home/bx2k/git/taichi/python/taichi/lang/ast/ast_transformer.py", line 736, in <lambda>
    ast.MatMult: lambda l, r: l @ r,
  File "/home/bx2k/git/taichi/python/taichi/lang/matrix.py", line 564, in __matmul__
    return other.transpose() @ self
  File "/home/bx2k/git/taichi/python/taichi/lang/matrix.py", line 566, in __matmul__
    assert self.m == other.n, f"Dimension mismatch between shapes ({self.n}, {self.m}), ({other.n}, {other.m})"
AssertionError: Dimension mismatch between shapes (2, 1), (2, 1)

```

**Additional comments**
This bug seems to be related to PR #6014 .

strongoier:
cc: @AD1024

AD1024:
Thanks! Just working on a PR

PR #6425, opened 2022-10-25
[Lang] MatrixType refactor part 2: add more ops

AD1024:
Issue: #5478 #5819


python/taichi/lang/matrix_ops_utils.py
```diff
@@ -50,9 +76,64 @@ def assert_tensor(m, msg='not tensor type: {}'):
...
+
+def check_matmul(x, y):
+    assert_tensor(x, f'left hand side is not a matrix: {type(x)}')
+    assert_tensor(y, f'right hand side is not a matrix: {type(y)}')
+    x_shape = x.get_shape()
+    y_shape = y.get_shape()
+    if len(x_shape) == 1:
+        if len(y_shape) == 1:
+            return True, None
+        if x_shape[0] != y_shape[1]:
```
strongoier, line 134:
```suggestion
        if x_shape[0] != y_shape[0]:
```
Let's pass CI tests - `y_shape[0]` is the number of rows of `y`.

python/taichi/lang/matrix_ops_utils.py
```diff
@@ -50,9 +76,64 @@ def assert_tensor(m, msg='not tensor type: {}'):
...
+            return False, f'dimension mismatch between {x_shape} and {y_shape} for left multiplication'
+    else:
+        if x_shape[0] != y_shape[0]:
```
strongoier, line 137:
```suggestion
        if x_shape[1] != y_shape[0]:
```
`x_shape[1]` is the number of columns of `x`.

python/taichi/lang/matrix_ops.py
```diff
@@ -1,12 +1,305 @@
...
+                 (E(x, i + 1, 2, 4) * E(x, i + 3, 3, 4) -
+                  E(x, i + 3, 2, 4) * E(x, i + 1, 3, 4)) + E(x, i + 3, 1, 4) *
+                 (E(x, i + 1, 2, 4) * E(x, i + 2, 3, 4) -
+                  E(x, i + 2, 2, 4) * E(x, i + 1, 3, 4))))
+        return det
+    # unreachable
+    return None
+
+
+def _vector_transpose(v):
```
strongoier, line 149: As we've already handled matrix multiplication with vectors correctly, is it possible to make vector transpose always return a vector?

python/taichi/lang/matrix_ops_utils.py
```diff
@@ -36,7 +36,33 @@ def check(*args, **kwargs):
...
+    return check
+
+
+def Or(f, g, msg=None):
+    def check(*args, **kwargs):
+        try:
+            do_check([f], *args, **kwargs)
+            return True, None
+        except TaichiCompilationError:
+            try:
```
jim19930609, line 60: The other thing is `try-catch` has a maximum depth, and it's better only use `try-catch` in the outter-most function (`precondition` in your case). The inner functions should directly `raise` an Exception in some manner like: 
```
if xxxx:
    raise ...
```

jim19930609 (commented):
I do think the decorators need another refactor. However, this shouldn't be blocking and I can help with that after all the attributes are finalized.

python/taichi/lang/matrix_ops.py
AD1024: This issue is related to our change here: #6067 
Shall we un-support this operation (we definitely can make it a `v1.outer_product(v2)` though)

python/taichi/lang/matrix_ops_utils.py
```diff
@@ -36,7 +36,33 @@ def check(*args, **kwargs):
...
+def foreach(*fns):
+    def check(args):
+        for x in args:
+            do_check(fns, x)
+        return True, None
+
+    return check
+
+
+def Or(f, g, msg=None):
```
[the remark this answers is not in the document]
AD1024: This sounds good! Thanks!

jim19930609 (commented):
Other parts LGTM.

strongoier (approved):
LGTM!

PR #6427, opened 2022-10-25
[lang] MatrixType bug fix: support star & loop operations over matrix-typed Expr

jim19930609:

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -30,6 +30,20 @@ else:
     from ast import unparse
 
 
+def reshape_list(target_list, target_shape):
+    if len(target_shape) < 2:
+        return target_list
+
+    curr_list = []
+    dim = target_shape.pop()
+    for i, elem in enumerate(target_list):
+        if i % dim == 0:
+            curr_list.append([])
+        curr_list[-1].append(elem)
+
+    return reshape_list(curr_list, target_shape)
+
+
 def boundary_type_cast_warning(expression):
     expr_dtype = expression.ptr.get_ret_type()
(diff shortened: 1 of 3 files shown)


python/taichi/lang/ast/ast_transformer.py
```diff
@@ -279,6 +279,9 @@ def process_generators(ctx, node, now_comp, func, result):
             return func(ctx, node, result)
         with ctx.static_scope_guard():
             _iter = build_stmt(ctx, node.generators[now_comp].iter)
+
+        if impl.current_cfg().real_matrix and isinstance(_iter, impl.Expr):
+            _iter = [Expr(x) for x in ctx.ast_builder.expand_expr([_iter.ptr])]
```
strongoier, line 284: It seems problematic if there's only one `for`.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -30,6 +30,20 @@
     from ast import unparse
 
 
+def reshape_list(flat_list, target_shape):
+    if len(target_shape) < 2:
+        return flat_list
+
+    curr_list = []
+    dim = target_shape.pop()
```
AD1024, line 38: It's internal use so maybe OK: this line will exhaust `target_shape` in a way the caller may not expect.

AD1024 (approved):
Looks good!

python/taichi/lang/ast/ast_transformer.py
jim19930609: Ah thanks for pointing it out. I should provide another entrance to make a copy of `target_shape`

strongoier (approved):
LGTM!

PR #6428, opened 2022-10-25
[lang] MatrixType bug fix: Add support for rows and cols when real_matrix is ON

jim19930609:


tests/python/test_linalg.py
```diff
@@ -304,10 +301,18 @@ def fill():
...
+@test_utils.test(real_matrix=True, real_matrix_scalarize=True)
+def test_init_matrix_from_vectors_matrix_scalarize():
+    _test_init_matrix_from_vectors()
+
+
 # TODO: Remove this once the APIs are obsolete.
-@pytest.mark.filterwarnings('ignore')
-@test_utils.test(arch=get_host_arch_list())
-def test_init_matrix_from_vectors_deprecated():
+def _test_init_matrix_from_vectors_deprecated():
```
strongoier, line 315: It seems that this test is exactly the same as the last one. Shall we only keep one of them?

strongoier (commented):
It seems that #6425 added implementation for `rows()` but didn't enable it. Shall we enable and test that implementation?

python/taichi/lang/matrix_ops.py
```diff
@@ -55,18 +55,25 @@ def _reduce(mat, fun: template()):
...
 def rows(rows):  # pylint: disable=W0621
     if isinstance(rows[0], (Matrix, Expr)):
         shape = rows[0].get_shape()
         assert len(shape) == 1, "Rows must be a list of vectors"
 
+        for i, row in enumerate(rows):
+            ast_builder = get_runtime().prog.current_ast_builder()
+            if current_cfg().real_matrix and isinstance(
+                    row, Expr) and row.is_tensor():
+                rows[i] = [Expr(x) for x in ast_builder.expand_expr([row.ptr])]
```
strongoier, line 69: Is this part necessary? IIRC `for row in rows` should be able to work after #6427.

jim19930609: Thanks!

python/taichi/lang/matrix_ops.py
```diff
@@ -84,27 +84,14 @@ def _rotation2d_matrix(alpha):
...
-            return Matrix([[row[i] for i in range(shape[0])] for row in rows])
-
-        return _rows()
-    if isinstance(rows[0], list):
-
-        @pyfunc
-        def _rows():
-            return Matrix([[x for x in row] for row in rows])
+    @pyfunc
+    def _rows():
```
strongoier, line 91: There's no need to have a inner function `_rows`?

python/taichi/lang/matrix_ops.py
```diff
@@ -84,27 +84,10 @@ def _rotation2d_matrix(alpha):
         foreach(
             Or(assert_vector(),
                assert_list,
-               msg="Cols/rows must be a list of lists, or a list of vectors")),
-        same_shapes))
+               msg="Cols/rows must be a list of lists, or a list of vectors")))
+)
 def rows(rows):  # pylint: disable=W0621
```
strongoier, line 89:
```suggestion
@pyfunc
def rows(rows):  # pylint: disable=W0621
```

python/taichi/lang/matrix_ops_utils.py
```diff
@@ -99,7 +99,23 @@ def check(*args, **kwargs):
...
-    shapes = [x.get_shape() for x in xs]
+    shapes = []
+    for x in xs:
+        if isinstance(x, Matrix):
+            shapes.append(x.get_shape())
+        elif isinstance(x, list):
+            shape = [len(x)]
+            for i in x:
+                if isinstance(i, list):
+                    shape.append(len(i))
```
strongoier, line 110: Hmm will this result in `[3, 4, 4, 4]` for a 3 x 4 list?

jim19930609: Ah thanks!

strongoier (approved):

PR #6433, opened 2022-10-26
[lang] MatrixType bug fix: Demote ret_type for ArgLoadStmt after scalarization

jim19930609:


taichi/transforms/scalarize.cpp
```diff
@@ -47,7 +47,12 @@ class Scalarize : public BasicStmtVisitor {
       auto dest_tensor_type = dest_dtype->template as<TensorType>();
       auto val_tensor_type = val_dtype->template as<TensorType>();
 
-      TI_ASSERT(dest_tensor_type->get_shape() == val_tensor_type->get_shape());
+      // Taichi used to treat VectorType(n) as MatrixType(n, 1),
```
AD1024, line 50: Probably worth discussing how we represent Vectors in CHI IR. Would be better to make the semantics consistent across different levels of IRs.

jim19930609: +1 for further discussion over Vector v.s. 1-dimensional Matrix semantics.

The other thing is that `type_check` probably does not allow `Vector(n) op Matrix(n, 1)`, but `scalarization` pass stays synchronized with the legacy `scalarization` happened in Python, which is why we decided to let `Vector(n) op Matrix(n, 1)` pass.

Let's have a thorough discussion then decide what we should do with this PR.

AD1024 (approved):
LGTM. Thx!

taichi/transforms/scalarize.cpp
AD1024: SGTM :)

taichi/transforms/lower_matrix_ptr.cpp
```diff
@@ -97,6 +97,11 @@ class RemoveMatrixOfPtr : public BasicStmtVisitor {
     modifier_.erase(stmt);
   }
 
+  void visit(ArgLoadStmt *stmt) override {
```
strongoier, line 100: Hmm.. should we move it elsewhere? `RemoveMatrixOfPtr` is for cleaning redundant stmts after `LowerMatrixPtr`.

strongoier (approved):
LGTM!

PR #6435, opened 2022-10-26
[ci] Swtich CPU jobs to use prebuilt LLVM15

ailzhang:
Note the `.bc` files are still compiled with clang-10 but this is fine, we can upgrade step by step

Issue: #


feisuzhu (approved):
LGTM!

tests/cpp/llvm/llvm_offline_cache_test.cpp
```diff
@@ -93,6 +93,9 @@ TEST_P(LlvmOfflineCacheTest, ReadWrite) {
   {
     auto llvm_ctx = std::make_unique<llvm::LLVMContext>();
 
+#ifdef TI_LLVM_15
+    llvm_ctx->setOpaquePointers(false);
```
python3kgae, line 97: Why set opaque pointers false?

PR #6440, opened 2022-10-26
[vulkan] Disable NoSignedWrap decoration on vulkan backend

ailzhang:
Issue: #

### Brief Summary
Here's my guess of source of validation errors in AOT, @dio-exploded would you mind help verify if this fixes your validation error?


PENGUINLIONG (approved):
lgtm. Decorations are ignored by the drivers so you don't need extra capability or vulkan   Extensions for that

PR #6442, opened 2022-10-26
[aot] Add VkDeviceMemory to TiVulkanMemoryInteropInfo

jim19930609:
For memory copy between CUDA and Vulkan devices, we need to fetch the underlying memory address of the `VkDeviceMemory`, so as to perform `cuMemcpyDtoD`. 

Since Vulkan API does not support accessing binded device memory of an existing `VkBuffer`, we have to explicitly export `VkDeviceMemory` to users.

### DIFF
--- a/c_api/src/taichi_vulkan_impl.cpp
+++ b/c_api/src/taichi_vulkan_impl.cpp
@@ -236,9 +236,14 @@ void ti_export_vulkan_memory(TiRuntime runtime,
   VulkanRuntime *runtime2 = ((Runtime *)runtime)->as_vk();
   taichi::lang::DeviceAllocation devalloc = devmem2devalloc(*runtime2, memory);
   vkapi::IVkBuffer buffer = runtime2->get_vk().get_vkbuffer(devalloc);
+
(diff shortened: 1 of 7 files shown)


c_api/src/taichi_vulkan_impl.cpp
```diff
@@ -236,9 +236,14 @@ void ti_export_vulkan_memory(TiRuntime runtime,
   VulkanRuntime *runtime2 = ((Runtime *)runtime)->as_vk();
   taichi::lang::DeviceAllocation devalloc = devmem2devalloc(*runtime2, memory);
   vkapi::IVkBuffer buffer = runtime2->get_vk().get_vkbuffer(devalloc);
+
+  auto tup = runtime2->get_vk().get_vkmemory_offset_size(devalloc);
+  VkDeviceMemory vk_mem = std::get<0>(tup);
```
bobcao3, line 241: I'd prefer the structured binding syntax instead of this. `auto &[a, b, c] = ...`

PENGUINLIONG (approved):
lgtm

PR #6443, opened 2022-10-26
[llvm] Fix broken llvm module on CUDA backend when TI_LLVM_15 is on

ailzhang:
Issue: #

### Brief Summary
I also changed broken module from TI_WARN to TI_ERROR.... not sure if we relied on this or not, let's see....


ailzhang:
Superseded by #6458 as I screwed up git history in this PR.

Issue #6448, opened 2022-10-27
Memory leaks in Taichi AOT runtime (GFX/Vulkan)


k-ye:
**Describe the bug**

I've run Taichi AOT with valgrind, which reported memory leaks in several places:

```
==3169151== 868,128 (288 direct, 867,840 indirect) bytes in 1 blocks are definitely lost in loss record 3,570 of 3,575
==3169151==    at 0x483BE63: operator new(unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==3169151==    by 0x525E254: taichi::lang::gfx::GfxRuntime::register_taichi_kernel(taichi::lang::gfx::GfxRuntime::RegisterParams) (in /libtaichi_c_api.so)
==3169151==    by 0x4D6E8C8: taichi::lang::gfx::KernelImpl::KernelImpl(taichi::lang::gfx::GfxRuntime*, taichi::lang::gfx::GfxRuntime::RegisterParams&&) (in /libtaichi_c_api.so)
==3169151==    by 0x52788A3: taichi::lang::gfx::(anonymous namespace)::AotModuleImpl::make_new_kernel(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (in /libtaichi_c_api.so)
==3169151==    by 0x4B3A193: taichi::lang::aot::Module::get_kernel(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (in /libtaichi_c_api.so)
==3169151==    by 0x5277D09: taichi::lang::gfx::(anonymous namespace)::AotModuleImpl::get_graph(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (in /libtaichi_c_api.so)
==3169151==    by 0x4AE2A0E: AotModule::get_cgraph(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (in /libtaichi_c_api.so)
==3169151==    by 0x4AE4B62: ti_get_aot_module_compute_graph (in /libtaichi_c_api.so)
```

---

```
==3169151== 109,720 (104 direct, 109,616 indirect) bytes in 1 blocks are definitely lost in loss record 3,556 of 3,575
==3169151==    at 0x483BE63: operator new(unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
[...]

PR #6449, opened 2022-10-27
[Bug] Fix memory leak in SPIRV module

k-ye:
Issue: https://github.com/taichi-dev/taichi/issues/6448

### Brief Summary

We need to explicitly destroy the SPIRV module

### DIFF
--- a/taichi/rhi/vulkan/vulkan_device.cpp
+++ b/taichi/rhi/vulkan/vulkan_device.cpp
@@ -300,6 +300,7 @@ void VulkanPipeline::create_descriptor_set_layout(const Params &params) {
                 graphics_pipeline_template_->blend_attachments.end(),
                 default_state);
     }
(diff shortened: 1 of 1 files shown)


taichi/rhi/vulkan/vulkan_device.cpp
```diff
@@ -300,6 +300,7 @@ void VulkanPipeline::create_descriptor_set_layout(const Params &params) {
                 graphics_pipeline_template_->blend_attachments.end(),
                 default_state);
     }
+    spvReflectDestroyShaderModule(&module);
```
k-ye, line 303: corresponds to https://github.com/taichi-dev/taichi/blob/9f96e514aaef2294107d63b0002e5ca55bac56b6/taichi/rhi/vulkan/vulkan_device.cpp#L213-L214

ailzhang (approved):
Thanks!

PR #6452, opened 2022-10-27
[Doc] Update meta.md

neozhaoliang:


docs/lang/articles/advanced/meta.md
```diff
@@ -4,129 +4,111 @@ sidebar_position: 1
...
+sum(f1d)
+sum(f2d)
+sum(f3d)
+sum(g3d)
 ```
 
-:::note
-For sparse fields, the full domain shape will be returned.
-:::
+As can be seen from the code above, the function `sum` works for fields of any shape and any dtype. There's no need to bother about the shape of the field. This is very handy for physical simulations as the same function can be used in both 2D and 3D scenarios.
```
Olinaaaloompa, line 74:
```suggestion

```

docs/lang/articles/advanced/meta.md
```diff
@@ -4,129 +4,111 @@ sidebar_position: 1
...
+
+Note the function `ti.group()` is critial to our dimensionality-independent programming: In general, you will need `d` independent indices to loop over a field of dimension `d`, with one index for each axis. Taichi's `ti.grouped` puts the `d` loop indices into a `d`-dimentional integer vector, and the `for` loop is parallelized for all such multidimensional indices.
```
Olinaaaloompa, line 76:
```suggestion
The function `ti.group()` is critical to dimensionality-independent programming. Normally, you need `d` independent indices to loop over a field of dimension `d`, with each index for an axis. Taichi's `ti.grouped` puts the `d` loop indices into a `d`-dimensional integer vector, and the `for` loop is parallelized for all multidimensional indices.
```

docs/lang/articles/advanced/meta.md
```diff
@@ -4,129 +4,111 @@ sidebar_position: 1
...
 
-## Matrix & vector metadata
+We should mention a difference between Taichi and C++ in template metaprogramming: C++ compilers will generate a version of `sum` for each different type `T`; meanwhile Taichi's compiler **recompiles** the kernel each time it finds an argument of a different type is encounted. In the example above, since the fields are of different shapes, or the same shape but of different dtypes, each of the four calls to `sum` will trigger a compilation:
```
Olinaaaloompa, line 78:
```suggestion
Here comes a major difference between Taichi and C++ in template metaprogramming: C++ compilers generate a version of `sum` for each different type `T`; while Taichi's compiler **recompiles** the kernel each time it encounters an argument of a different type. In the example above, the fields are of different shapes or different dtypes, and each of the four calls to `sum` triggers a compilation:
```

Olinaaaloompa, line 78: Can we just say "encounters a different argument"?

docs/lang/articles/advanced/meta.md
```diff
@@ -4,129 +4,111 @@ sidebar_position: 1
...
 ```
 
+:::
+
+:::note
+The template parameters are inlined into the generated kernel after compilation.
+:::
+
+
 ## Compile-time evaluations
```
Olinaaaloompa, line 104:
```suggestion
### Compile-time evaluations
```

docs/lang/articles/advanced/meta.md
```diff
@@ -4,129 +4,111 @@ sidebar_position: 1
...
 
+
 Using compile-time evaluation allows for some computation to be executed when kernels are instantiated. This helps the compiler to conduct optimization and reduce
```
Olinaaaloompa, line 107:
```suggestion
Using compile-time evaluations allow for some computation to be executed when kernels are instantiated. This helps the compiler to conduct optimization and reduce
```

docs/lang/articles/advanced/meta.md
```diff
@@ -4,129 +4,111 @@ sidebar_position: 1
...
-computational overhead at runtime:
+computational overhead at runtime. Or in other words, we do more at compile time to save effort for runime.
 
 ### Static Scope
+
 `ti.static` is a function which receives one argument. It is a hint for the compiler to evaluate the argument at compile time.
```
Olinaaaloompa, line 112:
```suggestion
The function `ti.static` receives one argument only and serves as a hint for the compiler to evaluate the argument at compile time.
```

docs/lang/articles/advanced/meta.md
```diff
@@ -166,6 +148,18 @@ def func():
...
+## When to use `ti.static` with for loops
+
+The main reason to use `ti.static` with for loops is to unroll the loop to improve runtime performance (see [Compile-time evaluations](#compile-time-evaluations)).
+
+An unrolling verison is often faster because it reduces branch overhead:
+
+1. It saves the instructions for incrementing the loop counter and checking the end-of-loop condition.
+2. It provides more instructions for the compiler to schedule across the whole iterations.
+
+This performance improvement does not come for free: it also costs longer complie time and generates larger binaries. Especially when the number of iterations is large or the loop body is complex.
```
Olinaaaloompa, line 160:
```suggestion
The performance improvement does not come for free but leads to longer compile time and larger binaries, especially when the number of iterations is large or when the loop body is complex.
```

docs/lang/articles/advanced/meta.md
```diff
@@ -166,6 +148,18 @@ def func():
   print(3)
 ```
 
+## When to use `ti.static` with for loops
```
Olinaaaloompa, line 151:
```suggestion
### When to use `ti.static` with for loops
```

docs/lang/articles/advanced/meta.md
```diff
@@ -189,7 +183,7 @@ That said, unrolling it will still help you reduce runtime overhead.
 
 ## Compile-time recursion of `ti.func`
 
-A compile-time recursive function is a function with recursion that can be recursively inlined at compile time. The condition which determines whether to recurse is evaluated at compile time.
+A compile-time recursive function is a function with recursion that can be recursively inlined at compile time. The condition which determines when to cease recursion is evaluated at compile time, hence must be known to the compiler.
```
Olinaaaloompa, line 186:
```suggestion
A compile-time recursive function is a function with recursion that can be recursively inlined at compile time. The condition that determines when to cease recursion is evaluated at compile time, hence must be known to the compiler.
```

Olinaaaloompa (changes requested):

docs/lang/articles/advanced/meta.md
```diff
@@ -166,6 +148,18 @@ def func():
...
 ```
 
+## When to use `ti.static` with for loops
+
+The main reason to use `ti.static` with for loops is to unroll the loop to improve runtime performance (see [Compile-time evaluations](#compile-time-evaluations)).
+
+An unrolling verison is often faster because it reduces branch overhead:
+
+1. It saves the instructions for incrementing the loop counter and checking the end-of-loop condition.
+2. It provides more instructions for the compiler to schedule across the whole iterations.
```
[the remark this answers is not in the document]
neozhaoliang: No, not schedule iterations, compiler schedule instructions in the iterations

docs/lang/articles/advanced/meta.md
neozhaoliang: I think the suggested change "... but leads to ..." is not as good as the original one?

Olinaaaloompa: I was thinking about reducing the number of verbs... We can revert to the original form with the colon but I would suggest phrasing it in this way: It takes longer compile time and generates larger binaries, especially when...

taichi-gardener:

PR #6321, opened 2022-10-14
[ci] Add end-to-end CI tests for meshtaichi

BillXu2000:
Continued from #6310.
Changes of mesh.py in taichi need to be merged before this CI change.

PR #6453, opened 2022-10-27
[ci] Add unix end-to-end CI tests for meshtaichi

BillXu2000:
Continued from #6321 

Meshtaichi repo is now public and the tests of which can be easily fetched.


g1n0st (approved):
LGTM

Issue #6447, opened 2022-10-26
Upgrade taichi to use LLVM15


ailzhang:
Thanks to @python3kgae we have pretty nice LLVM15 compatible code across the codebase now. And using an old version of LLVM (LLVM10) have limited us from applying advanced optimizations as well as adapting to newer hardware. So we're planning to upgrade taichi to depend on LLVM15. The overall process is kinda simple, fix the failed tests and enable TI_LLVM_15 in every CI job we have, and finally remove all the `ifdef TI_LLVM_15` in the codebase. 

A rough list of tasks:

- [x] linux cpu (#6435)
- [x] linux cuda (#6465)
- [x] mac cpu （#6634, #6483）
- [x] windows cpu
- [x] windows cuda (@python3kgae already did this, thanks!)
- [x] do we want to upgrade the compiler to compile `.bc` files? (#6678)
- [x] remove all the `ifdef TI_LLVM_15`! (#6685)
- [x] release prebuilt llvm binaries across different platforms & update docs (or shall we add llvm as a submodule?) (#6648)

python3kgae:
> Thanks to @python3kgae we have pretty nice LLVM15 compatible code across the codebase now. And using an old version of LLVM (LLVM10) have limited us from applying advanced optimizations as well as adapting to newer hardware. So we're planning to upgrade taichi to depend on LLVM15. The overall process is kinda simple, fix the failed tests and enable TI_LLVM_15 in every CI job we have, and finally remove all the `ifdef TI_LLVM_15` in the codebase.
> 
> A rough list of tasks:
> 
> * [ ]  linux cpu ([[ci] Swtich CPU jobs to use prebuilt LLVM15 #6435](https://github.com/taichi-dev/taichi/pull/6435))
> * [ ]  linux cuda
> * [ ]  mac cpu
> * [ ]  windows cpu
> * [x]  windows cuda (@python3kgae already did this, thanks!)
> * [ ]  do we want to upgrade the compiler to compile `.bc` files?
> * [ ]  remove all the `ifdef TI_LLVM_15`!
> * [ ]  release prebuilt llvm binaries across different platforms & update docs (or shall we add llvm as a submodule?)

> * [ ]  windows cpu
windows cpu is enabled too.

> * [ ]  do we want to upgrade the compiler to compile `.bc` files?
It would be nice to upgrade clang to 15, that will generate more test cases for opaque ptr.

[...]

PR #6458, opened 2022-10-27
[llvm] Fix broken llvm module on CUDA backend when TI_LLVM_15 is on

ailzhang:
Issue: #6447

### Brief Summary
Previously we inject some cuda kernel metadata after linking, that was problematic since when you modify metadata for a llvm::Function you'll have to use the exactly same LLVMContext it was created in. Thus we failed verifyModule check from LLVM15. And we happened to only TI_WARN there so didn't notice this issue. 
Screenshot from 2022-10-26 17-10-43

This PR fixes it by moving the mark as CUDA step up to parallel compilation so that it was guaranteed to be in the same LLVMContext.

### DIFF
--- a/taichi/codegen/llvm/codegen_llvm.cpp
+++ b/taichi/codegen/llvm/codegen_llvm.cpp
@@ -2749,6 +2749,15 @@ LLVMCompiledTask TaskCodeGenLLVM::run_compilation() {
   emit_to_module();
   eliminate_unused_functions();
 
+  if (config.arch == Arch::cuda) {
+    // CUDA specific metadata
+    for (const auto &task : offloaded_tasks) {
+      llvm::Function *func = module->getFunction(task.name);
+      TI_ASSERT(func);
+      tlctx->mark_function_as_cuda_kernel(func, task.block_dim);
+    }
+  }
+
   return {std::move(offloaded_tasks), std::move(module),
           std::move(used_tree_ids), std::move(struct_for_tls_sizes)};
 }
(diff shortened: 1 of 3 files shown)


taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2749,6 +2749,13 @@ LLVMCompiledTask TaskCodeGenLLVM::run_compilation() {
   emit_to_module();
   eliminate_unused_functions();
 
+  // CUDA specific metadata
+  for (const auto &task : offloaded_tasks) {
```
jim19930609, line 2753: This file is `codegen_llvm.cpp`, so this code block also marks CPU offloaded_tasks as cuda kernels?

ailzhang: oops I forgot the guard, thanks for catching!

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2749,6 +2749,15 @@ LLVMCompiledTask TaskCodeGenLLVM::run_compilation() {
   emit_to_module();
   eliminate_unused_functions();
 
+#ifdef TI_WITH_CUDA
```
jim19930609, line 2752: If you compiles with CUDA but executes on CPU, this is still problematic. Should check taichi::Arch at runtime as well?

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2749,6 +2749,15 @@ LLVMCompiledTask TaskCodeGenLLVM::run_compilation() {
...
+  // CUDA specific metadata
+  for (const auto &task : offloaded_tasks) {
```
lin-hitonami, line 2754: Should we wrap this within `if (config.arch == Arch::cuda)`?

taichi/codegen/llvm/codegen_llvm.cpp
ailzhang: you're absolutely right lol, thanks!

lin-hitonami (approved):
LGTM

jim19930609 (approved):

PR #6459, opened 2022-10-28
[ci] Run Meshtaichi tests only on cuda machine

BillXu2000:
related: #6453

Github CI machine sometimes gives results with large numerical error on Meshtaichi tests.
This pr tries running both cuda and CPU tests on cuda machine.

### DIFF
--- a/.github/workflows/scripts/unix_test.sh
+++ b/.github/workflows/scripts/unix_test.sh
@@ -113,13 +113,9 @@ else
 fi
 
 # meshtaichi end-to-end test
-if [[ $OSTYPE == "linux-"* && ($TI_WANTED_ARCHS == *"cuda"* || $TI_WANTED_ARCHS == *"cpu"*) ]]; then
+if [[ $OSTYPE == "linux-"* && $TI_WANTED_ARCHS == *"cuda"* ]]; then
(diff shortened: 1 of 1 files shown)


ailzhang:
@BillXu2000 is it possible to investigate what happened on the other machines?  It looks pretty suspicious to me the result from CPU vary depending on whether the machine has a GPU or not lol

BillXu2000:
> @BillXu2000 is it possible to investigate what happened on the other machines? It looks pretty suspicious to me the result from CPU vary depending on whether the machine has a GPU or not lol

The numerical error in [this run](https://github.com/taichi-dev/taichi/actions/runs/3339112012/jobs/5527652065) is way too large. I think the reason is that CPU CI machines are not as stable as the CUDA ones. Since CUDA CI machines can also run CPU tests, I rather use a more stable environment to run these tests.

ailzhang:
@BillXu2000 wait but shall we confirm if users may hit this large numerical error on their machines first?

BillXu2000:
> @BillXu2000 wait but shall we confirm if users may hit this large numerical error on their machines first?

I don't know. I don't know how to reproduce this bug, and MeshTaichi doesn't have many users now.

PR #6460, opened 2022-10-28
Revert "[ci] Add unix end-to-end CI tests for meshtaichi"

ailzhang:
Reverts taichi-dev/taichi#6453 since it broke nightly, https://github.com/taichi-dev/taichi/actions/workflows/release.yml


BillXu2000 (approved):
OK, maybe we can consider #6459 later

PR #6462, opened 2022-10-28
[Bug] Fix cache_loop_invariant_global_vars pass

### LINKED ISSUE #6441: [Bug] Numerical error with Taichi v1.2.0
Thanks to @lucifer1004 who encountered this bug in his development.

This bug is backend neural: cpu / cuda / vulkan all behave consistently.

[Minimal reproduction](https://github.com/turbo0628/TaichiCloud/blob/main/TaichiCloud/reprod.py)

Briefings:

The `init` part is good, but the main function `adj_cellwise_hlpr ` gets unexpected value in the field `self.f_r` if you comment out the `print` statement at [this line](https://github.com/turbo0628/TaichiCloud/blob/main/TaichiCloud/reprod.py#L109). 

The v1.1.3 has no numerical problem. We should check which newly added pass caused the error.

lin-hitonami:
Issue: fixes #6441

### DIFF
--- a/taichi/transforms/cache_loop_invariant_global_vars.cpp
+++ b/taichi/transforms/cache_loop_invariant_global_vars.cpp
@@ -122,9 +122,8 @@ class CacheLoopInvariantGlobalVars : public LoopInvariantDetector {
         std::make_unique<AllocaStmt>(dest->ret_type.ptr_removed());
     auto alloca_stmt = alloca_unique.get();
     modifier.insert_before(current_loop_stmt(), std::move(alloca_unique));
-    if (status == CacheStatus::Read) {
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Thanks for fixing it! I have two nits:
- is it possible to add a test to prevent a future breakage ;) ?
- is it a critical bug that requires a bug fix release?

lin-hitonami:
> * is it possible to add a test to prevent a future breakage ;) ?

Yes, I'll add one.
> * is it a critical bug that requires a bug fix release?

I think so.

ailzhang:
> is it a critical bug that requires a bug fix release?
> I think so.
oops cc: @jim19930609 @feisuzhu

PR #6463, opened 2022-10-28
[ci] Move implicit fem android demo to master branch

ailzhang:
If we pin on a commit taichi-aot-demo it's hard to catch all the errors.

Let's see if things already exploded... 

Issue: #


PENGUINLIONG (approved):
thx!

Issue #6434, opened 2022-10-26
Add support for AMDGPU


GaleSeLee:
**Purpose**
A clear and concise description of what you want.
I find Taichi users have demand for AMDGPU. (related issue https://github.com/taichi-dev/taichi/issues/4586)
There is also a discussion of the AMD GPU backend here https://github.com/taichi-dev/taichi/issues/412
Thus, I would like to add an AMDGPU backend to the compiler so that my 6900XT can be utilized. 

**Solution**
The purpose is to give Taichi preliminary support for AMDPU via LLVM and ROCm.
I started by completing the following steps step by step
1. the relevant dependencies in cmake
2. a specific version of Taichi-runtime module(taichi/runtime/llvm/runtime_module) for AMDGPU. During the implementation, I found that I couldn't reuse the CPU's llvm directly like taichi does for CUDA, I needed to do some addrspace conversions, so I temporarily used hipcc to help me deal with these issues.
3. interaction between Taichi and ROCm library functions. In cuda backend, some driver functions are used. However, It's convenient replace these with functions provided by ROCm, such as hipMemcpy, hipModuleLaunchKernel.
4. a codegen skeleton (like taichi/codegen/cuda)
5. convert llvm IR to amdgpu ISA / hsaco(ref [link](https://rocmdocs.amd.com/en/latest/Deep_learning/GCN-asm-tutorial.html?highlight=hsaco))
6. Unit test configuration for amdgpu.

GaleSeLee:
It seems better to add cmake at the end(after step5)

GaleSeLee:
Here is task list.
- [x] render hardware interface(https://github.com/taichi-dev/taichi/pull/6464)
- [x] codegen(https://github.com/taichi-dev/taichi/pull/6469)
- [x] runtime(https://github.com/taichi-dev/taichi/pull/6482)
- [ ] update runtime module
- [ ] linker math library
- [ ] amdgpu backend interface
- [ ] ti.amdgpu interface
- [ ] misc
- [ ] cmake related
- [ ] Unit test configuration

ITCJ:
**Amazing works.** 

I'm also considering this task since our new HPC is using Sugon gtx106 (known as Vega 20). I am quite interested is there any possible to cooperate ?

**Besides,**
Are you also working on ASC 2022 competition ? I notice that you forked DeePMD-kit which is one of the tasks of ASC2022.

GaleSeLee:
[...]

PR #6464, opened 2022-10-28
[amdgpu] Part0 add render hardware interface

GaleSeLee:

### Brief Summary
It contains four parts(`driver`, `context`, `device` and `caching_allocator`). The code is similar to `cuda/rhi`. However, there are still some differences between `amdgpu/rhi` and `cuda/rhi` 
#### context
1. The method of obtaining the hardware version
2. Context::launch
#### driver
1. ROCm/hip internal functions
#### cmake

The current cmake compilation system is sufficient to support the Unit test in https://github.com/taichi-dev/taichi/pull/6597


GaleSeLee:
\test

GaleSeLee:
Due to the addition of new content, the PR was updated in https://github.com/taichi-dev/taichi/pull/6975

cmake/TaichiCore.cmake
```diff
@@ -108,6 +117,12 @@ if (TI_WITH_CUDA)
   list(APPEND TAICHI_CORE_SOURCE ${TAICHI_CUDA_RUNTIME_SOURCE})
 endif()
 
+if (TI_WITH_AMDGPU)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DTI_WITH_AMDGPU")
+# file(GLOB TAICHI_AMDGPU_RUNTIME_SOURCE "taichi/runtime/amdgpu/runtime.cpp")
+  list(APPEND TAIHI_CORE_SOURCE ${TAICHI_AMDGPU_RUNTIME_SOURCE})
```
jim19930609, line 123: Does it work even if the former code has been commented out?

cmake/TaichiCore.cmake
```diff
@@ -3,6 +3,7 @@ option(TI_WITH_LLVM "Build with LLVM backends" ON)
 option(TI_WITH_METAL "Build with the Metal backend" ON)
 option(TI_WITH_CUDA "Build with the CUDA backend" ON)
 option(TI_WITH_CUDA_TOOLKIT "Build with the CUDA toolkit" OFF)
+option(TI_WITH_AMDGPU "Build with the AMDGPU backend" OFF)
```
jim19930609, line 6: What if the user sets both `TI_WITH_CUDA=ON` and `TI_WITH_AMDGPU=ON`?

taichi/rhi/amdgpu/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+    amdgpu_context.cpp
+    amdgpu_driver.cpp
+  )
+
+target_include_directories(${AMDGPU_RHI}
+  PRIVATE
+    ${PROJECT_SOURCE_DIR}
+    ${PROJECT_SOURCE_DIR}/external/eigen
+    ${PROJECT_SOURCE_DIR}/external/spdlog/include
+    ${PROJECT_SOURCE_DIR}/external/glfw/include
```
jim19930609, line 18: Wondering if you actually need `glfw` for AMDGPU, since we won't use AMDGPU to perform any rendering process?

taichi/rhi/amdgpu/amdgpu_caching_allocator.cpp
```diff
@@ -0,0 +1,40 @@
...
+    ret = it_blk->second;
+    mem_blocks_.erase(it_blk);
+  } else {
+    ret = device_->allocate_llvm_runtime_memory_jit(params);
+  }
+  return ret;
+}
+
+void AmdgpuCachingAllocator::release(size_t sz, uint64_t *ptr) {
+  mem_blocks_.insert({sz, ptr});
```
jim19930609, line 35: should it be `erase()` instead since you're releasing the memory?

jim19930609 (approved):

cmake/TaichiCore.cmake
GaleSeLee: It will compile both `CUDA` and `AMDGPU` devices. However, the exact device available at runtime depends on the GPU in the system.

cmake/TaichiCore.cmake
GaleSeLee: yeah, because RHI does not involve a platform.

cmake/TaichiCore.cmake
jim19930609: Given that only one of AMDGPU or CUDA will be available at runtime, I was thinking we should warn users not to enable both TI_WITH_CUDA and TI_WITH_AMDGPU at the same time. How about adding a check here to verify that?

taichi/rhi/amdgpu/CMakeLists.txt
GaleSeLee: yeah, it is redundant

taichi/rhi/amdgpu/amdgpu_caching_allocator.cpp
GaleSeLee: I think putting the memory block back into the available memory list also can be considered a release?
By the way, we do the same thing on the other backend(e.g. CUDA).

PR #6469, opened 2022-10-28
[amdgpu] Part1 add codegen

GaleSeLee:

### Brief Summary
This part contains `CHI IR->LLVM IR` part. Similar to part0, the skeleton of this part is similar to `cuda`. Some important notes are following.
1. `ocml` is the math library like 'libdevice' in `nvidia`. It provides most of the functions we need. Remains such as `sgn`, and `abs` have been completed by hand codes
2. The kernel parameter passing part is different from the `CUDA`. `extra_args` is the only way that could be accepted by `device kernel.`
3. still uses jargon from `nvidia`.(e.g. `sm`, `block` and `grid`)


cmake/TaichiCore.cmake
```diff
@@ -237,6 +237,9 @@ if(TI_WITH_LLVM)
     if (TI_WITH_AMDGPU)
         llvm_map_components_to_libnames(llvm_amdgpu_libs AMDGPU)
         add_subdirectory(taichi/rhi/amdgpu)
+        add_subdirectory(taichi/codegen/amdgpu)
+
+        target_link_libraries(${CORE_LIBRARY_NAME} PRIVATE amdgpu_codegen)
         target_link_libraries(${CORE_LIBRARY_NAME} PRIVATE amdgpu_rhi)
```
jim19930609, line 243: you can squeeze these two `target_link_libraries`

jim19930609 (approved):

Issue #6445, opened 2022-10-26
CI/CD Pipeline Overhaul 2022


feisuzhu:
This is an on-going work primarily to reduce CI/CD build time and improve maintainability.
Discussions and feature requests are welcome.

### Buildbots

- 5 Linux, 3 + 2(VM) Windows, 3 macOS, organization shared

![2022-10-26-183022_934x857_scrot](https://user-images.githubusercontent.com/857880/198003906-60931f70-902e-4e49-8bf9-8961873b83ea.png)

### Pipeline

- `release.yml`: Build nightly wheels, run complete tests, upload to pypi. GA releases also goes here.
![2022-10-26-183142_1257x534_scrot](https://user-images.githubusercontent.com/857880/198004181-f58a2e95-d635-48a5-bccd-787df298102d.png)

- `testing.yml`: Build PR and master, run partial or complete tests.

![2022-10-26-183127_1269x601_scrot](https://user-images.githubusercontent.com/857880/198004204-0e05d8f6-fd8c-44b8-ae58-e26eb82b52cb.png)

- `perf.yml`: Performance monitoring, only run on master.

![无标题-2022-10-26-1844](https://user-images.githubusercontent.com/857880/198009123-e3da093b-4151-4f35-a13c-e115037ded16.png)

### Code & logic duplication

- `release.yml` is largely the same with `testing.yml` except publishing and different build matrix.
- Different code path for CPU and GPU building tasks.

#### Latency !

- 1h30min for each complete run (master branch) (!!)
- Slightly over 1h for each PR run (!!!)

(pic)

#### Wasted cycles & bandwidth on same things

- For now C++ build cache are stored locally on each buildbot machine.
- Auxiliary git repos are not cached at all.
- `Perf Monitoring` and `Android Demo` needs to compile taichi binary separately.
- Pulling huge docker images for every run on GitHub Hosted Runners (cost 30min).
- ti.init dominates test case time consumption

#### Buildbot cluster of a low capacity
- Only a single M1 Mac Mini operating (SPOF)
- Unlimited(?) but weak GitHub Hosted Runners, better not to have them.
- Beefy runners are prohibitively expensive.

#### Under-utilized buildbot cluster

- Cores sitting idle due to extra low usage when running tests
- Different test suites run in serial

### Strong coupling with GitHub Actions

- Difficult for developers to run workflows locally

### Non-reproducible building environment (partially addressed)

- Windows and macOS builds still run on bare metal
[...]

PR #6476, opened 2022-10-31
[ci] GitHub Actions: Merge CPU & GPU tasks

feisuzhu:
Issue: #6445


feisuzhu:
@ailzhang  Please help merge this

ailzhang (approved):
Thanks!

.github/workflows/testing.yml
```diff
@@ -395,7 +306,7 @@ jobs:
              /home/dev/taichi/.github/workflows/scripts/unix_test.sh
         env:
           PY: py38
-          TI_WANTED_ARCHS: 'cuda,vulkan,opengl'
+          TI_WANTED_ARCHS: 'cpu,cuda,vulkan,opengl'
```
ailzhang, line 309: just to make sure we don't accidentally reduce coverage, shall we keep `cc` in this list as well? 
(which should be removed in a separate PR along with cc backend if we want. )

feisuzhu: Added `cc` to the list

ailzhang (approved):
Thanks!

PR #6483, opened 2022-10-31
[ci] Move m1 job to use build with LLVM 15

ailzhang:
Issue: #6447 

### Brief Summary
Just trying my luck in CI since my local mac is on macOS12.

### DIFF
--- a/.github/workflows/scripts/unix-build.sh
+++ b/.github/workflows/scripts/unix-build.sh
@@ -8,10 +8,17 @@ set -ex
 
 # TODO: Move llvm installation from container image to here
 if [[ "$LLVM_VERSION" == "15" ]]; then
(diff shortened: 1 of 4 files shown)


ailzhang:
FYI there's a real test failure that I can repro locally on my M1. A minimal repro script is 
```
import taichi as ti
ti.init(ti.cpu, print_ir=False, print_kernel_llvm_ir=False, debug=False)

N = 1024 * 100

@ti.kernel
def reduce() -> ti.f32:
  res = 0.
  for _ in range(N):
    res += 1.2
  return res

res = reduce()

ground_truth = N * 1.2
print(res, ground_truth)
```
Confirmed that this script runs fine with llvm10 but not llvm15.

feisuzhu (approved):
:shipit:

k-ye:
:shipit:

PR #6486, opened 2022-10-31
[amdgpu] Part3 update runtime module

GaleSeLee:

### Brief Summary
1. This is a special part of the Tacihi runtime module for the `AMDGPU` backend. Tacihi's runtime module uses `clang++` to generate `LLVM IR` is different in memory allocation differs from the cpu-generated `LLVM IR`. The following is an example.
```
C/C++ code
void func(int *a, int *b) {
    *a = *b;
}
x86_64 backend LLVM IR
define dso_local void @cpu_func(i32* %0, i32* %1) #2 {
  %3 = alloca i32*, align 8
  %4 = alloca i32*, align 8
  store i32* %0, i32** %3, align 8
  store i32* %1, i32** %4, align 8
  %5 = load i32*, i32** %4, align 8
  %6 = load i32, i32* %5, align 4
  %7 = load i32*, i32** %3, align 8
  store i32 %6, i32* %7, align 4
  ret void
}
__global__ function on AMDGPU
define protected amdgpu_kernel void @global_func(i32 addrspace(1)* %0, i32 addrspace(1)* %1) #4 {
  %3 = alloca i32*, align 8, addrspace(5)
  %4 = alloca i32*, align 8, addrspace(5)
  %5 = alloca i32*, align 8, addrspace(5)
  %6 = alloca i32*, align 8, addrspace(5)
  %7 = addrspacecast i32* addrspace(5)* %3 to i32**
  %8 = addrspacecast i32* addrspace(5)* %4 to i32**
  %9 = addrspacecast i32* addrspace(5)* %5 to i32**
  %10 = addrspacecast i32* addrspace(5)* %6 to i32**
  %11 = addrspacecast i32 addrspace(1)* %0 to i32*
  store i32* %11, i32** %7, align 8
  %12 = load i32*, i32** %7, align 8
  %13 = addrspacecast i32 addrspace(1)* %1 to i32*
  store i32* %13, i32** %8, align 8
  %14 = load i32*, i32** %8, align 8
  store i32* %12, i32** %9, align 8
  store i32* %14, i32** %10, align 8
  %15 = load i32*, i32** %10, align 8
  %16 = load i32, i32* %15, align 4
  %17 = load i32*, i32** %9, align 8
  store i32 %16, i32* %17, align 4
  ret void
}
__device__ function on AMDGPU
define hidden void @device_func(i32* %0, i32* %1) #2 {
  %3 = alloca i32*, align 8, addrspace(5)
  %4 = alloca i32*, align 8, addrspace(5)
  %5 = addrspacecast i32* addrspace(5)* %3 to i32**
  %6 = addrspacecast i32* addrspace(5)* %4 to i32**
  store i32* %0, i32** %5, align 8
  store i32* %1, i32** %6, align 8
  %7 = load i32*, i32** %6, align 8
  %8 = load i32, i32* %7, align 4
  %9 = load i32*, i32** %5, align 8
  store i32 %8, i32* %9, align 4
  ret void
}
```
2. There are some differences in the place about `allocainst`, specifically about addrspace (for `AMDGPU`, [this](https://llvm.org/docs/AMDGPUUsage.html#address-spaces) will be helpful). I have not found documentation describing how to write the correct `LLVM IR` on `AMDGPU`, through my observation of the `LLVM IR` generated by `clang++/hipcc`. We need to deal with the arguments of the `__global__` function and the `allocainst` (including specifying the addrspace of `allocainst` and performing addrspace-cast) while for the `__device__` function we do not need to deal with the arguments of the function.


taichi/runtime/llvm/llvm_context.cpp
```diff
@@ -399,93 +387,143 @@ std::unique_ptr<llvm::Module> TaichiLLVMContext::module_from_file(
...
-    // runtime_module->print(llvm::errs(), nullptr);
+    if (arch_ == Arch::amdgpu) {
+      module->setTargetTriple("amdgcn-amd-amdhsa");
+#ifdef TI_WITH_AMDGPU
+      for (auto &f : *module) {
+        f.addFnAttr("target-cpu", "");
+        f.addFnAttr("target-features", "");
+        for (auto &bb : f) {
+          std::vector<llvm::AllocaInst *> alloca_inst_vec;
+          for (llvm::Instruction &inst : bb) {
```
jim19930609, line 502: Essentially this is trying to handle AMD's address space semantics, where `Alloca` should always allocate memory on `Private (thread local)` memory represented by `address space (5)` in LLVM. (https://llvm.org/docs/AMDGPUUsage.html#amdgpu-address-spaces)

However, the address space interpretation is strictly hardware-specific, and we should migrate the following implementation to `AMDGPU::Codegen()`

taichi/runtime/llvm/llvm_context.cpp
```diff
@@ -817,6 +860,71 @@ void TaichiLLVMContext::update_runtime_jit_module(
...
+          new_func_params.push_back(arg.getType());
+        }
+      }
+      auto new_func_type = llvm::FunctionType::get(func_type->getReturnType(),
+                                                   new_func_params, false);
+      auto new_func = llvm::Function::Create(new_func_type, f->getLinkage(),
+                                             f->getAddressSpace());
+      // NF->copyAttributesFrom(f);
+      new_func->setCallingConv(llvm::CallingConv::AMDGPU_KERNEL);
+      new_func->addFnAttr("amdgpu-flat-work-group-size", "1, 1024");
```
jim19930609, line 897: `1` and `1024` are magic numbers, how were they picked? 

Should `work_group_size` be set by `block_size`?

taichi/runtime/llvm/llvm_context.cpp
```diff
@@ -817,6 +860,71 @@ void TaichiLLVMContext::update_runtime_jit_module(
...
+      bool is_kernel = false;
+      const std::string func_name = f.getName().str();
+      if (starts_with(func_name, "runtime_")) {
+        mark_function_as_amdgpu_kernel(&f);
+        is_kernel = true;
+      }
+      if (!is_kernel && !f.isDeclaration())
+        f.setLinkage(llvm::Function::PrivateLinkage);
+    }
+    std::vector<llvm::Function *> global_func;
```
jim19930609, line 874: I dont quite find a definition for `global function` in LLVM, there's `external function` only. 

Judged by `llvm::CallingConv::AMDGPU_KERNEL`, do you mean AMD GPU's device function?

taichi/runtime/llvm/llvm_context.cpp
```diff
@@ -817,6 +860,71 @@ void TaichiLLVMContext::update_runtime_jit_module(
...
+    for (auto &f : *module) {
+      if (f.getCallingConv() == llvm::CallingConv::AMDGPU_KERNEL)
+        global_func.push_back(&f);
+    }
+    for (auto &f : global_func) {
+      llvm::FunctionType *func_type = f->getFunctionType();
```
jim19930609, line 880: This is a bit too hacky, can we try to setup the function attributes for AMD GPU whenever the function is created, instead of patch it afterwards?

taichi/runtime/llvm/llvm_context.cpp
```diff
@@ -817,6 +860,71 @@ void TaichiLLVMContext::update_runtime_jit_module(
...
+                                             addrspacecast);
+          I->replaceAllUsesWith(addrspacecast);
+          I2->takeName(&*I);
+        } else {
+          I->replaceAllUsesWith(&*I2);
+          I2->takeName(&*I);
+        }
+      }
+
+      SmallVector<std::pair<unsigned, MDNode *>, 1> MDs;
```
jim19930609, line 920: I doubt if `SmallVector<.., 1>` has any benefit? Can you give an estimated number regarding the size of `MDs`?

jim19930609 (changes requested):
In general, for hardware-specifc code or configurations,  we should comment a reference link to a specific chapter/part of AMD-GPU-guide whenever we use some magic strings or magic numbers. This is especially true for setting up function attributes or so.

The legacy CUDA backend is filled with magic/hack and lacks enough comments/explanations. Let's improve this situation starting from AMD GPU. Thanks!

taichi/runtime/llvm/llvm_context.cpp
GaleSeLee: This part is trying to handle the address space semantics, which are generated from taichi-runtime(located taichi/runtime/llvm/runtime_module) and are unsuitable. Taichi will launch some functions that are in this module(such as runtime_initialize) during running ti.init before AMDGPU::codegen. Thus, I don't think this should be the scope of AMDGPU::codegen.

taichi/runtime/llvm/llvm_context.cpp
GaleSeLee: 1. According to https://releases.llvm.org/12.0.1/docs/AMDGPUUsage.html#llvm-ir-attributes, the function of attribution `amdgpu-flat-work-group-size` is that specify the minimum and maximum flat work group sizes that will be specified when the kernel is dispatched.
2. `workgroup` is similar to `block` in NVIDIA

taichi/runtime/llvm/llvm_context.cpp
GaleSeLee: 1. Ohhhh
Yeah, you're right.  `llvm::CallingConv::AMDGPU_KERNEL` has the similar function to `getOrInsertNamedMetadata("nvvm.annotations")` which is used in CUDA.
2. `kernel_function` may be a better choice.

taichi/runtime/llvm/llvm_runtime_executor.cpp
```diff
@@ -112,6 +112,15 @@ LlvmRuntimeExecutor::LlvmRuntimeExecutor(CompileConfig &config,
   }
 #endif
 
+#if defined(TI_WITH_AMDGPU)
+  if (config.arch == Arch::amdgpu) {
+    AMDGPUContext::get_instance().set_debug(config.debug);
```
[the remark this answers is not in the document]
GaleSeLee: 1. This does affect performance, but in debug mode, I think this is necessary, otherwise it will make it more difficult to locate the problem
2. By the way, I think the main reason of poor performance on AMDGPU backend is similar to this. And I will try to solve this problem when i finish to pull these code.

jim19930609 (approved):
LGTM!

PR #6490, opened 2022-11-01
[llvm] Add support for data types of different sizes on dynamic SNode

lin-hitonami:
Issue: #5420

### Brief Summary
This PR splits `append` operation of Dynamic SNode to two steps:
1. Allocate the space for the new element, increase the length by one, and return the address (void *) to the first byte of the new element.
2. Use `GetChStmt` to get the pointer of the element (of the right type) from the address returned in step 1, and use `GlobalStoreStmt` to store the value into the address.

**This PR breaks Dynamic SNode on Metal**

### DIFF
--- a/taichi/runtime/llvm/runtime_module/node_dynamic.h
+++ b/taichi/runtime/llvm/runtime_module/node_dynamic.h
@@ -58,11 +58,12 @@ void Dynamic_deactivate(Ptr meta_, Ptr node_) {
   }
 }
 
-i32 Dynamic_append(Ptr meta_, Ptr node_, i32 data) {
+Ptr Dynamic_allocate(Ptr meta_, Ptr node_, i32 *len) {
   auto meta = (DynamicMeta *)(meta_);
   auto node = (DynamicNode *)(node_);
   auto chunk_size = meta->chunk_size;
   auto i = atomic_add_i32(&node->n, 1);
+  *len = i;
   int chunk_start = 0;
   auto p_chunk_ptr = &node->ptr;
   while (true) {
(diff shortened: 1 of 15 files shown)


tests/python/test_dynamic.py
```diff
@@ -135,14 +135,14 @@ def func():
...
-@test_utils.test(require=ti.extension.sparse)
+@test_utils.test(require=ti.extension.sparse, exclude=[ti.metal])
 def test_dense_dynamic():
     # The spin lock implementation has triggered a bug in CUDA, the end result
     # being that appending to Taichi's dynamic node messes up its length. See
     # https://stackoverflow.com/questions/65995357/cuda-spinlock-implementation-with-independent-thread-scheduling-supported
     # CUDA 11.2 didn't fix this bug, unfortunately.
-    if ti.lang.impl.current_cfg().arch == ti.cuda:
-        pytest.skip('CUDA spinlock bug')
+    # if ti.lang.impl.current_cfg().arch == ti.cuda:
```
lin-hitonami, line 144: I tried to run this test on CUDA and it seems alright now. Should we delete those comments and enable this test on CUDA now? @k-ye

ailzhang: @lin-hitonami to isolate the changes, can you remove this skip & comment in a separate PR? ;)

taichi/runtime/llvm/runtime_module/node_dynamic.h
```diff
@@ -76,14 +77,12 @@ i32 Dynamic_append(Ptr meta_, Ptr node_, i32 data) {
...
-      *(i32 *)(*p_chunk_ptr + sizeof(Ptr) +
-               (i - chunk_start) * meta->element_size) = data;
-      break;
+      return *p_chunk_ptr + sizeof(Ptr) +
+             (i - chunk_start) * meta->element_size;
     }
     p_chunk_ptr = (Ptr *)(*p_chunk_ptr);
     chunk_start += chunk_size;
   }
-  return i;
```
ailzhang, line 86: oops, can we return some nullptr here and add a comment `NEVER REACH HERE` so that it's easier to read?

ailzhang (approved):
Thanks!!

PR #6492, opened 2022-11-01
[aot] Warn the user about out-of-range access in C++ wrapper

PENGUINLIONG:
We recently introduced a breaking change to `taichi.hpp` that `NdArray<T>::read()` and `NdArray<T>::write()` now consumes number of scalars rather than size-in-bytes. This PR sets the `TI_ERROR_ARGUMENT_OUT_OF_RANGE` error in case of out-of-range access.


ailzhang (approved):
LGTM!

PR #6493, opened 2022-11-01
[lang] Support matrix args on real function

lin-hitonami:
Issue: #602

### Brief Summary
We only need to disassemble the arguments and pass them to the real function one by one just as we pass matrix arguments to the kernel.

It does not support local matrix and element of matrix field as argument when real_matrix=True yet.


ailzhang (approved):
LGTM!

jim19930609 (approved):

PR #6494, opened 2022-11-01
[aot] Disable physical storage buffer in runtime

PENGUINLIONG:
The physical storage buffer codegen was disabled but the runtime can still use them. This is an attempt to disable psb at runtime. Will bring it back after devcap.


ailzhang (approved):
Thanks!

PR #6498, opened 2022-11-02
[aot] Select any of vkGetPhysicalDeviceMemoryProperties2

PENGUINLIONG:
Vulkan Validation layer intercepts the function pointer getter and returns nullptr for KHR-suffixed named if such interfaces have been incorporated into a version of Vulkan Core Profile. For example, validation layer returns `nullptr` for `vkGetPhysicalDeviceMemoryProperties2KHR` with a 1.1 instance because the function has been incorporated and is renamed to `vkGetPhysicalDeviceMemoryProperties2`. But this function is required by VMA and is asserted non-null, which leads to sophisticated assertion crash in AOT demos.


ailzhang (approved):
Thanks!

PR #6499, opened 2022-11-02
[lang] MatrixType refactor: Properly raise error when dynamic_index=False

strongoier:
Issue: #5819

### Brief Summary

This is to match the old behavior before the existence of `MatrixType`.

### DIFF
--- a/taichi/transforms/scalarize.cpp
+++ b/taichi/transforms/scalarize.cpp
@@ -554,7 +554,22 @@ class ScalarizePointers : public BasicStmtVisitor {
 
         // TODO(zhanlue): loose this contraint once dynamic indexing is properly
         // handled
-        TI_ASSERT(stmt->offset->is<ConstStmt>());
+        if (!stmt->offset->is<ConstStmt>()) {
+          modifier_.modify_ir();
+          throw TaichiSyntaxError(fmt::format(
+              "{}The index of a Matrix/Vector must be a compile-time constant "
+              "integer.\n"
+              "This is because matrix operations will be **unrolled** at "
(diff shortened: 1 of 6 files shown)


taichi/transforms/scalarize.cpp
```diff
@@ -554,7 +554,23 @@ class ScalarizePointers : public BasicStmtVisitor {
 
         // TODO(zhanlue): loose this contraint once dynamic indexing is properly
         // handled
-        TI_ASSERT(stmt->offset->is<ConstStmt>());
+        if (!stmt->offset->is<ConstStmt>()) {
+          modifier_.modify_ir();
```
AD1024, line 558: Shall we get rid of this line since we are immediately raising an exception.

strongoier: Removing this line will trigger an assertion error in the destructor of `modifier_`: https://github.com/taichi-dev/taichi/blob/b6b431764eb48f06a651f9afc60e3e28f8629c12/taichi/ir/ir.cpp#L409-L416

jim19930609: Destructors in C++ are marked as `noexcept` by default, thus use `TI_ASSERT` in this case can cause undefined behavior. 

The general idea is that `destructors` should try not get interrupted whatsoever, otherwise there's always a risk of resource leak since it violates RAII (part of the resources release codes won't run if a `throw` occurs)

It might worth a comment here.

strongoier: Comments have been added.

jim19930609 (approved):

PR #6500, opened 2022-11-02
[example] Fix out-of-bound access in examples

strongoier:
### Brief Summary

Before #6485 out-of-bound accesses of SNodes were silent - through `BitExtractStmt`  they became in-bound but might not result in expected results. The optimization in #6485 removed the in-bound transition in certain cases so it was likely that an out-of-bound access would behave differently. The CI failure of #6485 (https://github.com/taichi-dev/taichi/actions/runs/3367141462/jobs/5584351541) revealed out-of-bound access in examples, which could be confirmed by turning on `ti.init(debug=True)`. This PR aims at fixing those examples.


ailzhang (approved):
Thanks!

PR #6502, opened 2022-11-02
[lang] Add support for struct on Dynamic SNode

lin-hitonami:
Issue: #5420

### Brief Summary
This PR lets `SNodeOpExpression` store `vector<Expr> values` instead of `Expr value`, and enables passing a struct to `ti.append`.

### DIFF
--- a/taichi/ir/expr.h
+++ b/taichi/ir/expr.h
@@ -133,9 +133,18 @@ Expr expr_rand() {
   return taichi::lang::expr_rand(get_data_type<T>());
 }
 
+/*
+ * This function allocates the space for a new item (a struct or a scalar)
+ * in the Dynamic SNode, and assigns values to the elements inside it.
+ *
+ * When appending a struct, the size of vals must be equal to
+ * the number of elements in the struct. When appending a scalar,
+ * the size of vals must be one.
+ */
+
 Expr snode_append(SNode *snode,
                   const ExprGroup &indices,
-                  const std::vector<Expr> &val);
+                  const std::vector<Expr> &vals);
 
 Expr snode_is_active(SNode *snode, const ExprGroup &indices);
 
--- a/taichi/ir/expr.h
+++ b/taichi/ir/expr.h
@@ -133,7 +133,9 @@ Expr expr_rand() {
   return taichi::lang::expr_rand(get_data_type<T>());
 }
 
-Expr snode_append(SNode *snode, const ExprGroup &indices, const std::vector<Expr> &val);
+Expr snode_append(SNode *snode,
+                  const ExprGroup &indices,
+                  const std::vector<Expr> &val);
 
 Expr snode_is_active(SNode *snode, const ExprGroup &indices);
 
--- a/taichi/ir/expr.h
+++ b/taichi/ir/expr.h
@@ -133,7 +133,7 @@ Expr expr_rand() {
   return taichi::lang::expr_rand(get_data_type<T>());
 }
 
-Expr snode_append(SNode *snode, const ExprGroup &indices, const Expr &val);
+Expr snode_append(SNode *snode, const ExprGroup &indices, const std::vector<Expr> &val);
 
 Expr snode_is_active(SNode *snode, const ExprGroup &indices);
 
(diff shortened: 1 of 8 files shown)


taichi/ir/frontend_ir.h
```diff
@@ -730,7 +730,7 @@ class SNodeOpExpression : public Expression {
   SNode *snode;
   SNodeOpType op_type;
   ExprGroup indices;
-  Expr value;
+  std::vector<Expr> values;
```
jim19930609, line 733: Can we add a comment here stating that `values` are only used for `op_type == SNodeOpType::append`

taichi/ir/expr.h
```diff
@@ -133,7 +133,9 @@ Expr expr_rand() {
   return taichi::lang::expr_rand(get_data_type<T>());
 }
 
-Expr snode_append(SNode *snode, const ExprGroup &indices, const Expr &val);
+Expr snode_append(SNode *snode,
```
jim19930609, line 136: The semantics are little tricky here, since we won't append a `list`, but each element of it. 

Can we at least add a comment to say that this function flattens `std::vector<Expr> val` and append each element in sequence?

taichi/ir/frontend_ir.cpp
```diff
@@ -925,18 +925,19 @@ void SNodeOpExpression::flatten(FlattenContext *ctx) {
...
+      flatten_rvalue(value, ctx);
+    }
     auto alloca = ctx->push_back<AllocaStmt>(PrimitiveType::i32);
     auto addr =
         ctx->push_back<SNodeOpStmt>(SNodeOpType::allocate, snode, ptr, alloca);
-    auto ch_addr = ctx->push_back<GetChStmt>(addr, snode, 0);
-    ctx->push_back<GlobalStoreStmt>(ch_addr, value->stmt);
+    for (int i = 0; i < values.size(); i++) {
+      auto ch_addr = ctx->push_back<GetChStmt>(addr, snode, i);
+      ctx->push_back<GlobalStoreStmt>(ch_addr, values[i]->stmt);
```
jim19930609, line 936: Do we need to verify whether `ret_type` of `values[i]` fits with that of the `snode`, in `SNodeOpExpression::type_check`? (with type_promotion considered)

tests/python/test_dynamic.py
```diff
@@ -265,3 +265,28 @@ def make_list():
 
     for i in range(20):
         assert x[i] == i * i * i * 10000000000
+
+
+@test_utils.test(require=ti.extension.sparse, exclude=[ti.metal])
```
jim19930609, line 270: Comment on why `ti.metal` gets excluded?

taichi/ir/expr.h
lin-hitonami: The user passes in a struct, and this function allocates the space for the new struct and assigns value to every element in the struct in sequence.

tests/python/test_dynamic.py
lin-hitonami: Because according to @k-ye we may give up sparse support for metal backend. So for now I'm not considering metal when implementing dynamic SNode (Since Metal backend does not support dynamic memory allocating, we just allocate the whole memory when we materialize it. So it's easier than LLVM-based backends to implement if we decide not to give up metal backend).

taichi/ir/expr.h
jim19930609: sure, can we add this as a code comment to explain how `snode_append()` works? 

Suppose if you call `snode_append()` 3 times to append a `struct1`, a `scalar`, and another `struct2` consecutively, then essentially we'll flatten + squeeze everything, then store each value to the Place snodes under a Dynamic snode. This is much different compared with the former behavior (append scalar only) and does worth a comment to clarify how it works.

taichi/ir/frontend_ir.cpp
lin-hitonami: This is done inside the type checking of `GlobalStoreStmt` so we don't need to check it inside `SNodeOpExpression::type_check`?

jim19930609 (approved):
Thanks!

PR #6510, opened 2022-11-03
[cuda] Enable test on dense dynamic field

lin-hitonami:
Issue: #2179

### Brief Summary
It seems that CUDA has fixed this issue.


ailzhang (approved):
LGTM if CI passes ;)

PR #6512, opened 2022-11-03
[Lang] Add deactivate attribute to dynamic snodes

neozhaoliang:


python/taichi/lang/ast/ast_transformer.py
```diff
@@ -772,6 +772,17 @@ def build_Attribute(ctx, node):
             index = build_stmt(ctx, node.value.slice)
             node.value.ptr = None
             node.ptr = lambda val: append(x.parent(), index, val)
+        elif node.attr == "deactivate" and isinstance(node.value,
```
lin-hitonami, line 775: The logic processing `append` and `deactivate` are very similar, so maybe we can merge them together?

tests/python/test_dynamic_attributes.py
```diff
@@ -0,0 +1,37 @@
...
+            for j in range(i):
+                assert (x[i, j] == j)
+
+        for i in range(5):
+            x[i].deactivate([j for j in range(10)])
+            assert x[i, 0] == x[i, 9] == 0
+
+        for i in range(5, 10):
+            for j in range(10):
+                x[i].deactivate(j)
```
lin-hitonami, line 27: IIRC deactivate does not receive any argument and it clears the whole list?

lin-hitonami (approved):
LGTM

neozhaoliang:
> LGTM

Thanks for your review.

The newest commit has changed the `deactivate` API of a dynamic snode so that one can directly call `deactivate` without subscripting:

https://github.com/taichi-dev/taichi/pull/6512/files#diff-db5f62bc976e7c6a5b8ab4fb85480ca40d1240b626dc18330eb0094289525129R34

tests/python/test_dynamic_attributes.py
```diff
@@ -0,0 +1,62 @@
...
+            y[j] = j
+
+        ti.deactivate(y.parent(), [])
+        for j in range(n):
+            assert y[j] == 0
+
+    test()
+
+
+@test_utils.test(exclude=[ti.cc, ti.opengl, ti.vulkan, ti.metal],
```
ailzhang, line 58: shall we use arch= to explicitly list out supported arches?

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -762,16 +762,30 @@ def build_Module(ctx, node):
...
-            node.value.ptr = None
-            node.ptr = lambda val: append(x.parent(), index, val)
+        else:
+            x = build_stmt(ctx, node.value)
+
+        names = ("append", "deactivate")
+        if isinstance(x, Field) and x.parent(
+        ).ptr.type == _ti_core.SNodeType.dynamic and node.attr in names:
+
+            def build_dynamic_snode_hack():
```
ailzhang, line 775: OOC: why this local function here?

ailzhang (commented):
LGTM!

tests/python/test_dynamic_attributes.py
```diff
@@ -0,0 +1,62 @@
...
+                 default_fp=ti.f32,
+                 debug=True)
+def test_dynamic_append_length_f32():
+    _test_dynamic_append_length(ti.f32)
```
lin-hitonami, line 62: Do you mean `_test_dynamic_attributes`?

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -762,16 +762,30 @@ def build_Module(ctx, node):
...
+                node.value.ptr = None
+                if node.attr == "append":
+                    node.ptr = lambda val: append(x.parent(), index, val)
+                else:
+                    node.ptr = lambda: deactivate(x.parent(), index)
+
+            build_dynamic_snode_hack()
+
         else:
             build_stmt(ctx, node.value)
```
lin-hitonami, line 790: Emmm currently we can't call `build_stmt` on a node twice... I'll file a PR that enables us to call `build_stmt` on a node more than once immediately. Please wait for the PR.

lin-hitonami: emmm I found that I can't do that because we have static for...

lin-hitonami: We have to make sure that a node is built at most once for now...

lin-hitonami:
I've refactored the logic and let me explain it.

There are two valid cases:
1. `x[i, j].append` (where the dimension of the field (3 in this case) is equal to one plus the number of the indices (2 in this case) )
2. `x.append` (where the dimension of the field is one, equal to `x[()].append`)

For the first case, the AST(simplified) is like `node = Attribute(value=Subscript(value=x, slice=[i, j]), attr="append")`, when we `build_stmt(node.value)`(build the expression of the `Subscript` i.e. `x[i, j]`), it should build the expression of `node.value.value` (i.e. `x`) and `node.value.slice` (i.e. `[i, j]`), and raise a `TaichiIndexError` because the dimension of the field is not equal to the number of the indices. Therefore, when we meet the error, we can detect whether it is a method of Dynamic SNode and build the expression if it is by calling `build_attribute_if_is_dynamic_snode_method`. If we find that it is not a method of Dynamic SNode, we raise the error again.

For the second case, the AST(simplified) is like `node = Attribute(value=x, attr="append")`, and it does not raise error when we `build_stmt(node.value)`. Therefore, when we do not meet the error, we can also detect whether it is a method of Dynamic SNode and build the expression if it is by calling `build_attribute_if_is_dynamic_snode_method`. If we find that it is not a method of Dynamic SNode, we continue to process it as a normal attribute node.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -761,45 +762,73 @@ def build_Module(ctx, node):
...
+            indices = node.value.slice.ptr
+        else:
+            x = node.value.ptr
+            indices = []
+        if not isinstance(x, Field):
+            return False
+        if not x.parent().ptr.type == _ti_core.SNodeType.dynamic:
+            return False
+        snode = x._get_field_members()[0].ptr.snode()
+        field_dim = snode.num_active_indices()
```
strongoier, line 781:
```suggestion
        field_dim = x.snode.ptr.num_active_indices()
```
Writing in this way might be cleaner :-)

strongoier (approved):
I suggest turning some of your explanation into comments so that others can read the explanation while reading the code without diving into the PR.

PR #6519, opened 2022-11-04
[aot] Fixed vulkan texture usage export

PENGUINLIONG:
`alloc_info.usage` is the usage of memory allocation, to inform VMA how we gonna use the memory (for example, is it host accessible?) While `image_info.usage` is the one that should be exported.


ailzhang (approved):
Thanks!

Issue #5699, opened 2022-08-10
[bug] [aot] aot report bug on vulkan backend


Morcki:
**Describe the bug**
When running kernels and export aot files at the same time under vulkan backend, it reports bug. However, under cuda backend, there is no error information. In other words, cannot run kernels and export kernels at the same time.

```py
# sample code here
import taichi as ti

arch=ti.vulkan
ti.init(arch=arch)

g_init_builder = ti.graph.GraphBuilder()
mod = ti.aot.Module(arch)


@ti.kernel
def init_data(test_arr : ti.types.ndarray(field_dim=1)):
    test_arr[0] = 0

arr = ti.ndarray(ti.f32, shape=2)
sym_arr = ti.graph.Arg(ti.graph.ArgKind.NDARRAY,
                                'arr',
                                ti.f32,
                                field_dim=1,
                                element_shape=())
init_data(arr)


g_init_builder.dispatch(init_data, 
                        sym_arr)

g_init = g_init_builder.compile()

mod.add_graph('init', g_init)
mod.save("./temp", '')
```

ailzhang:
fyi after you construct and compile the graph you can run it via `g_init({'arr': arr})` and that's supposed to match the behavior in C++. 
But yea we should fix this :D

Morcki:
yeah, not urgent for this cause I can handle it in other way. but we need to fix this a bit later.

PR #6520, opened 2022-11-04
[bug] Fix that the cgraph doesn't respect the caps set by ti.aot.Module()

### LINKED ISSUE #6508: Offline cache should respect the caps set by ti.aot.Module()
An example to repro: 
Run https://github.com/taichi-dev/taichi-aot-demo/blob/master/2_mpm88/assets/mpm88.py with a simple change: `mod = ti.aot.Module(arch, caps=[])`. This essentially means when compiling the shader for AOT module, no caps should be enabled so the generated shaders shouldn't include extensions like below:

```
               OpExtension "SPV_KHR_storage_buffer_storage_class"
               OpExtension "SPV_KHR_non_semantic_info"
               OpExtension "SPV_KHR_variable_pointers"
               OpExtension "SPV_EXT_shader_atomic_float_add"
```
cc: @PGZXB @jim19930609

PGZXB:
Issue: #6508 
This PR fixes #6508 but causes the bug reported in #5699 (only) when enabling the offline cache.
Using `CacheManager` manage the compilation in `AotModuleBuilder` can fix the bug. I will fix it in next PR.

### DIFF
--- a/taichi/aot/graph_data.cpp
+++ b/taichi/aot/graph_data.cpp
@@ -1,6 +1,7 @@
 #include "taichi/aot/graph_data.h"
 #include "taichi/program/ndarray.h"
 #include "taichi/program/texture.h"
+#include "taichi/program/kernel.h"
 
 #include <numeric>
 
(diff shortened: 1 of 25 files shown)


ailzhang (approved):
Thanks a ton!

PR #6521, opened 2022-11-04
[lang] MatrixType refactor: Simplify reduction ops

strongoier:
Issue: #5819

### Brief Summary

The use of atomic ops in #6425 is unnecessary. This PR removes those usage and simplifies logic around reduction ops.


AD1024 (approved):
LGTM thanks!

jim19930609 (approved):

PR #6524, opened 2022-11-05
[misc] Add dynamic_index to offline cache key

PGZXB:
Fixed bug reported by https://github.com/taichi-dev/taichi/actions/runs/3396038407/jobs/5646653663
Repro: `python tests/run_tests.py -v -t1 -a cpu,cuda matrix_slice --with-offline-cache`

### DIFF
--- a/tests/python/test_offline_cache.py
+++ b/tests/python/test_offline_cache.py
@@ -511,9 +511,7 @@ def test_offline_cache_cleaning(curr_arch, factor, policy):
         only_init(max_size)
         for kernel, args, get_res, num_offloads in simple_kernels_to_test:
             assert kernel(*args) == test_utils.approx(get_res(*args))
-            if curr_arch in [ti.vulkan]:
-                sleep(
-                    1)  # make sure the kernels are not used in the same second
+            sleep(1)  # make sure the kernels are not used in the same second
 
     kernel_count = len(simple_kernels_to_test)
     count_of_cache_file = cache_files_cnt(curr_arch)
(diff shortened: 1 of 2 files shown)


tests/python/test_offline_cache.py
```diff
@@ -511,9 +511,7 @@ def run_simple_kernels(max_size):
         only_init(max_size)
         for kernel, args, get_res, num_offloads in simple_kernels_to_test:
             assert kernel(*args) == test_utils.approx(get_res(*args))
-            if curr_arch in [ti.vulkan]:
-                sleep(
-                    1)  # make sure the kernels are not used in the same second
+            sleep(1)  # make sure the kernels are not used in the same second
```
ailzhang, line 514: @PGZXB sorry I  just noticed this `sleep` lol, would you mind explaining more detail why this sleep is necessary?

PGZXB: Because both the `last_used_at` and `created_at` are timestamp in sec. To test the LRU/FIFO cache cleaning, we should make sure the kernels are used/created in different sec lol.

lin-hitonami: This is because The cache cleaning policy chooses which kernel to be cleaned based on the last used time of the kernels. Different kernels have different number of files, and this affects the assertion below which checks how many files remains in the directory. How about let the last used time use millisecond instead of second? @PGZXB

PGZXB: lgtm

ailzhang: aha I see, then it sounds good to me if you just add a comment about the timestamp being at second precision in this PR and switch to millisecond in a followup PR. Up to you @PGZXB

strongoier (approved):

PR #6525, opened 2022-11-06
[lang] Add fused foreach check

AD1024:
### Brief Summary

Foreach check in PR #6425 is verbose. This PR simplifies the usage.


AD1024:
Note: in the discussion of #6425, we mentioned `and` and `not` logic, but the real blocker is `Or` since we need a separate implementation to do `or` check, while the other two can be done sequentially or simply negating one of the checker funcs.

python/taichi/lang/matrix_ops_utils.py
```diff
@@ -75,19 +56,44 @@ def assert_tensor(m, msg='not tensor type: {}'):
...
 def assert_list(x, msg='not a list: {}'):
     if isinstance(x, list):
         return True, None
-    raise TaichiCompilationError(msg.format(type(x)))
+    return False, msg.format(type(x))
+
+
+def arg_foreach_check(*arg_indices, fns=[], logic='or', msg=None):
+    def check(*args, **kwargs):
+        for i in arg_indices:
```
strongoier, line 74: It seems that the logic below cannot work with multiple indices.

python/taichi/lang/matrix_ops_utils.py
```diff
@@ -75,19 +56,44 @@ def assert_tensor(m, msg='not tensor type: {}'):
     raise TaichiCompilationError(msg.format(type(m)))
```
strongoier, line 56:
```suggestion
    return False, msg.format(type(m))
```

python/taichi/lang/matrix_ops_utils.py
strongoier: Ah I'm saying the problem logic, not the `logic` argument :-) It seems that the function will return anyway after the first iteration, so if `arg_indices` contains multiple indices it may not work properly.

python/taichi/lang/matrix_ops_utils.py
```diff
@@ -90,7 +66,36 @@ def check(v):
...
+def arg_foreach_check(*arg_indices, fns=[], logic='or', msg=None):
+    def check(*args, **kwargs):
+        for i in arg_indices:
+            if i in kwargs:
+                arg = kwargs[i]
+            else:
+                arg = args[i]
+            if logic == 'or':
+                passed = False
+                for a in arg:
```
strongoier, line 81:
```suggestion
                for a in arg:
                passed = False
```

strongoier (approved):

PR #6526, opened 2022-11-07
[ci] Update slack message to include workflow_url instead of head commit

ailzhang:
I found an example of payload in
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

Issue: #


feisuzhu (approved):
LGTM!

PR #6528, opened 2022-11-07
[Lang] Matrix lib: Stop changing dimension in transpose()

strongoier:
Issue: #5819

### Brief Summary

This PR enforces that `vector.transpose()` returns a vector while `matrix.transpose()` returns a matrix, which mimics the behavior of numpy.


jim19930609 (approved):
Awesome! 

Btw, let's keep an eye on the nightly-run after this PR gets merged

PR #6531, opened 2022-11-07
[lang] MatrixType bug fix: Fix type error with struct-for loop index

jim19930609:

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -1157,7 +1157,6 @@ class ASTTransformer(Builder):
                 loop_var = build_stmt(ctx, node.iter)
                 loop_indices = expr.make_var_list(size=len(loop_var.shape),
                                                   ast_builder=ctx.ast_builder)
-
                 expr_group = expr.make_expr_group(loop_indices)
                 impl.begin_frontend_struct_for(ctx.ast_builder, expr_group,
                                                loop_var)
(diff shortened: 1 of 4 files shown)


python/taichi/lang/ast/ast_transformer.py
```diff
@@ -1157,12 +1157,24 @@ def build_struct_for(ctx, node, is_grouped):
                 loop_var = build_stmt(ctx, node.iter)
                 loop_indices = expr.make_var_list(size=len(loop_var.shape),
                                                   ast_builder=ctx.ast_builder)
+
```
strongoier, line 1160:
```suggestion
```

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -1157,12 +1157,24 @@ def build_struct_for(ctx, node, is_grouped):
...
                 expr_group = expr.make_expr_group(loop_indices)
                 impl.begin_frontend_struct_for(ctx.ast_builder, expr_group,
                                                loop_var)
-                ctx.create_variable(
-                    target, matrix.Vector(loop_indices,
-                                          dt=primitive_types.i32))
+                if impl.current_cfg().real_matrix:
+                    if len(loop_indices) > 1:
```
strongoier, line 1165: I'm wondering if it is OK to return a vector with length 1 (as the old implementation does) so that we don't need two cases here.

jim19930609: There's a special use case:
```
for I in ti.grouped(x):
    for i in range(10):
        arr[I, i, 5] = ....
```

If we return a Tensor[1 x i32] for `I`, `arr` will receiver `indices_ = [ expr(Tensor[ 1 x i32 ]), expr(i32), 5 ]` when compiling for subscription. This poses extra complicity for `_indices` expansion

jim19930609: Sure, let me adjust `indices expansion` logic when compiling for subscription.

strongoier (approved):
LGTM!

PR #6535, opened 2022-11-07
[lang] Add matrix support for Dynamic SNode

lin-hitonami:
Issue: #5420


tests/python/test_dynamic.py
```diff
@@ -283,3 +264,51 @@ def make_list():
             assert x[i, j].b == i * j * 10000 % 65536
             assert x[i, j].c == i * j * 100000000 % 4294967296
             assert x[i, j].d == i * j * 10000000000
+
+
+@test_utils.test(require=ti.extension.sparse, exclude=[ti.metal])
```
jim19930609, line 269: Can you try to add another test with `real_matrix=True, real_matrix_scalarize=True`? Something like:
```
@test_utils.test(require=ti.extension.sparse, exclude=[ti.metal], real_matrix=True, real_matrix_scalarize=True)
def test_append_matrix_scalarize():
     _test_append_matrix()
```

I can help with the fix if it fails.

tests/python/test_dynamic.py
```diff
@@ -283,3 +264,80 @@ def make_list():
...
+def test_append_matrix():
+    _test_append_matrix()
+
+
+@test_utils.test(require=ti.extension.sparse,
+                 exclude=[ti.metal],
+                 real_matrix=True,
```
ailzhang, line 298: parametrize?

jim19930609 (approved):

tests/python/test_dynamic.py
lin-hitonami: Can pytest.parametrize be used on the decorator?

ailzhang: @lin-hitonami do you mean https://github.com/taichi-dev/taichi/blob/master/tests/python/test_ndarray.py#L38? ;)

lin-hitonami: I mean we can't use it on parameters in `@test_utils.test()`. (Do you mean we can parametrize `real_matrix=True`?)

PR #6536, opened 2022-11-07
[bug] [aot] Fix texture struct for with cgraph

### LINKED ISSUE #6518: Texture structured-for doesn't work with AOT
**Describe the bug**
Writing to pixels iterating in the outermost loop doesn't results in any effective writes.

**To Reproduce**

```diff
import argparse
import os

import taichi as ti

parser = argparse.ArgumentParser()
parser.add_argument("--arch", default="vulkan", type=str)
args = parser.parse_args()

curr_dir = os.path.dirname(os.path.realpath(__file__))

if args.arch == "vulkan":
    arch = ti.vulkan
else:
    assert False
ti.init(arch=arch)

n = 320
canvas = ti.ndarray(dtype=ti.f32, shape=(n * 2, n))


@ti.func
def complex_sqr(z):
    return ti.Vector([z[0]**2 - z[1]**2, z[1] * z[0] * 2])


@ti.kernel
def fractal(t: ti.f32, canvas: ti.types.rw_texture(num_dimensions=2,
                                                   num_channels=1,
                                                   channel_format=ti.f32,
                                                   lod=0)):
-    for i, j in canvas: # doesn't work
+    for i, j in ti.ndrange(640, 320): # works
        c = ti.Vector([-0.8, ti.cos(t) * 0.2])
        z = ti.Vector([i / n - 1, j / n - 0.5]) * 2
        iterations = 0
        while z.norm() < 20 and iterations < 50:
            z = complex_sqr(z) + c
            iterations += 1

        color = 1 - iterations * 0.02
        canvas.store(ti.Vector([i, j]), ti.Vector([color, color, color, 1.0]))


sym_t = ti.graph.Arg(ti.graph.ArgKind.SCALAR,
                     "t",
                     ti.f32,
                     element_shape=())
sym_canvas = ti.graph.Arg(ti.graph.ArgKind.RWTEXTURE,
                          'canvas',
                          channel_format=ti.f32,
                          shape=(n * 2, n),
                          num_channels=1)

gb = ti.graph.GraphBuilder()
gb.dispatch(fractal, sym_t, sym_canvas)
graph = gb.compile()

mod = ti.aot.Module(ti.vulkan)
mod.add_graph('fractal', graph)

save_dir = os.path.join(curr_dir, "fractal")
os.makedirs(save_dir, exist_ok=True)
mod.save(save_dir, "fractal")
print('AOT done')

```

ailzhang:
Fixes #6518

We should handle graph args based on their symbolic_arg type instead of runtime type.

Issue: #

### DIFF
--- a/taichi/aot/graph_data.cpp
+++ b/taichi/aot/graph_data.cpp
@@ -23,7 +23,8 @@ void CompiledGraph::run(
       TI_ERROR_IF(found == args.end(), "Missing runtime value for {}",
                   symbolic_arg.name);
       const aot::IValue &ival = found->second;
-      if (ival.tag == aot::ArgKind::kNdarray) {
+      if (symbolic_arg.tag == aot::ArgKind::kNdarray) {
+        TI_ASSERT(ival.tag == aot::ArgKind::kNdarray);
         Ndarray *arr = reinterpret_cast<Ndarray *>(ival.val);
 
         TI_ERROR_IF(arr->get_element_shape() != symbolic_arg.element_shape,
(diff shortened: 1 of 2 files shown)


taichi/aot/graph_data.cpp
```diff
@@ -23,7 +23,7 @@ void CompiledGraph::run(
       TI_ERROR_IF(found == args.end(), "Missing runtime value for {}",
                   symbolic_arg.name);
       const aot::IValue &ival = found->second;
```
PENGUINLIONG, line 25: should we first assert the types are the same?

PENGUINLIONG (changes requested):

taichi/aot/graph_data.cpp
ailzhang: good point! added the asserts below since the runtime types may map to different compile time type. (like texture -> rwtexture & texture)

PENGUINLIONG (approved):

PR #6549, opened 2022-11-09
[aot] C-API device capability query

PENGUINLIONG:
Issue: #

### DIFF
--- a/docs/lang/articles/c-api/taichi_core.md
+++ b/docs/lang/articles/c-api/taichi_core.md
@@ -325,7 +325,6 @@ A collection of Taichi kernels (a compute graph) to launch on the offload target
 ```c
 // enumeration.error
 typedef enum TiError {
-  TI_ERROR_TRUNCATED = 1,
   TI_ERROR_SUCCESS = 0,
   TI_ERROR_NOT_SUPPORTED = -1,
   TI_ERROR_CORRUPTED_DATA = -2,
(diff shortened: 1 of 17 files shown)


taichi/inc/rhi_constants.inc.h
```diff
@@ -7,6 +7,7 @@
 // should not.
 #ifdef PER_DEVICE_CAPABILITY
 // SPIR-V Caps
+PER_DEVICE_CAPABILITY(reserved)
```
PENGUINLIONG, line 10: Breaking change because the 0 code is better reserved to helps us prefix capability from different backends.

docs/lang/articles/c-api/taichi_core.md
```diff
@@ -325,7 +325,6 @@ A collection of Taichi kernels (a compute graph) to launch on the offload target
 ```c
 // enumeration.error
 typedef enum TiError {
-  TI_ERROR_TRUNCATED = 1,
```
PENGUINLIONG, line 328: Success codes feel unnecessary at this point so it's removed and are confusing among these error codes.

taichi/aot/module_loader.h
```diff
@@ -94,6 +94,11 @@ class TI_DLL_EXPORT Module {
     TI_NOT_IMPLEMENTED;
   }
 
+  virtual const DeviceCapabilityConfig &get_required_caps() const {
+    static DeviceCapabilityConfig default_cfg;
```
ailzhang, line 98: Correct me if I'm missing anything, `default_cfg` should respect the caps specified by users but I didn't see that logic in this PR?

ailzhang (approved):
Thanks!

PR #6554, opened 2022-11-09
[aot] Deprecate filename arg in ti.aot.Module.save()

ailzhang:
Issue: #

### Brief Summary
`filename` argument in `ti.aot.Module.save()` has been no-op for a long time. This PR deprecates it and annouce that it'll be removed in v1.4.0 release. 

Also added a test file called `test_deprecation.py` which is designed to be the place to check all DeprecationWarnings so that before every release we can just check here to see if we can safely remove any deprecated apis.

### DIFF
--- a/python/taichi/aot/module.py
+++ b/python/taichi/aot/module.py
@@ -1,3 +1,4 @@
+import warnings
 from contextlib import contextmanager
 from glob import glob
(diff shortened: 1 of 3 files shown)


PENGUINLIONG (approved):
lgtm

jim19930609 (approved):

PR #6575, opened 2022-11-10
[lang] Unify Cuda and CPU Spmv API

FantasyVR:
Issue: #2906

### Brief Summary
The previous API of spmv is `A.spmv(x, y)` which means `y=A@x`. 
Now the spmv function is removed. We can directly use `y = A @ x` to compute the results on Cuda backend.


Hanke98 (approved):
Great! LGTM!

strongoier (approved):

Issue #6179, opened 2022-09-27
fflush after cuStreamSynchronize does not flush the contents of in-kernel printf in time


lin-hitonami:
**Describe the bug**
https://github.com/taichi-dev/taichi/pull/6173#issuecomment-1259105471

PR #6577, opened 2022-11-11
[bug] Sync after executing kernels which have print inside

### LINKED ISSUE #6540: Wrong sequence between the output of Python and Taichi kernel
**Describe the bug**
The output of Python and Taichi kernel are in the wrong sequence.

**To Reproduce**

```py
import taichi as ti

ti.init(ti.cuda)

@ti.kernel
def foo():
    print("inside kernel")

foo()
print("outside kernel")
```

**Log/Screenshots**

```
[Taichi] version 1.3.0, llvm 10.0.0, commit 7132459a, linux, python 3.9.12
outside kernel
inside kernel
```

lin-hitonami:
Issue: fixes #6540

### DIFF
--- a/python/taichi/lang/ast/ast_transformer.py
+++ b/python/taichi/lang/ast/ast_transformer.py
@@ -513,6 +513,9 @@ class ASTTransformer(Builder):
         keywords = dict(ChainMap(*[keyword.ptr for keyword in node.keywords]))
         func = node.func.ptr
 
+        if id(func) in [id(print), id(impl.ti_print)]:
+            ctx.func.has_print = True
+
         if isinstance(node.func, ast.Attribute) and isinstance(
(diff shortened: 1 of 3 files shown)


ailzhang (approved):
Thanks! 
(ideally we should have a test covering this but not sure how to do that...

lin-hitonami:
> Thanks! (ideally we should have a test covering this but not sure how to do that...

We can capture the output using the pytest capfd fixture, but this doesn't function well for CUDA on windows...  https://github.com/taichi-dev/taichi/issues/6179

bobcao3:
This feels like the wrong way to do things, shouldn't the runtime itself cares for the synchronization of its operations? This is more like a bug with the CUDA host runtime

PR #6578, opened 2022-11-11
[lang] Unify Cuda and CPU linear solve API

FantasyVR:
Issue: #2906


tests/python/test_sparse_linear_solver.py
```diff
@@ -98,18 +98,25 @@ def init_b(b: ti.types.ndarray(), nrows: ti.i32):
...
     solver = ti.linalg.SparseSolver()
-    solver.solve_cu(A_ti, b, x_ti)
+    x_ti = solver.solve_cu(A_ti, b)
     ti.sync()
 
     # solve Ax=b using numpy
     b_np = b.to_numpy()
     x_np = np.linalg.solve(A_psd, b_np)
-    assert (np.allclose(x_ti.to_numpy(), x_np, rtol=5e-3))
+    assert (np.allclose(x_ti.to_numpy(), x_np, rtol=1.0e-2))
```
Hanke98, line 107: Why is the `rtol` so large?

FantasyVR: It has a chance to fail on my machine if `rtol` is too small, I reset it using the previous value `5e-3` and see if it could pass the CI test.

python/taichi/linalg/sparse_solver.py
```diff
@@ -19,6 +19,7 @@ class SparseSolver:
         ordering (str): The method for matrices re-ordering.
     """
     def __init__(self, dtype=f32, solver_type="LLT", ordering="AMD"):
+        self.matrix_ = None
```
strongoier, line 22: nit: use `_matrix` in python

strongoier (approved):

Hanke98 (approved):
LGTM!

PR #6579, opened 2022-11-11
[autodiff] Clear adjoint after global store

erizmr:
To avoid invalid grad accumulation, adjoints of fields on left-hand sides of assignments (GlobalStoreStmt) need to be reset to zero after the corresponding adjoint assignments.

### DIFF
--- a/tests/python/test_loop_grad.py
+++ b/tests/python/test_loop_grad.py
@@ -27,8 +27,8 @@ def test_loop_grad():
     func.grad()
 
     for k in range(n):
-    # The grad of fields on left-hand sides of assignments (GlobalStoreStmt) need to be reset to zero after the corresponding adjoint assignments.
-    # Therefore, only the grad of the element with index 0 at second dimension is preserved here.
+        # The grad of fields on left-hand sides of assignments (GlobalStoreStmt) need to be reset to zero after the corresponding adjoint assignments.
+        # Therefore, only the grad of the element with index 0 at second dimension is preserved here.
         assert x[k, 0] == 2**0 * k
         assert x.grad[k, 0] == 2**(m - 1 - 0)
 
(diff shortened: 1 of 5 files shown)


tests/python/test_offline_cache.py
```diff
@@ -327,12 +327,12 @@ def compute_y():
             enable_fallback=False,
             **current_thread_ext_options())
     assert added_files(curr_arch) == expected_num_cache_files(
-        curr_arch, [1] * 8)
+        curr_arch, [1] * 9)
```
ailzhang, line 330: why did this change btw?

tests/python/test_loop_grad.py
```diff
@@ -27,9 +27,8 @@ def func():
     func.grad()
 
     for k in range(n):
-        for i in range(m):
-            assert x[k, i] == 2**i * k
-            assert x.grad[k, i] == 2**(m - 1 - i)
+        assert x[k, 0] == 2**0 * k
+        assert x.grad[k, 0] == 2**(m - 1 - 0)
```
ailzhang, line 31: `2 ** 0 *k` & `2**(m - 1- 0)` are weird expressions to show up here :P 
If you write it in this form, it'd be nice to have a comment explaining why ;)

ailzhang (commented):
Nice!

tests/python/test_offline_cache.py
erizmr: Since a `self.loss.grad.fill(1.0)` is added, an additional kernel call is taken into count.

tests/python/test_loop_grad.py
erizmr: Sure. Have added some comments.

erizmr:
\rebase

ailzhang (approved):
Thanks!

PR #6587, opened 2022-11-14
[ci] Terminate workers even earlier when test fails

feisuzhu:
Issue: #


ailzhang (approved):
Thanks!
nit: does that mean we don't allow any reruns? if so we should change the configurations in unix_test.sh as well?

PR #6597, opened 2022-11-14
[ci] Add C++ tests on AMDGPU RHI

GaleSeLee:

Add some unit tests on the AMDGPU RHI level


tests/cpp/backends/amdgpu_device_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+  EXPECT_TRUE(device != nullptr);
+  taichi::lang::Device::AllocParams params;
+  params.size = 1048576;
+  params.host_read = false;
+  params.host_write = false;
+  const taichi::lang::DeviceAllocation device_alloc =
+      device->allocate_memory(params);
+
+  // The purpose of the device_alloc_guard is to rule out double free
+  const taichi::lang::DeviceAllocationGuard device_alloc_guard(device_alloc);
```
ailzhang, line 24: nit: would you mind elaborating when the double free happens? Thanks!

ailzhang (commented):
Nice!! 
let's make sure we add an amd machine in ci before merging this PR tho ;)

tests/cpp/backends/amdgpu_device_test.cpp
GaleSeLee: This is to prevent manual dealloc_memory, but of course there is no double free problem here
I write this UT with reference to DX12, but I found redundancy in this UT and [this pr](https://github.com/taichi-dev/taichi/pull/6552)

turbo0628: Yes, this is from the DX12 test, and I didn't find any double free when migrating this test to vulkan and opengl.

tests/cpp/backends/amdgpu_device_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+  device->unmap(device_alloc);
+}
+
+TEST(AMDGPU, ImportMemory) {
+  std::unique_ptr<amdgpu::AmdgpuDevice> device =
+      std::make_unique<amdgpu::AmdgpuDevice>();
+  EXPECT_TRUE(device != nullptr);
+
+  int *ptr = nullptr;
+  AMDGPUDriver::get_instance().malloc_managed((void **)&ptr, 400,
```
jim19930609, line 49: Let's avoid using magic numbers. For example, rearrange these "400"s as follow:

```
size_t mem_size = 400;

device->import_memory(ptr, mem_size);

params.size = 400;
...
```

tests/cpp/backends/amdgpu_device_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+      device->import_memory(ptr, 400);
+
+  for (int i = 0; i < 100; i++) {
+    ptr[i] = i;
+  }
+
+  taichi::lang::Device::AllocParams params;
+  params.size = 400;
+  params.host_read = false;
+  params.host_write = false;
```
jim19930609, line 61: You'll have to at least set `host_read` to `true` in order to read from the mapped memory. (However, `host_read` isn't implemented nor validated on some backends including CUDA, which is why it still works even if set incorrectly.)

jim19930609 (approved):

tests/cpp/backends/amdgpu_device_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+}
+
+TEST(AMDGPU, CreateContextAndGetMemInfo) {
+  auto total_size = AMDGPUContext::get_instance().get_total_memory();
+  auto free_size = AMDGPUContext::get_instance().get_free_memory();
+  EXPECT_GE(total_size, free_size);
+  EXPECT_GE(free_size, 0);
+}
+
+TEST(AMDGPU, LaunchKernel) {
```
jim19930609, line 88: Let's not merge empty tests

PR #6605, opened 2022-11-15
[lang] Merge triplets  in the same position when building GPU sparse matrix

FantasyVR:
Issue: #2906 

### Brief Summary
When building GPU sparse matrix, cuSparse API requires three separated arrays: row index ptr, col index ptr, and values ptr. However, the sparse matrix builder only uses one ndarray to store all triplets, the memory layout is like: [row, col, value, row, col, value, ...]. In this pr, I retrieve all data from ndarray and merge all triplets in the same position of the sparse matrix. Then, all triplets are stored in three separate arrays. At last, these three arrays are used to build sparse matrix using cuSparse API.


taichi/program/sparse_matrix.cpp
```diff
@@ -145,6 +146,55 @@ std::unique_ptr<SparseMatrix> SparseMatrixBuilder::build() {
...
+  int count = 0;
+  for (auto entry : entries) {
+    auto [row, col, value] = entry.second;
+    row_host[count] = row;
+    col_host[count] = col;
+    value_host[count] = value;
+    count++;
+  }
+  void *row_device = nullptr, *col_device = nullptr, *value_device = nullptr;
+  CUDADriver::get_instance().malloc(&row_device, entry_size * sizeof(int));
```
strongoier, line 182: free?

Hanke98 (approved):
LGTM

strongoier (approved):

PR #6613, opened 2022-11-15
[lang] MatrixType bug fix: Fix result type of matmul

### LINKED ISSUE #6600: Example 22 broken after PR #6425
After this PR #6425, the exmaple 22 "Karman vortex street" is broken, this example still runs but the animation doesn't show correctly. And there is a warning in the `matrix_ops.py`:

```
[W 11/15/22 10:35:10.035 2582858] [type_check.cpp:type_check_store@36] [$32852] Atomic add may lose precision: i32 <- f32
File "/home/zhao/taichi/python/taichi/lang/matrix_ops.py", line 347, in _matmul_helper:
                result[i] += x[i, j] * y[j]
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The parent commit does not have this problem.

strongoier:
Issue: #5819, fix #6600

### Brief Summary

This PR also refactors the implementation of `matmul`, `fill` and `diag` a bit. Now `matmul` of two vectors will return a scalar just as numpy does, and `fill` will not have a return value.


jim19930609 (approved):
Thanks!

Issue #6590, opened 2022-11-14
Let real function/kernel support returning a struct/multiple values


lin-hitonami:
**Concisely describe the proposed feature**
Currently, the Taichi function supports returning a arbitrary number of return values, and the return values can be scalars, matrices and structs, and the Taichi function is inlined into the Taichi kernel at compile time. 

However, the Taichi kernel only supports returning a single return value, and the return value can only be a scalar or a matrix. 

The WIP real function (#602) is a Taichi function that does not inline into the kernel. The kernel calls it instead of inlining it. In LLVM-based backends, the real function receives an argument with type `RuntimeContext` where the arguments and the address of return value buffer are stored, same as the Taichi kernel. Currently, the real function only supports a single scalar return value.

I would like to let the real function and the Taichi kernel support returning a struct, and then let them support multiple return values (we can construct a struct containing the return values).

**Describe the solution you'd like (if any)**
First, we can implement a `StructType` in our IR, and get its respective `llvm::StructType`.

Then, we can construct return value to the `llvm::StructType` and store the pointer to it in the `RuntimeContext`. We can get the address of the elements of the return value via the GEP instructions.

Here comes a problem: where should we store the return value? 

There are two places we can store it: the stack or the heap. I suggest we store it on the stack if the size of the return value is very small, otherwise we should store it on the heap, because the stack of CUDA is small.

If we store it on the heap, we may want an allocator using preallocated memory because malloc on CUDA is slow. We may use the `NodeManager` to allocate small items which have the same size.

PR #6614, opened 2022-11-15
[llvm] Let real function support returning struct

lin-hitonami:
Issue: #602 #6590

### Brief Summary
Only supports scalar struct (every element in the struct is a scalar) for now. 

This PR does the following things:
1. Let `FuncCallStmt` return the `real_func_ret_struct *` result buffer instead of returning the return value directly.
2. Add `GetElementStmt` and `GetElementExpression` to get the i-th return value in a result buffer
3. Add `StructType.from_real_func_ret` to construct the returned struct to the `StructType` in Python

Will add support for nested struct and matrix in struct in the following PRs.


bobcao3:
Using U64 array types and doing pointer conversion is really bad. This prohibits the downstream compiler to do any sort of optimization on this array (because of pointer conversions), and it uses a lot of register space & causes larger spilling issue.

lin-hitonami:
> Using U64 array types and doing pointer conversion is really bad. This prohibits the downstream compiler to do any sort of optimization on this array (because of pointer conversions), and it uses a lot of register space & causes larger spilling issue.

Yeah, please look at the **Possible Improvement** section of #6590 , I will compile it into a real struct in the future.

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2823,21 +2823,29 @@ void TaskCodeGenLLVM::visit(FuncCallStmt *stmt) {
...
   }
   llvm::Value *result_buffer = nullptr;
-  if (stmt->ret_type->is<PrimitiveType>() &&
-      !stmt->ret_type->is_primitive(PrimitiveTypeID::unknown)) {
-    result_buffer = builder->CreateAlloca(tlctx->get_data_type<uint64>());
-    call("RuntimeContext_set_result_buffer", new_ctx, result_buffer);
-    call(llvm_func, new_ctx);
-    auto *ret_val_u64 = builder->CreateLoad(
+  result_buffer = builder->CreateAlloca(llvm::ArrayType::get(
+      tlctx->get_data_type<uint64>(), stmt->func->rets.size()));
```
bobcao3, line 2827: Wouldn't the process of making a struct type be this:

```c++
std::vector<llvm::Type *> member_types(stmt->func->rets.size());
for (int i = 0; i < stmt->func->rets.size(); i++) {
  member_types[i] = tlctx->get_data_type(stmt->func->rets[i].dt);
}
llvm::StructType *ret_struct_type = llvm::StructType::get(tlctx->get_this_thread_context(), llvm::ArrayRef(member_types));
result_buffer = builder->CreateAlloca(ret_struct_type);
```

And I think the rest of the code should just work after removing all the bitcasts. GEP works the same way in structs iirc

lin-hitonami: But it will become a bit complicated when there are nested structs (if we do not flatten the structs). I would like to add them properly in another PR.

taichi-gardener:

lin-hitonami:
Constructed the result buffer directly into a `llvm::StructType` thanks to @bobcao3 .

ailzhang (approved):
Awesome thanks!

PR #6618, opened 2022-11-16
[aot] Test for AOT device capability

PENGUINLIONG:
Issue: #


tests/python/test_aot.py
```diff
@@ -589,12 +589,32 @@ def test(a: ti.types.ndarray(), c: ti.u8):
 
     g.run({'a': a, 'c': c})
 
-    m = ti.aot.Module(caps=['spirv_has_int8'])
+    m = ti.aot.Module(caps=[ti.DeviceCapability.spirv_has_int8])
```
ailzhang, line 592: Nice! btw can we have another test to make sure when a weird string or enum is passed, we throw proper error message to users?

ailzhang (approved):
Thanks!! Had a minor nit inline but that can wait for a followup PR.

PR #6619, opened 2022-11-16
[misc] Cleanup unused files in root dir

ailzhang:
Issue: #

### Brief Summary
I believe these files are no longer used so we'd better just remove them to avoid confusion.


feisuzhu (approved):
:shipit:

Issue #6572, opened 2022-11-10
Ndarray type annotations & error message refinement


ailzhang:
We had a discussion about refining ndarray definition and type annotation in https://github.com/taichi-dev/taichi/discussions/4085 earlier this year and it's now a hard blocker if we want to properly release ndarray to users. Based on the discussions with @k-ye and @strongoier , we'll list  out the desired APIs as well as some examples of bad error messages we wanna fix here. 

## Type annotation

Currently [`ti.types.ndarray()` can take in `dtype/element_dim/element_shape/field_dim` as optional arguments](https://github.com/taichi-dev/taichi/blob/master/python/taichi/types/ndarray_type.py#L24-L27), we'd like to: 

- [ ] Rename `field_dim` to `ndim` (following numpy etc)
- [ ] Get rid of `element_dim` and  `element_shape` in the type annotation

And the desired ndarray type annotation API may only takes two optional argument `dtype` and `ndim`, which looks like below:

```
arr: ti.types.ndarray()
arr: ti.types.ndarray(dtype=ti.f32)
arr: ti.types.ndarray(dtype=ti.math.vec3)
arr: ti.types.ndarray(ndim=2)
arr: ti.types.ndarray(dtype=ti.f32, ndim=2)
```

## Error message
Currently we've got a lot to improve in terms of telling users what's wrong with their ndarray usage, listing a few examples below:

A running demo can be found at https://gist.github.com/ailzhang/d207a53c26720cccb7e7c5a9687009c2

```
tp_ivec3 = ti.types.vector(3, ti.i32)
tp_ivec2 = ti.types.vector(2, ti.i32)
tp_vec3 = ti.types.vector(3, ti.f32)

x = ti.ndarray(ti.f32, shape=(12, 13))
y = ti.ndarray(tp_ivec3, shape=(12,4))
z = ti.ndarray(ti.i32, shape=(12, 12))
m = ti.ndarray(tp_ivec2, shape=(2, 3))
n = ti.ndarray(ti.f32, shape=(12,))
k = ti.ndarray(tp_vec3, shape=(12,))

@ti.kernel
def test1(arr: ti.types.ndarray()):
    for I in ti.grouped(arr):
        arr[I] = 1

test1(x)
print(x.to_numpy())

# Case 1: The following currently errors out since the kernel implicitly assumes scalar element. We need better error message to users.
#test1(y)

@ti.kernel
def test2(arr: ti.types.ndarray(dtype=ti.f32)):
[...]

PR #6620, opened 2022-11-16
[Lang] [ci] Remove element shape and element dim in unit tests

turbo0628:
Issue: #6572

Major changes:
* Replace all the element_dim and element_shape arguments with matrix types in ndarray unit tests.
* [BREAK CHANGE] raises a TypeError when external array doesn't math matrix element type shape. Before this PR we don't check the consistency and silently trigger recompilation.
* Throw a warning if user specifies element_dim and element_shape

### DIFF
--- a/python/taichi/types/ndarray_type.py
+++ b/python/taichi/types/ndarray_type.py
@@ -1,5 +1,8 @@
 from taichi.lang.enums import Layout
-from taichi.types.compound_types import TensorType
+from taichi.types.compound_types import CompoundType, TensorType
+
+# We cannot use iomport Vector/MatrixType due to circular import
+# from taichi.lang.matrix import VectorType, MatrixType
 
 
 class NdarrayTypeMetadata:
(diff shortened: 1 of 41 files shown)


python/taichi/types/ndarray_type.py
```diff
@@ -25,6 +30,12 @@ def __init__(self,
                  element_dim=None,
                  element_shape=None,
                  field_dim=None):
+        # The element shape are deprecated. Use dtype to manage element-wise arguments.
+        if element_dim is not None or element_shape is not None:
+            warnings.warn(
```
ailzhang, line 35: nit: we'd better say explicitly when it'll be removed and add a test in `test_deprecations.py` to track ;)

ailzhang: Actually shall we remove changes in this file to the next PR? 
Since changes in annotations can go in immediately, but ideally there should be a separate PR with that updates the logic here. For example, after throwing the warning we should just translate `element_shape` and `element_dim` into corresonding `dtype`s and update the check logic to handle `dtype` only. 
Then it's much easier to maintain and delete deprecated `element_dim/element_shape` in the future. ;)

ailzhang (commented):
Thanks a lot! Had an inline comment suggesting splitting this PR into two but happy to discuss that if needed :D

python/taichi/types/ndarray_type.py
turbo0628: Great, then I should move this warning in the next PR that removes internal use of element shape!

ailzhang (approved):
LGTM!

Issue #6404, opened 2022-10-22
Any plan for improving type hint


noTmiHoYo:
I am using vscode with pylance. Autocomplete and typehint almost won't work..

Are there any one are current working on this? Or should I create a PR to fix some of these problems?

Several examples.

## Example 1
PyLance cannot figure out what does `VectorType` stand for..
<img width="410" alt="image" src="https://user-images.githubusercontent.com/112957922/197333648-127e6e6c-f5f4-491c-98b3-64e39788da27.png">

## Example 2
<img width="190" alt="image" src="https://user-images.githubusercontent.com/112957922/197333734-cca73b61-d98f-448a-9620-4de6706f253f.png">
possible fix for example 1&2: I found these problems are fixed after I add some import statement in __init__.py.

## Example 3
`ti.math.vec2(1, 2).x` is valid. But no completion is provided. 
possible fix: provide special type hint for vec2, vec3, vec4, ivec2, ivec3, etc..

## Example 4
though `ti.math.vec2` can be used as type when define a taichi struct, but it cannot be used in `isinstance(ti.math.vec2(1, 2), ti.math.vec2)` because it is not a real python type.
possible fix: How about make `ti.types` types real types inside C code ?

## Example #5
when i try to access a member variable of a object of classes decorated by `@ti.dataclass`, no auto completion is provided.
possible fix: [PEP 681](https://peps.python.org/pep-0681/)

neozhaoliang:
Hi, you are welcome to contribute! The suggestions for these examples are very valuable.

Let me explain why `vec2.x` does not have a completion: this attribute is added to the Matrix class via a decorator: https://github.com/taichi-dev/taichi/blob/master/python/taichi/lang/matrix.py#L370
and hence the code autocompletion needs to "run" this decorator to see it, which is infeasible.

PR #6621, opened 2022-11-16
[doc] Add faq about how to use code completion in vscode

ailzhang:
Issue: #6404 

### Brief Summary 

This PR doesn't solve the issue in #6404 but it helps a bit for beginners. 

Our type hint isn't perfect yet, but it should at least work for the python apis. ;) Adding a faq so that users can find them.


neozhaoliang (approved):
Thanks!

Issue #5662, opened 2022-08-07
[RFC] Argument pack API


bobcao3:
Similar to how we currently declare arguments for kernels in Taichi CGraph, we can use something similar for the regular kernel launch (with some additions). This API can be used to simplify kernel arguments and passing of resource for larger production projects, such as a raytracer or a production level PBD simulator.

The idea is that the user can create "argument templates", and "argument packs" that instantiates a template (or construct one implicitly by its content). When launching kernels, the argument templates can be used as the argument type, and the kernel should be able to access its members using subscript or `.member`

In addition, we should support cascading argument packs just like structs in C, and we should let the ordinary kernel arguments syntax be an implicit argument pack. This way we get a consistent semantics for all arguments.

Here is an example:
```python
# Create templates
view_params_tmpl = ti.types.argpack({ view_mtx : ti.mat4, proj_mtx : ti.mat4, far : ti.f32 })
scene_data_tmpl = ti.types.argpack({ n_objects : ti.i32, verticies : ti.types.resource_heap(ti.ndarray(...), dynamic_size=True), textures : ti.types.resource_heap(ti.types.texture(2), dynamic_size=True) })

# Instantiate packs
view_params = view_params_tmpl(view_mtx, ...)
textures = ti.arg_array()
textures.append(...)

...
scene_data = scene_data_tmpl(n_objects, verticies, textures)

renderA(view_params, scene_data, 2.0) # -> creates implicit argpack and its template { view_params :  view_params_tmpl, scene_data : scene_data_tmpl, c : ti.f32 }
renderB(scene_data) # -> always creates implicit argpack for kernel arguments { scene_data : scene_data_tmpl }
```

bobcao3:
I think we can approach this in steps. First step is to support basic argpacks without nesting and resource heaps. Second step is to support resource heap, and the third step would be to support nesting and implicit argument templates

listerily:
Argument pack is implemented in these Pull Requests. 

* #8273
* #8267
* #8266
* #8265
* #8264
* #8263
* #8262
* #8259
* #8257
* #8256
* #8255
* #8241
* #8240
* #8178
* #8176
* #8216
* #8204

However, this requirement is not implemented yet.
> All parameters of a kernel will be implicitly constructed as an ArgPack; same for return values

listerily:
Should we mark Argument Pack is completed, ignoring this requirement?

> All parameters of a kernel will be implicitly constructed as an ArgPack; same for return values

[...]

PR #6626, opened 2022-11-16
[refactor] Simplify callback logic for torch & paddle args

ailzhang:
related: #5662

We need runtime arg handling to be more modular to prepare for struct args. Will send a few refactor PRs on top of this one.

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -588,64 +588,6 @@ class Kernel:
             taichi_kernel)
         self.compiled_kernels[key] = taichi_kernel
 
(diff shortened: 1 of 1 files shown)


lin-hitonami (commented):
Why do we need to inline those functions into `get_function_body`? `get_function_body` is long enough, and writing functions that is too long is not a good coding style. We should split it into shorter functions if it is possible.

ailzhang:
@lin-hitonami Yup exactly, I plan to move the entire arg handling logic out of get_function_body but would like to do that in small steps :D

lin-hitonami (approved):
LGTM

Issue #6581, opened 2022-11-11
GGUI Vulkan error: Assertion failed: info->MinImageCount >= 2


amakiir:
Hi,
I get the following error when trying to run the ggui examples on Windows 10 with Vulkan:
```
Assertion failed: info->MinImageCount >= 2, file C:/Users/Administrator/actions-runner/_work/taichi/taichi/external/imgui/backends/imgui_impl_vulkan.cpp, line 1040
````

# To Reproduce
- latest Vulkan runtime from https://vulkan.lunarg.com/sdk/home
- python=3.10.6
- taichi=1.2.1

GPU: Radeon R9 280X

# Log/Screenshots
Output from trying to run `fractal3d_ggui.py`:
```
[Taichi] version 1.2.1, llvm 10.0.0, commit 12ab828a, win, python 3.10.6
[Taichi] Starting on arch=vulkan
[I 11/11/22 17:10:31.899 10228] [C:/Users/Administrator/actions-runner/_work/taichi/taichi/taichi/rhi/vulkan/vulkan_device_creator.cpp:pick_physical_device@394] Found Vulkan Device 0 (AMD Radeon R9 200 Series)
[I 11/11/22 17:10:31.899 10228] [C:/Users/Administrator/actions-runner/_work/taichi/taichi/taichi/rhi/vulkan/vulkan_device_creator.cpp:find_queue_families@148] Async compute queue 1, graphics queue 0
[I 11/11/22 17:10:31.900 10228] [C:/Users/Administrator/actions-runner/_work/taichi/taichi/taichi/rhi/vulkan/vulkan_device_creator.cpp:find_queue_families@148] Async compute queue 1, graphics queue 0
[I 11/11/22 17:10:31.900 10228] [C:/Users/Administrator/actions-runner/_work/taichi/taichi/taichi/rhi/vulkan/vulkan_device_creator.cpp:create_logical_device@462] Vulkan Device "AMD Radeon R9 200 Series" supports Vulkan 0 
version 1.2.170
[I 11/11/22 17:10:31.991 10228] [C:/Users/Administrator/actions-runner/_work/taichi/taichi/taichi/rhi/vulkan/vulkan_device.cpp:create_swap_chain@2416] Creating suface of 1280x720
Assertion failed: info->MinImageCount >= 2, file C:/Users/Administrator/actions-runner/_work/taichi/taichi/external/imgui/backends/imgui_impl_vulkan.cpp, line 1040
```

# ti diagnose
```
[...]

PR #6629, opened 2022-11-16
[gui] Tolerate swapchain image count

PENGUINLIONG:
Issue: #6581

### DIFF
--- a/taichi/rhi/vulkan/vulkan_device.cpp
+++ b/taichi/rhi/vulkan/vulkan_device.cpp
@@ -2438,7 +2438,7 @@ void VulkanSurface::create_swap_chain() {
   createInfo.pNext = nullptr;
   createInfo.flags = 0;
   createInfo.surface = surface_;
-  createInfo.minImageCount = capabilities.minImageCount;
+  createInfo.minImageCount = std::min<uint32_t>(capabilities.maxImageCount, 3);
   createInfo.imageFormat = surface_format.format;
   createInfo.imageColorSpace = surface_format.colorSpace;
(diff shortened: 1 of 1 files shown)


taichi/rhi/vulkan/vulkan_device.cpp
```diff
@@ -2438,7 +2438,7 @@ void VulkanSurface::create_swap_chain() {
   createInfo.pNext = nullptr;
   createInfo.flags = 0;
   createInfo.surface = surface_;
-  createInfo.minImageCount = capabilities.minImageCount;
+  createInfo.minImageCount = std::min<uint32_t>(capabilities.maxImageCount, 3);
```
ailzhang, line 2441: nit: would you mind adding a comment explaining why `3` is chosen :P? Thanks!

ailzhang (approved):
Thanks!

taichi/rhi/vulkan/vulkan_device.cpp
PENGUINLIONG: @ailzhang https://www.intel.com/content/www/us/en/support/articles/000006930/graphics.html

PENGUINLIONG: it's a rather common practice actually, you know it if you are a gamer 😎

amakiir:
Can I ask why the commit was reverted? Because the commit fixed the `Assertion failed: info->MinImageCount >= 2` bug for me. This means the bug is not fixed in the current (1.3.0) release.

PR #6630, opened 2022-11-16
[Lang] Remove sourceinspect dependency

neozhaoliang:
This PR replaces, and also fixes a bug in the `sourceinpect` module (cannot use cell magics %%time and %%timeit in jupyter notebooks). The dependency on `sourceinspect` is reserved and will be removed in future releases (if the current fix works fine)


python/taichi/lang/_wrap_inspect.py
```diff
@@ -0,0 +1,143 @@
...
+        raise IOError(
+            'Object `{obj.__name__}` is not defined in a .blend file!')
+
+    lines = bpy.data.texts[text_name].as_string()
+    # Now we have found the filename and code lines.
+    # We first check if they are already cached, to avoid file io in each query.
+    try:
+        filename = _blender_findsource._saved_inspect_cache[lines]  # pylint: disable=no-member
+    except KeyError:
+        fd, filename = tempfile.mkstemp(prefix='_Blender_',
```
ailzhang, line 48: who's responsible for removing these temp files?

python/taichi/lang/_wrap_inspect.py
```diff
@@ -0,0 +1,143 @@
...
+    def __exit__(self, *_):
+        inspect.findsource = _builtin_findsource
+
+
+def _getsourcelines(obj):
+    try:
+        with _InspectContextManager():
+            return inspect.getsourcelines(obj)
+    except:
+        warn(f'Using sourceinspect to get source lines of {obj}')
```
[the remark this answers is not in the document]
ailzhang: If we just automatically fallback to without asking users to open an issue, we'll probably never know when a bug happens and it'll bite us again when we remove source inspect.

python/taichi/lang/_wrap_inspect.py
```diff
@@ -0,0 +1,143 @@
...
+        with _InspectContextManager():
+            ret = inspect.getsourcefile(obj)
+            if ret is None:
+                try:
+                    ret = inspect.getfile(obj)
+                except:
+                    pass
+            return ret
+    except:
+        warn(f'Using sourceinspect to get source file of {obj}')
```
ailzhang, line 142: ditto

python/taichi/lang/_wrap_inspect.py
```diff
@@ -0,0 +1,143 @@
...
+            return inspect.getsourcelines(obj)
+    except:
+        warn(f'Using sourceinspect to get source lines of {obj}')
+        return sourceinspect.getsourcelines(obj)
+
+
+def _getsourcefile(obj):
+    try:
+        with _InspectContextManager():
+            ret = inspect.getsourcefile(obj)
```
ailzhang, line 134: can you add some comments when `ret` is None and why if `ret = inspect.getfile(obj)` throws it's safe to just pass?

python/taichi/lang/_wrap_inspect.py
```diff
@@ -0,0 +1,143 @@
...
+        if (filename in {"<timed exec>", "<magic-timeit>"}):
+            try:
+                ip = get_ipython()
+                if ip is not None:
+                    session_id = ip.history_manager.get_last_session_id()
+                    fd, filename = tempfile.mkstemp(prefix='_IPython_',
+                                                    suffix=f'_{session_id}.py')
+                    os.close(fd)
+
+                    lines = ip.history_manager._i00
```
ailzhang, line 86: need comment

python/taichi/lang/ast/checkers.py
```diff
@@ -1,7 +1,7 @@
 import ast
 
+from taichi.lang._wrap_inspect import _getsourcefile, _getsourcelines
```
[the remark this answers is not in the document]
neozhaoliang: Sure, fixed

python/taichi/lang/_wrap_inspect.py
neozhaoliang: Good catch... Updated to use `atexit.register` to remove the file.

python/taichi/lang/_wrap_inspect.py
neozhaoliang: Added

python/taichi/lang/_wrap_inspect.py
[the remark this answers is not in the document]
neozhaoliang: I'm afraid we cannot add a test here: the warning should only be raised in IDLE or Blender's interactive interpreter console, not the environment we can reproduce in a test?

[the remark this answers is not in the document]
ailzhang: agh right, I guess my point was more like we should add a global toggle(maybe env variable?)that when it's turned on, everything just falls back to source inspect. I think this is also what we agreed on last time right? (having a way for users to fallback instead of fallback for them automatically?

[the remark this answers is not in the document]
neozhaoliang: I see. Shall we make it an option in `ti.init()`?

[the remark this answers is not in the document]
ailzhang: I guess an env is much simpler?

python/taichi/lang/_wrap_inspect.py
```diff
@@ -0,0 +1,164 @@
...
+import inspect
+import os
+import tempfile
+
+import sourceinspect
+
+_builtin_getfile = inspect.getfile
+_builtin_findsource = inspect.findsource
+
+use_sourceinspect = int(os.getenv('USE_SOURCEINSPECT', 0)) == 1
```
ailzhang, line 11: IIUC now you can add a test in test_deprecations.py?

neozhaoliang: As the variable `use_sourceinspect` is determined at the time when taichi is imported, I think it's not that straight-forward to add such a test in `test_deprecations.py`?

ailzhang: you can set the env variable...

neozhaoliang: I don't understand why the test function doesn't raise the expected warning message here ...

neozhaoliang:
I have no idea why it doesn't raise the warnings, I used a test script and it works fine, but it always fails the pytest when running in the `test_deprecations.py`.

PR #6638, opened 2022-11-17
[ci] Revert "[Lang] [ci] Remove element shape and element dim in unit tests"

turbo0628:
Reverts taichi-dev/taichi#6620


ailzhang (approved):
Thanks!

PR #6641, opened 2022-11-17
[ci] Adaptively laying off test workers on CUDA OOM, 2nd try

feisuzhu:
Issue: #

### Brief Summary

Actual `pytest-xdist` fix: https://github.com/taichi-dev/pytest-xdist/commit/e3e1ccf5178ff53e5c9fa01adea18a2019cb764d

nothing is changed compared to last PR except the git commit hash in `requirements_test.txt`


ailzhang (approved):
Thanks!

PR #6643, opened 2022-11-17
[Lang] [ci] Clean element shape in tests

turbo0628:
Issue: #6572 

* Re-open #6620 which unexpectedly broke difftaichi and other user codes.


ailzhang (approved):
Thanks!

PR #6645, opened 2022-11-17
[bug] Add a mutex to protect counters_map in Statistics

### LINKED ISSUE #6644: Flaky test on M1: double free
**Describe the bug**
The following code fails on M1 from time to time.
```
import taichi as ti
ti.init(ti.cpu)

N = 1024 * 100

@ti.kernel
def reduce() -> ti.f32:
  res = 0.
  for _ in range(N):
    res += 0.2
  return res

res = reduce()

ground_truth = N * 0.2
print(res, ground_truth)
```
```
python(15557,0x170877000) malloc: Double free of object 0x1254c1210
python(15557,0x170877000) malloc: *** set a breakpoint in malloc_error_break to debug

```

lin-hitonami:
Issue: fixes #6644

### DIFF
--- a/taichi/util/statistics.cpp
+++ b/taichi/util/statistics.cpp
@@ -5,6 +5,7 @@ namespace taichi {
 Statistics stat;
 
 void Statistics::add(std::string key, Statistics::value_type value) {
+  std::lock_guard<std::mutex> _(counters_map_mutex);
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Thanks!

PR #6647, opened 2022-11-17
[build] Restore installed c_api headers in the python wheel

ailzhang:
Issue: #

### Brief Summary
The filter discarded our installed headers by accident, this PR restore them so that nightly users can get both lib/ and include there.


k-ye (approved):
LGTM!

PR #6648, opened 2022-11-17
[doc] Update prebuilt llvm15 links in dev install doc

ailzhang:
Issue: #6447 

### Brief Summary
Updating links in dev install doc. 

Nit: we still depend on clang-10/clang-11 to produce runtime.bc files and that part will be updated once those are cleaned up.


jim19930609 (approved):
LGTM!

PR #6659, opened 2022-11-18
[autodiff] Refactor independent block identification

erizmr:
Refactor independent block (IB) identification. 

Changes:
- Separate the identification of IB to `is_ib`and `is_smallest_ib`.
- The purpose is to find IBs as smallest as possible. The `MakeAdjoint` only  happens inside the IBs.
    1. No load/store to alloca outside the current block makes the block an IB.
    2. If it is an IB, either it is an inner most loop or there are global operations required gradient (such as atomics) makes it a smallest IB, i.e., should not dive deeper.

### DIFF
--- a/taichi/transforms/auto_diff.cpp
+++ b/taichi/transforms/auto_diff.cpp
@@ -80,11 +80,8 @@ class IndependentBlocksJudger : public BasicStmtVisitor {
     // - If the #1 is satisfied, either an inner most loop or a block without
     // global atomics / global load is an IB
     ib_meta_data.is_smallest_ib =
-        Judger.qualified_glb_operations_ || Judger.inner_most_loop_;
-
-    // Each smallest IB must be a IB
-    if (ib_meta_data.is_smallest_ib)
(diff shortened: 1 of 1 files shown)


taichi/transforms/auto_diff.cpp
```diff
@@ -45,7 +45,7 @@ class IndependentBlocksJudger : public BasicStmtVisitor {
     is_inside_loop_ = false;
   }
 
-  static bool run(IRNode *root) {
+  static void run(IRNode *root, bool &is_ib, bool &is_smallest_ib) {
```
ailzhang, line 48: aha is it possible to make a IBType enum to manager this? ;)

erizmr: I think there is an inclusion relation between `ib` and `smallest ib`, i.e., a `smallest ib` must be an `ib`. Will a IBType enum help with this?

ailzhang (approved):
LGTM！

PR #6662, opened 2022-11-18
[autodiff] Fix missing global load stmt in independent blocks

erizmr:
Issue: #6666 

Fix missing global load stmt in independent blocks.


ailzhang (approved):
Thanks!

PR #6665, opened 2022-11-18
[Lang] Deprecate element_dim and element_shape in ndarray annotations.

turbo0628:
Issue: #6572 

Changes:
* Add deprecation warning when `element_shape` and `element_dim` are specified.
* Internally remove `element_shape` and `element_dim` inside the type annotation. Added a helper function to create respective MatrixType in dtype for legacy code.
* Add deprecation test.
* Update docstring.


python/taichi/aot/utils.py
```diff
@@ -62,12 +64,14 @@ def produce_injected_args(kernel, symbolic_args=None):
...
                     f'{field_dim} from Arg {arg.name} doesn\'t match kernel\'s annotated field_dim={anno.field_dim}'
                 )
-
-            if anno.dtype is not None and not check_type_match(
-                    dtype, anno.dtype):
-                raise TaichiCompilationError(
-                    f' Arg {arg.name}\'s dtype {dtype.to_string()} doesn\'t match kernel\'s annotated dtype={anno.dtype.to_string()}'
-                )
+            anno_dtype = anno.dtype
+            if isinstance(anno.dtype, MatrixType):
```
ailzhang, line 68: nit: we need to be consistent of using `anno_dtype` or `anno.dtype`, it might cause confusions.

python/taichi/types/ndarray_type.py
```diff
@@ -9,78 +12,107 @@ def __init__(self, element_type, shape=None):
         self.layout = Layout.AOS
 
 
+# TODO(Haidong): This is a helper function that creates a MatrixType
+#                with respect to element_dim and element_shape.
+#                Remove this function when the two args are totally deprecated.
+def make_matrix_dtype_from_element_shape(element_dim, element_shape,
```
ailzhang, line 18:
```suggestion
def _make_matrix_dtype_from_element_shape(element_dim, element_shape,
```

python/taichi/types/ndarray_type.py
```diff
@@ -9,78 +12,107 @@ def __init__(self, element_type, shape=None):
...
+
+    # Scalars
+    if element_dim == 0 or (element_shape is not None
+                            and len(element_shape) == 0):
+        return primitive_dtype
+
+    # Cook element dim and shape into matrix type.
+    mat_dtype = None
+    if element_dim is not None:
+        if element_dim < 0 or element_dim > 2:
```
ailzhang, line 33: we may need to support dtype ndim > 2 in the future but a TODO comment is enough if we don't have a use case for now.

python/taichi/types/ndarray_type.py
```diff
@@ -9,78 +12,107 @@ def __init__(self, element_type, shape=None):
...
+            )
+        # Check dim consistency. The matrix dtype will be cooked later.
+        if element_shape is not None and len(element_shape) != element_dim:
+            raise ValueError(
+                f"Both element_shape and element_dim are specified, but shape doesn't match specified dim: {len(element_shape)}!={element_dim}"
+            )
+        mat_dtype = vector(None,
+                           primitive_dtype) if element_dim == 1 else matrix(
+                               None, None, primitive_dtype)
+    if element_shape is not None:
```
ailzhang, line 45: These `if`s are error prone in the future that it may slip away and return `mat_dtype` as None. 
Can we refactor this function to: 
```
if element_dim is not None:
  ...
  return ...
elif element_shape is not None:
  ...
  return ...
else:
  throw!
```

python/taichi/types/ndarray_type.py
```diff
@@ -9,78 +12,107 @@ def __init__(self, element_type, shape=None):
...
-                )
-
-            if self.element_shape != ndarray_type.element_type.shape():
-                raise ValueError(
-                    f"Invalid argument into ti.types.ndarray() - required element_shape={self.element_shape}, but {ndarray_type.element_type.shape()} is provided"
-                )
+        # FIXME(Haidong) Cannot use Vector/MatrixType due to circular import
+        # Use the CompuoundType instead to determine the specific typs.
+        # TODO Replace CompoundType with MatrixType and VectorType
+        if isinstance(self.dtype, CompoundType):
```
ailzhang, line 92: nit: can we just check if `self.dtype` and `ndarray_type.dtype` match instead of translating them back to element_dim and element_shape? 
This may look like:
```
if isinstance(self.dtype, CompoundType):
   assert isinstance(ndarray_type.dtype, CompoundType)
   # check self.dtype.element_shape, shape and layout matches
else:
  assert isinstance(ndarray_type.dtype, primitive_dtypes) # pseudo code
  # check self.dtype matches ndarray_type.dtype
```

ailzhang (approved):
Awesome thanks! 
Had a few suggestions inline to make the code more readable but otherwise LGTM!

python/taichi/aot/utils.py
turbo0628: Thanks!

python/taichi/types/ndarray_type.py
turbo0628: Exact match of `self.dtype` and `ndarray_type` is too strict for our use cases at the moment. There are `dtype` of MatrixType with `None` shape and `None` dim and `None` type. So the refactored match code is no cleaner than this bunch of if-else. I did the translation just to keep it backward compatible, and yes, we should clean up the logics for type matching for ndarrays. I've get rid of the `element_shape` and `element_dim` vars, but I think it's better to do a full sweep in a follow-up PR, WDYT?

python/taichi/types/ndarray_type.py
turbo0628: I used to place some return statement at each branch, but forgot to change that if back to elif when I unified the returns. Thanks for catching this!

PR #6671, opened 2022-11-18
[misc] Remove class Statistics

### LINKED ISSUE #6646: Unused feature: Statistics
**Describe the bug**
The class `Statistics` counts the number of IRs and offloads of different kinds, but it seems not in use now, and it has no interface exposed in Python. 
https://github.com/taichi-dev/taichi/blob/1fba52f2fcba069bda13c3e1d78f7ca1ab2357f6/taichi/util/statistics.h#L7-L28
Is the counting still useful now or will be useful in the future? If not, we should remove the related code out of our codebase.

lin-hitonami:
Issue: fixes #6646


ailzhang (approved):
Thanks!!

PR #6675, opened 2022-11-20
[misc] Add VS Chromium Code Search config

bobcao3:
This enables the VSChromium plugin (http://chromium.github.io/vs-chromium/#feature-summary) & the powerful Chromium Code Search tool to work:

image
image

Where you can search for code instantly with sane search queries (not regex).
By instantly I really mean instantly, way way way way faster than Ctrl-Shift-F


ailzhang (approved):
LGTM!

PR #6676, opened 2022-11-20
[misc] Fix warning with TI_DLL_EXPORT

bobcao3:
`enum class` only exists in headers and do not produce a symbol. `TI_DLL_EXPORT` has no meaning in this case...


PENGUINLIONG:
lgtm; cc @ailzhang

PENGUINLIONG (approved):

ailzhang (approved):
Aha nice catch, thanks!

PR #6677, opened 2022-11-20
[aot] Load AOT module from Zip archive

PENGUINLIONG:
Good news: It doesn't break any existing loading code. The only difference is that now `ti_load_aot_module` accepts a path to a `.tcm` file.

The next step is to load AOT module from memory with `ti_create_aot_module`.


c_api/tests/c_api_interface_test.cpp
```diff
@@ -138,3 +138,23 @@ TEST_F(CapiTest, DryRunOpenglAotModule) {
...
+
+    {
+      // Vulkan Runtime
+      TiArch arch = TiArch::TI_ARCH_VULKAN;
+      ti::Runtime runtime(arch);
+      ti::AotModule aot_mod = runtime.load_aot_module(aot_mod_ss.str());
+      ti::Kernel run = aot_mod.get_kernel("run");
+      ti::NdArray<int32_t> arr = runtime.allocate_ndarray<int32_t>({16});
+      run[0] = arr;
+      run.launch();
```
ailzhang, line 157: nit: how about also waiting and checking on values?

taichi/common/virtual_dir.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+  return false;
+}
+
+std::unique_ptr<VirtualDir> VirtualDir::open(const std::string &path) {
+  if (is_zip_file(path)) {
+    return ZipArchiveVirtualDir::create(path);
+  } else {
+    // (penguinliong) I wanted to use `std::filesyste::is_directory`. But it
+    // seems `<filesystem>` is only supported in MSVC.
```
ailzhang, line 119: Hmmmm <filesystem> is a standard C++17 header so it should be available on all platforms. 
background: macos10.14 didn't have it so we didn't use it before, but now that it's removed from our supported os we should be all good to use <filesystem>

ailzhang (approved):
LGTM!

taichi/common/virtual_dir.cpp
PENGUINLIONG: actually it failed linux ci 🥹

c_api/tests/c_api_interface_test.cpp
PENGUINLIONG: oof forgot about it

jim19930609 (commented):
Any particular reason why we want to copy MiniZ codes into Taichi, instead of leaving it as an external library?

( Btw, is it legal to copy codes from other open source? )

jim19930609:
Okay it seems legal to steal the codes, but I'm still wondering if it's better to integrate MiniZ as an external library so we dont have to maintain the source code.

image

PENGUINLIONG:
> Any particular reason why we want to copy MiniZ codes into Taichi, instead of leaving it as an external library?
> 
> 
> 
> ( Btw, is it legal to copy codes from other open source? )

@jim19930609 The previous version has glued .h and .c together so I couldn't include it from multiple source so the one in common is a recent release. The license info is attached in the front of miniz.h and it actually claims public domain code so I think that's fine.

PENGUINLIONG:
> Okay it seems legal to steal the code, but I'm still wondering if it's better to integrate MiniZ as an external library so we dont have to maintain the source code.
> 
> image

i see. but i would make it in another PR tho. i dont wanna wait another year for the ci..

jim19930609:
> > Any particular reason why we want to copy MiniZ codes into Taichi, instead of leaving it as an external library?
> > ( Btw, is it legal to copy codes from other open source? )
> 
> @jim19930609 The previous version has glued .h and .c together so I couldn't include it from multiple source so the one in common is a recent release. The license info is attached in the front of miniz.h and it actually claims public domain code so I think that's fine.

Sure, it's definitely welcome to upgrade the Miniz version. My question mainly focuses on how we integrate MiniZ into Taichi. Instead of copying its source code, it might be more maintainable if we make MiniZ a submodule of Taichi so it's easier to follow up MiniZ's future upgrades/bug-fixes.

However, the change in integration may not happen in this specific PR - we can file a follow-up PR to make that happen.

PR #6678, opened 2022-11-21
[build] Support any clang<=15 as CLANG_EXECUTABLE

ailzhang:
Issue: #6447

### Brief Summary
We can now support any clang version <=15 to compile runtime.cpp into `.bc` files. 
A few notes:
- clang13 and clang15 introduced new diagnostic flags that we have to temporarily disable in Werror until they're fixed. 
- The commit update introduced in SPIRV-Tools was also introduced to mitigate similar issues as shown above. 
- I have tested locally on my linux with clang-12, clang-14, clang-15 and MacBook with clang-10, clang-13.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,42 @@ if (WIN32)
   message("CMAKE_MSVC_RUNTIME_LIBRARY: ${CMAKE_MSVC_RUNTIME_LIBRARY}")
 endif()
 
+# Setup CLANG_EXECUTABLE
+if (CLANG_EXECUTABLE)
+  message("CLANG_EXECUTABLE defined: ${CLANG_EXECUTABLE}")
+elseif ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+  set (CLANG_EXECUTABLE ${CMAKE_CXX_COMPILER})
+  message("Clang executable using host compiler ${CLANG_EXECUTABLE}")
+else()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -86,6 +86,42 @@ if (WIN32)
...
+  execute_process(COMMAND ${CLANG_EXECUTABLE} --version OUTPUT_VARIABLE CLANG_VERSION_OUTPUT)
+  string(REGEX MATCH "([0-9]+)\\.[0-9]+(\\.[0-9]+)?" CLANG_VERSION "${CLANG_VERSION_OUTPUT}")
+  message("${CLANG_EXECUTABLE} --version: ${CLANG_VERSION}")
+
+  set(CLANG_VERSION_MAJOR "${CMAKE_MATCH_1}")
+endmacro()
+
+if (APPLE)
+  set(CLANG_OSX_FLAGS "-isysroot${CMAKE_OSX_SYSROOT}")
+  set(CLANG_HIGHEST_VERSION "13")
```
jim19930609, line 114: Can we document the specific issue with clang > 13 on MacOS?

jim19930609 (approved):
Super!

CMakeLists.txt
ailzhang: sure! I just haven't tried too many clang versions on macOS yet so set this tends to be more conservative. Will send a followup PR to relax this if possible or add a comment!

PR #6681, opened 2022-11-21
[autodiff] Enforce an empty block a smallest independent block

erizmr:
According to the definition of an independent block (IB) in autodiff, i.e., no local load/store to allocas outside the block itself.
An empty block should be a smallest IB.


ailzhang (approved):
nit: do we have a test that can cover this? ;)

erizmr:
> 
We have a nested empty for loop test to ensure there should at least one IB, i.e., an empty block should be recorded as an IB. 
https://github.com/taichi-dev/taichi/blob/master/tests/python/test_ad_for.py#L6

PR #6687, opened 2022-11-21
[Lang] Deprecate field_dim in ndarray annotation

turbo0628:
Issue: #6572 

* Deprecate `field_dim` in ndarray type annotation, replace with `ndim`.
* Change tests accordingly. Examples are left unchanged.

### DIFF
--- a/tests/cpp/aot/python_scripts/graph_aot_test.py
+++ b/tests/cpp/aot/python_scripts/graph_aot_test.py
@@ -11,17 +11,17 @@ def compile_graph_aot(arch):
         return
 
     @ti.kernel
-    def run0(base: int, arr: ti.types.ndarray(field_dim=1, dtype=ti.i32)):
+    def run0(base: int, arr: ti.types.ndarray(ndim=1, dtype=ti.i32)):
         for i in arr:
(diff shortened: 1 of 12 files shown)


tests/cpp/aot/python_scripts/graph_aot_test.py
```diff
@@ -11,17 +11,17 @@ def compile_graph_aot(arch):
         return
 
     @ti.kernel
-    def run0(base: int, arr: ti.types.ndarray(field_dim=1, dtype=ti.i32)):
+    def run0(base: int, arr: ti.types.ndarray(ndim=1, dtype=ti.i32)):
```
ailzhang, line 14: nit: shall we reorder this to dtype first given they're both keywords?

ailzhang (approved):
Awesome thanks!

tests/cpp/aot/python_scripts/graph_aot_test.py
turbo0628: There are lots of such cases in examples and tests. Even `any_arr` is still present in our codebase. Let's do a full sweep in the next PR.

PR #6689, opened 2022-11-21
[bug] MatrixType bug fix: Fix error with nested StructType and MatrixType

jim19930609:

### Brief Summary
1. Modified `Matrix::fill()` to broadcast `val` into VectorType if `ndim == 1`, and to MatrixType if `ndim == 2`
2. Modified `Struct::fill()` to apply `matrix_op.fill()` in case of `Expr` with TensorType

### DIFF
--- a/python/taichi/lang/struct.py
+++ b/python/taichi/lang/struct.py
@@ -222,16 +222,14 @@ class Struct(TaichiOperations):
         Args:
             val (Union[int, float]): Value to fill.
         """
-        entries = {}
         for k, v in self.items:
             if isinstance(v, impl.Expr) and v.ptr.is_tensor():
                 from taichi.lang import matrix_ops  # pylint: disable=C0415
                 matrix_ops.fill(v, val)
-                entries[k] = v
             elif isinstance(v, (Struct, Matrix)):
-                entries[k] = v._element_wise_binary(ops.assign, val)
+                v._element_wise_binary(ops.assign, val)
             else:
-                entries[k] = ops.assign(v, val)
+                ops.assign(v, val)
 
     def __len__(self):
         """Get the number of entries in a custom struct"""
--- a/python/taichi/lang/struct.py
+++ b/python/taichi/lang/struct.py
@@ -222,8 +222,6 @@ class Struct(TaichiOperations):
         Args:
             val (Union[int, float]): Value to fill.
         """
-
-        bcast_val = self._broadcast_copy(val)
         entries = {}
         for k, v in self.items:
             if isinstance(v, impl.Expr) and v.ptr.is_tensor():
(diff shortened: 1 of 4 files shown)


python/taichi/lang/struct.py
```diff
@@ -222,10 +222,20 @@ def fill(self, val):
         Args:
             val (Union[int, float]): Value to fill.
         """
-        def assign_renamed(x, y):
-            return ops.assign(x, y)
 
-        return self._element_wise_writeback_binary(assign_renamed, val)
+        bcast_val = self._broadcast_copy(val)
```
strongoier, line 226: Is it possible to simply use `val`? It seems that `_element_wise_binary` will also broadcast.

python/taichi/lang/struct.py
```diff
@@ -222,10 +222,20 @@ def fill(self, val):
...
+            if isinstance(v, impl.Expr) and v.ptr.is_tensor():
+                from taichi.lang import matrix_ops  # pylint: disable=C0415
+                matrix_ops.fill(v, val)
+                entries[k] = v
+            elif isinstance(v, (Struct, Matrix)):
+                entries[k] = v._element_wise_binary(ops.assign,
+                                                    bcast_val.entries[k])
+            else:
+                entries[k] = ops.assign(v, bcast_val.entries[k])
+        return Struct(entries)
```
strongoier, line 238: The original `fill()` seems a bit problematic. `fill()` should happen in-place and no return value is needed.

taichi/transforms/lower_ast.cpp
```diff
@@ -397,6 +397,7 @@ class LowerAST : public IRVisitor {
     auto fctx = make_flatten_ctx();
     flatten_rvalue(expr, &fctx);
     if (dest.is<IdExpression>()) {
+      flatten_lvalue(dest, &fctx);
```
strongoier, line 400: nit: move this outside as all branches do the same thing

taichi/transforms/lower_ast.cpp
```diff
@@ -397,6 +397,7 @@ class LowerAST : public IRVisitor {
...
       fctx.push_back<LocalStoreStmt>(
           assign->parent->lookup_var(assign->lhs.cast<IdExpression>()->id),
           expr->stmt);
```
strongoier, line 403:
```suggestion
      fctx.push_back<LocalStoreStmt>(dest->stmt, expr->stmt);
```
After `flatten_lvalue` we can simply do this.

python/taichi/lang/struct.py
```diff
@@ -222,10 +222,16 @@ def fill(self, val):
         Args:
             val (Union[int, float]): Value to fill.
         """
-        def assign_renamed(x, y):
-            return ops.assign(x, y)
-
-        return self._element_wise_writeback_binary(assign_renamed, val)
+        entries = {}
```
strongoier, line 225: Is `entries` also unnecessary?

jim19930609: good point, thanks!

Issue #6660, opened 2022-11-18
[RFC] Unify packed mode and non-packed mode


strongoier:
## Motivation

Packed mode was introduced to allow users to trade running time for memory usage:
https://github.com/taichi-dev/taichi/blob/db6ab0d30280f665da26a7c73e144ac7be2993ba/docs/lang/articles/basic/layout.md?plain=1#L362-L380

A more detailed introduction can be found at https://github.com/taichi-dev/taichicon/releases/download/taichicon02/Packed.Mode-Effectively.Reducing.Memory.Costs.of.Non-power-of-two.Fields-Yi.Xu.pdf.

However, simply providing users a switch is far from perfect. The default automatic padding behavior is non-intuitive to newcomers, so it is hard for them to diagnose related problems (e.g. 2-10x larger memory usage than expected, or sparsity manipulation in a wrong way) and then find out this feature.

To improve user experience, I propose to unify packed mode and the default non-packed mode - making the memory allocation and coordinate mapping behavior of Taichi SNodes as straightforward to understand as under packed mode but the runtime performance in most cases as optimized as under non-packed mode.

## Proposal

The main idea is to try our best to optimize the runtime overhead of packed mode and then make it the default mode.

#6444 has taken the first step. It has made overhead of SNode access under normal cases (no second-time division on one axis) the same for both packed mode and non-packed mode.

After a discussion with @yuanming-hu, we agree that when users make a second-time division on some axis, it is very likely that they are prioritizing runtime performance and in the process of fine-tuning memory access patterns. In this case, it makes no sense to introduce slow operations like mod or div in each access. Therefore, we can add a limitation on the definition of SNodes - the non-first-time division on one axis must be a power of two. With this limitation, we can make the runtime overhead of SNode access always the same for both modes.

The only remaining parts involving coordinate calculation are struct fors. As long as the struct for is operating on a SNode whose path to root contains no non-power-of-two division, we can make the runtime overhead of the struct for the same for both modes. Unfortunately, for the remaining case, where the struct for is operating on a SNode whose path to root does contain some non-power-of-two division, the runtime overhead of the struct for for packed mode will be higher. Luckily, the runtime overhead of a struct for only appears once in each thread and is unlikely to become the performance bottleneck. In the extreme case where it does become the bottleneck, the user can change that non-power-of-two division to power-of-two as an optimization strategy.

## Plan

- [x] Add the limitation above and fix tests violating the limitation.
[...]

PR #6690, opened 2022-11-21
[Lang] Limit non-first division of an axis on a SNodeTree path to a power of two

strongoier:
Issue: #6660

### DIFF
--- a/taichi/ir/snode.cpp
+++ b/taichi/ir/snode.cpp
@@ -37,7 +37,8 @@ SNode &SNode::insert_children(SNodeType t) {
 SNode &SNode::create_node(std::vector<Axis> axes,
                           std::vector<int> sizes,
                           SNodeType type,
-                          bool packed) {
+                          bool packed,
+                          const std::string &tb) {
   TI_ASSERT(axes.size() == sizes.size() || sizes.size() == 1);
   if (sizes.size() == 1) {
     sizes = std::vector<int>(axes.size(), sizes[0]);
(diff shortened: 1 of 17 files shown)


taichi/ir/snode.cpp
```diff
@@ -54,30 +54,26 @@ SNode &SNode::create_node(std::vector<Axis> axes,
...
-    auto &ind = axes[i];
-    new_node.extractors[ind.value].activate(
+    int ind = axes[i].value;
+    auto end = new_node.physical_index_position + new_node.num_active_indices;
+    bool is_first_division =
+        std::find(new_node.physical_index_position, end, ind) == end;
+    if (is_first_division) {
+      new_node.physical_index_position[new_node.num_active_indices++] = ind;
+    } else {
+      if (!bit::is_power_of_two(sizes[i])) {
```
jim19930609, line 64: Shall we limit this check to packed mode only? In non-packed mode, we always pad `sizes` to the power of 2, so this limitation should've never applied?

strongoier:
According to the discussions in #6660, I've turned the error into a warning.

> Interesting... If I understand it correctly, after this PR, the packed mode and non-packed mode will only influence the first axis of a snode tree?

More accurately, they will only influence the first division of an axis on a SNodeTree path.

taichi/ir/snode.cpp
strongoier: You're right. Although non-packed mode will be removed, this PR should make a proper intermediate state.

jim19930609 (approved):
LGTM!

PR #6697, opened 2022-11-22
[build] Integrate backward_cpp to test targets for enabling C++ stack trace

jim19930609:
Issue: #

### Brief Summary
When building with `-DTI_BUILD_TESTS=ON`, `taichi_cpp_tests`,`taichi_c_api_tests`, `taichi_python.so` will link to an external library: https://github.com/bombela/backward-cpp, which enables C++ stack trace upon test failures.

## Segfault with python tests 
`python3 test.py`

image

## Segfault with cpp tests
`./build/taichi_cpp_tests`

image

## Segfault with c_api tests
`./build/taichi_c_api_tests`

image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -96,7 +96,6 @@ jobs:
             -DTI_WITH_CC:BOOL=OFF
             -DTI_WITH_VULKAN:BOOL=ON
             -DTI_BUILD_TESTS:BOOL=ON
-            -DTI_WITH_BACKTRACE:BOOL=ON
 
       - name: Archive Wheel Artifacts
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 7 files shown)


ailzhang (commented):
@jim19930609 instead of coupling with TI_BUILD_TESTS, shall we add a new toggle for backward-cpp itself to make it cleaner? 
For example:
- if in CI we just set TI_WITH_BT on by default, and whatever targets (whether they're tests or not) can be built with backward-cpp. 

The main reason I'm asking for it is that we may want to build a taichi in debug mode (with backtrace) and it's a bit weird to say we need to turn on TI_BUILD_TESTS which is supposed to only toggle on/off of build test targets?

.github/workflows/release.yml
```diff
@@ -149,6 +150,7 @@ jobs:
             -DTI_WITH_CC:BOOL=OFF
             -DTI_WITH_VULKAN:BOOL=OFF
             -DTI_BUILD_TESTS:BOOL=ON
+            -DTI_WITH_BACKTRACE:BOOL=ON
```
ailzhang, line 153: hmmm I'd assume we only want to turn this on for nightly releases but not for official releases?

jim19930609: Ah my bad, thanks!

ailzhang (approved):
Thanks！

PR #6699, opened 2022-11-22
[misc] Headers dependency cleanup from RHI

bobcao3:
In this MR, LLVM and Eigen header dependencies have been removed from the RHI folder. In addition, `device.h` now uses cleaner and leaner includes so that we don't include `ir.h` and `program.h` unnecessarily.


k-ye (approved):
LGTM!

k-ye:
PR nits: could you associate PR with issues, thanks!

ailzhang (approved):
Thanks!

PR #6700, opened 2022-11-22
[autodiff] Separate non-linear operators to an individual class

erizmr:
Separate non-linear operators to an individual class so that can be accessed by other visitors, e.g., the visitors in IB identification will use it in further development.


ailzhang (approved):
LGTM!

PR #6718, opened 2022-11-23
[opt] Turn mod/div into bit_and/bit_shr if possible in packed mode

strongoier:
Issue: #6660

### Brief Summary

After this PR, the two main goals in #6660,

> - Make the runtime overhead of SNode access always the same for both modes if the limitation is obeyed.
> - Make the runtime overhead of struct for on a SNode whose path to root contains no non-power-of-two division always the same for both modes.

are both achieved.

See comments in the code for implementation details.


ailzhang (approved):
Thanks!

jim19930609 (approved):

PR #6739, opened 2022-11-25
[Metal] [error] Raise deprecate warning and error when using sparse snodes on metal

lin-hitonami:
Issue: https://github.com/taichi-dev/taichi/discussions/6358


ailzhang (approved):
Thanks!
(probably need a followup PR to update docs as well ;)

strongoier (approved):

Issue #6036, opened 2022-09-11
Do not the Kernel arguments support the class 'bool'?


17150934:
When I set the type-hinted of my Kernel arguments as 'bool', there existed an error:
`TaichiSyntaxError: Invalid type annotation (argument 0) of Taichi kernel: <class 'bool'>`
So, do not the Kernel arguments support the class 'bool'?

AD1024:
Thanks for the issue.
Currently `bool` type is not (explicitly) one of the primitive types in Taichi. There is a `u1` type, which is used as `bool`. A workaround is to use `int` (or other integer types) to mimic its behavior.

@strongoier I noticed that the `u1` type is defined in C++ but not exposed to the Python frontend (used in `export_lang.cpp` but not referenced in `primitive_types.py`). Should we use it when encountering the `bool` type during compilation?

strongoier:
`u1` type has not been fully supported yet (https://github.com/taichi-dev/taichi/issues/577), and we plan to support it in the future. For now we need to use `int` as a workaround.

re-xyr:
As a first step towards adding the proper bool type, we're going to add `bool` as an alias of `i32`.

PR #6742, opened 2022-11-26
[Lang] [type] Add bool type in python as an alias to i32

re-xyr:
Issue: #577 #6036

### Brief Summary

This PR adds `bool` as an alias to `ti.i32`. Specifically,

- `x: bool` is equivalent to `x: i32`
- `-> bool` is equivalent to `-> i32`
- `bool(x)` is equivalent to `ti.cast(x, i32)`.

This is a temporary solution while we work towards a standalone bool type.


strongoier (approved):
LGTM!

PR #6743, opened 2022-11-26
[ci] Add AMDGPU relected ci

GaleSeLee:

### Brief Summary
Add the logic of docker for AMDGPU ci
1. The llvm used which only contains AMDGPU and X86 targets is from docker image
2. Using AMDGPU in docker requires that `/dev/kfd` and the directory `/dev/dri` be mounted on. For `dev` user, there is no permission to access these character devices by default.
3. `TI_WITH_CUDA=OFF`
4. `TI_RUN_RELEASE_TESTS=OFF`
5. Currently only run cpu-relected test

### DIFF
--- a/.github/workflows/scripts/unix-build.sh
+++ b/.github/workflows/scripts/unix-build.sh
@@ -68,6 +68,10 @@ build_taichi_wheel() {
             EXTRA_ARGS="-p manylinux_2_27_x86_64"
         fi
     fi
+
+    if ["$AMDGPU_TEST"]; then
+      TAICHI_CMAKE_ARGS="$TAICHI_CMAKE_ARGS -DTI_WITH_CUDA:BOOL=OFF"
+    fi
     python3 misc/make_changelog.py --ver origin/master --repo_dir ./ --save
 
     python3 setup.py $PROJECT_TAGS bdist_wheel $EXTRA_ARGS
--- a/.github/workflows/scripts/unix-build.sh
+++ b/.github/workflows/scripts/unix-build.sh
@@ -7,7 +7,12 @@ set -ex
 [[ "$IN_DOCKER" == "true" ]] && cd taichi
 
 if [[ $OSTYPE == "linux-"* ]]; then
-  if [ -z "$AMDGPU_TEST"]; then
+  if ["$AMDGPU_TEST"]; then
+    sudo ln -s /usr/bin/clang++-10 /usr/bin/clang++
+    sudo ln -s /usr/bin/clang-10 /usr/bin/clang
+    sudo ln -s /usr/bin/ld.lld-10 /usr/bin/ld.lld
+    export LLVM_DIR="/taichi-llvm-15.0.0-linux"
+  else
     if [ ! -d ~/taichi-llvm-15 ]; then
       pushd ~
       if [ -f /etc/centos-release ] ; then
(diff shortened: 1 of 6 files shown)


.github/workflows/scripts/unix-build.sh
```diff
@@ -7,19 +7,26 @@ set -ex
...
 
 if [[ $OSTYPE == "linux-"* ]]; then
-  if [ ! -d ~/taichi-llvm-15 ]; then
-    pushd ~
-    if [ -f /etc/centos-release ] ; then
-      # FIXIME: prebuilt llvm15 on ubuntu didn't work on manylinux image of centos. Once that's fixed, remove this hack.
-      wget https://github.com/ailzhang/torchhub_example/releases/download/0.3/taichi-llvm-15-linux.zip
-    else
-      wget https://github.com/taichi-dev/taichi_assets/releases/download/llvm15/taichi-llvm-15-linux.zip
+  if ["$AMDGPU_TEST"]; then
```
feisuzhu, line 10: This is not right :(

I guess you wanted to test if `$AMDGPU_TEST` is set, please use

```bash
if [ ! -z $AMDGPU_TEST ]; then
    # do sth when set
else
    # do sth when not
fi
```

Note the spaces around the brackets, they are mandatory.

The condition expression means `if not (is-string-empty ($AMDGPU_TEST))`

.github/workflows/scripts/unix_test.sh
```diff
@@ -14,6 +14,11 @@ setup_python
 
 [[ "$IN_DOCKER" == "true" ]] && cd taichi
 
+if ["$AMDGPU_TEST"]; then
```
feisuzhu, line 17: ditto

.github/workflows/scripts/unix_test.sh
```diff
@@ -103,6 +108,9 @@ if [ -z "$GPU_TEST" ]; then
         fi
         python3 tests/run_tests.py -vr2 -t4 -k "not paddle" -a "$TI_WANTED_ARCHS"
     fi
+elif ["$AMDGPU_TEST"]; then
```
feisuzhu, line 111: ditto

.github/workflows/testing.yml
```diff
@@ -268,6 +268,76 @@ jobs:
...
+  build_and_test_amdgpu_linux:
+    name: Build and Test (AMDGPU)
+    needs: check_files
+    timeout-minutes: ${{ github.event.schedule != '0 18 * * *' && 90 || 120 }}
+    strategy:
+      matrix:
+        tags:
+          - [self-hosted, amdgpu]
+
+    runs-on: ${{ matrix.tags }}
```
feisuzhu, line 280: I didn't get why you are using matrix here, assuming you are just copying & pasting for convenience.
Then, please remove matrix here
```suggestion
    runs-on: [self-hosted, amdgpu]
```

.github/workflows/testing.yml
```diff
@@ -268,6 +268,76 @@ jobs:
...
+        run: |
+          [[ ${{needs.check_files.outputs.run_job}} == false ]] && exit 0
+          . .github/workflows/scripts/common-utils.sh
+
+          ci-docker-run-amdgpu --name taichi-test \
+             registry.taichigraphics.com/taichidev-ubuntu18.04.amdgpu:v0.0.3 \
+             /home/dev/taichi/.github/workflows/scripts/unix_test.sh
+        env:
+          PY: py38
+          TI_WANTED_ARCHS: 'cc,cpu,amdgpu'
```
feisuzhu, line 321: Is `cc` backend relevant here? Remove if not.

feisuzhu (changes requested):

.github/workflows/testing.yml
```diff
@@ -268,6 +268,75 @@ jobs:
           path: taichi-release-tests/bad-compare/*
           retention-days: 7
 
+  build_and_test_amdgpu_linux:
+    name: Build and Test (AMDGPU)
+    needs: check_files
+    timeout-minutes: ${{ github.event.schedule != '0 18 * * *' && 90 || 120 }}
+    strategy:
```
feisuzhu, line 275: Hmm, I mean, the `matrix` thing is redundant but you have `runs-on` removed :(

feisuzhu (approved):
The rest looks right, approving

PR #6749, opened 2022-11-28
[Misc] Refactored flattend_values() to avoid potential conflicts in flattened statements

jim19930609:

Overriding the flattened statement `stmt` of an `Expression` can cause conflicts, for example:
```
@ti.kernel
def test():
    x = ti.Vector([1, 2, 3, 4])
    tmp = x + x[0] # implicit broadcast
```

In `x + x[0]`, the `x` on the lhs serves as rvalue whereas the `x` in the `x[0]` serves as a lvalue, so the result of `flatten_rvalue()` and `flatten_lvalue()` will override each other.

To avoid such conflicts, this PR refactored the `flatten_values()` functions:
1. Flattened statement `stmt` of an `Expression` will only get modified by `Expression::flatten()`, any other overriding will be forbidden.
2. `flatten_rvalue()` and `flatten_lvalue()` now returns the flattened statement as the result. External users such as `irpass::lower_ast()` will turn to use the returned statement.

### DIFF
--- a/taichi/ir/frontend_ir.cpp
+++ b/taichi/ir/frontend_ir.cpp
@@ -1135,8 +1135,7 @@ void GetElementExpression::type_check(CompileConfig *config) {
 }
 
 void GetElementExpression::flatten(FlattenContext *ctx) {
-  auto src_stmt = flatten_lvalue(src, ctx);
-  ctx->push_back<GetElementStmt>(src_stmt, index);
+  ctx->push_back<GetElementStmt>(src->get_flattened_stmt(), index);
   stmt = ctx->back_stmt();
 }
 // Mesh related.
--- a/taichi/ir/frontend_ir.cpp
+++ b/taichi/ir/frontend_ir.cpp
@@ -332,7 +332,6 @@ void BinaryOpExpression::flatten(FlattenContext *ctx) {
   // if (stmt)
   //  return;
   auto lhs_stmt = flatten_rvalue(lhs, ctx);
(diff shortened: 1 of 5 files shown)


strongoier (approved):

taichi/ir/frontend_ir.cpp
```diff
@@ -919,10 +919,10 @@ void AtomicOpExpression::flatten(FlattenContext *ctx) {
...
   }
   // expand rhs
-  flatten_rvalue(val, ctx);
-  auto src_val = val->stmt;
-  flatten_lvalue(dest, ctx);
-  stmt = ctx->push_back<AtomicOpStmt>(op_type, dest->stmt, src_val);
+  auto val_stmt = flatten_rvalue(val, ctx);
+  auto src_val = val_stmt;
+  auto dest_stmt = flatten_lvalue(dest, ctx);
+  stmt = ctx->push_back<AtomicOpStmt>(op_type, dest_stmt, src_val);
```
strongoier, line 925:
```suggestion
  auto val_stmt = flatten_rvalue(val, ctx);
  auto dest_stmt = flatten_lvalue(dest, ctx);
  stmt = ctx->push_back<AtomicOpStmt>(op_type, dest_stmt, val_stmt);
```
After the refactoring I think we can eliminate
```cpp
auto b = a;
c = b;
```
to make the code cleaner. It seems that there are a few places with this situation.

jim19930609 (commented):
Code adjustment

PR #6753, opened 2022-11-28
[Lang] Add deprecation warning for the removal of the packed switch

strongoier:
Issue: #6660

### Brief Summary

Apart from the deprecated warning, the previously added limitation that "non-first division on an axis must be a power of two" is now made purely a warning because it can already let users learn the best practice.


ailzhang (approved):
Thanks!

PR #6755, opened 2022-11-28
[Doc] Stop mentioning packed mode

strongoier:
Issue: #6660

### Brief Summary

We no longer need the concept of packed mode. Now we only have to suggest users using power-of-two block size in hierarchical fields.


neozhaoliang (approved):
LGTM!

PR #6762, opened 2022-11-29
[build] Initial distributed compiling support

feisuzhu:
Issue: #6445

### Brief Summary

This PR addresses quirks in `sccache` distributed compiling:

1.  `sccache` would run preprocessing locally (`clang -E`) first and then send the expanded code to remote build server. This can trigger false warnings not present when compiling locally.
2.  Compile host and target can be different, `--target=` must be send to remote build server.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,4 +213,3 @@ if (TI_WITH_GRAPHVIZ)
       WORKING_DIRECTORY "${CMAKE_BINARY_DIR}"
   )
 endif()
-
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -133,6 +133,7 @@ if (TI_WITH_BACKTRACE)
   add_subdirectory(external/backward_cpp)
 endif()
 
+include(cmake/Distributed.cmake)
```
jim19930609, line 136: In the complete state, will this `Distributed Compilation` only apply to `Taichi CI System` and individual contributors won't be able to use it? 

If it's not designed for individual contributors, then we probably want to add a check and give warnings if it's not running on CI (just in case people mistakenly set DISTRIBUTED_COMPILE=ON)

jim19930609 (approved):

PR #6764, opened 2022-11-29
[Doc] [lang] Cherrypick commits related to dynamic SNode

lin-hitonami:
Issue: #


PENGUINLIONG (approved):
lgtm. tho, is it possible to remove sparse from metal's extension list instead of this?

```python
@test_utils.test(require=ti.extension.sparse, exclude=ti.metal)
```

PENGUINLIONG:
maybe in another pr. no rush

lin-hitonami:
maybe we can do it in 1.4.0 when we completely delete sparse snodes in metal backend.

PR #6766, opened 2022-11-29
[aot] C-API to get available archs

PENGUINLIONG:
This PR provides a new API to query all available APIs on the current platform. It is basically a public version of the test utils APIs (`is_vulkan_available`, etc.) @jim19930609 introduced before.

### DIFF
--- a/c_api/include/taichi/taichi_cuda.h
+++ b/c_api/include/taichi/taichi_cuda.h
@@ -1,5 +1,9 @@
 #pragma once
 
+#ifndef TI_WITH_CUDA
+#define TI_WITH_CUDA 1
+#endif  // TI_WITH_CUDA
+
 #include <taichi/taichi.h>
 
 #ifdef __cplusplus
--- a/c_api/include/taichi/taichi_cuda.h
+++ b/c_api/include/taichi/taichi_cuda.h
@@ -1,9 +1,5 @@
 #pragma once
 
-#ifndef TI_WITH_CUDA
-#define TI_WITH_CUDA 1
-#endif  // TI_WITH_CUDA
-
 #include <taichi/taichi.h>
 
 #ifdef __cplusplus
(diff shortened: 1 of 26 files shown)


c_api/include/taichi/taichi_cuda.h
```diff
@@ -1,5 +1,9 @@
 #pragma once
 
+#ifndef TI_WITH_CUDA
```
jim19930609, line 3: In general it's not suggested to change the macro definition in a normal header file, because .cpp files who included  `taichi_cuda.h` will have different `TI_WITH_CUDA` values compared to those who did not. This is likely gonna cause  binary mismatch at link stage. 

Personally I'd suggest ask the users to define `TI_WITH_CUDA` in thier CMakeLists.txt - similar to what we have done in taichi-aot-demo

jim19930609: (One more issue)
This also gonna invalidate all the `TI_WITH_CUDA` in `src/taichi_llvm_impl.cpp` since it includes `taichi_cuda.h` - all the `TI_WITH_CUDA` will be override to `ON` no matter what we set `TI_WITH_CUDA` in `TAICHI_CMAKE_ARGS`

python/taichi/examples/graph/texture_graph.py
```diff
@@ -79,7 +79,7 @@ def main():
         mod.save(tmpdir, '')
     else:
         t = 0.0
-        window = ti.ui.Window('UV', res)
+        window = ti.ui.Window('UV', res, True)
```
ailzhang, line 82: nit: is this `True` somehow required? If so it's better to write `vsync=True` ;)

ailzhang (approved):
LGTM!

jim19930609 (approved):
LGTM!

PR #6774, opened 2022-11-30
[vulkan] Make sure kernel recompiles when texture dtype changes

ailzhang:
Issue: #

### Brief Summary
As discussed with @PENGUINLIONG offline, kernel should recompile when `dtype & num_dimension` changes. Note currently `rwtexture` also recompiles if `num_channels` change, I didn't remove that in this PR.


PENGUINLIONG (approved):
lgtm, but actually i'm gonna deprecate `dtype` + `num_channel` and replace that with `format` solely in the future.

PR #6777, opened 2022-11-30
[doc] Create a demo in git tutorial

ailzhang:
Issue: #


setup.py
```diff
@@ -1,4 +1,4 @@
-# Optional environment variables supported by setup.py:
+# Optional envionment variables supported by setup.py:
```
ailzhang, line 1:
```suggestion
# Optional environment variables supported by setup.py:
```

PR #6780, opened 2022-11-30
[example] Redundant weight size in rgb layer of ngp example

erizmr:
The output size of the rgb layer is 3, however a size of 4 is allocated for the shared array of the weight.

@Linyou Would you mind helping double check? Thanks!


Linyou:
Yes, it actually only uses the first 3 elements, so is OK to change to 3 if it does not affect the performance.

erizmr:
Thanks for the reply. As mentioned in https://github.com/Linyou/taichi-ngp-renderer/issues/3, if it is for memory alignment use, I think it may be better to have it as is. :)

PR #6782, opened 2022-12-01
[vulkan] Deprecate num_channels and channel_format args in rw_texture type annotation

ailzhang:
Issue: #

### Brief Summary
#6338 added a way to specify fmt (`ti.Format`) which is more expressive and easier to understand than the combo of `num_channels + channel_format` so let's deprecate them to avoid confuse users.


PENGUINLIONG (commented):
LGTM, thank you! 🤩

PENGUINLIONG (approved):

PR #6784, opened 2022-12-01
[misc] Remove the default potential_bug label on bug report issues

ailzhang:
Issue: #

### Brief Summary
Unlike other issue templates, bug report issues can be very diverse and `potential bug` label isn't that helpful. We'd rather assign a proper label manually or make it more automatic in the future.


feisuzhu (approved):
LGDM

PR #6789, opened 2022-12-02
[vulkan] Support rw_texture in aot add_kernel

ailzhang:
Issue: #

### Brief Summary
https://github.com/taichi-dev/taichi/pull/6782 accidentally makes `channel_format/num_channels` optional and this PR fixes it by throwing an error if none of
`channel_format/num_channels/fmt` is specified.

Note `rw_texture` isn't a template type (yet) so it's okay to just compile it directly with its type hint.

TODO: support `ti.types.texture()` in add_kernel


PENGUINLIONG (approved):
lgtm!

PR #6796, opened 2022-12-02
[vulkan] Support texture type args in aot add_kernel

ailzhang:
Issue: #

### Brief Summary
`ti.types.texture` has been a template type so I didn't change that in this PR. But in the future `TextureType` may support args other than `num_dimensions` so that users can provide all necessary info in the type hint. But that's out of scope for this PR.

This PR just added support for texture type args in aot `add_kernel` so that it can be used in AOT without cgraph, and it worth noticing that you're currently required to provide a template_args for texture type. But since we require enough info in rw_texture type annotation, you don't need to provide any template_args for rw_texture type.


PENGUINLIONG (approved):
thank you! 🤩

PR #6800, opened 2022-12-03
[doc] Update docstring for grad replaced

erizmr:
Update docstring for grad replaced from `ti.grad_replaced` to `ti.ad.grad_replaced`.


ailzhang (approved):
Thanks!

PR #6808, opened 2022-12-05
[test] Remove unnecessary duplicated python runtime test runs

ailzhang:
Issue: #

### Brief Summary
Just noticed these tests always run on x64 during my local runs, even if I'm running on OpenGL backend. 
The reason is mainly due to the missing `test_utils.test()` guard so let's guard them to save some time.


feisuzhu (approved):
LGTM

PR #6812, opened 2022-12-06
[ir] Add ir_traits namespace to use less dynamic casts & Run CFG only ever once

bobcao3:
This speeds up CFG a little bit by removing most of the dynamic casts, and create `Load` and `Store` IR Traits that can be applied to statements that have this trait. (4~5s out of ~1.2 minutes)

In addition, since most of the time in the load/store source/destination analysis we only return a single statement or even none, using vector as a return type seems excessive. Here we create a `one_or_more` template that is a variant between one or more (vector of ones), and we apply the `begin`, `end`, `empty`, and `size` functions so that this can directly plug into our code, and we can directly use range-for loops on them.

Then we should disable to `while (!modified)` loop for CFG passes and only ever run them once in `full_simplify`. I observed no performance degradation with this change while reducing compile time further down to around 45s


k-ye (commented):
I think it's good to create small helpers like `Load`/`Store`. Can't really tell whether they are generic enough as IR building blocks, but we will see :-)

taichi/ir/statements.h
```diff
@@ -535,7 +576,9 @@ class AssertStmt : public Stmt {
 /**
  * Call an external (C++) function.
  */
-class ExternalFuncCallStmt : public Stmt {
+class ExternalFuncCallStmt : public Stmt,
```
k-ye, line 579: I wonder if concept would also allow us to avoid multi-inheritance..

taichi/util/meta.h
```diff
@@ -83,6 +85,84 @@ struct is_specialization : std::false_type {};
 template <template <class...> class Template, class... Args>
 struct is_specialization<Template<Args...>, Template> : std::true_type {};
 
+template <typename T, class Container = std::vector<T>>
```
k-ye, line 88: nit: better put this into a dedicated file, e.g. `one_or_more.h` in the future?

k-ye (approved):
LGTM!

taichi/ir/statements.h
bobcao3: I think it should? Since these "Traits" are tied to actual C++ types (not runtime)

turbo0628:
Compared fill, mpm and nbody performance on RTX3080 against 1.3.0, CUDA and Vulkan backends. No regression observed.

LGTM!

PR #6822, opened 2022-12-07
[bug] Matrix refactor bug fix: Fix cross scope matrix operations

jim19930609:


taichi-gardener:

python/taichi/lang/ops.py
```diff
@@ -7,7 +7,26 @@
...
+def uniform_matrix_inputs(*args):
+    has_real_matrix = False
+    for arg in args:
+        if is_taichi_expr(arg) and arg.ptr.is_tensor():
+            has_real_matrix = True
+            break
+
+    results = []
+    for arg in args:
+        if has_real_matrix and is_matrix_class(arg):
```
strongoier, line 23: Can we simply use `isinstance(arg, Matrix)` instead of adding `is_matrix_class`?

jim19930609: Interesting, the root reason is that Python doesn't support separate declaration and definition, therefore circular import can be a real headache. 

However, if it supports separate declaration and definition, we'll then suffer from symbol inconsistency instead :(

PR #6824, opened 2022-12-07
[opt] Re-enable constant folding when debug=True

ailzhang:
I tried this locally and it no longer fails so try removing the hack to see if anything breaks.

Issue: #

### DIFF
--- a/tests/python/test_assert.py
+++ b/tests/python/test_assert.py
@@ -107,6 +107,19 @@ def test_assert_ok():
     func()
 
 
+@test_utils.test(require=ti.extension.assertion,
+                 debug=True,
+                 check_out_of_bound=True,
+                 gdb_trigger=False)
+def test_assert_with_check_oob():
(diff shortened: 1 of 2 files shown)


tests/python/test_assert.py
```diff
@@ -107,6 +107,19 @@ def func():
     func()
 
 
+@test_utils.test(require=ti.extension.assertion,
+                 debug=True,
+                 check_out_of_bound=True,
+                 gdb_trigger=False)
+def test_assert_with_check_oob():
```
strongoier, line 114: What is the difference between this test and `test_assert_ok` above? IIRC `check_out_of_bound` is implicitly true when `debug=True`.

ailzhang: @strongoier this one is added since it's the exact repro which errored out before so I guess it's just to make sure it disappeared.

strongoier (approved):

PR #6827, opened 2022-12-07
[example] Solve implicit fem using sparsee solver

FantasyVR:
### Brief Summary
The original implicit fem example is actually a projective dynamics simulation. The system matrix could be constructed and factorized before starting the simulation. In this PR, the system matrix is explicitly constructed and factorized. The linear system could be solved using a sparse solver on the CPU/GPU backend.


BillXu2000 (approved):
LGTM

Issue #6793, opened 2022-12-02
[RFC] Roadmap to minimize libtaichi_c_api.so library size


jim19930609:
We roughly planed three stages to optimize the library size for libtaichi_c_api.so:
- Function level linking
- Symbol rearrangement and clean up
- Customized C-API features pruning

# 1. Turn on function-level linking
## 1.1 Convert CMake OBJECT targets to STATIC targets
Removes unnecessary entry files that compiles into libtaichi_c_api.so

[Before]
<img width="742" alt="image" src="https://user-images.githubusercontent.com/22334008/205221958-0e58cc7e-4b33-4ea7-a38d-68339399c5b3.png">

[After]
<img width="759" alt="image" src="https://user-images.githubusercontent.com/22334008/205222135-559241c1-71fa-4c25-9f02-77ee2be4bc28.png">

## 1.2 Apply version script to libtaichi_c_api.so
Convert unnecessary global symbols to local symbols, leading to more optimization opportunities.
[Before]
<img width="760" alt="image" src="https://user-images.githubusercontent.com/22334008/205222164-349cc48e-f621-4e59-8535-2298b65f39e6.png">

[After]
<img width="754" alt="image" src="https://user-images.githubusercontent.com/22334008/205222200-26165a73-a186-4031-8141-7d6598994878.png">

## 1.3 Add compile option `-ffunction-sections`, `-fdata-sections` and link option `--gc-sections`
Turns on function level linking eventually.
[Before]
<img width="751" alt="image" src="https://user-images.githubusercontent.com/22334008/205222247-354363d3-8553-4949-addc-515d65d0fade.png">

[After]
<img width="749" alt="image" src="https://user-images.githubusercontent.com/22334008/205222277-1161c4a6-3856-4a65-87dd-761282fa69f3.png">

Based on some experiment results, after turning on function-level-linking, we're expecting ~80% code size reduction from 12 MB to 1.7 MB (Vulkan C-API library, https://github.com/taichi-dev/taichi/pull/6388)

# 2. Symbol rearrangement and clean up
Further detect and clean up redundant symbols via symbol analysis (tooling: https://github.com/google/bloaty)

Symbol class | Symbol keyword | Size / KB | Side notes / KB
-- | -- | -- | --
libtaichi_c_api.so | [Total Size] | 1700 |  
 
RTTI + vtable | "vtable for" | 27 |  
RTTI + vtable | "typeinfo name for" | 9 |  
[...]

PR #6831, opened 2022-12-07
[build] Migrate cmake targets from OBJECT to STATIC for libtaichi_c_api.so

jim19930609:
Issue: https://github.com/taichi-dev/taichi/issues/6793


PENGUINLIONG (approved):
Thank you!

PR #6835, opened 2022-12-08
[test] Avoid constant folding in overflow tests

ailzhang:
Issue: #6824

### Brief Summary
https://github.com/taichi-dev/taichi/actions/runs/3641543203/jobs/6154890343 

We used to depend on the behavior that "no constant folding when debug=True"  in these tests and this PR fixes by avoid constant folding in the tests. 

Tested locally with ` python tests/run_tests.py overflow --with-offline-cache`


PGZXB (approved):
Thanks!

PR #6839, opened 2022-12-08
[bug] MatrixType bug fix: Fix error with static-grouped-ndrange

jim19930609:


python/taichi/lang/_ndrange.py
```diff
@@ -139,10 +139,12 @@ def ndrange(*args):
...
-class GroupedNDRange:
+class GroupedNDRangeBase:
     def __init__(self, r):
         self.r = r
 
+
+class GroupedNDRange(GroupedNDRangeBase):
     def __iter__(self):
         for ind in self.r:
             if impl.current_cfg().real_matrix:
```
strongoier, line 150: IIUC the `GroupedNDRange` class is used only in `ti.static`. So it seems that removing the `real_matrix` case here is already enough and no other modification is needed.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -1060,7 +1060,7 @@ def build_static_for(ctx, node, is_grouped):
                     f"Group for should have 1 loop target, found {len(targets)}"
                 )
             target = targets[0]
-            for value in impl.grouped(ndrange_arg):
+            for value in impl.static(impl.grouped(ndrange_arg)):
```
strongoier, line 1063:
```suggestion
            for value in impl.grouped(ndrange_arg):
```
Now we don't need to call `static` in Python.

tests/python/test_ast_refactor.py
```diff
@@ -1039,14 +1061,23 @@ def foo() -> ti.i32:
...
 
-@test_utils.test()
-def test_grouped_static_for_cast():
+def _test_grouped_static_for_cast():
     @ti.kernel
     def foo() -> ti.f32:
         ret = 0.
         for I in ti.static(ti.grouped(ti.ndrange((4, 5), (3, 5), 5))):
-            tmp = I.cast(float)
+            tmp = ti.cast(I, float)
```
strongoier, line 1069:
```suggestion
            tmp = I.cast(float)
```
The original code should also work.

strongoier (approved):

PR #6845, opened 2022-12-09
[build] Enable strip for libtaichi_c_api.so with Release Build

jim19930609:
Issue: https://github.com/taichi-dev/taichi/issues/6793

  | Size (Original) | Size (Optimized)
-- | -- | --
Vulkan | 14 MB | 1.5 MB
Vulkan + LLVM | 82 MB | 51 MB

libtaichi_c_api.so built with LLVM does not benefit much from function level linking, mainly because libLLVM.a is not compiled with `-ffunction-sections -fdata-sections`. We'll have that fixed later.


cmake/TaichiCAPI.cmake
```diff
@@ -19,9 +19,9 @@ endif()
 
 endfunction()
 
-
 set(TAICHI_C_API_NAME taichi_c_api)
 
+set(TAICHI_C_API_NAME taichi_c_api)
```
ailzhang, line 24: duplicated?

ailzhang (approved):
Thanks!

PR #6848, opened 2022-12-09
[ci] Workflow Rewrite: Building on Linux

feisuzhu:
Issue: #6445


.github/workflows/scripts/build.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/python3 -u
```
ailzhang, line 1: nit: hmmm this implicit deps on /usr/bin/python3 is a bit concerning, IMHO I'd prefer to be explicit about which python we use in CI :P

.github/workflows/scripts/build.py
```diff
@@ -0,0 +1,92 @@
...
+
+# -- prioritized --
+import sys
+import os.path
+sys.path.insert(0, os.path.join(os.path.dirname(__file__)))
```
ailzhang, line 6: agh would be nice if we can avoid manipulating `sys.path` as well :P

.github/workflows/scripts/build.py
```diff
@@ -0,0 +1,92 @@
...
+
+import common
```
ailzhang, line 8: maybe rename this folder to a more longer name like `ci_common`? Sometimes it's not so straightforward for readers to distinguish native modules and folders in the same path from names :P

.github/workflows/scripts/build.py
```diff
@@ -0,0 +1,92 @@
...
+        out = get_cache_home() / 'llvm15-m1'
+        url = 'https://github.com/taichi-dev/taichi_assets/releases/download/llvm15/taichi-llvm-15-m1.zip'
+    elif (u.system, u.machine) == ('Darwin', 'x86_64'):
+        out = get_cache_home() / 'llvm15-mac'
+        url = 'https://github.com/taichi-dev/taichi_assets/releases/download/llvm15/llvm-15-mac10.15.zip'
+    else:
+        raise RuntimeError(f'Unsupported platform: {u.system} {u.machine}')
+
+    download_dep(url, out, strip=1)
+    env['LLVM_DIR'] = str(out)
```
ailzhang, line 58: shall we just return LLVM_DIR to the caller instead of changing the environment variable directly? ;) 
In that case we can easily aggregate all the changes to environment variables in one function and it'll be easier for debugging I believe.

.github/workflows/scripts/build.py
```diff
@@ -0,0 +1,92 @@
...
+        if common.is_manylinux2014():
+            extra.extend(['-p', 'manylinux2014_x86_64'])
+        else:
+            extra.extend(['-p', 'manylinux_2_27_x86_64'])
+
+    python('misc/make_changelog.py', '--ver', 'origin/master', '--repo_dir', './', '--save')
+    python('setup.py', *proj_tags, 'bdist_wheel', *extra)
+
+
+setup_llvm()
```
ailzhang, line 84: Let's wrap these in the main function ;)

.github/workflows/scripts/build.py
```diff
@@ -0,0 +1,92 @@
...
+        url = 'https://github.com/taichi-dev/taichi_assets/releases/download/llvm15/llvm-15-mac10.15.zip'
+    else:
+        raise RuntimeError(f'Unsupported platform: {u.system} {u.machine}')
+
+    download_dep(url, out, strip=1)
+    env['LLVM_DIR'] = str(out)
+
+
+@banner('Build Taichi Wheel')
+def build_wheel(python, pip):
```
ailzhang, line 62: nit: need docstring, even better if these newly added python script are equipped with type hints  ;)

.github/workflows/scripts/common/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .bootstrap import init
+init()
```
ailzhang, line 2: ditto: let's wrap this in the main so that it's friendly for future imports from other scripts ;)

.github/workflows/scripts/common/escapes.py
```diff
@@ -0,0 +1,58 @@
+"""
+Generates a dictionary of ANSI escape codes.
+
+http://en.wikipedia.org/wiki/ANSI_escape_code
+
+Uses colorama as an optional dependency to support color on Windows
+
+---
+Copied from colorlog
+---
```
ailzhang, line 10: Hmm is it only used for pretty banner (and not related to ci/test)? If so let's add a simple line here explaining it's for banner only ;)

ailzhang (commented):
Awesome!!

.github/workflows/scripts/build.py
feisuzhu: Python running this build script and python actually running `setup.py` is not the same one. I'm writing this script with portability in mind, even dependencies are installed automatically, so don't worry :)

.github/workflows/scripts/build.py
feisuzhu: No, if we remove this hack the script would only work when current directory is at where the script lies. Or we have to cram all the dependencies into a single file which is apparently a worse solution.

.github/workflows/scripts/common/__init__.py
feisuzhu: This have to be here since it is doing some very early initialization (mainly automatically installing missing dependencies), put it into `main` would defeat its purpose.

.github/workflows/scripts/build.py
feisuzhu: I'll fix this particular site but leave others alone.

After a casual think I found it troublesome to strictly aggregate all env variables and change them at once:

1. `PATH` change in `setup_python` should immediately take effect or it would be a big surprise that the returned `python` and `pip` command couldn't work before env variable changes are applied.
2. `TAICHI_CMAKE_ARGS` is a hot spot and have been changed at multiple sites, aggregating them would require a not-so-trivial 'EnvVariableManager' to do this, that's overkill.

Debugging is a valid concern though, I'll just print the change instead of trying to 'manage' them.

feisuzhu: Hmm, I've changed my mind :|

After trying I found the code have a big smell of abstraction leak:

```python
def main() -> None:
    llvm_dir = setup_llvm()
    env['LLVM_DIR'] = str(llvm_dir)
    sccache = setup_sccache()
    python, pip = setup_python(env['PY'])
    build_wheel(python, pip)
    sccache('-s')
```

Reading code above, I think manipulating `LLVM_DIR` is definitely a wanted side-effect of `setup_llvm` here, at least such detail shouldn't appear in the `main` function.

.github/workflows/scripts/common/__init__.py
ailzhang: oops my bad, this one doesn't matter at all :D

.github/workflows/scripts/build.py
ailzhang: Hmmm correct me if I'm wrong, when you run `.github/workflows/scripts/build.py` in testing.yml it's implicitly using `/usr/bin/python3` right? IMHO others who modify this script may wonder which python is being used?

.github/workflows/scripts/build.py
ailzhang: Aha I see your concern, that's totally valid. One potential solution I have in mind is passing a dict of envs into the function instead of modifying the global process's environ on the fly. `PATH` is a valid exception but other envs like `LLVM_DIR/TAICHI_CAMKE_ARGS` can be set on the subprocess (if applicable) which runs `setup.py` only.  That way we don't have to pollute the process environ just in case we start other subprocesses which may inherit the environ from it, wdyt? 
Note this isn't a blocker for this PR to merge so I'm happy to take followup PRs for it, but just want to start the discussion here to make sure the ci infra is robust ;)

.github/workflows/scripts/build.py
feisuzhu: > Hmmm correct me if I'm wrong, when you run `.github/workflows/scripts/build.py` in testing.yml it's implicitly using `/usr/bin/python3` right? 

Right

> IMHO others who modify this script may wonder which python is being used?

There's no confusion here I think, `/usr/bin/python3` will point to system/container provided Python.

.github/workflows/scripts/build.py
feisuzhu: > ... `LLVM_DIR/TAICHI_CAMKE_ARGS` can be set on the subprocess (if applicable) which runs `setup.py` only. That way we don't have to pollute the process environ just in case we start other subprocesses which may inherit the environ from it, wdyt? 

Agree, but I don't think it worth the extra complexity and effort.  A simple try results pseudo-code below...

```python
def main() -> None:
    cmake_args = {}
    buildenv = {}
    cmake_args.update(split_cmake_args(env.get('TAICHI_CMAKE_ARGS', '')))
    llvm_dir, llvm_cmake_args = setup_llvm()
    cmake_args.update(llvm_cmake_args)
    buildenv['LLVM_DIR'] = llvm_dir
    sccache, sccache_cmake_args = setup_sccache()
    cmake_args.update(sccache_cmake_args)
    python, pip = setup_python(env['PY'])
    buildenv['TAICHI_CMAKE_ARGS'] = join_cmake_args(cmake_args)
    build_wheel(python, pip, buildenv)
    sccache('-s')

    distfiles = glob.glob('dist/*.whl')
    if len(distfiles) != 1:
        raise RuntimeError(f'Failed to produce exactly one wheel file: {distfiles}')
```

... which has significant more noises compared to current version.

feisuzhu: In bash environment vars always pollute all subprocesses but I have yet run into an issue because of this. Bash shell script writers understand this so they would careful design environment variable names.

feisuzhu: If env variable pollution is really a concern, I would like to use a whitelist mechanism(like what sudo does):

```python
with passthrough_envs(['TAICHI_CMAKE_ARGS', 'LLVM_DIR', 'SCCACHE_DIR']):
    python('setup.py', 'bdist_wheel')
```

ailzhang (approved):
Awesome!!

PR #6850, opened 2022-12-09
[bug] Fix flaky mass_spring_game_ggui.py on Mac M1 by setting up default values for VulkanCapabilities

jim19930609:
Mac M1 does not support `wide_line` with Vulkan. However, device capability was initialized to random values which is why some examples are flaky on M1.

Before this PR: 
image

image

### DIFF
--- a/taichi/rhi/vulkan/vulkan_device.h
+++ b/taichi/rhi/vulkan/vulkan_device.h
@@ -562,12 +562,12 @@ class VulkanStream : public Stream {
 };
 
 struct VulkanCapabilities {
-  uint32_t vk_api_version;
-  bool physical_device_features2;
-  bool external_memory;
-  bool wide_line;
-  bool surface;
-  bool present;
+  uint32_t vk_api_version = 0;
+  bool physical_device_features2 = false;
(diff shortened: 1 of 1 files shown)


PENGUINLIONG (approved):
Thank youuuuu

taichi/rhi/vulkan/vulkan_device.h
```diff
@@ -562,12 +562,12 @@ class VulkanStream : public Stream {
 };
 
 struct VulkanCapabilities {
-  uint32_t vk_api_version;
-  bool physical_device_features2;
-  bool external_memory;
-  bool wide_line;
-  bool surface;
-  bool present;
+  uint32_t vk_api_version = 0;
```
bobcao3, line 565: Nit: should use the bracket initializer syntax

`uint32_t vk_api_version{0}`

bobcao3:
> Question here: Why compiler didn't report 'use of uninitialized values' for the original code?

I think this is a struct and it's "default initialized" as a struct object, just that the struct did not specify initial values of its members

ailzhang (approved):

Issue #6841, opened 2022-12-08
For dynamic snodes, calling to_numpy() results a whole array padded with zeros, not the actual length


neozhaoliang:
Let's say we have an empty dynamic snode like this:
```python
import taichi as ti
ti.init(arch=ti.cpu)
x = ti.field(int)
ti.root.dynamic(ti.i, 1024, chunk_size=32).place(x)

@ti.kernel
def test():
    print(x.length())  #0

arr = x.to_numpy()  # array of size 1024
```

This behavior may cause problems, for example, using `arr` to draw `gui.circle` or `gui.lines` will give false results at the origin.
We should make the returned array have the same size with the actual length of the dynamic snode.

neozhaoliang:
This might be difficult to solve on the python frontend. Since `to_numpy` must be called in the python scope, and we cannot know the actual length of a dynamic snode in the python scope. Also we cannot define a suitable length when the dynamic snode is a multidimensional array. Should better raise a warning message here.

PR #6853, opened 2022-12-09
[Lang] Add warning message when converting dynamic snode to numpy

neozhaoliang:
This PR adds a warning message when the user tries to call the `to_numpy` method of a dynamic snode.

close: #6841


jim19930609 (approved):
thanks

PR #6870, opened 2022-12-12
[example] Add circle-packing example

neozhaoliang:
This PR adds a new example.

Figure_1


jim19930609 (approved):
Pip Installation for `cairocffi` on MacOs and Windows are both pretty troublesome though (https://www.csestack.org/macbook-m1-oserror-no-library-called-cairo-pango-was-found/), but works fine on Linux

PR #6873, opened 2022-12-12
[test] Remove tests with real_matrix=True and real_matrix_scalarize=True

strongoier:
Issue: #5819

### Brief Summary

As these two options are enabled by default (#6801), we no longer need separate tests for them.


jim19930609 (approved):
Cool, thanks!

PR #6886, opened 2022-12-13
[lang] Reorder sparse matrix before solving

FantasyVR:
Issue: #2906 

### Brief Summary
When solving a sparse linear system without reordering the system matrix, `analyze_pattern` and `factorize` would consume a lot of GPU memory. For example, the `res` in the `stable fluid` example can not be larger than 500. After reordering the sparse system matrix, we can set `res=512` now.


taichi/program/sparse_solver.cpp
```diff
@@ -210,12 +212,68 @@ void CuSparseSolver::analyze_pattern(const SparseMatrix &sm) {
   CUSPARSEDriver::get_instance().cpSetMatIndexBase(descr_,
                                                    CUSPARSE_INDEX_BASE_ZERO);
 
-  // step 1: create opaque info structure
+  // step 1: reorder the sparse matrix
+  float *h_csrValA = NULL;
```
strongoier, line 216: Let's use `nullptr`.

strongoier (approved):

Hanke98 (approved):
The changes look good to me ! Thanks!
PS: Sorry for the delayed response! I am busy working on my paper recently.

FantasyVR:
> The changes look good to me ! Thanks! PS: Sorry for the delayed response! I am busy working on my paper recently.

hhha, Never mind. If you don't have time, you can just ignore the review request.

PR #6887, opened 2022-12-13
[ci] Reenable AMDGPU CI, disable OpenGL tests in AMDGPU task

feisuzhu:
AMDGPU node is back.
And I couldn't make OpenGL work in container (probably because of mesa version / other user space AMDGPU driver), disabling for now.


ailzhang (approved):
Thanks!

PR #6893, opened 2022-12-13
[aot] Enable validation layers for C-API tests

jim19930609:
Issue: #


c_api/src/taichi_core_impl.cpp
```diff
@@ -225,7 +225,15 @@ TiRuntime ti_create_runtime(TiArch arch) {
   switch (arch) {
 #ifdef TI_WITH_VULKAN
     case TI_ARCH_VULKAN: {
-      out = (TiRuntime)(static_cast<Runtime *>(new VulkanRuntimeOwned));
+#ifdef TI_BUILD_TESTS
```
ailzhang, line 228: agh we'd better separate enabling validation layer with TI_BUILD_TESTS, how about we create another macro to control validation and let TI_IN_CI control that?

ailzhang (commented):
LGTM!

.github/workflows/scripts/win_build.ps1
```diff
@@ -72,11 +72,11 @@ if ($llvmVer -eq "10") {
 $env:TAICHI_CMAKE_ARGS += " -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang"
 
 if ($installVulkan) {
-    $env:VULKAN_SDK = "C:\VulkanSDK\1.2.189.0"
+    $env:VULKAN_SDK = "C:\VulkanSDK\1.3.236.0"
```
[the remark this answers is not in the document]
jim19930609: Looks like so?

image

CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 option(USE_LLD "Use lld (from llvm) linker" OFF)
 option(USE_MOLD "Use mold (A Modern Linker)" OFF)
+option(TI_BUILD_CI "(Internal Use) Enable certain validations and possible debuggers for Taichi CI system" OFF)
```
PENGUINLIONG, line 67: I somehow recall there is a TI_CI macro or something like that already. Can we reuse that name?

jim19930609: Ah that's right, thanks!

PENGUINLIONG (approved):
Thank you!

PR #6894, opened 2022-12-13
[Lang] Deprecate sourceinspect dependency

neozhaoliang:
This PR replaces #6630 (now closed)

It deprecates, and also fixes a bug in the sourceinpect module (cannot use cell magics %%time and %%timeit in jupyter notebooks). The dependency on sourceinspect is reserved and will be removed in future releases (if the current fix works fine)

Also added test in `test_deprecation.py`.


jim19930609 (approved):
thanks!

PR #6895, opened 2022-12-13
[example] Add simplex noise example

neozhaoliang:
This PR adds a new example based on simplex noise.

Result:

res


python/taichi/examples/simulation/alien_life.py
```diff
@@ -0,0 +1,174 @@
...
+time = 0
+theta = np.linspace(0, 2 * np.pi, 200)
+begin = np.array([np.cos(theta)[:-1], np.sin(theta[:-1])]).transpose()
+end = np.array([np.cos(theta)[1:], np.sin(theta[1:])]).transpose()
+begin = (begin + 1) / 2
+end = (end + 1) / 2
+
+while gui.running:
+    time += 1
+    update(time * 0.002, intensity, initial_positions)
```
bobcao3, line 171: This makes a copy of `intensity` and `initial_positions` every time we call the kernel. Although they are GCed, we shouldn't encourage this usage because it hinders performance

bobcao3 (changes requested):
Please change the ndarray copy thing and make them a field as well, or make them a Taichi ndarray

PR #6905, opened 2022-12-15
[ci] Try enabling MSVC and check build times

bobcao3:
Issue: #


.github/workflows/testing.yml
```diff
@@ -344,7 +344,7 @@ jobs:
           - os: windows-2019
             llvmVer : '15'
             archs: "cpu,cuda,opengl"
-            runsOn: [self-hosted, windows, cuda, OpenGL]
+            runsOn: [self-hosted, windows, cuda, OpenGL, msvc-debug]
```
feisuzhu, line 347: Should restore it

.github/workflows/scripts/win_build.ps1
```diff
@@ -69,7 +69,8 @@ if ($llvmVer -eq "10") {
     throw "Unsupported LLVM version"
 }
 
-$env:TAICHI_CMAKE_ARGS += " -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang"
+$env:TAICHI_USE_MSBUILD = 1
```
feisuzhu, line 72: Where is it used? I can only see assignments but not reads.

feisuzhu (changes requested):

.github/workflows/scripts/win_build.ps1
bobcao3: It's a flag for `setup.py`

feisuzhu (approved):
LGTM

PR #6906, opened 2022-12-15
[autodiff] Support grad in ndarray

PhrygianGates:
# TODO


taichi/program/context.h
```diff
@@ -109,7 +122,26 @@ struct RuntimeContext {
...
+    TI_ASSERT(shape.size() <= taichi_max_num_indices);
+    size_t total_size = 1;
+    for (int i = 0; i < shape.size(); i++) {
+      extra_args[arg_id][i] = shape[i];
+      total_size *= shape[i];
+    }
+    set_array_runtime_size(arg_id, total_size);
+  }
+
+  void set_arg_ndarray(int arg_id,
```
erizmr, line 137: It seems the code of this two `set_arg_ndarray` overlap largely. Shall we merge them into one ?

erizmr (commented):
Thanks!!

python/taichi/lang/kernel_impl.py
```diff
@@ -642,7 +642,15 @@ def func__(*args):
                 elif isinstance(needed,
                                 ndarray_type.NdarrayType) and isinstance(
                                     v, taichi.lang._ndarray.Ndarray):
-                    launch_ctx.set_arg_ndarray(actual_argument_slot, v.arr)
+                    has_external_arrays = True
```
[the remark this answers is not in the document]
PhrygianGates: It's not useful any more and I'll just delete it.

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -1264,6 +1266,18 @@ void TaskCodeGenLLVM::visit(ArgLoadStmt *stmt) {
   } else {
     llvm_val[stmt] = bitcast_from_u64(raw_arg, stmt->ret_type);
   }
+
+  // print
```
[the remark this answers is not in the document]
PhrygianGates: Yes, I'll remove that.

taichi/ir/frontend_ir.h
```diff
@@ -446,31 +446,39 @@ class ExternalTensorExpression : public Expression {
...
 
       auto tensor_type =
           taichi::lang::TypeFactory::get_instance().create_tensor_type(
               element_shape, dt);
-      init(tensor_type, dim, arg_id, element_dim);
+      init(tensor_type, dim, arg_id, element_dim, is_grad);
     }
   }
 
+  ExternalTensorExpression(Expr *expr, bool is_grad = true) {
```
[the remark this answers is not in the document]
PhrygianGates: Because [here](https://github.com/taichi-dev/taichi/blob/094c05a56a0cfe7f5b3c0d00525b48e3238e699d/python/taichi/lang/any_array.py#L43) we want to construct a new grad ExternalTensorExpression by just using its primal.

taichi/transforms/simplify.cpp
```diff
@@ -630,6 +630,7 @@ void full_simplify(IRNode *root,
         modified = true;
       if (die(root))
         modified = true;
+      // here
```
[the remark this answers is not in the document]
PhrygianGates: ok!

taichi/program/context.h
PhrygianGates: ok！

erizmr: It seems that the `has_grad[arg_id] = false` invoked by `ctx_->set_arg_ndarray(arg_id, ptr, arr.shape)`  will overwrite the `has_grad[arg_id] = true` before hand ?

erizmr (commented):
Thanks. Could you please also add some tests for the new rule for alias analysis of `ExternalPtrStmt` ?  FYI, test samples: https://github.com/taichi-dev/taichi/blob/master/tests/cpp/analysis/alias_analysis_test.cpp

taichi/program/context.h
PhrygianGates: Is it possible that we invoke `ctx_->set_arg_ndarray(arg_id, ptr, arr.shape)` after `set_arg_ndarray_with_grad(actual_argument_slot, v_primal, v_grad)` for a same variable?

PhrygianGates: I think it may overwrite if we do so.

erizmr (approved):
LGTM! Nice work!

taichi-gardener:

bobcao3:
I don't like the idea of augmenting ArgLoadStmt with an autodiff attribute. This complicates the semantics of CHI-IR too much, is it possible for grad ndarrays to just take two arguments instead of adding another set of arguments?

bobcao3:
Also due to recent CI issues please merge the latest master. I tried to run `/rebase` but it failed.

erizmr:
I recall the motivation to add a set of `grad_args` is to link the grad ndarray to primal ndarray during the kernel compilation. Otherwise, users are expected to pass all grad arrays as an additional arguments.

erizmr:
Sure. @PhrygianGates could you please help rebase the PR to the latest master?

PhrygianGates:
Yes, because the pointer to the ndarray points to a raw array on a device, there is no way to get the grad information as the snode does. One way to do this is to add the grad argument to the kernel argument automatically when passing the reference, or to add a grad_args and augment argloadstmt.

PhrygianGates:
Of course I'll rebase it.

PR #6909, opened 2022-12-16
[doc] Update debugging.md

vai9er:
Issue: #

### DIFF
--- a/docs/lang/articles/debug/debugging.md
+++ b/docs/lang/articles/debug/debugging.md
@@ -4,18 +4,18 @@ sidebar_position: 1
 
 # Debugging
 
-Taichi provides the following mechanisms to facilitate parallel programming debugging:
+To aid with simultaneous programming debugging, Taichi has the following mechanisms:
 
-- `print` in the Taichi scope checks the value of a variable.
-- Serialization of your program or a specific parallel for loop.
-- Activated with `ti.init(debug=True)`, debug mode detects out-of-bound array accesses.
-- Static or non-static `assert` verifies an assertion condition at compile time or runtime respectively.
-- `sys.tracebacklimit` produces a conciser traceback.
+1. `print` in the Taichi scope checks the value of a variable.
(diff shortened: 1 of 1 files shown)


docs/lang/articles/debug/debugging.md
```diff
@@ -225,13 +225,13 @@ When you are done with debugging, set `debug=False`. Then, the program ignores a
 
 ## Compile-time `ti.static_assert`
 
+Taichi, like `ti.static print`, includes a static version of `assert`; `ti.static assert` which may be used to create assertions on data types, dimensionality, and shapes.
+
```
neozhaoliang, line 229:
```suggestion
Taichi, like `ti.static_print`, includes a static version of `assert`; `ti.static_assert`, which may be used to create assertions on data types, dimensionality, and shapes.

```

docs/lang/articles/debug/debugging.md
```diff
@@ -225,13 +225,13 @@ When you are done with debugging, set `debug=False`. Then, the program ignores a
...
 ```python
 ti.static_assert(cond, msg=None)
 ```
 
-Like `ti.static_print`, Taichi also provides a static version of `assert`: `ti.static_assert`, which is useful to make assertions on data types, dimensionality, and shapes. It works regardless of whether `debug=True` is set or not. A false statement of `ti.static_assert` triggers an `AssertionError`, just as a false `assert` statement in the Python scope does.
-
-For example:
+It works whether or not `debug=True` is used. A false `ti.static assert` statement, like a false `assert` statement in the Python scope, causes a ` AssertionError`, as shown below:
```
neozhaoliang, line 234:
```suggestion
It works whether or not `debug=True` is used. A false `ti.static_assert` statement, like a false `assert` statement in the Python scope, causes an ` AssertionError`, as shown below:
```

neozhaoliang (approved):
LGTM!

PR #6912, opened 2022-12-16
[vulkan] Improve Vulkan RHI impl with lower overhead internal implementations

bobcao3:
Also added partial RHI support for Vulkan Dynamic Rendering, but there are some interactions and bugs to figure out between this and ImGUI, so the device creator will not actively enable this feature for now.

### DIFF
--- a/taichi/rhi/device.h
+++ b/taichi/rhi/device.h
@@ -17,6 +17,7 @@ enum class RhiResult {
   error = -1,
   invalid_usage = -2,
   not_supported = -3,
+  out_of_memory = -4,
 };
 
 constexpr size_t kBufferSizeEntireSize = size_t(-1);
@@ -54,7 +55,7 @@ struct LLVMRuntime;
 
 // TODO: Figure out how to support images. Temporary solutions is to have all
 // opque types such as images work as an allocation
-using DeviceAllocationId = uint32_t;
+using DeviceAllocationId = uint64_t;
 
 struct TI_DLL_EXPORT DeviceAllocation {
   Device *device{nullptr};
--- a/taichi/rhi/impl_support.h
+++ b/taichi/rhi/impl_support.h
@@ -3,6 +3,7 @@
 #include "taichi/rhi/device.h"
 #include <assert.h>
 #include <forward_list>
+#include <unordered_set>
 #include <mutex>
 
 namespace taichi::lang {
--- a/taichi/rhi/impl_support.h
+++ b/taichi/rhi/impl_support.h
@@ -92,37 +92,59 @@ struct BidirMap {
 };
 
 // A synchronized list of objects that is pointer stable & reuse objects
-// It does not mark objects as used, and it does not free objects (destructor is
-// not called)
 template <class T>
 class SyncedPtrStableObjectList {
+  using storage_block = std::array<uint8_t, sizeof(T)>;
+  
  public:
-  T &acquire() {
+  template <typename... Params>
+  T &acquire(Params &&...args) {
     std::lock_guard<std::mutex> _(lock_);
+    
+    void *storage = nullptr;
     if (free_nodes_.empty()) {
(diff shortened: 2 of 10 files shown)


taichi/rhi/impl_support.h
```diff
@@ -89,5 +91,37 @@ struct BidirMap {
   }
 };
 
+// A synchronized list of objects that is pointer stable & reuse objects
+// It does not mark objects as used, and it does not free objects (destructor is
+// not called)
+template <class T>
+struct SyncedPtrStableObjectList {
+  std::mutex lock;
```
k-ye, line 99: nit: make these members private and put them to the end?

taichi/rhi/device.h
```diff
@@ -17,6 +17,7 @@ enum class RhiResult {
   error = -1,
   invalid_usage = -2,
   not_supported = -3,
+  out_of_memory = -4,
```
PENGUINLIONG, line 20: 🤩

taichi/rhi/vulkan/vulkan_device.cpp
```diff
@@ -827,8 +858,11 @@ void VulkanCommandList::bind_pipeline(Pipeline *p) {
     return;
 
   if (pipeline->is_graphics()) {
```
PENGUINLIONG, line 860: i wonder if we will have an way to opt out graphics support in the distributed runtime library in the future?

taichi/rhi/impl_support.h
```diff
@@ -89,5 +91,39 @@ struct BidirMap {
   }
 };
 
+// A synchronized list of objects that is pointer stable & reuse objects
+// It does not mark objects as used, and it does not free objects (destructor is
+// not called)
+template <class T>
+class SyncedPtrStableObjectList {
```
PENGUINLIONG, line 98: so to my understanding the the contained objects should be explicitly constructed/destructed?

taichi/rhi/vulkan/vulkan_device.cpp
```diff
@@ -1526,21 +1671,7 @@ DeviceAllocation VulkanDevice::allocate_memory(const AllocParams &params) {
...
-  assert(map_pair != allocations_.end() && "Invalid memory handle");
-  return map_pair->second;
-}
-
-const VulkanDevice::AllocationInternal &VulkanDevice::get_alloc_internal(
-    const DeviceAllocation &alloc) const {
-  const auto &map_pair = allocations_.find(alloc.alloc_id);
-  assert(map_pair != allocations_.cend() && "Invalid memory handle");
-  return map_pair->second;
+  return DeviceAllocation{this, (uint64_t)&alloc};
```
PENGUINLIONG, line 1674: finally!

PENGUINLIONG (approved):
lgtm! thank you

taichi/rhi/impl_support.h
k-ye: That's a good point. It seems like we should add a check that `T` is a POD-ish type, which doesn't have fancy ctor/dtor. (Dtor is especially important here, IMO. Right now the implementation would require the caller to call `~T()`)

bobcao3: Fixed it now, the the object list now handles ctor/dtor & uses RAII.

taichi/rhi/vulkan/vulkan_device.cpp
bobcao3: Should be doable. (really doubt there will be significant size difference tho)

taichi/rhi/impl_support.h
k-ye: nit: I suggest that we add some unit test on this (https://github.com/taichi-dev/taichi/tree/master/tests/cpp), given that this is a very low-level infra, and has certain behaviors that is not trivial (manually calling `~T()`, placement new, ...)

PR #6916, opened 2022-12-17
[doc] Link to LLVM 15 built for Visual Studio 2022

PENGUINLIONG:
Built LLVM 15 with Visual Studio 2022. Note that this build has the experimental DXIL target enabled so Taichi's DX12 backend should work. Not tested tho.


ailzhang (approved):
Thanks!

PR #6919, opened 2022-12-18
[misc] Show suggestion when locking metadata.lock fails

PGZXB:
Issue: #4401

### Brief Summary
I found that some users' programs warn the `"[W ... Lock C:/taichi_cache/ticache/llvm/metadata.lock failed"`.
I think the crash when loading old version cache files which were generated by old version taichi-nightly could cause the warning. But I'm not sure.

Temporarily, show a message to suggest the user remove this `.lock` file manually.
And I will provide `python -m taichi clean_ticache` to allow user to clean the cache files manually.


ailzhang (approved):
LGTM!

PR #6920, opened 2022-12-18
[bug] Fixed running taichi-example-tests on single thread fails

PGZXB:
Fixed bug reported by https://github.com/taichi-dev/taichi/actions/runs/3715416438/jobs/6300563453
Repro: `python tests/run_tests.py -v -t1 -m "run_in_serial" -a cpu --with-offline-cache`

### DIFF
--- a/tests/python/examples/rendering/test_cornell_box.py
+++ b/tests/python/examples/rendering/test_cornell_box.py
@@ -1,9 +1,10 @@
 import argparse
 import os
-import pytest
-import taichi as ti
-
 from importlib import reload
+
+import pytest
+
+import taichi as ti
 
 FRAMES = 200
 
--- a/tests/python/examples/rendering/test_cornell_box.py
+++ b/tests/python/examples/rendering/test_cornell_box.py
@@ -1,9 +1,9 @@
 import argparse
 import os
-
 import pytest
-
 import taichi as ti
+
+from importlib import reload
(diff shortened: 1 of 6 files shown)


tests/python/examples/rendering/test_cornell_box.py
```diff
@@ -11,19 +12,22 @@
 @pytest.mark.run_in_serial
 @pytest.mark.skipif(os.environ.get('TI_LITE_TEST') or '0', reason='Lite test')
 def test_cornell_box():
-    from taichi.examples.rendering.cornell_box import render, tonemap
+    import taichi.examples.rendering.cornell_box as cornell_box
+    reload(cornell_box)
```
ailzhang, line 16: would you mind explaining why reloading is required?

PGZXB: The tests to test taichi examples are not similar with others. They import the example as module to init the taichi, register the global fields and get the functions to be called in test body. Running them one time is ok, but running pytest with  --with-offline-cache option will run them many times. Next time, they are run repeatedly, but the module will not be imported again so the initialization will not be done properly. `reload` is to ensure that the example module is reimport.
And note that because of the implementation of pytest, run them on multi threads is ok.

PGZXB:
These tests can be removed.

PR #6923, opened 2022-12-19
[lang] Support 'len' with Matrix-typed operands

### LINKED ISSUE #6915: `len()` cannot be applied to `ti.Vector()` types with `real_matrix` enabled
This no longer works starting from #6801 
```py
@ti.kernel
def k():
    v = ti.Vector([1,2,3])
    print(ti.static(len(v)))

k()
```
The error thrown is 
```
  File "c:\users\ldfra\code\taichi\taichi\python\taichi\lang\ast\ast_transformer_utils.py", line 25, in __call__
    return method(ctx, node)
  File "c:\users\ldfra\code\taichi\taichi\python\taichi\lang\ast\ast_transformer.py", line 542, in build_Call
    node.ptr = func(*args, **keywords)
  File "c:\users\ldfra\code\taichi\taichi\python\taichi\lang\impl.py", line 1062, in static
    raise ValueError(
ValueError: Input to ti.static must be compile-time constants or global pointers, instead of <class 'taichi.lang.expr.Expr'>
```
I prefer `len()` over `v.n`, because `len()` can also be applied to tuple types. So I could write
```py
@ti.func
def f(v):
    len(v) ..
```
and this would work for both `ti.Vector()`s and lists/tuples.

The workaround I used before #6885 is to do `ti.init(real_matrix = False)`, but this no longer works after the switch is removed.

jim19930609:
Issue: fix #6915


AmesingFlank (approved):
many thanks!

tests/python/test_matrix.py
```diff
@@ -1181,3 +1181,16 @@ def foo() -> ti.i32:
         return sig.sum() + p.sum()
 
     assert foo() == 4
+
+
+@test_utils.test(debug=True)
```
ailzhang, line 1186: OOC: why do we need debug=True for this test btw?

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -427,6 +427,15 @@ def build_call_if_is_builtin(ctx, node, args, keywords):
             id(pow): pow,
             id(operator.matmul): matrix_ops.matmul,
         }
+        if id(func) == id(len):
+            if len(args) != 1:
```
ailzhang, line 431: nit: add a test which triggers this error?

ailzhang (approved):
LGTM!

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -427,6 +427,15 @@ def build_call_if_is_builtin(ctx, node, args, keywords):
...
+                raise TaichiSyntaxError("len() takes exactly one argument.")
+            if isinstance(args[0], Expr) and args[0].ptr.is_tensor():
+                node.ptr = args[0].get_shape()[0]
+            else:
+                node.ptr = len(*args, **keywords)
```
strongoier, line 436: What about only handling the special case (one argument with `TensorType`) and let other cases fall into the default `build_Call`? Then we don't have to report the `SyntaxError` above either.

tests/python/test_matrix.py
```diff
@@ -1181,3 +1181,16 @@ def foo() -> ti.i32:
...
+def test_matrix_len():
+    @ti.kernel
+    def test():
+        x = ti.Vector([1, 0])
+        y = ti.Matrix([[1, 1, 1], [2, 2, 2], [3, 3, 3]])
+
+        assert ti.static(len(x)) == 2
```
strongoier, line 1193: The modification in this PR doesn't require `ti.static` here.

tests/python/test_matrix.py
strongoier: For in-kernel assertions :-)

strongoier (approved):

Issue #6807, opened 2022-12-05
Enable vulkan validation layer in CI


ailzhang:
We'll need to fix existing errors and enable crashing on error in the end.

PR #6925, opened 2022-12-19
[aot] Fix validation warning: OpImageFetch should operate on OpImage instead of OpSampledImage

jim19930609:
Issue: https://github.com/taichi-dev/taichi/issues/6807

### Brief Summary
`spv::OpImageFetch` should take `OpImage` instead of `OpSampledImage` as operand. Therefore we extract the underlying `OpImage` during code generation for `TextureOpStmt("fetch")`

### DIFF
--- a/taichi/codegen/spirv/spirv_ir_builder.cpp
+++ b/taichi/codegen/spirv/spirv_ir_builder.cpp
@@ -392,6 +392,14 @@ SType IRBuilder::get_pointer_type(const SType &value_type,
   return t;
 }
 
+SType IRBuilder::get_underlying_image_type(const SType &primitive_type,
+                                           int num_dimensions) {
+  auto key = std::make_pair(primitive_type.id, num_dimensions);
+  TI_ASSERT(sampled_image_underlying_image_type_.find(key) !=
+            sampled_image_underlying_image_type_.end());
+  return sampled_image_underlying_image_type_[key];
+}
+
 SType IRBuilder::get_sampled_image_type(const SType &primitive_type,
                                         int num_dimensions) {
   auto key = std::make_pair(primitive_type.id, num_dimensions);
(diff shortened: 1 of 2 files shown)


taichi/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -392,6 +392,14 @@ SType IRBuilder::get_pointer_type(const SType &value_type,
   return t;
 }
 
+SType IRBuilder::get_underlying_image_type(const SType &primitive_type,
+                                           int num_dimensions) {
```
bobcao3, line 396: I would like to instead of getting a value from a map, have this function construct the image type. Essentially moving this code to here:
```c++
int img_id = id_counter_++;
  spv::Dim dim;
  if (num_dimensions == 1) {
    dim = spv::Dim1D;
  } else if (num_dimensions == 2) {
    dim = spv::Dim2D;
  } else if (num_dimensions == 3) {
    dim = spv::Dim3D;
  } else {
    TI_ERROR("Unsupported number of dimensions: {}", num_dimensions);
  }
  ib_.begin(spv::OpTypeImage)
      .add_seq(img_id, primitive_type, dim,
               /*Depth=*/0, /*Arrayed=*/0, /*MS=*/0, /*Sampled=*/1,
               spv::ImageFormatUnknown)
      .commit(&global_);
```

taichi/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -415,13 +423,20 @@ SType IRBuilder::get_sampled_image_type(const SType &primitive_type,
                /*Depth=*/0, /*Arrayed=*/0, /*MS=*/0, /*Sampled=*/1,
                spv::ImageFormatUnknown)
       .commit(&global_);
+
   SType sampled_t;
   sampled_t.id = id_counter_++;
   sampled_t.flag = TypeKind::kImage;
   ib_.begin(spv::OpTypeSampledImage)
       .add_seq(sampled_t, img_id)
```
bobcao3, line 431: Then here, we can get the image type value by calling `get_underlying_image_type`.

In a IR builder API, when we call a function `get_xxx_type` we expect to get a value that holds the type, and this value can be cached from a previous construction. But the gist is calling this function shouldn't fail and should always get us the correct type.

bobcao3 (approved):
LGTM

PR #6926, opened 2022-12-19
[opt] Add pass eliminate_immutable_local_vars

strongoier:
Issue: #6933

### Brief Summary

There are many redundant copies of local vars in the initial IR:
```
  <[Tensor (3, 3) f32]> $128 = [$103, $106, $109, $112, $115, $118, $121, $124, $127]
  $129 : local store [$100 <- $128]
  <[Tensor (3, 3) f32]> $130 = alloca
  $131 = local load [$100]
  $132 : local store [$130 <- $131]
  <[Tensor (3, 3) f32]> $133 = alloca
  $134 = local load [$130]
  $135 : local store [$133 <- $134]
  <[Tensor (3, 3) f32]> $136 = alloca
  $137 = local load [$133]
  $138 : local store [$136 <- $137]
// In fact, `$128` can be used wherever `$136` is loaded.
```

These can come from many places; one of the main sources is the pass-by-value convention of `ti.func`. The consequence is that the number of instructions is unnecessarily large, which significantly slows down compilation.

My solution here is to identify and eliminate such redundant instructions in the first place so all later passes can take a much smaller number of instructions as input. These redundant local vars are essentially immutable ones - they are assigned only once and only loaded after the assignment. In this PR, I add an optimization pass `eliminate_immutable_local_vars` as the first pass.

(P.S. The type check processes of `MatrixExpression` and `LocalLoadStmt` are fixed by the way to make the pass work properly.)

Let's study the effects in two cases: #6933 and [voxel-rt2](https://github.com/taichi-dev/voxel-rt2/blob/main/example7.py).

First, let's compare the number of instructions after `scalarization` pass (which happens immediately after the first pass).

| Kernel | Before this PR | After this PR | Rate of decrease |
| ------ | ------ | ------ | ------ |
| `test` (#6933) | 45859  | 26452 | 42% |
| `spatial_GRIS` (voxel-rt2) | 48519 | 17713 | 63% |

Then, let's compare the total time of `compile()`.

| Case | Before this PR | After this PR | Rate of decrease |
| ------ | ------ | ------ | ------ |
| #6933 | 20.622s | 8.550s | 59% |
| voxel-rt2  | 27.676s  | 9.495s | 66% |


taichi/analysis/gather_immutable_local_vars.cpp
```diff
@@ -0,0 +1,92 @@
+#include "taichi/ir/ir.h"
+#include "taichi/ir/statements.h"
+#include "taichi/ir/transforms.h"
+#include "taichi/ir/visitors.h"
+
+namespace taichi::lang {
+
+// The GatherImmutableLocalVars pass gathers all immutable local vars as input
+// to the EliminateImmutableLocalVars pass. An immutable local var is an alloca
+// which is stored only once (in the same block) and only loaded after that
```
bobcao3, line 10: Speaking of this, since the most amount of time spent was on `store_to_load_forwarding` in CFG, we can make a smaller & simpler specialized `store_to_load_forwarding` that only processes alloca load stores in the same block

bobcao3 (approved):
LGTM

taichi/analysis/gather_immutable_local_vars.cpp
```diff
@@ -0,0 +1,92 @@
...
+// store.
+class GatherImmutableLocalVars : public BasicStmtVisitor {
+ private:
+  using BasicStmtVisitor::visit;
+
+  enum class AllocaStatus { kCreated = 0, kStoredOnce = 1, kInvalid = 2 };
+  std::unordered_map<Stmt *, AllocaStatus> alloca_status_;
```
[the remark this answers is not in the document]
strongoier: Ah `+= 2` is also a way to mark the state as invalid. Well it's just an implementation choice :-)

taichi/analysis/gather_immutable_local_vars.cpp
strongoier: Yep I considered that as well and may try that in the future. The current one is chosen first because its cost-benefit is surprisingly high. Meanwhile, it's at least a global optimization so it sounds not that hacky :-)

PR #6934, opened 2022-12-20
[lang] Migrate TensorType expansion for SNode indices from Python to Frontend IR

jim19930609:


taichi/ir/frontend_ir.cpp
```diff
@@ -923,6 +928,25 @@ void AtomicOpExpression::flatten(FlattenContext *ctx) {
...
+  this->indices.exprs = std::move(expanded_indices);
+}
+
+SNodeOpExpression::SNodeOpExpression(ASTBuilder *builder,
+                                     SNode *snode,
+                                     SNodeOpType op_type,
+                                     const ExprGroup &indices,
+                                     const std::vector<Expr> &values)
+    : SNodeOpExpression(builder, snode, op_type, indices) {
+  this->values = builder->expand_expr(values);
```
strongoier, line 947: Do we need to expand `values`?

jim19930609: Yup, because we supported appending Struct and Matrix to Dynamic SNode: https://github.com/taichi-dev/taichi/pull/6535/files. 

So we have to scalarize `val` either during SNodeOpExpression construction or at `irpass::scalarize()`, and I'd prefer that happens earlier than later.

strongoier: Then maybe we need to cancel the scalarization in Python https://github.com/taichi-dev/taichi/blob/be153583417523096448beaaf50cc7a176f808cd/python/taichi/lang/snode.py#L426
to test this part?

jim19930609: Looks like `_get_flattened_ptrs()` is also used by `build_Return`, and I do have a separate PR to remove that: https://github.com/taichi-dev/taichi/pull/6946. 

To make this PR focus on SNodes-related changes, how about let's keep `_get_flattened_ptrs()` as is for now and leave the cleanup to #6946?

strongoier: Sure!

strongoier (approved):

PR #6939, opened 2022-12-20
Fixed C-API behavior tests

PENGUINLIONG:
Fixed allocation leakage, unchecked values, camel cases and uninitialized structs in behavior tests.


c_api/include/taichi/cpp/taichi.hpp
```diff
@@ -663,60 +685,6 @@ class AotModule {
   }
 };
 
-class Event {
```
jim19930609, line 666: Are we intended to deprecate the C++ wrapper for `TiEvent` or is it by mistake?

PENGUINLIONG: yes, it was introduced as a hack for unity integration but we no longer need it now. but anyway, let's defer it to another or

jim19930609 (approved):
thanks~

PR #6942, opened 2022-12-21
[lang] Migrate TensorType expansion for subscription indices from Python to Frontend IR

jim19930609:

### Brief Summary
For indices of TensorType, instead of scalarizing them at Python level, it is up to the Frontend IR's consumer to decide whether TensorType'd indices are acceptable and if we should have it scalarized.

This PR removes `expand_expr` in Expression subscription and migrate the scalarization logics to the following constructors: 

1. MeshIndexConversionExpression::MeshIndexConversionExpression
2. IndexExpression::IndexExpression


taichi/ir/frontend_ir.cpp
```diff
@@ -1477,63 +1603,72 @@ Expr ASTBuilder::snode_get_addr(SNode *snode, const ExprGroup &indices) {
 }
 
 std::vector<Expr> ASTBuilder::expand_expr(const std::vector<Expr> &exprs) {
```
strongoier, line 1605: Now more than one expr is expanded. Shall we rename it to `expand_exprs`?

taichi/ir/frontend_ir.cpp
```diff
@@ -688,6 +688,39 @@ void MatrixExpression::flatten(FlattenContext *ctx) {
...
+IndexExpression::IndexExpression(ASTBuilder *builder,
+                                 const Expr &var,
+                                 const ExprGroup &indices,
+                                 std::string tb)
+    : var(var) {
+  // IndexExpression without ret_shape is used for matrix indexing,
+  // where each entry of ExprGroup is interpreted as indexing into a specific
+  // axis. For example, mat[0][3, 4] has indices_group={0, [3, 4]}, where [3, 4]
+  // corresponds to "n"-axis and "m"-axis of the matrix. Therefore we expand
+  // indices_group={0, [3, 4]} into {0, 3, 4} to avoid TensorType in indices.
```
strongoier, line 700: The comment confuses me a bit. Why does a matrix need three indices `0, 3, 4`?

python/taichi/lang/impl.py
```diff
@@ -204,28 +202,28 @@ def subscript(ast_builder, value, *_indices, skip_reordered=False):
                     f"Gradient {_var.get_expr_name()} has not been placed, check whether `needs_grad=True`"
                 )
         field_dim = snode.num_active_indices()
```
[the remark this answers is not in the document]
jim19930609: yep, migrated the corresponding validation checks to Frontend IR

taichi/ir/frontend_ir.cpp
jim19930609: Ah bad example, let me change it to {3, 4}

python/taichi/lang/impl.py
```diff
@@ -204,28 +202,28 @@ def subscript(ast_builder, value, *_indices, skip_reordered=False):
...
-        if field_dim != index_dim:
-            raise TaichiIndexError(
-                f'Field with dim {field_dim} accessed with indices of dim {index_dim}'
-            )
+
         if isinstance(value, MatrixField):
-            return make_index_expr(value.ptr, indices_expr_group)
+            return Expr(
```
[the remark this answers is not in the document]
jim19930609: Personally I'd prefer a unified interface to create `Expressions`, so maybe we should remove `make_index_expr` and turn to use `ast_builder.expr_subscript()`. WDYT?

taichi/ir/frontend_ir.cpp
```diff
@@ -1160,6 +1263,20 @@ void MeshRelationAccessExpression::flatten(FlattenContext *ctx) {
   stmt = ctx->back_stmt();
 }
 
+MeshIndexConversionExpression::MeshIndexConversionExpression(
```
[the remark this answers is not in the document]
jim19930609: Looks like so. Not familiar with the MeshInstance implementation, shall I add an assertion here instead?
 
image

python/taichi/lang/impl.py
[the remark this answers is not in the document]
strongoier: SG. Let's remove `make_index_expr` then.

taichi/ir/frontend_ir.cpp
```diff
@@ -721,28 +741,98 @@ bool IndexExpression::is_global() const {
...
+
 void IndexExpression::type_check(CompileConfig *) {
   // TODO: Change to type-based solution
   // Currently, dimension compatibility check happens in Python
   TI_ASSERT(indices_group.size() == std::accumulate(begin(ret_shape),
                                                     end(ret_shape), 1,
                                                     std::multiplies<>()));
+  int index_dim = indices_group.empty() ? 0 : indices_group[0].size();
+  bool has_slice = !ret_shape.empty();
   if (!ret_shape.empty()) {
```
strongoier, line 809: The `has_slice` case has been checked here. So maybe we don't need the `has_slice` variable?

taichi/ir/frontend_ir.cpp
```diff
@@ -721,28 +741,98 @@ bool IndexExpression::is_global() const {
   return is_field() || is_matrix_field() || is_ndarray();
 }
 
+static void field_validation(FieldExpression *field_expr,
+                             int index_dim,
+                             bool has_slice) {
+  TI_ASSERT(field_expr != nullptr);
+  if (field_expr->snode == nullptr)
```
strongoier, line 748: `field_expr->snode` should not be `nullptr`.

taichi/ir/frontend_ir.cpp
```diff
@@ -721,28 +741,98 @@ bool IndexExpression::is_global() const {
...
+    return;
+  int field_dim = field_expr->snode->num_active_indices;
+
+  int has_dynamic = 0;
+  if (field_expr->snode->parent &&
+      field_expr->snode->parent->type == SNodeType::dynamic) {
+    has_dynamic = 1;
+  }
+  /*
+    Indexing to dynamic snode can be special, you can either index all the way
```
strongoier, line 758: `dynamic` SNode is already handled in a special way: https://github.com/taichi-dev/taichi/blob/59d8909112a924125ca6a13ceaa05d2219a359c0/python/taichi/lang/ast/ast_transformer.py#L800-L819

So maybe we don't need to handle this again here? BTW we need to raise `TaichiIndexError` instead of `TaichiTypeError`.

taichi/ir/frontend_ir.cpp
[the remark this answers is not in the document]
strongoier: Sounds a bit weird. But for now let's just leave it as it is.

taichi/ir/frontend_ir.cpp
jim19930609: Personally I'd prefer to have that checker in `IndexExpression::type_check()` as well, as an ultimate validation for the legality of an `IndexExpression` in Frontend IR semantics. 

On the other hand, the checker at Python level can trigger an early stop to provide a more accurate error message, but it could get bypassed since this may not be the only interface to create an `IndexExpression`

strongoier: Ah, the main point here is that `a[i, j].append(x)` is currently a syntax sugar for `ti.append(a.parent(), (i, j), x)` in Python, and `a[i, j]` itself is not valid. So IMO we don't need to (or shouldn't?) allow it in frontend IR.

jim19930609: I see, I can remove this check in IndexExpression. But honestly this "syntax sugar" allows the use of intermediate(non-place) SNode and does hacks IndexExpression. If other frontend languages (say Rust) doesn't have the same "syntax sugar" then there will be "Python-specifc hacks" at Frontend IR level.

strongoier: I agree. I guess the root problem here is that SNodes are not in the type system yet, so handling these intermediate SNodes always needs hacks somewhere. Once we have SNodes in our type system, we can precisely describe rules like `a[i, j]` is allowed to have `append` attribute.

jim19930609: Btw, there's no `TaichiIndexError ` at C++ level yet. Let's comply to `TaichiSyntaxError` then?

image

strongoier: Hmm I mention `TaichiIndexError` because it seems that the dynamic SNode hack relies on it: https://github.com/taichi-dev/taichi/blob/f4e3aa41e050408974bc6dac9bebe1c3ebc57d5d/python/taichi/lang/ast/ast_transformer.py#L822

taichi/ir/frontend_ir.cpp
```diff
@@ -721,28 +741,70 @@ bool IndexExpression::is_global() const {
...
+                    field_dim, index_dim));
+  }
+}
+
+static void matrix_field_validation(MatrixFieldExpression *matrix_field_expr,
+                                    int index_dim) {
+  TI_ASSERT(matrix_field_expr != nullptr);
+  TI_ASSERT(!matrix_field_expr->fields.empty());
+  auto field_expr = matrix_field_expr->fields[0].cast<FieldExpression>();
+  int element_dim = matrix_field_expr->element_shape.size();
```
strongoier, line 761: For matrix field there's no need to consider `element_dim`.

jim19930609: I was thinking we have to handle the following two cases, one of which has element_dim involved?

```
x = ti.Vector.field(n=2, shape=(10, 20, 30))

@ti.kernel
def test():
  x[0, 1, 2][0] = 1   # outter_dim + element_dim == index_dim
  x[0, 1, 2] = [10, 20] # outter_dim == index_dim
```

strongoier: Here `x[0, 1, 2][0]` is always becoming `IndexExpr(IndexExpr(x, [0, 1, 2]), [0])`.

jim19930609: Ah that's right~

strongoier (approved):

Issue #6884, opened 2022-12-13
Initialising nested `@ti.dataclass` throws error when argument name not explicitly given


AmesingFlank:
Consider this code snippet
```py
import taichi as ti
ti.init(ti.cpu)


@ti.dataclass
class X:
    i: int


@ti.dataclass
class Y:
    x: X


x = X(0)
y = Y(x)
```
it fails with
```
  File "test.py", line 17, in <module>
    y = Y(x)
  File "/Users/dunfanlu/Code/Taichi/taichi/python/taichi/lang/struct.py", line 677, in __call__
    d[name] = dtype(args[index])
  File "/Users/dunfanlu/Code/Taichi/taichi/python/taichi/lang/struct.py", line 687, in __call__
    struct = self.cast(entries)
  File "/Users/dunfanlu/Code/Taichi/taichi/python/taichi/lang/struct.py", line 717, in cast
    entries[k] = int(
TypeError: int() argument must be a string, a bytes-like object or a number, not 'Struct'
```
But if the final line `y = Y(x)` is replaced with `y = Y(x=x)`, then it works.

PR #6949, opened 2022-12-21
[Lang] Fix struct type problem

neozhaoliang:
This PR close: #6884


python/taichi/lang/struct.py
```diff
@@ -672,20 +672,22 @@ def __init__(self, **kwargs):
...
-                    d[name] = kwargs.get(name, 0)
+            name, dtype = pair  # (member name, member type)
+            if index < len(args):  # set from args
+                data = args[index]
+            else:  # set from kwargs
+                data = kwargs.get(name, 0)
+
+            if isinstance(dtype, CompoundType) and not isinstance(
+                    data, (dict, Struct)):
+                data = dtype(data)
```
strongoier, line 688: Why do we need `dtype(data)` here? There's already `struct = self.cast(entries)` below.

strongoier (commented):
Shall we add a test?

python/taichi/lang/struct.py
neozhaoliang: This is because when `dtype` is a struct type and `data` is a scalar, `self.cast(entries)` will raise an error.

neozhaoliang: This is because if there is a struct member that leaves uninitialized, `self.cast` will try to cast the int `0` to it, hence raising an error. For example:

```python
import taichi as ti
ti.init(ti.cpu)

vec2 = ti.math.vec2

D = dict({"i": 3, "k":2.})
@ti.dataclass
class X:
    i: int

@ti.dataclass
class Z:
    i: int
    x: X

x = X(1)
z = Z(1)
print(z.x)
```

I have  updated the code so that the intention should be more clear.

strongoier (approved):

Issue #6832, opened 2022-12-07
[RFC] Roadmap to a smaller, faster, and cleaner Unified Device API


bobcao3:
The long term goal for the Unified Device API, or the RHI, is to be isolated into its own repo and maintained separately from Taichi. Where we will create its own unit tests and integration tests to improve robustness and maintain strong isolation boundary between them.

In the end, we want to achieve these goals for the future RHI:
1. RHI can be trimmed down through discarding unused modules. We should make RHI per-backend dynamically linked and optionally statically linked.
2. RHI contains its own full functional specification and tests that uphold these specifications
3. Provide a zero-overhead C-API
4. Do not leak symbols to linked libraries & support different drivers and external dependencies robustly
5. Able to use shared library if the client of RHI already uses them
6. Being a fully functioned RHI that we can build other non-taichi applications on top of it
7. Make it smaller in size and complexity

To achieve this goal while not hugely breaking the workflow and goals within Taichi, we will start by progressively trimming away dependencies of RHI. This hopefully will eventually lead to a clean separation when we move RHI outside of Taichi repo. Here are the first steps:
- [ ] Remove all Taichi IR and related dependencies (`ir.h`, `program.h`, LLVM, eigen, etc.)
- [ ] Remove / avoid `TI_XXX` and `spdlog` based logging when required (use MACRO, log less, and use what's provided by standard library)
- [ ] Remove exceptions (use `noexcept`) and use clear status return values. (This reduces generated code size and helps with C-API impl)
- [ ] Re-organize the API to make most-of-the-time optional features being binary-wise optional. (e.g. Device/Context creation, Window/Surface creation, etc.)
- [ ] Re-organize the binding model to help with integration into 3rd party code bases

bobcao3:
DX11 Impl tracker: #6879
Vulkan Impl tracker: #6880
OpenGL Impl tracker: #6881 
Metal Impl tracker: #6882

PR #6954, opened 2022-12-21
[gfx] Update Device API: Splitting ResourceBinder into seperate ShaderResourceSet & RasterResources

bobcao3:
Issue: #6832

### Brief Summary

ResourceBinder is not split into two structures, one controls binding of shared-accessible resources, the other is dedicated to control binding of rasterizer states (vertex buffers, etc.)

This makes the mapping onto Vulkan DescriptorSets easier, and cleans up the implementation of resource binding all over the place. In addition, these binding states are now no longer attached to the program and can be thus pre-filled / pre-allocated to achieve lower overhead.


taichi/rhi/dx/dx_device.h
```diff
@@ -174,6 +175,15 @@ class Dx11Device : public GraphicsDevice {
 
   DeviceAllocation allocate_memory(const AllocParams &params) override;
   void dealloc_memory(DeviceAllocation handle) override;
+
+  ShaderResourceSet *create_resource_set() final {
```
k-ye, line 179: Why raw pointer instead of `unique_ptr`?

k-ye (approved):
LGTM!

taichi/rhi/dx/dx_device.h
bobcao3: It is STL, so there is a wrapper function that makes a unique one in the device.h headers (so that we should be able to link to the library from a different compiler, while still having unique pointer access)

bobcao3:
https://github.com/taichi-dev/taichi-aot-demo/pull/105

PR #6960, opened 2022-12-22
[aot] Fix rwtexture with template_args

ailzhang:
Currently we take rwtexture as non-template arg so we need to properly initialize it from its annotation for aot.

Issue: #


PENGUINLIONG (approved):
lgtm! thank you!

PR #6479, opened 2022-10-31
[ci] Package C-API in nightly builds

feisuzhu:
Issue: #

PR #6962, opened 2022-12-22
[ci] Fix missing c_api.so in linux nightly

ailzhang:
Previous hack to add enable capi build in unix-build.sh (#6479) is gone so this PR tries to add it back.

Issue: #


PENGUINLIONG (approved):
thx!

PR #6964, opened 2022-12-22
[bug] Allow StructType as type hint to ti.func

### LINKED ISSUE #6855: Struct types cannot be used with type hints as arguments to `@ti.func`
The following code works:
```py
import taichi as ti
ti.init(ti.cpu)

@ti.dataclass
class C:
    i: int

@ti.func
def f(c):
    return c.i

@ti.kernel
def k():
    c = C(i = 2)
    print(f(c))

k()
```
while this doesn't:
```py
import taichi as ti
ti.init(ti.cpu)

@ti.dataclass
class C:
    i: int

@ti.func
def f(c: C):
    return c.i

@ti.kernel
def k():
    c = C(i = 2)
    print(f(c))

k()
```
The only difference is the type hint: `c: C`
When the type hint is added, the following error is thrown:
```
  File "/Users/dunfanlu/Code/Taichi/taichi/python/taichi/lang/kernel_impl.py", line 341, in extract_arguments
    raise TaichiSyntaxError(
taichi.lang.exception.TaichiSyntaxError: Invalid type annotation (argument 0) of Taichi function: <taichi.lang.struct.StructType object at 0x10875e220>
```

strongoier:
Issue: fix #6855

### Brief Summary

Note that this PR is a quick hack to allow the type hint but doesn't check invalid cases. We would like to wait for the upcoming new struct type representation to officially support struct type as kernel/func parameters.

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -334,6 +334,8 @@ class Func:
                     pass
                 elif isinstance(annotation, MatrixType):
                     pass
+                elif isinstance(annotation, StructType):
+                    pass
                 elif id(annotation) in primitive_types.type_ids:
                     pass
                 elif isinstance(annotation, template):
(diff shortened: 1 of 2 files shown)


ailzhang (approved):
Thanks!

PR #6967, opened 2022-12-23
[Lang] Support  LU sparse solver on CUDA backend

FantasyVR:
Issue: #2906 

### Brief Summary
To be consistent with API on CPU backend, this pr provides LU sparse solver on CUDA backend. CuSolver just provides a CPU version API of LU sparse solver which is used in this PR. The cuSolverRF provides a GPU version LU solve, but it only supports `double` datatype. Thus, it's not used in this PR.

Besides, the `print_triplets` is refactored to resolve the ndarray `read` constraints (the `read` and `write` data should be the same datatype).


taichi/program/sparse_matrix.cpp
```diff
@@ -99,17 +99,53 @@ SparseMatrixBuilder::SparseMatrixBuilder(int rows,
...
+}
+
+void SparseMatrixBuilder::print_triplets_cuda() {
+#ifdef TI_WITH_CUDA
+  CUDADriver::get_instance().memcpy_device_to_host(
+      &num_triplets_, (void *)get_ndarray_data_ptr(), sizeof(int));
+  fmt::print("n={}, m={}, num_triplets={} (max={})\n", rows_, cols_,
+             num_triplets_, max_num_triplets_);
+  auto len = 3 * num_triplets_ + 1;
+  std::vector<float32> trips(len);
```
strongoier, line 138: Is only `float32` supported for CUDA sparse matrix?

strongoier (approved):

taichi/program/sparse_matrix.cpp
FantasyVR: Yeah, we only support `float32` right now.

PR #6974, opened 2022-12-23
Revert "[ci] Fix missing c_api.so in linux nightly"

ailzhang:
Reverts taichi-dev/taichi#6962
We don't want to add c_api.so to official releases *yet*, so #6969 is a better fix. :D


feisuzhu (approved):
LGTM

PR #6980, opened 2022-12-26
[lang] Migrate TensorType expansion for FuncCallExpression from Python code to Frontend IR

jim19930609:

### DIFF
--- a/taichi/ir/frontend_ir.cpp
+++ b/taichi/ir/frontend_ir.cpp
@@ -35,6 +35,13 @@ FrontendSNodeOpStmt::FrontendSNodeOpStmt(ASTBuilder *builder,
   }
 }
 
+FuncCallExpression::FuncCallExpression(ASTBuilder *builder,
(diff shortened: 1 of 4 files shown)


taichi/ir/frontend_ir.cpp
```diff
@@ -35,6 +35,13 @@ FrontendSNodeOpStmt::FrontendSNodeOpStmt(ASTBuilder *builder,
   }
 }
 
+FuncCallExpression::FuncCallExpression(ASTBuilder *builder,
+                                       Function *func,
+                                       const ExprGroup &args)
+    : func(func), args(args) {
+  this->args.exprs = builder->expand_expr(args.exprs);
```
lin-hitonami, line 42: Why not do this in `ASTBuilder::insert_func_call_expr` and then pass the processed `ExprGroup` into the construction function?

strongoier (approved):

PR #6982, opened 2022-12-26
[ir] Add struct type to CHI-IR

lin-hitonami:
Issue: #6983

### DIFF
--- a/taichi/ir/type.h
+++ b/taichi/ir/type.h
@@ -40,6 +40,14 @@ class TI_DLL_EXPORT Type {
     return p;
   }
 
+  template <typename T>
+  const T *as() const {
+    auto p = dynamic_cast<const T *>(this);
+    TI_ASSERT_INFO(p != nullptr, "Cannot treat {} as {}", this->to_string(),
+                   typeid(T).name());
+    return p;
+  }
+
   bool is_primitive(PrimitiveTypeID type) const;
 
   virtual Type *get_compute_type() {
@@ -56,7 +64,7 @@ class TI_DLL_EXPORT DataType {
   DataType();
 
   // NOLINTNEXTLINE(google-explicit-constructor)
-  DataType(Type *ptr) : ptr_(ptr) {
+  DataType(const Type *ptr) : ptr_((Type *)ptr) {
   }
 
   DataType(const DataType &o) : ptr_(o.ptr_) {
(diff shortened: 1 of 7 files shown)


taichi/ir/type_factory.h
```diff
@@ -57,6 +60,10 @@ class TypeFactory {
   // TODO: use unordered map
   std::map<std::pair<std::string, Type *>, std::unique_ptr<Type>> tensor_types_;
 
+  std::unordered_map<std::vector<const Type *>, std::unique_ptr<Type>>
```
bobcao3, line 63: Are we sure that `Type *` comparison is what we expect it to be? If not, I would be more comfortable in this case to have a custom hasher and comparison function

taichi/util/hash.h
```diff
@@ -0,0 +1,25 @@
...
+namespace std {
+
+namespace {
+inline void hash_combine(size_t &seed, size_t value) {
+  seed ^= value + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+}
+}  // namespace
+
+template <typename T>
+struct hash<std::vector<T>> {
```
bobcao3, line 15: This is quite dangerous, don't think we should be extending std namespace with a template that's as wide as this. We should move this into a custom hasher

taichi/ir/type_factory.h
lin-hitonami: Yes, `Type *` identifies a type, and two types are equal iff the `Type *`s are equal.

lin-hitonami: If we need to add debug info (name of the elements) to it, maybe we can change to `std::vector<std::pair<const Type *, std::string>>` later.

taichi/ir/type.h
```diff
@@ -193,6 +201,42 @@ class TensorType : public Type {
   Type *element_{nullptr};
 };
 
+class StructType : public Type {
+ public:
+  explicit StructType(std::vector<const Type *> elements)
+      : elements_(std::move(elements)) {
```
jim19930609, line 207: Do we allow PointerType and QuantType in a StructType? Otherwise let's add a checker here?

taichi/util/hash.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <functional>
+#include <stddef.h>
+
+namespace taichi::hashing {
+
+namespace {
+inline void hash_combine(size_t &seed, size_t value) {
+  seed ^= value + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```
jim19930609, line 10: Since we're implementing a custom hash algorithm, can you comment on how did you decide the math model and the parameters "0x9e3779b9", "<< 6" and ">> 2" ? 

Btw, do you also have a rough estimation of the collision rate?

jim19930609 (approved):

taichi/util/hash.h
lin-hitonami: It's borrowed from boost, I'll add an comment about the reference.

strongoier (approved):

ailzhang (approved):
lgtm!

taichi/ir/type.h
lin-hitonami: `PointerType` is allowed in a `StructType` for passing references. Though it is not yet tested, I think `QuantType` can be put into a struct naturally as well.

lin-hitonami: Emmm it seems that we don't support `QuantType` yet... I'll add a checker then...

PR #6985, opened 2022-12-26
[Lang] [metal] Support dynamic indexing in metal

strongoier:
Issue: #2590

### Brief Summary

1. Dynamic indexing in metal is supported by mimicking the implementation of that in LLVM backends.
2. `ret_type` change of `ArgLoadStmt` is moved out of the `ScalarizePointers` pass, which is designed to scalarize `alloca` and will be removed once dynamic indexing becomes default.
3. Configs of dynamic indexing tests are fixed.

PR #6990, opened 2022-12-27
[Lang] [spirv] Support dynamic indexing in spirv

strongoier:
Issue: #2590

### Brief Summary

This PR supports dynamic indexing in spirv (sister PR: #6985). `_test_local_matrix_non_constant_index()` is modified due to lack of `AssertStmt` support in spirv.

### DIFF
--- a/taichi/codegen/spirv/spirv_codegen.cpp
+++ b/taichi/codegen/spirv/spirv_codegen.cpp
@@ -248,10 +248,13 @@ class TaskCodegen : public IRVisitor {
     auto dt = stmt->element_type().ptr_removed();
     if (stmt->offset_used_as_index()) {
       if (stmt->origin->is<AllocaStmt>()) {
-        spirv::SType ptr_type = ir_->get_pointer_type(ir_->get_primitive_type(dt), origin_val.stype.storage_class);
-        ptr_val = ir_->make_value(spv::OpAccessChain, ptr_type, origin_val, offset_val);
+        spirv::SType ptr_type = ir_->get_pointer_type(
+            ir_->get_primitive_type(dt), origin_val.stype.storage_class);
+        ptr_val = ir_->make_value(spv::OpAccessChain, ptr_type, origin_val,
+                                  offset_val);
       } else if (stmt->origin->is<GlobalTemporaryStmt>()) {
-        spirv::Value dt_bytes = ir_->int_immediate_number(ir_->i32_type(), ir_->get_primitive_type_size(dt), false);
+        spirv::Value dt_bytes = ir_->int_immediate_number(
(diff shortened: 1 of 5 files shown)


taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -219,40 +219,53 @@ class TaskCodegen : public IRVisitor {
...
+      } else if (stmt->origin->is<GlobalTemporaryStmt>()) {
+        spirv::Value dt_bytes = ir_->int_immediate_number(
+            ir_->i32_type(), ir_->get_primitive_type_size(dt), false);
+        spirv::Value offset_bytes = ir_->mul(dt_bytes, offset_val);
+        ptr_val = ir_->add(origin_val, offset_bytes);
+        ptr_to_buffers_[stmt] = ptr_to_buffers_[stmt->origin];
+      } else {
+        TI_NOT_IMPLEMENTED;
+      }
+    } else {  // offset used as bytes
```
bobcao3, line 264: What is offset used as bytes?

taichi/program/extension.cpp
```diff
@@ -23,8 +23,9 @@ bool is_extension_supported(Arch arch, Extension ext) {
       {Arch::metal,
        {Extension::adstack, Extension::assertion, Extension::dynamic_index,
         Extension::sparse}},
-      {Arch::opengl, {Extension::extfunc}},
+      {Arch::opengl, {Extension::dynamic_index, Extension::extfunc}},
```
bobcao3, line 26: GFX backend also covers dx11

taichi/codegen/spirv/spirv_codegen.cpp
strongoier: It is for dynamically indexing a matrix element of a matrix field, which might have arbitrary layout. We identity cases where the layout is dynamically indexable, and turn those access into `base_pointer + offset`. See #3810 for details.

bobcao3: that sounds terrible... what if offset is not aligned to 4 bytes compared to the starting point of the buffer? Is there a guarantee somewhere, and can we turn this into a linear index?

strongoier: hmm I think the problem is not about the offset calculation, but about the struct compiler... Because the offset calculation directly takes `cell_stride` and `mem_offset_in_parent_cell` from the struct compiler, whether the offset is aligned depends on whether `mem_offset_in_parent_cell` is aligned. If we have this guarantee then we can certainly turn this into a linear index.

ailzhang (approved):
LGTM!

PR #6993, opened 2022-12-28
[ci] Switch Windows build script to build.py

feisuzhu:
Issue: #6445


.github/workflows/scripts/build.py
```diff
@@ -32,6 +33,30 @@ def setup_llvm(env_out: dict) -> None:
...
             lnsf('/usr/bin/ld.lld-10', '/usr/bin/ld.lld')
+    elif (u.system, u.machine) == ('Windows', 'AMD64'):
+        out = get_cache_home() / 'clang-15'
+        url = 'https://github.com/python3kgae/taichi_assets/releases/download/llvm15_vs2022_clang/clang-15.0.0-win.zip'
+        download_dep(url, out)
+        env_out['PATH'] = concat_paths(out / 'bin', env_out.get('PATH'))
+        env_out[
+            'TAICHI_CMAKE_ARGS'] += ' -DCLANG_EXECUTABLE=clang++.exe'  # TODO: Can this be omitted?
+        env_out[
+            'TAICHI_CMAKE_ARGS'] += ' -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang'
```
bobcao3, line 44: FYI, there's a msbuild+MSVC effort going on rn. #6905

.github/workflows/scripts/build.py
```diff
@@ -47,10 +72,15 @@ def setup_llvm(env_out: dict) -> None:
     elif (u.system, u.machine) == ('Darwin', 'x86_64'):
         out = get_cache_home() / 'llvm15-mac'
         url = 'https://github.com/taichi-dev/taichi_assets/releases/download/llvm15/llvm-15-mac10.15.zip'
+    elif (u.system, u.machine) == ('Windows', 'AMD64'):
+        out = get_cache_home() / 'llvm15-win'
+        url = 'https://github.com/python3kgae/taichi_assets/releases/download/llvm15_vs2019_clang/taichi-llvm-15.0.0-msvc2019.zip'
+        env_out['TAICHI_CMAKE_ARGS'] += " -DLLVM_AS_EXECUTABLE=llvm-as.exe"
```
bobcao3, line 78: I'm not sure whether we still use `llvm-as` anymore

feisuzhu: Nice, I'll try removing it!

.github/workflows/scripts/build.py
```diff
@@ -1,25 +1,75 @@
...
+def setup_clang(as_compiler=True) -> None:
+    '''
+    Setup Clang.
+    '''
+    u = platform.uname()
+    if u.system == 'Linux':
+        pass
+    elif (u.system, u.machine) == ('Windows', 'AMD64'):
+        out = get_cache_home() / 'clang-15'
+        url = 'https://github.com/python3kgae/taichi_assets/releases/download/llvm15_vs2022_clang/clang-15.0.0-win.zip'
```
bobcao3, line 29: Don't think we need the entire Clang? I think Visual Studio ships with a compatible version of Clang to our LLVM.

.github/workflows/scripts/build.py
```diff
@@ -1,25 +1,75 @@
...
+        download_dep(url, out)
+        path_prepend('PATH', out / 'bin')
```
bobcao3, line 31: Don't pollute path, and we actually don't need to do this at all.

.github/workflows/scripts/build.py
```diff
@@ -1,25 +1,75 @@
...
+        os.environ['TAICHI_CMAKE_ARGS'] += ' -DCLANG_EXECUTABLE=clang++.exe'
```
bobcao3, line 32: Instead point to full path here

.github/workflows/scripts/build.py
bobcao3: Though this depends on whether the user wants to use their own install of Visual Studio or just the build tools. I guess this is fine.

.github/workflows/scripts/build.py
```diff
@@ -1,25 +1,75 @@
...
+
+        if as_compiler:
+            os.environ['TAICHI_CMAKE_ARGS'] += (' -DCMAKE_CXX_COMPILER=clang++'
```
bobcao3, line 35: This flag might not do anything under MSBuild as it uses its own "Toolset" system

.github/workflows/scripts/build.py
```diff
@@ -37,54 +87,82 @@ def setup_llvm(env_out: dict) -> None:
...
         out = get_cache_home() / 'llvm15-m1'
         url = 'https://github.com/taichi-dev/taichi_assets/releases/download/llvm15/taichi-llvm-15-m1.zip'
+        download_dep(url, out, strip=1)
     elif (u.system, u.machine) == ('Darwin', 'x86_64'):
         out = get_cache_home() / 'llvm15-mac'
         url = 'https://github.com/taichi-dev/taichi_assets/releases/download/llvm15/llvm-15-mac10.15.zip'
+        download_dep(url, out, strip=1)
+    elif (u.system, u.machine) == ('Windows', 'AMD64'):
+        out = get_cache_home() / 'llvm15'
+        url = 'https://github.com/python3kgae/taichi_assets/releases/download/llvm15_vs2019_clang/taichi-llvm-15.0.0-msvc2019.zip'
```
bobcao3, line 101: Just to double check, shouldn't `'https://github.com/python3kgae/taichi_assets/releases/download/llvm15_vs2022_clang/clang-15.0.0-win.zip'` contain LLVM already?

.github/workflows/scripts/build.py
```diff
@@ -37,54 +87,82 @@ def setup_llvm(env_out: dict) -> None:
...
+        os.environ['VULKAN_SDK'] = str(sdk)
+        path_prepend('PATH', sdk / "bin")
+        path_prepend('LD_LIBRARY_PATH', sdk / 'lib')
+        os.environ['VK_LAYER_PATH'] = str(sdk / 'etc' / 'vulkan' /
+                                          'explicit_layer.d')
     # elif (u.system, u.machine) == ("Darwin", "arm64"):
     # elif (u.system, u.machine) == ("Darwin", "x86_64"):
-    # elif u.system == "Windows":
+    elif (u.system, u.machine) == ('Windows', 'AMD64'):
+        url = 'https://sdk.lunarg.com/sdk/download/1.3.236.0/windows/VulkanSDK-1.3.236.0-Installer.exe'
```
bobcao3, line 126: IMO, in a one-click build script we shouldn't install a full SDK on the user's machine, it's too intrusive. Building Vulkan backend does not requrie the SDK, the SDK is only needed to load validation layer. We should instruct users install their own SDK (and the windows installer will setup the VULKAN_SDK variable correctly)

.github/workflows/scripts/build.py
feisuzhu: This is just a standalone `clang++` binary which has a reasonable size. Installing Clang++ from VS would download at least 2GiB of stuff, thus I stopped following the original approach.

.github/workflows/scripts/build.py
feisuzhu: No, the clang zip only contains a clang++.exe binary.

.github/workflows/scripts/build.py
feisuzhu: (Today I learned!

The flag will not be set when using MSVC (guarded by `as_compiler`, ). See https://github.com/taichi-dev/taichi/pull/6993/files#diff-08e92060a244d51600d95a8a4e32047aa4f3c4bbb856becf14d67c83b6d12907R183

.github/workflows/scripts/build.py
feisuzhu: - For now this script is only targeting CI environment so this is fine.
- I totally agree the 'too intrusive' concern so every dependency is explicitly placed at a cache dir (See `get_cache_home`) including Vulkan SDK.
- When this script evolves into a developer facing script, all these dependency installing will be optional and be documented.

bobcao3 (approved):
LGTM

Issue #6983, opened 2022-12-26
Compile the arguments and return values to struct types to replace `RuntimeContext`


lin-hitonami:
**Concisely describe the proposed feature**
We want to compile the arguments and return values to struct types to replace `RuntimeContext` because `RuntimeContext` uses fixed-size arrays that always occupies maximum amount of space which are mostly wasted. If we compile them to struct types, we can save the memory.

PR #6994, opened 2022-12-28
[lang] Get the CHI-IR struct type in python

lin-hitonami:
Issue: #6983

### DIFF
--- a/python/taichi/lang/matrix.py
+++ b/python/taichi/lang/matrix.py
@@ -1436,8 +1436,10 @@ class MatrixType(CompoundType):
         #                 Remove the None dtype when we are ready to break legacy code.
         if dtype is not None:
             self.dtype = cook_dtype(dtype)
+            self.tensor_type = TensorType((n, m) if ndim == 2 else (n,), self.dtype)
         else:
             self.dtype = None
+            self.tensor_type = None
 
     def __call__(self, *args):
         """Return a matrix matching the shape and dtype.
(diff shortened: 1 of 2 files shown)


python/taichi/lang/matrix.py
```diff
@@ -1436,8 +1436,11 @@ def __init__(self, n, m, ndim, dtype):
         #                 Remove the None dtype when we are ready to break legacy code.
         if dtype is not None:
             self.dtype = cook_dtype(dtype)
```
jim19930609, line 1438: Should reuse `self.dtype` to store the TensorType.

lin-hitonami: But it seems that `MatrixType.dtype` is used in many places and it will take much time to rename it. Do you plan to remove `MatrixType` in the future? If so, I don't think we should waste time on this.

jim19930609: MatrixType will be preserved since we'll need a type annotation for Matrix object anyway. I feel that having both `dtype` and `tensor_type` in the MatrixType is too confusing.

lin-hitonami: Maybe we can refactor in another PR so that it will not block my successive PRs？

ailzhang: +1 on having both `dtype` and `tensor_type` is confusing, @lin-hitonami if this PR is blocking some other PRs please feel free to merge first but a followup PR to eliminate `tensor_type` is pretty necessary to avoid making this part error prone. ;)

ailzhang (approved):
lgtm!

PR #6995, opened 2022-12-28
[ir] Record the return types to a StructType

lin-hitonami:
Issue: #6983

### DIFF
--- a/taichi/program/callable.cpp
+++ b/taichi/program/callable.cpp
@@ -33,7 +33,7 @@ int Callable::insert_texture_arg(const DataType &dt) {
 
 void Callable::finalize_rets() {
   std::vector<const Type *> types;
-  for (auto &ret : rets) {
+  for (const auto &ret : rets) {
     types.push_back(ret.dt);
   }
   ret_type =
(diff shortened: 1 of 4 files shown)


taichi/program/callable.cpp
```diff
@@ -31,6 +31,15 @@ int Callable::insert_texture_arg(const DataType &dt) {
   return (int)args.size() - 1;
 }
 
+void Callable::finalize_rets() {
+  std::vector<const Type *> types;
```
bobcao3, line 35: nit:
```suggestion
types.reserve(rets.size());
```

taichi/program/callable.cpp
```diff
@@ -31,6 +31,15 @@ int Callable::insert_texture_arg(const DataType &dt) {
...
+  for (auto &ret : rets) {
```
bobcao3, line 36: tiny nit: `auto` does not deduce const (use const auto)

bobcao3 (approved):
LGTM

ailzhang (approved):
LGTM!

PR #6998, opened 2022-12-28
[aot] AOT compat test

damnkk:


PENGUINLIONG (approved):
lgtm

PENGUINLIONG (approved):

ailzhang (commented):
nit: in case that a known breakage happens, which script should we run to re-generate this tcm?

PENGUINLIONG:
> nit: in case that a known breakage happens, which script should we run to re-generate this tcm? 

@ailzhang The TCM is currently manually unzipped, had the `required_caps` field removed and again packed up. I'm not quite sure if there is an expected pattern to break the thing. Or maybe we should automate the packing-edit-unpacking process in a python script?

ailzhang:
@PENGUINLIONG yup it'll be nice to automate the process since anyone who fails this test in CI can have an idea what to do next ;)

Issue #6933, opened 2022-12-20
Updating Taichi results in slower compilation speed


strongoier:
Original user post: https://forum.taichi-lang.cn/t/topic/3710
Script to reproduce: https://github.com/JinliBot7/taichi-version-update-problem

The user reports that the script needs <10s to compile with Taichi v1.1.3 but needs ~30s to compile with Taichi v1.3.0.

On my MacBook Pro (13-inch, M1, 2020), I observe that the script needs ~5s to compile with Taichi v1.1.3 and needs ~20s to compile at 6f4ce42, which shows similar slowdown.

PR #7001, opened 2022-12-28
[opt] Add ImmediateIRModifier to provide amortized constant-time replace_usages_with()

strongoier:
Issue: #6933

### Brief Summary

`ImmediateIRModifier` aims at replacing `Stmt::replace_usages_with`, which visits the whole tree for a single replacement. `ImmediateIRModifier` is currently associated with a pass, visits the whole tree once at the beginning of that pass, and performs a single replacement with amortized constant time. It is now used in two most recent passes, `eliminate_immutable_local_vars` and `scalarize`. More passes can be modified to leverage it in the future.

After this PR, the profiling result of the script in #6933 shows that the time of `eliminate_immutable_local_vars` reduces from `0.956 s` to `0.162 s`, the time of `scalarize` reduces from `3.510 s` to `0.478 s`, and the total time of `compile` reduces from `8.550 s` to `4.696 s`.

### DIFF
--- a/taichi/analysis/gather_statement_usages.cpp
+++ b/taichi/analysis/gather_statement_usages.cpp
@@ -39,7 +39,8 @@ class GatherStatementUsages : public BasicStmtVisitor {
     default_visit(stmt);
   }
 
-  static std::unordered_map<Stmt *, std::vector<std::pair<Stmt *, int>>> run(IRNode *node) {
+  static std::unordered_map<Stmt *, std::vector<std::pair<Stmt *, int>>> run(
+      IRNode *node) {
     GatherStatementUsages pass;
     node->accept(&pass);
     return pass.stmt_usages_;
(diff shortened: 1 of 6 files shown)


taichi/analysis/gather_statement_usages.cpp
```diff
@@ -0,0 +1,59 @@
...
+    invoke_default_visitor = true;
+  }
+
+  void default_visit(Stmt *stmt) {
+    auto ops = stmt->get_operands();
+    for (int i = 0; i < ops.size(); i++) {
+      auto &op = ops[i];
+      if (op != nullptr) {
+        if (stmt_usages_.find(op) == stmt_usages_.end()) {
+          stmt_usages_[op] = {{stmt, i}};
```
lin-hitonami, line 26: IIRC `stmt_usages_[op].push_back({stmt, i})` works no matter `op` is inserted before or not.

taichi/analysis/gather_statement_usages.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "taichi/ir/visitors.h"
+
+namespace taichi::lang {
+
+class GatherStatementUsages : public BasicStmtVisitor {
+ private:
+  using BasicStmtVisitor::visit;
+
+  // maps a stmt to all its usages <stmt, operand>
+  std::unordered_map<Stmt *, std::vector<std::pair<Stmt *, int>>> stmt_usages_;
```
bobcao3, line 13: Can we benchmark whether this or `std::unordered_multimap` with `equal_range` iterators are faster?

strongoier: My attempt shows that they run at the same speed :-)

bobcao3 (approved):

Issue #6999, opened 2022-12-28
Draw circles/particles with different radii in GGUI


YouJiacheng:
**Concisely describe the proposed feature**
Currently `canvas.circles` and `scene.particles` only supports a single scalar radius, prevent users drawing many circles/particles with different radii in batch. I would like to draw many circles/particles with different radii efficiently, using one `canvas.circles` or `scene.particles` call.

**Describe the solution you'd like (if any)**
Allow parameter `radius` to be a taichi field. 

**Additional comments**
`ti.GUI.circles` allows parameter `radius` to be a numpy ndarray. BTW, the documentation seems outdated:
https://github.com/taichi-dev/taichi/blob/449a7f60a57fd0e2213bc8fe984cd6627feef9ce/python/taichi/ui/gui.py#L355-L365

https://github.com/taichi-dev/taichi/blob/449a7f60a57fd0e2213bc8fe984cd6627feef9ce/python/taichi/ui/gui.py#L430-L434

It should be `radius (Union[Number, np.array], optional)` like `ti.GUI.arrows`.

YouJiacheng:
Hmmm. I don't think this issue should be closed by a docstring fix. 
1. Does the `canvas.circles` really support array `radius`? And I afraid it should be taichi field instead of numpy array in GGUI.
2. What about `scene.particles`?

bobcao3:
Duplicate of #3888 ?

PR #7003, opened 2022-12-29
[Lang] Fix gui docstring

neozhaoliang:
Close: #6999

### DIFF
--- a/python/taichi/ui/canvas.py
+++ b/python/taichi/ui/canvas.py
@@ -111,7 +111,7 @@ class Canvas:
         Args:
             centers: a taichi 2D Vector field, where each element indicate the \
                 3D location of a vertex.
-            radius (float): radius of the circles, relative to the height of the screen.
+            radius (Union[float, numpy.array]): radius of the circles in pixel.
             color: a global color for the triangles as 3 floats representing \
                 RGB values. If `per_vertex_color` is provided, this is ignored.
             per_vertex_color (Tuple[float]): a taichi 3D vector field, where \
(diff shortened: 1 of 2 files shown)


python/taichi/ui/canvas.py
```diff
@@ -111,7 +111,7 @@ def circles(self,
         Args:
             centers: a taichi 2D Vector field, where each element indicate the \
                 3D location of a vertex.
-            radius (float): radius of the circles, relative to the height of the screen.
+            radius (Union[float, numpy.array]): radius of the circles in pixel.
```
ailzhang, line 114: nit: do we limit the dimension of the numpy array?

python/taichi/ui/canvas.py
```diff
@@ -111,7 +111,9 @@ def circles(self,
         Args:
             centers: a taichi 2D Vector field, where each element indicate the \
                 3D location of a vertex.
-            radius (float): radius of the circles, relative to the height of the screen.
+            radius (Union[float, numpy.array]): radius of the circles in pixels. \
+                Can be either a single number, which will be applied to all circles, or a 1D NumPy array of the same length as `centers`.\
+                The `i-th` item in the array will
```
ailzhang, line 116: nit: unfinished sentence?

neozhaoliang: Sorry, that's so awkward....

ailzhang (approved):
LGTM!

YouJiacheng:
According to the shader
https://github.com/taichi-dev/taichi/blob/3305acc76dade9844a9617648d9931940902d1c9/python/taichi/shaders/Circles_vk.vert

I don't think that GGUI support array/field `radius`.

neozhaoliang:
@YouJiacheng Thanks for pointing out this. The `GUI.circles` does support passing a numpy array for the radius, but `GGUI.circles` doesn't. This is will be fixed in another PR #7065

PR #7007, opened 2022-12-29
[ir] Let GetElementExpression&Statement support index list

lin-hitonami:
Issue: #6983


bobcao3:
IIRC there was a ExpressionStatement wrapper or something (turns a rvalue into a lvalue), should we use that instead of duplicating the statements?

lin-hitonami:
> IIRC there was a ExpressionStatement wrapper or something (turns a rvalue into a lvalue), should we use that instead of duplicating the statements?

I've never heard of it. Where is it?

bobcao3:
There's a FrontendExprStmt. Is the expression used only in frontend? (I would assume so)

lin-hitonami:
> There's a FrontendExprStmt. Is the expression used only in frontend? (I would assume so)

The `Frontend***Stmt`s are frontend statements that receive `Expr`s as arguments while other statements that do not start with `Frontend` receive `Stmt`s as arguments. The `Frontend***Stmt`s are replaced by other statements in the `lower_ast` pass, and they do no reach the codegen.

`FrontendExprStmt` adds an `Expr` to the IR without needing a `Stmt` containing it. The `FrontendExprStmt` is replaced by the lowered `Stmt` of the `Expr` in the `lower_ast` pass. I don't think this can solve the problem that we must have an `***Expr` or an `Frontend***Stmt` for frontend and a `***Stmt` for backend...

lin-hitonami:
Reopened with ghstack. Closing this one. Let's discuss in https://github.com/taichi-dev/taichi/pull/7026

PR #7012, opened 2022-12-30
[cuda] Fix LLVM15 rsqrt perf regression

### LINKED ISSUE #6957: CUDA performance regression of n-body after LLVM15 upgrade
Charts:

Compare with previous releases
image

Perfmon records
image

turbo0628:
Fixes #6957

### DIFF
--- a/taichi/runtime/cuda/jit_cuda.cpp
+++ b/taichi/runtime/cuda/jit_cuda.cpp
@@ -170,12 +170,10 @@ std::string JITSessionCUDA::compile_module_to_ptx(
        * https://github.com/llvm/llvm-project/commit/a4451d88ee456304c26d552749aea6a7f5154bde#diff-6fda74ef428299644e9f49a2b0994c0d850a760b89828f655030a114060d075a
        */
       fn.addFnAttr("denormal-fp-math-f32", "preserve-sign");
-    }
-  }
 
-  // Use unsafe fp math for sqrt.approx instead of sqrt.rn
-  for (llvm::Function &fn : *module) {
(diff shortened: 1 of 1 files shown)


taichi/runtime/cuda/jit_cuda.cpp
```diff
@@ -165,10 +165,19 @@ std::string JITSessionCUDA::compile_module_to_ptx(
...
-      fn.addFnAttr("nvptx-f32ftz", "true");
+      /* nvptx-f32ftz was deprecated.
+       *
+       * https://github.com/llvm/llvm-project/commit/a4451d88ee456304c26d552749aea6a7f5154bde#diff-6fda74ef428299644e9f49a2b0994c0d850a760b89828f655030a114060d075a
+       */
+      fn.addFnAttr("denormal-fp-math-f32", "preserve-sign");
     }
   }
 
+  // Use unsafe fp math for sqrt.approx instead of sqrt.rn
```
ailzhang, line 176: aha can we reorganize this into below to save a for-loop? 

```
for (llvm::Function &fn: *module) {
    fn.addFnAttr("unsafe-fp-math", "true");
    if (kFTZDenorms) {
      // comment 
      fn.addFnAttr("denormal-fp-math-f32", "preserve-sign");
    }
}
```

ailzhang (approved):
Awesome!!

taichi/runtime/cuda/jit_cuda.cpp
turbo0628: sure!

PR #7013, opened 2022-12-30
[ir] Replace FuncCallExpression with FrontendFuncCallStmt

lin-hitonami:
Issue: #6983 

### Brief Summary
FuncCall is not just an expression. It executes code and returns the result. It should be more like `FrontendAllocaStmt` rather than being `FuncCallExpression`. If it is an expression, the IR printer handles it like an expression and prints the detail of the func call everywhere its return value is used. If it is a statement, it will only be printed once, and other statements/expressions reference its return value by its statement number.


lin-hitonami:
Reopened with ghstack. Closing this one. Let's discuss in https://github.com/taichi-dev/taichi/pull/7027

PR #7022, opened 2022-12-31
[amdgpu] Add amdgpu module call

GaleSeLee:

### Brief Summary
1. Add exclusive way to process kernel argument.
2. Upate `launch` api in `AMDGPUContext` .

### DIFF
--- a/taichi/jit/jit_module.h
+++ b/taichi/jit/jit_module.h
@@ -6,6 +6,7 @@
 #include "taichi/inc/constants.h"
 #include "taichi/util/lang_util.h"
 #include "taichi/program/kernel_profiler.h"
+#include "taichi/rhi/arch.h"
 
 namespace taichi::lang {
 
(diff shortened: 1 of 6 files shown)


taichi/jit/jit_module.h
```diff
@@ -44,15 +45,53 @@ class JITModule {
...
+  // Note: AMDGPU need to pass args by extra_arg currently
   template <typename... Args>
   void call(const std::string &name, Args... args) {
     if (direct_dispatch()) {
       get_function<Args...>(name)(args...);
     } else {
-      auto arg_pointers = JITModule::get_arg_pointers(args...);
-      call(name, arg_pointers);
+      if (module_arch() == Arch::cuda) {
+#if defined(TI_WITH_CUDA)
```
jim19930609, line 77: `jit_module.h` should only contain `base class` methods. 

You can modify:
```
virtual void call(const std::string &name,
                    const std::vector<void *> &arg_pointers) {
    TI_NOT_IMPLEMENTED
  }
```

to 

```
virtual void call(const std::string &name,
                    const std::vector<void *> &arg_pointers,
                    const std::vector<uint32_t>& arg_bytes) {
    TI_NOT_IMPLEMENTED
  }
```

and make different implementations in `jit_amd.h` and `jit_cuda.h`

taichi/rhi/amdgpu/amdgpu_context.cpp
```diff
@@ -63,15 +63,14 @@ std::string AMDGPUContext::get_device_name() {
...
                            unsigned grid_dim,
                            unsigned block_dim,
-                           std::size_t dynamic_shared_mem_bytes,
-                           int arg_bytes) {
+                           std::size_t dynamic_shared_mem_bytes) {
   if (grid_dim > 0) {
     std::lock_guard<std::mutex> _(lock_);
-    void *config[] = {(void *)0x01, const_cast<void *>(arg_pointers),
-                      (void *)0x02, &arg_bytes, (void *)0x03};
+    void *config[] = {(void *)0x01, arg_pointers[0], (void *)0x02,
```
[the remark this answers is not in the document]
GaleSeLee: 1. 0x01, 0x02 and 0x03 are the delimiters passed to `extra_arg` to identify arguments and the length of the arguments.
2. Sure! I will assert `arg_pointers.size() == 2`

taichi/jit/jit_module.h
GaleSeLee: I considered this design. If we adopt this design and don't judge module's arch, it will not work well when both enable `TI_WITH_CUDA` and `TI_WITH_AMDGPU`.

jim19930609: Basically the idea is that the compiler will create `JITModuleCUDA` if initialized with `ti.init(arch=ti.cuda)`, and `JITModuleAMD` if initialized with `ti.init(arch=ti.amdgpu)`. Can you explain into details in terms of why it doesn't work with both `TI_WITH_CUDA` and `TI_WITH_AMDGPU`?

GaleSeLee: In detail, `TI_WITH_CUDA` and `TI_WITH_AMDGPU` are the control variants in complie-scope. If user enables both `TI_WITH_CUDA` and `TI_WITH_AMDGPU`, we should use somehow to get the arch info. In code. However,  if we use macros(`TI_WITH_CUDA` and `TI_WITH_AMDGPU`) to control the which `call` function we use. 
```
#if defined(TI_WITH_CUDA)
use cuda-related call
#else ifdefined(TI_WITH_AMDGPU)
use amdgpu-related call
#endif
```
We will never call amdgpu-related `call` function, even `ti.init(arch=ti.amdgpu)`

jim19930609 (approved):
Thanks~

PR #7026, opened 2023-01-03
[ir] Let GetElementExpression&Statement support index list

lin-hitonami:
* #7027
* __->__ #7026

### DIFF
--- a/taichi/codegen/llvm/codegen_llvm.cpp
+++ b/taichi/codegen/llvm/codegen_llvm.cpp
@@ -2735,13 +2735,15 @@ void TaskCodeGenLLVM::visit(FuncCallStmt *stmt) {
 
 void TaskCodeGenLLVM::visit(GetElementStmt *stmt) {
   auto *real_func = stmt->src->as<FuncCallStmt>()->func;
-  auto &rets = real_func->rets;
-  auto *ret_type = get_real_func_ret_type(real_func);
-  auto *gep = builder->CreateGEP(
-      ret_type, llvm_val[stmt->src],
-      {tlctx->get_constant(0), tlctx->get_constant(stmt->index)});
-  auto *val =
-      builder->CreateLoad(tlctx->get_data_type(rets[stmt->index].dt), gep);
+  auto *real_func_ret_type = tlctx->get_data_type(real_func->ret_type);
+  std::vector<llvm::Value *> index;
+  index.push_back(tlctx->get_constant(0));
+  for (auto &i : stmt->index) {
+    index.push_back(tlctx->get_constant(i));
+  }
+  auto *gep =
+      builder->CreateGEP(real_func_ret_type, llvm_val[stmt->src], index);
+  auto *val = builder->CreateLoad(tlctx->get_data_type(stmt->ret_type), gep);
   llvm_val[stmt] = val;
 }
 
--- a/taichi/ir/frontend_ir.cpp
+++ b/taichi/ir/frontend_ir.cpp
@@ -1124,8 +1124,8 @@ void GetElementExpression::type_check(CompileConfig *config) {
   TI_ASSERT_TYPE_CHECKED(src);
   auto func_call = src.cast<FuncCallExpression>();
   TI_ASSERT(func_call);
-  TI_ASSERT(index < func_call->func->rets.size());
-  ret_type = func_call->func->rets[index].dt;
+  TI_ASSERT(index[0] < func_call->func->rets.size());
+  ret_type = func_call->func->rets[index[0]].dt;
 }
 
 void GetElementExpression::flatten(FlattenContext *ctx) {
--- a/taichi/ir/statements.h
+++ b/taichi/ir/statements.h
@@ -1082,8 +1082,9 @@ class ReferenceStmt : public Stmt, public ir_traits::Load {
 class GetElementStmt : public Stmt {
  public:
   Stmt *src;
-  int index;
-  GetElementStmt(Stmt *src, int index) : src(src), index(index) {
+  std::vector<int> index;
+  GetElementStmt(Stmt *src, std::vector<int> index)
+      : src(src), index(std::move(index)) {
     TI_STMT_REG_FIELDS;
   }
 
--- a/taichi/transforms/ir_printer.cpp
+++ b/taichi/transforms/ir_printer.cpp
@@ -825,6 +825,16 @@ class IRPrinter : public IRVisitor {
     print(result);
   }
 
+  void visit(GetElementStmt *stmt) override {
+    std::string index = "";
+    for (int i = 0; i < stmt->index.size(); i++) {
+      index += ", ";
+      index += std::to_string(stmt->index[i]);
+    }
+    print("{}{} = get_element({}{})", stmt->type_hint(), stmt->name(),
+          stmt->src->name(), index);
+  }
+
  private:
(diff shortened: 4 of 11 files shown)


lin-hitonami:
Discussions from https://github.com/taichi-dev/taichi/pull/7007:

[bobcao3](https://github.com/bobcao3) commented [3 days ago](https://github.com/taichi-dev/taichi/pull/7007#issuecomment-1368108245)
IIRC there was a ExpressionStatement wrapper or something (turns a rvalue into a lvalue), should we use that instead of duplicating the statements?

@lin-hitonami
Member
Author
[lin-hitonami](https://github.com/lin-hitonami) commented [3 days ago](https://github.com/taichi-dev/taichi/pull/7007#issuecomment-1368151671)
IIRC there was a ExpressionStatement wrapper or something (turns a rvalue into a lvalue), should we use that instead of duplicating the statements?

I've never heard of it. Where is it?

@bobcao3
Member
[bobcao3](https://github.com/bobcao3) commented [2 days ago](https://github.com/taichi-dev/taichi/pull/7007#issuecomment-1368192324)
There's a FrontendExprStmt. Is the expression used only in frontend? (I would assume so)

@lin-hitonami
Member
Author
[lin-hitonami](https://github.com/lin-hitonami) commented [53 minutes ago](https://github.com/taichi-dev/taichi/pull/7007#issuecomment-1369347864) • 
There's a FrontendExprStmt. Is the expression used only in frontend? (I would assume so)

The `Frontend***Stmts` are frontend statements that receive `Expressions` as arguments while other statements that do not start with Frontend receive Stmts as arguments. The `Frontend***Stmts` and `***Expressions` are replaced by lowered statements in the lower_ast pass, and they do no reach the codegen.

`FrontendExprStmt` adds an `Expression` to the IR without needing a `Stmt` containing it. The `FrontendExprStmt` is replaced by the lowered `Stmt` of the `Expression` in the `lower_ast` pass. I don't think this can solve the problem that we must have an `***Expression` or a `Frontend***Stmt` for frontend and a `***Stmt` for backend...

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2735,13 +2735,15 @@ void TaskCodeGenLLVM::visit(FuncCallStmt *stmt) {
...
   auto *real_func = stmt->src->as<FuncCallStmt>()->func;
-  auto &rets = real_func->rets;
-  auto *ret_type = get_real_func_ret_type(real_func);
-  auto *gep = builder->CreateGEP(
-      ret_type, llvm_val[stmt->src],
-      {tlctx->get_constant(0), tlctx->get_constant(stmt->index)});
-  auto *val =
-      builder->CreateLoad(tlctx->get_data_type(rets[stmt->index].dt), gep);
+  auto *real_func_ret_type = tlctx->get_data_type(real_func->ret_type);
+  std::vector<llvm::Value *> index;
```
bobcao3, line 2739: nit: reserve space or pre-allocate storage

taichi/ir/frontend_ir.cpp
```diff
@@ -1124,8 +1124,8 @@ void GetElementExpression::type_check(CompileConfig *config) {
   TI_ASSERT_TYPE_CHECKED(src);
   auto func_call = src.cast<FuncCallExpression>();
   TI_ASSERT(func_call);
-  TI_ASSERT(index < func_call->func->rets.size());
-  ret_type = func_call->func->rets[index].dt;
+  TI_ASSERT(index[0] < func_call->func->rets.size());
```
bobcao3, line 1127: This seems a bit confusing, what is `index[0]` and what's special about it?

bobcao3: Should probably put a comment somewhere denoting that `index[0]` is different

taichi/ir/statements.h
```diff
@@ -1082,8 +1082,9 @@ class ReferenceStmt : public Stmt, public ir_traits::Load {
 class GetElementStmt : public Stmt {
  public:
   Stmt *src;
-  int index;
-  GetElementStmt(Stmt *src, int index) : src(src), index(index) {
+  std::vector<int> index;
+  GetElementStmt(Stmt *src, std::vector<int> index)
```
bobcao3, line 1086:
```suggestion
  GetElementStmt(Stmt *src, const std::vector<int> &index)
```

taichi/ir/statements.h
```diff
@@ -1082,8 +1082,9 @@ class ReferenceStmt : public Stmt, public ir_traits::Load {
...
+      : src(src), index(std::move(index)) {
```
bobcao3, line 1087:
```suggestion
      : src(src), index(index) {
```

bobcao3: Either we move it from the caller (which is destructive, not sure that's the best idea)
Or we make a copy from a const reference

taichi/transforms/ir_printer.cpp
```diff
@@ -825,6 +825,16 @@ class IRPrinter : public IRVisitor {
     print(result);
   }
 
+  void visit(GetElementStmt *stmt) override {
+    std::string index = "";
+    for (int i = 0; i < stmt->index.size(); i++) {
```
bobcao3, line 830: We can use `fmt::join(stmt->index, ", ")` here, much cleaner

@k-ye Is it correct that we will very likely keep `spdlog` and `libfmt` a part of the compiler infra?

taichi/transforms/type_check.cpp
```diff
@@ -414,8 +414,8 @@ class TypeCheck : public IRVisitor {
   void visit(GetElementStmt *stmt) override {
     TI_ASSERT(stmt->src->is<FuncCallStmt>());
     auto *func = stmt->src->as<FuncCallStmt>()->func;
-    TI_ASSERT(stmt->index < func->rets.size());
-    stmt->ret_type = func->rets[stmt->index].dt;
+    TI_ASSERT(stmt->index[0] < func->rets.size());
```
bobcao3, line 417: Same as above, what's special about index 0, leave a comment here or make it a named parameter

taichi/ir/statements.h
lin-hitonami: If an lvalue is passed, it will first be copied and then the copied one is moved, so it's not destructive. IIUC If an rvalue is passed, if we pass by const reference, it is copied (or maybe it will be optimized out, I'm not sure). However if we pass by value, it is not copied. So  I think passing by value is better than by const reference, at least not worse than.

taichi/ir/frontend_ir.cpp
lin-hitonami: The return value is still flattened now, so the length of the index is always 1. I will refactor in the subsequent PRs to let it compile to nested structs.

taichi/ir/statements.h
bobcao3: If rvalue is passed and if it's const reference it won't be copied. In the end the result is the same. However a const reference is cleaner than passing in a copy then move because in the constructor access to `index` will then be invalid.

taichi/ir/frontend_ir.cpp
lin-hitonami: In fact they are refactored in #7027 lol

bobcao3 (approved):
LGTM

PR #7027, opened 2023-01-03
[ir] Replace FuncCallExpression with FrontendFuncCallStmt

lin-hitonami:
* #7059
* __->__ #7027

issue: #6983
FuncCall is not just an expression. It executes code and returns the result. It should be more like FrontendAllocaStmt rather than being FuncCallExpression. If it is an expression, the IR printer handles it like an expression and prints the detail of the func call everywhere its return value is used. If it is a statement, it will only be printed once, and other statements/expressions reference its return value by its statement number.


taichi/ir/frontend_ir.cpp
```diff
@@ -1332,6 +1310,14 @@ Expr ASTBuilder::expr_alloca() {
   return var;
 }
 
+Expr ASTBuilder::insert_func_call(Function *func, const ExprGroup &args) {
+  auto var = Expr(std::make_shared<IdExpression>(get_next_id()));
```
bobcao3, line 1314: Are functions guaranteed to return a value?

taichi/transforms/lower_ast.cpp
```diff
@@ -82,6 +82,21 @@ class LowerAST : public IRVisitor {
     }
   }
 
+  void visit(FrontendFuncCallStmt *stmt) override {
+    auto block = stmt->parent;
+    auto ident = stmt->ident;
+    TI_ASSERT(block->local_var_to_stmt.find(ident) ==
+              block->local_var_to_stmt.end());
+    std::vector<Stmt *> args;
```
bobcao3, line 90:
```suggestion
    std::vector<Stmt *> args(stmt->args.exprs.size());
```

and switch the push back to indexing

taichi/transforms/lower_ast.cpp
```diff
@@ -82,6 +82,21 @@ class LowerAST : public IRVisitor {
     }
   }
 
+  void visit(FrontendFuncCallStmt *stmt) override {
+    auto block = stmt->parent;
```
bobcao3, line 86: nit: Use the types here. `auto` doesn't deduce references and if the underlying type is a simple pointer it's easier to read & less likely to mess up

taichi/ir/frontend_ir.cpp
lin-hitonami: If it has no return value, it returns an empty struct.

bobcao3: This feels iffy... Some IR might not like the idea of an empty struct

lin-hitonami: Maybe we can deal with this in the codegen. We can return void in the codegen if there is no return value (the return type is an empty struct in CHI-IR). In this case the return value is not used anywhere so we don't need to care about its value.

bobcao3: Maybe it's better to return a std::optional

bobcao3 (approved):
LGTM

PR #7031, opened 2023-01-03
[opengl] Enable more gles tests in CI

ailzhang:
Issue: #

### Brief Summary
Reland of #7010  with a hack that running gles tests alone. Wanted to enable the tests to prevent regression, will fix the crash that happens when testing vulkan,gles,opengl at the same time.


feisuzhu (approved):
LGTM

Issue #7034, opened 2023-01-03
Support more iterative solvers


FantasyVR:
**Concisely describe the proposed feature**
Sparse matrix and direct sparse solvers are supported in Taichi. To solve very large linear systems, iterative solvers like conjugate gradient solvers (CG)  are more suitable. 

**Describe the solution you'd like (if any)**

- [x] Wrap Eigen's CG solver on CPU backend
- [x] Implement a CG solver on CUDA backend

PR #7035, opened 2023-01-03
[lang] Experimental support of conjugate gradient solver

FantasyVR:
Issue: #7034


python/taichi/linalg/sparse_matrix.py
```diff
@@ -254,10 +255,10 @@ def build(self, dtype=f32, _format='CSR'):
         taichi_arch = get_runtime().prog.config().arch
         if taichi_arch == _ti_core.Arch.x64 or taichi_arch == _ti_core.Arch.arm64:
             sm = self.ptr.build()
-            return SparseMatrix(sm=sm)
+            return SparseMatrix(sm=sm, dtype=dtype)
         if taichi_arch == _ti_core.Arch.cuda:
             sm = self.ptr.build_cuda()
-            return SparseMatrix(sm=sm)
+            return SparseMatrix(sm=sm, dtype=dtype)
```
houkensjtu, line 261: Can we give an error message here for `ti.f64` dtype on CUDA backend?

tests/python/test_cg.py
```diff
@@ -0,0 +1,39 @@
+import numpy as np
+import pytest
+
+import taichi as ti
+from tests import test_utils
+
+
+@pytest.mark.parametrize("ti_dtype", [ti.f32, ti.f64])
+@test_utils.test(arch=ti.cpu)
```
houkensjtu, line 9: Shall we include a test for CUDA backend here, too?

FantasyVR: CUDA backend CG solver hasn't been implemented yet.

python/taichi/linalg/sparse_matrix.py
houkensjtu: It's probably better to raise the error before doing `sm = self.ptr.build_cuda()`.

houkensjtu (approved):
Thank you and LGTM!

taichi/program/conjugate_gradient.cpp
```diff
@@ -0,0 +1,94 @@
...
+  if (!CUBLASDriver::get_instance().is_loaded()) {
+    bool load_success = CUBLASDriver::get_instance().load_cublas();
+    if (!load_success) {
+      TI_ERROR("Failed to load cublas library!");
+    }
+  }
+  CUBLASDriver::get_instance().cubCreate(&handle);
+  // int version;
+  // CUBLASDriver::get_instance().cubGetVersion(handle, &version);
+  // fmt::print("CUBLAS version: {}\n", version);
```
strongoier, line 15: Are we intended to keep these comments?

FantasyVR: I'm not sure if we should print out the version of CuBlas.

strongoier: We can print info like this using `TI_TRACE`.

strongoier (approved):

PR #7037, opened 2023-01-04
[rhi] Update compute CommandList APIs (except dispatch)

bobcao3:
Issue: #6832

### Brief Summary

Compute-only CommandList APIs other than dispatch has been changed to use `noexcept`, and the behavior specifications has been added, and relevant checks are now in place for Vulkan and partially for DX11.


k-ye (approved):
LGTM!

PR #7038, opened 2023-01-04
[Doc] Update type.md

neozhaoliang:
Since we have made `matrix` and `vector` to be different types in v1.4.0, the documentation needs to be updated.


Olinaaaloompa (approved):
LGTM! Thanks

Issue #7002, opened 2022-12-28
Refactor kernel compilation


PGZXB:
(This is part of **_Refactor kernel compilation and launch_**, which aims to split kernel compilation and kernel launching totally.)

### Brief Introduction
1. Isolate the stuff which is not related with kernel compilation from `lang::Kernel`.
2. **`lang::CompileConfig` per compilation** instead of `lang::CompileConfig` per `lang::Program`.
3. Introduce `lang::CompiledKernelData`, `lang::KernelCompiler` to **unify** the compilation result and the interfaces of kernel compilation.
4. Introduce `lang::KernelCompilationManager` to unify the implementation of _cache_ on **ALL** backends, including _online cache_ and _[offline cache](https://github.com/taichi-dev/taichi/issues/4401)_
5. Re-impl the JIT in Python frontend and remove the _online cache_ in frontend 

### Prototype
* An **incomplete** implementation of this proposal: see https://github.com/taichi-dev/taichi/pull/6819 or [PGZXB:dev-refactor-kernelCR-1st-prototype](https://github.com/PGZXB/taichi/tree/dev-refactor-kernelCR-1st-prototype)

### Provide `@ti.pkernel`
> _**NOTE:** The `@ti.pkernel` is **only an attempt** based on the prototype in order to **verify the effect of the proposal**, which is **NOT** a "feature request"._

* `@ti.pkernel` = `@ti.kernel` + `CompileConfig`, which allows the user to specify different compile configuration options for each kernel.
#### Implementation
* See [PGZXB:dev-pkernel-based_on-refactorKCaL_1st_prototype](https://github.com/PGZXB/taichi/tree/dev-pkernel-based_on-refactorKCaL_1st_prototype)
  
#### Example
```python
import taichi as ti

ti.init(arch=ti.vulkan, offline_cache=True, print_preprocessed_ir=False)

@ti.kernel
def K1(a: ti.i32, b: ti.i32) -> ti.i32:
  return a + b


@ti.pkernel(offline_cache=False, print_preprocessed_ir=True)
def PK1(a: ti.i32, b: ti.i32) -> ti.i32:
  return a * b

# Not allowed before finishing "Refactor kernel launch"
# @ti.pkernel(arch=ti.cpu) # ti.cpu != ti.vulkan
# def PK2(a: ti.i32, b: ti.i32) -> ti.i32:
#   return a * b

K1(10, 10)
PK1(10, 10)
# PK2(10, 100)
```

### To-Do List

[...]

PR #7048, opened 2023-01-04
[refactor] Move Kernel::lower() outside the taichi::lang::Kernel

PGZXB:
Issue: #7002


taichi/transforms/compile_to_offloads.cpp
```diff
@@ -41,6 +41,14 @@ void compile_to_offloads(IRNode *ir,
   auto print = make_pass_printer(verbose, kernel->get_name(), ir);
   print("Initial IR");
 
+  if (config.print_preprocessed_ir && start_from_ast) {
+    TI_INFO("[{}] {}:", kernel->get_name(), "Preprocessed IR");
+    std::cout << std::flush;
+    irpass::re_id(ir);
+    irpass::print(ir);
+    std::cout << std::flush;
```
strongoier, line 49: Maybe we don't need to print preprocessed IR again when `verbose == true`. WDYT @lin-hitonami

PGZXB: Agree!

lin-hitonami: makes sense.

strongoier (approved):

PR #7051, opened 2023-01-04
[opt] Add ExtractPointers pass for dynamic index

strongoier:
Issue: #2590

### Brief Summary

Under pure `dynamic_index` setting, `MatrixPtrStmt`s are not scalarized. It actually produces `2n` more instructions (`n` `ConstStmt`s and n `MatrixPtrStmt`s) than the scalarized setting, where `n` is the number of usages of `MatrixPtrStmt`s. This PR adds `ExtractPointers` pass to eliminate all the redundant instructions. See comments in the code for details.

After this PR, the number of instructions after the `scalarize()` pass of the script in #6933 under dynamic index reduces from 49589 to 26581, and the compilation time reduces from 20.02s to 7.82s.

### DIFF
--- a/taichi/transforms/scalarize.cpp
+++ b/taichi/transforms/scalarize.cpp
@@ -676,9 +676,14 @@ class ExtractPointers : public BasicStmtVisitor {
   ImmediateIRModifier immediate_modifier_;
   DelayedIRModifier delayed_modifier_;
 
-  std::unordered_map<std::pair<Stmt *, int>, Stmt *, hashing::Hasher<std::pair<Stmt *, int>>>
-      first_matrix_ptr_;  // mapping an (AllocaStmt, integer) pair to the first MatrixPtrStmt representing it
-  std::unordered_map<int, Stmt *> first_const_;  // mapping an integer to the first ConstStmt representing it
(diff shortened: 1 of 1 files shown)


jim19930609 (approved):
I was thinking `MatrixInitStmt` might have the same optimization opportunity, if we change `std::vector<Stmt*> elements` to sth like `std::vector<std::optional<Stmt*, TypedConstant>> elements`:
```
%1 = ConstStmt(1);
%2 = ConstStmt(2);
%3 = MatrixInitStmt({%1, %1, %2, %2});
```

strongoier:
> I was thinking `MatrixInitStmt` might have the same optimization opportunity, if we change `std::vector<Stmt*> elements` to sth like `std::vector<std::optional<Stmt*, TypedConstant>> elements`:
> 
> ```
> %1 = ConstStmt(1);
> %2 = ConstStmt(2);
> %3 = MatrixInitStmt({%1, %1, %2, %2});
> ```

Ah nice point. My concern here is that `MatrixInitStmt` is composed of actual computation results, which might not be that repetitive. But yeah we might explore this idea in other scenarios in the future. Or we should have a better CSE pass in general (it is currently too slow to handle the case in this PR).

PR #7055, opened 2023-01-05
[opt] Automatically identify allocas to scalarize

strongoier:
Issue: #2590

### Brief Summary

Previously `ScalarizePointers` (which is for allocas) takes place only when `dynamic_index=False`. In this PR, we automatically identify those allocas only indexed with constants and get them scalarized, and let the remaining allocas go through the code path for dynamic indexing. In this way, we make old user programs behave the same regardless of the `dynamic_index` option, and provide dynamic indexing support for new user programs at the same time.

### DIFF
--- a/taichi/transforms/scalarize.cpp
+++ b/taichi/transforms/scalarize.cpp
@@ -573,7 +573,10 @@ class ScalarizeLocalPointers : public BasicStmtVisitor {
   // { original_alloca_stmt : [scalarized_alloca_stmt0, ...] }
   std::unordered_map<Stmt *, std::vector<Stmt *>> scalarized_local_tensor_map_;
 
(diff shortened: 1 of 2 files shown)


lin-hitonami (commented):
Should this pass be placed after a simplify pass? IIUC the offset may be optimized to a constant after constant folding.

strongoier:
> Should this pass be placed after a simplify pass? IIUC the offset may be optimized to a constant after constant folding.

Currently `scalarize()` has to be placed before any other pass :-( The current version ensures the behavior is the same as before (no optimized constants are considered), but in the future `scalarize()` will be postponed and able to use those optimized constants.

lin-hitonami (approved):
LGTM

PR #7059, opened 2023-01-05
[ir] Let real function return nested StructType

lin-hitonami:
* __->__ #7059
* #7027

issue: #6983

Before this PR, if we want to return a struct type in a real function, we flatten the struct in LLVM. For example, `struct(a: Ta, s: struct(b: Tb, c: Tc))` is flattened to `struct(a: Ta, b: Tb, c: Tc)`. After this PR, we do not flatten anymore. We directly return `struct(a: Ta, s: struct(b: Tb, c: Tc))`. 

Note that we still need  to pass the flattened return values from Python to CHI-IR because we can't pass a Python struct into C++. We compose the target struct type from the flattened values and the return type in the codegen.


taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2757,6 +2749,49 @@ llvm::Type *TaskCodeGenLLVM::get_real_func_ret_type(Function *real_func) {
...
+  }
+}
+
+void TaskCodeGenLLVM::create_return(const std::vector<Stmt *> &elements) {
+  auto buffer = call("RuntimeContext_get_result_buffer", get_context());
+  auto ret_type = now_real_func->ret_type;
+  auto buffer_type = tlctx->get_data_type(ret_type);
+  buffer = builder->CreatePointerCast(buffer,
+                                      llvm::PointerType::get(buffer_type, 0));
+  int now_element = 0;
```
jim19930609, line 2790: Better to use `current_element_index`, `current_real_func` and `current_indices`?

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2757,6 +2749,49 @@ llvm::Type *TaskCodeGenLLVM::get_real_func_ret_type(Function *real_func) {
   return llvm::StructType::get(*llvm_context, tps);
 }
 
+void TaskCodeGenLLVM::create_return(llvm::Value *buffer,
+                                    llvm::Type *buffer_type,
```
jim19930609, line 2753: Can we obtain buffer_type from `buffer->getType()`?

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2757,6 +2749,49 @@ llvm::Type *TaskCodeGenLLVM::get_real_func_ret_type(Function *real_func) {
...
+    int i = 0;
+    for (const auto &element_type : struct_type->elements()) {
+      now_ind.push_back(tlctx->get_constant(i++));
+      create_return(buffer, buffer_type, elements, element_type, now_element,
+                    now_ind);
+      now_ind.pop_back();
+    }
+  } else {
+    auto tensor_type = now_type->as<TensorType>();
+    int num_elements = tensor_type->get_num_elements();
```
jim19930609, line 2773: assert `tensor_type != nullptr` before use?

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2757,6 +2749,49 @@ llvm::Type *TaskCodeGenLLVM::get_real_func_ret_type(Function *real_func) {
   return llvm::StructType::get(*llvm_context, tps);
 }
 
+void TaskCodeGenLLVM::create_return(llvm::Value *buffer,
+                                    llvm::Type *buffer_type,
+                                    const std::vector<Stmt *> &elements,
+                                    const Type *now_type,
```
jim19930609, line 2755: Let's rename it to `current_type`?

taichi/codegen/llvm/codegen_llvm.h
```diff
@@ -138,6 +138,15 @@ class TaskCodeGenLLVM : public IRVisitor, public LLVMModuleBuilder {
 
   llvm::Value *create_print(std::string tag, llvm::Value *value);
 
+  void create_return(const std::vector<Stmt *> &elements);
+
+  void create_return(llvm::Value *buffer,
```
jim19930609, line 143: Looks like this overloaded `create_return()` is an implementation of `void create_return(const std::vector<Stmt *> &elements);`, which won't get called by other compile unit? If that is the case, let's make it a static function in `codegen_llvm.cpp`?

taichi/codegen/llvm/codegen_llvm.cpp
lin-hitonami: The `as` method already does that for us.

taichi/codegen/llvm/codegen_llvm.h
lin-hitonami: Do you mean private function? `create_return` uses a lot of non-static members of class `CodeGenLLVM` so it's not possible to be a static function.

jim19930609: yep I meant `private funtion`

jim19930609 (approved):

taichi/codegen/llvm/codegen_llvm.cpp
lin-hitonami: No... While it works fine on linux machines, it doesn't work on Windows because the pointers are opaque pointers, and `buffer->getType()` only returns `ptr` in LLVM 15. I don't know why it works on linux machines... maybe the LLVM binaries are different....

Issue #6438, opened 2022-10-26
ndarrays should be initialized to zero by default


ailzhang:
Per offline discussion, since we already do this for [fields](https://docs.taichi-lang.org/docs/field), ndarrays should match this behavior to avoid surprises on user end. 
We may also provide an option to skip this default initialization to maximize performnace if users need. 
cc: @neozhaoliang @k-ye @yuanming-hu

PR #7060, opened 2023-01-05
[test] Add test to make sure ndarrays are initialized as zero

ailzhang:
Issue: #6438 

### Brief Summary
This PR only enables zero fill for ndarrays created from Python frontend so it doesn't affect performance of ndarray created from C-API. 
In the future we may provide an option for python users to skip the zero fill as needed but that will be a separate PR.

### DIFF
--- a/python/taichi/lang/_ndarray.py
+++ b/python/taichi/lang/_ndarray.py
@@ -225,7 +225,7 @@ class ScalarNdarray(Ndarray):
         super().__init__()
         self.dtype = cook_dtype(dtype)
         self.arr = impl.get_runtime().prog.create_ndarray(
-            self.dtype, arr_shape)
+            self.dtype, arr_shape, Layout.NULL, True)
         self.shape = tuple(self.arr.shape)
         self.element_type = dtype
 
--- a/taichi/program/program.cpp
+++ b/taichi/program/program.cpp
@@ -455,8 +455,20 @@ std::size_t Program::get_snode_num_dynamically_allocated(SNode *snode) {
 
 Ndarray *Program::create_ndarray(const DataType type,
                                  const std::vector<int> &shape,
-                                 ExternalArrayLayout layout) {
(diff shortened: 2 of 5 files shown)


python/taichi/lang/_ndarray.py
```diff
@@ -225,7 +225,7 @@ def __init__(self, dtype, arr_shape):
         super().__init__()
         self.dtype = cook_dtype(dtype)
         self.arr = impl.get_runtime().prog.create_ndarray(
-            self.dtype, arr_shape)
+            self.dtype, arr_shape, Layout.NULL, True)
```
lin-hitonami, line 228:
```suggestion
            self.dtype, arr_shape, Layout.NULL, zero_fill=True)
```

taichi/program/program.cpp
```diff
@@ -455,8 +455,25 @@ std::size_t Program::get_snode_num_dynamically_allocated(SNode *snode) {
...
 Ndarray *Program::create_ndarray(const DataType type,
                                  const std::vector<int> &shape,
-                                 ExternalArrayLayout layout) {
-  ndarrays_.emplace_back(std::make_unique<Ndarray>(this, type, shape, layout));
+                                 ExternalArrayLayout layout,
+                                 bool zero_fill) {
+  auto arr = std::make_unique<Ndarray>(this, type, shape, layout);
+  if (zero_fill) {
+    Arch arch = this_thread_config().arch;
+    if (arch == Arch::x64 || arch == Arch::cuda) {
```
lin-hitonami, line 463: or arch == Arch::arm64? I guess it will fail on M1 because arm64 uses CpuDevice as well. Maybe we can use `arch_is_cpu(arch)` here.

ailzhang: good point!

lin-hitonami (approved):
LGTM

PR #7071, opened 2023-01-05
[Lang] Deprecate the dynamic_index switch

strongoier:
Issue: #2590

### Brief Summary

Dynamic indexing is now automatically supported and no switch is needed.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -192,14 +192,10 @@ class TestParam:
 
 if os.environ.get('TI_LITE_TEST', ''):
     _test_features = {
-        "dynamic_index": [TestParam(True, [])],
     }
 else:
     _test_features = {
-        #"packed":
-        # [TestValue(True, []),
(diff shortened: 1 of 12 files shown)


tests/test_utils.py
```diff
@@ -192,16 +192,14 @@ def required_extensions(self):
 
 if os.environ.get('TI_LITE_TEST', ''):
     _test_features = {
-        "dynamic_index": [TestParam(False, [])],
+        "dynamic_index": [TestParam(True, [])],
```
lin-hitonami, line 195: Do we still need these code since it's automatically on?

strongoier: Nope. I had an incorrect assumption that `_test_features` had to be non-empty :-)

Fixed with only one commented line left to serve as an example for potential future use.

lin-hitonami (approved):
LGTM

PR #7072, opened 2023-01-05
[lang] Free ndarray memory when it's GC-ed in Python

### LINKED ISSUE #5048: Ndarray lifetime management
Current ndarray implementation doesn't respect python. For example if you do `del ndarray1` only the ndarray object is deleted but the underlying memory is not freed. (managed by Program) 

Ideally we'd want ndarray memory to be freed either when python GCs it or C++ Program gets destructed, whichever comes first. 

- Ndarray cannot be managed solely by Python (like holding a shared pointer to the Device). Otherwise `ti.reset()` won't be able to destruct `Device` object as it does before. 
- Ndarray cannot be solely managed by C++ `Program` class. Otherwise the underlying memory of ndarray won't be freed even after user calls `del ndarray`. 

We gave this a try but had to revert it in https://github.com/taichi-dev/taichi/pull/5019. Need to understand the root cause of the bug first.

ailzhang:
Issue: fixes #5048 

### Brief Summary
  Ndarray's memory allocation is managed by Taichi and Python can control this via Taichi indirectly.
  For example, when an ndarray is GC-ed in Python, it signals Taichi to free its memory allocation. 
  But Taichi will make sure **no pending kernels to be executed needs the ndarray** before it actually 
  frees the memory.

### DIFF
--- a/taichi/python/export_lang.cpp
+++ b/taichi/python/export_lang.cpp
@@ -432,8 +432,7 @@ void export_lang(py::module &m) {
           },
           py::arg("dt"), py::arg("shape"),
           py::arg("layout") = ExternalArrayLayout::kNull,
-          py::arg("zero_fill") = false, py::return_value_policy::reference,
-          py::return_value_policy::reference)
+          py::arg("zero_fill") = false, py::return_value_policy::reference)
       .def("delete_ndarray", &Program::delete_ndarray)
       .def(
           "create_texture",
--- a/taichi/python/export_lang.cpp
+++ b/taichi/python/export_lang.cpp
@@ -432,7 +432,9 @@ void export_lang(py::module &m) {
           },
           py::arg("dt"), py::arg("shape"),
           py::arg("layout") = ExternalArrayLayout::kNull,
-          py::arg("zero_fill") = false, py::return_value_policy::reference)
+          py::arg("zero_fill") = false, py::return_value_policy::reference,
+          py::return_value_policy::reference)
+      .def("delete_ndarray", &Program::delete_ndarray)
       .def(
           "create_texture",
           [&](Program *program, const DataType &dt, int num_channels,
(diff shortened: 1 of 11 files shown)


taichi/python/export_lang.cpp
```diff
@@ -463,7 +463,9 @@ void export_lang(py::module &m) {
           },
           py::arg("dt"), py::arg("shape"),
           py::arg("layout") = ExternalArrayLayout::kNull,
-          py::arg("zero_fill") = false, py::return_value_policy::reference)
+          py::arg("zero_fill") = false, py::return_value_policy::reference,
+          py::return_value_policy::reference)
```
lin-hitonami, line 467: `py::return_value_policy::reference` repeats twice, what does this mean?

ailzhang: oops it's a typo I think...

strongoier (commented):
The tests may be enhanced to verify whether the memory is actually freed.

ailzhang:
> The tests may need some enhancement to verify whether the memory is actually freed.

Nice catch! Yea I've updated the test so that it allocates a lot of memory repeatedly and would've failed if memories were never freed. 

Confirmed this test fails on current master branch and works on this PR, the only quirk is that I also discovered that our ndarray memory on CPU is not properly handled/deallocated so had to exclude cpu backend in this test. Will send a followup PR to fix that separately. PTAL, thanks!

tests/python/test_ndarray.py
```diff
@@ -700,3 +700,28 @@ def test_ndarray_init_as_zero():
...
+    def test():
+        z = ti.ndarray(float, (8192, 8192))
+
+    for i in range(300):
+        test()
+
+
+@test_utils.test(arch=supported_archs_taichi_ndarray)
+def test_ndarray_simple_alloc():
+    z = ti.ndarray(float, (8192))
```
strongoier, line 727:
```suggestion
```
This test is unnecessary.

strongoier (approved):
LGTM!

PR #7089, opened 2023-01-06
[refactor] Remove dependencies on Program::this_thread_config() in gfx::run_codegen

PGZXB:
Issue: #7002

### DIFF
--- a/taichi/cache/gfx/cache_manager.cpp
+++ b/taichi/cache/gfx/cache_manager.cpp
@@ -107,8 +107,11 @@ namespace gfx {
 CacheManager::CacheManager(Params &&init_params)
     : mode_(init_params.mode),
       runtime_(init_params.runtime),
+      compile_config_(*init_params.compile_config),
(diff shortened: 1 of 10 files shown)


taichi/cache/gfx/cache_manager.cpp
```diff
@@ -161,7 +165,8 @@ CompiledKernelData CacheManager::load_or_compile(CompileConfig *config,
     spirv::lower(kernel);
     return gfx::run_codegen(kernel, runtime_->get_ti_device()->arch(),
                             runtime_->get_ti_device()->get_caps(),
-                            compiled_structs_);
+                            compiled_structs_,
+                            config->external_optimization_level > 0);
```
bobcao3, line 169: Can we get the full compile config into run_codegen instead of a single flag?

bobcao3: in case we want to add more backend options / configs down the road

bobcao3 (approved):
LGTM

PR #7102, opened 2023-01-09
[ci] More Windows git hacks

feisuzhu:
Issue: #

### Brief Summary

Sometimes we can see `filename too long` errors in `actions/checkout` step, with the added `core.longpaths=true` we can (hopefully) solve this.

Still didn't figure out the exact reason why checked-out repos go corrupt, but this seems a reasonable candidate.

2023-01-09-151745_1951x428_scrot


ailzhang (approved):
nit: would you mind updating the description to explain why setting core.longpath may help ;)? Thanks!

PR #7104, opened 2023-01-09
[Lang] Remove deprecated packed switch

strongoier:
### Brief Summary

This is a follow-up PR of #7125 to fully remove the `packed` switch. Changes to `ArithmeticInterpretor` are included to pass the `ScalarPointerLowererTest.Basic` test, which now switches running mode (packed) and generates different IR.


ailzhang (approved):
Epic!

PR #7107, opened 2023-01-10
[metal] Update Metal RHI impl & add support for shared arrays

bobcao3:
1. Remove most of the `TI_XXX` macros from Metal RHI impl
2. Utilize some features from `impl_support.h`
3. Enable shared array support on all SPIR-V backends


PENGUINLIONG (approved):
lgtm, but what's npy_models/lego.npy tho?

bobcao3:
> lgtm, but what's npy_models/lego.npy tho?

ah..... I ran the taichi ngp example and it put the downloaded stuff there...

PR #7110, opened 2023-01-10
[bug] Unify error message matching with/without validation layers for CapiTest.FailMapDeviceOnlyMemory

jim19930609:
Issue: #


ailzhang (approved):
Thanks!

PR #7113, opened 2023-01-10
[Error] Raise errors when using metal sparse

lin-hitonami:
Issue: #


bobcao3 (approved):
LGTM! This is for v1.4.0 right?

Issue #7111, opened 2023-01-10
inspect module cannot find source in interactive interpreter (v1.4.0 blocker)


neozhaoliang:
After removing the `sourceinspect` dependency in v1.4.0, users cannot run Taichi in the interactive interpreter:
```
$ python
>>> import taichi
>>> ti.init()
>>> @ti.kernel
>>> def test():
...     pass
>>> test()
```
As the interactive interpreter is widely used in some illustrating cases, we should better add this functionality back.

PR #7114, opened 2023-01-10
[Lang] Fix cannot use taichi in REPL

neozhaoliang:
This PR fixes the issue: #7111


python/taichi/lang/_wrap_inspect.py
```diff
@@ -1,48 +1,80 @@
...
 
     return None
 
 
 def _blender_findsource(obj):
     try:
         import bpy  # pylint: disable=import-outside-toplevel
-    except ImportError:
-        raise IOError('Not in Blender environment!')
+    except：
```
lin-hitonami, line 58:
```suggestion
    except:
```

lin-hitonami (approved):
Thanks!

PR #7124, opened 2023-01-11
[lang] Remove redundant TensorType expansion for function returns

jim19930609:

### DIFF
--- a/taichi/transforms/type_check.cpp
+++ b/taichi/transforms/type_check.cpp
@@ -196,9 +196,17 @@ class TypeCheck : public IRVisitor {
   }
 
   void visit(UnaryOpStmt *stmt) override {
-    stmt->ret_type = stmt->operand->ret_type;
+    auto operand_type = stmt->operand->ret_type;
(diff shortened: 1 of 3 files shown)


taichi/transforms/type_check.cpp
```diff
@@ -196,9 +196,17 @@ class TypeCheck : public IRVisitor {
   }
 
   void visit(UnaryOpStmt *stmt) override {
-    stmt->ret_type = stmt->operand->ret_type;
+    auto operand_type = stmt->operand->ret_type;
```
strongoier, line 199: Are these changes related to this PR?

jim19930609: Yup, we do have some test cases with `real_matrix_scalarize=False` such as `test_matrix_return_real_matrix`. These tests used to work because of the scalarization at Python level, so after removing the `expand_exprs()` we'll have to fix the issues with CodeGen and runtime.

strongoier (approved):

PR #7125, opened 2023-01-11
[Lang] Raise errors when using the packed switch

strongoier:
### Brief Summary

The `packed` switch of `ti.init()` should be removed according to the [deprecation notice](https://github.com/taichi-dev/taichi/releases/tag/v1.3.0). However, to avoid a large changeset immediately before release (#7104), let's raise errors first and get that in later.


lin-hitonami (approved):
LGTM

PR #7126, opened 2023-01-11
[Misc] Strictly check ndim with external array

turbo0628:
Issue: #6572 

* Check external ndarray total dim against type annotation
* Error out when element shape mismatch
* Refine error messages

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -420,8 +420,7 @@ class TaichiCallableTemplateMapper:
                     if len(shape) != anno.dtype.ndim + anno.ndim:
                         raise ValueError(
                             f"Invalid argument into ti.types.ndarray() - required array has ndim={anno.ndim} element_dim={anno.dtype.ndim}, "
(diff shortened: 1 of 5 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -416,11 +416,24 @@ def extract_arg(arg, anno):
             shape = tuple(shape)
             element_shape = ()
             if isinstance(anno.dtype, MatrixType):
-                if len(shape) < anno.dtype.ndim:
-                    raise ValueError(
-                        f"Invalid argument into ti.types.ndarray() - required element_dim={anno.dtype.ndim}, "
-                        f"but the argument has only {len(shape)} dimensions")
+                if anno.ndim is not None:
```
ailzhang, line 419: Can we add a few tests to test against these errors each? Thanks!

turbo0628: Done!

ailzhang (approved):
Awesome thanks!

PR #7128, opened 2023-01-11
[cpu] Reuse VirtualMemoryAllocator for CPU ndarray memory allocation

### LINKED ISSUE #7119: [bug] Ndarray on CPU are not handled by caching allocator
It's listed as TODO here https://github.com/taichi-dev/taichi/blob/master/taichi/rhi/cpu/cpu_device.cpp#L33. 
Without this being implemented, ndarray will always allocate new memories instead of reusing freed cached memory blocks. 
Found this in #7072

ailzhang:
Fixes #7119

### Brief Summary
It seems unnecessary for CPU to use runtime jit memory allocation for ndarray, let's see if anything breaks...


lin-hitonami (approved):
LGTM

PR #7137, opened 2023-01-11
[metal] Retain Metal commandBuffers & build command buffers directly

bobcao3:
Issue: #

### Brief Summary

Changed the command buffer building to directly retaining a command buffer and encoding commands into it (instead of building up a list of lambda functions)

### DIFF
--- a/cmake/TaichiCXXFlags.cmake
+++ b/cmake/TaichiCXXFlags.cmake
@@ -55,13 +55,13 @@ else()
     # [Global] CXX compilation option to enable all warnings.
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall ")
 
-# Due to limited CI coverage, -Werror is only turned on with Clang-compiler for now.
-if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
-    if (NOT ANDROID) # (penguinliong) Blocking builds on Android.
-        # [Global] CXX compilation option to treat all warnings as errors.
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror ")
+    # Due to limited CI coverage, -Werror is only turned on with Clang-compiler for now.
+    if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+        if (NOT ANDROID) # (penguinliong) Blocking builds on Android.
+            # [Global] CXX compilation option to treat all warnings as errors.
+            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror ")
+        endif()
     endif()
-endif()
 
     # [Global] By default, CXX compiler will throw a warning if it decides to ignore an attribute, for example "[[ maybe unused ]]".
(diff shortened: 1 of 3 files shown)


cmake/TaichiCXXFlags.cmake
```diff
@@ -79,14 +79,13 @@ endif()
...
 
-    # CLANG_VERSION_MAJOR is set in check_clang_version
-    if (${CLANG_VERSION_MAJOR} VERSION_GREATER_EQUAL 15)
-      # [Global] FIXME: Newly introduced flag in clang-15
+    check_cxx_compiler_flag("-Wno-unqualified-std-cast-call" CXX_HAS_Wno_unqualified_std_cast_call)
+    if (${CXX_HAS_Wno_unqualified_std_cast_call})
       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unqualified-std-cast-call ")
     endif()
 
-    if (${CLANG_VERSION_MAJOR} VERSION_GREATER_EQUAL 13)
```
bobcao3, line 88: This is the clang compiler version for the runtime, it might not be the host compiler!

bobcao3: Also AppleClang take their branch cuts at a different time from main line Clang. There are feature mismatches

taichi/rhi/metal/metal_device.mm
```diff
@@ -179,9 +179,13 @@
   return *this;
 }
 
-MetalCommandList::MetalCommandList(const MetalDevice &device)
-    : device_(&device) {}
-MetalCommandList::~MetalCommandList() {}
+MetalCommandList::MetalCommandList(const MetalDevice &device,
+                                   MTLCommandQueue_id cmd_queue)
+    : device_(&device) {
+  cmdbuf_ = [[cmd_queue commandBuffer] retain];
```
PENGUINLIONG, line 185: This should be guarded by `@autoreleasepool` I suppose?

taichi/rhi/metal/metal_device.mm
```diff
@@ -315,24 +319,18 @@
...
-    MTLCommandBuffer_id cmdbuf = [[mtl_command_queue_ commandBuffer] retain];
-    for (auto &command : cmdlist2->pending_commands_) {
-      command(cmdbuf);
-    }
-    cmdlist2->pending_commands_.clear();
-
-    [cmdbuf commit];
-    pending_cmdbufs_.emplace_back(cmdbuf);
-  }
+  MTLCommandBuffer_id cmdbuf = [cmdlist2->finalize() retain];
```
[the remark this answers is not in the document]
bobcao3: I think I do? `MetalCommandList` might be deallocated when we execute sync, so I need to add a reference to the object. When we then do the command sync (or when we release the pending cmdbuf vector) we can release these references.

taichi/rhi/metal/metal_device.mm
bobcao3: I guess if I do `[cmd_queue commandBuffer]` it returns a object with references already incremented once? If that's the case and if we are releasing manually in dtor I don't think this should be in autorelease?

PENGUINLIONG (approved):

PR #7138, opened 2023-01-12
[lang] Sign python library for Apple M1

PENGUINLIONG:
Apple requires any shared library built for Apple Silicon to be codesigned with a local development certificate. This certificate is generated by Xcode automatically.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -177,9 +177,8 @@ def sign_development_for_apple_m1():
         try:
             for path in glob.glob("python/taichi/_lib/core/*.so"):
                 print(f"signing {path}..")
-                subprocess.check_call([
-                    'codesign', '--force', '--deep', '--sign', '-', path
-                ])
+                subprocess.check_call(
+                    ['codesign', '--force', '--deep', '--sign', '-', path])
         except:
             print("cannot sign python shared library for macos arm64 build")
 
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,7 @@ import multiprocessing
 import os
 import platform
 import shutil
+import subprocess
 import sys
 from distutils.command.clean import clean
 from distutils.dir_util import remove_tree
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -166,6 +167,22 @@ def cmake_install_manifest_filter(manifest_files):
...
+    """
+    Apple enforces codesigning for arm64 targets even for local development
+    builds. See discussion here:
+        https://github.com/supercollider/supercollider/issues/5603
+    """
+    if sys.platform == "darwin" and platform.machine() == "arm64":
+        try:
+            subprocess.check_call([
+                'codesign', '--force', '--deep', '--sign', '-',
+                'python/taichi/_lib/core/taichi_python.cpython-310-darwin.so'
```
ailzhang, line 180: hmmm do we hardcode 310 here?

bobcao3 (approved):
LGTM

setup.py
```diff
@@ -166,6 +167,22 @@ def cmake_install_manifest_filter(manifest_files):
...
 
+def sign_development_for_apple_m1():
+    """
+    Apple enforces codesigning for arm64 targets even for local development
+    builds. See discussion here:
+        https://github.com/supercollider/supercollider/issues/5603
+    """
+    if sys.platform == "darwin" and platform.machine() == "arm64":
+        try:
+            for path in glob.glob("python/taichi/_lib/core/*.so"):
```
ailzhang, line 178: hmmm this won't work if you ship the .so right? It only signs the copy in your local directory but not the one "installed" or "saved in .whl"?

PENGUINLIONG: The thing is that we have to sign with *our official certificate in the future* if we want to distribute this `.so` because any local certificate is for local use only. If it's been downloaded from the internet then Apple would sigkill dyld when the user attempts to load it. I have a personal developer account so I can sign with my certificate to workaround this at least for the 1.4 release. But we ultimately have to ship it with Taichi's official account in the future, which we don't have for now. This PR is setup just for any local build to work.

ailzhang: I see, yea makes sense, would you mind opening an issue to track (like we need an official account to sign binaries) ?Thanks!

ailzhang (approved):
Thanks!

PR #7142, opened 2023-01-12
[aot] Support texture arguments for AOT kernels

jim19930609:
Issue: #

### DIFF
--- a/python/taichi/aot/utils.py
+++ b/python/taichi/aot/utils.py
@@ -7,7 +7,7 @@ from taichi.types.annotations import template
 from taichi.types.ndarray_type import NdarrayType
 from taichi.types.texture_type import TY_CH2FORMAT, RWTextureType, TextureType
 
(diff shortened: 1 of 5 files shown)


python/taichi/aot/utils.py
```diff
@@ -7,7 +7,7 @@
 from taichi.types.ndarray_type import NdarrayType
 from taichi.types.texture_type import TY_CH2FORMAT, RWTextureType, TextureType
 
-template_types = (NdarrayType, TextureType, template)
+template_types = (NdarrayType, RWTextureType, TextureType, template)
```
ailzhang, line 10: Is this accidentally included?

ailzhang (approved):
Thanks!

python/taichi/aot/utils.py
jim19930609: Ah this is by mistake, thanks!

YuCrazing (approved):
LGTM, Thanks!

PR #7143, opened 2023-01-12
[lang] Interface refactors for MatrixType and VectorType

jim19930609:

### Brief Summary
1. Removed redundant interface `fill_with_scalar()`
2. Refactored `cast()` into internal interface `_instantiate()`
3. MatrixType/VectorType will only return `ti.Matrix` in python_scope, while `Expr(TensorType)` in taichi_scope

After this PR, both MatrixType and VectorType should have single public interface `__call__()`


python/taichi/lang/matrix.py
```diff
@@ -1476,21 +1479,11 @@ def __call__(self, *args):
             elif isinstance(x, np.ndarray):
                 entries += list(x.ravel())
             elif isinstance(x, Matrix):
-                entries += x.entries
+                entries += [x(i, j) for j in range(x.m) for i in range(x.n)]
```
strongoier, line 1482: Why do we change this?

jim19930609: Ah just realized `Matrix::entries` is already flattened. Thanks!

strongoier (approved):

PR #7146, opened 2023-01-12
[ci] Run test when pushing to rc branches

lin-hitonami:
We should directly push the cherry-picks to the rc branches instead of merging it using PRs.


feisuzhu (approved):
LGTM

PR #7149, opened 2023-01-12
[gui] Fix particle size limits

bobcao3:
Related: #6999

image

### Brief Summary

Change particle rendering from point rasterization to point instanced quads, so that there's no longer a size limit. Screenshot came from macOS.


taichi/ui/backends/vulkan/renderables/particles.cpp
```diff
@@ -69,15 +69,32 @@ void Particles::init_particles(AppContext *app_context,
...
 }
 
+void Particles::record_this_frame_commands(CommandList *command_list) {
+  command_list->bind_pipeline(pipeline_.get());
+  command_list->bind_raster_resources(raster_state_.get());
+  command_list->bind_shader_resources(resource_set_.get());
+
+  // We draw num_particles * 6, 6 forms a quad
+  if (indexed_) {
+    command_list->draw_indexed_instance(6, config_.draw_index_count,
```
bobcao3, line 89: hmmmmmmm.... why would "indexed" particles make any sense? I need to double check whether GGUI API provides indexed version for particles.

taichi/ui/backends/vulkan/renderable.h
```diff
@@ -42,6 +42,7 @@ struct RenderableConfig {
       taichi::lang::TopologyType::Triangles};
   taichi::lang::PolygonMode polygon_mode{taichi::lang::PolygonMode::Fill};
   VertexAttributes vbo_attrs{VboHelpers::all()};
+  bool point_instancing{false};
```
AmesingFlank, line 45: How about just calling this `vertex_input_rate_instance`? So that the meaning of this flag is more obvious (and generic).

AmesingFlank (approved):
LGTM!

jim19930609 (approved):
Thanks!

Issue #6114, opened 2022-09-20
Changelog should be generated before tagging latest version


strongoier:
In the release of v1.1.3, the workflow generates an incomplete changelog (https://github.com/taichi-dev/taichi/actions/runs/3087352313/jobs/4993540670):
```
Highlights: Full changelog: - [bug] [gui] Fix a bug of drawing mesh instacing that cpu/cuda objects have an offset when copying to vulkan object (#6028) (by **Mocki**)
```
It seems that the latest version has been tagged before changelog generation. However, the setting looks correct: https://github.com/taichi-dev/taichi/blob/02822e70d1100b1adca9bc3796f7cee9408053fa/.github/workflows/release.yml#L454-L472

There must be something wrong. If we don't fix it, we will have to locally generate the changelog and manually paste it immediately after each release.

PR #7152, opened 2023-01-13
[ci] Use make_changelog.py to generate the full changelog

lin-hitonami:
Issue: #6114

### Brief Summary
The changelog automatically generated by GitHub only shows the PR list related to the branch, and the cherry-picked commits are not inside the list. So we should use our commit-based `make_changelog.py` to generate the full changelog.


ailzhang (approved):
@lin-hitonami would you mind adding a bit context in the pr description about why we add it back? Thanks!

PR #7156, opened 2023-01-13
[mac] Add .dylib and .cmake to built wheel

ailzhang:
Issue: #

### Brief Summary

Adding these to the built python package. 
Screenshot 2023-01-13 at 6 32 32 PM

cc: @PENGUINLIONG however the built wheels are with LLVM on so we might need to write a separate CI job just to build a c_api.so with desired configuration for release?


PENGUINLIONG (approved):
lol thx

PR #7159, opened 2023-01-13
[refactor] Remove dependencies on Program::this_thread_config() in irpass::constant_fold

PGZXB:
Issue: #7002 

### Brief Summary
1. The dependencies on `Program::this_thread_config()` in `irpass::constant_fold` were removed;
2. The race condition of `Program::config` (concurrent write) was killed, so we can remove the multi-thread version `Program::this_thread_config()`. I will do it in next PR;
3. The `Program::compile(Kernel *kernel)` was refactored to `Program::compile(const CompileConfig &compile_config, Kernel *kernel)`, which is a temporary solution because I will introduce new classes to provide compilation interfaces (see #7002 for more information).

### DIFF
--- a/taichi/cache/gfx/cache_manager.h
+++ b/taichi/cache/gfx/cache_manager.h
@@ -37,7 +37,8 @@ class CacheManager {
 
   explicit CacheManager(Params &&init_params);
 
-  CompiledKernelData load_or_compile(CompileConfig *config, Kernel *kernel);
+  CompiledKernelData load_or_compile(const CompileConfig *config,
+                                     Kernel *kernel);
   void dump_with_merging() const;
   void clean_offline_cache(offline_cache::CleanCachePolicy policy,
                            int max_bytes,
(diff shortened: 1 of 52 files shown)


taichi/cache/gfx/cache_manager.h
```diff
@@ -37,7 +37,8 @@ class CacheManager {
 
   explicit CacheManager(Params &&init_params);
 
-  CompiledKernelData load_or_compile(CompileConfig *config, Kernel *kernel);
+  CompiledKernelData load_or_compile(const CompileConfig *config,
```
lin-hitonami, line 40: Why is the argument sometimes `CompileConfig *` and sometimes `CompileConfig &`? Can we unify them to `CompileConfig &`?

PGZXB: History reason lol.
I will fix all `const CompileConfig *` to `const CompileConfig &` in future, which is part of #7002.

image

lin-hitonami (approved):
LGTM!

PR #7161, opened 2023-01-14
[doc] Update accelerate_python.md to use ti.max

tao-j:
newer version of ti deprecates `max`

Issue: #


tao-j:
https://docs.taichi-lang.org/docs/cloth_simulation
also need to updated min to ti.min.

I think add more commits now will trigger another CI run, and it seems that you have a long pipeline

neozhaoliang (approved):
Thank you!

PR #7162, opened 2023-01-15
[aot] C-API C++ wrapper feature completion and CMake find package infra

PENGUINLIONG:
Issue: #

### Brief Summary

- Added wrapping C++ function to `ti_get_version` and `ti_get_last_error`.
- Added `FindTaichi.cmake` for easy integration of Taichi Runtime in external projects.
- Fixed various compilation issue on C++11.
- Removed unused archs from C-API.
- Allowed CMake-build (without using setup.py) Taichi to be versioned


taichi/common/logging.cpp
```diff
@@ -59,7 +59,7 @@ Logger::Logger() {
   console_ = spdlog::android_logger_mt("android", "taichi");
   console_->flush_on(spdlog::level::trace);
 #else
-  console_ = spdlog::stdout_color_mt("console");
+  console_ = spdlog::stderr_color_mt("stderr");
```
PENGUINLIONG, line 62: The log contaminate stdout which makes it difficult to CI external projects

PR #7163, opened 2023-01-16
[gui] Update circles rendering to use quads

bobcao3:
This removes the size limit on circles. This works in a similar manner as the particle rendering size limits.


k-ye (approved):
LGTM!

PR #7165, opened 2023-01-16
[aot] Fixed ti_get_last_error signature

PENGUINLIONG:
Issue: #

### Brief Summary

Initially `ti_get_last_error` returns `TI_EROR_TRUNCATED` to hint the user that the message output is not complete (because the buffer size is insufficient), but this 'non-failing' error code has been removed in later commits. So at this point this API doesn't have a reliable way to know whether the error message is complete. This is considered problematic.


ailzhang (approved):
lgtm!

PR #7168, opened 2023-01-16
[aot] FindTaichi CMake module to help outside project integration

PENGUINLIONG:
Issue: #

### Brief Summary

This PR introduces a CMake module `FindTaichi.cmake` to help external project to integrate Taichi Runtime into their applications following the CMake convention to minimize the lines of codes they need in their `CMakeLists.txt`.

The following is the integration CMake snippet:

```cmake
# We need to set `FindTaichi.cmake` in module path to `find_package(taichi)`.
set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../cmake)
find_package(taichi REQUIRED)

add_executable(${EXAMPLE_NAME} main.cpp)
target_link_libraries(${EXAMPLE_NAME} PRIVATE taichi::runtime)

# Copy redistributed libraries (`.so` and `.dll`) next to the executable.
add_custom_command(
    TARGET ${EXAMPLE_NAME}
    PRE_BUILD
    COMMENT "-- Copy redistributed libraries to output directory (if different): ${taichi_REDIST_LIBRARIES}"
    COMMAND ${CMAKE_COMMAND} -E copy_if_different ${taichi_REDIST_LIBRARIES} "$<TARGET_FILE_DIR:${EXAMPLE_NAME}>")
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,8 +8,8 @@ project(taichi)
 
 include("cmake/utils.cmake")
 
-message(WARNING "It seems that you are running cmake manually, which may cause issues. Please use setup.py to build taichi from source, see https://docs.taichi-lang.org/docs/dev_install for more details.")
 if (NOT DEFINED TI_VERSION_MAJOR)
+    message(WARNING "It seems that you are running cmake manually, which may cause issues. Please use setup.py to build taichi from source, see https://docs.taichi-lang.org/docs/dev_install for more details.")
     file(READ "${CMAKE_CURRENT_LIST_DIR}/version.txt" TI_VERSION_LITERAL)
     string(REGEX MATCH "v([0-9]+)\\.([0-9]+)\\.([0-9]+)" TI_VERSION_LITERAL ${TI_VERSION_LITERAL})
     set(TI_VERSION_MAJOR ${CMAKE_MATCH_1})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,11 +8,13 @@ project(taichi)
 
 include("cmake/utils.cmake")
 
+message(WARNING "It seems that you are running cmake manually, which may cause issues. Please use setup.py to build taichi from source, see https://docs.taichi-lang.org/docs/dev_install for more details.")
 if (NOT DEFINED TI_VERSION_MAJOR)
-    message(WARNING "It seems that you are running cmake manually, which may cause issues. Please use setup.py to build taichi from source, see https://docs.taichi-lang.org/docs/dev_install for more details.")
-    set(TI_VERSION_MAJOR 0)
-    set(TI_VERSION_MINOR 0)
-    set(TI_VERSION_PATCH 0)
+    file(READ "${CMAKE_CURRENT_LIST_DIR}/version.txt" TI_VERSION_LITERAL)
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -8,11 +8,13 @@ project(taichi)
 
 include("cmake/utils.cmake")
 
+message(WARNING "It seems that you are running cmake manually, which may cause issues. Please use setup.py to build taichi from source, see https://docs.taichi-lang.org/docs/dev_install for more details.")
```
ailzhang, line 11: nit: why do we take this out of the if branch? ;) The old logic was: if we run cmake from setup.py it'll have `TI_VERSION_MAJOR` set so it won't trigger this warning, not sure if it's still applicable.

ailzhang (commented):
lgtm! leaving this to @jim19930609 for a final check ;)

CMakeLists.txt
PENGUINLIONG: oof it's a miss

cmake/TaichiCAPI.cmake
```diff
@@ -32,14 +45,17 @@ endif()
 
 if (TI_WITH_OPENGL)
   list(APPEND C_API_SOURCE "c_api/src/taichi_opengl_impl.cpp")
+  list(APPEND C_API_PUBLIC_HEADERS "c_api/include/taichi/taichi_opengl.h")
 endif()
 
 if (TI_WITH_METAL)
   list(APPEND C_API_SOURCE "c_api/src/taichi_metal_impl.mm")
+  #list(APPEND C_API_PUBLIC_HEADERS "c_api/include/taichi/taichi_metal.h")
```
jim19930609, line 53: uncomment this?

c_api/cmake/FindTaichi.cmake
```diff
@@ -0,0 +1,172 @@
...
+
+# Support `IN_LIST` in CMake `if` command.
+if(POLICY CMP0057)
+    cmake_policy(SET CMP0057 NEW)
+endif()
+
+# Find TiRT in the installation directory. The installation directory is
+# traditionally specified by an environment variable
+# `TAICHI_C_API_INSTALL_DIR`.
+if(NOT EXISTS ${TAICHI_C_API_INSTALL_DIR})
```
jim19930609, line 63: Will `FindTaichi.cmake` get shipped along with C-API? If so,  ${TAICHI_C_API_INSTALL_DIR} should be a relative path to `FindTaichi.cmake` (like `./..`?)

jim19930609 (approved):

cmake/TaichiCAPI.cmake
PENGUINLIONG: We don't have this file atm, lol.

c_api/cmake/FindTaichi.cmake
PENGUINLIONG: This one should be set up by the user. This FindTaichi.cmake is expected to be copied by the user to their repositories, in a folder called `cmake`. Then, when the user builds their project, CMake can find that package using this script following the rules in `https://cmake.org/cmake/help/v3.0/command/find_package.html#command:find_package`.

PR #7170, opened 2023-01-16
[aot] Replace std::exchange with local implementation for C++11

PENGUINLIONG:
Issue: #

### Brief Summary

Just noticed that `std::exchange` doesn't exists in C++11.


ailzhang (approved):
Thanks!

PR #7173, opened 2023-01-16
[Doc] Update debugging.md

neozhaoliang:
This PR deletes the obsolete note because:

1. Taichi will drop the support for IDLE from v1.4.0.
2. print is now supported in Jupyter notebook.


lin-hitonami (approved):
LGTM!

PR #7183, opened 2023-01-17
[ci] Don't copy to build unless it's in CI

PENGUINLIONG:
Issue: #

### DIFF
--- a/cmake/TaichiCAPI.cmake
+++ b/cmake/TaichiCAPI.cmake
@@ -77,6 +77,7 @@ set_target_properties(${TAICHI_C_API_NAME} PROPERTIES
     LIBRARY_OUTPUT_DIRECTORY ${C_API_OUTPUT_DIRECTORY}
     ARCHIVE_OUTPUT_DIRECTORY ${C_API_OUTPUT_DIRECTORY})
 
+if ($ENV{TI_CI})
 if (${CMAKE_GENERATOR} MATCHES "^Visual Studio")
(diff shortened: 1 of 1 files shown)


cmake/TaichiCAPI.cmake
```diff
@@ -77,6 +77,7 @@ set_target_properties(${TAICHI_C_API_NAME} PROPERTIES
     LIBRARY_OUTPUT_DIRECTORY ${C_API_OUTPUT_DIRECTORY}
     ARCHIVE_OUTPUT_DIRECTORY ${C_API_OUTPUT_DIRECTORY})
 
+if ($ENV{TI_CI})
```
ailzhang, line 80: why do we want to remove this copy btw? ;)

PENGUINLIONG: it basically fails raw cmake builds out of source. it's kind of a problem if you want to keep multiple cross-compiled runtime build folders on a single machine at the same time. i understand that ccache can help saving build time when we shift from a target to another but the point is i have to keep multiple build folders alive at the same time.

ailzhang: hmmm then I'm a bit confused why these were added originally. cc: @bobcao3 would you mind giving some context here?

bobcao3: Without this CAPI tests will fail because it can't find its DLL files. Unless we want to modify our run tests script to adapt to multi config generator.

bobcao3: Also I don't get about the cross compiled part, do we support actual cross compiling？

PR #7186, opened 2023-01-17
[Doc] Rename tutorial doc

neozhaoliang:
The name tutorial is a bit misleading as it's not a tutorial on Taichi but on taichi in C++ apps. Also, we need to use the older name to make our doc deployment bot happy.


Olinaaaloompa (approved):
Thanks!

PR #7187, opened 2023-01-17
[example] Fix ti gallery close warning

neozhaoliang:
The terminal will report an error when users run the command `ti gallery` but close the window without choosing any example. This PR fixes this problem.


lin-hitonami (approved):
LGTM

PR #7196, opened 2023-01-17
[misc] Remove legacy latex generation

strongoier:
### Brief Summary

The legacy helper functions only work under specific environment and should no longer be part of this project.


ailzhang (approved):
Thanks!

PR #7199, opened 2023-01-17
[refactor] Program::this_thread_config() -> Program::compile_config()

PGZXB:
Issue: #7002 
Removed multi-thread version `Program::this_thread_config()` (see https://github.com/taichi-dev/taichi/pull/7159#issue-1532269774)

### DIFF
--- a/taichi/program/program.h
+++ b/taichi/program/program.h
@@ -92,12 +92,6 @@ class StructCompiler;
 class TI_DLL_EXPORT Program {
  public:
   using Kernel = taichi::lang::Kernel;
-  // We let every thread has its own config because the constant folding pass
-  // wants to change the CompileConfig so that it can compile the evaluator,
-  // but we don't want it to change the global config. We will refactor it
-  // later when we make Taichi thread-safe.
-  std::unordered_map<std::thread::id, CompileConfig> configs;
-  std::thread::id main_thread_id_;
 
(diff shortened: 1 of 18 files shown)


taichi/program/program.h
```diff
@@ -121,22 +115,9 @@ class TI_DLL_EXPORT Program {
...
-      configs[thread_id] = configs[main_thread_id_];
-      return configs[thread_id];
-    }
-    return configs[thread_id];
-  }
-
-  const CompileConfig &config() {
-    return configs[main_thread_id_];
+  // TODO(PGZXB): Return `const CompileConfig &` (blocked by some tests)
+  CompileConfig &compile_config() {
```
ailzhang, line 119: nit: is it possible that we have another method that returns a const reference, and make sure only the problematic tests use this non-const version? Thanks!

PGZXB: Maybe we can provide `Program::set_compile_config`(only for tests)?
p.s. And I will refactor those tests to remove the temporary `Program::set_compile_config`.

ailzhang: sg!

PGZXB: I found that the dependencies on the non-const version in tests can be removed directly, so I killed them in this PR and fixed the ` CompileConfig &compile_config()` to ` const CompileConfig &compile_config()`. Please review again.

ailzhang (approved):
Thanks!

PR #7209, opened 2023-01-18
[refactor] Split constructing and compilation of lang::Function

PGZXB:
Issue: #7002 

### Brief Summary
Removed dependencies on `Program::this_thread_config()` in `lang::Function` compilation (AST->IR part)
* Push off the compilation of `lang::Function`: Introduce the `irpass::compile_called_function(IRNode *root, const CompileConfig &config)`, which compiles the AST/IR of `Function`s called in `root` to the final IR.

### DIFF
--- a/taichi/ir/transforms.h
+++ b/taichi/ir/transforms.h
@@ -200,7 +200,7 @@ void ast_to_ir(const CompileConfig &config,
                Kernel &kernel,
                bool to_executable = true);
 
-void compile_called_functions(IRNode *ir, const CompileConfig &compile_config);
+void compile_taichi_functions(IRNode *ir, const CompileConfig &compile_config);
 }  // namespace irpass
 
 }  // namespace taichi::lang
--- a/taichi/ir/transforms.h
+++ b/taichi/ir/transforms.h
@@ -199,6 +199,8 @@ void compile_function(IRNode *ir,
 void ast_to_ir(const CompileConfig &config,
                Kernel &kernel,
                bool to_executable = true);
+
+void lower_called_functions(const CompileConfig &compile_config, IRNode *ir);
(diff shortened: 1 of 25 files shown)


taichi/program/function.h
```diff
@@ -9,6 +9,8 @@ class Program;
 
 class Function : public Callable {
  public:
+  enum class IRType { None, AST, InitialIR, OptimizedIR };
```
ailzhang, line 12: nit: might need better names here. cc: @strongoier

taichi/ir/transforms.h
```diff
@@ -199,6 +199,8 @@ void compile_function(IRNode *ir,
 void ast_to_ir(const CompileConfig &config,
                Kernel &kernel,
                bool to_executable = true);
+
+void compile_called_functions(IRNode *ir, const CompileConfig &compile_config);
```
ailzhang, line 203: nit: `compile_called_function` isn't very intuitive. how about `compile_taichi_function`?

taichi/program/function.h
strongoier: It seems OK for now :-)

taichi/transforms/compile_called_functions.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+ private:
+  const CompileConfig &compile_config_;
+};
+
+namespace irpass {
+
+void compile_taichi_functions(IRNode *ir, const CompileConfig &compile_config) {
+  TI_AUTO_PROF;
+  CompileCalledFunctions::run(ir, compile_config);
```
strongoier, line 45: nit: name mismatch?

ailzhang (approved):
Thanks!

PR #7214, opened 2023-01-19
[aot] Fix ndarray aot with information from type hints

### LINKED ISSUE #7172: [aot] [bug] aot add_kernel behavior when the kernel has ndarray args
There're a few things we can improve here:
- When the kernel type annotation is good enough for compilation, `add_kernel` should allow the following usage:
Screenshot from 2023-01-16 11-09-24
- Currently we have simply type match checking logic in `produce_injected_args` https://github.com/taichi-dev/taichi/blob/master/python/taichi/aot/utils.py#L41 and this can be unified with the logic in JIT runtime.

ailzhang:
Issue: fixes #7172 

### Brief Summary
Ideally we should reconstruct the dtype to the Tensortype from taichi_core instead of python ones but that can be a separate PR.

### DIFF
--- a/python/taichi/aot/utils.py
+++ b/python/taichi/aot/utils.py
@@ -1,7 +1,8 @@
 from taichi.lang._ndarray import ScalarNdarray
 from taichi.lang._texture import Texture
 from taichi.lang.exception import TaichiCompilationError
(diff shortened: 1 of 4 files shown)


PENGUINLIONG (approved):
lgtm, should we update the docs too tho?

ailzhang:
@PENGUINLIONG sure thing, will send a followup PR for it!

PR #7218, opened 2023-01-20
[gui] Force GGUI to go through host memory (nuking interops)

bobcao3:
Issue: #

Contains:
- [x] #7230
- [x] #7232
- [x] #7233
- [x] #7234
Merge those before this PR (to reduce the size of the change set)

### Brief Summary

Nuking all interops, in order to reduce code complexity of GGUI & provide path ways to other rendering frontends (e.g. metal)


k-ye:
nit: Do you plan to break this PR down?

bobcao3:
> nit: Do you plan to break this PR down?

I will put all the RHI changes / bug fixes in their own PR, and then the GGUI changes kind of have to be a big blob to pass all tests.

I think that change set would be around +750 lines.

I can seperate them through ghstack perhaps but individual ones won't pass CI.

PENGUINLIONG (approved):
lgtm, @ailzhang and @AmesingFlank might wanna double check this huge pr.

ailzhang (approved):
Thanks!

PR #7220, opened 2023-01-20
[examples] Remove dependency on `ti.u8` compute type for ngp

bobcao3:
### Brief Summary

Many backends support half / float16, but in comparison int8 has very pretty bad support in general. This change enables backends like `opengl` and `dx12` to run this demo (assuming the feature-enabling PR is also merged in those)


ailzhang (approved):
LGTM!

PR #7227, opened 2023-01-20
[refactor] Remove legacy num_bits and acc_offsets from AxisExtractor

strongoier:
### Brief Summary

This is a follow-up PR of #7104 to fully remove legacy code for the SNode padding behavior. Note that the experimental code of struct for on bitmasked SNode for SPIR-V is removed here because it assumes the existence of the padding behavior.


bobcao3 (approved):
LGTM

PR #7230, opened 2023-01-22
[opengl] Fix GLFW global context issue

bobcao3:
Related: #7218

Without this GGUI will mess with the OpenGL backend because both tries to initialize & destroy GLFW.

There is an issue tho with dependency. Which target should `rhi/window_system` belong to? Maybe we need a device API target that collects all the specific backends and generate a static lib with all backends and optional dependencies like GLFW? (Anyways, that will be a TODO for another PR)

### DIFF
--- a/taichi/rhi/window_system.cpp
+++ b/taichi/rhi/window_system.cpp
@@ -21,7 +21,7 @@ static void glfw_error_callback(int code, const char *description) {
   RHI_LOG_ERROR(buf.data());
 }
 
-bool glfwContextAcquire() {
+bool glfw_context_acquire() {
   std::lock_guard lg(glfw_state.mutex);
   if (glfw_state.glfw_ref_count == 0) {
     auto res = glfwInit();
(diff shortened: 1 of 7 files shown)


taichi/rhi/window_system.cpp
```diff
@@ -0,0 +1,60 @@
...
+static void glfw_error_callback(int code, const char *description) {
+  std::array<char, 1024> buf;
+  snprintf(buf.data(), buf.size(), "GLFW Error %d: %s", code, description);
+  RHI_LOG_ERROR(buf.data());
+}
+
+bool glfw_context_acquire() {
+  std::lock_guard lg(glfw_state.mutex);
+  if (glfw_state.glfw_ref_count == 0) {
+    auto res = glfwInit();
```
k-ye, line 27: How expensive are `glfwInit()` and `glfwTerminate()`. If they are costly, put these calls outside the mutex?

bobcao3: Because they are global states, they need to be serial

k-ye (approved):

PR #7233, opened 2023-01-25
[misc] Export DeviceAllocation into Python & support devalloc in field_info

bobcao3:
Part of #7218

### DIFF
--- a/taichi/program/field_info.h
+++ b/taichi/program/field_info.h
@@ -17,8 +17,9 @@ enum class FieldSource : int {
   TaichiCuda = 0,
   TaichiX64 = 1,
   TaichiVulkan = 2,
-  TaichiOpenGL = 3
-  // support np array / torch tensor in the future?
+  TaichiOpenGL = 3,
+  TaichiNDarray = 4,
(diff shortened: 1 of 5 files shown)


taichi/program/field_info.h
```diff
@@ -29,6 +30,7 @@ struct FieldInfo {
   DEFINE_PROPERTY(int, matrix_rows);
   DEFINE_PROPERTY(int, matrix_cols);
   DEFINE_PROPERTY(std::vector<int>, shape);
+  DEFINE_PROPERTY(uint64_t, num_elements);
   DEFINE_PROPERTY(FieldSource, field_source);
```
k-ye, line 34: Out of curiosity, when we get rid of the interops, can we remove all field stuff in GGUI? I'd assume they are converted to ndarrays first?

k-ye (approved):
LGTM!

taichi/program/field_info.h
bobcao3: That's the plan. FieldInfo will become basically a wrapped DeviceAllocation

PR #7236, opened 2023-01-26
[DOC] Update ndarray.md

vai9er:


neozhaoliang (approved):
LGTM!

Olinaaaloompa (approved):
Fix dangling modifiers. Also, we tend to use the present tense throughout even in conditional sentences to imply that something always happens, but we can discuss it later :)
Otherwise, it LGTM! Thanks!

PR #7238, opened 2023-01-28
[Lang] Add better error message for dynamic snode

neozhaoliang:
Why this PR:

Currently, when users allocate dynamic SNodes of dimension >=  2**31, the error message is not easy to understand:
```
   self.ptr.dynamic(axis[0], dimension, chunk_size, get_traceback()))
TypeError: dynamic(): incompatible function arguments. The following argument types are supported:
    1. (self: taichi._lib.core.taichi_python.SNode, arg0: taichi._lib.core.taichi_python.Axis, arg1: int, arg2: int, arg3: str) -> taichi._lib.core.taichi_python.SNode
```

This PR makes the error clear.

### DIFF
--- a/python/taichi/_snode/fields_builder.py
+++ b/python/taichi/_snode/fields_builder.py
@@ -98,10 +98,12 @@ class FieldsBuilder:
 
         if dimension >= 2**31:
             raise TaichiRuntimeError(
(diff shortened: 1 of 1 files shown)


lin-hitonami (approved):
LGTM

PR #7249, opened 2023-01-29
[llvm] Unify the llvm context of host and device

### LINKED ISSUE #7251: Unify the llvm context of host and device
**Describe the bug**
The `llvm_context_host_` and `llvm_context_device_` are never both used in the same backend, so we can unify them and only use a single `llvm_context_`.

lin-hitonami:
Issue: fixes #7251
The` llvm_context_host_` and `llvm_context_device_` are never both used in the same backend, so we can unify them and only use a single `llvm_context_`.

### DIFF
--- a/taichi/runtime/llvm/llvm_context.cpp
+++ b/taichi/runtime/llvm/llvm_context.cpp
@@ -95,6 +95,11 @@ TaichiLLVMContext::TaichiLLVMContext(CompileConfig *config, Arch arch)
     llvm::InitializeNativeTargetAsmParser();
 #endif
   } else if (arch == Arch::dx12) {
+    // FIXME: Must initialize these before initializing Arch::dx12
+    // because it uses the jit of CPU right now.
+    llvm::InitializeNativeTarget();
+    llvm::InitializeNativeTargetAsmPrinter();
+    llvm::InitializeNativeTargetAsmParser();
 #if defined(TI_WITH_DX12)
     LLVMInitializeDirectXTarget();
     LLVMInitializeDirectXTargetMC();
--- a/taichi/runtime/llvm/llvm_context.cpp
+++ b/taichi/runtime/llvm/llvm_context.cpp
@@ -100,12 +100,14 @@ TaichiLLVMContext::TaichiLLVMContext(const CompileConfig &config, Arch arch)
     llvm::InitializeNativeTarget();
     llvm::InitializeNativeTargetAsmPrinter();
     llvm::InitializeNativeTargetAsmParser();
+    /*
 #if defined(TI_WITH_DX12)
     LLVMInitializeDirectXTarget();
     LLVMInitializeDirectXTargetMC();
     LLVMInitializeDirectXTargetInfo();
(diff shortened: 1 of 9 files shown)


taichi/runtime/llvm/llvm_context.cpp
```diff
@@ -95,6 +95,11 @@ TaichiLLVMContext::TaichiLLVMContext(CompileConfig *config, Arch arch)
     llvm::InitializeNativeTargetAsmParser();
 #endif
   } else if (arch == Arch::dx12) {
+    // FIXME: Must initialize these before initializing Arch::dx12
```
ailzhang, line 98: OOC: could you explain a bit why it wasn't required before this PR? :P Thanks!

ailzhang (commented):
LGTM!

taichi/runtime/llvm/llvm_context.cpp
lin-hitonami: Because we initialized the TaichiLLVMContext of CPU before initializing the context of the backend specified by config.arch regardless of what the backend is before. We don't initialized it now if the backend is not CPU.

taichi-gardener:

ailzhang (approved):
Thanks!

taichi/runtime/llvm/llvm_context.cpp
```diff
@@ -95,12 +95,19 @@ TaichiLLVMContext::TaichiLLVMContext(const CompileConfig &config, Arch arch)
...
+    llvm::InitializeNativeTargetAsmPrinter();
+    llvm::InitializeNativeTargetAsmParser();
+    /*
 #if defined(TI_WITH_DX12)
     LLVMInitializeDirectXTarget();
     LLVMInitializeDirectXTargetMC();
     LLVMInitializeDirectXTargetInfo();
     LLVMInitializeDirectXAsmPrinter();
 #endif
+    */
```
lin-hitonami, line 110:
```suggestion
#if defined(TI_WITH_DX12)
    LLVMInitializeDirectXTarget();
    LLVMInitializeDirectXTargetMC();
    LLVMInitializeDirectXTargetInfo();
    LLVMInitializeDirectXAsmPrinter();
#endif
```

taichi/runtime/llvm/llvm_context.cpp
```diff
@@ -95,6 +95,11 @@ TaichiLLVMContext::TaichiLLVMContext(const CompileConfig &config, Arch arch)
     llvm::InitializeNativeTargetAsmParser();
 #endif
   } else if (arch == Arch::dx12) {
+    // FIXME: Must initialize these before initializing Arch::dx12
+    // because it uses the jit of CPU right now.
+    llvm::InitializeNativeTarget();
+    llvm::InitializeNativeTargetAsmPrinter();
+    llvm::InitializeNativeTargetAsmParser();
```
lin-hitonami, line 102:
```suggestion
    llvm::InitializeNativeTargetAsmParser();
    // The dx target is used elsewhere, so we need to initialize it too.
```

PR #7254, opened 2023-01-29
[misc] Update MSVC flags

bobcao3:
Update performance flags & enable LTO on MSVC

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -1,7 +1,9 @@
 function(target_enable_function_level_linking TARGET)
     if(APPLE)
         target_link_options(${TARGET} PRIVATE -Wl,-dead_strip)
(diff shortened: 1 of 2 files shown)


cmake/utils.cmake
```diff
@@ -1,7 +1,9 @@
 function(target_enable_function_level_linking TARGET)
     if(APPLE)
         target_link_options(${TARGET} PRIVATE -Wl,-dead_strip)
-    else() # LINUX AND WIN32
+    elseif(MSVC) # LINUX AND WIN32
```
jim19930609, line 4: comment should be "# WIN32 only"

jim19930609 (approved):
thanks~

jim19930609 (approved):

PR #7261, opened 2023-01-30
[ir] Get the offsets of elements in StructType

lin-hitonami:
This PR adds `get_struct_type_with_data_layout` to `ProgramImpl` to get a struct type (with the offset of each component) with a specified data layout. In LLVM-based backends, it uses `llvm::DataLayout` to get the layout of a struct type. 

This PR also adds `TensorType::get_element_offset` and `StructType::get_element_offset` to enable getting the offset of any component recursively inside a compound type.

* #7310
* #7262
* __->__ #7261


bobcao3:
Get element offset will not make sense without its layout context? Even with the same backend (say CUDA or SPIR-V), a struct might have different layouts in different storage (e.g. local vs global, shared memory, etc.)

lin-hitonami:
> Get element offset will not make sense without its layout context? Even with the same backend (say CUDA or SPIR-V), a struct might have different layouts in different storage (e.g. local vs global, shared memory, etc.)

The offset is currently only planned to be used when reading the returned struct and constructing the arguments of the Kernel, so it only have one layout per backend now. If we have other usages later, we can add the layout argument at that time.

bobcao3:
> 

This will be a problem. Real function returns are local, while in SPIR-V the actual whole kernel input struct and output struct will have different layout, and they are also different from the thread local layout.

bobcao3:
I think for this to work correctly, layout of a struct can only be queried after kernel compile, and the query needs to be based on not only the type but also usage. Or we can add two query just for kernel argument and returns?

lin-hitonami:
> I think for this to work correctly, layout of a struct can only be queried after kernel compile, and the query needs to be based on not only the type but also usage. Or we can add two query just for kernel argument and returns?

How about this: we add a field `layout` (with type `std::string` or an enum?) to `StructType` and offer a function to generate the `StructType` with another `layout`?

bobcao3:
I think we can have something similar to the LLVM layout string thing, hmmmm

bobcao3:
If you take a look at SPIRV codegen in the TinyIR section I have layout code for SPIRV std140 and 430, and the API for "layout context" . You can try to implement those two alignments into IR struct type.

lin-hitonami:
@bobcao3 I've added the function getting a struct of a specified layout in the ProgramImpl so that every backend can have its own implementation (in LLVM-based backends we use llvm::DataLayout to get the offsets). Does this satisfy your need?

bobcao3:
> @bobcao3 I've added the function getting a struct of a specified layout in the ProgramImpl so that every backend can have its own implementation (in LLVM-based backends we use llvm::DataLayout to get the offsets). Does this satisfy your need?

This should work, thanks!

bobcao3 (approved):
LGTM!

taichi/ir/type.cpp
```diff
@@ -84,6 +84,26 @@ std::string PrimitiveType::to_string() const {
   return data_type_name(DataType(const_cast<PrimitiveType *>(this)));
 }
 
+size_t PrimitiveType::type_size() {
```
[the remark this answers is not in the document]
lin-hitonami: Ok, I'll do it in another PR.

strongoier, line 87: Is https://github.com/taichi-dev/taichi/blob/4e6da3da971dbd630d2fb8878aea0f09530b8c4b/taichi/ir/type_utils.cpp#L30-L69 enough for this purpose?

taichi/ir/type.cpp
```diff
@@ -103,6 +103,10 @@ std::string TensorType::to_string() const {
   return s;
 }
 
+size_t TensorType::get_element_offset(int ind) const {
+  return data_type_size(element_) * ind;
```
strongoier, line 107: I'm a bit confused about why only the element offset of `TensorType` needs the `data_type_size()` function.

lin-hitonami: We calculate the offset of a `StructType` by calling `llvm::DataLayout::getStructLayout` because we don't know how it pads the struct type. For a nested struct type, we calculate it by adding the struct/tensor offsets of every depth. We don't need `data_type_size()` for struct type because the offsets of scalars are calculated by LLVM.

strongoier (approved):
LGTM!

PR #7262, opened 2023-01-30
[ir] Refactor result fetching

lin-hitonami:
This PR adds code that fetches the return values out of a StructType instead of uint64 arrays. The next PR will let the kernel return a StructType instead of a uint64 array.

Note that before the refactoring, f16 was converted to f32 before returning. However after the refactor, we convert the f16 to f32 on host after it is returned (in the next PR). Therefore, I added an implementation of f16-to-f32 conversion (from stackoverflow) to util/bits.cpp.

* #7310
* __->__ #7262
* #7261


taichi/runtime/llvm/llvm_runtime_executor.h
```diff
@@ -85,6 +85,15 @@ class LlvmRuntimeExecutor {
         fetch_result_uint64(i, result_buffer));
```
bobcao3, line 85: We can finally remove this API afterwards right? `fetch_result_uint64` also crept into `device.h` and should be removed imo

bobcao3 (approved):
LGTM!

taichi/runtime/llvm/llvm_runtime_executor.h
lin-hitonami: Yeah, after the refactoring of all backends completes.

strongoier (approved):

PR #7271, opened 2023-02-01
[Gui] GGUI use shader "factory" (GGUI rework n/N)

bobcao3:
### Brief Summary

Use shader factory system so that we will be able to create and destroy Renderable objects on-the-fly (next step for reducing complexity & removing reuse related code)

The ultimate goal here is to make Renderable immutable and thus fully switch to an immediate mode implementation


taichi/ui/backends/vulkan/app_context.h
```diff
@@ -37,13 +37,36 @@ class TI_DLL_EXPORT AppContext {
 
   AppConfig config;
 
+  taichi::lang::Pipeline *get_raster_pipeline(
```
k-ye, line 40: style nit: pack these params into a dedicated params struct, so that we don't run into risks of confusing param order && avoids changing the interface when we need to add more parameters

taichi/ui/backends/vulkan/app_context.h
```diff
@@ -37,13 +37,36 @@ class TI_DLL_EXPORT AppContext {
...
+      const std::string &frag_path,
+      const std::string &vert_path,
+      taichi::lang::TopologyType prim_topology,
+      bool depth = false,
+      taichi::lang::PolygonMode polygon_mode = taichi::lang::PolygonMode::Fill,
+      bool blend = true,
+      bool vbo_instanced = false);
+
+  taichi::lang::Pipeline *get_customized_raster_pipeline(
```
k-ye, line 49: what's the different between these two methods?

k-ye (approved):

taichi/ui/backends/vulkan/app_context.h
PENGUINLIONG: is this about vertex normal/color support?

PENGUINLIONG (approved):
lgtm

taichi/ui/backends/vulkan/app_context.h
bobcao3: yes, and currently used for line drawing because we use a compute shader to generate translated vertices and that vertex buffer is slimmer to save space

bobcao3: will provide comments

PR #7278, opened 2023-02-01
[rhi] Device upload readback functions

bobcao3:
Issue: #

### Brief Summary

Add device upload and readback functions, and use backend-specific fast path.

### DIFF
--- a/taichi/rhi/device.h
+++ b/taichi/rhi/device.h
@@ -516,6 +516,7 @@ enum class AllocUsage : int {
   Uniform = 2,
   Vertex = 4,
   Index = 8,
+  Upload = 16,
 };
 
 MAKE_ENUM_FLAGS(AllocUsage)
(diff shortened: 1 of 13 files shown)


taichi/rhi/opengl/opengl_device.cpp
```diff
@@ -546,10 +538,6 @@ DeviceAllocation GLDevice::allocate_memory(const AllocParams &params) {
     target_hint = GL_SHADER_STORAGE_BUFFER;
   } else if (params.usage && AllocUsage::Uniform) {
     target_hint = GL_UNIFORM_BUFFER;
-  } else if (params.host_write && params.host_read) {
```
ailzhang, line 549: nit: why do we remove these btw?

taichi/rhi/device.h
```diff
@@ -678,6 +679,18 @@ class TI_DLL_EXPORT Device {
         this->allocate_memory(params));
   }
 
+  virtual RhiResult upload_data(DevicePtr *device_ptr,
```
ailzhang, line 682: nit: docstrings ;)?

ailzhang (approved):
Thanks!!

taichi/rhi/device.h
bobcao3: Ah I forgot

taichi/rhi/opengl/opengl_device.cpp
bobcao3: They don't actually modify the hint right now.

PR #7285, opened 2023-02-02
[Bug] Fix caching same loop invariant global vars inside nested fors

### LINKED ISSUE #7263: Thread Local Storage Optimisation Bug
There issue provides a minimal repro example for the bug in #7255, and discuss why this is likely a bug with our TLS optimization.

Example:
```py
import taichi as ti
ti.init(ti.cpu, print_ir = True, advanced_optimization = True)

N = 1
p = ti.field(float, N)

@ti.kernel
def k():
    for n in range(N):
        for t in range(2):
            print("outside", p[n])
            for m in range(1):
                print("inside", p[n])
            p[n] = p[n] + 1.0

k()
```
The result printed is
```
outside 0.000000
inside 0.000000
outside 1.000000
inside 0.000000
```
while the expected result is
```
outside 0.000000
inside 0.000000
outside 1.000000
inside 1.000000
```

I suspect this is due to a bug in the TLS optimisation. If we look at the IR, we can see that the reference to `p[n]` in `print("outside", p[n])` and in `p[n] = p[n] + 1.0` are optimised into local storages, while the reference to `p[n]` in `print("inside", p[n])` is still a global read:
```
kernel {
$0 = offloaded range_for(0, 1) grid_dim=0 block_dim=32 
body {
  <f32> $1 = const 1.0
  <i32> $2 = const 1
  <i32> $3 = const 0
  <i32> $4 = loop $0 index 0
  <i32> $5 = const 2
  <f32> $6 = alloca
  <*gen> $7 = get root [S0root][root]
  <*gen> $8 = [S0root][root]::lookup($7, $3) activate = false
  <*gen> $9 = get child [S0root->S1dense] $8
  <*gen> $10 = [S1dense][dense]::lookup($9, $4) activate = false
  <*f32> $11 = get child [S1dense->S2place<f32>] $10
  <f32> $12 = global load $11
  <f32> $13 : local store [$6 <- $12]
  $14 : for in range($3, $5) block_dim=adaptive {
    <f32> $15 = local load [$6]
    print "outside ", $15, "\n"
    $17 : for in range($3, $2) block_dim=adaptive {
      print "inside ", $12, "\n"
    }
    <f32> $19 = add $15 $1
    <f32> $20 : local store [$6 <- $19]
  }
  <f32> $21 = local load [$6]
  $22 : global store [$11 <- $21]
}
}
```

Also, I verified that if `advanced_optimization` is disabled, the program behaves correctly.

cc @strongoier

lin-hitonami:
Issue: fixes #7263

### Brief Summary
The bug is caused by caching a same global variable in two different local vars in two nested loops. To avoid this, for a cache-able global var, we find the outermost loop that it can be moved out of instead of just moving out of the innermost loop.

### DIFF
--- a/taichi/transforms/cache_loop_invariant_global_vars.cpp
+++ b/taichi/transforms/cache_loop_invariant_global_vars.cpp
@@ -16,7 +16,7 @@ class CacheLoopInvariantGlobalVars : public LoopInvariantDetector {
 
   typedef std::unordered_map<Stmt *, std::pair<CacheStatus, AllocaStmt *>>
       CacheMap;
-  std::stack<CacheMap> cached_maps;
+  std::vector<CacheMap> cached_maps;
 
   DelayedIRModifier modifier;
   std::unordered_map<const SNode *, GlobalPtrStmt *> loop_unique_ptr_;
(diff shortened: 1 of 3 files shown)


AmesingFlank (approved):
thank you!!

PR #7293, opened 2023-02-03
[ci] [amdgpu] Enable amdgpu backend python unit tests

GaleSeLee:

### Brief Summary
1. fix amdgpu backend bugs:
    a. codegen typos
    b. add more types support for `sgn`
    c. use temporary method to handle `runtime.cpp` error
2. enable amdgpu backend python unit test
    a. because of the lack of `print` support, temporarily disable all `print` related tests on amdgpu backend.
    b. there is still something wrong in `gdar_mpm` and `ad_if` tests.

### DIFF
--- a/tests/python/test_cfg_continue.py
+++ b/tests/python/test_cfg_continue.py
@@ -14,8 +14,5 @@ def test_cfg_continue():
                 x[p] = 1
                 continue
 
-            if state[p] != 0:
-                print('test')
-
     foo()
     assert x[0] == 1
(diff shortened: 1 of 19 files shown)


tests/python/test_cfg_continue.py
```diff
@@ -14,8 +14,5 @@ def foo():
                 x[p] = 1
                 continue
 
-            if state[p] != 0:
```
jim19930609, line 17: If you remove this line, there's a chance that the `continue` get removed due to redundancy. 

Btw, this test case should verify that the codes following `continue` did not get executed. Can you file a separate PR to make a fix for this test? Thanks!

```
    @ti.kernel
    def foo():
        for p in range(1):
            if state[p] == 0:
                x[p] = 1
                continue

            x[p] = 100
```

jim19930609 (approved):

PR #7294, opened 2023-02-03
[Bug] Fix copy_from() of StructField

### LINKED ISSUE #7290: `copy_from()` crashes for fields of struct types
The following code crashes
```py
@ti.dataclass
class C:
    i: int
    f: float

cf1 = C.field(shape = (2,))
cf2 = C.field(shape = (2,))

cf1.copy_from(cf2)
```
with the error message
```
TypeError: Cannot copy from a non-field object
```

Interestingly, if the copy is performed on 2D fields, the crash has a different error message:
```py
@ti.dataclass
class C:
    i: int
    f: float

cf1 = C.field(shape = (2,2))
cf2 = C.field(shape = (2,2))

cf1.copy_from(cf2)
```
```
AssertionError: Slicing is not supported on ti.field
```

strongoier:
Issue: fix #7290

### DIFF
--- a/tests/python/test_copy_from.py
+++ b/tests/python/test_copy_from.py
@@ -23,3 +23,32 @@ def test_scalar():
     assert y[0] == 1
     assert y[1] == 0
     assert y[2] == 3
+
+
+@test_utils.test()
+def test_struct():
+    @ti.dataclass
+    class C:
+        i: int
+        f: float
+
+    n = 16
+
+    x = C.field(shape=n)
+    y = C.field(shape=n)
+
+    x[1].i = 2
+    x[2].i = 4
+
+    y[0].f = 1.0
+    y[2].i = 3
+
+    x.copy_from(y)
+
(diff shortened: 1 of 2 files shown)


tests/python/test_copy_from.py
```diff
@@ -23,3 +23,32 @@ def test_scalar():
...
+def test_struct():
+    @ti.dataclass
+    class C:
+        i: int
+        f: float
+
+    n = 16
+
+    x = C.field(shape=n)
+    y = C.field(shape=n)
```
AmesingFlank, line 38: nit, add a test for 2d fields? since that was failing with a different error message

AmesingFlank (approved):
thank you!!

tests/python/test_copy_from.py
strongoier: That might not be necessary because they have the same root cause :-)

PR #7305, opened 2023-02-06
[test] Update the wrong branch impl in test_cfg_continue

GaleSeLee:
Issue: # related to https://github.com/taichi-dev/taichi/pull/7293 conversation

### DIFF
--- a/tests/python/test_cfg_continue.py
+++ b/tests/python/test_cfg_continue.py
@@ -14,8 +14,7 @@ def test_cfg_continue():
                 x[p] = 1
                 continue
 
-            if state[p] != 0:
-                print('test')
+            x[p] = 100
 
     foo()
     assert x[0] == 1


GaleSeLee:
@jim19930609  
Unfortunately, when I run the code below an assertion error will be thrown. 
```
import taichi as ti

ti.init(ti.cpu, print_kernel_llvm_ir=True)

def test_cfg_continue():

    @ti.kernel
    def foo():
        for p in range(1):
            if state[p] == 0:
                x[p] = 1
                continue

            x[p] = 100


    foo()
    print(x[0])
    assert x[0] == 1

test_cfg_continue()
```
Error info(and x[0] will be 0)
```
  File "cfg.py", line 23, in <module>
    test_cfg_continue()
  File "cfg.py", line 21, in test_cfg_continue
    assert x[0] == 1
AssertionError
```

PR #7310, opened 2023-02-06
[llvm] Let the kernel return struct

lin-hitonami:
This PR lets the kernel return a StructType like the real function does. To achieve this, we have to do `kernel->finalize_rets()` after we declared the return values of the kernel like the real function.

I added `MatrixType.from_kernel_struct_ret`, `StructType.from_kernel_struct_ret` and `Kernel.construct_kernel_ret` to read the return values from kernel and construct them to a Python object.

It only works on LLVM-based backends now, and other backends stay the same as before.

* __->__ #7310
* #7262
* #7261


python/taichi/lang/kernel_impl.py
```diff
@@ -841,26 +842,33 @@ def call_back():
...
-                    if is_signed(cook_dtype(ret_dt.dtype)):
-                        it = iter(t_kernel.get_ret_int_tensor(0))
-                    else:
-                        it = iter(t_kernel.get_ret_uint_tensor(0))
-                    ret = Matrix([[next(it) for _ in range(ret_dt.m)]
-                                  for _ in range(ret_dt.n)],
-                                 ndim=getattr(ret_dt, 'ndim', 2))
+                taichi_arch = impl.current_cfg().arch
+                if taichi_arch == _ti_core.Arch.x64 or taichi_arch == _ti_core.Arch.arm64 \
+                        or taichi_arch == _ti_core.Arch.cuda or taichi_arch == _ti_core.Arch.amdgpu:
```
bobcao3, line 847: do we have a `arch_is_llvm` thing in python scope?

bobcao3 (approved):
LGTM

python/taichi/lang/kernel_impl.py
lin-hitonami: Let me add one.

PR #7319, opened 2023-02-07
[bug] Fix u64 field cannot be assigned value >= 2 ** 63

### LINKED ISSUE #7313: u64 field can't be assigned values greater than 2 ** 63
**Describe the bug**
u64 field can't be assigned values greater than 2 ** 63

**To Reproduce**

```py
import taichi as ti
ti.init()
a = ti.field(ti.u64, shape=())
a[None] = 2 ** 63
```

**Log/Screenshots**
Please post the **full log** of the program (instead of just a few lines around the error message, unless the log is > 1000 lines). This will help us diagnose what's happening. For example:

```
  File "<stdin>", line 1, in <module>
  File "/Users/lin/taichi/python/taichi/lang/util.py", line 310, in wrapped
    return func(*args, **kwargs)
  File "/Users/lin/taichi/python/taichi/lang/field.py", line 362, in __setitem__
    self.host_accessors[0].setter(value, *self._pad_key(key))
  File "/Users/lin/taichi/python/taichi/lang/field.py", line 418, in setter
    snode.write_int(key, value)
TypeError: write_int(): incompatible function arguments. The following argument types are supported:
    1. (self: taichi._lib.core.taichi_python.SNode, arg0: List[int], arg1: int) -> None

Invoked with: <taichi._lib.core.taichi_python.SNode object at 0x1278aadf0>, (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0), 9223372036854775808
```

lin-hitonami:
Issue: fixes #7313

### DIFF
--- a/python/taichi/lang/field.py
+++ b/python/taichi/lang/field.py
@@ -384,47 +384,30 @@ class ScalarField(Field):
 class SNodeHostAccessor:
     def __init__(self, snode):
         if _ti_core.is_real(snode.data_type()):
-
-            def getter(*key):
-                assert len(key) == _ti_core.get_max_num_indices()
-                return snode.read_float(key)
-
-            def setter(value, *key):
-                assert len(key) == _ti_core.get_max_num_indices()
-                snode.write_float(key, value)
-                # we only capture write kernels in tape scope with no grad_repaced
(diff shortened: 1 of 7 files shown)


strongoier (approved):
Thanks!

PR #7327, opened 2023-02-09
[gui] Remove renderable reuse & make renderable immediate

bobcao3:
Removed all the renderable reusing logic and since we have added the shader factory, we can now make creation of `renderable` immediate (and release them per frame). This allows the next step of immediate data transfer (so that we can remove the host numpy array cache)


k-ye (approved):
LGTM!

Issue #7240, opened 2023-01-28
Don't broadcast the `cond` in the select statement when `cond` is a scalar


lin-hitonami:
**Concisely describe the proposed feature**
Currently we broadcast the `cond` in the select statement when `cond` is a scalar, but LLVM and SPIR-V both support the select statement with a scalar `cond`, so we don't need to broadcast at all.

PR #7344, opened 2023-02-13
[lang] Stop broadcasting scalar cond in select statements

dream189free:
Issue: #7240 

### Brief Summary
1. Stop broadcasting `cond` when it is a scalar in a `select` statement.
2. Since `config.real_matrix_scalarize` is currently set to `True`, we need to scalarize the `select`. I added another code path to do this, which seems ugly.
3. We can see the expected behavior when `config.real_matrix_scalarize` is unset.


jim19930609 (approved):
Good job! Btw you can add a C++ test to:
1. create a TernaryOpExpression at Frontend IR level
2. make a function call to TernaryOpExpression::type_check(...)
3. Verify that the `cond` wasn't broadcast-ed anymore

you can refer to: `tests/cpp/ir/frontend_type_inference_test.cpp`

jim19930609 (changes requested):
Hmmmm, looks like it's poluted by some other branches?

jim19930609 (approved):

PR #7346, opened 2023-02-13
[amdgpu] Update device memory free

GaleSeLee:
### Brief Summary
1. remove redundant device stream_synchronize which will damage performance seriously.

### DIFF
--- a/taichi/runtime/llvm/llvm_runtime_executor.cpp
+++ b/taichi/runtime/llvm/llvm_runtime_executor.cpp
@@ -207,6 +207,7 @@ void LlvmRuntimeExecutor::synchronize() {
   } else if (config_.arch == Arch::amdgpu) {
 #if defined(TI_WITH_AMDGPU)
     AMDGPUDriver::get_instance().stream_synchronize(nullptr);
+    AMDGPUContext::get_instance().free_kernel_arg_pointer();
 #else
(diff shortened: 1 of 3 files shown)


taichi/runtime/llvm/llvm_runtime_executor.cpp
```diff
@@ -207,6 +207,7 @@ void LlvmRuntimeExecutor::synchronize() {
   } else if (config_.arch == Arch::amdgpu) {
 #if defined(TI_WITH_AMDGPU)
     AMDGPUDriver::get_instance().stream_synchronize(nullptr);
+    AMDGPUContext::get_instance().free_kernel_arg_pointer();
```
jim19930609, line 210: I don't think device memory release should happen during `synchronize()` - very likely it's gonna cause double free. 

This function should get called immediately after you've done launching the kernel, so better call `synchronize() + free_kernel_arg_pointer()` at the very end of `AMDGPUModuleToFunctionConverter::convert()`

GaleSeLee: Thanks for the discussion with @jim19930609, a better way is to use hipfreeasync. Due to the rocm version problem, I think we can switch to the hipfreeasync implementation afterwards.

jim19930609 (approved):

PR #7357, opened 2023-02-14
[Amdgpu] Add print kernel amdgcn

GaleSeLee:
Issue: #

### Brief Summary
1. add `print_kernel_amdgcn` while `gcn` is the ISA on `AMDGPU`


GaleSeLee:
There will be a simpler way to get the gcn isa when `AsmVerbose` is False

GaleSeLee:
> There will be a simpler way to get the gcn isa when `AsmVerbose` is False
Unfortunately, it doesn't work. :-)

turbo0628 (approved):

PR #7365, opened 2023-02-15
[aot] Fix blockers in static library build with XCode

PENGUINLIONG:
Issue: #

### Brief Summary

It seems iOS applications can only consume archives built from XCode.


jim19930609 (approved):
LGTM, thanks~

PR #7376, opened 2023-02-16
[Doc] Update type.md

neozhaoliang:
Issue: #

### DIFF
--- a/docs/lang/articles/type_system/type.md
+++ b/docs/lang/articles/type_system/type.md
@@ -19,7 +19,7 @@ def test():
 
 - Line 3: `x` is an integer since it is assigned an integer value when it is declared for the first time.
 - Line 4: `x` is reassigned a floating-point number 3.14. However, the value of `x` is 3 instead of 3.14. This is because 3.14 is automatically cast to an integer 3 to match the type of `x`.
-- Line 5: In this line, we are assigning a `ti.Vector` to `x`. However, the system throws an error because a `ti.Vector` cannot be cast into an integer.
+- Line 5: The system throws an error, because `ti.Vector` cannot be cast into an integer.
 
 The `ti.types` module in Taichi defines all of the supported data types. These data types are categorized into two groups: primitive and compound.
 
--- a/docs/lang/articles/type_system/type.md
+++ b/docs/lang/articles/type_system/type.md
@@ -79,7 +79,7 @@ def example_cast() -> int:  # the returned type is ti.i64
     y = int(x)  # equivalent to ti.i64(x)
 ```
 
-Furthermore, in the Python scope, when declaring Taichi's data containers using ti.field, ti.Vector, ti.Matrix, ti.ndarray, int, or float, these keywords also serve as aliases for the default integer and floating-point types. For example:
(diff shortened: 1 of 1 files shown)


docs/lang/articles/type_system/type.md
```diff
@@ -9,17 +9,17 @@ Taichi is a statically typed programming language, meaning that the type of a va
...
     x = 3.14  # x is an integer, so the value 3.14 is cast to 3 and x takes the value 3
     x = ti.Vector([1, 1])  # Error!
 ```
 
-- Line 3: `x` is an integer because it is assigned an integer value the first time it is declared.
-- Line 4: `x` is reassigned a floating-point number 3.14 but takes the value 3. This is because 3.14 is automatically cast to integer 3 to match the type of `x`.
-- Line 5:  The system throws an error, because `ti.Vector` cannot be cast into an integer.
+- Line 3: `x` is an integer since it is assigned an integer value when it is declared for the first time.
+- Line 4: `x` is reassigned a floating-point number 3.14. However, the value of `x` is 3 instead of 3.14. This is because 3.14 is automatically cast to an integer 3 to match the type of `x`.
+- Line 5: In this line, we are assigning a `ti.Vector` to `x`. However, the system throws an error because a `ti.Vector` cannot be cast into an integer.
```
strongoier, line 22: Hmm why do we change this part? IMO the original sentences are OK :-)

docs/lang/articles/type_system/type.md
```diff
@@ -233,15 +244,15 @@ sphere2 = sphere_type(center=vec3([1, 1, 1]), radius=1.0)
 When defining a struct with numerous members, the use of `ti.types.struct` can lead to cluttered and unorganized code. Taichi provides a more elegant solution with the `@ti.dataclass` decorator, which acts as a lightweight wrapper around the struct type.
 
 ```python
-@ti.dataclass
+@ti.dataclass skip-ci
```
strongoier, line 248: `skip-ci` should not be put in this line?

docs/lang/articles/type_system/type.md
```diff
@@ -66,26 +66,37 @@ The numeric literals in Taichi's scope have default integer or floating-point ty
...
+```python skip-ci
 ti.init(default_ip=ti.i64, default_fp=ti.f64)
 
+@ti.kernel
+def example_cast() -> int:  # the returned type is ti.i64
+    x = 3.14    # x is of ti.f64 type
+    y = int(x)  # equivalent to ti.i64(x)
+```
+
+Furthermore, in the Python scope, when declaring Taichi's data containers using `ti.field`, `ti.Vector`, `ti.Matrix`, `ti.ndarray`, these two names also serve as aliases for the default integer and floating-point types. For example:
```
strongoier, line 83: `dt` of `ti.Vector` in Python scope doesn't really have effects. Let't not mention them here.

docs/lang/articles/type_system/type.md
neozhaoliang: It's polished by chatgpt so that more like a tech doc style ~

Olinaaaloompa (approved):
LGTM!

Issue #7338, opened 2023-02-10
GGUI  shows blank screen without errors on Intel GT2 card


neozhaoliang:
When running the `mass_spring_game_ggui` example, it shows a blank window without reporting any warning or error messages:

<img src="https://user-images.githubusercontent.com/23307174/217988925-b0787d49-947c-438b-b188-d970f5a4f9b5.png" width=40%>
System info: Ubuntu 22.04, intel GT2 card.

```
ti example
[Taichi] version 1.5.0, llvm 14.0.0, commit c3606b0f, linux, python 3.10.6
```

Issue #7291, opened 2023-02-03
[gui] GGUI swapchain image count issue.


PENGUINLIONG:
This is a notorious issue that have been fixed several times but cannot be properly resolved without crashing either Intel integrated GPU or AMD GPU.

See https://github.com/taichi-dev/taichi/issues/6581, https://github.com/taichi-dev/taichi/pull/3604, https://github.com/taichi-dev/taichi/issues/3601.

Issue #6684, opened 2022-11-21
Code breaks on vulkan backend with v1.1.2


neozhaoliang:
This code crashes on intel GT2 card, the commit is `e2fb83bdc4f3b4cc6cc86cca4ea35433883a49e6`:

```
import taichi as ti
ti.init(arch=ti.vulkan)

d = 3
num_round = 1000000
max_steps = 1000000

ivec = ti.types.vector(d, int)
origin = ivec([0, 0, 0]) # ivec([0] * d)

dirs = ti.Vector.field(d, int, shape=2*d)
for k in range(d):
    dirs[2 *k][k] = 1
    dirs[2 * k + 1][k] = -1

print(dirs)

@ti.func
def choose_random_direction():
    ind = int(ti.random() * 2 * d)
    return dirs[ind]

@ti.kernel
def walk() -> float:
    success = 0
    for _ in range(num_round):
        pos = origin
        for step in range(max_steps):
            pos += choose_random_direction()
            if all(pos == origin):
                success += 1
                break

    return success / num_round

print(walk())
```
Error message:
```
[E 11/21/22 12:57:53.167 1630179] [vulkan_device.cpp:submit@1708] Vulkan Error : -4 : failed to submit command buffer


Traceback (most recent call last):
  File "/home/zhao/China_R_Meeting/return_probability_Z^d.py", line 36, in <module>
    print(do_experiments())
  File "/home/zhao/.local/lib/python3.10/site-packages/taichi/lang/kernel_impl.py", line 942, in wrapped
    return primal(*args, **kwargs)
  File "/home/zhao/.local/lib/python3.10/site-packages/taichi/lang/kernel_impl.py", line 869, in __call__
    return self.runtime.compiled_functions[key](*args)
  File "/home/zhao/.local/lib/python3.10/site-packages/taichi/lang/kernel_impl.py", line 792, in func__
    runtime_ops.sync()
  File "/home/zhao/.local/lib/python3.10/site-packages/taichi/lang/runtime_ops.py", line 8, in sync
    impl.get_runtime().sync()
  File "/home/zhao/.local/lib/python3.10/site-packages/taichi/lang/impl.py", line 453, in sync
    self.prog.synchronize()
RuntimeError: [vulkan_device.cpp:submit@1708] Vulkan Error : -4 : failed to submit command buffer
[...]

PR #7382, opened 2023-02-16
[vulkan] Fix GGUI and vulkan swapchain on AMD drivers

bobcao3:
Issue: #7291

This might also fix Intel. Can anyone with intel iGPU check for me?


PENGUINLIONG:
@neozhaoliang could you check if the fix could break on your intel integrated graphics?

neozhaoliang:
Looks great! This PR fixes the issue: #7338 
However, the error in #6684 still occurs.

PENGUINLIONG:
@neozhaoliang That should be a separate issue. I'm approving since #7338 has been resolved.

PENGUINLIONG (approved):

PR #7387, opened 2023-02-17
[aot] Add test for shared array

### LINKED ISSUE #7274: [aot] Add tests with block_dim and sharearray
It seems setting `block_dim` in `ti.loop_config` and `SharedArray` just works with AOT, but we should definitely add necessary tests to make sure we don't regress in the future.

ailzhang:
Fixes #7274


PENGUINLIONG (approved):
lgtm, thank you

PR #7395, opened 2023-02-18
[vulkan] Change command list submit error message & misc device API cleanups

bobcao3:
### Brief Summary

The confusing "failed to submit command buffer" error message has been replaced with a device lost message (which is the typical type of failure that occurs when failing to submit cmdbuf).

In addition, framebuffer caching code is removed, the reason not to do so is addressed in the comments. There are other misc changes / code quality changes included.


ailzhang (approved):
Thanks!

PR #7407, opened 2023-02-21
[aot] Simplify device capability assignment

PENGUINLIONG:
Issue: #

### Brief Summary

Avoid case specific processing to devcaps


ailzhang (approved):
Thanks!

PR #7410, opened 2023-02-21
[ir] Get the StructType of the kernel parameters

lin-hitonami:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7441
* #7420
* __->__ #7410

PR #7411, opened 2023-02-21
[llvm] Allocate the result buffer from preallocated memory

### LINKED ISSUE #7386: `result_buffer` seems not be freed after the program's destruction in CUDA backend
**Describe the bug**
We currently use cudaMalloc to allocate the `result_buffer` instead of using the preallocated memory, and `result_buffer` is never freed. We should allocate it in the preallocated memory.
https://github.com/taichi-dev/taichi/blob/4af54d49b38a40e1035ca36227ddaf342b0b15c0/taichi/runtime/llvm/llvm_runtime_executor.cpp#L578-L580

lin-hitonami:
issue: closes #7386

* #7420
* #7410
* __->__ #7411


taichi/runtime/llvm/llvm_runtime_executor.cpp
```diff
@@ -600,14 +597,16 @@ void LlvmRuntimeExecutor::materialize_runtime(MemoryPool *memory_pool,
 
     CUDADriver::get_instance().memset(preallocated_device_buffer_, 0,
```
ailzhang, line 598: nit: shall well add a note here to tell readers the first `result_buffer_size` of preallocated memory is always result buffer? 

```
| ==================preallocated device buffer =========================== |
|<-reserved for return ->|                        
```
or a prettier one :P ?

ailzhang (approved):
LGTM!

taichi/runtime/llvm/llvm_runtime_executor.cpp
bobcao3: This seems quite important tbh. I guess a better option is to store two pointers one with this offset baked in for the device to suballocate runtime memory?

lin-hitonami: Do we need to keep the original pointer? IIUC this pointer is passed into the runtime allocator and never used again on the host.

lin-hitonami: I will add the comment in another PR because the comment will change in that PR.

PR #7412, opened 2023-02-21
[aot] Reorganize graph argument creation process

PENGUINLIONG:
Issue: #

### Brief Summary

Make it clean.

Note that it's intentionally not modifying the tests. Let's ensure the new interface doesn't lead to any breakage over the existing test cases.

### DIFF
--- a/python/taichi/examples/graph/texture_graph.py
+++ b/python/taichi/examples/graph/texture_graph.py
@@ -51,19 +51,14 @@ def main():
 
     _rw_tex = ti.graph.Arg(ti.graph.ArgKind.RWTEXTURE,
                            'rw_tex',
-                           channel_format=ti.f32,
-                           shape=(128, 128),
-                           num_channels=1)
+                           fmt=ti.Format.r32f,
+                           ndim=2)
     g_init_builder = ti.graph.GraphBuilder()
     g_init_builder.dispatch(make_texture, _rw_tex)
     g_init = g_init_builder.compile()
 
     g_init.run({'rw_tex': texture})
-    _tex = ti.graph.Arg(ti.graph.ArgKind.TEXTURE,
-                        'tex',
-                        channel_format=ti.f32,
-                        shape=(128, 128),
-                        num_channels=1)
+    _tex = ti.graph.Arg(ti.graph.ArgKind.TEXTURE, 'tex', ndim=2)
     g_builder = ti.graph.GraphBuilder()
     g_builder.dispatch(paint, _t, _pixels_arr, _tex)
     g = g_builder.compile()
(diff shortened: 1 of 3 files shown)


python/taichi/examples/graph/texture_graph.py
```diff
@@ -51,19 +51,14 @@ def main():
...
     g_init_builder.dispatch(make_texture, _rw_tex)
     g_init = g_init_builder.compile()
 
     g_init.run({'rw_tex': texture})
-    _tex = ti.graph.Arg(ti.graph.ArgKind.TEXTURE,
-                        'tex',
-                        channel_format=ti.f32,
-                        shape=(128, 128),
-                        num_channels=1)
+    _tex = ti.graph.Arg(ti.graph.ArgKind.TEXTURE, 'tex', ndim=2)
```
ailzhang, line 61: OOC: do we still need fmt info here?

python/taichi/graph/_graph.py
```diff
@@ -97,77 +99,227 @@ def run(self, args):
...
+            if tag == ArgKind.RWTEXTURE:
+                fmt = TY_CH2FORMAT[(kwargs["channel_format"],
+                                    kwargs["num_channels"])]
+                kwargs["fmt"] = fmt
+                warnings.warn(
+                    "The channel_format and num_channels arguments for texture will be deprecated in v1.5.0, use fmt instead.",
+                    DeprecationWarning)
+            else:
+                warnings.warn(
+                    "The channel_format and num_channels arguments are only required for RW textures since v1.5.0, you can remove them safely.",
```
ailzhang, line 170: are they "only required" or " no longer required" :P ?

python/taichi/graph/_graph.py
```diff
@@ -97,77 +99,227 @@ def run(self, args):
...
+    if tag == ArgKind.RWTEXTURE or tag == ArgKind.TEXTURE:
+        if "dtype" in kwargs:
+            warnings.warn(
+                "The dtype argument for texture will be deprecated in v1.5.0, use format instead.",
+                DeprecationWarning)
+            del kwargs["dtype"]
+
+        if "shape" in kwargs:
+            warnings.warn(
+                "The shape argument for texture will be deprecated in v1.5.0, use ndim instead. (Note that you no longer need the exact texture size.)",
```
ailzhang, line 151: OOC: can various textures with the same shape share the same taichi kernel?

ailzhang (approved):
Thanks!

python/taichi/graph/_graph.py
PENGUINLIONG: yes, we have struct for over rw textures

python/taichi/examples/graph/texture_graph.py
PENGUINLIONG: only required for *rw* texture, not for textures

python/taichi/graph/_graph.py
PENGUINLIONG: nice catch

Issue #4455, opened 2022-03-07
ti.atomic_min gives wrong result when an argument is a matrix


neozhaoliang:
For all atomic operations `ti.atomic_min`, `ti.atomic_max`, `ti.atomic_add`, `ti.atomic_sub`, `ti.atomic_and`, `ti.atomic_xor`, they always return the first argument when this argument is a matrix, and raises error if the first argument is a scalar and the second one is a matrix:

```python
import taichi as ti
ti.init()

@ti.kernel
def foo():
    x = ti.Vector([0, 1, 2])
    z = ti.atomic_min(x, 1)  # [0, 1, 2]
    z = ti.atomic_min(1, x)  # TaichiSyntaxError

foo()
```

strongoier:
1. According to https://docs.taichi.graphics/lang/articles/basic/operator#supported-atomic-operations, the return value of an atomic operation is the old value of first argument.
2. The first argument must be some writable destination, so using a constant doesn't make sense.

neozhaoliang:
Thanks, just learned that.

ailzhang:
Closing via https://github.com/taichi-dev/taichi/issues/4455#issuecomment-1060193714

ailzhang:
Reopening since the error message here can be improved indeed, I currently get `cannot augassign taichi class <class taichi.lang.matrix.Matrix> to scalar expr` which is less obvious to users what they did wrong. Will send a PR for it.

PR #7414, opened 2023-02-21
[lang] Add is_lvalue() to Expr to check writeback_binary operand

dream189free:
Issue: #4455

### Brief Summary
Currently, the following error message appears when the user uses the constant int as the first operand of the `atomic_*()` series of functions: `AttributeError: 'int' object has no attribute 'ptr'`. To make matters worse, taichi gives an internal assertion failure when the user uses some right value expression, e.g., `-x`. 

This PR adds a helper function exposed in the Expr python class called `is_lvalue()`, which the python AST parser can use to check the validity of operands and throw an error message when a rvalue is used as a write target.

### DIFF
--- a/taichi/python/export_lang.cpp
+++ b/taichi/python/export_lang.cpp
@@ -752,12 +752,7 @@ void export_lang(py::module &m) {
              return expr->cast<FieldExpression>()->snode_grad_type ==
                     SNodeGradType::kPrimal;
            })
-      .def("is_lvalue",
-           [](Expr *expr) {
-             return expr->is<IdExpression>() || expr->is<IndexExpression>() ||
-                    (expr->is<ArgLoadExpression>() &&
-                     expr->cast<ArgLoadExpression>()->is_lvalue());
-           })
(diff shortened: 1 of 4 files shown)


taichi/python/export_lang.cpp
```diff
@@ -752,6 +752,12 @@ void export_lang(py::module &m) {
              return expr->cast<FieldExpression>()->snode_grad_type ==
                     SNodeGradType::kPrimal;
            })
+      .def("is_lvalue",
+           [](Expr *expr) {
+             return expr->is<IdExpression>() || expr->is<IndexExpression>() ||
+                    (expr->is<ArgLoadExpression>() &&
+                     expr->cast<ArgLoadExpression>()->is_lvalue());
```
strongoier, line 759: Why not directly return `expr->is_lvalue()`?

dream189free: Thanks! I’ll fix it.

strongoier (approved):
LGTM!

PR #7420, opened 2023-02-22
[llvm] Let the RuntimeContext use the host result buffer

lin-hitonami:
In this PR, we allocate and deallocate the result buffer on the device when launching instead of using the preallocated memory. However, the preallocated result buffer is still used by the runtime JIT calls. We use an async memory pool on CUDA to allocate and deallocate the result buffer, which is almost as fast as using the preallocated memory.

Async memory operations on AMDGPU is not very stable so we use sync Malloc and Free on AMDGPU.

In following PRs the host result buffer will be managed by `LaunchContextBuilder` instead of `Program`.

* #7441
* #7460
* #7448
* __->__ #7420


GaleSeLee:
I will update `ROCm` version both on dev and CI machines next week. Hopefully the async apis in new `ROCm` will be stable.

lin-hitonami:
Per offline sync, @GaleSeLee found that Taichi can't run stably on the new version of the ROCm. Therefore, we will not use async apis in this PR.

ailzhang (approved):
Thanks!

PR #7433, opened 2023-02-23
[aot] Properly handle texture attributes

PENGUINLIONG:
Issue: #

### DIFF
--- a/python/taichi/lang/_texture.py
+++ b/python/taichi/lang/_texture.py
@@ -5,8 +5,7 @@ from taichi.lang.expr import Expr, make_expr_group
 from taichi.lang.matrix import Matrix
 from taichi.lang.util import taichi_scope
 from taichi.types import vector
-from taichi.types.primitive_types import f32, u8
-from taichi.types.texture_type import FORMAT2TY_CH
+from taichi.types.primitive_types import f32
 
 
 def _get_entries(mat):
@@ -130,12 +129,8 @@ class Texture:
         shape (Tuple[int]): Shape of the Texture.
     """
     def __init__(self, fmt, arr_shape):
-        dtype, num_channels = FORMAT2TY_CH[fmt]
-        self.tex = impl.get_runtime().prog.create_texture(
-            dtype, num_channels, arr_shape)
+        self.tex = impl.get_runtime().prog.create_texture(fmt, arr_shape)
         self.fmt = fmt
-        self.dtype = dtype
-        self.num_channels = num_channels
         self.num_dims = len(arr_shape)
         self.shape = arr_shape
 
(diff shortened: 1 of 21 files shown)


taichi/runtime/gfx/aot_module_builder_impl.h
```diff
@@ -23,6 +23,11 @@ class AotModuleBuilderImpl : public AotModuleBuilder {
   void dump(const std::string &output_dir,
             const std::string &filename) const override;
 
+  void mangle_aot_data();
+  void merge_with_old_meta_data(const std::string &path);
+  std::optional<GfxRuntime::RegisterParams> try_get_kernel_register_params(
+      const std::string &kernel_name) const;
```
PGZXB, line 29: We need these?

taichi/runtime/gfx/aot_module_loader_impl.h
```diff
@@ -20,6 +20,7 @@ struct TI_DLL_EXPORT AotModuleParams {
   std::string module_path{};
   const io::VirtualDir *dir{nullptr};
   GfxRuntime *runtime{nullptr};
+  bool enable_lazy_loading{false};
```
PGZXB, line 23: We need these?

PENGUINLIONG: actually no, im gonna get rid of this

python/taichi/lang/_texture.py
```diff
@@ -188,8 +181,6 @@ def to_image(self):
             img (PIL.Image.Image): a PIL image in RGB mode, with the same size as source texture.
         """
         assert self.num_dims == 2
-        assert self.dtype == u8
```
ailzhang, line 191: OOC: do we no longer enforce `dtype` here? If so, the docstring above need to be updated as well. Otherwise we probably still need to assert on `fmt`? ;)

tests/python/test_texture.py
```diff
@@ -103,8 +103,10 @@ def paint(t: ti.f32, tex: ti.types.texture(num_dimensions=2), n: ti.i32):
     paint(0.2, texture2, n2)
     assert impl.get_runtime().get_num_compiled_functions() == 3
 
+    # (penguinliong) Remember that non-RW textures don't enforce a format so
+    # it's the same as the first call to `paint`.
     paint(0.3, texture3, n1)
-    assert impl.get_runtime().get_num_compiled_functions() == 4
+    assert impl.get_runtime().get_num_compiled_functions() == 3
```
ailzhang, line 109: nit: shall we also assert on sampled `pixels` values to make sure it's numerically correct? IIRC we don't have such test yet.

ailzhang (commented):
LGTM!

ailzhang (approved):

tests/python/test_texture.py
PENGUINLIONG: shouldn't this be covered by the texture aot example?

PR #7441, opened 2023-02-24
[llvm] Compile the kernel arguments to a StructType

lin-hitonami:
This PR does things as follows:
1. Add `char *arg_buffer` to `RuntimeContext` as the argument buffer.
2. Manage the memory of `arg_buffer` in `LaunchContextBuilder`.
3. Record `arg_buffer_size` and `args_type` to RuntimeContext for processing the arguments right before the kernel executes in `ModuleToFunctionConverter`. (They will be moved to `LaunchContextBuilder` as soon as the argument of the function returned by `ModuleToFunctionConverter` becomes `LaunchContextBuilder`.)
4. Deal with the argument compiled to struct type in LLVM in the codegen.
5. Allocate the argument buffer on device right before the kernel executes in `ModuleToFunctionConverter`.
6. Support f16

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7576
* #7550
* __->__ #7441

PR #7448, opened 2023-02-27
[refactor] Let the LaunchContextBuilder manage the result buffer

lin-hitonami:
Before this PR, the result buffer of Taichi kernels is managed in `Program`(`program->result_buffer`/`program->host_result_buffer`). This PR lets the `LaunchContextBuilder` manage the result buffer instead of the `Program`. 

The `result_buffer` in `Program` is not removed because it still serves as the result buffer of runtime JIT calls which are not Taichi kernels.

After this PR, we fetch the result from `LaunchContextBuilder` instead of `Kernel` or `Program` in LLVM-based backends.

In future PRs, we will further separate `LaunchContextBuilder` and `Kernel`/`Program`, making it available for AOT as well.

* #7441
* #7460
* __->__ #7448


ailzhang (approved):
Thanks!

PR #7443, opened 2023-02-25
[bug] Returning nan for ti.sym_eig on identity matrix

TachikakaMin:
Issue: #7092

### Brief Summary

Function [_sym_eig3x3](https://github.com/taichi-dev/taichi/blob/ae61418d66b79c4af2173acb00ba8a8b28b5dd54/python/taichi/_funcs.py#L284) is copied from https://www.mpi-hd.mpg.de/personalhomes/globes/3x3/, but the taichi function removed one key "if" statement from original code. 

Just add the full code into taichi.

PR #7449, opened 2023-02-27
[misc] Add an explicit error in cc backend codegen for dynamic indexing

ailzhang:
Issue: #7443 

### Brief Summary
For backends that don't support dynamic indexing, we should raise an explicit error.


strongoier (approved):
Thanks!

PR #7450, opened 2023-02-27
[lang] Fixed texture assertions

PENGUINLIONG:
Issue: #


ailzhang (commented):
nit: after a second thought, I do agree that it's better to let it just throw a compilation error :P It's cleaner than this assert :P 

Btw can we also add a test that triggers the compilation error in this PR? Thanks a lot!

ailzhang (approved):
Thanks!!

PR #7452, opened 2023-02-27
[CUDA] Support large shared memory for CUDA backend

### LINKED ISSUE #6385: Adding support for larger SharedArray

SharedArray helps store the intermediate results so we don't need to write back to global memory, which really increases the performance. 

However, as quoted from CUDA doc: [link](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#shared-memory-8-x)

> Devices of compute capability 8.0 allow a single thread block to address up to 163 KB of shared memory, while devices of compute capabilities 8.6 and 8.9 allow up to 99 KB of shared memory. Kernels relying on shared memory allocations over 48 KB per block are architecture-specific, and must use dynamic shared memory rather than statically sized shared memory arrays. These kernels require an explicit opt-in by using cudaFuncSetAttribute() to set the cudaFuncAttributeMaxDynamicSharedMemorySize; see [Shared Memory](https://docs.nvidia.com/cuda/cuda-c-programming-guide/topics/compute-capabilities.html#shared-memory-7.x) for the Volta architecture.

Right now, we can only allocate 48KB of shared memory (SharedArray), even with new high-end GPUs which already have 99KB or more shared memory. So, is this possible for Taichi to support dynamic shared memory in CUDA that allows us to allocate more shared memory?

turbo0628:
closes #6385

This PR adds dynamic shared memory support for large allocations above 48KB.


taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -147,6 +148,38 @@ class TaskCodeGenCUDA : public TaskCodeGenLLVM {
...
+          tensor_type->get_num_elements() *
+          data_type_size(tensor_type->get_element_type());
+      if (shared_array_bytes > shared_array_threshold) {
+        if (dynamic_shared_array_bytes > 0) {
+          TI_ERROR(
+              "Only one single shared array instance is allowed in current "
+              "version.")
+        }
+        // Clear tensor shape for dynamic shared memory.
+        tensor_type->set_shape(std::vector<int>({0}));
```
ailzhang, line 166: OOC: why do we need to clear tensor shape here? This is a bit weird by changing tensor shape in the codegen phase...

taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -699,8 +733,8 @@ FunctionType CUDAModuleToFunctionConverter::convert(
     for (auto task : offloaded_tasks) {
       TI_TRACE("Launching kernel {}<<<{}, {}>>>", task.name, task.grid_dim,
                task.block_dim);
-      cuda_module->launch(task.name, task.grid_dim, task.block_dim, 0,
-                          {&context}, {});
+      cuda_module->launch(task.name, task.grid_dim, task.block_dim,
+                          task.dynamic_shared_array_bytes, {&context}, {});
```
ailzhang, line 737: maybe I missed this in the code, why do we need to pass `dynamic_shared_array_bytes` to runtime?

taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -161,6 +162,39 @@ class TaskCodeGenCUDA : public TaskCodeGenLLVM {
     llvm_val[stmt] = create_print(formats, types, values);
   }
 
+  void visit(AllocaStmt *stmt) override {
+    // Override shared memory codegen logic for large shared memory
+    if (stmt->ret_type->is<TensorType>() && stmt->is_shared) {
+      auto tensor_type = stmt->ret_type->cast<TensorType>();
+      size_t shared_array_threshold = 49152;
```
ailzhang, line 169: shall we move this to taichi/inc/constants.h and make it a constexpr?

taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -161,6 +162,39 @@ class TaskCodeGenCUDA : public TaskCodeGenLLVM {
...
+      size_t shared_array_bytes =
+          tensor_type->get_num_elements() *
+          data_type_size(tensor_type->get_element_type());
+      if (shared_array_bytes > shared_array_threshold) {
+        if (dynamic_shared_array_bytes > 0) {
+          TI_ERROR(
+              "Only one single large shared array instance is allowed in "
+              "current "
+              "version.")
```
ailzhang, line 178:
```suggestion
              "Only one single large shared array instance is allowed in "
              "current version.")
```

taichi/ir/type.h
```diff
@@ -230,6 +230,10 @@ class TI_DLL_EXPORT TensorType : public Type {
     return shape_;
   }
 
+  void set_shape(std::vector<int> shape) {
```
ailzhang, line 233:
```suggestion
  void set_shape(const std::vector<int>& shape) {
```

taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -39,6 +39,7 @@ static bool is_half2(DataType dt) {
 class TaskCodeGenCUDA : public TaskCodeGenLLVM {
  public:
   using IRVisitor::visit;
+  int dynamic_shared_array_bytes{0};
```
ailzhang, line 42: size_t since this will never go negative?

taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -161,6 +162,39 @@ class TaskCodeGenCUDA : public TaskCodeGenLLVM {
...
+      if (shared_array_bytes > shared_array_threshold) {
+        if (dynamic_shared_array_bytes > 0) {
+          TI_ERROR(
+              "Only one single large shared array instance is allowed in "
+              "current "
+              "version.")
+        }
+        // Clear tensor shape for dynamic shared memory.
+        tensor_type->set_shape(std::vector<int>({0}));
+        dynamic_shared_array_bytes += shared_array_bytes;
```
ailzhang, line 182: OOC:
1. from the logic here we only allow creating one large shared array but why does `test_multiple_shared_array` also pass? Did I miss anything obvious? :P 
2. Looks like `dynamic_shared_array_bytes` will be incremented at most once, shall we use `=` instead of `+=` to make it more clear?

ailzhang (approved):
LGTM!

taichi/codegen/cuda/codegen_cuda.cpp
turbo0628: or make it a compile config?

PR #7466, opened 2023-03-01
[aot] Remove graph kernel interfaces

PENGUINLIONG:
Issue: #

### Brief Summary

Avoid generating kernels that belongs to a graph in module interface.


ailzhang (approved):
Thanks!
Shall we move the tutorial example in taichi-aot-demo to use this generated header after this stabilizes ;P

Issue #7447, opened 2023-02-27
Simplify the implementation of Taichi ops in Python


strongoier:
Currently, Taichi ops in Python are implemented in a complicated way. `Matrix`, `Struct`, `Expr` classes inherit from `TaichiOperations` to make sure they support the same set of ops. The actual implementation of the ops are in https://github.com/taichi-dev/taichi/blob/master/python/taichi/lang/ops.py, which contains complicated dispatching according to whether a value is a Taichi expr or a Python value, a scalar or a compound type value, ... However, much of the logic is unnecessary because only a small set of code paths are actually valid. It will be much cleaner to only keep those valid paths. Therefore, I propose to simplify the logic in the following steps:

- [x] Deprecate arithmetic ops and `fill` of `Struct` in v1.5.0, and remove them in v1.6.0. Such support is simply uncommon and unnecessary. 
- [x] Stop making `Struct` inherit from `TaichiOperations` by implementing necessary ops like `assign`.
- [x] Replace the internal representation of `Matrix` with numpy arrays and remove redundant code paths in `ops.py`.

PR #7474, opened 2023-03-02
[Lang] Stop letting ti.Struct inherit from TaichiOperations

strongoier:
Issue: #7447

### Brief Summary

This PR corresponds to step 2 of #7447.


lin-hitonami (approved):
LGTM. BTW make sure this does not get into v1.5.0

PR #7481, opened 2023-03-03
[ci] New CI docker images

feisuzhu:
- Remove unused parts
- Use `clang-15` in Linux builds
- Use RedHat devtoolset-11's libstdc++ (for compatibility)
- Complete rewrite of Dockerfiles


feisuzhu:
Merged using ghstack

PR #7484, opened 2023-03-04
[lang] Fixed double release of Metal command buffer

PENGUINLIONG:
Issue: #

### Brief Summary

It leads to a crash when there is an outer `autoreleasepool` attempting to release the command buffer. Note the command buffer is already released in the destructor, and the `autoreleasepool` will attempt to release it again because the command buffer is assigned to it.

### DIFF
--- a/taichi/rhi/metal/metal_device.mm
+++ b/taichi/rhi/metal/metal_device.mm
@@ -240,7 +240,10 @@ ShaderResourceSet &MetalShaderResourceSet::rw_image(uint32_t binding,
 MetalCommandList::MetalCommandList(const MetalDevice &device,
                                    MTLCommandQueue_id cmd_queue)
     : device_(&device) {
-  cmdbuf_ = [cmd_queue commandBuffer];
+  @autoreleasepool {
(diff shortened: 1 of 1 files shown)


taichi/rhi/metal/metal_device.mm
```diff
@@ -240,7 +240,10 @@
 MetalCommandList::MetalCommandList(const MetalDevice &device,
                                    MTLCommandQueue_id cmd_queue)
     : device_(&device) {
-  cmdbuf_ = [cmd_queue commandBuffer];
+  @autoreleasepool {
```
bobcao3, line 243: Do you mean that even when we didn't write `@autoreleasepool {` there is a outer release pool already?

PENGUINLIONG: yes, there is an outer pool by default in the cocoa/UIKit event loop.

bobcao3: okok. (That seems like a absolute crap design by Apple)

ailzhang (approved):
Thanks!

PR #7488, opened 2023-03-05
[docs] Reword words of warning about building from source

schuelermine:
### Brief Summary

“We recommend that there is no need” is not grammatical.


ailzhang (approved):
Thanks!

PR #7489, opened 2023-03-06
[llvm] Let the offline cache record the type info of arguments and return values

lin-hitonami:
* #7441
* #7504
* __->__ #7489
* #7494


ailzhang (approved):
So much cleaner, thanks!

PR #7494, opened 2023-03-06
[ir] Separate LaunchContextBuilder from Kernel

lin-hitonami:
This PR separates `LaunchContextBuilder` from `Kernel`, in order to let `LaunchContextBuilder` able to be used in AOT.

This PR also extracts a class `CallableBase` from `Callable`, and let `Callable` inherit `CallableBase`. `CallableBase` contains information sufficient for the `LaunchContextBuilder` to construct arguments and read return values, and it does not contain any runtime information (like `program`) and things related to compiling (like `ir`). In this way, `CallableBase` will be able to be used in AOT.

* #7441
* #7504
* #7489
* __->__ #7494


ailzhang (approved):
Hooray!

PR #7496, opened 2023-03-06
[example] Remove cv2, cairo dependency

neozhaoliang:
This PR removes the `cv2` and `cairocffi` dependency in `requirements_dev.txt`.


lin-hitonami (approved):
LGTM

PR #7504, opened 2023-03-07
[AOT] [llvm] Let AOT kernel inherit CallableBase and use LaunchContextBuilder

lin-hitonami:
After this PR the AOT kernel also have information about the types of the arguments and the return values. 

As a result, AOT kernels of LLVM-based backends can have return values now.

* #7441
* __->__ #7504
* #7489
* #7494


taichi-gardener:
ghstack bot failed: PR #7494 check-run `Build and Test Windows`'s status `failure` is not success!

PR #7506, opened 2023-03-07
[metal] Choose the proper msl version according to the device capability

YuCrazing:
Issue: #

### Brief Summary
Subgroups are only supported in Metal 2.1 and up, we upgrade the default msl version to 2.1.0 to enable subgroup functionalities, otherwise, metal crashes when launching subgroups-involved kernels.


PENGUINLIONG (approved):
Thank you!

bobcao3:
Is this going to make older devices incompatible? If so this might need to be based on device capability

YuCrazing:
@PENGUINLIONG @bobcao3 
Update: 
* Choose a proper msl version according to the device capability .
* Add the exception handling to report compilation errors (if any).

PR #7514, opened 2023-03-08
[lang] [ir] Extract and save the format specifiers in the f-string

dream189free:
Issue: #

### Brief Summary
This PR extracts the format specifiers in f-strings, e.g., `:.2f` in `f'{x:.2f}'`, from print functions in the taichi scope, and saves them in CHI IR.
The formats are checked with the data type inferred by taichi, during front-end type checking.
It is ensured that each content in PrintStmt corresponds to an optional format, before and after any IR pass.

Finally, this PR discards all of these specifiers during codegen, and the specifier inferred from the data type of Expr will be used as before, so there should be no observable differences.

There're will be a follow-up PR which implements specifier merging, after which the validity of specifiers can be checked, and unit tests for that part will be added then.

This PR focuses on extracting formats from f-strings. For formats in constant string with `format()` method, e.g., `'{:.2f}'.format(x)`, extraction is left to a subsequent PR.


taichi/codegen/cuda/codegen_cuda.cpp
```diff
@@ -105,11 +105,16 @@ class TaskCodeGenCUDA : public TaskCodeGenLLVM {
 
     std::string formats;
     size_t num_contents = 0;
-    for (auto const &content : stmt->contents) {
+    TI_ASSERT(stmt->contents.size() == stmt->formats.size());
```
jim19930609, line 108: If there's no CHI IR Pass that transforms `contents` and `formats`, we can migrate this validation to `irpass::type_check()` or even `FrontendPrintStmt::type_check()` - these are common places for verifications.

taichi/ir/statements.h
```diff
@@ -858,13 +858,21 @@ class IfStmt : public Stmt {
 class PrintStmt : public Stmt {
  public:
   using EntryType = std::variant<Stmt *, std::string>;
-  std::vector<EntryType> contents;
+  using FormatType = std::optional<std::string>;
+  const std::vector<EntryType> contents;
+  const std::vector<FormatType> formats;
 
   explicit PrintStmt(const std::vector<EntryType> &contents_)
```
jim19930609, line 865: Any place still using this legacy constructor after your patch? Otherwise we should consider removing it.

taichi/codegen/codegen_utils.h
```diff
@@ -8,4 +8,12 @@ inline bool codegen_vector_type(const CompileConfig &config) {
   return !config.real_matrix_scalarize;
 }
 
+inline std::string merge_printf_specifier(
+    std::optional<std::string> const &from_user,
+    std::string const &from_data_type,
+    Arch arch) {
+  // TODO: implement specifier merging
```
jim19930609, line 15: Can you put an example in the  comment to explain what "specifier merging" be like?

python/taichi/lang/impl.py
```diff
@@ -813,6 +813,12 @@ def ti_format_list_to_content_entries(raw):
     def entry2content(_var):
         if isinstance(_var, str):
             return _var
+        # handle optional format specifier
+        if isinstance(_var, list):
+            assert len(_var) == 2 and (isinstance(_var[1], str)
+                                       or _var[1] is None)
+            _var[0] = Expr(_var[0]).ptr
+            return _var
         return Expr(_var).ptr
```
jim19930609, line 822: Let's try to make the return from different branches as consistent as possible, otherwise it's gonna pollute the code that make use of the function result (`extract_formats` in your case). For example, always return a tuple of "content, format" even if the format is left with None

python/taichi/lang/impl.py
```diff
@@ -854,9 +864,24 @@ def fused_string(entries):
         if accumated:
             yield accumated
 
+    def extract_formats(entries):
+        contents = []
+        formats = []
+        for entry in entries:
+            if isinstance(entry, list):
```
jim19930609, line 871: Let's make `entry2content` have consistent return types so the `entry` here doesn't have dynamic type.

python/taichi/lang/impl.py
```diff
@@ -832,6 +838,10 @@ def vars2entries(_vars):
                 if len(_var) > 0 and isinstance(
                         _var[0], str) and _var[0] == '__ti_format__':
                     res = _var[1:]
+                elif len(_var) > 0 and isinstance(
+                        _var[0], str) and _var[0] == '__ti_fmt_list__':
+                    yield _var[1:]
```
jim19930609, line 843: put `res = _val[1:]` here so the whole function yields at the same place?

python/taichi/lang/impl.py
```diff
@@ -832,6 +838,10 @@ def vars2entries(_vars):
                 if len(_var) > 0 and isinstance(
                         _var[0], str) and _var[0] == '__ti_format__':
                     res = _var[1:]
+                elif len(_var) > 0 and isinstance(
+                        _var[0], str) and _var[0] == '__ti_fmt_list__':
```
jim19930609, line 842: I was thinking `__ti_fmt_list__` functions the same way as `__ti_format__`? What's this attribute used for?

dream189free: These are less relevant code for handling formats extraction from the `str.format()` method. I have removed them in this PR to focus on f-strings.

dream189free: I rewrite this part of code and added a tag `'__ti_fmt_value__'` when building `FormattedValue` (the Python AST counterpart of an expr-format pair `{x:.2f}` in the f-string).

In this way, the Formatted value is converted to the form `['__ti_fmt_value__', Expr, format]` and can therefore be distinguished from a normal list to be printed containing two elements. Thus in `ti_format()` we can use another code path to yield this pair and prevent it being interpreted as a normal list.

python/taichi/lang/impl.py
dream189free: I rewrote this part of code and use a tag named `'__ti_fmt_value__'` when referring to an expr-format pair (for example a `{x:.2f}` in the f-string).

`'__ti_format__'` is used to indicate an artifact list generated in translating `str.format()`, instead of a user inputted one. After removing this tag, the remaining part is still a list and should be handled recursively in `vars2entries()`, so rest part of the `_val` should be pass to `vars2entries()` again then.

On the other hand, `'__ti_fmt_value__'` indicates a single Expr with format, so I think we should create a new code path to yield Expr and format together.

tests/python/test_print.py
```diff
@@ -103,7 +103,7 @@ def func(k: ti.f32):
     ti.sync()
 
 
-@test_utils.test(exclude=[ti.cc, ti.dx11, vk_on_mac, ti.amdgpu], debug=True)
```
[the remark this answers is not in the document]
dream189free: get~

taichi/codegen/cuda/codegen_cuda.cpp
dream189free: Type checking was migrated to FrontendPrintStmt::type_check()

taichi/ir/statements.h
dream189free: Old constructor was removed~

taichi/codegen/codegen_utils.h
dream189free: I added some comments~

python/taichi/lang/impl.py
dream189free: `entry2content()` now has consistent return type~

python/taichi/lang/impl.py
dream189free: get~

tests/python/test_print.py
```diff
@@ -67,6 +67,72 @@ def func(k: ti.f32):
...
+
+    @ti.kernel
+    def func(k: ti.f32):
+        x[None][0, 0] = -1.0
+        y[2] += 1.0
+        print(f'hello {x[None]:.2f} world!')
+        print(f'{(y[2] * k):.3f} {(x[None] / k):e} {y[2]:.2}')
+        print(f'hello {z[None]:+d} world!')
+
+    func(233.3)
```
jim19930609, line 100: Is there a way to capture the print message from stdout and verify that it's expected?

jim19930609 (approved):

tests/python/test_print.py
dream189free: Now stdout is captured by `capfd`, and we can check whether the output is desired~

dream189free: BTW, it seems that Python unit test for printing runs on **all** architectures except`dx11`, `vulkan_on_mac` and `amdgpu`. I corrected this by setting the desired architecture to  `[cpu, cuda, vulkan]`.

jim19930609: yep, printing's only supported on cpu, cuda and vulkan so far. Thanks!

jim19930609 (approved):
Super!

Issue #7244, opened 2023-01-28
[aot] Runtime device is null by surprise with too many TiArgument s


PENGUINLIONG:
This is exposed by the following usage:
```cpp
ti::Kernel kernel = ...
for (int i = 0; i < N_ITER; ++i) {
  // user forgot to `kernel.clear_args();` here.
  kernel.push_arg(i);
  kernel.launch();
}
```
cc @ailzhang

PR #7518, opened 2023-03-09
[aot] Kernel argument count limit

PENGUINLIONG:
Resolves issue: #7244

### Brief Summary

Limit the number of arguments that can be pushed into the context.


ailzhang (approved):
Thanks! 
Nits: in the future when our kernel carries information input & returns we can improve the checks here :P

PR #7540, opened 2023-03-12
[refactor] Remove the difficult-to-implement CompiledKernelData::size()

PGZXB:
Issue: #7002 

### Brief Summary
It is challenging to implement the function `CompiledKernelData::size()`, particularly for the `llvm::CompiledKernelData` class, as it is not possible to obtain the size of the `llvm::Module` without dumping it.


ailzhang (approved):
Thanks! 
OOC: why do we need to record the total size of spirv files btw? :P

PGZXB:
> Thanks! OOC: why do we need to record the total size of spirv files btw? :P

To implement offline cache cleaning.
```
IF the total size > compile_config.offline_cache_max_size_of_files:
    Clean some cache files
```

ailzhang:
@PGZXB aha right, thanks!

PR #7546, opened 2023-03-13
[ir] Let the reference type declare a pointer argument

lin-hitonami:
An argument with reference type is declared as a scalar argument before, which is not right because the reference is passed to the real function as a pointer. This PR fixes this and declares it a pointer argument.

* #7550
* #7441
* #7561
* #7560
* #7558
* #7548
* #7547
* __->__ #7546


ailzhang (approved):
LGTM!

PR #7547, opened 2023-03-13
[aot] Let C-API use LaunchContextBuilder to manage RuntimeContext

lin-hitonami:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7550
* #7441
* #7561
* #7560
* #7558
* #7548
* __->__ #7547
* #7546

PR #7548, opened 2023-03-13
[ir] Let void struct type represent void type

lin-hitonami:
This PR lets `kernel->ret_type` and `kernel->args_type` be a void struct type instead of nullptr when there is no return value or argument.

* #7550
* #7441
* #7561
* #7560
* #7558
* __->__ #7548
* #7547
* #7546


ailzhang (approved):
LGTM!

Issue #7555, opened 2023-03-14
Doc preview isn't available


lin-hitonami:
**Concisely describe the proposed feature**
A clear and concise description of what you want. For example,
> I would like to add an Apple Metal backend to the compiler so that my Macbook GPU can be utilized.

**Describe the solution you'd like (if any)**
A clear and concise description of what you want to achieve and implement.

**Additional comments**

PR #7556, opened 2023-03-14
[Doc] Add doc about returning struct

lin-hitonami:
Includes the change set of #7552. Can't use ghstack because I need the doc preview.
See https://github.com/taichi-dev/taichi/pull/7556/commits/de6f0a8b5dc3aa345880a57241703e0ff5893cbe for the real change set of this PR.
Issue: #7555

### DIFF
--- a/docs/lang/articles/kernels/kernel_function.md
+++ b/docs/lang/articles/kernels/kernel_function.md
@@ -135,7 +135,23 @@ print(x)  # Prints [5, 7, 9]
 
 ### Return value
 
-In Taichi, a kernel can have at most one return value, which can be a scalar, `ti.Matrix`, or `ti.Vector`. Here are the rules to follow when defining the return value of a kernel:
+In Taichi, a kernel can have at most one return value, which can be a scalar, `ti.Matrix`, or `ti.Vector`.
+The return value can also be a `ti.Struct` in the LLVM-based backends (CPU and CUDA backends).
+
+Here is an example of a kernel that returns a struct:
+
+```python
+s0 = ti.types.struct(a=ti.math.vec3, b=ti.i16)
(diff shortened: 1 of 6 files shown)


docs/lang/articles/kernels/kernel_function.md
```diff
@@ -135,7 +135,23 @@ print(x)  # Prints [5, 7, 9]
...
+s1 = ti.types.struct(a=ti.f32, b=s0)
+
+@ti.kernel
+def foo() -> s1:
+    return s1(a=1, b=s0(a=ti.math.vec3(100, 0.2, 3), b=1))
+
+print(foo())  # {'a': 1.0, 'b': {'a': [100.0, 0.2, 3.0], 'b': 1}}
+```
+
+Here are the rules to follow when defining the return value of a kernel:
```
jim19930609, line 154: In Taichi, a kernel is allowed to have a maximum of one return value, which could either be a scalar, `ti.Matrix`, or `ti.Vector`.
Moreover, in the LLVM-based backends (CPU and CUDA backends), a return value could also be a `ti.Struct`.

Here is an example of a kernel that returns a `ti.Struct`:
```python
s0 = ti.types.struct(a=ti.math.vec3, b=ti.i16)
s1 = ti.types.struct(a=ti.f32, b=s0)
@ti.kernel
def foo() -> s1:
    return s1(a=1, b=s0(a=ti.math.vec3(100, 0.2, 3), b=1))
print(foo())  # {'a': 1.0, 'b': {'a': [100.0, 0.2, 3.0], 'b': 1}}
```

When defining the return value of a kernel in Taichi, it is important to follow these rules:

jim19930609 (approved):
Polished with ChatGPT :p

PR #7560, opened 2023-03-14
[refactor] Remove LaunchContextBuilder::set_arg_raw

lin-hitonami:
`set_arg_raw` no longer works when the type of the arguments are not uniformly 64 bits long, so I removed it in this PR and added `set_arg(int arg_id, TypedConstant d)` to set the argument correctly.

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7550
* #7441
* #7561
* __->__ #7560
* #7558
* #7548
* #7547
* #7546

PR #7550, opened 2023-03-14
[refactor] Migrate {set, get}_{arg, ret} functions from RuntimeContext to LaunchContextBuilder

lin-hitonami:
Will delete the functions in `RuntimeContext` in future PRs.

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7612
* #7611
* #7610
* #7608
* #7576
* #7601
* __->__ #7550

PR #7561, opened 2023-03-14
[bug] Fix crashing on printing FrontendFuncCallStmt with no return value

lin-hitonami:
Taichi crashes when printing FrontendFuncCallStmt with no return value because `stmt->ident` (which is the identifier of its return value) is null when it does not have a return value. This PR fixes this bug.

* #7550
* #7441
* __->__ #7561
* #7560
* #7558
* #7548
* #7547
* #7546


bobcao3 (approved):
LGTM

jim19930609 (approved):

PR #7562, opened 2023-03-14
[lang] Fixed useless serial shader to blit ExternalTensorShapeAlongAxisStmt on Metal

PENGUINLIONG:
Issue: #

### Brief Summary

It's already fixed in other SPIR-V powered archs tho. Could we come up with a better way than this sequence of arch checks?


lin-hitonami (approved):
lgtm

PR #7573, opened 2023-03-15
[opt] Constant folding for unary ops on host

ailzhang:
Note this PR might cause some precision differences since we used to do constant folding on device while it's now on the host. 

Issue: #

### DIFF
--- a/taichi/transforms/constant_fold.cpp
+++ b/taichi/transforms/constant_fold.cpp
@@ -118,33 +118,6 @@ class ConstantFold : public BasicStmtVisitor {
     return true;
   }
 
-  bool jit_evaluate_unary_op(TypedConstant &ret,
-                             UnaryOpStmt *stmt,
-                             const TypedConstant &operand) {
-    if (!is_good_type(ret.dt))
-      return false;
-    JITEvaluatorId id{std::this_thread::get_id(),
-                      (int)stmt->op_type,
-                      ret.dt,
-                      operand.dt,
-                      stmt->cast_type,
-                      "",
-                      false};
-    auto *ker = get_jit_evaluator_kernel(id);
-    auto launch_ctx = ker->make_launch_context();
-    launch_ctx.set_arg(0, operand);
-    {
-      std::lock_guard<std::mutex> _(program->jit_evaluator_cache_mut);
-      (*ker)(compile_config, launch_ctx);
-      if (arch_uses_llvm(compile_config.arch)) {
-        ret = launch_ctx.fetch_ret({0});
-      } else {
-        ret.val_i64 = program->fetch_result<int64_t>(0);
-      }
-    }
-    return true;
-  }
-
   void visit(BinaryOpStmt *stmt) override {
     auto lhs = stmt->lhs->cast<ConstStmt>();
     auto rhs = stmt->rhs->cast<ConstStmt>();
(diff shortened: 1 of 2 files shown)


taichi/transforms/constant_fold.cpp
```diff
@@ -176,30 +149,76 @@ class ConstantFold : public BasicStmtVisitor {
...
+      return;
     }
+    const auto dt = operand->val.dt;
+    if (!is_good_type(dt))
+      return;
+    const auto dst_type = stmt->ret_type;
+    switch (stmt->op_type) {
+#define HANDLE_REAL_AND_INTEGRAL_UNARY(OP_TYPE, OP_CPP)                     \
+  case UnaryOpType::OP_TYPE: {                                              \
+    if (dt->is_primitive(PrimitiveTypeID::f32) ||                           \
```
lin-hitonami, line 165: We can use `is_real`, `is_signed`, and `is_unsigned` functions here

taichi/transforms/constant_fold.cpp
```diff
@@ -176,30 +149,76 @@ class ConstantFold : public BasicStmtVisitor {
...
+        dt->is_primitive(PrimitiveTypeID::f64)) {                           \
+      auto res = TypedConstant(dst_type, OP_CPP(operand->val.val_float())); \
+      insert_and_erase(stmt, res);                                          \
+    } else if (dt->is_primitive(PrimitiveTypeID::i32) ||                    \
```
lin-hitonami, line 169: Don't we support 8-bit and 16-bit data types?

ailzhang: @lin-hitonami we didn't do constant folding for 8-bit and 16-bits before (see `is_good_type` function ;) We can definitely add those as well but I'd prefer to add them in a separate PR.

taichi/transforms/constant_fold.cpp
ailzhang: Yup the reason that we list types explicitly here is to align with `is_good_type` but we can update these for sure!

taichi/transforms/constant_fold.cpp
```diff
@@ -176,30 +149,76 @@ class ConstantFold : public BasicStmtVisitor {
...
+      HANDLE_REAL_AND_INTEGRAL_UNARY(acos, std::acos)
+      HANDLE_REAL_AND_INTEGRAL_UNARY(tan, std::tan)
+      HANDLE_REAL_AND_INTEGRAL_UNARY(tanh, std::tanh)
+      HANDLE_REAL_AND_INTEGRAL_UNARY(log, std::log)
+      HANDLE_REAL_AND_INTEGRAL_UNARY(exp, std::exp)
+      HANDLE_REAL_AND_INTEGRAL_UNARY(cast_value, )
+      HANDLE_REAL_AND_INTEGRAL_UNARY(rsqrt, 1.0 / std::sqrt)
+#undef HANDLE_REAL_AND_INTEGRAL_UNARY
+
+#define HANDLE_INTEGRAL_UNARY(OP_TYPE, OP_CPP)                             \
```
jim19930609, line 199: Correct me if were wrong, but looks like `HANDLE_INTEGRAL_UNARY` is a subset of `HANDLE_REAL_AND_INTEGRAL_UNARY`? Why don't we just put `HANDLE_REAL_AND_INTEGRAL_UNARY` here?

jim19930609 (approved):

taichi/transforms/constant_fold.cpp
ailzhang: @jim19930609 the main reason is we actually don't support bit_not/logical_not on real types so I didn't want to fake the support here. (but yea we should have done the type check beforehand...

ailzhang: I'm going to send a separate PR for the type check & error message ;)

PR #7576, opened 2023-03-16
Let LaunchContextBuilder be the argument of the kernel launch function

lin-hitonami:
* (to be filled)

### DIFF
--- a/c_api/src/taichi_core_impl.cpp
+++ b/c_api/src/taichi_core_impl.cpp
@@ -717,8 +717,8 @@ void ti_launch_kernel(TiRuntime runtime,
   auto ti_kernel = (taichi::lang::aot::Kernel *)kernel;
 
   Runtime &runtime2 = *((Runtime *)runtime);
-  taichi::lang::RuntimeContext &runtime_context = runtime2.runtime_context_;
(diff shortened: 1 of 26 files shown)


PENGUINLIONG (commented):
I'm kind of confused.. Is it actually a builder or yet another RuntimeContext? In usual builder pattern there should be something like .build() to create or pull a context instance out of itself but it doesn't seem so?

PGZXB (approved):
LGTM

lin-hitonami:
> I'm kind of confused.. Is it actually a builder or yet another RuntimeContext? In usual builder pattern there should be something like .build() to create or pull a context instance out of itself but it doesn't seem so?

The real runtime context that the host passes to the device (the pointer to argument buffer and result buffer, etc.) is still inside `RuntimeContext`. In LLVM-based backends, the parameter of the LLVM function is still a `RuntimeContext`. `LaunchContextBuilder` is expected to stores information about how to build the `RuntimeContext` (e.g. the type and size of the arguments and return values, `has_grad` and `device_allocation_type`, etc.), and provides APIs to set and retrieve the arguments and return values.

PR #7580, opened 2023-03-17
[refactor] Impl llvm::CompiledKernelData

PGZXB:
* #7587
* #7586
* #7585
* #7584
* #7583
* #7582
* #7581
* __->__ #7580

### DIFF
--- a/taichi/codegen/compiled_kernel_data.cpp
+++ b/taichi/codegen/compiled_kernel_data.cpp
@@ -92,10 +92,9 @@ bool CompiledKernelDataFile::update_hash() {
   return true;
 }
 
-// FIXME: Uncomment after impl LLVMCompiledKernelData
-// #if !defined(TI_WITH_LLVM)
+#if !defined(TI_WITH_LLVM)
 CompiledKernelData::Creator *const CompiledKernelData::llvm_creator = nullptr;
-// #endif
+#endif
 
 #if !defined(TI_WITH_VULKAN) && !defined(TI_WITH_OPENGL) && \
     !defined(TI_WITH_DX11) && !defined(TI_WITH_METAL)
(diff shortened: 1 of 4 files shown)


taichi/codegen/compiled_kernel_data.cpp
```diff
@@ -92,10 +92,9 @@ bool CompiledKernelDataFile::update_hash() {
   return true;
 }
 
-// FIXME: Uncomment after impl LLVMCompiledKernelData
-// #if !defined(TI_WITH_LLVM)
+#if !defined(TI_WITH_LLVM)
```
ailzhang, line 95: OOC: why do we only need this when TI_WITH_LLVM is off?

PGZXB: Note that the `llvm_creator` is used by `CompiledKernelData::load(std::istream&, Err*)` to create an instance of `llvm::CompiledKernelData`. If `TI_WITH_LLVM` is off, the llvm/compiled_kernel_data.* won't be built, which means that `llvm::CompiledKernelData` won't be usable. Therefore, the `llvm_creator ` should be initialized to `nullptr`.

The part may help you:
https://github.com/taichi-dev/taichi/blob/0578b8f9b7c3b00d08d055264c0d4b6e96370845/taichi/codegen/llvm/compiled_kernel_data.cpp#L8-L13

ailzhang: @PGZXB I see, are these creators going to be moved into `llvm::CompiledKernelData` in the future?

PGZXB: OK, but Why? The `lang::CompiledKernelData::llvm_creator` is used by `lang::CompiledKernelData::load(std::istream&, Err*)` but not `llvm::CompiledKernelData::*`.

PGZXB: No, these creators are(and will be) static members of `lang::CompiledKernelData`, because they are used by `lang::CompiledKernelData::load(std::istream&, Err*)` but not `llvm::CompiledKernelData`.

ailzhang (approved):
Thanks!

PR #7584, opened 2023-03-17
[refactor] Implement llvm::KernelCompiler

PGZXB:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7587
* #7586
* #7585
* __->__ #7584
* #7583

PR #7583, opened 2023-03-17
[refactor] Gen code for KernelCodeGen::ir instead of KernelCodeGen::kernel->ir

PGZXB:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7587
* #7586
* #7585
* #7584
* __->__ #7583

PR #7585, opened 2023-03-17
[refactor] Re-impl JIT and Offline Cache on LLVM backends

PGZXB:
* #7587
* #7586
* __->__ #7585
* #7584
* #7583


tests/python/test_offline_cache.py
```diff
@@ -558,29 +543,29 @@ def added_files(arch):
 
     def helper1():
         @ti.experimental.real_func
-        def sum(l: ti.i32, r: ti.i32) -> ti.i32:
+        def sum(l: ti.i32, r: ti.i32, R: ti.i32) -> ti.i32:
```
lin-hitonami, line 546: Why do we change this test?

PGZXB: The re-implemented Offline Cache cache the jit evaluators. The old `sum` generates more than one cache files because of constant folding. To avoid constant folding, let `sum` use the `R` parameter instead of `2`.

PGZXB: But after supporting host constant folding, this change will be not needed.

PGZXB: I disabled caching the evaluators.

lin-hitonami (approved):
LGTM

PR #7590, opened 2023-03-20
[lang] Add assembly printer for CPU backend

jim19930609:
Inspired by GPT-4

### DIFF
--- a/docs/lang/articles/contribution/development_tips.md
+++ b/docs/lang/articles/contribution/development_tips.md
@@ -42,7 +42,7 @@ to make the Taichi compiler print out IRs in different stages:
   kernel compilers.
 - `print_kernel_llvm_ir_optimized=True`: save the optimized LLVM IR
   of each kernel.
-- `print_kernel_nvptx=True`: save the emitted NVPTX of each kernel
+- `print_kernel_asm=True`: save the emitted NVPTX of each kernel
   (CUDA only).
 
 :::note
--- a/taichi/runtime/cpu/jit_cpu.cpp
+++ b/taichi/runtime/cpu/jit_cpu.cpp
@@ -289,9 +289,26 @@ void JITSessionCPU::global_optimize_module_cpu(llvm::Module *module) {
   module_pass_manager.add(llvm::createSeparateConstOffsetFromGEPPass(false));
   module_pass_manager.add(llvm::createEarlyCSEPass(true));
 
+  llvm::SmallString<8> outstr;
+  raw_svector_ostream ostream(outstr);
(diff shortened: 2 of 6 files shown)


taichi/runtime/cpu/jit_cpu.cpp
```diff
@@ -289,11 +289,28 @@ void JITSessionCPU::global_optimize_module_cpu(llvm::Module *module) {
   module_pass_manager.add(llvm::createSeparateConstOffsetFromGEPPass(false));
   module_pass_manager.add(llvm::createEarlyCSEPass(true));
 
+  llvm::SmallString<8> outstr;
+  raw_svector_ostream ostream(outstr);
+  ostream.SetUnbuffered();
+  if (this->config_.print_kernel_nvptx) {
```
ailzhang, line 295: lol shall we also update the doc to make sure devs understand `print_kernel_nvptx` also controls cpu asm? (in fact we should rename this hh

ailzhang (approved):
Thanks!

GaleSeLee:
Thanks! 
But I noticed that we use `IRCompileLayer` for processing during execution, is the asm generated in this pass and the asms during our execution equivalent

jim19930609:
> Thanks! But I noticed that we use `IRCompileLayer` for processing during execution. Is the asm generated in this pass equivalent to the asm in our execution

Good point! However, I did skim through the source code of IRCompileLayer, especially focused on the `IRCompileLayer::add()` function which turns an `llvm::Module` into a JIT-executable `llvm::dylib`. 

It seems that `IRCompileLayer::add()` only places a `MaterializationUnit` onto the `llvm::Module` and does not further transform the IR. In that sense, the assembly code generated from the finalized LLVM IR should be in sync with what executed by the JIT module.

docs/lang/articles/contribution/development_tips.md
```diff
@@ -42,7 +42,7 @@ to make the Taichi compiler print out IRs in different stages:
   kernel compilers.
 - `print_kernel_llvm_ir_optimized=True`: save the optimized LLVM IR
   of each kernel.
-- `print_kernel_nvptx=True`: save the emitted NVPTX of each kernel
+- `print_kernel_asm=True`: save the emitted NVPTX of each kernel
```
lin-hitonami, line 45:
```suggestion
- `print_kernel_asm=True`: save the emitted assembly code of each kernel
```

lin-hitonami (approved):
Thanks!

GaleSeLee (approved):

PR #7604, opened 2023-03-21
[aot] Hide map memory failure

PENGUINLIONG:
Issue: #

### Brief Summary

Vulkan memory mapping can fail when the memory is not visible. But this is checked by both our RHI and the validation layers. We ignore the validation message in this case to unblock the CI.


feisuzhu (approved):
LGTM! CI has passed, many thanks!

PR #7613, opened 2023-03-21
[build] Remove libSPIRV-Tools-shared.(so|dll) in wheel

feisuzhu:
* (to be filled)


ailzhang (approved):
thanks!

PR #7616, opened 2023-03-21
[ci] Do not run tests with mixed backends

feisuzhu:
* #7620
* #7619
* #7618
* #7617
* __->__ #7616


ailzhang (approved):
LGTM！Would you mind opening an issue to track the TLS issue? ;)

feisuzhu:
> LGTM！Would you mind opening an issue to track the TLS issue? ;)

In my opinion, it might not be necessary to address this issue for the following reasons:

- The problem only occurs when repeatedly `ti.init` in the same process, which essentially means it happens during testing.
- Since it's not our responsibility and there's already a workaround in place, the new issue is not actionable.

PR #7622, opened 2023-03-21
[refactor] Use f16 function from external lib

lin-hitonami:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #7622
* #7612
* #7623
* #7611
* #7610
* #7608
* #7576

PR #7612, opened 2023-03-21
[refactor] Migrate members from RuntimeContext to LaunchContextBuilder

lin-hitonami:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #7622
* __->__ #7612
* #7623
* #7611
* #7610
* #7608
* #7576

PR #7623, opened 2023-03-21
[bug] Fix setting arguments exceeding the max arg num

lin-hitonami:
Before this PR, if the number of arguments exceeds the maximum number of arguments, it still sets the argument to the array beyond the index. 

* #7622
* #7612
* __->__ #7623
* #7611
* #7610
* #7608
* #7576

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -655,10 +655,15 @@ class Kernel:
 
             actual_argument_slot = 0
             launch_ctx = t_kernel.make_launch_context()
+            max_arg_num = 64 if impl.current_cfg().arch != _ti_core.cc else 8
+            exceed_max_arg_num = False
             for i, v in enumerate(args):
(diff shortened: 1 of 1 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -655,10 +655,15 @@ def func__(*args):
 
             actual_argument_slot = 0
             launch_ctx = t_kernel.make_launch_context()
```
bobcao3, line 657: I think this check should be done within launch context...

lin-hitonami: In fact this should be done when compiling the kernel. Let's keep this as a temporary solution. I'll refactor it later.

lin-hitonami: BTW the limit will be removed as soon as the refactoring completes.

ailzhang (approved):
Yea the whole section below can be refactored/simplified in compilation, looking forward to it!

PR #7631, opened 2023-03-22
[cpu] Remove atomics demotion for single-thread CPU targets.

turbo0628:
Issue: #7442

Remove the single thread CPU atomic demotion, so that LLVM could do auto-reduction for better performance.

Performance on Apple M1 CPU, single thread:
1273 -> 5122 MFlops

### DIFF
--- a/taichi/transforms/demote_atomics.cpp
+++ b/taichi/transforms/demote_atomics.cpp
@@ -29,10 +29,6 @@ class DemoteAtomics : public BasicStmtVisitor {
     bool demote = false;
     bool is_local = false;
     if (current_offloaded) {
-      if (arch_is_cpu(current_offloaded->device) &&
-          current_offloaded->num_cpu_threads == 1) {
-        demote = true;
-      }
       if (stmt->dest->is<ThreadLocalPtrStmt>()) {
         demote = true;
       }


jim19930609 (approved):

jim19930609 (changes requested):
Wait, demoting the atomic operation when there's no multi-threading tasks should be expected behavior? Why are we removing this logic

turbo0628:
@lin-hitonami I'm not so sure, current test cases are all reducible and LLVM did a good job at vectorization. 

@jim19930609 It's not really expected. For single threaded CPUs, it could leverage the vector units to accelerate atomic ops, which has implicit parallelism under the hood.

jim19930609:
> @lin-hitonami I'm not so sure, current test cases are all reducible and LLVM did a good job at vectorization.
> 
> @jim19930609 It's not really expected. For single threaded CPUs, it could leverage the vector units to accelerate atomic ops, which has implicit parallelism under the hood.

Hmmm, I still don't quite get it. Say you have an `atomic_add_i32()` which runs on single-thread CPU. How does demoting it to `add()` hampers the vector optimization?

turbo0628:
@jim19930609  You are right. Seems that the root cause is not atomic add, it's more about how to properly deal with global tmps. Let's use a simpler example program:
```Python
@ti.kernel
def calc_sum(NPTS:int)->float:
    s = 0.0
    for i in range(0,(NPTS)):
       s += 1.0
    return s
```


If I turn on the atomic add demotion (as we are doing in current. master), the CHI IR is
```llvm
kernel {
$0 = offloaded range_for(0, 1) grid_dim=0 block_dim=1 
body {
  <f64> $1 = const 1.0
  <i32> $2 = const 512
  <i32> $3 = loop $0 index 0
  <*i32> $4 = global tmp var (offset = 0 B)
  <i32> $5 = global load $4
  <i32> $6 = max $5 $2
  <i32> $7 = mul $6 $3
  <i32> $8 = add $7 $6
  <i32> $9 = min $5 $8
  $10 : for in range($7, $9) block_dim=1 {
    <*f64> $11 = global tmp var (offset = 8 B)
    <f64> $12 = global load $11
    <f64> $13 = add $12 $1
    $14 : global store [$11 <- $13]
  }
}
}
```

However if we reserve the atomic add, it then becomes

```llvm
kernel {
$0 = offloaded range_for(0, tmp(offset=0B)) grid_dim=0 block_dim=32 
tls prologue {
  <*f64> $1 = thread local ptr (offset = 0 B)
  <f64> $2 = const 0.0
  $3 : global store [$1 <- $2]
}
body {
  <*f64> $4 = thread local ptr (offset = 0 B)
  <f64> $5 = const 1.0
  <f64> $6 = global load $4
  <f64> $7 = add $6 $5
  $8 : global store [$4 <- $7]
}
tls_epilogue {
  <*f64> $9 = thread local ptr (offset = 0 B)
  <f64> $10 = global load $9
  <*f64> $11 = global tmp var (offset = 8 B)
  <f64> $12 = atomic add($11, $10)
}
}
```
with TLS explicitly enabled. 

So, the problem is that global tmps are not properly transformed when atomic add is demoted. We might have a better fix than brutally keep the atomic add :)

turbo0628:
The performance differences of tuning the atomics on and off is ignorable. The test case is single threaded saxpy.
image

jim19930609 (approved):

PR #7633, opened 2023-03-22
[ci] Resolve "Needed single revision" workaround failure when the rep…

feisuzhu:
…o directory is empty

Issue: #


ailzhang (approved):
nit: this seems like a workaround, do we know the root cause for this error already ;) ?

jim19930609 (approved):

feisuzhu:
> nit: this seems like a workaround, do we know the root cause for this error already ;) ?

Yes, it fails when runs on new machine(without a previously checked-out taichi repo), so we just ignore it.

PR #7636, opened 2023-03-22
[opt] Move binary_jit_evaluator on host

ailzhang:
Note that `logical_or` and `logical+and` are lowered in the frontend ir so constant folding shouldn't see them at all. That's why we updated the C++ test in this PR.

nit: will send out a followup PR to remove `constant_fold`'s dependency on `CompileConfig`.

### DIFF
--- a/taichi/transforms/constant_fold.cpp
+++ b/taichi/transforms/constant_fold.cpp
@@ -12,6 +12,12 @@
 #include "taichi/program/program.h"
 
 namespace taichi::lang {
+namespace {
+template <typename T>
+T sar(T value, T amount) {
+  return value < 0 ? ~(~value >> amount) : value >> amount;
+}
+}  // namespace
 
 class ConstantFold : public BasicStmtVisitor {
  public:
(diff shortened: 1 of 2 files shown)


taichi/transforms/constant_fold.cpp
```diff
@@ -135,8 +65,85 @@ class ConstantFold : public BasicStmtVisitor {
...
+          dst_type, PREFIX(lhs->val.val_uint() OP_CPP rhs->val.val_uint())); \
+      insert_and_erase(stmt, res);                                           \
+    }                                                                        \
+    break;                                                                   \
+  }
+
+      HANDLE_REAL_AND_INTEGRAL_BINARY(mul, , *)
+      HANDLE_REAL_AND_INTEGRAL_BINARY(add, , +)
+      HANDLE_REAL_AND_INTEGRAL_BINARY(sub, , -)
+      HANDLE_REAL_AND_INTEGRAL_BINARY(truediv, , /)
```
lin-hitonami, line 97: `truediv` should all be transformed into `div` in https://github.com/taichi-dev/taichi/blob/6a0b63ef709ae4da836492712d7f43c2c226d479/taichi/transforms/type_check.cpp#L318-L329

taichi/transforms/constant_fold.cpp
```diff
@@ -135,8 +65,85 @@ class ConstantFold : public BasicStmtVisitor {
...
+    }                                                                        \
+    break;                                                                   \
+  }
+
+      HANDLE_INTEGRAL_BINARY(mod, , %)
+      HANDLE_INTEGRAL_BINARY(bit_and, , &)
+      HANDLE_INTEGRAL_BINARY(bit_or, , |)
+      HANDLE_INTEGRAL_BINARY(bit_xor, , ^)
+      HANDLE_INTEGRAL_BINARY(bit_shl, , <<)
+      HANDLE_INTEGRAL_BINARY(bit_shr, , >>)
```
lin-hitonami, line 134: `>>` in C++ is sar if the data type is signed, not shr

lin-hitonami: `bit_shr` is transformed into `bit_sar` in `demote_operation` but this pass is after `constant_fold` so we have to handle it correctly here. Though I think this logic is a bit strange. Maybe it should be handled in `type_check` like `truediv` so we don't need to handle it here. Also it has a bug (unrelated to this PR) ...... https://github.com/taichi-dev/taichi/issues/7650

ailzhang:
Superseded by #7657

PR #7641, opened 2023-03-23
[bug] Fix logical comparison returns -1

ailzhang:
Issue: #

### Brief Summary
We used to return -1 for comparison ops which is kinda confusing... This PR makes it return 1 for true. Note this change might have wide numerical impact on programs so we should pay additional attention when reviewing & bisecting for bugs in the future.


lin-hitonami (approved):
LGTM

Issue #7599, opened 2023-03-20
[RFC] Refactor plan for memory pool & allocators


jim19930609:
# Existing Architecture
![image](https://user-images.githubusercontent.com/22334008/226300202-6c7c4955-599f-4e3d-aff5-39de8c93d1a5.png)

The existing MemoryPool and Allocators has a bunch of issues:

1. Multiple classes such as `VirtualMemoryAllocator`, `SNodeTreeBufferManager`, `Caching Allocator` may directly interact with "OS/Driver" to allocate and **own** raw memory. This is the root reason why memory allocated for `Ndarray` cannot be reused for `SNodeTree` even if it's released explicitly: https://github.com/taichi-dev/taichi/issues/7300

2. MemoryPool, on the other hand, neither allocates nor owns any raw memory from the "OS / Driver" - a useless class at this moment.

3. Weird design pattern and unclear architecture

# Proposed Architecture

![image](https://user-images.githubusercontent.com/22334008/226301587-613de6eb-570a-428e-9169-1a33ded13f2c.png)

In the proposed architecture, we make `MemoryPool` the center of runtime memory management system. `MemoryPool` will be the only class that directly interact with the "OS/Driver" to allocate for raw memory. It holds various `Allocators` that helps arrange the allocated raw memory it owns. 

For example:
```
class MemoryPool {
  // Singleton Implementations

  // ThreadSafe Implementations

  public:
    DeviceAllocation allocate(size_t size) {
        allocator_->allocate(this, size);
    }

    void release(DeviceAllocation devalloc) {
        allocator_->release(devalloc);
    }

  private:
    void allocate_raw_memory(size_t size) {
        // call runtime driver functions to allocate for raw memory
        // designed to be called by Allocator::allocate()
    } 

  private:
    map<std::size, uint8_t*> memory_blocks_;
    std::unique_ptr<Allocator> allocator_;
};
```

Any higher-level class should use `MemoryPool::allocate()` to apply for runtime memory. Such class may include: `SNodeTreeBufferManager`, `NodeManager`, `allocate_ndarray()` and so on.

PR #7648, opened 2023-03-24
[refactor] Remove MemoryPool daemon in LLVM runtime

jim19930609:
Issue: #7599

### Brief Summary
MemoryPool used to startup a `daemon` thread and takes `MemoryRequest` to allocate for runtime memory allocations. However, this is redundant or even harmful because the runtime process still have to wait util the memory is allocated, plus some overhead preparing for the `MemoryRequest` object:

[Previous] Runtime memory allocation via `MemoryRequest`:
```
    auto i = atomic_add_i32(&mem_req_queue->tail, 1);
    taichi_assert_runtime(this, i <= taichi_max_num_mem_requests,
                          "Too many memory allocation requests.");
    auto volatile r = &mem_req_queue->requests[i];
    atomic_exchange_u64((uint64 *)&r->size, size);
    atomic_exchange_u64((uint64 *)&r->alignment, alignment);

    // wait for host to allocate
    while (r->ptr == nullptr) {
#if defined(ARCH_cuda)
      system_memfence();
#endif
    };
```

This PR simplifies the above process into a mutex-based allocation. Tested with `python/taichi/examples/algorithm/mgpcg_advanced.py` (on x64) which made use of `ti.pointer`, and observed 5% performance improvement.

```
[Before Patch] avg = 0.69s
0.686
0.678
0.676
0.688
0.701
0.705
0.704
0.685
0.702
0.679
0.677
0.699

[After Patch] avg = 0.65s
0.642
0.671
0.645
0.651
0.652
0.644
0.657
0.650
0.650
0.650
0.646

```

### DIFF
--- a/taichi/program/program.cpp
+++ b/taichi/program/program.cpp
@@ -315,7 +315,6 @@ void Program::finalize() {
   TI_TRACE("Program finalizing...");
 
   synchronize();
-  memory_pool_->terminate();
   if (arch_uses_llvm(compile_config().arch)) {
     program_impl_->finalize();
   }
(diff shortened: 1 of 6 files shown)


lin-hitonami (approved):
LGTM

PR #7659, opened 2023-03-27
[misc] Rename taichi::lang::llvm to taichi::lang::LLVM

PGZXB:
Rename this namespace because it may introduce conflicts with `::llvm`.


lin-hitonami (approved):
lgtm

PR #7661, opened 2023-03-27
[opt] Cleanup unncessary options in constant fold pass

### LINKED ISSUE #6834: [RFC] Replace JIT evaluator (on device) in constant folding pass with a minimal CHI IR interpreter on host
## Background:

Currently when we do constant folding we have[ this JIT evaluator](https://github.com/taichi-dev/taichi/blob/master/taichi/transforms/constant_fold.cpp#L25-L72) which creates another taichi kernel and then compile and run on the fly and then inject the evaluated result back to the original compilation process. And this JIT evaluator kernel runs on the exact same device as initiated taichi arch. 

## Goal:

We want to replace this JIT evaluator on device with a minimal CHI IR interpreter running on host. The main motivation is: 
- Currently JIT evaluator only handles unary cast op and binary ops, these ops are pretty lightweight, creating a taichi kernel and evaluating them on the device often creates overhead. 
- Well the main motivation is actually simplify the codebase, esp. compiling JIT evaluator kernels requires a different compileconfig than its parent compilation and it has been a headache to maintain this. Replacing it with a small interpreter on host can greatly simplify the process. And we have this [arithmetic interpreter prototype](https://github.com/taichi-dev/taichi/blob/master/taichi/analysis/arithmetic_interpretor.cpp) from @k-ye. 

## Planned steps

- [ ] Add tests for arithmetic interpreter to make sure it's still behaving correctly
- [ ] Expand binary op coverage in arithmetic interpreter so that we can replace the binary jit evaluator kernels
- [ ] Support unary cast in arithmetic interpreter so that we can replace the unary jit evaluator kernels
- [ ] There might be some precision issues since we now evaluate on host instead of taichi device, we think the precision issue is tolerable and we'll just "fix" the tests failures as they show up.

ailzhang:
fixes #6834

### DIFF
--- a/taichi/ir/transforms.h
+++ b/taichi/ir/transforms.h
@@ -106,9 +106,7 @@ void differentiation_validation_check(IRNode *root,
  * AD-stacks before this pass.
  */
 bool determine_ad_stack_size(IRNode *root, const CompileConfig &config);
-bool constant_fold(IRNode *root,
-                   const CompileConfig &config,
-                   const ConstantFoldPass::Args &args);
+bool constant_fold(IRNode *root);
 void offload(IRNode *root, const CompileConfig &config);
 bool transform_statements(
     IRNode *root,
--- a/taichi/transforms/constant_fold.cpp
+++ b/taichi/transforms/constant_fold.cpp
@@ -29,15 +29,6 @@ class ConstantFold : public BasicStmtVisitor {
  public:
   using BasicStmtVisitor::visit;
   DelayedIRModifier modifier;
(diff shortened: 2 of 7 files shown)


PGZXB (approved):
LGTM!

PGZXB:
@ailzhang, these fields can be removed too?

https://github.com/taichi-dev/taichi/blob/97061e4000bdb977fb5d2d9f656f8966ef93e595/taichi/program/program.h#L104-L106

PR #7667, opened 2023-03-27
[misc] Move optimize cpu module to KernelCodeGen

PGZXB:
Issue: #

NOTE: Because setting data layout (`llvm::Module::setDataLayout`) is moved from `JITSession ` to `KernelCodeGen`, we should fix the `llvm_offline_cache_test`.

I will move cuda module optimization to `KernelCodeGen` too.


PGZXB:
NOTE: Because setting data layout (`llvm::Module::setDataLayout`) is moved from `JITSession ` to `KernelCodeGen`, we should fix the `llvm_offline_cache_test`.

lin-hitonami (approved):
LGTM

PR #7676, opened 2023-03-28
[vulkan] Update SPIR-V codegen to emit FP16 consts

bobcao3:
Issue: #

### Brief Summary

After constant folding will fold FP16, CHI-IR passes will now emit FP16 consts. Adding support in codegen & updated SPIRV libraries and vulkan libraries to ensure proper support on the backend


ailzhang (approved):
Thanks!

PR #7681, opened 2023-03-29
[bug] Fix camera controls

liblaf:
### Brief Summary

1. scale movement distance based on frame time
2. support multiple keys pressed at the same time
3. fix conversion between Euler angles and front vector

### DIFF
--- a/python/taichi/ui/camera.py
+++ b/python/taichi/ui/camera.py
@@ -200,9 +200,9 @@ class Camera:
     def track_user_inputs(
         self,
         window,
-        movement_speed: float = 10.0,
-        yaw_speed: float = 100.0,
-        pitch_speed: float = 100.0,
+        movement_speed: float = 1.0,
+        yaw_speed: float = 2.0,
+        pitch_speed: float = 2.0,
         hold_key=None,
     ):
         """Move the camera according to user inputs.
@@ -227,7 +227,7 @@ class Camera:
         time_elapsed = (time.perf_counter_ns() - self.last_time) * 1e-9
         self.last_time = time.perf_counter_ns()
 
-        movement_speed *= time_elapsed
+        movement_speed *= time_elapsed * 60.0
         if window.is_pressed("w"):
             position_change += front * movement_speed
         if window.is_pressed("s"):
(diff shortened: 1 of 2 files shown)


python/taichi/ui/camera.py
```diff
@@ -195,12 +197,14 @@ def get_projection_matrix(self, aspect):
...
-    def track_user_inputs(self,
-                          window,
-                          movement_speed=1.0,
-                          yaw_speed=2,
-                          pitch_speed=2,
-                          hold_key=None):
+    def track_user_inputs(
+        self,
+        window,
+        movement_speed: float = 10.0,
```
bobcao3, line 203: I suggest keeping the values to previously used so we don't break user code. See comments below for speed adjustments

python/taichi/ui/camera.py
```diff
@@ -217,50 +221,49 @@ def track_user_inputs(self,
...
-        self.up(0, 1, 0)
+        if (hold_key is None) or window.is_pressed(hold_key):
+            if (self.last_mouse_x is None) or (self.last_mouse_y is None):
+                self.last_mouse_x, self.last_mouse_y = curr_mouse_x, curr_mouse_y
+            dx = curr_mouse_x - self.last_mouse_x
+            dy = curr_mouse_y - self.last_mouse_y
+
+            yaw, pitch = vec_to_euler(front)
+
+            yaw_speed *= time_elapsed
```
bobcao3, line 255:
```suggestion
            yaw_speed *= time_elapsed * 60.0f
```

So that we match the speed assuming most people are using 60hz vsync

bobcao3 (commented):
LGTM, thank you! Just need to fix the minor issues regarding speed and API

python/taichi/ui/camera.py
liblaf: How about adding an optional parameter `fps: float = 60.0` for more flexibility?

liblaf: I have tested & committed the change. Please let me know if I made anything wrong.

python/taichi/ui/camera.py
liblaf: I have tested & committed the change. Please let me know if I made anything wrong.

python/taichi/ui/camera.py
bobcao3: Let's just hard code the value, since the time independent update is actually the better way, this is just to ensure it's mostly similar the experience of before

liblaf: 👌🏻 I have modified the last commit.

liblaf:
Any other issue?

bobcao3 (approved):
LGTM

PR #7682, opened 2023-03-29
[gui] Fluid visualization utilities

houkensjtu:
Issue: #

## Brief Summary
This PR implemented several visualization utilities to help simulation users to quickly visualize their Scalar/Vector fields.

1. `gui.contour()`

Show the contour of a 2D scalar field (`ti.field`) on a gui canvas. The main purpose of this function is that it will try to handle the case where user's input scalar field's shape does **NOT** match the window size with a bilinear interpolation. For example, a user might create a scalar field as follows
```python
f = ti.field(dtype=ti.f32, shape=(100,100))
```
but later wants to display this field on a canvas of different size
```python
gui = ti.GUI('Title', res=(512, 512)
```
This function will handle this case well and stretch the image. 
```python
while gui.running:
    gui.contour(f)
    gui.show()
```
![image](https://user-images.githubusercontent.com/2747993/228480402-698b294e-6b5f-4cc5-a6b1-f5360dcc49b5.png)

2. `gui.vector_field()`

Similar to `gui.contour()`, this function provides a quick way for user to visualize their vector field. The implementation uses the existing `gui.arrows()` to plot the vectors. It tries to automatically scale the vectors so that the largest vector will occupy ~ 10% of the window. 
```python
v = ti.Vector.field(n=2, dtype=ti.f32, shape=(400,400))                                                                                        
                                                                                                                                               
@ti.kernel                                                                                                                                     
def init():                                                                                                                                    
    for i, j in v:                                                                                                                             
        x, y = i / v.shape[0] - 0.5 , j / v.shape[1] - 0.5                                                                                     
        v[i, j] = ti.Vector([y, -x])                                                                                                           
                                                                                                                                               
init()                                                                                                                                         
gui = ti.GUI('Vector field', res = (400,400), background_color=0xFFFFFF)                                                                       
                                                                                                                                               
while gui.running:                                                                                                                             
    gui.vector_field(v, arrow_spacing=20, color=0x000000)                                                                                      
    gui.show()
```
![image](https://user-images.githubusercontent.com/2747993/228490539-dfaae5f3-4663-47be-a1c5-d101a613e846.png)

4. `write_vtk()`

This function write a 2D `ti.field` into a `.vtr` file which can be read by popular vtk toolset such as Paraview. Example usage
```python
f = ti.field(dtype=ti.f64, shape=(100,100,100))

@ti.kernel                                                                                                                                     
def init():
    for i, j, k in f:
        x = i / 100
        y = j / 100
        z = k / 100
        f[i, j, k] = ti.random()

init()
ti.tools.write_vtk(f, 'output3d')
```
will generate the following `output.vtr` file:
![image](https://user-images.githubusercontent.com/2747993/228492844-18990b51-8e1c-4027-b94e-d3f92f6033d3.png)

### DIFF
--- a/python/taichi/tools/vtk.py
+++ b/python/taichi/tools/vtk.py
@@ -1,6 +1,7 @@
 import numpy as np
 from pyevtk.hl import gridToVTK
 
+
 def write_vtk(scalar_field, filename):
     scalar_field_np = scalar_field.to_numpy()
     field_shape = scalar_field_np.shape
@@ -14,11 +15,11 @@ def write_vtk(scalar_field, filename):
         zcoords = np.array([0, 1])
     elif dimensions == 3:
         zcoords = np.arange(0, field_shape[2])
-    gridToVTK(
-        filename,
-        x=np.arange(0,field_shape[0]),
-        y=np.arange(0,field_shape[1]),
-        z=zcoords,
-        cellData={filename: scalar_field_np})
+    gridToVTK(filename,
+              x=np.arange(0, field_shape[0]),
+              y=np.arange(0, field_shape[1]),
+              z=zcoords,
+              cellData={filename: scalar_field_np})
 
-__all__ = ['write_vtk']    
+
+__all__ = ['write_vtk']
--- a/python/taichi/tools/vtk.py
+++ b/python/taichi/tools/vtk.py
@@ -3,7 +3,8 @@ import numpy as np
 
 def write_vtk(scalar_field, filename):
     try:
-        from pyevtk.hl import gridToVTK  # pylint: disable=import-outside-toplevel
+        from pyevtk.hl import \
+            gridToVTK  # pylint: disable=import-outside-toplevel
     except ImportError:
         raise RuntimeError('Failed to import pyevtk. Please install it via /\
         `pip install pyevtk` first. ')
--- a/python/taichi/tools/vtk.py
+++ b/python/taichi/tools/vtk.py
@@ -1,8 +1,13 @@
 import numpy as np
-from pyevtk.hl import gridToVTK
 
 
 def write_vtk(scalar_field, filename):
+    try:
+        from pyevtk.hl import gridToVTK  # pylint: disable=import-outside-toplevel
+    except ImportError:
+        raise RuntimeError('Failed to import pyevtk. Please install it via /\
+        `pip install pyevtk` first. ')
+
     scalar_field_np = scalar_field.to_numpy()
     field_shape = scalar_field_np.shape
     dimensions = len(field_shape)
--- a/python/taichi/tools/vtk.py
+++ b/python/taichi/tools/vtk.py
@@ -1,0 +1,24 @@
+import numpy as np
+from pyevtk.hl import gridToVTK
+
+def write_vtk(scalar_field, filename):
+    scalar_field_np = scalar_field.to_numpy()
+    field_shape = scalar_field_np.shape
+    dimensions = len(field_shape)
+
+    if dimensions not in (2, 3):
+        raise ValueError("The input field must be a 2D or 3D scalar field.")
+
+    if dimensions == 2:
+        scalar_field_np = scalar_field_np[np.newaxis, :, :]
+        zcoords = np.array([0, 1])
+    elif dimensions == 3:
+        zcoords = np.arange(0, field_shape[2])
+    gridToVTK(
+        filename,
+        x=np.arange(0,field_shape[0]),
+        y=np.arange(0,field_shape[1]),
+        z=zcoords,
+        cellData={filename: scalar_field_np})
+
+__all__ = ['write_vtk']    
--- a/python/taichi/ui/gui.py
+++ b/python/taichi/ui/gui.py
@@ -356,9 +356,11 @@ class GUI:
             Default is False.
         """
         try:
-            from matplotlib import cm  # pylint: disable=import-outside-toplevel
+            from matplotlib import \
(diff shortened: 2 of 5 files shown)


python/taichi/tools/vtk.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+from pyevtk.hl import gridToVTK
```
feisuzhu, line 2:
```suggestion
```

python/taichi/tools/vtk.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def write_vtk(scalar_field, filename):
```
feisuzhu, line 5:
```suggestion
def write_vtk(scalar_field, filename):
    try:
        from pyevtk.hl import gridToVTK
    except ImportError:
        raise RuntimeError('Failed to import pyevtk, please install it via `pip install pyevtk` !')
```

feisuzhu (changes requested):

python/taichi/tools/vtk.py
houkensjtu: Thanks for the suggestions!

python/taichi/ui/gui.py
```diff
@@ -342,6 +342,42 @@ def set_image(self, img):
 
         self.core.set_img(self.img.ctypes.data)
 
+    def contour(self, scalar_field, normalize=False):
```
bobcao3, line 345: Please refactor this and put it in GGUI instead of the old GUI, so that it is `canvas.contour`.

The `ti.GUI` class is not maintained anymore

python/taichi/ui/gui.py
```diff
@@ -699,6 +735,31 @@ def arrow_field(self,
                                       2)
         self.arrows(base, direction, radius=radius, color=color, **kwargs)
 
+    def vector_field(self, vector_field, arrow_spacing=5, color=0xFFFFFF):
```
bobcao3, line 738: ditto, use GGUI. Arrows there needs to be colored lines maybe...

python/taichi/ui/gui.py
houkensjtu: Will do! Thanks for the suggestion.

python/taichi/ui/gui.py
houkensjtu: After briefly looking into the existing functions in `ui/canvas.py`, I think I will combine a call to `canvas.lines()` and a call to `canvas.triangles()` to make the arrows. Let me know if you have further comments!

houkensjtu:
Per @bobcao3 's suggestion, I refactored the plot functions into `canvas.py`. Here are two examples:
```python
import taichi as ti                                                                                                                                                                                                                                                                                                                                                                                                                                                           
ti.init(arch=ti.gpu)                                                                                                                                                                                                                    
                                                                                                                                                                                                                                        
nx, ny = 300, 300                                                                                                                                                                                                                       
im = ti.field(dtype=ti.f64, shape=(nx, ny))                                                                                                                                                                                             
                                                                                                                                                                                                                                        
@ti.kernel                                                                                                                                                                                                                              
def draw():                                                                                                                                                                                                                             
    for i, j in im:                                                                                                                                                                                                                     
        im[i, j] = (i * j) * 1.0 / (nx * ny)                                                                                                                                                                                            
                                                                                                                                                                                                                                        
window = ti.ui.Window('Image', (300, 300))                                                                                                                                                                                              
canvas = window.get_canvas()                                                                                                                                                                                                            
draw()                                                                                                                                                                                                                                  
while window.running:                                                                                                                                                                                                                   
    canvas.contour(im,'plasma', normalize=True)                                                                                                                                                                                         
    window.show() 
```
![image](https://user-images.githubusercontent.com/2747993/229511560-bc2dda31-b94a-41fc-81fc-1fa4772dbedf.png)

And example for `vector_field()` :
```python
import taichi as ti                                                                                                                                                                                                                     
ti.init(arch=ti.cuda)                                                                                                                                                                                                                   
                                                                                                                                                                                                                                        
nx, ny = 300, 300                                                                                                                                                                                                                       
                                                                                                                                                                                                                                        
vf = ti.Vector.field(2, dtype=ti.f32, shape=(nx, ny))                                                                                                                                                                                   
                                                                                                                                                                                                                                        
@ti.kernel                                                                                                                                                                                                                              
def init_vec():                                                                                                                                                                                                                         
    for i, j in vf:                                                                                                                                                                                                                     
        x = (i / nx - 0.5)                                                                                                                                                                                                              
        y = (j / ny - 0.5)                                                                                                                                                                                                              
        vf[i, j] = ti.Vector([-y, x])                                                                                                                                                                                                   
                                                                                                                                                                                                                                        
window = ti.ui.Window("Test for Drawing 2d-lines", (300, 300))                                                                                                                                                                          
canvas = window.get_canvas()                                                                                                                                                                                                            
canvas.set_background_color((1,1,1))                                                                                                                                                                                                    
init_vec()                                                                                                                                                                                                                              
while window.running:                                                                                                                                                                                                                   
    canvas.vector_field(vf, scale=0.1, arrow_spacing=10)                                                                                                                                                                                
    window.show() 
```
![image](https://user-images.githubusercontent.com/2747993/229511722-97840287-1b1f-4fb5-a4c1-423ec785feb1.png)

`vector_field()` is usable but does raise performance concerns as I defined a customized kernel in it to calculate the points for plotting the arrows, plus there are other overheads. WDYT @bobcao3 ?

bobcao3 (approved):
LGTM

feisuzhu (approved):
LGTM

PR #7686, opened 2023-03-29
[Lang] Support formatted printing in str.format() and f-strings

dream189free:
Issue: #5949

### Brief Summary
This PR enables formatted printing in Taichi scope using `str.format()` and f-strings.


jim19930609 (commented):
Logic-wise it looks good to me! 

However, I was thinking `merge_printf_specifier ` should aim at returning a backend-agnostic result of formatted string (only does the "merge" operation), and the adaptions for specific backends (vulkan, cuda, etc) can be done inside each `codegen_cuda.cpp`, `codegen_spirv.cpp` of `visit(PrintStmt *stmt)`. What do you think?

jim19930609 (approved):
Great job!

PR #7698, opened 2023-03-30
[refactor] Unify implementation of ProgramImpl::compile()

PGZXB:
* (to be filled)

### DIFF
--- a/taichi/runtime/program_impls/vulkan/vulkan_program.cpp
+++ b/taichi/runtime/program_impls/vulkan/vulkan_program.cpp
@@ -70,19 +70,6 @@ VulkanProgramImpl::VulkanProgramImpl(CompileConfig &config)
     : ProgramImpl(config) {
 }
 
-FunctionType VulkanProgramImpl::compile(const CompileConfig &compile_config,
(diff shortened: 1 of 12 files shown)


taichi/runtime/program_impls/vulkan/vulkan_program.cpp
```diff
@@ -222,6 +209,11 @@ std::unique_ptr<KernelLauncher> VulkanProgramImpl::make_kernel_launcher() {
   return std::make_unique<gfx::KernelLauncher>(std::move(cfg));
 }
 
+DeviceCapabilityConfig VulkanProgramImpl::get_device_caps() {
```
lin-hitonami, line 212: I think we may need a `GfxProgramImpl` base class for these functions shared in all the gfx backends in the future.

lin-hitonami (approved):
LGTM

taichi/runtime/program_impls/vulkan/vulkan_program.cpp
PGZXB: Agree!

PR #7718, opened 2023-04-03
[test] [cc] Skip tests that cc backend doesn't support

lin-hitonami:
cc backend does not support ndarray and offline cache, so we should skip them.

* #7719
* __->__ #7718


ailzhang (approved):
thanks!

PR #7745, opened 2023-04-06
[Doc] Update write_test.md

houkensjtu:
Mention to reader that they need to install dependencies before running tests.

### DIFF
--- a/docs/lang/articles/contribution/write_test.md
+++ b/docs/lang/articles/contribution/write_test.md
@@ -15,7 +15,7 @@ want to write a **test** to ensure that it functions properly.
 
 :::note
 
-Before running the test set by `pytest`, you need to install the corresponding
+Before running the test launcher `tests/run_tests.py`, you need to install the corresponding
 dependencies:
 ```bash
 pip install -r requirements_test.txt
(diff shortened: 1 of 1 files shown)


docs/lang/articles/contribution/write_test.md
```diff
@@ -13,6 +13,16 @@ Normally we write functional tests in Python.
 For example, you've just added a utility function `ti.log10`. Now you
 want to write a **test** to ensure that it functions properly.
 
+:::note
+
+Before running the test set by `pytest`, you need to install the corresponding
```
feisuzhu, line 18: We should use the launcher named `tests/run_tests.py` , directly calling `pytest` will fail.

feisuzhu (changes requested):
test

feisuzhu (approved):
The rest are LGTM.

docs/lang/articles/contribution/write_test.md
houkensjtu: Thank you for the suggestion. I modified the command in the latest commit.

neozhaoliang (approved):
LGTM!

PR #7762, opened 2023-04-07
[Lang] Fix math module circular import bugs

neozhaoliang:
This PR fixes circular import bugs in the `math` module


ailzhang (commented):
Thanks! btw there's also `_complex.py` :P

ailzhang (approved):
Thanks a ton!

PR #7783, opened 2023-04-11
[gui] Only destruct ImGui context once

### LINKED ISSUE #7769: cache filed not generated
Minimal example:
```python
import taichi as ti

ti.init(arch=ti.gpu, device_memory_GB=8, debug=False)
window = ti.ui.Window('x', (128, 128), vsync=True, show_window=False)
# window2 = ti.ui.Window('x2', (128, 128), vsync=True, show_window=False) # if we uncomment this, cache files won't be generated


# x = ti.field(dtype=ti.f32, shape=10, needs_grad=True)

@ti.kernel
def funcc():
    print(2)

funcc()
```
I noticed cache file is not generated in various cases. This is a minimal example to reproduce it. If we uncomment the code for creating `window2`, there won't be cache files generated.

ailzhang:
Hmmm it's kinda bad practice that we init/destruct global ImGui resources in a `Gui` object owned by `Window` (ImGui seems to be shared by all `Gui` instances) but let's refactor that in a separate PR.

fixes #7769 

### <samp>🤖 Generated by Copilot at ff72743</samp>

This pull request fixes a crash bug in the `Vulkan` backend of the `taichi/ui` module and adds a test case for creating multiple windows with the same module.

### <samp>🤖 Generated by Copilot at ff72743</samp>

* Fix a crash bug in the Vulkan GUI backend by adding a null check before destroying the ImGui context ([link](https://github.com/taichi-dev/taichi/pull/7783/files?diff=unified&w=0#diff-aa837ca266d11faa9968cd0aad76e18080fd5a9d4110ac1a6d6ccd128aee7a4bL238-R240))
* Add a feature test for creating multiple windows with the Taichi UI module and verify their separate rendering ([link](https://github.com/taichi-dev/taichi/pull/7783/files?diff=unified&w=0#diff-7f78177ffeb214e2e6edde3b288143eb941d514aaf0d1471e4038b5ddf11538eR1019-R1024))

### DIFF
--- a/taichi/ui/backends/vulkan/gui.cpp
+++ b/taichi/ui/backends/vulkan/gui.cpp
@@ -235,7 +235,9 @@ Gui::~Gui() {
 #endif
   }
   cleanup_render_resources();
-  ImGui::DestroyContext();
+  if (ImGui::GetCurrentContext() != NULL) {
+    ImGui::DestroyContext();
+  }
 }
 
 bool Gui::is_empty() {
--- a/tests/python/test_ggui.py
+++ b/tests/python/test_ggui.py
@@ -1016,3 +1016,10 @@ def test_wireframe_mode():
     render()
     verify_image(window.get_image_buffer_as_numpy(), 'test_wireframe_mode')
     window.destroy()
+
+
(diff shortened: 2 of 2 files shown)


PGZXB (approved):
Thanks!

PR #7786, opened 2023-04-11
[lang] Split MemoryPool into DeviceMemoryPool and HostMemoryPool

jim19930609:
Issue: https://github.com/taichi-dev/taichi/issues/7599

### Brief Summary
This PR simply split the `MemoryPool` into `HostMemoryPool` & `DeviceMemoryPool` and removed the base class. 

The underlying reason is that `allocation` interface for host memory and CUDA memory can be very different. For example, CUDA backend supports allocating runtime memory inside a "CUDA" kernel, which relies on more parameters such as `LlvmDevice` and `result_buffer`. Therefore, we decided to remove the common base class of MemoryPool and the Device API will switch to use `HostMemoryPool` & `DeviceMemoryPool` directly.


ailzhang (approved):
Thanks!

PR #7797, opened 2023-04-12
[misc] Optimize import overhead: pytorch and get_clangpp

### LINKED ISSUE #7711: Decrease Import time of TaiChi
> I found import TaiChi first time will cost about 2-3 s, which is relatively long than other Python package, such as Numpy. Is it possible to reduce the import time and init time?

turbo0628:
Fix #7711 

Reduce import overhead by:
* Don't import pyTorch when importing Taichi. (It will be imported when necessary)
* Don't unconditionally get clangpp executable path.

The former would cutdown Taichi import overhead from ~2s down to 0.7s, while the latter further reduce it from 0.7s to 0.5s.

### DIFF
--- a/python/taichi/lang/kernel_impl.py
+++ b/python/taichi/lang/kernel_impl.py
@@ -640,10 +640,6 @@ class Kernel:
         self.compiled_kernels[key] = taichi_kernel
 
     def get_function_body(self, t_kernel):
-        if has_pytorch():
-            import torch  # pylint: disable=C0415
-        if has_paddle():
-            import paddle  # pylint: disable=C0415
         # The actual function body
         def func__(*args):
             assert len(args) == len(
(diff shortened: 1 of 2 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -646,6 +640,10 @@ def taichi_ast_generator(kernel_cxx):
         self.compiled_kernels[key] = taichi_kernel
 
     def get_function_body(self, t_kernel):
+        if has_pytorch():
```
ailzhang, line 643: we can even move it further down to where torch/paddle is used? ;)

python/taichi/lang/util.py
```diff
@@ -47,26 +36,37 @@ def has_paddle():
...
+        try:
+            import paddle  # pylint: disable=C0415
+            _has_paddle = True
+        except:
+            pass
     return _has_paddle
 
 
-from distutils.spawn import find_executable
+_clangpp_presence = False
```
ailzhang, line 50: this seems unnecessary any more?

ailzhang (approved):
Thanks!

python/taichi/lang/util.py
turbo0628: Thanks!

PR #7803, opened 2023-04-14
[Lang] Seprate out the scalarization for MatrixOfMatrixPtrStmt and MatrixOfGlobalPtrStmt

jim19930609:
Issue: #

### Brief Summary
MatrixOfGlobalPtrStmt and MatrixOfMatrixPtrStmt can greatly harm the aliasing analysis, since the semantics of which allows recursively-formed multilayered pointers, such as `MatrixOfMatrixPtrStmt[MatrixPtrStmt(AllocaStmt), ...]`

Therefore we separate out the scalarization logics for `MatrixOfGlobalPtrStmt` and `MatrixOfMatrixPtrStmt`, so that these two IRs will be lowered before going into any optimization passes.

### DIFF
--- a/taichi/ir/statements.cpp
+++ b/taichi/ir/statements.cpp
@@ -91,6 +91,7 @@ MatrixPtrStmt::MatrixPtrStmt(Stmt *origin_input,
   origin = origin_input;
   offset = offset_input;
   this->tb = tb;
+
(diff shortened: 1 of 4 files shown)


dream189free (approved):
LGTM!

PR #7809, opened 2023-04-14
[refactor] Compile the Ndarray argument to a struct

lin-hitonami:
**In this PR we represent an ndarray with type `tp` argument with element type `tp` as` struct{primal: *tp}`. We will add fields to the struct in following prs. It will become `struct{shape: array(n), primal: *tp, grad: *tp}` in the end.**

**We may add an ndarray type in CHI-IR and lower it to struct type in codegen in the future, but we temporarily represent the ndarray type with struct type in CHI-IR in this PR.**

### Summary
🔄🐛🆕

This pull request refactors the ndarray support in Taichi, making ndarrays first-class citizens in the IR and kernel representations. It introduces new helper functions, such as `create_ndarray_arg_load` and `insert_ndarray_param`, to simplify the creation and insertion of ndarrays as kernel arguments and parameters. It also updates the code generation and runtime logic for various backends, such as LLVM, SPIRV, and CUDA, to handle ndarrays correctly. It also modifies the tests and the Python interface to reflect the new ndarray support.

> _`ndarray` is the new way to pass the data_
> _Structs and pointers are the keys to the gate_
> _We refactor the code, we generate the IR_
> _We unleash the power of the kernel fire_

### Walkthrough
[...]

PR #7788, opened 2023-04-11
[spirv] Support struct as kernel argument

lin-hitonami:
Only `is_array` is used in `KernelContextAttribute.args()` so we don't need to set the rest of them. We will remove `ArgAttribute` as soon as `is_array` is no longer used.

Will refactor `compile_args_struct` to let it translate the CHI-IR struct type to spirv struct type without using tinyir after we get rid of the extra_args.

### Summary
🔄🆕🛠️

This pull request improves the argument handling and type translation logic for the SPIR-V backend. It simplifies the `KernelContextAttributes` constructor, refactors the `TaskCodegen` class, adds a new method `from_taichi_type` to the `IRBuilder` class, and fixes a test failure in `test_argument.py`. It also introduces some utility functions for aligning and translating struct types.

> _Sing, O Muse, of the valiant codegen heroes_
> _Who tamed the fiery SPIR-V with skill and craft_
> _They forged the `IRBuilder` with wise methods_
> _To shape the types from Taichi as they wished_

### Walkthrough
[...]

PR #7814, opened 2023-04-14
[spirv] Fix the ret type of frexp

lin-hitonami:
Changed the type of `frexp` from struct to a pointer to struct.

* #7809
* #7788
* #7836
* __->__ #7814


ailzhang (approved):
Thanks!

PR #7831, opened 2023-04-18
[Doc] Add kernel sync doc

neozhaoliang:
This  PR adds a new documentation on the sync between taichi and python scope

### DIFF
--- a/docs/lang/articles/kernels/kernel_sync.md
+++ b/docs/lang/articles/kernels/kernel_sync.md
@@ -53,9 +53,8 @@ print(end - start)
 Most of the time, Taichi automatically handles data synchronization. For example, Taichi will automatically call `ti.sync()` to synchronize data in the following cases:
 
 1. The kernel has a return value.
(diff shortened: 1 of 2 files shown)


docs/lang/articles/kernels/kernel_sync.md
```diff
@@ -0,0 +1,64 @@
...
+benchmark()
+ti.sync()
+end = time.time()
+print(end - start)
+```
+
+Most of the time, Taichi automatically handles data synchronization. For example, Taichi will automatically call `ti.sync()` to synchronize data in the following cases:
+
+1. The kernel has a return value.
+2. The kernel contains a `print` statement.
```
jim19930609, line 56: let's remove this line, cause it's not guaranteed to perform `ti.sync` at the very end of the kernel (Let's say you have a `print` resides somewhere in the middle, so it only syncs everytime it's about to print)

jim19930609 (approved):

jim19930609 (approved):

PR #7846, opened 2023-04-19
[lang] Support TensorType for AutoDiff

jim19930609:
Issue: #

## Brief Summary
This PR added TensorType support for `irpass::autodiff()`, which includes the following significant changes:

1. Minor adjustments to original `adjoint()` and `dual()` functions to adapt statements with TensorType
2. IR Change: Added `return_ptr` to `AdStackLoadTopStmt`. With `return_ptr=true`, `AdStackLoadTopStmt` returns a pointer to the object at stack top without further loading its value. This new parameter will be removed in a following PR to keep the semantic clean.
3. Added support for `MatrixPtrStmt` and `MatrixInitStmt` for `irpass::MakeAdjoint()` and `irpass::MakeDual()`, both statements simply trans-passes the `adjoint`/`dual` to the `operands`/`output`.

4. Handles the special case of `LocalStore(MatrixPtr(AdStackAlloca(TensorType<...>)))` in `irpass::ReplaceLocalVarWithStacks`:

    Basically we replace the original `scalar-LocalStore` with `tensor-LocalStore`
    ```
                Fwd:
                $0 = adstack alloca <4 x i32>
                $0 = adstack load top
                $1 = matrix ptr $0, 2 // offset = 2
                $2 : local store $1, $val
    
                Replaced:
                $0 = adstack alloca <4 x i32>
                $0 = adstack load top
                $1 = matrix ptr $0, 2 // --> erase
    
                $2 = matrix ptr $0, 0
                $3 = load $2
    
                $4 = matrix ptr $0, 1
                $5 = load $4
    
                $6 = matrix ptr $0, 3
                $7 = load $6
    
                $5 = matrix init [$3, $5, $val, $7]
    
                $2 : adstack push $5
    ```

5. Added scalarization for AdStack-related statements for `irpass::scalarize()`

### Taichi Code
```
@ti.kernel
def test(...):
    b = ti.Vector([0, 1, 2, 3])
    b[2] = 100
    y = b[3] * b[2] * x
```

### Gradient
截图 2023-04-24 15-26-26


tests/python/test_ad_tensor.py
```diff
@@ -0,0 +1,34 @@
...
+
+try:
+    import autograd.numpy as np
+    from autograd import grad
+
+    has_autograd = True
+except:
+    pass
+
+
```
dream189free, line 19: Should we add a decorator similar to the one used in `test_ad_basics.py` at this location?

```suggestion
def if_has_autograd(func):
    # functools.wraps is nececssary for pytest parametrization to work
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        if has_autograd:
            func(*args, **kwargs)

    return wrapper
    
    
    @if_has_autograd
```

dream189free (approved):
LGTM!

taichi/transforms/auto_diff.cpp
```diff
@@ -474,22 +502,192 @@ class ReplaceLocalVarWithStacks : public BasicStmtVisitor {
...
+
+            $4 = matrix ptr $0, 1
+            $5 = load $4
+
+            $6 = matrix ptr $0, 3
+            $7 = load $6
+
+            $5 = matrix init [$3, $5, $val, $7]
+
+            $2 : adstack push $5
```
erizmr, line 555: nit: the numberings of sample IR in the comments are somehow misordered

taichi/transforms/auto_diff.cpp
```diff
@@ -749,33 +970,42 @@ class MakeAdjoint : public ADTransform {
...
     }
   }
 
   Stmt *adjoint(Stmt *stmt) {
-    if (!is_real(stmt->ret_type) || stmt->is<ConstStmt>()) {
+    DataType adjoint_dtype = stmt->ret_type.ptr_removed();
+    if (stmt->ret_type->is<TensorType>()) {
+      adjoint_dtype = TypeFactory::get_instance().get_tensor_type(
+          stmt->ret_type->as<TensorType>()->get_shape(), PrimitiveType::f32);
+    } else if (stmt->is<MatrixPtrStmt>()) {
```
erizmr, line 995: Why we skip the adjoint of `MatrixPtrStmt`?
e.g., in the case below, IIUC the adjoint of $val comes from that of $1, which is a adjoint of a `MatrixPtrStmt`
```
        $0 = Tensor alloca <4 x i32>
        $1 = matrix ptr $0, 2 // offset = 2
        $2 : local store $1, $val
```

taichi/transforms/auto_diff.cpp
```diff
@@ -1140,6 +1375,122 @@ class MakeAdjoint : public ADTransform {
...
+
+  void visit(MatrixPtrStmt *stmt) override {
+    if (stmt->origin->is<GlobalPtrStmt>() ||
+        stmt->origin->is<ExternalPtrStmt>()) {
+      // Handled in GlobalLoadStmt and GlobalStoreStmt
+      return;
+    }
+
+    Stmt *adjoint_value = nullptr;
+    if (stmt->offset->is<ConstStmt>()) {
```
erizmr, line 1387: Shall we move the adjoint creation into `Stmt *adjoint(Stmt *stmt)` ? Maybe we can only handle the logic of chain rule here.

taichi/transforms/auto_diff.cpp
```diff
@@ -1140,6 +1375,122 @@ class MakeAdjoint : public ADTransform {
     accumulate(stmt->val, insert<GlobalLoadStmt>(adjoint_ptr));
     stmt->parent->erase(stmt);
   }
+
+  void visit(MatrixPtrStmt *stmt) override {
+    if (stmt->origin->is<GlobalPtrStmt>() ||
+        stmt->origin->is<ExternalPtrStmt>()) {
+      // Handled in GlobalLoadStmt and GlobalStoreStmt
```
erizmr, line 1382: Sorry I am not familiar with the usage of `MatrixPtrStmt`. Is it possible to use it to load a  `GlobalPtrStmt` or `ExternalPtrStmt`? And I am wondering why we skip them here?

erizmr (commented):
Thanks!!

taichi/transforms/auto_diff.cpp
jim19930609: On a second thought, I feel like it's actually working on the chain rule instead of the `adjoint` (I can change the name a little bit).

Taking the following example, say we have a MatrixPtrStmt in the forward pass:

```
f32* $1 = MatrixPtrStmt(Tensor<4 x f32>* $origin, i32 $offset)
```

The problem is that there's shape mismatch between the adjoint of `$origin` (`Tensor<4 x f32>*`) and the adjoint of `$1` (`f32*`). The `adjoint_value` here denotes the adjoint for `$origin` and the entire effort here is to construct the `Tensor<4 x f32>` adjoint from `f32*` adjoint.

So it's actually working on the chain rule for MatrixPtrStmt

taichi/transforms/auto_diff.cpp
erizmr: Thanks for the clarification.

taichi/transforms/auto_diff.cpp
```diff
@@ -501,22 +529,192 @@ class ReplaceLocalVarWithStacks : public BasicStmtVisitor {
...
+          auto stack_push = Stmt::make<AdStackPushStmt>(stack_top->stack,
+                                                        matrix_init_stmt.get());
+
+          stmt->insert_before_me(std::move(matrix_init_stmt));
+          stmt->replace_with(std::move(stack_push));
+          return;
+
+        } else {
+          /*
+            [Dynamic index]
```
erizmr, line 623: nit: numberings issue as above ;)

erizmr (approved):
LGTM! Thanks!

taichi/transforms/auto_diff.cpp
jim19930609: thx~

PR #7857, opened 2023-04-20
[ci] Polishing build.py, wave 4

feisuzhu:
- Update mambaforge && sccache
- Option to use developer managed Python environment
- Update docs about the Python thing, and behaviors considered intrusive


ailzhang (approved):
Thanks!

PR #7860, opened 2023-04-21
[Lang] Fix a bug on PosixPath

neozhaoliang:
When calling  `ti gallery,` the `_exec_python_file` function in `_main.py` simply passes a `pathlib.PosixPath` object to the command. This may cause problems on Windows platforms. This PR fixes this issue by forcing using the absolute string path.


ailzhang (approved):
Thanks!

PR #7861, opened 2023-04-21
[Doc] Update faq and ggui, and add them to CI

neozhaoliang:
This PR updates the `bool` section and `fps_limit` for ggui.

### DIFF
--- a/docs/conftest.py
+++ b/docs/conftest.py
@@ -92,6 +92,9 @@ def show(orig, self, *args, **kwargs):
     return orig(self, *args, *kwargs)
 
 
+GGUI_WINDOW = None
+
+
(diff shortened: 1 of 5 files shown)


docs/conftest.py
```diff
@@ -103,6 +103,11 @@ def show(orig, self, *args, **kwargs):
     return orig(self, *args, *kwargs)
 
 
+@hook(plt)
```
jim19930609, line 106: where is this function used?

feisuzhu: `lang/articles/faqs/faq.md`, there are `plt.show()` in docs

jim19930609 (approved):

PR #7877, opened 2023-04-24
[lang] Support 0 dim ndarray read & write in python scope

ailzhang:
* #7878
* __->__ #7877

### DIFF
--- a/python/taichi/lang/_ndarray.py
+++ b/python/taichi/lang/_ndarray.py
@@ -208,8 +208,8 @@ class Ndarray:
 
     @python_scope
     def _pad_key(self, key):
-        if key is None or key == ():
-            return (0,)
+        if key is None:
+            key = ()
         if not isinstance(key, (tuple, list)):
             key = (key,)
         if len(key) != len(self.arr.total_shape()):
--- a/python/taichi/lang/_ndarray.py
+++ b/python/taichi/lang/_ndarray.py
@@ -209,7 +209,7 @@ class Ndarray:
     @python_scope
     def _pad_key(self, key):
         if key is None:
-            key = ()
+            return (0,)
         if not isinstance(key, (tuple, list)):
             key = (key,)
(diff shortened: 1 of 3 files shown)


taichi/program/ndarray.cpp
```diff
@@ -14,6 +14,10 @@ namespace {
 
 size_t flatten_index(const std::vector<int> &shapes,
                      const std::vector<int> &indices) {
+  if (shapes.size() == 0) {
+    TI_ASSERT(indices.size() == 1 && indices[0] == 0);
+    return 0;
+  }
   TI_ASSERT(shapes.size() == indices.size());
   if (indices.size() == 1) {
```
lin-hitonami, line 22: Why do we need this if-else? IIUC we can write
```
size_t ind = 0;
for (int i = 0; i < indices.size(); i++) {
  ind = ind * shapes[i] + indices[i];
}
```
and then we don't need to care about the size of `indices`, and it works if the size is 0.

ailzhang: lol good point, let me update

python/taichi/lang/_ndarray.py
```diff
@@ -209,7 +209,7 @@ def _fill_by_kernel(self, val):
     @python_scope
     def _pad_key(self, key):
         if key is None:
-            key = ()
+            return (0,)
```
lin-hitonami, line 212: Does it work if we just pass a void tuple?

ailzhang: Nice catch! let me add that since both numpy and ti.field support it...

lin-hitonami: I mean can we return `()` instead of `(0, )` here now since the backend supports the case when `indices.size() == 0`?

ailzhang: oh I see what you meant, makes sense!

lin-hitonami (approved):
LGTM

PR #7882, opened 2023-04-24
[ir] [lang] Support ti.u1 as boolean type

listerily:
Issue: #577 

### Brief Summary

This PR added ti.u1 as boolean type to Taichi Lang.

#### Features
+ Full ti.u1 support on LLVM backends and limited support on SPIR-V backends;

#### Changes
+ If and while statements now accept ti.u1 and other integers
+ Ternary expressions accepts ti.u1 and other integers
+ Comparison expressions yields ti.u1 instead of ti.i32
+ ti.is_active yields ti.u1 instead of ti.i32
+ Constant Fold support for boolean type during irpass
+ Replaced bitwise_and with logical_and in some cases; removed the constraint that logical_and should be demoted during irpass
+ Added u1 support during type check.

#### Limitations
+ On spir-v backends, ti.u1 as dtype in compound types are not supported currently
+ On spir-v backends, all ti.u1 in return/args/buffer were casted to i32 for the lack of support of boolean pointers on spir-v
+ Boolean types are printed as 0/1 instead of False/True. The issue with printing boolean values in Vulkan is that there are no options to print them as "True" or "False", so they are printed as 0 or 1. To solve this, one possible solution is to split the printf statements into three parts: one to print the messages before the boolean value, one to print the messages after the boolean value, and one with an if-else statement that prints "True" or "False" based on the boolean value. However, this solution can cause problems when the printf message is long, as the printf buffer can run out of memory and the message can be truncated. This issue occurs even if there are no boolean variables to be printed, as we always split the printf statements no matter what type the variable to be printed is.

#### Support ti.u1 to be used as dtype in compound types
On spir-v, ti.u1 used as compound type like `a = ti.field(ti.u1, 10)` would yield:
```
taichi/rhi/opengl/opengl_device.cpp:924: virtual void taichi::lang::opengl::GLCommandList::CmdBufferFill::execute(): Assertion 'false' failed.
```
#### Support better ti.u1 in print
Find possible solutions to correctly print boolean variables.

### <samp>🤖 Generated by Copilot at 7b21443</samp>

This pull request introduces a new boolean type `u1` in Taichi, and updates various parts of the language, compiler, and runtime to support and use the new type. The purpose is to enable the use of boolean values in Taichi and its interoperability with other frameworks, and to fix some bugs and errors related to the boolean type. The pull request modifies several files in the `python`, `taichi`, and `docs` directories, such as `matrix_ops.py`, `codegen_llvm.cpp`, and `llvm_sparse_runtime.md`.


listerily:
**This PR will be split into a few separate and stacked PRs. It will be closed.**

PR #7884, opened 2023-04-24
[Bug] Fix wrong datatype size when writing to ndarray from Python scope

ailzhang:
* #7878
* __->__ #7884

We might be able cleanup the interfaces that `Ndarray` class provides
but will leave that for a separate PR.


lin-hitonami (approved):
LGTM

Issue #7678, opened 2023-03-29
calling x.fill(1) for a field x of type ti.u8 raises warning


neozhaoliang:
To reproduce:

```python
x = ti.field(ti.u8, 100)
x.fill(1)
```

raise the warning message

```
[W 03/29/23 10:53:27.516 328640] [type_check.cpp:type_check_store@36] [$25] Global store may lose precision: u8 <- i32
File "/home/zhao/taichi/python/taichi/_kernels.py", line 24, in fill_tensor:
        tensor[I] = val
        ^^^^^^^^^^^^^^^
```

However, when one tries to use `ti.u8(1)` to cast the constant to `ti.u8` type, it raises error

```
TypeError: 'taichi._lib.core.taichi_python.DataType' object is not callable
```

PR #7899, opened 2023-04-26
[Lang] Add cast to field.fill() interface

jim19930609:
Issue: #https://github.com/taichi-dev/taichi/issues/7678

There's one potential hazard: if the user mistakenly filled a ti.field with out-of-scope values (like filling a ti.field(ti.u8) with > 255), it's just gonna overflow silently.

### DIFF
--- a/python/taichi/_kernels.py
+++ b/python/taichi/_kernels.py
@@ -20,9 +20,9 @@ from taichi.math import vec3
 # A set of helper (meta)functions
 @kernel
 def fill_field(field: template(), val: template()):
-    tmp = ops.cast(val, field.dtype)
+    value = ops.cast(val, field.dtype)
     for I in grouped(field):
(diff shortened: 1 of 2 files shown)


python/taichi/_kernels.py
```diff
@@ -19,9 +19,10 @@
 
 # A set of helper (meta)functions
 @kernel
-def fill_tensor(tensor: template(), val: template()):
-    for I in grouped(tensor):
-        tensor[I] = val
+def fill_field(field: template(), val: template()):
+    value = ops.cast(val, field.dtype)
```
neozhaoliang, line 23: Have a question here: does this `cast` operation introduces extra instructions when `val` and `field` are of the same type?

bobcao3 (approved):
LGTM

PR #7903, opened 2023-04-26
[example] Fix ti example bugs

neozhaoliang:
This PR fixes the following bugs in the `ti example` command:

1. The command will crash if the user type a non-digits string.
2. It prints the traceback log if the user kills the command via the keyboard.


python/taichi/_main.py
```diff
@@ -307,10 +307,15 @@ def colormap(index, name):
...
-            while not 0 <= index < len(names):
-                index = int(input(f"Example [{index}] does not exist. Please try again: "))
-            example_name = names[index]
+            try:
+                index = input(f"Please input the example index (between 0 and {len(names)}): ")
+                while not index.isdigit() or int(index) >= len(names):
+                    index = input(f"Example [{index}] does not exist. Please try again: ")
+                example_name = names[int(index)]
+            except KeyboardInterrupt as e:
+                print("\nCacelled by user, exit...")
```
ailzhang, line 316:
```suggestion
                print("\nCancelled by user, exiting...")
```

ailzhang (approved):
Thanks!

PR #7906, opened 2023-04-27
[cuda] Only set CU_LIMIT_STACK_SIZE when necessary

ailzhang:
If we set it to 8192 unconditionally it'll increase CUDA memory usage by ~800MB on my 3090. It's not necessary when deep recursively real function call isn't involved so let's default to not set it.

Issue: #

### <samp>🤖 Generated by Copilot at 378709f</samp>

Removed redundant and potentially harmful stack size limit settings for CUDA kernels using the LLVM backend. Added a single stack size limit setting based on user preference or default value in the `ti.init` function. These changes aim to improve the performance and compatibility of Taichi on CUDA devices.

### <samp>🤖 Generated by Copilot at 378709f</samp>

*  Remove redundant and potentially harmful stack size limit settings for every kernel launch on CUDA devices using the LLVM backend ([link](https://github.com/taichi-dev/taichi/pull/7906/files?diff=unified&w=0#diff-d8cf951dea16cc738fc36faee3a158f37a605abccd312c4a6f18d73fd2504332L137-L138))
*  Set the stack size limit only once per device, and only if the user specifies a non-zero value in the `ti.init` function, in the function that defines the Taichi runtime functions on CUDA devices using the LLVM backend ([link](https://github.com/taichi-dev/taichi/pull/7906/files?diff=unified&w=0#diff-b9155792159f392bd8bacd44cb1819be5239b022d707499fc364c0f93dd8c5e5R124-R127))
*  Change the default value of `cuda_stack_limit` in the `CompileConfig` struct from 8192 to 0, to avoid unnecessary stack size limit settings unless the user explicitly requests them ([link](https://github.com/taichi-dev/taichi/pull/7906/files?diff=unified&w=0#diff-604e33db5b7c4f4b819098dee1f2634a36c0974822c1538aec0d57d28ca00dd9L102-R102))

### DIFF
--- a/tests/python/test_function.py
+++ b/tests/python/test_function.py
@@ -214,7 +214,7 @@ def test_different_argument_type():
 
 
 @pytest.mark.run_in_serial
-@test_utils.test(arch=[ti.cpu, ti.cuda], cuda_stack_limit=8092)
+@test_utils.test(arch=[ti.cpu, ti.cuda], cuda_stack_limit=8192)
 def test_recursion():
     @ti.experimental.real_func
     def sum(f: ti.template(), l: ti.i32, r: ti.i32) -> ti.i32:
--- a/tests/python/test_function.py
+++ b/tests/python/test_function.py
@@ -214,7 +214,7 @@ def test_different_argument_type():
 
 
 @pytest.mark.run_in_serial
-@test_utils.test(arch=[ti.cpu, ti.cuda])
+@test_utils.test(arch=[ti.cpu, ti.cuda], cuda_stack_limit=8092)
 def test_recursion():
     @ti.experimental.real_func
     def sum(f: ti.template(), l: ti.i32, r: ti.i32) -> ti.i32:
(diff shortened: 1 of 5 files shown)


lin-hitonami (approved):
LGTM

tests/python/test_function.py
```diff
@@ -214,7 +214,7 @@ def run() -> ti.i32:
 
 
 @pytest.mark.run_in_serial
-@test_utils.test(arch=[ti.cpu, ti.cuda])
+@test_utils.test(arch=[ti.cpu, ti.cuda], cuda_stack_limit=8092)
```
lin-hitonami, line 217: 8192?

ailzhang: lol thanks!

ailzhang, line 217:
```suggestion
@test_utils.test(arch=[ti.cpu, ti.cuda], cuda_stack_limit=8192)
```

tests/python/test_offline_cache.py
```diff
@@ -469,6 +469,7 @@ def added_files():
 @pytest.mark.run_in_serial
 @pytest.mark.parametrize("curr_arch", {ti.cpu, ti.cuda} & supported_archs_offline_cache)
 @_test_offline_cache_dec
+@test_utils.test(cuda_stack_limit=8092)
```
ailzhang, line 472:
```suggestion
@test_utils.test(cuda_stack_limit=8192)
```

PR #7918, opened 2023-05-04
[Lang] Fix the api doc search problem

neozhaoliang:
The docstrings of `inf, nan, pi, e` from the math module are missing, this PR fixes this.

### DIFF
--- a/python/taichi/math/mathimpl.py
+++ b/python/taichi/math/mathimpl.py
@@ -2,7 +2,7 @@
 """
 Math functions for glsl-like functions and other stuff.
 """
-from math import e, inf, nan, pi
+import math
 
 from taichi.lang import impl, ops
 from taichi.lang.impl import static, zero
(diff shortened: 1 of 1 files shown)


ailzhang (changes requested):
nah I don't think this can fix the problem, for example, https://docs.taichi-lang.org/search?category=api&q=vec2 didn't show up in search result but it has a docstring.

neozhaoliang:
I have tested the docstring-gen repo, sphinx correctly generated items for `vec2/3/4` because [they have docstrings](https://github.com/taichi-dev/docstring-gen/blob/master/experimental/_autoapi_templates/python/module.rst?plain=1#L10), you can verify this here: https://docs.taichi-lang.org/api/taichi/math/

The problem should happen on the algolia part.

ailzhang:
@neozhaoliang agh I see thanks! shall we compile a list of all apis that are missing docstrings (thus hidden in doc site) and decide whether we want to add docstrings for them as well?

neozhaoliang:
Sure, opened an issue to record: https://github.com/taichi-dev/taichi/issues/7924

ailzhang (approved):

Issue #8015, opened 2023-05-15
you can try "ti ticache clean" but ticache is not a valid command


Nanase-Nishino:
taichi 1.5 
report a warnning and you can try "ti ticache clean -p {dir}"
but it should be "ti cache clean"

ailzhang:
@Nanase-Nishino Nice catch! would you mind sending a quick fix for it? Thanks!

Nanase-Nishino:
> @Nanase-Nishino Nice catch! would you mind sending a quick fix for it? Thanks!

Of course! I will send a pr about this issue.

PR #8020, opened 2023-05-16
[lang] Update ticache to cache

Nanase-Nishino:
you can run "ti ticache clean -p {dir}"
but ticache is not a valid command

Issue: #8015


ailzhang (approved):
Thanks!

